PR #145, opened 2015-12-03
cli: pull command migrated to docopts

sergiusens:

### DIFF
--- a/snapcraft/commands/pull.py
+++ b/snapcraft/commands/pull.py
@@ -1,0 +1,39 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2015 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+snapcraft pull
+
+Download or retrieve artifacts defined for a part.
+
+Usage:
+  pull [options] [PART ...]
+
+Options:
+  -h --help             show this help message and exit.
+
+"""
+
+from docopt import docopt
+
+from snapcraft import lifecycle
+
+
+def main(argv=None):
+    argv = argv if argv else []
+    args = docopt(__doc__, argv=argv)
+
+    lifecycle.execute('pull', args['PART'])
--- a/snapcraft/lifecycle.py
+++ b/snapcraft/lifecycle.py
@@ -17,14 +17,16 @@
 import contextlib
 import glob
 import importlib
-import jsonschema
 import logging
 import os
 import sys
 import shutil
+
+import jsonschema
 import yaml
 
 import snapcraft
+import snapcraft.yaml
 from snapcraft import common
 from snapcraft import repo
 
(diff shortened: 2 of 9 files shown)


snapcraft/lifecycle.py
```diff
@@ -17,14 +17,18 @@
 import contextlib
 import glob
 import importlib
-import jsonschema
 import logging
 import os
 import sys
 import shutil
+
+# Non-Standard Library modules
```
elopio: dd :)

snapcraft/lifecycle.py
```diff
@@ -17,14 +17,18 @@
...
+import jsonschema
 import yaml
 
+# Snapcraft modules
```
elopio: dd

snapcraft/commands/pull.py
```diff
@@ -0,0 +1,41 @@
...
+
+Usage:
+  pull [options] [PART ...]
+
+Options:
+  -h --help             show this help message and exit.
+
+"""
+
+# Non-Standard Library modules
```
elopio: dd

snapcraft/commands/pull.py
```diff
@@ -0,0 +1,41 @@
...
+from docopt import docopt
+
+# Snapcraft modules
```
elopio: dd

snapcraft/lifecycle.py
```diff
@@ -36,6 +40,21 @@ def _local_plugindir():
...
+
+    if part_names:
+        config.validate_parts(part_names)
+        parts = {p for p in config.all_parts if p.name in part_names}
+    else:
+        parts = config.all_parts
+
+    for part in parts:
+        common.env = config.build_env_for_part(part)
+        part.pull()
```
elopio: I'm confused here. It's doing only pull because you haven't migrated the other commands yet?

snapcraft/tests/test_commands_pull.py
```diff
@@ -0,0 +1,120 @@
...
+                        'Expected a part directory for the pull1 part')
+        self.assertTrue(os.path.exists(parts[1]['state_file']),
+                        'Expected a state file for the pull1 part')
+
+        with open(parts[1]['state_file']) as sf:
+            state = sf.readlines()
+        self.assertEqual(len(state), 1, 'Expected only one line in the state '
+                         'file for the pull1 part')
+        self.assertEqual(state[0], 'pull', 'Expected the state file for pull1 '
+                         'to be \'pull\'')
```
elopio: why don't you use "pull" instead of escaping the quotes?

snapcraft/tests/test_commands_pull.py
```diff
@@ -0,0 +1,120 @@
...
+
+        for i in [0, 2]:
+            self.assertFalse(os.path.exists(parts[i]['part_dir']),
+                             'Expected for only to be a part dir for pull1')
```
elopio: This is hard to read. I'm not sure what are you trying to say.
Maybe something like: "Pulled wrong part."

snapcraft/lifecycle.py
sergiusens: yes and `pull` is always going to be executed anyways; an

```go
if cmds == 'pull':
    part.pull()
```

will always evaluate to `True` when I loop over the commands as it is the first one in the list.

snapcraft/tests/test_commands_pull.py
sergiusens: For message consistency, but I can change it.

snapcraft/tests/test_commands_pull.py
sergiusens: Sure.

snapcraft/lifecycle.py
sergiusens: I added a comment

sergiusens:
Updated with all your comments @elopio
Thanks for the review.

elopio:
:+1: , pending the s/stage/step

PR #148, opened 2015-12-03
clean cmd: migrate to use new lifecycle attribs

sergiusens:
This also migrates clean to use real elements instead of
over mocking.

### DIFF
--- a/snapcraft/commands/clean.py
+++ b/snapcraft/commands/clean.py
@@ -30,7 +30,6 @@ Options:
 import os
 import logging
 import shutil
-import sys
 
 from docopt import docopt
 
@@ -46,12 +45,8 @@ def main(argv=None):
 
     config = snapcraft.yaml.load_config()
 
-    part_names = {part.name for part in config.all_parts}
-    for part_name in args['PART']:
-        if part_name not in part_names:
-            logger.error('The part named {!r} is not defined in '
-                         '\'snapcraft.yaml\''.format(part_name))
-            sys.exit(1)
+    if args['PART']:
+        config.validate_parts(args['PART'])
 
     for part in config.all_parts:
         if not args['PART'] or part.name in args['PART']:
@@ -62,7 +57,9 @@ def main(argv=None):
             not os.listdir(common.get_partsdir())):
         os.rmdir(common.get_partsdir())
 
-    clean_stage = not args['PART'] or part_names == set(args['PART'])
+    parts_match = set(config.part_names) == set(args['PART'])
+    # Only clean stage if all the parts were cleaned up.
+    clean_stage = not args['PART'] or parts_match
     if clean_stage and os.path.exists(common.get_stagedir()):
         logger.info('Cleaning up staging area')
         shutil.rmtree(common.get_stagedir())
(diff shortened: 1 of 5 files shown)


snapcraft/commands/clean.py
```diff
@@ -62,7 +57,8 @@ def main(argv=None):
             not os.listdir(common.get_partsdir())):
         os.rmdir(common.get_partsdir())
 
-    clean_stage = not args['PART'] or part_names == set(args['PART'])
+    parts_match = set(config.part_names) == set(args['PART'])
+    clean_stage = not args['PART'] or parts_match
     if clean_stage and os.path.exists(common.get_stagedir()):
```
elopio: This deserves a comment. "Delete the stage dir only if all the parts were cleaned"

did I get it right?

snapcraft/tests/test_cmds.py
```diff
@@ -31,12 +30,6 @@
 )
 
 
-class _IO(io.StringIO):
-
-    def fileno(self):
-        return 1
```
elopio: this looks crazy. Good delete.

snapcraft/tests/test_commands_clean.py
```diff
@@ -14,168 +14,122 @@
...
+parts:
+{parts}"""
+
+    yaml_part = """  clean{:d}:
+    plugin: nil
+    source: ."""
+
+    def make_snapcraft_yaml(self, n=1, create=True):
+        parts = '\n'.join([self.yaml_part.format(i) for i in range(n)])
+        super().make_snapcraft_yaml(self.yaml_template.format(parts=parts))
```
elopio: calling supper here is weird. Why not just self?

snapcraft/tests/test_commands_clean.py
```diff
@@ -14,168 +14,122 @@
...
-            mock.call(common.get_stagedir()),
-            mock.call(common.get_snapdir()),
-        ])
+        for part in parts:
+            self.assertFalse(
+                os.path.exists(part['part_dir']),
+                'Expected for {!r} to be wiped'.format(part['part_dir']))
+            self.assertFalse(
+                os.path.exists(part['state_file']),
+                'Expected for {!r} to be wiped'.format(part['state_file']))
```
elopio: This might be unnecessary, isn't it enough to check that partsdir doesn't exist?
Feel free to leave the extra checks if you think they might be useful.

snapcraft/tests/test_commands_clean.py
```diff
@@ -14,168 +14,122 @@
...
```
elopio: same comment as above.

elopio:
:100: I like it a lot better without the mocks.
I just left a couple of comments because.

snapcraft/commands/clean.py
sergiusens: That is correct, I'll add a comment.

snapcraft/tests/test_cmds.py
sergiusens: It was important when trying to mock stdout for subprocess, that code was killed, so this is not used anymore.

snapcraft/tests/test_commands_clean.py
sergiusens: True that!

snapcraft/tests/test_commands_clean.py
sergiusens: ditto.

sergiusens:
all good then?

elopio:
all good.

PR #150, opened 2015-12-04
cli: stage command migrated to docopts

sergiusens:
What differs from the previous implementation is that the looping is done on the `steps` first and then on the `parts`, this makes the fact that `after` handling needs fixing right now instead of later as a *part* depending on another needs to have its assets `staged`.


snapcraft/lifecycle.py
```diff
@@ -48,10 +49,24 @@ def execute(step, part_names=None):
...
         parts = config.all_parts
+        part_names = config.part_names
 
     step_index = common.COMMAND_ORDER.index(step) + 1
     for step in common.COMMAND_ORDER[0:step_index]:
+        if step == 'stage':
+            _check_for_collisions(config.all_parts)
         for part in parts:
+            prereqs = config.part_prereqs(part.name)
+            if prereqs and not prereqs.issubset(part_names):
```
elopio: what if you ran the prereqs before?
Like on the test you added, you could have:
lifecycle.execute('stage', part_names=['part1'])
lifecycle.execute('pull', part_names=['part2'])

should that succeed?

sergiusens: Some things in `after` need things to be staged first; since after doesn't state it, I am prefering to default to the safest step for the dependant to work

snapcraft/tests/test_commands_stage.py
```diff
@@ -0,0 +1,149 @@
...
+                        'Expected a state file for the build0 part')
+
+        with open(parts[0]['state_file']) as sf:
+            state = sf.readlines()
+        self.assertEqual(len(state), 1, 'Expected only one line in the state '
+                         'file for the build0 part')
+        self.assertEqual(state[0], 'stage', "Expected the state file for "
+                         "build0 to be 'stage'")
+
+        stage.main()
```
elopio: What is the assertion here?
Maybe we can refactor the step functions to be:
def step(...):
   if not should_step_run(): 
       return
    _step(...)

Put mocks in those _step methods, and check that they were not called.

Or maybe checking that the logger has the skip messages.

snapcraft/tests/test_lifecycle.py
```diff
@@ -300,3 +302,103 @@ def test_clean_part_already_clean(self, mock_exists, mock_rmtree):
...
+        with open(part2.installdir + '/2', mode='w') as f:
+            f.write('1')
+        open(part3.installdir + '/a/2', mode='w').close()
+
+        self.part1 = part1
+        self.part2 = part2
+        self.part3 = part3
+
+    def test_no_collisions(self):
+        lifecycle._check_for_collisions([self.part1, self.part2])
```
elopio: For tests like this, I like to put a comment: In case of error, _check_for_collisions will raise an exception.
Maybe to be extra pedantic, also docstring on _check_for_collisions the exception it raises, and when.

snapcraft/tests/test_lifecycle.py
```diff
@@ -300,3 +302,103 @@ def test_clean_part_already_clean(self, mock_exists, mock_rmtree):
...
+  part1:
+    plugin: nil
+  part2:
+    plugin: nil
+    after:
+      - part1
+""")
+        open('icon.png', 'w').close()
+
+        lifecycle.execute('pull')
```
elopio: You are missing the assertions here.

elopio:
I think this is ok. With snapcraft.yaml.load_config() we prevent the loop to be infinite. But I feel that we are missing tests.
We have tests for check_for_collisions, but we don't have a test that checks that execute raises an exception on collision.
The tests for after in test_yaml.py only have a simple loop check. We need to add at least a loop with three parts.
And I think we don't have nice coverage for should_stage_run, which makes it risky to rely on that for this complex loop.

I asked @fgimenez to take a look too. Also it would be nice for @chipaca to check it, if he's in a good mood today :)

sergiusens:
Ok I fixed up your comments, I can add more complex scenarios for execute.

I didn't add a collisions one as we had separate tests for that, but I don't mind migrating those over here.

sergiusens:
I am checking log message for should stage run now btw

snapcraft/lifecycle.py
```diff
@@ -48,10 +49,27 @@ def execute(step, part_names=None):
...
+                    'Requested {!r} of {!r} but there are unsatisfied '
+                    'prerequisites: {!r}'.format(
+                        step, part.name, ' '.join(prereqs)))
+            elif prereqs:
+                # prerequisites need to build all the way to the staging
+                # step to be able to share the common assets that make them
+                # a dependency.
+                logger.info(
+                    '{!r} has prerequisites that need to be staged: {}'.format(
+                        part.name, ' '.join(prereqs)))
```
sergiusens: @elopio how does this look?

chipaca:
CHIPACA SMASH!!!

chipaca:
Also, :+1:

snapcraft/lifecycle.py
elopio: Looks pretty.

elopio:
From my side, merge when you are happy and we'll add coverage for the other func and exceptions in later branches.

PR #152, opened 2015-12-04
Catkin plugin: Add support for ROS tools.

kyrofa:
The current Catkin plugin distributes only the built version of a ROS package that would typically go into the devel workspace, but that version doesn't include non-source, non-binary files, such as .launch files. This PR changes the plugin to actually install ROS packages, thereby including any non-source, non-binary files specified by the developer.

This PR also updates the wrapper environment to correctly setup the ROS environment, which allows for the use of typical ROS tools within the snapcraft.yaml, such as roslaunch, rosrun, etc., and updates the ROS examples to make use of them.

Finally, this PR also increases the test coverage of the Catkin plugin twofold.

### DIFF
--- a/examples/ros/src/beginner_tutorials/CMakeLists.txt
+++ b/examples/ros/src/beginner_tutorials/CMakeLists.txt
@@ -1,4 +1,3 @@
-# %Tag(FULLTEXT)%
 cmake_minimum_required(VERSION 2.8.3)
 project(beginner_tutorials)
 
@@ -25,7 +24,6 @@ add_executable(listener src/listener.cpp)
 target_link_libraries(listener ${catkin_LIBRARIES})
 
 ## Build service client and server
-# %Tag(SRVCLIENT)%
 add_executable(add_two_ints_server src/add_two_ints_server.cpp)
 target_link_libraries(add_two_ints_server ${catkin_LIBRARIES})
 add_dependencies(add_two_ints_server beginner_tutorials_gencpp)
@@ -34,6 +32,9 @@ add_executable(add_two_ints_client src/add_two_ints_client.cpp)
 target_link_libraries(add_two_ints_client ${catkin_LIBRARIES})
 add_dependencies(add_two_ints_client beginner_tutorials_gencpp)
 
-# %EndTag(SRVCLIENT)%
-
-# %EndTag(FULLTEXT)%
+## Mark executables for installation
+install(TARGETS talker listener add_two_ints_server add_two_ints_client
+  ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+  LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+  RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
+)
--- a/snapcraft/plugins/catkin.py
+++ b/snapcraft/plugins/catkin.py
@@ -32,6 +32,7 @@ import os
 import tempfile
 import logging
 import shutil
+import re
 
 import snapcraft
 from snapcraft import repo
@@ -91,7 +92,6 @@ deb http://${security}.ubuntu.com/${suffix} trusty-security main universe
                     'lib',
                     self.python_version,
                     'dist-packages')),
-            'DESTDIR={0}'.format(self.installdir),
             # ROS needs it but doesn't set it :-/
             'CPPFLAGS="-std=c++11 $CPPFLAGS -I{0} -I{1}"'.format(
                 os.path.join(root, 'usr', 'include', 'c++', self.gcc_version),
@@ -105,6 +105,10 @@ deb http://${security}.ubuntu.com/${suffix} trusty-security main universe
                 root,
                 self.options.rosversion),
             'ROS_MASTER_URI=http://localhost:11311',
+
+            # Various ROS tools (e.g. rospack) keep a cache, and they determine
+            # where the cache goes using $ROS_HOME.
+            'ROS_HOME=$SNAP_APP_USER_DATA_PATH/.ros',
             '_CATKIN_SETUP_DIR={}'.format(os.path.join(
                 root, 'opt', 'ros', self.options.rosversion)),
             'echo FOO=BAR\nif `test -e {0}` ; then\n. {0} ;\nfi\n'.format(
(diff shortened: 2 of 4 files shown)


examples/ros/src/beginner_tutorials/CMakeLists.txt
```diff
@@ -34,6 +32,9 @@ add_executable(add_two_ints_client src/add_two_ints_client.cpp)
...
-# %EndTag(SRVCLIENT)%
-
-# %EndTag(FULLTEXT)%
\ No newline at end of file
+## Mark executables for installation
+install(TARGETS talker listener add_two_ints_server add_two_ints_client
+  ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+  LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+  RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
+)
```
sergiusens: :100:

snapcraft/plugins/catkin.py
```diff
@@ -105,6 +105,7 @@ def env(self, root):
                 root,
                 self.options.rosversion),
             'ROS_MASTER_URI=http://localhost:11311',
```
sergiusens: Not part of the PR, just something to keep in mind.

snapcraft/tests/test_plugin_catkin.py
```diff
@@ -29,16 +30,67 @@ class _IOError(IOError):
...
+                      <package>
+                        <buildtool_depend>buildtool_depend</buildtool_depend>
+                        <build_depend>build_depend</build_depend>
+                        <run_depend>run_depend</run_depend>
+                      </package>"""
+
+        plugin = catkin.CatkinPlugin('test-part', self.properties)
+
+        with mock.patch.object(builtins, 'open',
+                               mock.mock_open(read_data=package_xml)):
```
elopio: Could you remove this mock by writting the xml file to the sourcedir?

snapcraft/plugins/catkin.py
kyrofa: Heh, I don't quite understand what you're trying to draw my attention to. That you'd like that variable removed eventually?

snapcraft/tests/test_plugin_catkin.py
```diff
@@ -29,16 +30,67 @@ class _IOError(IOError):
...
+            plugin.pull()
+
+        self.ubuntu_mock.assert_has_calls([
+            mock.call().get([
+                'ros-foo-buildtool-depend',
+                'ros-foo-build-depend',
+                'ros-foo-run-depend']),
+            mock.call().unpack(plugin.installdir)])
```
elopio: This looks good.

snapcraft/plugins/catkin.py
```diff
@@ -105,6 +105,7 @@ def env(self, root):
...
+            'ROS_HOME=$SNAP_APP_USER_DATA_PATH/.ros',
```
sergiusens: To avoid future misgivings, can we add a comment on why this is needed?

snapcraft/tests/test_plugin_catkin.py
kyrofa: Indeed I could, but I feel like actually hitting the filesystem when I don't need to feels a bit brittle and slow.

snapcraft/plugins/catkin.py
```diff
@@ -245,12 +246,18 @@ def build(self):
 
         self.run(
             ['rm', '-f',
-             'opt/ros/' + self.options.rosversion + '/.catkin',
-             'opt/ros/' + self.options.rosversion + '/.rosinstall',
-             'opt/ros/' + self.options.rosversion + '/setup.sh',
-             'opt/ros/' + self.options.rosversion + '/_setup_util.py'],
+             'opt/ros/' + self.options.rosversion + '/.rosinstall'],
```
sergiusens:
```python
os.remove(os.path.join(self.installdir, self.options.rosversion, '.rosinstall))
```

snapcraft/plugins/catkin.py
```diff
@@ -245,12 +246,18 @@ def build(self):
...
             cwd=self.installdir)
 
+        # Fix the shebang in _setup_util.py to be portable
```
elopio: It would be nice to have a unittest for the shebang.
Aren't we doing this already in the python plugin? Maybe we can share some code, and unit test that.

snapcraft/tests/test_plugin_catkin.py
elopio: In unit tests, I prefer to access the filesystem in temp directories instead of mocking that part.
But your test is correct, if you prefer it this way, I'm ok with that.

snapcraft/plugins/catkin.py
kyrofa: Good call, done.

snapcraft/plugins/catkin.py
sergiusens: That maybe it eventually is expected to run on a different port. But I'm not sure that is configurable.

snapcraft/plugins/catkin.py
kyrofa: Haha, alright I'll clean that up now.

snapcraft/plugins/catkin.py
sergiusens: It's been moved to meta. If we want to share it I'd move it to another module.

snapcraft/plugins/catkin.py
elopio: This deserves an integration test, to check that the files you want end up in the right dirs.
Feel free to add that in a later branch, because that's debt from before.

snapcraft/plugins/catkin.py
kyrofa: I agree, but ALL the build steps here need test coverage, which I plan on adding in a later PR.

snapcraft/plugins/catkin.py
sergiusens: Oh, and if you know why it is being removed, a comment would be great :smile:

snapcraft/plugins/catkin.py
kyrofa: Ah, okay. It is actually configurable, I just need to figure out how to do that with Snappy, which I'll make a different PR.

snapcraft/tests/test_plugin_catkin.py
kyrofa: I don't feel strongly about this. I'll update it :) .

sergiusens:
I like this very much. One followup question is; do we need the `ros` plugin at all? I always found it redundant as maybe the core can be pulled in through some logic here.

kyrofa:
@sergiusens I was wondering that myself. Let me experiment, this PR may have enabled the removal of it.

snapcraft/plugins/catkin.py
kyrofa: @elopio agreed 100%. I'll do it in a later PR though.

ted-gould: It shouldn't need to be removed, but can be filtered now that the filters are fixed. Never got around to replacing the rm with proper stage filters.

kyrofa:
@sergiusens very close. As soon as https://bugs.launchpad.net/snappy/+bug/1466234 is resolved we can remove the roscore plugin.

snapcraft/plugins/catkin.py
kyrofa: I got rid of all the removals here, none of them seemed to be necessary.

snapcraft/tests/test_plugin_catkin.py
kyrofa: Done.

kyrofa:
Alright I believe I've resolved all the concerns for this one. I will continue to increase test coverage throughout subsequent PRs, and will remove the roscore plugin once the [blocking bug](https://bugs.launchpad.net/snappy/+bug/1466234) has been resolved.

elopio:
The example fails to build with:
Error: parts catkin-tutorials and roscore have the following file paths in common which have different contents:

  usr/bin/xml2-config

kyrofa:
@elopio oh interesting, I didn't consider the interplay between plugins! I will add that to my manual test procedure. xml2-config isn't a required file, so I have temporarily added the removal of it back to the catkin plugin until we can remove roscore.

sergiusens:
@elopio and @kyrofa fwiw, plugins aren't supposed to solve all the conflicts in the world, that is to some extent part of the part author's responsibility. That does no rule out plugins themselves should try to deliver the cleanest environment possible, but not in detriment of its functionality.

In the example of `xml-config`, if I as an author ever create another part that also ships `xml-config` I would need to solve if somehow, luckily, that is somewhat what the `stage`, `snap` and `organize` keywords are for.

sergiusens:
I'm :+1: on this branch btw :wink:

elopio:
Me too. Merging.

PR #155, opened 2015-12-04
Making the icon optional

sergiusens:
The store should provide an icon if it is missing.


elopio:
phew, are we done?
:100:

PR #158, opened 2015-12-07
Resolve xml2-config FIXME in Catkin plugin

kyrofa:
This simply moves the hack into the roscore plugin, which already has more of them anyway.


elopio:
okay. :+1:

PR #159, opened 2015-12-07
Add a tutorial for turning a ROS project into a .snap.

kyrofa:


docs/ros-snap.md
```diff
@@ -0,0 +1,354 @@
...
+  the update goes sideways?
+
+These are all problems solved by Snappy; the combination of the two is a perfect
+match for consumer robotic systems.
+
+
+## Package your current ROS project as a .snap
+
+Let's assume you already have a ROS project. It can be as simple or as
+complicated as you like, but for this example, our project will be made up of:
```
elopio: would it be good to mention the project in the examples dir as a reference?

docs/ros-snap.md
```diff
@@ -0,0 +1,354 @@
...
+    ...
+
+### Put it all in a .snap
+
+Create a `snapcraft.yaml` in the workspace root, containing the following:
+
+```yaml
+name: ros-talker-and-listener
+version: 1.0
+vendor: example <example@example.com>
```
elopio: no vendor anymore.

docs/ros-snap.md
```diff
@@ -0,0 +1,354 @@
...
+summary: ROS Example
+description: Contains talker/listener ROS packages and a .launch file.
+icon: icon.svg
```
elopio: icon is optional, so I would leave it out of the examples, for brevity.

docs/ros-snap.md
```diff
@@ -0,0 +1,354 @@
...
+    exec: roslaunch listener talk_and_listen.launch
+# ...
+```
+
+Even though the `talker` and `listener` packages will be installed, they'll be
+within a confined ROS installation, so the user won't be able to simply call
+`rosrun` or `roslaunch`. Instead, you have control over how your .snap is used,
+and here we specify that we only want a single binary, called "launch_project",
+which results in the `roslaunch` call you see. If this seems confusing now, it
+might make more sense when we actually use it.
```
elopio: nitpicking here, I think you should mention that we'll give it a try later in the document.
"it WILL make more sense later in this document when we actually use it."

docs/ros-snap.md
```diff
@@ -0,0 +1,354 @@
...
+
+    $ sudo snappy install ros-talker-and-listener_1.0_amd64.snap \
+      --allow-unauthenticated
+
+Take a look in `/apps/bin/`, and you'll see the binary you requested, called
+`ros-talker-and-listener.launch_project`. Test it out:
+
+    $ ros-talker-and-listener.launch_project
+
+And you should see the talker and listener communicating like before.
```
elopio: How to stop them?

elopio:
Thanks @kyrofa, this is great.
I'd like @dholbach to take a look, so we make sure we are consistent with the style in his docs.

docs/ros-snap.md
kyrofa: Agreed, good point.

docs/ros-snap.md
kyrofa: Ah, right.

docs/ros-snap.md
kyrofa: Removed, thank you.

docs/ros-snap.md
kyrofa: I actually plan on replacing that example with a more realistic one when I have a better solution for roscore.

docs/ros-snap.md
kyrofa: ctrl+c, I updated it to reflect, thank you :) .

docs/ros-snap.md
```diff
@@ -0,0 +1,355 @@
...
+    [ INFO] [1449512660.415941529]: Hello world 3
+    [INFO] [WallTime: 1449512660.416330] I heard Hello world 3
+    [ INFO] [1449512660.515954882]: Hello world 4
+    [INFO] [WallTime: 1449512660.516307] I heard Hello world 4
+    [ INFO] [1449512660.615954473]: Hello world 5
+    [INFO] [WallTime: 1449512660.616306] I heard Hello world 5
+    ...
+
+### Put it all in a .snap
+
```
dholbach: Maybe add half a sentence about what `snapcraft.yaml` is, what snapcraft aims to do - just so people who get directed to this specific article have a little bit of context.

docs/ros-snap.md
```diff
@@ -0,0 +1,355 @@
...
+
+### Take the .snap for a test drive
+
+Transfer the newly-created .snap to your Ubuntu Core machine or instance, and
+install it. For example:
+
+    $ sudo snappy install ros-talker-and-listener_1.0_amd64.snap \
+      --allow-unauthenticated
+
```
dholbach: Is `snappy-remote` maybe easier to use or explain?

docs/ros-snap.md
kyrofa: Done. Does that look better?

docs/ros-snap.md
kyrofa: Probably. I've updated the document to use `snappy-remote`. Does that seem simpler?

docs/ros-snap.md
dholbach: +1 - great work!

elopio:
I think the last comment by @dholbach means he's happy, so let's merge.

PR #160, opened 2015-12-07
cli: snap command migrated to docopts

sergiusens:


snapcraft/commands/snap.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def main(argv=None):
+    argv = argv if argv else []
+    args = docopt(__doc__, argv=argv)
+
+    if args['DIRECTORY']:
+        # TODO: needs design
+        # TODO: overly simplistic, no meta/snap.yaml check
+        snap_dir = os.path.abspath(args['DIRECTORY'])
```
elopio: This needs an integration test.

snapcraft/commands/snap.py
```diff
@@ -0,0 +1,77 @@
...
+        snap = _snap_data_from_dir(snap_dir)
+    else:
+        # make sure the full lifecycle is executed
+        snap_dir = common.get_snapdir()
+        snap = lifecycle.execute('strip')
+
+    snap['arch'] = snap['arch'][0] if len(snap['arch']) == 1 else 'multi'
+    snap_name = '{name}_{version}_{arch}.snap'.format(**snap)
```
elopio: I would wrap these two statements in an internal function.

snapcraft/commands/snap.py
```diff
@@ -0,0 +1,77 @@
...
+
+    logger.info('Snapping {}'.format(snap_name))
+    subprocess.check_call(
+        ['mksquashfs', snap_dir, snap_name, '-noappend', '-comp', 'xz'])
```
elopio: a new deb dependency?

snapcraft/lifecycle.py
```diff
@@ -80,6 +80,10 @@ def execute(step, part_names=None):
         if step == 'strip' and part_names == config.part_names:
             meta.create(config.data)
 
+    return {'name': config.data['name'],
+            'version': config.data['version'],
+            'arch': config.data['architectures']}
+
```
elopio: gentle reminders about the docstring ;)

elopio:
lgtm. Minor comments, and I would like to see an integration test for snapping a dir.

snapcraft/commands/snap.py
sergiusens: @elopio The integration tests will come in the next PR where I will enable them all again. That `TODO` is about `snap.yaml` is outdated though.

snapcraft/commands/snap.py
sergiusens: got it.

snapcraft/commands/snap.py
sergiusens: yes, and effectively moving away from the requirement to have `ubuntu-snappy-cli` :smile:

snapcraft/lifecycle.py
sergiusens: Ah, this is the right time to do it!

sergiusens:
@elopio great, I'll write the `TODO` to get an integration test, but won't do it in this PR as integration tests as a whole aren't working here yet. It is just after this PR that I will overcome that issue

snapcraft/lifecycle.py
```diff
@@ -44,6 +44,23 @@ def _local_plugindir():
...
+    Lifecycle execution will happen for each step iterating over all
+    the available parts, if part_names is specified, only those parts
+    will run.
+
+    If one of the parts to execute has an after keyword, execution is
+    forced until the stage step for such part. If part_names was provided
+    and after is not in this set, an exception will be raised.
+
+    :param str step: A valid step in the lifecycle: pull, build, strip or snap.
+    :raises RuntimeError: If an prerequesite of the part needs to be staged
```
elopio: typo: an -> a

elopio:
Thanks. :+1:

snapcraft/commands/snap.py
```diff
@@ -0,0 +1,81 @@
...
+    else:
+        # make sure the full lifecycle is executed
+        snap_dir = common.get_snapdir()
+        snap = lifecycle.execute('strip')
+
+    snap_name = _format_snap_name(snap)
+
+    logger.info('Snapping {}'.format(snap_name))
+    subprocess.check_call(
+        ['mksquashfs', snap_dir, snap_name, '-noappend', '-comp', 'xz'])
```
kyrofa: If `mksquashfs` exits with non-zero, this throws exceptions, right? Would it be more user-friendly to catch those here and fail gracefully, or are you intending to provide a stack trace?

sergiusens: @kyrofa nothing returns a stack trace by default. I learned my lesson and there is a catch all in our main execution entry point to take care of this; no matter what (I intend to add a `--debug` mode so such stack trace can be seen).

sergiusens: We can eventually catch here and throw something nice; I also don't like exposing `squashfs` to the build console.

PR #161, opened 2015-12-09
Go plugin creating invalid env

jocave:
Possible fix for this bug: https://bugs.launchpad.net/snapcraft/+bug/1524374

### DIFF
--- a/snapcraft/plugins/go.py
+++ b/snapcraft/plugins/go.py
@@ -70,7 +70,7 @@ class GoPlugin(snapcraft.BasePlugin):
         # usr/lib/go/bin on newer Ubuntus, usr/bin on trusty
         env = [
             'GOPATH={}/go'.format(root),
-            'CGO_LDFLAGS=$CGO_LDFLAGS"' + ' '.join([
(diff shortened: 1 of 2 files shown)


elopio:
This makes sense to me, but I would feel safer with a regression test. Have you been able to reduce the issue to a minimum scenario?

cwayne18:
@elopio We haven't been able to reduce it to a minimum scenario quite yet (so far we're only seeing it with building our snap that we'd rather share in public repos yet).  Will this block us from landing this? It'd be great to not need a fork of snapcraft to build our snap :)

elopio:
If this is urgent for you, sure, we can land it. But if it can wait, I would prefer to fully understand the issue before making a fix, by writing a regression test.
Can you share the snapcraft with me?

elopio:
Merging. Now I see this is a clear fix for the issue and the test updated ensures the space is in there. 
I'll ask Sergio next week why we are sharing a single env. The error happens when wrapping the python3 executable, and CGO_LDFLAGS are not needed in there. I'll keep digging.
Thanks for the report, and thanks for the fix!

kyrofa:
This needs to be backported to 1.x.

PR #162, opened 2015-12-09
Wrapping exe should not be over match to replace

sergiusens:
The regex used in wrap_exe was over doing it and globbing more than
necessary for replacement. This makes the match not consider `.*`
but instead uses the match used for valid part names and considers `/`.

Fixes LP: #1523912


elopio:
ok, I did some tests. :+1:

PR #163, opened 2015-12-11
Catkin plugin: use rosdep for dependency resolution

kyrofa:
The Catkin plugin currently does its own dependency resolution via manually parsing each package.xml. It makes some naive assumptions while doing so, and a tool already exists in ROS for doing this: rosdep. This PR gets rid of the manual parsing in favor of rosdep.

### DIFF
--- a/snapcraft/plugins/catkin.py
+++ b/snapcraft/plugins/catkin.py
@@ -27,12 +27,12 @@ Additionally, this plugin uses the following plugin-specific keywords:
       List of catkin packages to build.
 """
 
-import lxml.etree
 import os
 import tempfile
 import logging
 import shutil
 import re
+import subprocess
 
 import snapcraft
 from snapcraft import repo
@@ -53,7 +53,7 @@ deb http://${security}.ubuntu.com/${suffix} trusty-security main universe
     @classmethod
     def schema(cls):
         schema = super().schema()
-        schema['properties']['rosversion'] = {
+        schema['properties']['rosdistro'] = {
             'type': 'string',
             'default': 'indigo'
         }
@@ -73,15 +73,33 @@ deb http://${security}.ubuntu.com/${suffix} trusty-security main universe
 
     def __init__(self, name, options):
         super().__init__(name, options)
+
         self.packages = set(options.catkin_packages)
-        self.dependencies = ['ros-core']
         self.package_deps_found = False
         self.package_local_deps = {}
         self._deb_packages = []
+        self.dependencies = []
 
-    def pull(self):
-        super().pull()
-        self._setup_deb_packages()
+        source_subdir = self.options.source_subdir
+        if not source_subdir:
+            source_subdir = 'src'
+
+        self._ros_package_path = os.path.join(self.sourcedir, source_subdir)
+
+        # Ensure the `source` key is set to the root of a Catkin workspace,
+        # where the Catkin source space can be remapped via the source-subdir
+        # key. We're this strict because `source` CANNOT be set to the Catkin
+        # source space (likewise source_subdir cannot be set to the root of a
+        # Catkin workspace), as it will result in a circular symlink upon
+        # pull() which will break rosdep.
+        if os.path.abspath(self.sourcedir) == os.path.abspath(
+                self._ros_package_path):
+            raise RuntimeError(
+                'source-subdir cannot be the root of the Catkin workspace')
+
+        self._rosdep = _Rosdep(self.options.rosdistro, self._ros_package_path,
+                               os.path.join(self.partdir, 'rosdep'),
+                               self.PLUGIN_STAGE_SOURCES)
 
     def env(self, root):
         return [
@@ -103,18 +121,30 @@ deb http://${security}.ubuntu.com/${suffix} trusty-security main universe
                              self.gcc_version)),
             'LD_LIBRARY_PATH=$LD_LIBRARY_PATH:{0}/opt/ros/{1}/lib'.format(
                 root,
-                self.options.rosversion),
+                self.options.rosdistro),
             'ROS_MASTER_URI=http://localhost:11311',
 
             # Various ROS tools (e.g. rospack) keep a cache, and they determine
             # where the cache goes using $ROS_HOME.
             'ROS_HOME=$SNAP_APP_USER_DATA_PATH/.ros',
             '_CATKIN_SETUP_DIR={}'.format(os.path.join(
-                root, 'opt', 'ros', self.options.rosversion)),
+                root, 'opt', 'ros', self.options.rosdistro)),
             'echo FOO=BAR\nif `test -e {0}` ; then\n. {0} ;\nfi\n'.format(
                 os.path.join(
-                    root, 'opt', 'ros', self.options.rosversion, 'setup.sh'))
+                    root, 'opt', 'ros', self.options.rosdistro, 'setup.sh'))
         ]
+
+    def pull(self):
+        super().pull()
+
+        # Make sure the package path exists before continuing
+        if not os.path.exists(self._ros_package_path):
+            raise FileNotFoundError(
+                'Unable to find package path: "{}"'.format(
+                    self._ros_package_path))
+
+        self._rosdep.setup()
+        self._setup_deb_packages()
 
     @property
     def python_version(self):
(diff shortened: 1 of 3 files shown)


snapcraft/plugins/catkin.py
```diff
@@ -73,15 +73,36 @@ def schema(cls):
...
-        self.dependencies = ['ros-core']
         self.package_deps_found = False
         self.package_local_deps = {}
         self._deb_packages = []
+        self.dependencies = []
 
-    def pull(self):
-        super().pull()
-        self._setup_deb_packages()
+        source_subdir = getattr(self.options, 'source_subdir', None)
```
sergiusens: this should always be defined in the schema (as in it will only fail due to a programmatic error) so there is no need to `getattr`

snapcraft/plugins/catkin.py
```diff
@@ -127,77 +160,59 @@ def gcc_version(self):
...
-            sourcedir = os.path.join(self.sourcedir, source_subdir)
-        else:
-            sourcedir = self.sourcedir
+        if not system_dependency:
+            raise RuntimeError(
+                'Package "{}" isn\'t a valid system dependency. '
+                'Did you forget to add it to catkin-packages? If '
+                'not, add the Ubuntu package containing it to '
+                'stage-packages until you can get it into the '
+                'rosdep database.'.format(dependency))
```
sergiusens: nice error message :smile:

sergiusens:
*to be continued*

snapcraft/plugins/catkin.py
kyrofa: Ah, I probably stole it from a constructor somewhere, where it wasn't necessarily defined.

kyrofa:
What?! No! I added 30% coverage to the Catkin plugin... how did coverage drop?! I was looking forward to seeing how high it would go :( .

kyrofa:
Oh... plainbox and examples dropped somehow.

snapcraft/plugins/catkin.py
```diff
@@ -127,77 +160,59 @@ def gcc_version(self):
...
     def rosdir(self):
         return os.path.join(self.installdir,
-                            'opt', 'ros', self.options.rosversion)
+                            'opt', 'ros', self.options.rosdistro)
 
-    def _deps_from_packagesxml(self, f, pkg):
-        tree = lxml.etree.parse(f)
+    def _find_dependencies(self):
+        if self.package_deps_found:
+            return
```
elopio: It would be good to explain this in a comment, because it's weird. What are the cases that makes us call find_dependencies twice?

snapcraft/plugins/catkin.py
```diff
@@ -127,77 +160,59 @@ def gcc_version(self):
     @property
     def rosdir(self):
         return os.path.join(self.installdir,
-                            'opt', 'ros', self.options.rosversion)
+                            'opt', 'ros', self.options.rosdistro)
 
-    def _deps_from_packagesxml(self, f, pkg):
-        tree = lxml.etree.parse(f)
+    def _find_dependencies(self):
```
elopio: I would find this clearer if it is a get function that returns the dependencies, instead of a method with no return value that has a side effect of changing the dependencies attribute. Would that be hard to change?

snapcraft/plugins/catkin.py
kyrofa: Great question: I'm not sure (that was already in there). However, I *believe* it's because it's called from both `pull()` and `build()` (since both steps need to know it), but it doesn't want to find deps twice in case it's being run via `snapcraft` instead of `snapcraft <step>`.

snapcraft/plugins/catkin.py
kyrofa: I don't believe so, although not saving it as an attribute would mean I'd have to interrogate the sources for their dependencies in both `pull()` and `build()`. Depending on the size of the workspace that could be a bit heavy.

snapcraft/plugins/catkin.py
```diff
@@ -127,77 +160,59 @@ def gcc_version(self):
...
-                        dep.replace('_', '-') in self._deb_packages):
-                    continue
+        # Query rosdep for the list of dependencies for this package
+        dependencies = self._rosdep.get_dependencies(package)
 
-                # Get the ROS package for it
-                self._deb_packages.append(
-                    'ros-'+self.options.rosversion+'-'+dep.replace('_', '-'))
+        for dependency in dependencies:
+            if not dependency:
```
elopio: Why would get_dependencies return an invalid dependency? It seems better to make sure that it returns a clean list of valid deps than to check them on the caller.

kyrofa: Yeah you're right, this should be checked in `get_dependencies()`. And to answer your question, it shouldn't, but since I'm parsing stdout I'm trying to account for small changes in the output of `rosdep`.

snapcraft/plugins/catkin.py
kyrofa: I might be able to refactor the build step to not need these though... I'll get back to you.

snapcraft/plugins/catkin.py
```diff
@@ -312,3 +326,102 @@ def _handle_package(self, pkg):
...
+        # necessarily known at pull-time-- we need to interrogate the Catkin
+        # packages first. Then we pull them down ourselves.
+        ubuntu = repo.Ubuntu(self._rosdep_path, sources=self._ubuntu_sources)
+
+        # rosdep is used for analyzing the dependencies specified within each
+        # project's package.xml. But it introduces a complication, in that it's
+        # not a dependency of the project, and we don't want to bloat the .snap
+        # more than necessary. So we'll unpack it somewhere else, and use it
+        # from there.
+        logger.info("Fetching rosdep...")
```
elopio: single quotes. I'm sorry, it's an obsession of mine :)

snapcraft/plugins/catkin.py
```diff
@@ -312,3 +326,102 @@ def _handle_package(self, pkg):
...
+        ubuntu.get(['python-rosdep'])
+        ubuntu.unpack(self._rosdep_install_path)
+
+        logger.info("Initializing rosdep database...")
```
elopio: single quotes.

snapcraft/plugins/catkin.py
```diff
@@ -312,3 +326,102 @@ def _handle_package(self, pkg):
...
+        try:
+            self._run(['init'])
+        except subprocess.CalledProcessError as e:
+            raise RuntimeError(
+                'Error initializing rosdep database: {}'.format(e.output))
+
+        logger.info("Updating rosdep database...")
```
elopio: single quotes.

snapcraft/plugins/catkin.py
kyrofa: Ah, no thanks! I've been trying to adhere but I knew I was going to miss some, haha :) .

snapcraft/plugins/catkin.py
elopio: ok. And if you need the sideeffect for efficiency, maybe split it in two: one _get, and one _update that calls the _get. Or if it makes a bigger mess, just add a docstring to what you have explaining the side effect.

snapcraft/plugins/catkin.py
kyrofa: This will be removed completely  in the refactor coming after this PR, so I won't change it here.

snapcraft/plugins/catkin.py
kyrofa: I'm going to fix this in the next PR.

snapcraft/plugins/catkin.py
kyrofa: I'm going to fix this in the next PR.

snapcraft/plugins/catkin.py
kyrofa: Fixed.

kyrofa:
Unless there are more comments, I've either fixed the concerns here or will address them in a PR I'll make right after this one is merged.

snapcraft/plugins/catkin.py
kyrofa: It doesn't seem to be defined here in `__init__()` (particularly when unit testing), so `getattr` seems necessary unless I should make sure all tests explicitly pass it in.

sergiusens: Right, the `options` class used in the tests needs to have the attrib. If testing from the plugin handler itself, this should be created with the defaults.

kyrofa: Ah, alright done. And you're right, it works fine :) .

snapcraft/tests/test_plugin_catkin.py
```diff
@@ -183,3 +331,136 @@ class Options:
...
+        class check_env():
+            def __eq__(self, env):
+                return (
+                    env['PATH'] == os.path.join(rosdep._rosdep_install_path,
+                                                'usr', 'bin') and
+                    env['PYTHONPATH'] == os.path.join(
+                        rosdep._rosdep_install_path, 'usr', 'lib', 'python2.7',
+                        'dist-packages') and
+                    env['ROSDEP_SOURCE_PATH'] == rosdep._rosdep_sources_path
+                    and env['ROS_HOME'] == rosdep._rosdep_cache_path and
```
elopio: the and operator should be at the end of the line. flake8 complaints in xenial, it seems to ignore it in the travis's version.

snapcraft/tests/test_plugin_catkin.py
```diff
@@ -44,107 +45,255 @@ class props:
...
 
-        self.assertTrue('my_package' in plugin.package_local_deps,
-                        'Expected "my_package" to be in the dependencies')
-        self.assertEqual(plugin.package_local_deps['my_package'],
-                         {'package_2'},
-                         'Expected "my_package" to depend upon "package_2"')
-        self.assertFalse(self.ubuntu_mock.called,
-                         "Ubuntu packages were unexpectedly pulled down")
+        # Verify that rosdep was setup
+        self.rosdep_mock.assert_has_calls([
```
elopio: This could be a checker method, you are using it twice.

snapcraft/tests/test_plugin_catkin.py
```diff
@@ -44,107 +45,255 @@ class props:
...
+    def test_valid_catkin_workspace_src(self):
+        # sourcedir is expected to be the root of the Catkin workspace. Since
+        # it contains a 'src' directory, this is a valid Catkin workspace.
+        try:
+            plugin = catkin.CatkinPlugin('test-part', self.properties)
+            os.makedirs(os.path.join(plugin.sourcedir, 'src'))
+            plugin.pull()
+        except FileNotFoundError:
+            self.fail('Unexpectedly raised an exception when the Catkin '
+                      'workspace was valid')
```
elopio: I generally just leave the exception make the test fail, with a comment above the step that can raise the exception. The exception should say something useful already to understand what went wrong, otherwise the exception is not very good.
If you prefer the try catch, that's good too. Just not common in the projects I've worked.

snapcraft/tests/test_plugin_catkin.py
```diff
@@ -44,107 +45,255 @@ class props:
...
-        plugin = catkin.CatkinPlugin('test-part', self.properties)
+            return ''
+
+        self.rosdep_mock.return_value.get_dependencies.side_effect = get
+        self.rosdep_mock.return_value.resolve_dependency.return_value = None
+
+        with self.assertRaises(RuntimeError) as raised:
+            plugin.pull()
 
+        self.assertEqual(raised.exception.args[0],
```
elopio: I think it's better to check
str(raised.exception)
instead of
raises.exception.args[0]

snapcraft/tests/test_plugin_catkin.py
```diff
@@ -44,107 +45,255 @@ class props:
...
+    def test_invalid_catkin_workspace_no_src(self):
+        # sourcedir is expected to be the root of the Catkin workspace. Since
+        # it does not contain a `src` folder and `source-subdir` wasn't
+        # specified, this should fail.
+        with self.assertRaises(FileNotFoundError) as raised:
+            plugin = catkin.CatkinPlugin('test-part', self.properties)
+            plugin.pull()
+
+        self.assertEqual(
+            raised.exception.args[0],
```
elopio: same str(raised.exception)

snapcraft/tests/test_plugin_catkin.py
```diff
@@ -183,3 +331,136 @@ class Options:
...
+
+        self.rosdep = catkin._Rosdep('ros_distro', 'package_path',
+                                     'rosdep_path', 'sources')
+
+        patcher = mock.patch('snapcraft.repo.Ubuntu')
+        self.ubuntu_mock = patcher.start()
+        self.addCleanup(patcher.stop)
+
+        patcher = mock.patch('subprocess.check_output')
+        self.run_mock = patcher.start()
```
elopio: I would understand the tests easier if the mock is called check_output_mock.

elopio:
@kyrofa sorry it took so long.
As I told you, I'm :+1: here. I just left some PITA comments about the tests, that you can safely ignore.

snapcraft/tests/test_plugin_catkin.py
kyrofa: Normally I agree with you, but I'm explicitly making sure this code doesn't die if I know it can't find the file. I can make the failure message much more friendly than trying to navigate a traceback. There's also something to be said for making the tests **fail** if what I'm testing for doesn't pass rather than **error**. If an exception was raised as a result of some setup code then I feel like that's a valid error, but if it's exactly what I'm testing for I prefer to catch it and turn it into a failure.

snapcraft/tests/test_plugin_catkin.py
kyrofa: \*face-palm\* so much better! Thank you :)

sergiusens: Do you prefer `str(raised.exception)` over `raised.exception.__str__()` ?
Just asking since other parts of the code base use the latter in tests.

snapcraft/tests/test_plugin_catkin.py
kyrofa: Ah interesting, okay. Yeah, it passes on trusty.

kyrofa: Fixed.

snapcraft/tests/test_plugin_catkin.py
kyrofa: Fixed.

snapcraft/tests/test_plugin_catkin.py
kyrofa: Fixed.

snapcraft/tests/test_plugin_catkin.py
kyrofa: Agreed, thank you! Copy-paste error, I think.

kyrofa:
@elopio excellent review, thank you! PITA comments are appreciated ;) . We'll just wait for @sergiusens to give his OK here, then.

snapcraft/tests/test_plugin_catkin.py
kyrofa: Ah, hey @sergiusens I missed your comment here. I feel like the former is better-looking, but I'm happy doing either.

snapcraft/plugins/catkin.py
```diff
@@ -73,15 +73,33 @@ def schema(cls):
 
     def __init__(self, name, options):
         super().__init__(name, options)
+
         self.packages = set(options.catkin_packages)
-        self.dependencies = ['ros-core']
         self.package_deps_found = False
         self.package_local_deps = {}
         self._deb_packages = []
+        self.dependencies = []
```
sergiusens: is order important or can this be a set?

kyrofa: It could indeed be a set, but it's completely removed in the next PR.

sergiusens:
Just one comment related to names; it is mostly a nitpick and from being away from the problem, but I'd like to see member names to be `self.catkin_packages` instead of `self.packages` and similar for the `system` counterparts; this also applies to the names the methods have.

Preemptive :+1: take the comment or leave it :wink:

kyrofa:
@sergiusens yeah, good point. There's a bit of technical debt to be paid there, but I'll take a look at that in the refactor coming up.

PR #164, opened 2015-12-14
Updated the autopkgtests.

elopio:
On autopkgtests, run the integration tests, build the examples but do not install them, do not run the unittests.

The unittests are run when the package is built, so no need to run them again during package update.
The examples install tests will run only daily, and on PRs.


kyrofa:
Looks good to me.

PR #166, opened 2015-12-16
Translated the plainbox tests to python.

elopio:
There are still a couple of things that I don't understand, and plenty of things that I would like to clean up and split. But in this PR I just focused on a literal translation, to start improving the tests in future PRs.
In particular it's bad that the generated binaries can't be executed in trusty. On the plainbox execution we were ignoring the exit code, so for now I'm just doing the same in python. Marked with a TODO to dig further on the cause.


elopio:
\o/ greeeeen!

kyrofa:
Can this be closed, then?

PR #169, opened 2015-12-16
Python{2,3} plugins: Make site-packages link relative.

kyrofa:
Currently the Python2 and Python3 plugins create a `<install>/usr/lib/<pythonversion>/site-packages` symlink to `<install>/usr/lib/<pythonversion>/dist-packages`, but that absolute link doesn't exist when the .snap is installed. This commit makes the link relative instead of absolute so the link stays valid.

This may be worth cherry-picking to 1.x.

Fixes LP: [#1523384](https://bugs.launchpad.net/snapcraft/+bug/1523384)

### DIFF
--- a/snapcraft/plugins/python2.py
+++ b/snapcraft/plugins/python2.py
@@ -97,11 +97,11 @@ class Python2Plugin(snapcraft.BasePlugin):
         site_packages_dir = os.path.join(
             prefix, 'lib', self.python_version, 'site-packages')
 
+        # If site-packages doesn't exist, make sure it points to the
+        # dist-packages in the same directory (this is a relative link so that
+        # it's still valid when the .snap is installed).
         if not os.path.exists(site_packages_dir):
-            os.symlink(
-                os.path.join(prefix, 'lib', self.python_version,
-                             'dist-packages'),
-                site_packages_dir)
+            os.symlink('dist-packages', site_packages_dir)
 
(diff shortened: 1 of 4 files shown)


snapcraft/plugins/python2.py
```diff
@@ -97,11 +97,11 @@ def _pip(self):
...
 
+        # If site-packages doesn't exist, make sure it points to the
+        # dist-packages in the same directory (this is a relative link so that
+        # it's still valid when the .snap is installed).
         if not os.path.exists(site_packages_dir):
-            os.symlink(
-                os.path.join(prefix, 'lib', self.python_version,
-                             'dist-packages'),
-                site_packages_dir)
+            os.symlink('dist-packages', site_packages_dir)
```
elopio: Should we do the same for python3?

elopio:
I think we should have an integration test for this too, simplifying the snapcraft that was mentioned in the bug.

kyrofa:
@elopio yeah good idea, both on the test and on waiting, heh. The new bug: https://bugs.launchpad.net/snapcraft/+bug/1527270

snapcraft/plugins/python2.py
kyrofa: Ah, good call, I didn't even look at python3. Fixing now.

kyrofa: Done.

sergiusens:
:+1:

PR #174, opened 2015-12-18
New spec implementation for the cli

sergiusens:
This implements the move to the new spec'ed cli, also using docopts instead of argparse.
The integration tests are now also python based instead of using plainbox


elopio:
:+1: Merge!
A huge change right at the end of the year, what could possibly go wrong? ^_^.
Next year we can focus on making the coverage more meaningful. For now it's good enough.
I'll let @kyrofa hit the Merge button, so the three of us get an equal share of responsibility when things terribly break ;)

kyrofa:
Haha, @elopio you suck. Merging!

PR #175, opened 2015-12-18
Catkin plugin: Refactor build.

kyrofa:
This PR contains a number of small refactors:

- Change `env()` to only contain variables for run-time, and document each of them.
- Only obtain dependencies upon `pull()`-- for `build()` they should just be ready to go.
- Extract dependency resolution into a standalone function that can be tested in isolation.
- Change `build()` to be less clever and more simple/verbose/robust.
- Add `source-space` keyword to get rid of custom `build()`.
- Get unit testing coverage of the Catkin plugin to 100%.

### DIFF
--- a/snapcraft/plugins/catkin.py
+++ b/snapcraft/plugins/catkin.py
@@ -25,6 +25,9 @@ Additionally, this plugin uses the following plugin-specific keywords:
     - catkin-packages:
       (list of strings)
       List of catkin packages to build.
+    - source-space:
+      (string)
+      The source space containing Catkin packages. By default this is 'src'.
 """
 
 import os
@@ -66,6 +69,10 @@ deb http://${security}.ubuntu.com/${suffix} trusty-security main universe
             },
             'default': [],
         }
+        schema['properties']['source-space'] = {
+            'type': 'string',
+            'default': 'src',
+        }
 
         schema['required'].append('catkin-packages')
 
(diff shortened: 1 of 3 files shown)


snapcraft/plugins/catkin.py
```diff
@@ -25,6 +25,9 @@
     - catkin-packages:
       (list of strings)
       List of catkin packages to build.
+    - source-space:
```
sergiusens: Is `source-space` the word used upstream, if not, it feels weird. Maybe just [`workspace`](http://wiki.ros.org/catkin/Tutorials/create_a_workspace)?

snapcraft/plugins/catkin.py
```diff
@@ -66,75 +69,88 @@ def schema(cls):
...
-        self.packages = set(options.catkin_packages)
-        self.package_deps_found = False
-        self.package_local_deps = {}
-        self._deb_packages = []
-        self.dependencies = []
+        Specifically, the path created via the `source` key (or a combination
+        of `source` and `source-subdir` keys) needs to point to a valid Catkin
+        workspace containing another subdirectory called the "source space." By
+        default, this is a directory named "src," but it can be remapped via
+        the `source-space` key. It's important that the source space is not the
```
sergiusens: just a pointer for change if the above schema is changed

snapcraft/plugins/catkin.py
kyrofa: Yeah, line 262. Workspace would be confusing since that's what `source`/`source-subdir` should be pointing to (the catkin workspace).

snapcraft/plugins/catkin.py
```diff
@@ -66,75 +69,88 @@ def schema(cls):
...
+            'default': 'src',
+        }
 
         schema['required'].append('catkin-packages')
 
         return schema
 
     def __init__(self, name, options):
-        super().__init__(name, options)
+        """Sanity-check the plugin config regarding source code paths.
```
sergiusens: [`__init__`](http://www.rafekettler.com/magicmethods.html#construction) is much more than this. I am not entirely sure this is the best one liner docstring

snapcraft/plugins/catkin.py
```diff
@@ -66,75 +69,88 @@ def schema(cls):
...
                 os.path.join(
                     root, 'opt', 'ros', self.options.rosdistro, 'setup.sh'))
         ]
 
     def pull(self):
+        """Copy source into build directory and fetch dependencies.
+
+        Catkin packages can specify their system dependencies in their
+        package.xml. In order to support that, the Catkin packages are
+        interrogated for their dependencies here. Since "stage-packages" are
```
sergiusens: We might to check for consistency on use of "" versus `` and document it.

snapcraft/plugins/catkin.py
kyrofa: Is it though? Even the super class only sets a few variables in preparation for more work... doesn't it?

snapcraft/plugins/catkin.py
```diff
@@ -252,71 +229,55 @@ def build(self):
             '{}', ';'
         ])
```
sergiusens: no incentive to change this as well :wink:

Nothing that can't be done with `os.walk` and `re` or `str().replace`

snapcraft/plugins/catkin.py
```diff
@@ -252,71 +229,55 @@ def build(self):
...
         catkincmd.append('--pkg')
-        catkincmd.append(pkg)
+        catkincmd.extend(self.catkin_packages)
 
         # Don't clutter the real ROS workspace-- use the Snapcraft build
         # directory
         catkincmd.extend(['--directory', self.builddir])
 
+        # Account for a non-default source space by always specifying it
+        catkincmd.extend(['--source-space', os.path.join(
```
sergiusens: ah it indeed is called `source-space` ...

snapcraft/plugins/catkin.py
```diff
@@ -324,7 +285,58 @@ def _handle_package(self, pkg):
                 os.path.join(self.installdir, 'usr', 'bin', 'g++'))
         ])
 
-        self._rosrun(catkincmd)
+        # This command must run in bash due to a bug in Catkin that causes it
+        # to explode if there are spaces in the cmake args (which there are).
+        # This has been fixed in Catkin Tools... perhaps we should be using
+        # that instead.
```
sergiusens: not sure what *Catkin Tools* are but I'm inclined to say yes

snapcraft/plugins/catkin.py
```diff
@@ -340,40 +352,41 @@ def __init__(self, ros_distro, ros_package_path, rosdep_path,
         self._rosdep_cache_path = os.path.join(self._rosdep_path, 'cache')
 
     def setup(self):
+        # Make sure we can run multiple times without error, while leaving the
+        # capability to re-initialize, by making sure we clear the sources.
+        shutil.rmtree(self._rosdep_sources_path, ignore_errors=True)
```
sergiusens: What is the intention of `ignore_errors` here? If it is to clean after use, maybe contextlib is better here.

This will ignore the errors even if only some files failed to remove, I'll make it your call though.

kyrofa: Ah, good point. I threw that in there to make sure it was okay if the files didn't exist, but I should just be catching that exception. Good catch. Fixing.

snapcraft/plugins/catkin.py
kyrofa: Yeah, I'll be investigating that.

snapcraft/tests/test_plugin_catkin.py
```diff
@@ -45,69 +61,129 @@ class props:
...
-    def verify_rosdep_setup(self, plugin):
-        self.rosdep_mock.assert_has_calls([
-            mock.call(self.properties.rosdistro,
-                      os.path.join(plugin.sourcedir, 'src'),
-                      os.path.join(plugin.partdir, 'rosdep'),
-                      plugin.PLUGIN_STAGE_SOURCES),
-            mock.call().setup()])
+    def test_schema(self):
+        plugin = catkin.CatkinPlugin('test-part', self.properties)
+        schema = plugin.schema()
```
sergiusens:
```python
schema = catkin.CatkinPlugin.schema()
```

as `schema` is a `classmethod`

snapcraft/tests/test_plugin_catkin.py
```diff
@@ -233,41 +259,36 @@ def test_invalid_catkin_workspace_invalid_subdir(self):
...
+    def test_invalid_catkin_workspace_source_space_same_as_source(self):
+        self.properties.source_space = '.'
 
         # sourcedir is expected to be the root of the Catkin workspace. Since
-        # source_subdir was specified to be the same as the root, this should
+        # source_space was specified to be the same as the root, this should
         # fail.
         with self.assertRaises(RuntimeError) as raised:
             plugin = catkin.CatkinPlugin('test-part', self.properties)
             plugin.pull()
```
sergiusens: this call is never reached. Might want to check the test coverage for this file to spot others and remove dead test code :wink:

sergiusens:
:100: on this, looks really good in the readability domain!

Just had a couple nitpicks here and there.
Nice job on the 100% coverage.

snapcraft/tests/test_plugin_catkin.py
kyrofa: Oh this isn't here to DO anything. This test is saying "at least one of these lines had better throw an exception." I went back a forth a few times where I wanted the safety checks, so this makes the test less fragile to future changes.

snapcraft/plugins/catkin.py
sergiusens: I have no quarrel with what it does, just the docstring, not sure you noticed, but `__init__` above is also a link :innocent:

kyrofa: Alright, I removed the docstring and put the comment back where it was.

snapcraft/plugins/catkin.py
kyrofa: Done.

snapcraft/plugins/catkin.py
kyrofa: Fixed.

snapcraft/tests/test_plugin_catkin.py
kyrofa: Fixed, thanks!

kyrofa:
Thanks for the good review @sergiusens! I've pushed a new version that should address your concerns.

snapcraft/tests/test_plugin_catkin.py
sergiusens: You can make this a oneline then ;-)

`catkin.CatkinPlugin('test-part', self.properties).pull()`

sergiusens:
Looks good, just adding one more comment about `try/except` versus checking. Using `if` here might be better, no functionality is lost so I won't block (another minor one about the testing strategy)

you have my :+1: regardless

snapcraft/tests/test_plugin_catkin.py
kyrofa: Ah, indeed! Fixed.

kyrofa:
@sergiusens Fixed both. You're right, the `if` probably conveys my intentions better. I'm going to go ahead and merge this, then.

PR #193, opened 2015-12-23
node.js plugin: add node-folder property (default: '.')

asac:
This adds support for building snap parts from sources that dont have their
package.json in the top level directory (such as https://github.com/asac/etherpad-lite).

### DIFF
--- a/snapcraft/plugins/nodejs.py
+++ b/snapcraft/plugins/nodejs.py
@@ -28,6 +28,10 @@ Additionally, this plugin uses the following plugin-specific keywords:
     - node-packages:
       (list)
       A list of dependencies to fetch using npm.
+    - node-folder:
+      (string)
+      A relative path to the folder containing the package.json to use with
+      npm install
 """
 
 import logging
@@ -65,6 +69,11 @@ class NodePlugin(snapcraft.BasePlugin):
             'default': [],
         }
 
+        schema['properties']['node-folder'] = {
+            'type': 'string',
+            'default': '.',
+        }
+
         if 'required' in schema:
             del schema['required']
 
@@ -85,8 +94,11 @@ class NodePlugin(snapcraft.BasePlugin):
         self._nodejs_tar.provision(self.installdir)
         for pkg in self.options.node_packages:
             self.run(['npm', 'install', '-g', pkg])
-        if os.path.exists(os.path.join(self.builddir, 'package.json')):
-            self.run(['npm', 'install', '-g'])
+        if os.path.exists(os.path.join(self.builddir,
+                                       self.options.node_folder,
+                                       'package.json')):
+            self.run(['npm', 'install', '-g',
+                     os.path.join(self.builddir, self.options.node_folder)])
 
 
 def _get_nodejs_base():
--- a/snapcraft/tests/test_plugin_nodejs.py
+++ b/snapcraft/tests/test_plugin_nodejs.py
@@ -44,6 +44,7 @@ class NodePluginTestCase(tests.TestCase):
         class Options:
             source = '.'
             node_packages = []
+            node_folder = '.'
 
         plugin = nodejs.NodePlugin('test-part', Options())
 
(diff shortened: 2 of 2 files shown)


snapcraft/plugins/nodejs.py
```diff
@@ -85,8 +94,11 @@ def build(self):
         self._nodejs_tar.provision(self.installdir)
         for pkg in self.options.node_packages:
             self.run(['npm', 'install', '-g', pkg])
-        if os.path.exists(os.path.join(self.builddir, 'package.json')):
-            self.run(['npm', 'install', '-g'])
+        if os.path.exists(os.path.join(self.builddir,
+                                       self.options.node_folder,
+                                       'package.json')):
+            self.run(['npm', 'install', '-g',
+                     os.path.join(self.builddir, self.options.node_folder)])
```
kyrofa: Just to verify (not a nodejs expert here): it's safe to do this even when the `package.json` IS in the root of the project?

snapcraft/tests/test_plugin_nodejs.py
```diff
@@ -58,10 +59,34 @@ class Options:
...
+
+        os.makedirs(plugin.sourcedir)
+        os.makedirs(os.path.join(plugin.sourcedir, plugin.options.node_folder))
+        open(os.path.join(plugin.sourcedir, plugin.options.node_folder,
+                          'package.json'), 'w').close()
+
+        plugin.build()
+
+        self.run_mock.assert_has_calls([
+            mock.call(['npm', 'install', '-g'], cwd=plugin.builddir)])
```
kyrofa: Yeah you're on the right track here. You just need to add the expected path to the `package.json` after the `'-g'` argument in the call.

snapcraft/plugins/nodejs.py
asac: 2015-12-23 19:14 GMT+01:00 Kyle Fazzari <notifications@github.com>:
>
> In snapcraft/plugins/nodejs.py:
>
> > @@ -85,8 +94,11 @@ def build(self):
> >          self._nodejs_tar.provision(self.installdir)
> >          for pkg in self.options.node_packages:
> >              self.run(['npm', 'install', '-g', pkg])
> > -        if os.path.exists(os.path.join(self.builddir, 'package.json')):
> > -            self.run(['npm', 'install', '-g'])
> > +        if os.path.exists(os.path.join(self.builddir,
> > +                                       self.options.node_folder,
> > +                                       'package.json')):
> > +            self.run(['npm', 'install', '-g',
> > +                     os.path.join(self.builddir, self.options.node_folder)])
>
> Just to verify (not a nodejs expert here): it's safe to do this even when the package.json IS in the root of the project?

node.path definition api doc says:

```
Note: If the arguments to join have zero-length strings, unlike other
path module functions, they will be ignored. If the joined path string
is a zero-length string then '.'will be returned, which represents the
current working directory.
``` - https://nodejs.org/api/path.html#path_path_join_path1_path2

so if node_folder is '' (zero-length) or '.' the result will be same
as $builddir and $builddir/. always poing to same directory as this
one... or just builddir, which should be fine ...

> path.join ("asdda", "subdir", "package.json");
'asdda/subdir/package.json'
> path.join ("asdda", "", "package.json");
'asdda/package.json'
> path.join ("asdda", ".", "package.json");
'asdda/package.json'


> —
> Reply to this email directly or view it on GitHub.

asac: OK,

I think its correct this way, but i now see that its smarter to use
the default '' rather than ''. In that way the dir of join will be
exactly the same as without this patch for the current parts that dont
specify node-folder... let me change that and then it should safe...

Wait for resubmit!

 - Alexander


2015-12-24 15:59 GMT+01:00 Alexander Sack <asacasa@gmail.com>:

kyrofa:
Discussion on IRC led to the discovery that `node-packages: ['src/']` seems to solve this problem. Closing this-- please reopen if you discover that it's still necessary.

PR #194, opened 2015-12-23
node.js plugin: add node-folder property (default: '.')

asac:
This adds support for building snap parts from sources that dont have their
package.json in the top level directory (such as https://github.com/asac/etherpad-lite).


kyrofa:
Discussion on IRC led to the discovery that node-packages: ['src/'] seems to solve this problem. Closing this-- please reopen if you discover that it's still necessary.

PR #199, opened 2016-01-04
Fix ordering for conditional negations in catkin tests

sergiusens:
The configtional is `if not ... in ...` when it should be `if ... not in ...`


elopio:
:+1:

PR #200, opened 2016-01-05
Replace use of `assert_not_called` in the catkin tests

sergiusens:
Mocks using `assert_not_called` don't take a message, to keep the message
we use `assertFalse` and the `called` attribute of the mock.


kyrofa:
Thanks @sergiusens! Looks great. I'll just wait for Travis to finish.

PR #201, opened 2016-01-05
Add licensing support

sergiusens:
(Closes LP: #1527453)

### DIFF
--- a/integration_tests/test_license.py
+++ b/integration_tests/test_license.py
@@ -1,0 +1,39 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2016 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+
+from testtools.matchers import (
+    DirExists,
+    FileContains,
+    FileExists,
+)
+
+import integration_tests
+
+
+class LicenseTestCase(integration_tests.TestCase):
+
+    def test_license(self):
+        project_dir = 'license'
+        self.run_snapcraft('strip', project_dir)
+
+        hooks_dir = os.path.join(project_dir, 'snap', 'meta', 'hooks')
+        self.assertThat(hooks_dir, DirExists())
+        license_hook = os.path.join(hooks_dir, 'license')
+        self.assertThat(license_hook, FileExists())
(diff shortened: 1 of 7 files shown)


integration_tests/test_license.py
```diff
@@ -0,0 +1,39 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+
+from testtools.matchers import (
+    DirExists,
+    FileExists,
+    FileContains,
```
elopio: FileContains before FileExists, so it's alphabetical.

snapcraft/tests/test_meta.py
```diff
@@ -317,6 +334,41 @@ def test_create_meta_with_vararg_config(self, mock_the_open,
...
+            ),
+        ])
+
+        self.mock_copyfile.assert_has_calls([
+            call('my-icon.png', os.path.join(self.meta_dir,
+                 'my-icon.png')),
+            call('file.apparmor', os.path.join(self.meta_dir,
+                 'file.apparmor')),
+            call('file.seccomp', os.path.join(self.meta_dir,
+                 'file.seccomp')),
```
elopio: These three calls are not related to the license. You can use mock.ANY instead, I think here we only care about the license call.

snapcraft/tests/test_meta.py
```diff
@@ -317,6 +334,41 @@ def test_create_meta_with_vararg_config(self, mock_the_open,
...
+
+        mock_the_open.assert_has_calls(self.expected_open_calls)
+        mock_wrap_exe.assert_has_calls([
+            call(
+                '$SNAP_APP_PATH/bin/bash',
+                os.path.join(os.path.abspath(os.curdir),
+                             'snap/bin/bash.wrapper'),
+                args=None, shebang=None,
+            ),
+        ])
```
elopio: How is this assertion related to the license? You might be testing too much in this test, this seems to me like something to do in a separate test.

snapcraft/tests/test_yaml.py
```diff
@@ -542,6 +542,38 @@ def test_invalid_part_name_plugin_raises_exception(self):
...
+
+    def test_license_version_without_license_raises_exception(self):
+        self.data['license-version'] = '1.1'
+
+        with self.assertRaises(snapcraft.yaml.SnapcraftSchemaError) as raised:
+            snapcraft.yaml._validate_snapcraft_yaml(self.data)
+
+        expected_message = "'license' is a dependency of 'license-version'"
+        self.assertEqual(raised.exception.message, expected_message,
+                         msg=self.data)
```
elopio: What about a scenario for version without agreement?

elopio:
You have here some files without the trailing new line. I think we should adopt the practice of ending all files with a new line.
Also I left some comments for discussion.

sergiusens:
@chipaca maybe you want to take a look.

snapcraft/tests/test_yaml.py
sergiusens: added `def test_license_with_license_version(self):`

sergiusens:
@elopio I think I gone through all your requests.

chipaca:
:+1:

elopio:
yes, thanks, :+1:

PR #204, opened 2016-01-06
Add support for mesa libraries.

kyrofa:
Currently Snapcraft adds only a few standard paths to LD_LIBRARY_PATH. This PR expands that to include paths used by mesa packages.

This will need to be backported to 1.x once approved.

Resolves LP: [#1531620](https://bugs.launchpad.net/snapcraft/+bug/1531620)

### DIFF
--- a/snapcraft/libraries.py
+++ b/snapcraft/libraries.py
@@ -1,0 +1,50 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2016 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import re
+import glob
+
+
+def determine_ld_library_path(root):
+    # If more ld.so.conf files need to be supported, add them here.
+    ld_config_globs = {
+        '{}/usr/lib/*/mesa*/ld.so.conf'.format(root)
+    }
+
+    ld_library_paths = []
+    for this_glob in ld_config_globs:
+        for ld_conf_file in glob.glob(this_glob):
+            ld_library_paths.extend(_extract_ld_library_paths(ld_conf_file))
+
+    return [root + path for path in ld_library_paths]
+
+
+def _extract_ld_library_paths(ld_conf_file):
+    # From the ldconfig manpage, paths can be colon-, space-, tab-, newline-,
+    # or comma-separated.
+    path_delimiters = re.compile(r'[:\s,]')
+    comments = re.compile(r'#.*$')
+
+    paths = []
+    with open(ld_conf_file, 'r') as f:
+        for line in f:
+            # Remove comments from line
+            line = comments.sub('', line).strip()
+
+            if line:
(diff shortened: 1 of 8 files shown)


snapcraft/tests/test_yaml.py
```diff
@@ -346,6 +347,107 @@ def test_config_expands_filesets(self, mock_loadPlugin):
...
 
+class TestYamlEnvironment(tests.TestCase):
+
+    def setUp(self):
+        super().setUp()
+        dirs.setup_dirs()
+
+        tempdirObj = tempfile.TemporaryDirectory()
+        self.addCleanup(tempdirObj.cleanup)
+        os.chdir(tempdirObj.name)
```
elopio: We already have this on the base setup, when this test runs the cwd is already a tempdir that will be deleted. Tha path to that dir is in self.path.
https://github.com/ubuntu-core/snapcraft/blob/master/snapcraft/tests/__init__.py#L33

snapcraft/tests/test_yaml.py
```diff
@@ -346,6 +347,107 @@ def test_config_expands_filesets(self, mock_loadPlugin):
...
+        with open('icon.svg', 'w') as icon:
+            icon.write('<svg />')
```
elopio: aren't the icons optional now? I can't remember if that landed or not yet.

kyrofa: Oh right, I forgot about that! I do believe it's landed.

elopio: ah, yes it is because you don't have it in your example! So this can be removed.

snapcraft/tests/test_yaml.py
kyrofa: Ah good find! I was copying what was already there-- I've fixed the rest of the tests as well.

snapcraft/tests/test_yaml.py
kyrofa: Fixed.

elopio:
lgtm. Maybe, I would move the library handling to a new module. snapcraft/libraries?
I don't like that we have the mesa lib hardcoded and that we'll have to keep adding paths there, but I don't know much about that and can't think of a better way to do it.

Thanks for the example!

kyrofa:
@elopio alright I moved it into `snapcraft.libraries`. I agree with you on the hard-coding, but there doesn't seem to be another way to do it since the .debs typically configure themselves in ld.so.conf.d via postinst scripts, which don't get run with how we're unpacking them.

sergiusens:
I re triggered the examples test run

snapcraft/libraries.py
```diff
@@ -0,0 +1,50 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2015 Canonical Ltd
```
sergiusens: 2016

snapcraft/tests/test_libraries.py
```diff
@@ -0,0 +1,45 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2015 Canonical Ltd
```
sergiusens: 2016

snapcraft/tests/test_yaml.py
```diff
@@ -317,7 +300,6 @@ def test_config_expands_filesets(self, mock_loadPlugin):
 version: "1"
 summary: test
 description: test
-icon: my-icon.png
```
sergiusens: shouldn't these be a separate commit?

sergiusens:
What wishlist bug is this closing? :wink:

snapcraft/tests/test_yaml.py
kyrofa: They could be. These small refactors were to simply make the tests consistent with the ones I was adding rather than being consistent the other way and adding more cruft.

kyrofa:
@sergiusens heh. Fixed :smiley: .

sergiusens:
:+1:

PR #206, opened 2016-01-06
Autotools plugin: Set autogen.sh executable.

kyrofa:
This PR fixes LP: [#1530995](https://bugs.launchpad.net/snapcraft/+bug/1530995) by setting autogen.sh to be executable (if necessary). It also adds tests for the autotools plugin.

It will need to be backported to 1.x when approved.

### DIFF
--- a/snapcraft/tests/test_plugin_autotools.py
+++ b/snapcraft/tests/test_plugin_autotools.py
@@ -1,0 +1,163 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2016 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+import stat
+
+from unittest import mock
+
+from snapcraft import tests
+from snapcraft.plugins import autotools
+
+
+class AutotoolsPluginTestCase(tests.TestCase):
+
+    def setUp(self):
+        super().setUp()
+
+        class Options:
+            configflags = []
+
+        self.options = Options()
+
+        patcher = mock.patch('snapcraft.repo.Ubuntu')
+        self.ubuntu_mock = patcher.start()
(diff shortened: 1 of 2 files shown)


snapcraft/tests/test_plugin_autotools.py
```diff
@@ -0,0 +1,155 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2015 Canonical Ltd
```
sergiusens: 2016

snapcraft/tests/test_plugin_autotools.py
```diff
@@ -0,0 +1,155 @@
...
+        self.options = Options()
+
+        patcher = mock.patch('snapcraft.repo.Ubuntu')
+        self.ubuntu_mock = patcher.start()
+        self.addCleanup(patcher.stop)
+
+    def test_schema(self):
+        schema = autotools.AutotoolsPlugin.schema()
+
+        # Check rosdistro property
```
sergiusens: this comment seems out of place :wink:

snapcraft/tests/test_plugin_autotools.py
```diff
@@ -0,0 +1,155 @@
...
+
+        self.assertEqual(3, run_mock.call_count)
+        run_mock.assert_has_calls([
+            mock.call(['autoreconf', '-i']),
+            mock.call(['./configure', '--prefix=']),
+            mock.call(['make', 'install',
+                       'DESTDIR={}'.format(plugin.installdir)])
+        ])
+
+    @mock.patch('os.chmod')
```
sergiusens: why are you mocking `os.chmod` but actually creating the file?

It seems less error prone to not mock `os.chmod` and `os.stat` it

kyrofa: Actually I think I found something even better-- test exactly what I'm trying to do: make `autogen.sh` runnable.

kyrofa:
@sergiusens Alright I believe I've addressed your concerns here.

snapcraft/tests/test_plugin_autotools.py
```diff
@@ -0,0 +1,165 @@
...
+
+        # We want to mock out every run() call except the one to autogen
+        def _run(cmd):
+            if './autogen.sh' in cmd:
+                patcher.stop()
+                output = plugin.run(cmd)
+                patcher.start()
+                return output
+
+            return None
```
sergiusens: maybe remove this statement

```
In [1]: def a():
   ...:     pass
   ...: 

In [2]: a() is None
Out[2]: True
```

or `return ''` so it is consistent.

kyrofa: Ah, good catch. Fixed.

sergiusens:
:+1: feel free to merge once the tools are in; also, the commit message is missing the (Closes LP: #XXXXX) comment.

PR #209, opened 2016-01-08
Added symlinks option to the copy plugin

femdom:
We need this change since we need to copy large directories.

Please, consider adding this change to your codebase.

Thank you,
Renat

### DIFF
--- a/snapcraft/plugins/copy.py
+++ b/snapcraft/plugins/copy.py
@@ -36,10 +36,9 @@ class CopyPlugin(snapcraft.BasePlugin):
                 },
                 'follow-symlinks': {
                     'type': 'boolean',
-                    'default': False,
-                    'description': 'symbolic links in the source tree are'
-                                   ' represented as symbolic links'
-                                   ' in the new tree',
+                    'default': True,
+                    'description': 'copy linked file content instead of'
+                                   ' copying only symlink metadata',
                 }
             },
             'required': [
@@ -64,6 +63,7 @@ class CopyPlugin(snapcraft.BasePlugin):
             os.makedirs(os.path.dirname(dst), exist_ok=True)
 
             if os.path.isdir(src):
-                shutil.copytree(src, dst, symlinks=self.options.follow_symlinks)
+                shutil.copytree(
+                    src, dst, symlinks=not self.options.follow_symlinks)
             else:
                 shutil.copy2(src, dst)
--- a/snapcraft/plugins/copy.py
+++ b/snapcraft/plugins/copy.py
@@ -34,6 +34,13 @@ class CopyPlugin(snapcraft.BasePlugin):
                 'files': {
                     'type': 'object',
                 },
+                'follow-symlinks': {
+                    'type': 'boolean',
+                    'default': False,
+                    'description': 'symbolic links in the source tree are'
+                                   ' represented as symbolic links'
+                                   ' in the new tree',
+                }
             },
             'required': [
                 'files',
@@ -57,6 +64,6 @@ class CopyPlugin(snapcraft.BasePlugin):
             os.makedirs(os.path.dirname(dst), exist_ok=True)
 
             if os.path.isdir(src):
-                shutil.copytree(src, dst)
+                shutil.copytree(src, dst, symlinks=self.options.follow_symlinks)
             else:
                 shutil.copy2(src, dst)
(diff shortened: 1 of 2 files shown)


kyrofa:
@femdom thank you for the PR! Please check out our [contributing guidelines](https://github.com/ubuntu-core/snapcraft/blob/master/CONTRIBUTING.md) and make sure you've completed all the steps. In particular: 

- Create a bug for this issue.
- Squash.

Thank you!

femdom:
@kyrofa, thanks!  I remember. Just wanted to know if you need this change.

femdom:
Here is the launchpad bug page
https://bugs.launchpad.net/snapcraft/+bug/1532515

snapcraft/plugins/copy.py
```diff
@@ -34,6 +34,13 @@ def schema(cls):
                 'files': {
                     'type': 'object',
                 },
+                'symlinks': {
```
sergiusens: how about `follow-symlinks`? It has a more semantic meaning for a `boolean`

femdom: Understood.

snapcraft/plugins/copy.py
```diff
@@ -57,6 +64,6 @@ def build(self):
             os.makedirs(os.path.dirname(dst), exist_ok=True)
 
             if os.path.isdir(src):
-                shutil.copytree(src, dst)
+                shutil.copytree(src, dst, symlinks=self.options.follow_symlinks)
```
kyrofa: This line is too long, and is failing the static checks.

kyrofa: Quick question: If the symlink is pointing outside of the tree, this will create a .snap with dangling symlinks. Right? Reading the docs it sounds like `ignore_dangling_symlinks` only has an affect if `symlinks` is false.

femdom: Sorry, my mistake. Follow symlniks - mean "copy file where symlink points"
Let me update a PR then.

kyrofa: @femdom good catch, I missed that switch.

@sergiusens, @elopio we could use your input on the symlinks potentially reaching out of the .snap. Should we make sure that doesn't happen with a smarter copy function, or allow them to shoot themselves in the foot if they don't want to follow symlinks?

elopio: I'm +1 on doing sanity checks for the resulting snap and/or preventing it from generating stupid things.

snapcraft/tests/test_plugin_copy.py
```diff
@@ -31,6 +32,42 @@ def setUp(self):
...
+    def test_copy_plugin_with_follow_symlink_true_copies_symlinked_file(self):
+        self.mock_options.files = {
+            'src': 'dst'
+        }
+
+        self.mock_options.follow_symlinks = True
+
+        os.makedirs('src')
+        open('src/file.txt', 'w').close()
+        os.symlink('file.txt', 'src/link.txt')
```
elopio: shouldn't this be src/file.txt?

femdom:
@kyrofa, is showing warning enough when symlink points outside the snap, or you want to stop build process on that case?

snappy-m-o:

snappy-m-o:

snappy-m-o:

kyrofa:
@femdom actually we'd like to redesign this bit of the copy plugin completely. Rather than support following symlinks and not following symlinks with an option, we'd rather get rid of the options and copy in the most efficient manner possible. Something like:

- For every file:
    - If it is not a symlink:
        - Copy it
    - If it is a symlink:
        - If it is a relative symlink, pointing to something else within the snap:
            - Copy the symlink
        - If it is an absolute symlink, or pointing outside the snap:
            - Follow the symlink and copy what it's pointing to

This would be the best of both worlds, would it not?

femdom:
@kyrofa I'm not sure about the last step. What if it's pointing to the something like /etc/timezone? What if it's absolute symlink - copy it as a symlink?

kyrofa:
@femdom if the final .snap contains a symlink to /etc/timezone it won't pass the automated review tools anyway. If it's an absolute symlink or pointing outside the snap, follow the symlink and copy what it's pointing to, _not_ the symlink itself. The only exception that the review tools will allow are symlinks to libc-- think we'd run into that with the copy plugin?

kyrofa:
@femdom I'll close this for now. Please reopen (or create a new one) if you'd like to continue on this.

PR #211, opened 2016-01-08
Update CONTRIBUTING.md to require a bug.

kyrofa:
This PR resolves LP: [#1532195](https://bugs.launchpad.net/snapcraft/+bug/1532195) by making the first step of the contribution guidelines "Make sure there's a bug."

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,9 +5,12 @@ you want to make Snapcraft even better. However, we do ask that you follow some
 general guidelines while doing so, just so we can keep things organized around
 here.
 
-1. Sign the [contributor license agreement][1].
+**Prerequisite:** Sign the [contributor license agreement][1]. This is how you
+give us permission to use your contributions.
 
-   This is how you give us permission to use your contributions.
+1. Make sure a [Snapcraft bug][2] is filed for the bug you're about to fix, or
+   feature you're about to add. If it's a feature, mark the status as
+   "Wishlist."
 
 2. We use a forking, feature-based workflow.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -15,18 +18,21 @@ here.
    feature on which you'd like to work. Make your changes there, adding new
    tests as needed, and make sure the existing tests continue to pass when your
    changes are complete (for information about running the tests, see the
-   [HACKING][2] document).
+   [HACKING][3] document).
 
-3. Squash commits into one, well-formatted commit.
+3. Squash commits into one, well-formatted commit. Mention the bug being
+   resolved in the commit message on a line all by itself like `LP: #<bug>`.
```
elopio: Does it work also making it a link?

elopio:
:+1:
Only a comment about making the LP comment a link.

CONTRIBUTING.md
kyrofa: You mean only in the PR instead of the commit? If so, no-- the tools we use for changelog generation parse the commit logs, not the PR descriptions.

kyrofa: If you mean adding a link to the commit, that also doesn't work since git-dch parses for the `LP: #<bug>` pattern, which a link wouldn't match.

sergiusens:
:+1: a link in the PR like @kyrofa does is an added bonus, not sure if necessary,

kyrofa:
Agreed. Let's merge this, and if the lack of links becomes painful we can revamp.

PR #212, opened 2016-01-08
Ensure rosdep resolves dependencies using Trusty.

kyrofa:
This PR fixes LP: [#1532241](https://bugs.launchpad.net/snapcraft/+bug/1532241) by instructing rosdep to resolve dependencies using Trusty, even if Snapcraft is not being run on Trusty.

This will need to be cherry-picked to 1.x as well.

### DIFF
--- a/snapcraft/plugins/catkin.py
+++ b/snapcraft/plugins/catkin.py
@@ -419,8 +419,16 @@ class _Rosdep:
 
     def resolve_dependency(self, dependency_name):
         try:
+            # rosdep needs three pieces of information here:
+            #
+            # 1) The dependency we're trying to lookup.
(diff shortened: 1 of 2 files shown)


sergiusens:
:+1: but I would like to test on xenial as well (the resulting snap), if @elopio tests it, then I'm good for merge

elopio:
Tested in xenial:
It builds: Snapped ros-example_1.0_amd64.snap
Installing on rolling edge #310 works, but when I run the binaries I get python not found, rosrun not found, cat not found.

kyrofa:
Yeah we're probably going to run into other issues with rolling as well. This is verified to at least work on 15.04, built from trusty and xenial, so we'll roll with it.

PR #221, opened 2016-01-12
Documentation refresh.

sergiusens:
Refreshing the docs refering to yaml keywords and lifecycle changes affecting
2.0

LP: #1533021


docs/snapcraft-advanced-features.md
```diff
@@ -189,26 +195,28 @@ directory structure.
...
+	    - usr/lib
+	  go-server:
+	    - bin/golang-*
+	stage:
+	  - $fswebcam
+	  - $go-server
+	snap:
+	  - $fswebcam
+	  - $go-server
+	  - -usr/share/doc
```
elopio: I guess we should be consistent and use only spaces.

docs/snapcraft-syntax.md
```diff
@@ -20,24 +20,33 @@ contain.
...
    Path to the runnable in snap that will be used as the [Snappy config
    interface](https://developer.ubuntu.com/snappy/guides/config-command/).
- * `services` (yaml subsection)  
-   A set of keys representing service names with values as defined by the
-   [Snappy packaging spec](https://developer.ubuntu.com/snappy/guides/packaging-format-apps/).
- * `binaries` (yaml subsection)  
-   A set of keys representing binary names with values as defined by the
-   [Snappy packaging spec](https://developer.ubuntu.com/snappy/guides/packaging-format-apps/).
+ * `apps` (yaml subsection)  
+   A map of keys for integration points for the system. Either as a daemon
```
elopio: "integration points for the system" says nothing to me. Do we have a better way to tell what's an app?

docs/snapcraft-syntax.md
```diff
@@ -20,24 +20,33 @@ contain.
...
+   or command line accessible binary.
+   * `command` (string)
+     Specifies the internal command to expose.
```
elopio: Maybe mention that for a daemon, this is the command that starts it.

docs/snapcraft-syntax.md
```diff
@@ -20,24 +20,33 @@ contain.
...
+   * `daemon` (string)
+     If present, integrates the runnable as a system service. Valid values are
+     forking and simple.
```
elopio: Can you please explain what forking and simple are?

docs/snapcraft-syntax.md
```diff
@@ -20,24 +20,33 @@ contain.
...
+   * `stop-command` (string)
+     Requires `daemon` to be specified and represents the command to run to
+     effectively stop the service.
```
elopio: "effectively" seems to serve no purpose here. The sentence says the same if you remove it.

docs/snapcraft-syntax.md
```diff
@@ -20,24 +20,33 @@ contain.
...
+   * `stop-timeout` (integer)
+     Requires `daemon` to be specified. It is number in seconds to indicate how
```
elopio: It is number of seconds that the system will wait for the service to stop.
I'm not sure if this sounds better, so @kyrofa please.

docs/snapcraft-syntax.md
```diff
@@ -20,24 +20,33 @@ contain.
...
-     If true, license acceptance is required for the package to be activated.
-     A good example for this one is the Sun JRE/JDK being bundled in a snap.
- * `framework-policy` (string)  
-   A relative path to a directory containing additional policies, used if
-   creating a framework and want to extend permissions to snap apps.
+ * `license` (string)  
+   Path to a license file.
+ * `license-agreement` (string)  
+   Requires `license` to be set. The only valid value for this entry is
+   `explicit` stating that if a license is present, acceptance to it must
```
elopio: acceptance OF it? @kyrofa: help again.

docs/snapcraft-syntax.md
```diff
@@ -20,24 +20,33 @@ contain.
...
+   require an explicit action.
+   A good example for this one is the Sun JRE/JDK being bundled in a snap.
+ * `license-version` (string)  
+   Requires `license` to be set. Effectively the version for the license.
```
elopio: I would remove "Effectively"

docs/snapcraft-syntax.md
elopio: What about: stating that if a license is present, the user must accept it in order to install the snap.

elopio:
I summon also @dholbach.

docs/snapcraft-syntax.md
```diff
@@ -20,24 +20,41 @@ contain.
...
+     fork() as part of its start-up. The parent process is expected to exit
+     when start-up is complete and all communication channels are set up.
+     The child continues to run as the main daemon process. This is the
+     behavior of traditional UNIX daemons.
+   * `stop-command` (string)
+     Requires `daemon` to be specified and represents the command to run to
+     stop the service.
+   * `stop-timeout` (integer)
+     Requires `daemon` to be specified. It is number in seconds that the
+     system will wait for the service to stop.
```
kyrofa: Sounds threatening. Should we specify what happens if the daemon takes too long to stop?

docs/snapcraft-syntax.md
```diff
@@ -20,24 +20,41 @@ contain.
...
-     A good example for this one is the Sun JRE/JDK being bundled in a snap.
- * `framework-policy` (string)  
-   A relative path to a directory containing additional policies, used if
-   creating a framework and want to extend permissions to snap apps.
+ * `license` (string)  
+   Path to a license file.
+ * `license-agreement` (string)  
+   Requires `license` to be set. The only valid value for this entry is
+   `explicit` which requires the license to be accepted for the snap to
+   install.
```
kyrofa: Yeah this reads well.

docs/snapcraft-syntax.md
sergiusens: I don't know how to change reality :-)

> Configures the time to wait for stop. If a service is asked to stop, but does not terminate in the specified time, it will be terminated forcibly via SIGTERM, and after another timeout of equal duration with SIGKILL (see KillMode= in systemd.kill(5)). Takes a unit-less value in seconds, or a time span value such as "5min 20s". Pass "0" to disable the timeout logic. Defaults to DefaultTimeoutStopSec= from the manager configuration file (see systemd-system.conf(5)).

Unless we disable the possibility of allowing 0

kyrofa: Do we pass this straight to systemd? Heck, we might as well include almost that entire text. How about changing this to:

```
Requires `daemon` to be specified. It is the length of time that the system will wait for the service to stop before terminating it via SIGTERM (and SIGKILL if that doesn't work). Accepts a unit-less value in seconds, or a time span value such as "5min 20s." A value of 0 indicates that the system should wait forever for the service to stop.
```

kyrofa: Oh wait-- it's an integer. Okay, how about this:

```
Requires `daemon` to be specified. It is the length of time in seconds that the
system will wait for the service to stop before terminating it via SIGTERM (and
SIGKILL if that doesn't work).
```

sergiusens: This sounds sensible, we might be removing the `0` or disallow it so this is good

sergiusens: done

PR #222, opened 2016-01-12
Support the new security-override semantics

sergiusens:
LP: [#1533293](https://bugs.launchpad.net/snapcraft/+bug/1533293)

### DIFF
--- a/schema/snapcraft.yaml
+++ b/schema/snapcraft.yaml
@@ -1,7 +1,7 @@
 $schema: http://json-schema.org/draft-04/schema#
 
 definitions:
-  security:
+  security-policy:
     type: object
     required:
       - apparmor
@@ -13,6 +13,47 @@ definitions:
       seccomp:
         type: string
         format: file-path
+  security-override:
+      type: object
+      properties:
+        read-paths:
+          type: array
+          description: a list of additional paths that the app can read.
+          minItems: 1
+          uniqueItems: true
+          items:
+              type: string
+        write-paths:
+          type: array
+          description: a list of additional paths that the app can write.
+          minItems: 1
+          uniqueItems: true
+          items:
+              type: string
+        abstractions:
+          type: array
+          description: a list of additional AppArmor abstractions for the app.
+          minItems: 1
+          uniqueItems: true
+          items:
+              type: string
+        syscalls:
+          type: array
+          description: a list of additional syscalls the app can use.
+          minItems: 1
+          uniqueItems: true
+          items:
+              type: string
+  security-caps:
+    type: array
+    description: Security caps.
+    minItems: 1
+    uniqueItems: true
+    items:
+      type: string
+  security-template:
+      type: string
+      description: alternate security template to use instead of default.
 
 title: snapcraft schema
 type: object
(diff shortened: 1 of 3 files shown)


schema/snapcraft.yaml
```diff
@@ -13,6 +13,44 @@ definitions:
...
+        write-paths:
+          type: array
+          description: a list of additional paths that the app can write.
+          minItems: 1
+          uniqueItems: true
+          items:
+              type: string
+        abstractions:
+          type: array
+          description: a list of additional Apparmor abstractions for the app.
```
kyrofa: AppArmor. Only slightly nit-picky.

snapcraft/tests/test_yaml.py
```diff
@@ -614,6 +615,77 @@ def test_license_version_without_license_raises_exception(self):
...
+                    'abstractions': ['abstraction1', 'abstraction2'],
+                    'syscalls': ['open', 'close'],
+                },
+                'caps': ['cap1', 'cap2'],
+                'security-policy': {
+                    'seccomp': 'file.seccomp',
+                    'apparmor': 'file.apparmor',
+                },
+            },
+        }
```
kyrofa: You're missing three combos here, right. Worth adding them?

kyrofa:
Couple nits, but overall looks good.

schema/snapcraft.yaml
sergiusens: fair enough :-)

snapcraft/tests/test_yaml.py
sergiusens: what combos? with `subTest` I cover

- *all*
- `caps` & `security-policy`
- `security-override` & `security-policy`
- `security-override` & `caps`

kyrofa: Ah, I read too quickly. However, I can break your code above by removing all but one invalid combo and this test would still pass, correct? Should you be explicitly testing each invalid combination rather than throwing them all together?

kyrofa: Ah, I missed the deletion below. You do exactly as I ask, just by altering the app on the fly a few times. It could be a bit clearer, but I'm fine with this.

schema/snapcraft.yaml
```diff
@@ -13,6 +13,44 @@ definitions:
...
+          uniqueItems: true
+          items:
+              type: string
+  security-caps:
+    type: array
+    description: Security caps.
+    minItems: 1
+    uniqueItems: true
+    items:
+      type: string
```
jdstrand: You added caps, security-override and security-policy but left out security-template. Since you seem to be going for completeness of the current implementation, you should include security-template. It is of type string. It may be used with caps and security-override, but not security-policy.

sergiusens: Added. Thanks!

jdstrand: I think you misunderstood my comment. There are four pieces of yaml that refer to security policy:
 * caps
 * security-override
 * security-policy
 * security-template

The PR I commented on dealt with the first 3 so I suggested adding 'security-template' of type string, but instead you changed caps and left out security-template.

jdstrand: I'm sorry, please disregard my last comment-- I misread the email notification.

PR #224, opened 2016-01-12
Split the static and unit tests.

elopio:
@sergiusens wants to run the static checks in something newer than trusty. We can do that with the lxc branch, but we have to split them from the unit tests that we do want to run in trusty.

Fixes LP: [#1533397](https://bugs.launchpad.net/snapcraft/+bug/1533397)


sergiusens:
This is missing the LP bug link in the actual commit, other than that :+1:

elopio:
agh, I think I did it wrong because I added "Fixes" on the commit message :(

PR #226, opened 2016-01-12
Added the option to output subunit results in examples tests.

elopio:
Fixes LP: [#1533403](https://bugs.launchpad.net/snapcraft/+bug/1533403)


sergiusens:
This is missing the bug # in the commit message. Aside from that, :+1:

PR #228, opened 2016-01-13
Updated the tomcat example network capability.

elopio:
LP: 1533736


sergiusens:
:+1: from me.

I think that `LP: 1533736` should be `LP: #1533736` in the commit message though

elopio:
grrrr, damn it, I never get it right. I amended and pushed. Thanks for checking.

PR #229, opened 2016-01-13
Fix environment variable handling

sergiusens:
Environment variables now correctly point to stagedir and snapdir and
don't expose the internal installdir paths for a part.

This required an intermediate change in the handling of stage-packages
but it also gives it a semantic meaning by unpacking everything into
the staging directory. Future work can make the handling of
stage-package more transactional or in a single shot.

LP: #1531481

### DIFF
--- a/snapcraft/tests/test_yaml.py
+++ b/snapcraft/tests/test_yaml.py
@@ -342,13 +342,19 @@ description: test
 
 parts:
   part1:
-    plugin: go
-    stage-packages: [fswebcam]
+    plugin: nil
 """)
 
-    def test_config_runtime_environment(self):
+        patcher = unittest.mock.patch('snapcraft.common.get_arch_triplet')
+        mock_arch = patcher.start()
+        mock_arch.return_value = 'x86_64-linux-gnu'
+        self.addCleanup(patcher.stop)
+
+    @unittest.mock.patch('snapcraft.common.get_snapdir')
+    def test_config_snap_environment(self, mock_snapdir):
+        mock_snapdir.return_value = 'foo'
         config = snapcraft.yaml.Config()
-        environment = config.runtime_env('foo')
+        environment = config.snap_env()
         self.assertTrue('PATH="foo/bin:foo/usr/bin:$PATH"' in environment)
 
         # Ensure that LD_LIBRARY_PATH is present and it contains only the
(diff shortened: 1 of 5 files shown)


sergiusens:
This can probably be refactored some more to not need `build_env_for_part`

snapcraft/tests/test_yaml.py
```diff
@@ -399,6 +403,39 @@ def test_config_runtime_environment_ld(self):
...
+        self.assertTrue(
+            'LDFLAGS="-Lfoo/lib -Lfoo/usr/lib -Lfoo/lib/x86_64-linux-gnu '
+            '-Lfoo/usr/lib/x86_64-linux-gnu $LDFLAGS"' in environment)
+        self.assertTrue(
+            'PKG_CONFIG_PATH=foo/lib/pkgconfig:'
+            'foo/lib/x86_64-linux-gnu/pkgconfig:foo/usr/lib/pkgconfig:'
+            'foo/usr/lib/x86_64-linux-gnu/pkgconfig:foo/usr/share/pkgconfig:'
+            'foo/usr/local/lib/pkgconfig:'
+            'foo/usr/local/lib/x86_64-linux-gnu/pkgconfig:'
+            'foo/usr/local/share/pkgconfig:$PKG_CONFIG_PATH' in environment)
```
kyrofa: Are you actually wanting to test the ordering here, or do you only really only want to test that each of these are included in `PKG_CONFIG_PATH`? If you're not wanting to test the order, this test is perhaps overly fragile.

sergiusens: Well I also want to test that the output is correct and no missing colon gets in there. The order is fixed though (and should be as it determines the search order too).

kyrofa: > Well I also want to test that the output is correct and no missing colon gets in there.

You can do this without testing the order of them.

> The order is fixed though (and should be as it determines the search order too).

Good enough for me. Just wanted to make sure.

kyrofa:
Yeah this looks good to me!

snapcraft/tests/test_yaml.py
```diff
@@ -342,13 +342,19 @@ def setUp(self):
...
-    plugin: go
-    stage-packages: [fswebcam]
+    plugin: nil
 """)
 
-    def test_config_runtime_environment(self):
+        patcher = unittest.mock.patch('snapcraft.common.get_arch_triplet')
+        mock_arch = patcher.start()
+        mock_arch.return_value = 'x86_64-linux-gnu'
+        self.addCleanup(patcher.stop)
```
kyrofa: Nice.

sergiusens:
I am at a point where the only thing failing is ROS. Let's see how this goes.

sergiusens:
Everything builds now; most things built work too

elopio:
:+1: 
I see on the bug report that you requested an example. That would be awesome. Should we wait for it?

sergiusens:
I did a rebase of these onto the catkin one and it all works great.

PR #230, opened 2016-01-14
Fix catkin plugin python calls

sergiusens:
10.ros.sh which is sourced with setup.sh has a hardcoded
/usr/bin/python which needs to adapt to use the snaps
python interpreter.

LP: #1533297

### DIFF
--- a/snapcraft/tests/test_plugin_catkin.py
+++ b/snapcraft/tests/test_plugin_catkin.py
@@ -447,17 +447,27 @@ class CatkinPluginTestCase(tests.TestCase):
     @mock.patch.object(catkin.CatkinPlugin, 'run_output', return_value='foo')
     def test_finish_build(self, run_output_mock, run_mock):
         plugin = catkin.CatkinPlugin('test-part', self.properties)
-        os.makedirs(plugin.rosdir)
+        profile_dir = os.path.join(plugin.rosdir, 'etc', 'catkin', 'profile.d')
+        os.makedirs(profile_dir)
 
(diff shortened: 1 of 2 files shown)


sergiusens:
I now get 

```
[rosrun] Couldn't find executable named listener below /snaps/ros-example.sideload/IKKNUNfbERGD/opt/ros/indigo/share/beginner_tutorials
```

Maybe we can fix that here too

snapcraft/tests/test_plugin_catkin.py
```diff
@@ -447,17 +447,27 @@ def test_prepare_build(self):
...
         plugin._finish_build()
 
         # Verify that _setup_util.py had its shebang rewritten correctly
-        with open('{}/_setup_util.py'.format(
-                plugin.rosdir), 'r') as f:
+        with open(setup_util_path, 'r') as f:
+            self.assertEqual(f.read(), '#!/usr/bin/env python',
+                             'The shebang was not replaced as expected')
+
+        # Verify that _setup_util.py had its shebang rewritten correctly
```
kyrofa: Copy-paste in the comment.

kyrofa:
Small comment issue, but otherwise :+1: from me.

kyrofa:
Alright I'm good merging this as-is, since it fixes an obvious problem. However, I'm also good waiting to see if we can completely fix the problem.

kyrofa:
Hey wait... am I missing something? My 10.ros.sh doesn't have a shebang at all.

sergiusens:
I'm closing this for now so it doesn't accidentally merge, @kyrofa said he'd take a look.

PR #231, opened 2016-01-14
Wrap apps with the app name

sergiusens:
We wrap with the app name as the params are now globbed into the
wrapper

LP: #1534150

### DIFF
--- a/snapcraft/tests/test_meta.py
+++ b/snapcraft/tests/test_meta.py
@@ -147,7 +147,7 @@ class CreateTest(tests.TestCase):
 
         meta.create(self.config_data)
 
-        app1_wrapper_path = os.path.join(self.snap_dir, 'app1.sh.wrapper')
+        app1_wrapper_path = os.path.join(self.snap_dir, 'command-app1.wrapper')
         self.assertTrue(
             os.path.exists(app1_wrapper_path),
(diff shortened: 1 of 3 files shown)


snapcraft/tests/test_meta.py
```diff
@@ -209,6 +209,34 @@ def test_wrap_exe_must_write_wrapper(self, mock_assemble_env):
...
+        mock_assemble_env.return_value = """\
+PATH={0}/part1/install/usr/bin:{0}/part1/install/bin
+""".format(common.get_partsdir())
+
+        snapdir = common.get_snapdir()
+        os.mkdir(snapdir)
+
+        relative_exe_path = 'test_relexepath'
+        with open(os.path.join(snapdir, relative_exe_path), 'w'):
+            pass
```
kyrofa: Is this better than `open(os.path.join(snapdir, relative_exe_path), 'w').close()`?

kyrofa:
Small ignorable comment-- looks good to me.

elopio:
lgtm.
I want to get all merged to easily try all the binaries.

snapcraft/tests/test_meta.py
sergiusens: no it's not, I'll fix this in this test and the test I copied from, thanks for spotting this!

PR #232, opened 2016-01-14
Catkin plugin: Use in-snap python instead of OS-provided.

kyrofa:
This works pre-Xenial since python2 is provided on the image. However, it's been removed in Xenial which uncovered this bug. The fix simply rewrites all /usr/bin/python shebangs to use /usr/bin/env python, and also fixes up 10.ros.sh to use python from the path instead of /usr/bin/python.

LP: [#1533297](https://bugs.launchpad.net/snapcraft/+bug/1533297)

### DIFF
--- a/snapcraft/plugins/catkin.py
+++ b/snapcraft/plugins/catkin.py
@@ -234,25 +234,23 @@ deb http://${security}.ubuntu.com/${suffix} trusty-security main universe
             return '"' + ';'.join(paths) + '"'
 
         # Looking for any path-like string
-        pattern = re.compile(r'"(.*?/.*?)"')
-        for root, directories, files in os.walk(self.rosdir):
-            for fileName in files:
-                if fileName.endswith('Config.cmake'):
-                    with open(os.path.join(root, fileName), 'r+') as f:
-                        original = f.read()
-                        replaced = pattern.sub(rewrite_paths, original)
-                        if replaced != original:
-                            f.seek(0)
-                            f.truncate()
-                            f.write(replaced)
+        _search_and_replace(self.rosdir, re.compile(r'.*Config.cmake$'),
+                            re.compile(r'"(.*?/.*?)"'),
+                            rewrite_paths)
 
     def _finish_build(self):
-        # Fix the shebang in _setup_util.py to be portable
-        setup_util_file = os.path.join(self.rosdir, '_setup_util.py')
+        # Fix all shebangs to use the in-snap python.
+        _search_and_replace(self.rosdir, re.compile(r''),
+                            re.compile(r'#!.*python'),
+                            r'#!/usr/bin/env python')
+
+        # Also replace the python usage in 10.ros.sh to use the in-snap python.
+        setup_util_file = os.path.join(self.rosdir,
+                                       'etc/catkin/profile.d/10.ros.sh')
         if os.path.isfile(setup_util_file):
             with open(setup_util_file, 'r+') as f:
-                pattern = re.compile(r'#!.*python')
-                replaced = pattern.sub(r'#!/usr/bin/env python', f.read())
+                pattern = re.compile(r'/usr/bin/python')
+                replaced = pattern.sub(r'python', f.read())
                 f.seek(0)
                 f.truncate()
                 f.write(replaced)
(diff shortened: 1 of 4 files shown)


snapcraft/plugins/catkin.py
```diff
@@ -247,12 +247,30 @@ def rewrite_paths(match):
                             f.write(replaced)
 
     def _finish_build(self):
-        # Fix the shebang in _setup_util.py to be portable
-        setup_util_file = os.path.join(self.rosdir, '_setup_util.py')
+        # Fix all shebangs to use the in-snap python.
+        pattern = re.compile(r'#!.*python')
+        for root, directories, files in os.walk(self.rosdir):
+            for fileName in files:
+                with open(os.path.join(root, fileName), 'r+') as f:
```
elopio: I would extract this into a method.

kyrofa: Which part exactly-- the walking bit?

snapcraft/plugins/roscore.py
```diff
@@ -67,6 +68,8 @@ def build(self):
 
         os.chmod(service_wrapper, 0o755)
 
+        self._finish_build()
```
elopio: This feels like a bad name. It doesn't say what it does, it has many statements and no docstring.
Maybe something like _patch_snap_python? I'm not sure if that's better.

kyrofa: Well that's sort of the point though-- there are two generic functions, _prepare_build() and _finish_build(). The names indicate when they run, not what they do (since what they do is kind of a smattering of things).

snapcraft/plugins/roscore.py
```diff
@@ -81,3 +84,35 @@ def snap_fileset(self):
...
+        # Also replace the python usage in 10.ros.sh to use the in-snap python.
+        setup_util_file = os.path.join(rospath,
+                                       'etc/catkin/profile.d/10.ros.sh')
+        if os.path.isfile(setup_util_file):
+            with open(setup_util_file, 'r+') as f:
+                pattern = re.compile(r'/usr/bin/python')
+                replaced = pattern.sub(r'python', f.read())
+                f.seek(0)
+                f.truncate()
+                f.write(replaced)
```
elopio: I think this is clearly composed of two blocks, so I would extract two methods. One like _fix_shebangs or _patch_shebangs. The other something like _use_snap_python.
If you find really nice names for them, maybe you can get rid of the call to _finish_build (which has a name that doesn't say much) by directly calling these methods.
I like small methods with nice names. Lots of them. Sometimes I go too far, so feel free to disagree :)

snapcraft/tests/test_plugin_catkin.py
```diff
@@ -447,19 +447,50 @@ def test_prepare_build(self):
...
+        with open(irrelevant_file, 'r') as f:
+            self.assertEqual(f.read(), 'foo',
+                             'Expected the contents of "{}" to remain '
+                             'unchanged.'.format(irrelevant_file))
+
+        # Verify that the absolute path in 10.ros.sh was rewritten correctly
+        with open(ros_profile, 'r') as f:
+            self.assertEqual(f.read(), 'python foo',
+                             'The absolute path to python was not replaced as '
+                             'expected')
```
elopio: Huge test. What about making one test per #Verify comment?

snapcraft/plugins/catkin.py
elopio: first, the code inside the second for.
Then, maybe the two fors too, but this might be too much.

snapcraft/plugins/roscore.py
elopio: Personally, I would find it more readable if you called that smattering of things one by one instead of grouping them in a function that will have more than one responsibility and with a name that will force you to read the code to understand what it's doing.
But I'm just leaving this as a suggestion, I'm not blocking your PR on this. I understand why you have to patch things afterwards, and I'm not expecting that to be pretty at all.

sergiusens:
```
'RosCorePlugin' object has no attribute 'rosdir'
```

sergiusens:
Once that is fixed you will run into:
```
OSError: Parts 'catkin-tutorials' and 'roscore' have the following file paths in common which have different contents:
opt/ros/indigo/bin/catkin_find
opt/ros/indigo/bin/catkin_init_workspace
opt/ros/indigo/bin/catkin_make
opt/ros/indigo/bin/catkin_make_isolated
opt/ros/indigo/bin/catkin_package_version
opt/ros/indigo/bin/catkin_prepare_release
opt/ros/indigo/bin/catkin_test_results
opt/ros/indigo/bin/catkin_topological_order
opt/ros/indigo/bin/rosconsole
opt/ros/indigo/bin/rosgraph
opt/ros/indigo/etc/catkin/profile.d/10.ros.sh
opt/ros/indigo/lib/python2.7/dist-packages/genmsg/gentools.py
opt/ros/indigo/lib/python2.7/dist-packages/roslib/gentools.py
opt/ros/indigo/lib/python2.7/dist-packages/roslib/launcher.py
opt/ros/indigo/lib/python2.7/dist-packages/roslib/manifest.py
opt/ros/indigo/lib/python2.7/dist-packages/roslib/manifestlib.py
opt/ros/indigo/lib/python2.7/dist-packages/roslib/stack_manifest.py
opt/ros/indigo/lib/python2.7/dist-packages/roslib/stacks.py
opt/ros/indigo/lib/roslib/gendeps
```

Due to the global regexing (I guess)

kyrofa:
@sergiusens oops, knew I did that refactor too quickly. And no, the reason roscore has that clean step as well is to avoid just that.

snapcraft/tests/test_plugin_catkin.py
kyrofa: I instead refactored the test into two, one testing the shebang rewrite and the other testing the absolute path rewrite. It can be refactored further at some point if we get rid of _finish/_prepare_build(), but that's for a different PR.

snapcraft/plugins/roscore.py
kyrofa: I'll consider this in a future PR, but it's too heavy of a change right here.

PR #235, opened 2016-01-15
Added the network-listener capability.

elopio:
LP: #1534784

### DIFF
--- a/examples/gopaste/snapcraft.yaml
+++ b/examples/gopaste/snapcraft.yaml
@@ -9,7 +9,9 @@ apps:
     command: bin/gopaste
     daemon: simple
     caps:
+      - network-listener
       - network-service
+      - network-management
 
 parts:
   gopaste:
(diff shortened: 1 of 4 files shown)


sergiusens:
missing gopasted and 
tomcat-maven-webapp

elopio:
@sergiusens tomcat is already working.
gopasted after adding network-listener and network-management fails with:
sysctl: permission denied on key 'kernel.printk_ratelimit'
= Seccomp =
Time: Jan 15 19:12:58
Log: auid=4294967295 uid=0 gid=0 ses=4294967295 pid=1142 comm="gopasted" exe="/snaps/gopaste.sideload/IKNPLdMMXNPK/bin/gopasted" sig=31 arch=c000003e 93(fchown) compat=0 ip=0x7f801d9a9467 code=0x0
Syscall: fchown
Suggestions:
* don't copy ownership of files (eg, use 'cp -r --preserve=mode' instead of 'cp -a')
* adjust program to not use 'fchown'

elopio:
gopaste issue worked around in #237.

PR #240, opened 2016-01-18
Initialize CMAKE_PREFIX_PATH to [] for ros

sergiusens:
The ros plugins are setting _setup_util.py's CMAKE_PREFIX_PATH to
build time paths and later inserting the current path. This fails on
runtime due to confinement as the directories accessed don't exist.
(this is the case for armhf).

LP: #1535309


kyrofa:
I don't see the harm in this change, though I'm confused as to why this bites us on arm and not anything else. Access will obviously be denied regardless of the arch. But if it works, :+1: !

PR #243, opened 2016-01-19
Add doc about how to specify metadata (binaries, services or config).

dholbach:
LP: #1535905

### DIFF
--- a/docs/metadata.md
+++ b/docs/metadata.md
@@ -1,0 +1,72 @@
+# Adding metadata
+
+## Adding binaries
+
+Every part may build a binary that may need to be exposed in the system, with
+that in mind, the way to expose it would be to use the **binaries** keyword
+and declare each binary and its functionality. So, if there were two binaries,
+one called *app1* and another called *app2*, the extract would look like
+
+```yaml
+binaries:
+  app1:
+    exec: bin/app1
+  app2:
+    exec: opt/bin/app2
+    security-policy:
+      apparmor: app2.apparmor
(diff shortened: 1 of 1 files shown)


docs/metadata.md
```diff
@@ -0,0 +1,72 @@
...
+      apparmor: app2.apparmor
+      seccomp: app2.seccomp
+```
+
+So in the example, *service1* will declare its start to the relative path
+`bin/app1` and use the standard security template, while *service2* defines
+a different start to `bin/app2`, declares an explicit stop mechanism and uses
+customized **security-policy**, `app2.apparmor` and `app2.seccomp` are
+relative paths to these security artifacts within your project. More about
+security templating can be read about in the [security section] [sec].
```
kyrofa: I suggest adding a note that only simple daemons (i.e. non-forking) are supported in 1.x.

docs/metadata.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Snap config
+
+Snappy config is an entry point to configure your snap, in order to have the
+integration with the config system in place, the **config** keyword needs to
+be used. This is a relative path to the binary that will manage the
+configuration for the snap. If *config* is the part that would handle this
+and building its source creates `bin/config`, the extract would look like
```
kyrofa: This section needs a link to "read more about Snappy config" or something.

kyrofa:
Couple thoughts, but overall looks good.

sergiusens:
I'm :+1: on this.

PR #245, opened 2016-01-20
Added a unit tests for pulling bazaar sources from URIs.

fazerlicourice7:
I created a test for bazaar type sources. The test ensures that sources.py correctly recognizes and calls the correct methods for bazaar sources.

LP: #1536283

### DIFF
--- a/snapcraft/tests/test_sources.py
+++ b/snapcraft/tests/test_sources.py
@@ -335,7 +335,24 @@ class TestUri(tests.TestCase):
 
         for source in sources:
             with self.subTest(key=source):
+                options = tests.MockOptions(source=source)
+                snapcraft.sources.get(
+                    sourcedir='dummy',
+                    builddir='dummy',
+                    options=options)
 
+                mock_pull.assert_called_once_with()
+                mock_pull.reset_mock()
+
+    @unittest.mock.patch('snapcraft.sources.Bazaar.pull')
+    def test_get_bzr_source_from_uri(self, mock_pull):
+        sources = [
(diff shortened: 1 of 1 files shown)


fazerlicourice7:
I'm trying to squash the merge commit, but when I attempt to rebase I can't see it as a commit.

debian/changelog
```diff
@@ -86,6 +86,7 @@ snapcraft (2.0) xenial; urgency=medium
 
   [ Vamshi Balanaga ]
   * Added test for git sources in sources.py (LP: #1534411)
+  * Added test for bzr sources in sources.py (LP: #1536283)
```
elopio: You shouldn't edit the changelog of the released version.
We could open a new section for the unreleased 2.1 version, but we are currently trying to do that automatically with git-buildpackage, so just remove this line, please.

elopio:
The test looks good to me. I will let @kyrofa help you with the squash, because I might just make things worst.
Once you are able to rewrite the commit, please make sure to write it something like this:

Added a unit tests for pulling bazaar sources from URIs.

LP: #1536283

(note the capital letter at the start, and the period at the end. And it's not necessary to mention the file you changed but I consider it good to give a little more context of what you are doing)
We are taking care a lot about commit messages now because we are trying to autogenerate the changelog from them.

Thanks again @fazerlicourice7!

kyrofa:
Looks good, though since the PR title turns into the merge commit message, let's leave it clean and leave off the bug ref there (just in the PR description and commit message is perfect).

fazerlicourice7:
Thanks.

PR #248, opened 2016-01-21
Git source: Make a shallow clone.

kyrofa:
This PR resolves LP: [#1536696](https://bugs.launchpad.net/snapcraft/+bug/1536696) by making the git source use a shallow (depth of 1) clone rather than fetching the entire history.


elopio:
:+1:

PR #249, opened 2016-01-21
Git source: Support submodules.

kyrofa:
This PR resolves LP: [#1536698](https://bugs.launchpad.net/snapcraft/+bug/1536698) by cloning recursively (thus retrieving submodules), and updates them when pulled as well.


sergiusens:
The code looks good, can we maybe have an example?

sergiusens:
Or better yet, an integration test, it shouldn't be hard to create and do everything locally (check the other git integration tests for inspiration).

PR #251, opened 2016-01-22
Remove 'vendor' from snapcraft init template.

dholbach:
LP: #1536860

### DIFF
--- a/snapcraft/commands/init.py
+++ b/snapcraft/commands/init.py
@@ -37,11 +37,9 @@ logger = logging.getLogger(__name__)
 
 _TEMPLATE_YAML = r'''name: # the name of the snap
 version: # the version of the snap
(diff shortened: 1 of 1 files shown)


kyrofa:
Oops, that's a little embarrassing, thanks for catching that! Perhaps we should remove `icon` as well and make this as bare-bones as possible. Thoughts?

elopio:
:+1:
I'm not sure about removing the icon. Do we want to make it bare-bones, or to showcase all the features? Maybe we can put in a comment that it is optional and can be removed.

kyrofa:
@elopio yeah I'm good with that. @dholbach can you just add a "This is optional." sentence to the end of the icon comment?

kyrofa:
We should add a test to validate this yaml, but that's a different PR. Thanks @dholbach!

PR #253, opened 2016-01-22
Autotools plugin: Call make before make install

kyrofa:
This PR resolves LP: [#1536700](https://bugs.launchpad.net/snapcraft/+bug/1536700) by making the Autotools plugin call `make` before it calls `make install`.


elopio:
ok. I think it's not worth it to make an integration or example test just for this change.
But it would be nice to have an integration one that requires all the steps of make. We can check about this next week
:+1:

kyrofa:
@elopio sounds good, I think we can swing that. I don't think it would make the greatest example, though-- so probably just an integration test.

PR #256, opened 2016-01-25
make sure that version is set correctly no matter what (version for 2…

dholbach:
….0.1 release was '2.0')

LP: #1537580

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,14 +15,25 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import codecs
 import os
+import re
 
 from setuptools import setup
+
+version = 'devel'
+# look/set what version we have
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -15,14 +15,24 @@
...
 
 from setuptools import setup
 
+# look/set what version we have
+changelog = 'debian/changelog'
+if os.path.exists(changelog):
+    head = codecs.open(changelog, encoding='utf-8').readline()
+    match = re.compile('.*\((.*)\).*').match(head)
+    if match:
+        version = match.group(1)
```
kyrofa: I know the possibility of this happening is small, but should `version` default to "devel" so if this fails there's still a valid version?

dholbach: Good thinking.

kyrofa:
This looks excellent, thank you!

PR #257, opened 2016-01-25
Fix apt cache FD leakage and performance boost lp:1537705

asac:


asac:
test coverage decrease... would be nice if someone deep into code could write a simple test for this... if you need test case yaml its in the lp bug https://bugs.launchpad.net/snapcraft/+bug/1537705

asac:
lool pointed out that this patch does not close cache when pkg isnt in cache... we could make this cleaner, but practically we probably always want to bail if someone asks for a build-package that isnt even known to cache, no?

snapcraft/repo.py
```diff
@@ -58,11 +58,16 @@
...
 def install_build_packages(packages):
+    seen = set()
+    seen_add = seen.add
+    unique_packages = [x for x in packages if not (x in seen or seen_add(x))]
     new_packages = []
-    for pkg in packages:
+    for pkg in unique_packages:
         try:
-            if not apt.Cache()[pkg].installed:
+            apt_cache = apt.Cache()
```
sergiusens: I wonder if something like this is allowed (apt cache being context aware, I think it is), so mind replacing that with

```python
    if not apt_cache[pkg].installed:
        new_package.append(pkg)
```

It might also be more performant to open the apt cache before looping over the packages.

asac: uploaded new rev with comments addressed

asac:
ok, ran static/unit tests locally on latest push and all green... hope good enough :)

snapcraft/repo.py
```diff
@@ -58,15 +58,19 @@
 
 
 def install_build_packages(packages):
+    seen = set()
+    seen_add = seen.add
+    unique_packages = [x for x in packages if not (x in seen or seen_add(x))]
```
sergiusens: it seems my comment from before did not stick :hushed: 

Is this a way to get only one item? If so, instead of

```python
seen = set()
seen_add = seen.add
unique_packages = [x for x in packages if not (x in seen or seen_add(x))]
```

do 

```python
unique_packages = set(packages)
```

Or bellow in

```python
for pkg in unique_packages
```

Change that to

```python
for pkg in set(packages)
```

asac: pushed fix

sergiusens:
I'm closing this in favor of #268 which still keeps @asac as the author but squashed.

PR #260, opened 2016-01-26
add note explaining that only simple daemons are supported in snapcra…

dholbach:
Fixed two small issues @kyrofa pointed out in https://github.com/ubuntu-core/snapcraft/pull/243


kyrofa:
:+1: Just waiting on the tools.

PR #262, opened 2016-01-26
Refactored the local source pull.

elopio:
Added full test coverage to the method.

LP: #1538250


elopio:
This makes me think that having a source and a source_dir makes things confusing.

kyrofa:
@elopio yeah I agree-- takes some tracing back into the code to figure out the difference. We should rename them-- though if we started using docstrings for the functions that describe the parameters that would improve things as well. We should do both.

But this looks good to me.

PR #264, opened 2016-01-26
Make plugin: Support non-standard makefile name.

kyrofa:
This PR fixes LP: [#1500759](https://bugs.launchpad.net/snapcraft/+bug/1500759) by adding a "makefile" property to the Make plugin, which, if specified, results in a `make -f <makefile>` instead of just `make`. This supports projects with makefiles named e.g. makefile.linux, etc.


elopio:
:+1: 
An integration tests without the mocks wouldn't hurt :)

kyrofa:
@elopio good idea-- done.

PR #268, opened 2016-01-27
fix FD leakage by closing apt cache in repo.py

sergiusens:
LP: #1537705


sergiusens:
I'm landing @asac's change here.

PR #269, opened 2016-01-27
Go plugin: Unset all Go-related environment variables.

kyrofa:
This PR fixes LP: [#1537728](https://bugs.launchpad.net/snapcraft/+bug/1537728) by making sure the Go plugin runs in a clean Go environment.

### DIFF
--- a/snapcraft/plugins/go.py
+++ b/snapcraft/plugins/go.py
@@ -35,6 +35,7 @@ Additionally, this plugin uses the following plugin-specific keywords:
 
 import os
 import shutil
+import subprocess
 
 import snapcraft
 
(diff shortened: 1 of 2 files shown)


snapcraft/plugins/go.py
```diff
@@ -122,6 +123,17 @@ def _remote_build(self):
...
+        env = os.environ.copy()
+        env['GOPATH'] = self._gopath
+
+        for variable in ['GOARCH', 'GOBIN', 'GOCHAR', 'GOEXE', 'GOHOSTARCH',
+                         'GOHOSTOS', 'GOOS', 'GORACE', 'GOROOT', 'GOTOOLDIR',
+                         'TERM', 'CC', 'GOGCCFLAGS', 'CXX', 'CGO_ENABLED']:
+            if variable in env:
+                del env[variable]
+
+        subprocess.check_call(cmd, cwd=self._gopath_src, env=env)
```
elopio: What if we extend the parent self.run to understand the env kwarg and pass it to check_call?
I'm discussing about something similar for @sergiusens' #270.

sergiusens: @elopio That won't work for when this plugin's vars is used by its dependant ones.

sergiusens: Oh, I'm just seeing you are using `check_call` directly; this will miss out on any LDFLAG or CFLAG you might depend on.

I think that the better solution to this problem is to set `GOROOT` and others in the call the `env` if not caring at all either (cleanbuild should solve this mostly)

kyrofa:
Yeah... this opens a can of worms. The real solution to this is cleanbuild, which will be completed soon. Closing this.

PR #270, opened 2016-01-27
Support $SNAPCRAFT_STAGE for parts

sergiusens:
When a part is loaded, instances of $SNAPCRAFT_STAGE
are replaced with the absolute path to the stage directory
during runtime.

LP: #1538688

### DIFF
--- a/docs/snapcraft-advanced-features.md
+++ b/docs/snapcraft-advanced-features.md
@@ -127,6 +127,25 @@ will be built after `libpipeline`. Especially if you need specific
 functionality during a build or as part of checks during the `stage` phase,
 this will be handy.
 
+If any part built using the `after` keyword needs to explicitly access
+assets in the stage directory with configuration flags (e.g.; `configure`
+in the case of autotools) it can use of the `SNAPCRAFT_STAGE` environment
+variable, like this:
+
+```yaml
+parts:
+    my-part:
+        plugin: autotools
+        source: .
+        configFlags:
+            - --with-swig $SNAPCRAFT_STAGE/swig
+        after:
+            - swig
+    swig:
+        plugin: autotools
+        source: ./swig
+```
+
 ### Re-using parts
 
 With snapcraft we want to make it easy to learn from other app vendors and
--- a/integration_tests/snaps/stage_env/snapcraft.yaml
+++ b/integration_tests/snaps/stage_env/snapcraft.yaml
@@ -1,0 +1,19 @@
+name: test-package
+version: 0.1
+summary: try out $SNAPCRAFT_STAGE
+description: |
+  The cmake-project expects to have a copied directory in the stage directory
+  to consume during its build time.
+
+parts:
+  cmake-project:
+    plugin: cmake
+    source: .
+    configflags:
+        - -DDEP_DIR=$SNAPCRAFT_STAGE/copied
+    after:
+      - copy
+  copy:
+    plugin: copy
+    files:
+        copied: copied
--- a/snapcraft/tests/test_pluginhandler.py
+++ b/snapcraft/tests/test_pluginhandler.py
@@ -355,3 +355,193 @@ class CollisionTestCase(tests.TestCase):
             raised.exception.__str__(),
             "Parts 'part2' and 'part3' have the following file paths in "
             "common which have different contents:\n1\na/2")
+
+
+class StageEnvTestCase(tests.TestCase):
+
+    def test_string_replacements(self):
+        stagedir = common.get_stagedir()
+
+        replacements = (
+            (
+                'no replacement',
+                'snapcraft_stage/usr/bin',
+                'snapcraft_stage/usr/bin',
+            ),
+            (
+                'replaced start',
+                '$SNAPCRAFT_STAGE/usr/bin',
+                '{}/usr/bin'.format(stagedir),
+            ),
+            (
+                'replaced between',
+                '--with-swig $SNAPCRAFT_STAGE/usr/swig',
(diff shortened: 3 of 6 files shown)


sergiusens:
I'll get an integration test in tonight.

docs/snapcraft-advanced-features.md
```diff
@@ -127,6 +127,25 @@ will be built after `libpipeline`. Especially if you need specific
 functionality during a build or as part of checks during the `stage` phase,
 this will be handy.
 
+If any part built through after needs to explicitly access assets in the
```
elopio: "built through after" is hard to read.
Maybe: If a part built using the `after` keyword needs ...

docs/snapcraft-advanced-features.md
```diff
@@ -127,6 +127,25 @@ will be built after `libpipeline`. Especially if you need specific
...
+stage directory with configuration flags (e.g.; `configure` in the case
+of autotools) it can through the use of the `SNAPCRAFT_STAGE` environment
```
elopio: replace "it can through the use of the" with "it can use the"

docs/snapcraft-advanced-features.md
```diff
@@ -127,6 +127,25 @@ will be built after `libpipeline`. Especially if you need specific
...
+flag, such that the following is possible:
```
elopio: I think it's a variable, not a flag.

elopio:
why is this needed? integration_tests/snaps/stage_env/copied/s

docs/snapcraft-advanced-features.md
elopio: I would replace "such that the following is possible" with "like this"

snapcraft/tests/test_pluginhandler.py
```diff
@@ -355,3 +355,193 @@ def test_collisions_between_two_parts(self):
...
+                    '2': '--with-swig {}/usr/swig'.format(stagedir),
+                },
+            ),
+        )
+
+        for test_name, subject, expected in replacements:
+            self.subTest(key=test_name)
+            self.assertEqual(pluginhandler._expand_env(subject), expected)
+
+    def test_complex_data(self):
```
elopio: test_string_replacement_with_complex_data

integration_tests/snaps/stage_env/snapcraft.yaml
```diff
@@ -0,0 +1,19 @@
...
+description: |
+  The cmake-project expects to have a copied directory in the stage directory
+  to consume during its build time.
+
+parts:
+  cmake-project:
+    plugin: cmake
+    source: .
+    configflags:
+        - -DVAR=$SNAPCRAFT_STAGE/copied
```
elopio: I hate cmake. I don't understand how it translates VAR into DEP_DIR.

elopio:
This makes sense and looking at the tests it seems to work.
I think I would prefer instead of replacing the vars ourselfs, to find a way to set the variables while the step is running. So you would call something like SNAPCRAFT_STAGE=stage; cmake -DVAR=$SNAPCRAFT_STAGE/copied. Could we add it to the env argument of subprocess.check_output?
Just asking. I'm ok with this if you prefer it.

sergiusens:
@elopio any idea why this and #269 dropped coverage so badly? It reports 80% now :-/

integration_tests/snaps/stage_env/snapcraft.yaml
sergiusens: you caught an actual error! I changed var names after testing, but only half way through. I'm fixing this. And since it has a `/` at the end it didn't result in an error

sergiusens:
@elopio *why is this needed? `integration_tests/snaps/stage_env/copied/s`*

git did not let me add an empty dir, but from the error above, now I'm going to use it as well.

sergiusens:
@elopio so why no envvar? I guess it is because I would like to improve that code before polluting it more. Every env var used now involves a write to a file.

I'll let @kyrofa weigh in on this.

docs/snapcraft-advanced-features.md
```diff
@@ -127,6 +127,25 @@ will be built after `libpipeline`. Especially if you need specific
 functionality during a build or as part of checks during the `stage` phase,
 this will be handy.
 
+If any part built using the `after` keyword needs to explicitly access
+assets in the stage directory with configuration flags (e.g.; `configure`
```
kyrofa: I don't think the semicolon is necessary here.

kyrofa:
I agree that an environment variable might make the code a little cleaner since that stuff already exists. @sergiusens I understand where you're coming from regarding the file writes. However, the environment variable handling isn't going away, even if we're able to refactor the run functions, so I feel like this would be a safe addition. It just needs to be in a place that doesn't cause it to end up in any wrappers, heh.

I'm fine with this as well, I just think environment variables may have already solved this problem for us.

docs/snapcraft-advanced-features.md
sergiusens: I picked this habit from scientific writing classes for my thesis.

Although they refer to using a comma instead of a semicolon, a separator is required http://english.stackexchange.com/questions/16172/should-i-always-use-a-comma-after-e-g-or-i-e

kyrofa: Hahaha, okay you win! :grinning:

sergiusens:
@kyrofa @elopio so merging this now. If we need something fancier or to rely on the OS, then we can look into this.

PR #273, opened 2016-01-28
Use the git http protocol in the examples tests.

elopio:
LP: #1539139


elopio:
Requires #272.

sergiusens:
It makes no sense to my poor little brain how coverage can drop for a thing we no longer consider for coverage :weary: 

Looks good :+1:

PR #274, opened 2016-01-28
make sure otp is included in login request only if it's not empty

ricardokirkner:
LP: #1539208


kyrofa:
Thanks @ricardokirkner! Looks good to me, just waiting on the tools. Mind also adding a `#` to the bug reference in the commit message?

PR #275, opened 2016-01-28
Remove legacy formatting and implement skills

sergiusens:
This adds functionality for the migration-skill and effectively removes
support for caps, security-override, security-policy and security-template
as part of the app definition.

In addition to that, package.yaml and readme.md support is also removed.


sergiusens:
@mvo5 mind taking a look?

sergiusens:
Closing in favor of #278

PR #276, opened 2016-01-28
Update debian/changelog for the 1.0.1 release.

kyrofa:
Also update the version number in setup.py. I _think_ that's all of them :smiley: .

Please merge only if this is the last commit for 1.0.1.


sergiusens:
Looks good, thanks for remembering to change the version! :+1:

PR #278, opened 2016-01-28
Implement skills (the migration-skill) and remove support for legacy assets in meta

sergiusens:
This adds functionality for the migration-skill and effectively removes
support for caps, security-override, security-policy and security-template
as part of the app definition.

In addition to that, package.yaml and readme.md support is also removed.

LP: #1539122


schema/snapcraft.yaml
```diff
@@ -54,6 +54,32 @@ definitions:
...
+  migration-skill:
+    type: object
+    additionalProperties: false
+    properties:
+        type:
+          type: string
+          enum:
+            - migration-skill
+        security-policy:
+          ref: "#definitions/security-policy"
```
zyga: ref or $ref?

snapcraft/meta.py
```diff
@@ -140,16 +140,18 @@ def _copy(meta_dir, relpath, new_relpath=None):
...
-    for app in apps:
-        if 'security-policy' in apps[app]:
-            apps[app]['security-policy']['apparmor'] = \
-                _copy(meta_dir, apps[app]['security-policy']['apparmor'])
-            apps[app]['security-policy']['seccomp'] = \
-                _copy(meta_dir, apps[app]['security-policy']['seccomp'])
+def _copy_security_profiles(meta_dir, uses):
+    # TODO: remove once skills are implemented.
+    for slot in uses:
+        if uses[slot]['type'] != 'migration-skill':
```
zyga: if slot name itself is migration-skill then this should work too (type defaults to slot name)

snapcraft/tests/test_meta.py
```diff
@@ -200,12 +202,114 @@ def test_create_meta_with_app_with_security_policy(self):
...
+        open(os.path.join(os.curdir, 'stub-sec'), 'w').close()
+        self.config_data['apps'] = {
+            'app1': {
+                'command': 'app1.sh',
+                'uses': ['migration'],
+            },
+        }
+        self.config_data['uses'] = {
+            'migration': {
+                'type': 'migration-skillz',
```
zyga: skillz?

zyga:
A few comments inline, otherwise +1

schema/snapcraft.yaml
sergiusens: Ah, nice catch

snapcraft/meta.py
sergiusens: Right, I made type mandatory in the schema (I think). Will look into how this can be done

snapcraft/tests/test_meta.py
sergiusens: Yeah, I'll change the type to something more obvious

sergiusens:
@zyga thanks for the review

schema/snapcraft.yaml
sergiusens: Now I see this is a classic press of `x` by mistake :disappointed:

snapcraft/meta.py
```diff
@@ -123,10 +122,11 @@ def _compose_snap_yaml(meta_dir, config_data):
...
     if 'apps' in config_data:
-        apps = config_data['apps']
-        apps = _wrap_apps(apps)
-        snap_yaml['apps'] = \
-            _copy_security_profiles(meta_dir, apps)
+        snap_yaml['apps'] = _wrap_apps(config_data['apps'])
+
+    if 'uses' in config_data:
+        snap_yaml['uses'] = \
+            _copy_security_profiles(meta_dir, config_data['uses'])
```
kyrofa: Should you also copy over the `offers` here since, as you say, you don't want to block them from being used even if we don't have a good definition for them?

sergiusens: There are no `offers` to copy. The `migration-skill` only applies to `uses`. In theory it could for frameworks, but we removed support for them in snapcraft.

This code will go away as soon as skill `type`s start landing into the `os` and the `migration-skill` is gone.

kyrofa:
Yeah this looks good to me. Consider changing the PR title, but :+1: !

schema/snapcraft.yaml
```diff
@@ -152,14 +168,18 @@ properties:
...
-            $ref: "#definitions/security-template"
-          caps:
-            $ref: "#definitions/security-caps"
+          uses:
+            type: array
+            minitems: 1
+            uniqueItems: true
+            items:
+              type: string
+          offers:
```
mvo5: Fwiw, we don't have code for this in snappy yet.

snapcraft/tests/test_yaml.py
```diff
@@ -661,6 +661,12 @@ def test_valid_security_policy_for_apps(self):
         self.data['apps'] = {
             'app1': {
                 'command': 'binary',
+                'uses': ['migration'],
```
mvo5: It says uses "migration" but the skill is called "file-migration2 below?

mvo5:
Looks good, just one question in the tests.

snapcraft/tests/test_yaml.py
sergiusens: Thanks for catching this!

sergiusens:
Confirmed working on the latest image

PR #279, opened 2016-01-29
Try to capture what the articles are about in a more concise and clearer way.

dholbach:
LP: #1539502


sergiusens:
:+1:

sergiusens:
Don't drop coverage so badly next time (*Coverage decreased (-0.0%) to 93.269%*) :wink:

PR #280, opened 2016-01-29
debian/control: add python3-responses

sergiusens:
python3-responses is needed to run the recently added unit tests
for store related comms.

LP: #1539199


kyrofa:
:+1: . Let's watch the coverage drop.

PR #282, opened 2016-01-29
Make package installation dynamic to reduce deps

sergiusens:
debian/control has a huge amount of dependencies that could very well
be installed depending on the project. There are also dependencies
included that are not relevant anymore.

The work here:

- cleans up debian/control
- adds logic to install the correct vcs tool depending on the source
  in snapcraft.yaml
- makes sure all the examples have complete dependencies
- cleans up .travis.yml

LP: #1539146

### DIFF
--- a/snapcraft/common.py
+++ b/snapcraft/common.py
@@ -17,6 +17,7 @@
 # Data/methods shared between plugins and snapcraft
 
 import os
+import platform
 import subprocess
 import tempfile
 import urllib
@@ -29,7 +30,6 @@ _plugindir = _DEFAULT_PLUGINDIR
 _DEFAULT_SCHEMADIR = '/usr/share/snapcraft/schema'
 _schemadir = _DEFAULT_SCHEMADIR
 _arch = None
-_arch_triplet = None
 
 env = []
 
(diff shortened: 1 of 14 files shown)


snapcraft/common.py
```diff
@@ -61,21 +61,40 @@ def run_output(cmd, **kwargs):
...
+            'https://bugs.launchpad.net/snapcraft'.format(platform.machine()))
+
+
+_GNU_ARCH_TRIPLET = {
+    'armv7l': 'arm-linux-gnueabihf',
+    'aarch64': 'aarch64-linux-gnu',
+    'i686': 'i386-linux-gnu',
+    'ppc64le': 'powerpc64le-linux-gnu',
+    'x86_64': 'x86_64-linux-gnu',
+}
```
kyrofa: This information is very closely related to the `_DEB_ARCH` above. Would the codebase be less prone to an update in one spot and forgetting to update the other if you stored this information within the same container? For example, a nested map of the structure:

```python
{
   'armv7l': {
      'arch': 'armhf',
      'triplet': 'arm-linux-gnueabihf',
   }
   # ...
}
```

You get the idea. Less duplication, still clear?

kyrofa:
Small comment about combining the data structures for the arch-specific stuff, but looks good to me :+1: .

snapcraft/common.py
sergiusens: yeah, let me do that

PR #283, opened 2016-01-29
Upload: Clean up output.

kyrofa:
This PR fixes LP: [#1538692](https://bugs.launchpad.net/snapcraft/+bug/1538692) by change the OTP prompt to be more clear, and adding progress indicators. The upload output now looks something like this:

    $ snapcraft upload

    Uploading foo_1.7_amd64.snap [==============================================================] 100%
    Checking package status... | # <-- this is an animated pinwheel indicator

    Application uploaded successfully (as revision 6)
    Please check out the application at: https://myapps.developer.ubuntu.com/dev/click-apps/1337/

Note that the test logger output is not cleaned here-- I bisected it back to 23290980fe9dbbc407cb0edb3cd1a22f781d411b but I can't figure out what's causing it.


sergiusens:
Nice work! I'll try and review soon; don't worry about the unit test dilemma; I have it covered

snapcraft/commands/login.py
```diff
@@ -47,7 +47,8 @@ def main(argv=None):
     print('Enter your Ubuntu One SSO credentials.')
     email = input('Email: ')
     password = getpass.getpass('Password: ')
-    otp = input('OTP: ')
+    otp = input('One-time password (just press enter if you don\'t use '
+                'two-factor authentication): ')
```
elopio: I like this!

snapcraft/storeapi/_upload.py
```diff
@@ -33,6 +38,7 @@
 )
 
 
+logging.getLogger("requests").setLevel(logging.WARNING)
```
elopio: we generally call the logger __file__. Any special reason for using "requests" here instead?

snapcraft/storeapi/_upload.py
```diff
@@ -48,7 +54,8 @@ def upload(binary_filename, metadata_filename='', metadata=None, config=None):
         return
     name = match.groupdict()['name']
 
-    logger.info('Uploading files...')
+    logger.info('')
```
elopio: Why did you keep this one? I might be missing something, but I fail to see a reason for logging nothing :)

snapcraft/storeapi/_upload.py
```diff
@@ -65,19 +71,23 @@ def upload(binary_filename, metadata_filename='', metadata=None, config=None):
...
     revision = result.get('revision')
 
+    logger.info('')
+
     if success:
-        logger.info('Application uploaded successfully.')
+        message = 'Application uploaded successfully'
         if revision:
-            logger.info('Uploaded as revision %s.', revision)
+            message += ' (as revision {})'.format(revision)
```
elopio: :100:

snapcraft/storeapi/_upload.py
```diff
@@ -92,26 +102,52 @@ def upload_files(binary_filename, config=None):
...
+        progress_bar = ProgressBar(
+            widgets=['Uploading {} '.format(binary_filename),
+                     Bar(marker='=', left='[', right=']'), ' ', Percentage()],
+            maxval=os.path.getsize(binary_filename)).start()
+
+        # This callback simply updates the progress bar as the upload
+        # progresses.
+        def upload_progress_callback(monitor):
+            if monitor.bytes_read <= binary_file_size:
+                progress_bar.update(monitor.bytes_read)
```
elopio: Could you extract a function for this block? maybe call it get_upload_progress_callback, that receives the binary_filename and returns the callback func.

snapcraft/storeapi/_upload.py
```diff
@@ -125,9 +161,8 @@ def upload_files(binary_filename, config=None):
             'An unexpected error was found while uploading files.')
         result['errors'] = [str(err)]
     finally:
-        # make sure to close any open files used for request
-        for fd in files.values():
-            fd.close()
+        # Close the open file
+        binary_file.close()
```
elopio: @sergiusens copied the same block in #285. I prefer your removal of the confusing dictionary of 1 element, so I think this one should land first.

elopio:
Nice work indeed.

snapcraft/storeapi/_upload.py
kyrofa: Ah, yeah notice I don't save the logger off here-- this line was for shutting up the requests lib. But perhaps this isn't the best place to put such a line?

snapcraft/storeapi/_upload.py
kyrofa: This is only here to print a newline before the progress bar (it can't be contained within the progress bar or it keeps printing newlines). I just thought it looked a little nicer with some spacing.

snapcraft/storeapi/_upload.py
kyrofa: I extracted it, but I thought binding the arguments with `functools.partial` was even better than returning a callback.

snapcraft/commands/login.py
kyrofa: It's a bit verbose as we discussed, but I think it's a good intermediate solution until the uploader is smart enough to know whether or not it's actually required.

snapcraft/storeapi/_upload.py
sergiusens: I say do a `print` instead

snapcraft/storeapi/_upload.py
sergiusens: Right, this may need to be refactored to support multiple snap uploads when library snaps come (hope we can have something for next week).

snapcraft/storeapi/_upload.py
kyrofa: Done.

snapcraft/storeapi/_upload.py
kyrofa: Ah, yeah we'll have to discuss how we want to do the progress bar in that case.

snapcraft/storeapi/_upload.py
elopio: We have log.py, with a configure method.

snapcraft/storeapi/_upload.py
elopio: Yes, the progress goes to stdout, not to the log. Thanks for the change.

elopio:
lgtm

sergiusens:
:+1:

PR #284, opened 2016-01-30
Use the log fixture for logging by default for ut

sergiusens:
Unit tests now have the logging fixture set by default to avoid leaking
messages to the console

LP: #1539817


elopio:
:+1:

kyrofa:
Thanks for this :smiley: .

PR #285, opened 2016-01-30
Close fd on invalid credentials for upload

sergiusens:


sergiusens:
this depends on #284

elopio:
ok to me.

ricardokirkner:
LGTM. Would it no be possible to add a test to ensure file is closed after error from wrong credentials?

kyrofa:
Looks fine, though check the title before merge.

kyrofa:
Note that this bug is fixed in PR #283

PR #287, opened 2016-02-03
Use docopt to handle example tests arguments.

elopio:
LP: #1533395

### DIFF
--- a/examples_tests/__main__.py
+++ b/examples_tests/__main__.py
@@ -14,9 +14,28 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import argparse
+"""Snapcraft examples tests.
+
+Usage:
+  examples_tests [--skip-install] [--ip IP_OR_HOSTNAME]
+                 [--port PORT_NUMBER] [--filter REGEXP]
+                 [--subunit]
+
+Options:
+  --skip-install       skip the tests that install the example snaps into a
+                       snapp test bed.
+  --ip IP_OR_HOSTNAME  IP of the testbed. If no IP is passed, a virtual
+                       machine will be created for the test.
+  --port PORT_NUMBER   SSH port of the testbed. Defaults to use port 22.
+  --filter REGEXP      a regular expression to filter the examples to test.
+  --subunit            generate subunit results.
+
+"""
+
 import logging
 import sys
+
+import docopt
 
 from examples_tests import tests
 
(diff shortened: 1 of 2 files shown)


snappy-m-o:
test

kyrofa:
Looks good to me :+1: .

examples_tests/__main__.py
```diff
@@ -14,48 +14,47 @@
...
-    if args.port:
-        tests.config['port'] = args.port
-    if args.filter:
-        tests.config['filter'] = args.filter
+    if arguments['--ip']:
+        tests.config['ip'] = arguments['--ip']
+    if arguments['--port']:
+        tests.config['port'] = arguments['--port']
+    if arguments['--filter']:
+        tests.config['filter'] = arguments['--filter']
```
sergiusens: Is there a problem down under if we just do this:
```python
test.config['ip'] = arguments['--ip']
tests.config['port'] = arguments['--port']
tests.config['filter'] = arguments['--filter']
```

then for `test.config` you don't need to check for the existence of the key but the key directly i.e. `if test.config['filter']`

Does this make sense, I have plenty of assumptions in this comment.

sergiusens:
Looks good, but what is **test** or why is it still running?

elopio:
test is not running, it was just a test that I could communicate to this PR from jenkins. Now I don't know how to delete it :)

examples_tests/__main__.py
elopio: Only the difference between having a dictionary element with None as a value and not having it.
>>> {}.get('port', 22)
22
>>> {'port': None}.get('port', 22)
>>>

I'll make the change to see if you like it better. I slightly prefer to put the complexity here than on the test; but on the other hand it sounds better to handle '', None and unexisting.

examples_tests/tests.py
```diff
@@ -214,14 +214,17 @@ def setUpClass(cls):
         cls.temp_dir = tempfile.mkdtemp()
         if not config.get('skip-install', False):
             ip = config.get('ip', None)
-            if ip is None:
+            if not ip:
```
sergiusens: I like this even if not relevant :smile:

examples_tests/tests.py
```diff
@@ -214,14 +214,17 @@ def setUpClass(cls):
...
                 cls.testbed_ip = 'localhost'
                 cls.testbed_port = '8022'
                 cls.vm_process = _start_snappy_testbed(
                     cls.temp_dir, cls.testbed_port)
             else:
                 cls.testbed_ip = ip
-                cls.testbed_port = config.get('port', '22')
+                port = config.get('port', None)
```
sergiusens: why not change `None` to `'22'` and discard the 
```python
if not port:
    port = '22'
```

So all these lines look like
```python
cls.testbed_port = config.get('port', '22')
```

elopio: Because when the argument is not passed docopt sets the port to None, and {'port': None}.get('port', 22) returns None, not 22.

sergiusens: Ah; for some reason I prefer
```python
cls.testbed_port = config.get('port', None) or '22'
```

But I'm :+1: either way!

elopio: That's nice. Done.

sergiusens:
So :+1: then

PR #288, opened 2016-02-03
Make requests-toolbelt and progressbar a Depends.

kyrofa:
This PR fixes LP: [#1541349](1541349), where if Snapcraft is installed via the .deb and a user tries to login/upload a package, they get modules not found.


debian/control
```diff
@@ -32,8 +32,10 @@ Depends: python3-apt,
          python3-jsonschema,
          python3-lxml,
          python3-pkg-resources,
+         python3-progressbar
```
sergiusens: missing `,`

debian/control
```diff
@@ -32,8 +32,10 @@ Depends: python3-apt,
...
          python3-requests,
          python3-requests-oauthlib,
+         python3-requests-toolbelt
```
sergiusens: missing `,`

sergiusens:
:+1: but I want @dholbach to also ack :wink:

dholbach:
:+1: - go go go! :smile:

PR #296, opened 2016-02-04
Improve snapcraft.yaml validation errors.

kyrofa:
This PR fixes LP: [#1524663](https://bugs.launchpad.net/snapcraft/+bug/1524663) by improving YAML validation errors across the board. Specifically:

-  Improve error message if icon isn't a .png or .svg, or doesn't exist.

    Was: "'icon.png' is not an icon-path"
    Now: "Specified icon 'icon.png' does not exist" or "'icon' must be either a .png or a .svg"

- Improve error message if apparmor or seccomp profiles don't exist.

    Was: "The 'migration' property does not match the required schema: {'security-policy': {'apparmor': 'path/profile', 'seccomp': 'path/profile'}, 'type': 'migration-skill'} is not valid under any of the given schemas"
    Now: "Specified file 'path/profile' does not exist"

- Improve generic jsonschema errors to include the properties which had the error.

    Was: "None is not of type 'object'"
    Now: "The 'parts' property does not match the required schema: None is not of type 'object'"

### DIFF
--- a/snapcraft/tests/test_yaml.py
+++ b/snapcraft/tests/test_yaml.py
@@ -37,8 +37,8 @@ class TestYaml(tests.TestCase):
         dirs.setup_dirs()
 
         patcher = unittest.mock.patch('os.path.exists')
-        mock_wrap_exe = patcher.start()
-        mock_wrap_exe.return_value = True
+        self.mock_path_exists = patcher.start()
+        self.mock_path_exists.return_value = True
         self.addCleanup(patcher.stop)
 
     @unittest.mock.patch('snapcraft.yaml.Config.load_plugin')
@@ -262,7 +262,7 @@ parts:
             snapcraft.yaml.Config()
 
         self.assertEqual(raised.exception.message,
-                         '\'name\' is a required property')
+                         "'name' is a required property")
 
     @unittest.mock.patch('snapcraft.yaml.Config.load_plugin')
     def test_invalid_yaml_invalid_name_as_number(self, mock_loadPlugin):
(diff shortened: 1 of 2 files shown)


snapcraft/tests/test_yaml.py
```diff
@@ -283,7 +283,89 @@ def test_invalid_yaml_invalid_name_as_number(self, mock_loadPlugin):
             snapcraft.yaml.Config()
 
         self.assertEqual(raised.exception.message,
-                         '1 is not of type \'string\'')
+                         'The \'name\' property does not match the required '
+                         'schema: 1 is not of type \'string\'')
```
elopio: If you use double quotes then you won't have to escape the single quotes. Same for the rest of the tests.

kyrofa: Yeah, but jsonschema's error messages use single-quotes, so I had to bite the bullet and maintain consistency.

sergiusens:
Looks good, jsonschema could be better and make this easier instead of needing to go into hacks`^w` clever tricks for the formatter :wink:

Let's wait for @elopio and the tests!

snapcraft/tests/test_yaml.py
elopio: right. I mean the exception to the rule of using single quotes to avoid me going nuts is when the string contains single quotes. In that case, use double quotes. Hah, confusing. So, the style we are trying to follow in the code is:
"The 'name' property does not match the required "
"schema: 1 is not of type 'string'"

Of course not all the old code is following it, as your deleted lines show.

kyrofa:
@sergiusens heh, no kidding! I couldn't agree more.

snapcraft/tests/test_yaml.py
kyrofa: Haha! Good to know @elopio, I'll update this real quick.

snapcraft/yaml.py
```diff
@@ -38,14 +38,26 @@
...
+
+    return True
 
 
 @jsonschema.FormatChecker.cls_checks('icon-path')
 def _validate_icon(instance):
     normalized_icon = instance.lower()
-    ext = normalized_icon.endswith('.png') or normalized_icon.endswith('.svg')
-    return os.path.exists(instance) and ext
+    if not (normalized_icon.endswith('.png') or
```
elopio: You could use os.path.splitext once, and then just check for equality.
I'm not too convinced that it would be much clearer, so feel free to ignore.

elopio:
:+1: 
just an ignorable comment about using splitext.

snapcraft/yaml.py
sergiusens: if you do that, I'd prefer to use one of the fancy python things `https://docs.python.org/3.5/library/functions.html#any`

snapcraft/tests/test_yaml.py
kyrofa: Fixed.

snapcraft/yaml.py
kyrofa: Alright I revamped this a bit-- I think it's cleaner. Though @sergiusens I didn't see a good use for `any` here.

sergiusens:
:+1:

PR #298, opened 2016-02-04
Upload: Obtain name from YAML instead of snap.

kyrofa:
This PR fixes LP: [#1541353](https://bugs.launchpad.net/snapcraft/+bug/1541353) by using the `snapcraft.yaml` to obtain the snap name rather than using regex against the snap package, which was rather fragile and not in accordance with the spec.

### DIFF
--- a/snapcraft/commands/upload.py
+++ b/snapcraft/commands/upload.py
@@ -54,13 +54,13 @@ def main(argv=None):
 
     # make sure the full lifecycle is executed
     yaml_config = snapcraft.yaml.load_config()
-    snap_name = _format_snap_name(yaml_config.data)
+    snap_filename = _format_snap_name(yaml_config.data)
 
-    if not os.path.exists(snap_name):
+    if not os.path.exists(snap_filename):
         logger.info(
             'Snap {} not found. Running snap step to create it.'.format(
-                snap_name))
+                snap_filename))
         snap.main(argv=argv)
 
     config = load_config()
-    upload(snap_name, config=config)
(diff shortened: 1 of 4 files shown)


sergiusens:
What editor do you have that does not have a save hook to check the static bits :wink:
`snapcraft/tests/test_storeapi_upload.py:174:9: F841 local variable 'name' is assigned to but never used`

snapcraft/commands/upload.py
```diff
@@ -54,13 +54,13 @@ def main(argv=None):
 
     # make sure the full lifecycle is executed
     yaml_config = snapcraft.yaml.load_config()
-    snap_name = _format_snap_name(yaml_config.data)
+    snap_package = _format_snap_name(yaml_config.data)
```
sergiusens: Don't call it package; its a language we are trying to get rid of, just call it `snap`.

kyrofa:
@sergiusens argh, I'm still in 1.x mode, where the static tests were run with the units. Sorry!

snapcraft/commands/upload.py
kyrofa: Yeah I'd use snap but it clashes with the import, so I went with `snap_filename`.

sergiusens:
:+1:

We don't have upload tests, so does this still integrate well?

kyrofa:
Yeah there's a good test case YAML on the bug itself-- without this fix it fails, with it, it succeeds (and uploads).

sergiusens:
:100:

PR #302, opened 2016-02-05
integration-tests: add login and logout tests

elopio:


elopio:
I get this now. The encrypted env vars are not usable from any pull request. I was assuming that they would be usable from the PRs of members of the team, and that they just could never be printed. Of course, that's totally wrong. So the coveralls token must be a lie, and I spent like a whole week trying to make it work, sooo sad.
Let's keep an eye on travis-ci/travis-ci/issues/1946 to see if they ever enable this.
In the mean time, this login tests will work only for PRs made to snappy, or when running from master.
If you change something that has to do with the login, make sure that your branch is in the upstream repo.

integration_tests/test_login_logout.py
```diff
@@ -0,0 +1,83 @@
...
+        process.expect_exact(
+            'Enter your Ubuntu One SSO credentials.\r\n'
+            'Email: ')
+        process.sendline(email)
+        process.expect_exact('Password: ')
+        process.sendline(password)
+        process.expect_exact(
+            "One-time password (just press enter if you don't use two-factor "
+            "authentication): ")
+        process.sendline('')
```
kyrofa: Ah, this is nice.

kyrofa:
Yeah, I like it! Too bad on the variables, but it makes sense. We'll just need to keep it in mind, I suppose.

PR #303, opened 2016-02-05
Support setting an output file when snapping

sergiusens:
This adds support for -o or --output to have a deterministic
output file for a snap.

LP: #1542479


sergiusens:
This is a requirement to support cleanbuild so we will have to release after this

elopio:
lgtm. The unit test seems to be testing a lot of things not related to this branch, but it's ok to me, better safe than sorry.

elopio:
:+1:

PR #304, opened 2016-02-05
integration_tests: add a successful upload test

elopio:


elopio:
Requires #302.

integration_tests/test_upload.py
```diff
@@ -48,3 +51,25 @@ def test_upload(self):
...
+        self.addCleanup(self.logout)
+        self.login()
+
+        project_dir = self.copy_project_to_tmp(project_dir)
+        yaml_file = os.path.join(project_dir, 'snapcraft.yaml')
+        for line in fileinput.input(yaml_file, inplace=True):
+            if 'version: ' in line:
+                # Change to a random version.
+                # The maximum size is 32 chars.
+                print('version: ' + str(uuid.uuid4().int)[:32])
```
kyrofa: Nice. Do we have any concern about needing to blow these away every once in a while?

elopio: Not as far as I can see. It assigns a sequential version starting from 1, ignoring what we pass in this string. I'd say we have some tens of thousands of executions before worrying. And I hope that before we hit that many executions in staging we have a way to fake or deploy the server locally, so we will have to run fewer slow tests hitting staging.

kyrofa: Ah, I'm not worried about the randomness-- I'm worried about the store folks getting angry that we have a staging account with 30,000 snaps in it :stuck_out_tongue: .

elopio: no, don't worry about that. I have 4 years of test users from the suite I wrote when I worked in ubuntu one ^_^.

kyrofa: Hahaha!

PR #305, opened 2016-02-05
adt fixes for maven and noninteractive deb install

sergiusens:
This makes sure we setup noninteractive for deb installation and
also sets up the proxy configuration for maven.

LP: #1542511

### DIFF
--- a/examples_tests/tests.py
+++ b/examples_tests/tests.py
@@ -23,6 +23,7 @@ import subprocess
 import tempfile
 import time
 
+import fixtures
 import testscenarios
 import testtools
 from testtools import content
@@ -240,6 +241,8 @@ class TestSnapcraftExamples(testscenarios.WithScenarios, testtools.TestCase):
         snapcraft_bin = os.getenv('SNAPCRAFT', 'snapcraft')
         self.snapcraft_command = os.path.join(
             os.getcwd(), 'bin', snapcraft_bin)
+        self.useFixture(
+            fixtures.EnvironmentVariable('SNAPCRAFT_SETUP_PROXIES', '1'))
 
     def run_command_through_ssh(self, command):
         try:
--- a/integration_tests/test_maven_plugin.py
+++ b/integration_tests/test_maven_plugin.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright (C) 2015 Canonical Ltd
+# Copyright (C) 2015, 2016 Canonical Ltd
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License version 3 as
@@ -14,10 +14,17 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import fixtures
+
 import integration_tests
 
 
 class MavenPluginTestCase(integration_tests.TestCase):
+
+    def setUp(self):
+        super().setUp()
+        self.useFixture(
+            fixtures.EnvironmentVariable('SNAPCRAFT_SETUP_PROXIES', '1'))
 
     def test_build_maven_plugin(self):
         project_dir = 'simple-maven'
--- a/snapcraft/plugins/maven.py
+++ b/snapcraft/plugins/maven.py
@@ -33,6 +33,7 @@ Additionally, this plugin uses the following plugin-specific keywords:
 import glob
 import logging
 import os
+from urllib.parse import urlparse
 
 import snapcraft
 import snapcraft.common
@@ -40,6 +41,26 @@ import snapcraft.plugins.jdk
 
 
 logger = logging.getLogger(__name__)
+
+
+_MVN_SETTINGS_FORMAT = (
+    '<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"\n'
+    '          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"\n'
+    '          xsi:schemaLocation="http://maven.apache.org/SETTINGS/'
+    '1.0.0 http://maven.apache.org/xsd/settings-1.0.0.xsd">\n'
+    '  <interactiveMode>false</interactiveMode>\n'
+    '  <proxies>\n'
+    '    <proxy>\n'
+    '      <id>proxy</id>\n'
+    '      <active>true</active>\n'
+    '      <protocol>http</protocol>\n'
+    '      <host>{}</host>\n'
+    '      <port>{}</port>\n'
+    '      <nonProxyHosts>{}</nonProxyHosts>\n'
+    '    </proxy>\n'
+    '  </proxies>\n'
+    '</settings>\n'
+)
 
 
 class MavenPlugin(snapcraft.plugins.jdk.JdkPlugin):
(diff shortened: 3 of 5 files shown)


examples_tests/tests.py
```diff
@@ -32,6 +34,24 @@
...
+    '    <proxy>\n' + \
+    '      <id>proxy</id>\n' + \
+    '      <active>true</active>\n' + \
+    '      <protocol>http</protocol>\n' + \
+    '      <host>{}</host>\n' + \
+    '      <port>{}</port>\n' + \
+    '      <nonProxyHosts>.internal</nonProxyHosts>\n' + \
+    '    </proxy>\n' + \
+    '  </proxies>\n' + \
+    '</settings>\n'
```
elopio: If you put this string inside parentheses, you don't need the \ nor the +, which I find really useful.
There's also the option of using three quotes and textwrap.dedent, which I don't like that much but it's common.

examples_tests/tests.py
```diff
@@ -210,6 +230,21 @@ class TestSnapcraftExamples(testscenarios.WithScenarios, testtools.TestCase):
...
+    def _setup_mvn_proxy(cls):
+        settings_path = os.path.expanduser('~/.m2/settings.xml')
+        if os.path.exists(settings_path):
+            return
+
+        if 'http_proxy' in os.environ:
+            proxy = urlparse(os.environ['http_proxy'])
+            if not os.path.exists(os.path.dirname(settings_path)):
+                cls.addCleanup(os.rmdir, os.path.dirname(settings_path))
+            os.makedirs(os.path.dirname(settings_path), exist_ok=True)
```
elopio: You are calling dir name three times . I would put it in a var.

integration_tests/test_maven_plugin.py
```diff
@@ -14,11 +14,51 @@
...
+            return
+
+        if 'http_proxy' in os.environ:
+            proxy = urlparse(os.environ['http_proxy'])
+            if not os.path.exists(os.path.dirname(settings_path)):
+                self.addCleanup(os.rmdir, os.path.dirname(settings_path))
+            os.makedirs(os.path.dirname(settings_path), exist_ok=True)
+            with open(settings_path, 'w') as f:
+                f.write(_MVN_SETTINGS.format(proxy.hostname, proxy.port))
+            self.addCleanup(os.remove, settings_path)
```
elopio: We can start moving common helpers to snapcraft/tests/utils, or you can turn this into a fixture and put it in snapcraft/tests/fixture_setup.
I think we shouldn't duplicate things, even if they are in different suites. That makes the suite rely on the snapcraft library, not just the snapcraft binary, but that's ok for me. What do you think?

elopio:
Thanks a lot for working on this, because I made no useful progress...

kyrofa:
Looks good to me! :+1:

examples_tests/tests.py
```diff
@@ -16,6 +16,7 @@
 
 import logging
 import os
+import os.path
```
elopio: There's no need to import os.path. By importing os you can just use it.

snapcraft/plugins/maven.py
```diff
@@ -42,6 +43,26 @@
 logger = logging.getLogger(__name__)
 
 
+_MVN_SETTINGS = (
```
elopio: A nit, I find it useful to suffix the format constants so it's easier to know that they need an argument.
Like _MVN_SETTINGS_FORMAT.

snapcraft/tests/test_plugin_maven.py
```diff
@@ -0,0 +1,131 @@
...
+            maven_options['uniqueItems'],
+            'Expected "maven-options" "uniqueItems" to be "True"')
+
+    @mock.patch.object(maven.MavenPlugin, 'run')
+    @mock.patch('glob.glob')
+    def test_build(self, glob_mock, run_mock):
+        plugin = maven.MavenPlugin('test-part', self.options)
+        os.makedirs(plugin.sourcedir)
+        glob_mock.return_value = [
+            os.path.join(plugin.builddir, 'target', 'fake')]
```
elopio: You are not using this value in the test, it seems it's just needed to be able to test the the call to run. To make that clear I usually call the value 'dummy' instead of 'fake'.

snapcraft/tests/test_plugin_maven.py
```diff
@@ -0,0 +1,131 @@
...
+
+        os.environ['SNAPCRAFT_SETUP_PROXIES'] = '1'
+        os.environ['http_proxy'] = 'http://localhost:3132'
+
+        plugin = maven.MavenPlugin('test-part', self.options)
+
+        settings_path = os.path.join(plugin.partdir, 'm2', 'settings.xml')
+        os.makedirs(plugin.sourcedir)
+        glob_mock.return_value = [
+            os.path.join(plugin.builddir, 'target', 'fake')]
```
elopio: Same s/fake/dummy.

snapcraft/plugins/maven.py
```diff
@@ -79,3 +111,19 @@ def build(self):
...
+    with open(settings_path, 'w') as f:
+        f.write(_MVN_SETTINGS.format(
+            proxy.hostname,
+            proxy.port,
+            _get_no_proxy_string()))
+
+
+def _get_no_proxy_string():
+    no_proxy = [k.strip() for k in
+                os.environ.get('no_proxy', 'localhost').split(',')]
```
elopio: You are missing tests for when no_proxy has one value, and when it has multiple values.

elopio:
Thanks for adding this. I'm :+1: with a couple of ignorable comments and a note about a missing test.
Also repo is severely undertested. I'll move it to the top of my pending coverage tasks.

snapcraft/tests/test_plugin_maven.py
```diff
@@ -0,0 +1,131 @@
...
+
+    @mock.patch.object(maven.MavenPlugin, 'run')
+    @mock.patch('glob.glob')
+    def test_build_with_snapcraft_proxy(self, glob_mock, run_mock):
+        env_copy = os.environ.copy()
+        self.addCleanup(os.environ.clear)
+        self.addCleanup(os.environ.update, env_copy)
+
+        os.environ['SNAPCRAFT_SETUP_PROXIES'] = '1'
+        os.environ['http_proxy'] = 'http://localhost:3132'
```
elopio: Oh, one more. This will pollute the environ for the following tests.
Use self.useFixture(fixtures.EnvironmentVariable(var, value))

snapcraft/plugins/maven.py
sergiusens: fixed

snapcraft/tests/test_plugin_maven.py
sergiusens: fixed

snapcraft/tests/test_plugin_maven.py
sergiusens: fixed

snapcraft/plugins/maven.py
sergiusens: fixed

examples_tests/tests.py
sergiusens: fixed

snapcraft/tests/test_plugin_maven.py
sergiusens: I'm using `addCleanup`, I don't mind trying this though

elopio:
Tests passed and we have +0.6%. Merging...
You earned a QA :star:

PR #306, opened 2016-02-08
Use image link which actually works

dholbach:
(raw.githubusercontent.com shows SVG as raw XML, rawgit.com works as expected).

LP: #1542422


kyrofa:
Well that's not annoying. Thanks @dholbach :smiley: .

PR #309, opened 2016-02-09
travis: remove the coveralls token

elopio:
LP: #1543675

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,8 @@ virtualenv:
   system_site_packages: true
 env:
   global:
-  # Encrypted Coveralls token. The secret token for your repository can be found at the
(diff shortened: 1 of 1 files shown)


elopio:
Let's wait for the coveralls results to confirm that I am correct.

kyrofa:
Why on earth isn't this required? Does it mean anyone can upload coverage for snapcraft?

elopio:
Anyone can, but it will be reported to their branches. Coveralls separates the master coverage, and no PRs affect it.
I suppose that the trigger from travis hits an API that doesn't need a token, that's somehow not accessible to everybody. The details about this I don't understand.

kyrofa:
Huh. Alright, I'm :+1: .

PR #312, opened 2016-02-10
Pre release unit test fix for 2.2

sergiusens:
This considers builder environment variables that could potentially affect
unit tests.


sergiusens:
Tested before (with failure) and after the fix with:
`no_proxy='127.0.0.1' python3 -m unittest -v -b snapcraft.tests.test_plugin_maven.MavenPluginTestCase.test_build_with_snapcraft_proxy`

elopio:
Confirmed here that the test fails without this change, and passes with it.

PR #314, opened 2016-02-11
Replace dpkg-architecure calls in tests

sergiusens:
The replacement is interesting since it allows to capture dependency
issues in the snapcraft.yaml's.

LP: #1544340

### DIFF
--- a/integration_tests/test_snap.py
+++ b/integration_tests/test_snap.py
@@ -27,18 +27,10 @@ from testtools.matchers import (
 
 import integration_tests
 
-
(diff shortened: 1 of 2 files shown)


elopio:
:+1: 
I'm just wondering why you imported the symbols instead of the style we usually follow:

from snapcraft import common.

sergiusens:
@elopio because I few months ago I asked @chipaca and he told me I should not be so java-ish :wink:

kyrofa:
:+1:

PR #316, opened 2016-02-11
Adding python-minimal to plugins that need it

sergiusens:
python-minimal is needed by the python2 and the catkin plugin.

LP: #1541451

### DIFF
--- a/snapcraft/common.py
+++ b/snapcraft/common.py
@@ -16,6 +16,7 @@
 
 # Data/methods shared between plugins and snapcraft
 
+import glob
 import os
 import platform
 import subprocess
@@ -131,6 +132,17 @@ def set_schemadir(schemadir):
 
 def get_schemadir():
     return _schemadir
+
+
+def get_python2_path(root):
+    """Return a valid PYTHONPATH or raise an exception."""
+    python_paths = glob.glob(os.path.join(
+        root, 'usr', 'lib', 'python2*', 'dist-packages'))
+    try:
+        return python_paths[0]
+    except IndexError:
+        raise EnvironmentError(
+            'PYTHONPATH cannot be set for {!r}'.format(root))
 
 
 def isurl(url):
--- a/snapcraft/plugins/catkin.py
+++ b/snapcraft/plugins/catkin.py
@@ -110,13 +110,7 @@ deb http://${security}.ubuntu.com/${suffix} trusty-security main universe
     def env(self, root):
         """Runtime environment for ROS binaries and services."""
 
-        return [
-            # The ROS packaging system tools (e.g. rospkg, etc.) don't go
-            # into the ROS install path (/opt/ros/$distro), so we need the
-            # PYTHONPATH to include the dist-packages in /usr/lib as well.
-            'PYTHONPATH={0}'.format(os.path.join(root, 'usr', 'lib',
-                                    self.python_version, 'dist-packages')),
-
+        env = [
             # This environment variable tells ROS nodes where to find ROS
             # master. It does not affect ROS master, however-- this is just the
             # default URI.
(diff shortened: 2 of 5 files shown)


elopio:
lgtm. The tests failed because of a hash mismatch. Retriggered...

snapcraft/plugins/python2.py
```diff
@@ -67,6 +67,7 @@ def __init__(self, name, options):
         super().__init__(name, options)
         self.stage_packages.extend([
             'python-dev',
+            'python-minimal',
```
kyrofa: I'm afraid the fix for this one isn't so easy (fix doesn't work for me on xenial Classic Dimension). Perhaps the `env()` is used before `pluginhandler.pull()`?

sergiusens: It indeed might be, I pushed it here to see what happened on travis; I'm not sure how to fix this one since `run` is called from the class which generates an env by calling env.

kyrofa:
Yeah that looks better. :+1:

sergiusens:
@kyrofa did it work for you? Or is it just looks :smiley:

kyrofa:
@sergiusens testing it out now, but yeah, based on how it looks it seems to solve the issue I had.

snapcraft/plugins/python2.py
```diff
@@ -137,23 +145,11 @@ def build(self):
         os.makedirs(self.dist_packages_dir, exist_ok=True)
```
kyrofa: You're still calling this, but you removed it.

sergiusens: Weird that it would need to be created; no I see our unit test coverage sucks for these python plugins :sob:

kyrofa: Indeed :disappointed:

kyrofa:
@sergiusens alright, tested and verified: :+1: from me.

snapcraft/common.py
```diff
@@ -133,6 +134,17 @@ def get_schemadir():
     return _schemadir
 
 
+def get_python2_path(root):
+    """Returns a valid PYTHONPATH or raises an exception."""
```
elopio: s/Returns/Return

snapcraft/plugins/catkin.py
```diff
@@ -140,6 +134,16 @@ def env(self, root):
                     root, 'opt', 'ros', self.options.rosdistro, 'setup.sh'))
         ]
 
+        # The ROS packaging system tools (e.g. rospkg, etc.) don't go
+        # into the ROS install path (/opt/ros/$distro), so we need the
+        # PYTHONPATH to include the dist-packages in /usr/lib as well.
+        try:
+            env.append('PYTHONPATH={0}'.format(common.get_python2_path(root)))
+        except EnvironmentError as e:
+            logger.debug(e)
```
elopio: So no python is not a fatal error? Sounds weird.
But if that's the was, I think it should be warn, not debug.

kyrofa: It's because this WILL happen when run is called, until the stage-packages have been installed.

snapcraft/plugins/python2.py
```diff
@@ -169,3 +165,10 @@ def snap_fileset(self):
         fileset.append('-usr/lib/python*/*/*/*/*/*/*/*/*/__pycache__/*.pyc')
         fileset.append('-usr/lib/python*/*/*/*/*/*/*/*/*/*/__pycache__/*.pyc')
         return fileset
+
+
+def _get_python2_include(root):
+    try:
+        return glob.glob(os.path.join(root, 'usr', 'include', 'python2*'))[0]
+    except IndexError:
+        raise EnvironmentError('python development headers not installed')
```
elopio: What about a clue about how to fix this? Like "Install the package ..."

elopio:
Some minor comments. :+1: from me.

snapcraft/common.py
sergiusens: and "or *raise* an" I suppose

snapcraft/plugins/catkin.py
sergiusens: @elopio I debated the case of using warn; but as @kyrofa mentions, it **will** happen a couple of times before it doesn't

snapcraft/plugins/python2.py
sergiusens: This is a programmatic error though as python-dev should be a required package for the python plugin (as it is now).

snapcraft/plugins/catkin.py
elopio: Got it. That's worth commenting in the code.

PR #318, opened 2016-02-11
Generate reference from snapcraft's "help" command.

dholbach:
Run:
```
  ./snapcraft/scripts/generate_reference.py
```
or
```
  ./snapcraft/scripts/generate_reference.py --output-file=/tmp/something.md
```
LP: #1544540

### DIFF
--- a/.bzrignore
+++ b/.bzrignore
@@ -1,3 +1,4 @@
+docs/reference.md
 examples/*/parts/
 examples/*/snap/
 examples/*/stage/
--- a/snapcraft/scripts/generate_reference.py
+++ b/snapcraft/scripts/generate_reference.py
@@ -1,0 +1,118 @@
+#!/usr/bin/env python3
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2016 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import argparse
+import os
+import subprocess
+import sys
+
+DEFAULT_REFERENCE_FILE = 'docs/reference.md'
+
+
+def _get_help_output(topic, lines=False):
+    TRANSLATE = {
+        'tar-content': 'tar_content',
+    }
+    if topic in TRANSLATE:
+        topic = TRANSLATE[topic]
+    return _run_snapcraft(['help', topic], lines)
+
+
+def _run_snapcraft(cmd, lines=False):
+    cmd = ['./snapcraft'] + cmd
+    output = subprocess.check_output(
+        cmd, cwd=os.path.join(os.path.dirname(__file__), '../../bin'))
+    output = output.decode('utf-8')
+    if lines:
+        return output.split('\n')[:-1]  # drop last empty line
+    return output
+
+
+class Reference:
+    def __init__(self, output_file):
+        self.docs = []
+        self.topics = []
+        self.plugins = []
+        self.output_file = output_file
+        assert os.path.isdir(os.path.dirname(self.output_file)), \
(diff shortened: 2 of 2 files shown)


dholbach:
This is just a first cut to get an idea of you generally like the idea. Having submitted this I wonder if I shouldn't generate just one docs/reference.md file with an auto-generated ToC instead...

dholbach:
Ah... maybe I should explain the rationale for this: it'd be great if (once the md importer for developer.u.c finally lands) we could show an automatically updated reference without having to duplicate this information.

dholbach:
I changed it to only write to one file.

Let me know if you want me to break out the functionality into a separate file outside `setup.py`.

setup.py
```diff
@@ -31,6 +33,99 @@
...
+        for topic in self.topics:
+            text = _get_help_output(topic)
+            if not text.startswith('The plugin has no documentation'):
+                self.docs += [
+                    (topic, text)
+                ]
+        self._write_index_doc()
+
+    def _write_index_doc(self):
+        text = '''
```
sergiusens: Use three """ instead, also, maybe call this text_template or something... @elopio will have ideas wrt to consistency on this one.

setup.py
```diff
@@ -31,6 +33,99 @@
         version = match.group(1)
 
 
+def _get_help_output(topic, lines=False):
+    TRANSLATE = {
+        'tar-content': 'tar_content',
+    }
```
elopio: Does this apply to all the topics with a dash? If so, you could just replace the dash.

setup.py
```diff
@@ -31,6 +33,99 @@
...
+    if topic in TRANSLATE:
+        topic = TRANSLATE[topic]
+    return _run_snapcraft(['help', topic], lines)
+
+
+def _run_snapcraft(cmd, lines=False):
+    cmd = ['./bin/snapcraft']+cmd
```
elopio: pep8 doesn't check this file, which is probably a bug. You should use spaces between the operators:
cmd = ['./bin/snapcraft'] + cmd

setup.py
```diff
@@ -31,6 +33,99 @@
...
+    command = subprocess.Popen(cmd, stdout=subprocess.PIPE)
+    output = command.communicate()[0]
```
elopio: why don't you use subprocess.check_output here?

setup.py
```diff
@@ -31,6 +33,99 @@
...
+
+    if sys.version_info.major == 3:
+        output = output.decode('utf-8')  # bytes to str
```
elopio: I think I would prefer just to ignore python2. @sergiusens do you know if there's a good reason to keep this check?

setup.py
```diff
@@ -31,6 +33,99 @@
...
+    def initialize_options(self):
+        self.output_file = None
+
+    def finalize_options(self):
+        if not self.output_file:
+            self.output_file = os.path.abspath(self.DEFAULT_REFERENCE_FILE)
+
+    def run(self):
+        assert os.getcwd() == os.path.abspath(os.path.dirname(__file__)), \
+            'Must be run in snapcraft root.'
```
elopio: This might be confusing. Root is also the user. So maybe:
'Must be run in the root directory of snapcraft.'

elopio:
I like the idea. I don't like that it's in the setup.py.
Could we move it to something like snapcraft/scripts/generate_reference.py ?

setup.py
dholbach: Right now it's the only one and I thought I'd rather make this explicit than just changing anything which comes up with a dash.

setup.py
dholbach: We might run this on developer.u.c where I don't know which python versions are available.

snappy-m-o:

snappy-m-o:

snappy-m-o:

dholbach:
I hope it's all fine now.

elopio:
Looks good. Your replies make a lot of sense, and thanks for moving it.
Now that you are inside the snapcraft, the static tests will check the script. (it was a trap! :)
Travis is giving this error:
snapcraft/scripts/generate_reference.py:38:9: E128 continuation line under-indented for visual indent

dholbach:
It's fixed now.

setup.py
dholbach: I removed the python2-ism. We have python3 available.

elopio:
add to whitelist

dholbach:
What does "add to whitelist" mean? Do I need to do anything?

kyrofa:
@dholbach I *think* he's experimenting with the bot.

elopio:
retest this please

snappy-m-o:

elopio:
landing...

PR #320, opened 2016-02-12
Set after test cmake project to use no language

sergiusens:
The after integration test had a cmake project set to use
the generic language template, as this cmake project does
nothing we explicitly set it to NONE.

LP: #1544792


kyrofa:
:+1: just waiting on the tools.

PR #324, opened 2016-02-15
catkin plugin: do not overwrite the PYTHONPATH

elopio:
LP: #1544790


elopio:
I think it's wrong to just overwrite the PYTHONPATH, and this makes the ros test pass. But this might have other consequences I'm not fully understanding.

kyrofa:
@elopio excellent fix, thank you! Good detective work on the bisect. This fix works by keeping the PYTHONPATH that was set by the earlier call to `setup.sh` (which included the path in which catkin was shipped, `opt/ros/<version>/lib/python2.7/dist-packages`) rather than wiping it out, which resulted in the bug.

PR #325, opened 2016-02-17
examples tests: fix typo and update libpipeline test

elopio:
LP: #1546771


elopio:
wait wait wait, scratch this. I want to do this in a different way. Each test will be more verbose, but clearer, which is something I've been missing when debugging them.

PR #326, opened 2016-02-18
examples_tests: fix the libpipeline test.

elopio:
Refactor the examples tests to be able to write them independently, each
on its own file.

LP: #1546771

### DIFF
--- a/examples_tests/__init__.py
+++ b/examples_tests/__init__.py
@@ -1,0 +1,167 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2015, 2016 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import atexit
+import logging
+import os
+import re
+import tempfile
+import shutil
+import subprocess
+
+import fixtures
+import testtools
+from testtools import content
+from testtools.matchers import MatchesRegex
+
+from examples_tests import testbed
+
+logger = logging.getLogger(__name__)
+
+config = {}
+
+
+def _get_latest_ssh_private_key():
+    """Return the latest private key in ~/.ssh.
+
+    :returns:
+        Path of the most-recently-modified private SSH key
+    :raises LookupError:
+        If no such key was found.
+
+    This function tries to mimic the logic found in ``ubuntu-device-flash``.
+    It will look for the most recently modified private key in the users' SSH
+    configuration directory.
+    """
+    candidates = []
+    ssh_dir = os.path.expanduser('~/.ssh/')
+    for filename in os.listdir(ssh_dir):
+        # Skip public keys, we want the private key
+        if filename.endswith('.pub'):
+            continue
+        ssh_key = os.path.join(ssh_dir, filename)
+        # Skip non-files
+        if not os.path.isfile(ssh_key):
+            continue
+        # Ensure that it is a real ssh key
+        with open(ssh_key, 'rb') as stream:
+            if stream.readline() != b'-----BEGIN RSA PRIVATE KEY-----\n':
+                continue
+        candidates.append(ssh_key)
+    # Sort the keys by modification time, pick the most recent key
+    candidates.sort(key=lambda f: os.stat(f).st_mtime, reverse=True)
+    logger.debug('Available ssh public keys: %r', candidates)
+    if not candidates:
+        raise LookupError('Unable to find any private ssh key')
+    return candidates[0]
+
+
+class ExampleTestCase(testtools.TestCase):
+
+    example_dir = None
+
+    def setUp(self):
+        filter_ = config.get('filter', None)
+        if filter_:
+            if not re.match(filter_, self.example_dir):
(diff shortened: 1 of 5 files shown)


elopio:
I ended up doing a massive refactor because I hated writing the installed tests. I'm sorry about the size of this one.

examples_tests/__init__.py
```diff
@@ -0,0 +1,161 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2015, 2016 Canonical Ltd
```
fgimenez: If this is a new file it should only include 2016

examples_tests/__init__.py
```diff
@@ -0,0 +1,161 @@
...
+        private_key = _get_latest_ssh_private_key()
+        snappy_image = config.get('snappy_image', None)
+        if not snappy_image:
+            temp_dir = tempfile.mkdtemp()
+            snappy_image = testbed.create_snappy_image(temp_dir)
+            # Store the image path in the config so it's only create once
+            # per execution.
+            config['snappy_image'] = snappy_image
+            # Delete the image when the execution exits.
+            atexit.register(shutil.rmtree, temp_dir)
```
fgimenez: Why not using addCleanup for this?

examples_tests/test_libpipeline.py
```diff
@@ -0,0 +1,34 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2015, 2016 Canonical Ltd
```
fgimenez: If this is a new file it should only include 2016

examples_tests/testbed.py
```diff
@@ -0,0 +1,112 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2015, 2016 Canonical Ltd
```
fgimenez: If this is a new file it should only include 2016

examples_tests/test_libpipeline.py
```diff
@@ -0,0 +1,34 @@
...
+        self.build_snap(self.example_dir)
+        self.install_snap(self.example_dir, 'pipelinetest', '1.0')
+        output = self.run_command_in_snappy_testbed(
+            '/snaps/bin/pipelinetest.pipelinetest')
+        expected = (
+            'running ls | grep c\n'
+            'custom libpipeline called\n'
+            'command-pipelinetest.wrapper\n'
+            'include\n')
+        self.assertEqual(output, expected)
```
fgimenez: Very nice! :)

examples_tests/__init__.py
```diff
@@ -0,0 +1,161 @@
...
+            raise
+
+    def install_snap(self, example_dir, snap_name, version):
+        if not config.get('skip-install', False):
+            snap_file_name = '{}_{}_amd64.snap'.format(
+                snap_name, version)
+            snap_local_path = os.path.join(
+                'examples', example_dir, snap_file_name)
+            self.snappy_testbed.copy_file(snap_local_path, '/home/ubuntu')
+            self.snappy_testbed.run_command([
```
fgimenez: Maybe you could check the output from run_command here, or catch any possible exception.

examples_tests/testbed.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def create_snappy_image(output_directory):
+    logger.info('Creating a snappy image to run the tests.')
+
+    image_path = os.path.join(output_directory, 'snappy.img')
+    subprocess.check_call(
+        ['sudo', 'ubuntu-device-flash', '--verbose',
+         'core', '15.04', '--channel', 'stable',
+         '--output', image_path, '--developer-mode'])
```
fgimenez: You could check the return code here, or possible exceptions raised. The textual output from udf could also be helpful, maybe at a high verbosity log level.

examples_tests/testbed.py
```diff
@@ -0,0 +1,112 @@
...
+    return image_path
+
+
+class SshTestbed:
```
fgimenez: Have you considered using a library like paramiko? http://docs.paramiko.org/en/1.16/api/client.html

fgimenez:
Looks very good, only real concerns about checking the output and eventual exceptions from command calls.

examples_tests/__init__.py
elopio: The file is new, but I started writing this code last year. I'm just refactoring it now. So as I understand it, it should have both years.

examples_tests/__init__.py
elopio: The idea is to create the image only once per execution, because it's what takes more time.
I tried using setupModule, but with this refactor it will mean once per example, which is also a lot of time. So the remaining options were to do it on __main__.py, or do a lazy creation and delete when the main thread exits. I liked a little more this last one, but I'm ok to change it if you have a better option in mind.

examples_tests/test_libpipeline.py
elopio: ditto.

examples_tests/testbed.py
elopio: ditto.

examples_tests/testbed.py
elopio: That could be nice. And pexpect + paramiko might be a good combination. After finishing my tasks for this week, I'll take a look at rewriting this with the lib.

examples_tests/testbed.py
elopio: check_call will raise an exception if the return code is different than 0. That will stop the test and show the traceback, which I think it's enough.

examples_tests/__init__.py
elopio: I pushed a check for the output. Any possible exception will be raised and the test will fail, which I think is enough.

fgimenez:
:+1:

elopio:
retest this please

sergiusens:
So the plan is to slowly move away from the big fixture of scenarios? I actually like this :+1:

PR #328, opened 2016-02-19
New command cleanbuild (using lxd)

sergiusens:
This new command's intention is to make building with a clean
slate and easy chore.

LP: #1480144

### DIFF
--- a/snapcraft/commands/cleanbuild.py
+++ b/snapcraft/commands/cleanbuild.py
@@ -1,0 +1,87 @@
+#!/usr/bin/python3
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2015, 2016 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+snapcraft cleanbuild
+
+Creates a lxd container to build the snap.
+
+This is a way to guarantee that the snapcraft.yaml used with associated
+local sources is not using any dependencies local to the developer system.
+
+The cleanbuild command requires a properly setup lxd environment that
+can connect to external networks. Refer to the "Ubuntu Desktop and
+Ubuntu Server" section on
+https://linuxcontainers.org/lxd/getting-started-cli
+to get started.
+
+Usage:
+  cleanbuild [options]
+
+Options:
+  -h --help             show this help message and exit.
+"""
+
+import logging
+import os.path
+import tarfile
+
+from docopt import docopt
+
+from snapcraft import repo
+from snapcraft.lxd import Cleanbuilder
+from snapcraft.common import format_snap_name
+
+from snapcraft.yaml import load_config
+
+logger = logging.getLogger(__name__)
+
+
+def _create_tar_filter(tar_filename):
+    def _tar_filter(tarinfo):
+        fn = tarinfo.name
+        if fn.startswith('./parts/') and not fn.startswith('./parts/plugins'):
+            return None
+        elif fn in ('./stage', './snap', tar_filename):
+            return None
+        elif fn.endswith('.snap'):
+            return None
+        return tarinfo
+    return _tar_filter
+
+
+def main(argv=None):
+    argv = [] if argv is None else argv
+    docopt(__doc__, argv=argv)
+
+    if not repo.is_package_installed('lxd'):
+        raise EnvironmentError(
+            'The lxd package is not installed, in order to use `cleanbuild` '
+            'you must install lxd onto your system. Refer to the '
+            '"Ubuntu Desktop and Ubuntu Server" section on '
+            'https://linuxcontainers.org/lxd/getting-started-cli/'
+            '#ubuntu-desktop-and-ubuntu-server to enable a proper setup.')
+
+    config = load_config()
+    tar_filename = '{}_{}_source.tar.bz2'.format(
+        config.data['name'], config.data['version'])
+
+    with tarfile.open(tar_filename, 'w:bz2') as t:
+        t.add(os.path.curdir, filter=_create_tar_filter(tar_filename))
+
+    snap_filename = format_snap_name(config.data)
+    Cleanbuilder(snap_filename, tar_filename).execute()
--- a/snapcraft/common.py
+++ b/snapcraft/common.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright (C) 2015 Canonical Ltd
+# Copyright (C) 2015, 2016 Canonical Ltd
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License version 3 as
@@ -86,22 +86,34 @@ _DEB_TRANSLATIONS = {
 }
 
 
+class PlatformError(Exception):
+
+    def __init__(self):
+        super().__init__(
+            '{0} is not supported, please log a bug at '
+            'https://bugs.launchpad.net/snapcraft/+filebug?'
+            'field.title=please+add+support+for+{0}'.format(
+                platform.machine()))
+
+
 def get_arch():
     try:
         return _DEB_TRANSLATIONS[platform.machine()]['arch']
     except KeyError:
-        raise EnvironmentError(
-            '{} is not supported, please log a bug at'
-            'https://bugs.launchpad.net/snapcraft'.format(platform.machine()))
+        raise PlatformError()
 
 
 def get_arch_triplet():
     try:
         return _DEB_TRANSLATIONS[platform.machine()]['triplet']
     except KeyError:
-        raise EnvironmentError(
-            '{} is not supported, please log a bug at'
-            'https://bugs.launchpad.net/snapcraft'.format(platform.machine()))
+        raise PlatformError()
+
+
+def format_snap_name(snap):
+    snap['arch'] = (snap['architectures'][0]
+                    if len(snap['architectures']) == 1 else 'multi')
+    return '{name}_{version}_{arch}.snap'.format(**snap)
 
 
 def get_partsdir():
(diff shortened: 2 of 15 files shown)


sergiusens:
Hello there, this is `cleanbuild`, it used to use `pylxd` but not anymore as the API changed, does not work with python3 with that change and does not have tests anymore.

snapcraft/commands/cleanbuild.py
```diff
@@ -0,0 +1,85 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+snapcraft cleanbuild
+
+Uses a container using lxd to create the resulting snap.
```
elopio: Instead of "Uses a container using lxd", just "Uses a lxd container".

elopio: Maybe also instead of "to create the resulting snap", just "to build the snap".

snapcraft/commands/cleanbuild.py
```diff
@@ -0,0 +1,85 @@
...
+
+This is a way to guarantee that the snapcraft.yaml used with associated
+local sources is not using any dependencies local to the developer system.
+
+The cleanbuild command requires a properly setup lxd environment that
+can connect to external networks.
```
elopio: Maybe a link to how to set up a lxd environment?

snapcraft/commands/cleanbuild.py
```diff
@@ -0,0 +1,85 @@
...
+
+import logging
+import os.path
+import tarfile
+
+from docopt import docopt
+
+from snapcraft import repo
+from snapcraft.lxd import Cleanbuilder
+from snapcraft.common import format_snap_name
```
elopio: hum, I dislike the import of symbols instead of modules.
But if you prefer it this way, as long as we are consistent I can live with it. Should we start using this style?

snapcraft/commands/cleanbuild.py
```diff
@@ -0,0 +1,85 @@
...
+
+from snapcraft.yaml import load_config
+
+logger = logging.getLogger(__name__)
+_DEFAULT_IMAGE_SERVER = 'https://images.linuxcontainers.org:8443'
+
+
+def _create_tar_filter(tar_filename):
+    def _tar_filter(tarinfo):
+        fn = tarinfo.name
```
elopio: why not file_name instead of fn? oh no, you are a go programmer again! :D

snapcraft/commands/cleanbuild.py
```diff
@@ -0,0 +1,85 @@
...
+from docopt import docopt
+
+from snapcraft import repo
+from snapcraft.lxd import Cleanbuilder
+from snapcraft.common import format_snap_name
+
+from snapcraft.yaml import load_config
+
+logger = logging.getLogger(__name__)
+_DEFAULT_IMAGE_SERVER = 'https://images.linuxcontainers.org:8443'
```
elopio: This constant is not used here.

snapcraft/lxd.py
```diff
@@ -0,0 +1,106 @@
...
+    def _create_container(self):
+        try:
+            remote_tmp = petname.Generate(2, '-')
+            check_call(['lxc', 'remote', 'add', remote_tmp, self._server])
+            check_call(['lxc', 'launch',
+                        '{}:ubuntu/xenial/{}'.format(remote_tmp, get_arch()),
+                        self._container_name])
+            yield
+        finally:
+            check_call(['lxc', 'stop', self._container_name])
```
elopio: No delete after the build is done?

snapcraft/lxd.py
```diff
@@ -0,0 +1,106 @@
...
+            self._container_run(['apt-get', 'update'])
+            self._container_run(['apt-get', 'install', 'snapcraft', '-y'])
+            self._container_run(
+                ['snapcraft', 'snap', '--output', self._snap_output])
+            self._pull_snap()
+
+    def _setup_project(self):
+        logger.info('Setting up container with project assets')
+        dst = os.path.join('/root', os.path.basename(self._project))
+        self._push_file(self._project, dst)
```
elopio: I thought it would be nice (and easier) to mount the directory in the container. Then you could inspect the part dirs after the container was killed, and no need to copy things around.

snapcraft/lxd.py
```diff
@@ -0,0 +1,106 @@
...
+        finally:
+            check_call(['lxc', 'stop', self._container_name])
+            check_call(['lxc', 'remote', 'remove', remote_tmp])
+
+    def execute(self):
+        with self._create_container():
+            self._setup_project()
+            self._wait_for_network()
+            self._container_run(['apt-get', 'update'])
+            self._container_run(['apt-get', 'install', 'snapcraft', '-y'])
```
elopio: Oh, I was kind of hoping that we could use the same snapcraft from the host. If we get it from the archive then it's not really useful for testing snapcraft itself.

snapcraft/tests/test_commands_cleanbuild.py
```diff
@@ -0,0 +1,111 @@
...
+    part1:
+      plugin: nil
+"""
+
+    def make_snapcraft_yaml(self, n=1):
+        super().make_snapcraft_yaml(self.yaml_template)
+        self.state_file = os.path.join(common.get_partsdir(), 'part1', 'state')
+
+    @mock.patch('snapcraft.lxd.check_call')
+    def test_cleanbuild(self, mock_call):
```
elopio: You should fake here that lxd is installed, I think.

snapcraft/common.py
```diff
@@ -91,7 +91,7 @@ def get_arch():
         return _DEB_TRANSLATIONS[platform.machine()]['arch']
     except KeyError:
         raise EnvironmentError(
-            '{} is not supported, please log a bug at'
+            '{} is not supported, please log a bug at '
```
elopio: You didn't forget, great! :)

elopio:
I like this very much. I left many comments just because I'm cranky and want to start my vacations :)
I would like that invasive integration test. Maybe we can just skip it if lxd is not installed.

snapcraft/commands/cleanbuild.py
sergiusens: done

snapcraft/commands/cleanbuild.py
sergiusens: done

snapcraft/commands/cleanbuild.py
sergiusens: We already have this all over; it really depends on the length of what you import.

snapcraft/commands/cleanbuild.py
sergiusens: hah, it's because the resulting lines below would have the horrible python semantics due to line length restrictions :wink:

snapcraft/lxd.py
sergiusens: It's ephemeral so no need.

snapcraft/lxd.py
sergiusens: Mounting requires root though. I prefer this non requiring root scenario as an initial implementation.

snapcraft/lxd.py
sergiusens: The problem with that is that you won't be able to drive builds from a different snapcraft version on a different series.

snapcraft/tests/test_commands_cleanbuild.py
sergiusens: I don't understand what this means

snapcraft/commands/cleanbuild.py
```diff
@@ -0,0 +1,87 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def _create_tar_filter(tar_filename):
+    def _tar_filter(tarinfo):
+        fn = tarinfo.name
+        if fn.startswith('./parts/') and not fn.startswith('./parts/plugins'):
+            return None
+        elif any(fn == d for d in ('./stage', './snap', tar_filename)):
```
chipaca: I find `fn in ('./stage', './snap', tar_filename)` more readable than this

snapcraft/commands/cleanbuild.py
```diff
@@ -0,0 +1,87 @@
...
+            return None
+        elif fn.endswith('.snap'):
+            return None
+        return tarinfo
+    return _tar_filter
+
+
+def main(argv=None):
+    argv = argv if argv else []
```
chipaca: `argv = [] if argv is None else argv`?

snapcraft/common.py
```diff
@@ -100,10 +100,16 @@ def get_arch_triplet():
         return _DEB_TRANSLATIONS[platform.machine()]['triplet']
     except KeyError:
         raise EnvironmentError(
-            '{} is not supported, please log a bug at'
+            '{} is not supported, please log a bug at '
             'https://bugs.launchpad.net/snapcraft'.format(platform.machine()))
```
chipaca: ooh :-) can I suggest you change the link to go to e.g. `https://bugs.launchpad.net/snapcraft/+filebug?field.title=please+add+support+for+{1}` ?

snapcraft/tests/test_commands_cleanbuild.py
chipaca: I think the tests will fail if lxd isn't installed?

snapcraft/lxd.py
```diff
@@ -0,0 +1,106 @@
...
+        while not_connected:
+            sleep(5)
+            try:
+                self._container_run(['python3', '-c', _NETWORK_PROBE_COMMAND])
+                not_connected = False
+            except CalledProcessError as e:
+                retry_count -= 1
+                if retry_count == 0:
+                    raise e
+                pass
```
chipaca: `pass` is spurious there.

Personally I would've done this without the flag variable, but that's probably because I don't like flag variables :-)
```
while True:
    sleep(5)
    try: ...
    except: ...
    else: break # done
```

chipaca:
:+1:

snapcraft/lxd.py
sergiusens: Yeah, I just find the `try/except/else` mechanics confusing

snapcraft/common.py
sergiusens: done

snapcraft/commands/cleanbuild.py
sergiusens: done

snapcraft/commands/cleanbuild.py
sergiusens: done

snapcraft/lxd.py
chipaca: Fair point.

PR #331, opened 2016-02-22
Catkin plugin: Handle rosdep empty and multiple dependencies.

kyrofa:
This PR resolves both LP: [#1548406](https://bugs.launchpad.net/snapcraft/+bug/1548406) and [#1548404](https://bugs.launchpad.net/snapcraft/+bug/1548404) by refactoring the `rosdep` dependency resolution code to account for an empty result as well as multiple results on one line.

Note that both of these bugs are fixed here since, while they're separate bugs, the fix for both is very closely related and it would be more work to split them apart than to just do both.

### DIFF
--- a/snapcraft/plugins/catkin.py
+++ b/snapcraft/plugins/catkin.py
@@ -377,26 +377,31 @@ def _find_system_dependencies(catkin_packages, rosdep):
             # weren't instructed to build. It's probably a system dependency,
             # but the developer could have also forgotten to tell us to build
             # it.
-            system_dependency = rosdep.resolve_dependency(dependency)
-
-            if not system_dependency:
+            try:
+                these_dependencies = rosdep.resolve_dependency(dependency)
+            except SystemDependencyNotFound:
                 raise RuntimeError(
-                    'Package "{}" isn\'t a valid system dependency. '
-                    'Did you forget to add it to catkin-packages? If '
-                    'not, add the Ubuntu package containing it to '
-                    'stage-packages until you can get it into the '
-                    'rosdep database.'.format(dependency))
+                    "Package {!r} isn't a valid system dependency. "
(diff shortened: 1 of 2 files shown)


snapcraft/plugins/catkin.py
```diff
@@ -350,9 +350,9 @@ def _find_system_dependencies(catkin_packages, ros_distro, ros_package_path,
...
             # but the developer could have also forgotten to tell us to build
             # it.
-            system_dependency = rosdep.resolve_dependency(dependency)
-
-            if not system_dependency:
+            try:
+                these_dependencies = rosdep.resolve_dependency(dependency)
+            except SystemDependencyNotFound:
                 raise RuntimeError(
                     'Package "{}" isn\'t a valid system dependency. '
```
sergiusens: s/"{}"/{!r}/

snapcraft/plugins/catkin.py
```diff
@@ -446,20 +451,16 @@ def resolve_dependency(self, dependency_name):
             output = self._run(['resolve', dependency_name, '--rosdistro',
                                 self._ros_distro, '--os', 'ubuntu:trusty'])
         except subprocess.CalledProcessError:
-            return None
+            raise SystemDependencyNotFound(
+                '"{}" does not resolve to a system dependency'.format(
```
sergiusens: s/"{}"/{!r}/

snapcraft/plugins/catkin.py
kyrofa: Done.

snapcraft/plugins/catkin.py
kyrofa: Done.

sergiusens:
:+1:

thanks!

PR #333, opened 2016-02-23
Calling snapcraft with no args defaults to snap

sergiusens:
Calling snapcraft with no arguments should default to just calling
the `snap` command for easier use.

LP: #1548915

### DIFF
--- a/snapcraft/main.py
+++ b/snapcraft/main.py
@@ -20,6 +20,7 @@ snapcraft
 
 Usage:
   snapcraft [--version | --help] [options] COMMAND [ARGS ...]
+  snapcraft [options]
 
 Options:
   -h --help        show this help message and exit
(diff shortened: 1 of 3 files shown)


snapcraft/main.py
```diff
@@ -81,25 +86,27 @@
     'upload',
 ]
 
-try:
-    version = pkg_resources.require('snapcraft')[0].version
-except pkg_resources.DistributionNotFound:
-    version = 'devel'
+
+def _version():
```
elopio: I would call this _get_version().

snapcraft/main.py
```diff
@@ -81,25 +86,27 @@
...
-        commands.load(args['COMMAND']).main(argv=args['ARGS'])
+        commands.load(cmd).main(argv=args['ARGS'])
     except Exception as e:
         sys.exit(textwrap.fill(str(e)))
 
 
-if __name__ == '__main__':
-    main()
+if __name__ == '__main__':  # pragma: no cover
+    main()                  # pragma: no cover
```
elopio: But we are covering it with integration tests, aren't we?

elopio:
lgtm. Just a couple of comments.

snapcraft/main.py
sergiusens: sure

sergiusens: but wait 'getters' in python were just the name, same for 'setters'; wouldn't this be a property in some sense?

snapcraft/main.py
sergiusens: integration tests do cover this, but it is meaningless (and unittests don't get 100% either :wink:)

elopio:
:+1:

PR #334, opened 2016-02-23
More attributes for the squashfs'ed snap

sergiusens:
To make security better and easier to recreate snaps
we need to add -all-root, we also add -no-xattrs since
it is also an improvement that helps in preventing
people trying to circumvent the system.

LP: #1546821

### DIFF
--- a/snapcraft/commands/snap.py
+++ b/snapcraft/commands/snap.py
@@ -74,6 +74,9 @@ def main(argv=None):
     snap_name = args['--output'] or _format_snap_name(snap)
 
     logger.info('Snapping {}'.format(snap_name))
+    # These options need to match the review tools:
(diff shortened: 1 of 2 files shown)


elopio:
:+1:, pending a sanity check from Jamie.

sergiusens:
@elopio so jamie already commented before you :wink:

sergiusens:
I swear I saw jamie ack and I based this out from a diff he was working on.

jdstrand:
I did ACK this-- not sure where it went...

sergiusens:
Thanks for confirming I'm not crazy! :-)

elopio:
due to the lack of evidence, I can't be sure about who is not crazy here.

PR #335, opened 2016-02-25
Redo pkg-config handling with library collecting

sergiusens:
pkg-config only handled PKG_CONFIG_SYSROOT_DIR as the stage dir
providing many issues when using packages from installdir for a part
and also build-packages from the host.

For a full implementation, elf dependency tracking from the host
to the snap is also implemented here

LP: #1549570


libraries/16.04
```diff
@@ -0,0 +1,513 @@
...
+libunwind-ptrace.so.0
+libunwind-ptrace.so.0.0.0
+libunwind-x86_64.so.8
+libunwind-x86_64.so.8.0.1
+libunwind.so.8
+libunwind.so.8.0.1
+libustr-1.0.so.1
+libustr-1.0.so.1.0.4
+libyaml-0.so.2
+libyaml-0.so.2.0.4
```
elopio: How did you generate this list?
A README file next to this one explaining why and how would be great.

snapcraft/common.py
```diff
@@ -30,6 +31,9 @@
 _plugindir = _DEFAULT_PLUGINDIR
 _DEFAULT_SCHEMADIR = '/usr/share/snapcraft/schema'
 _schemadir = _DEFAULT_SCHEMADIR
+_DEFAULT_LIBRARIESDIR = 'usr/share/snapcraft/libraries'
```
elopio: _DEFAULT_LIBRARIES_DIR

elopio: ah, nevermind, it's following the style of the other constants.

snapcraft/common.py
```diff
@@ -58,8 +62,9 @@ def run_output(cmd, **kwargs):
         f.write('\n')
         f.write('exec $*')
         f.flush()
-        return subprocess.check_output(['/bin/sh', f.name] + cmd,
-                                       **kwargs).decode('utf8').strip()
+        return subprocess.check_output([
+            '/bin/sh', f.name] + cmd, **kwargs).decode(
+                sys.getfilesystemencoding()).strip()
```
elopio: This looks good, but it means that if f.name prints a character that's not ascii when the system doesn't have utf8, it will raise an exception. This was the problem we had in a couple of places. So this possible error might need better handling.

snapcraft/libraries.py
```diff
@@ -48,3 +56,40 @@ def _extract_ld_library_paths(ld_conf_file):
...
+def _get_system_libs():
+    global _libraries
+    if _libraries:
+        return _libraries
+
+    release = platform.linux_distribution()[1]
+    lib_path = os.path.join(common.get_librariesdir(), release)
+
+    if not os.path.exists(lib_path):
+        logger.warning('No libraries to exclude from this release')
```
elopio: From the text in this message, it's not clear why it is a warning. How can affect the execution that there are no libraries to exclude?

elopio:
lgtm, but the elf bits are flying over my head. I'll read more about that tomorrow.
You might want to get chipaca or mvo to take a look.

snapcraft/yaml.py
```diff
@@ -320,30 +323,95 @@ def _build_env(root):
...
+    env['PKG_CONFIG_SYSROOT_DIR'] = {{basedir}}
+
+    return current_env.update(env)
+
+
+def modules_exist(modules, env):
+    try:
+        check_call(['/usr/bin/pkg-config', '--exists'] + modules, env=env)
+        return True
+    except:
```
elopio: except subprocess.CalledProcessError.
We don't want to ignore any other exceptions.

snapcraft/libraries.py
sergiusens: ah, `logger.debug` was supposed to be it

snapcraft/common.py
sergiusens: I went under the assumption that stdout and stderr are also just files. In fairness I was trying to get rid of this error:
`'utf-8' codec can't encode character '\\udcc3' in position 137:\nsurrogates not allowed\n"`

from http://162.213.35.179:8080/job/github-snapcraft-integration-tests-cloud/23/testReport/tests/TestSnapcraftExamples/test_example_java_hello_world_/

snapcraft/common.py
sergiusens: yeah; I really want to to a massive replace of all the `<something>dir` to be `<something>_dir`

libraries/16.04
sergiusens: Sounds like a good idea, so far I did it manually with find from an OS snap. This will probably need updating from my Qt comments from last night

sergiusens:
@elopio wrt `ELF` I can ldd against ELF (Executable Linking Format) files. I am walking the whole tree in case someone goes with a non standard layout.

snapcraft/common.py
chipaca: This changes the signature of `run_output` from returning a string to returning bytes. This on its own won't raise an exception, but the caller needs to handle it differently from before (because it's different).

snapcraft/libraries.py
```diff
@@ -48,3 +56,40 @@ def _extract_ld_library_paths(ld_conf_file):
...
+
+def get_dependencies(elf):
+    """Return a list of libraries that are needed to satisfy elf's runtime."""
+    logger.debug('Getting dependencies for {!r}'.format(elf))
+    ldd_out = common.run_output(['ldd', elf]).split('\n')
+    ldd_out = [l.split() for l in ldd_out]
+    ldd_out = [l[2] for l in ldd_out if len(l) > 2 and os.path.exists(l[2])]
+
+    # Anything that is part of the project we don't need to check for
+    libs = list(filter(lambda x: not x.startswith(os.getcwd()), ldd_out))
```
chipaca: two (silly) things that bother me about this line: you're calling `os.getcwd` in a loop without expecting it to change, and you switched from list comprehensions to filter+lambda.

snapcraft/libraries.py
```diff
@@ -48,3 +56,40 @@ def _extract_ld_library_paths(ld_conf_file):
...
+
+    # Now lets filter out what would be on the system
+    system_libs = _get_system_libs()
+    libs = list(filter(lambda x: not os.path.basename(x) in system_libs, libs))
```
chipaca: and here it seems you switched to filter+lambda entirely. Not sure I understand why you're using both idioms in a single function (and why you use filter but flatten it into a list for an intermediate result)

snapcraft/libraries.py
```diff
@@ -48,3 +56,40 @@ def _extract_ld_library_paths(ld_conf_file):
...
+        return _libraries
+
+    release = platform.linux_distribution()[1]
+    lib_path = os.path.join(common.get_librariesdir(), release)
+
+    if not os.path.exists(lib_path):
+        logger.warning('No libraries to exclude from this release')
+
+    with open(lib_path) as fn:
+        _libraries = fn.read().split()
```
chipaca: From what I can tell, you could make `_libraries` a `frozenset` and carry on using it as you are, but slightly faster.

snapcraft/pluginhandler.py
```diff
@@ -334,6 +339,32 @@ def _migrate_files(snap_files, snap_dirs, srcdir, dstdir, missing_ok=False):
...
+
+    elf_files = set()
+    for root, dirs, files in os.walk(workdir):
+        for entry in itertools.chain(files, dirs):
+            path = os.path.join(root, entry)
+            if os.path.islink(path):
+                logger.debug('Skipped link {!r} when parsing {!r}'.format(
+                    path, workdir))
+                continue
+            if ms.file(path).startswith('ELF'):
```
chipaca: you should check it's a dynamic executable, also

sergiusens:
thanks @chipaca

libraries/16.04
kyrofa: Agreed, this could use some documentation.

sergiusens:
retest this please

sergiusens:
retest this please

sergiusens:
retest this please

examples_tests/__init__.py
```diff
@@ -125,8 +125,8 @@ def build_snap(self, example_dir):
             [self.snapcraft_command, 'clean'], cwd=working_dir)
         try:
             subprocess.check_output(
-                [self.snapcraft_command, 'snap'], cwd=working_dir,
-                stderr=subprocess.STDOUT).decode('utf-8')
+                [self.snapcraft_command, '--debug', 'snap'], cwd=working_dir,
+                stderr=subprocess.STDOUT)
```
sergiusens: @elopio any idea what is going on here?

@elopio I'm lost as with what is going on here:

```
  File "/home/jenkins-slave/workspace/github-snapcraft-examples-tests-cloud/examples_tests/tests.py", line 98, in test_example
  File "/home/jenkins-slave/workspace/github-snapcraft-examples-tests-cloud/examples_tests/__init__.py", line 129, in build_snap
    stderr=subprocess.STDOUT).decode('utf-8')
    **kwargs).stdout
    output=stdout, stderr=stderr)
```

Is jenkins running `python3 setup.py develop` by any chance?

sergiusens:
retest this please

libraries/16.04
sergiusens: @elopio and @kyrofa done https://github.com/ubuntu-core/snapcraft/pull/335/files#diff-2cb561c5e46512465f54a0f84d421c80R66

But is also makes me think if it is worth to make this dynamic.

HACKING.md
```diff
@@ -62,3 +62,14 @@ environment variables:
 - **tests:** Tests, obviously. `unit` holds Python unit tests and `plainbox` holds plainbox integration tests.
 
 - **snapcraft:** The Python module that houses the core snapcraft logic. The `plugins` subdirectory holds the code for each plugin.
+
+### Updating library filter
+
+To update the list of libraries that get excluded from inclusion into a
+snap run:
+
+    ./libraries/generate_list.py libraries/<release>
```
kyrofa: Should this be `generate_lib_list.py`?

libraries/16.04
kyrofa: How would you go about making it dynamic? Would you have to download the OS snap every time?

sergiusens:
retest this please

HACKING.md
```diff
@@ -62,3 +62,14 @@ environment variables:
...
+### Updating library filter
+
+To update the list of libraries that get excluded from inclusion into a
+snap run:
+
+    ./libraries/generate_lib_list.py libraries/<release>
+
+e.g.; to update the list for 16.04,
+
+    ./libraries/generate_lib_list.py libraries/16.04
```
didrocks: What do you think of having this running as a git release hook? I'm afraid that updates will be sporadic otherwise and ofc, there is no way of getting that running during package build (as accessing external servers).

kyrofa:
I'm investigating why this doesn't work for the opencv example.

kyrofa:
So the reason this doesn't work with opencv is the same reason is won't work for anything with a lib in a non-standard area (non-standard being defined as anything not placed in LD_LIBRARY_PATH by `yaml.py`): it doesn't include the ld.so.conf files parsed to determine which libs to include. What's the best way to work around this?

snapcraft/pluginhandler.py
```diff
@@ -399,6 +398,7 @@ def strip(self, force=False):
         self.notify_stage('Stripping')
         snap_files, snap_dirs = self.migratable_fileset_for('snap')
         _migrate_files(snap_files, snap_dirs, self.stagedir, self.snapdir)
+        _copy_dependencies(self.snapdir)
```
sergiusens: Needs tracking and library paths used need to be exported to `env`
hint for @kyrofa

snapcraft/pluginhandler.py
```diff
@@ -265,7 +262,7 @@ def pull(self, force=False):
         package_files = set()
         package_directories = set()
         if self.code.stage_packages:
-            package_files, package_directories = self._setup_stage_packages()
+            self._setup_stage_packages()
         self.code.pull()
 
         # Record the files and directories unpacked from the stage packages
```
sergiusens: We probably don't need this anymore
hint @kyrofa

snapcraft/pluginhandler.py
```diff
@@ -633,6 +662,56 @@ def _clean_migrated_files(snap_files, snap_dirs, directory):
...
 
+def _linktree(source, destination):
+    for root, directories, files in os.walk(source):
+        for entry in itertools.chain(files, directories):
+            source_path = os.path.join(root, entry)
+            relpath = os.path.relpath(source_path, source)
+            destination_path = os.path.join(destination, relpath)
+
+            if os.path.isdir(source_path) and not os.path.islink(source_path):
+                os.makedirs(destination_path, exist_ok=True)
```
sergiusens: @kyrofa do we need to create these potential empty dirs? The `else` clause seems will be able to ensure the directories exist

sergiusens:
@elopio given these pass on `amd64` can we give this a run on `armhf` ?

sergiusens:
@elopio I'm still seeing *testtools.testresult.real._StringException: output: {{{No module named 'magic'}}}*

snapcraft/yaml.py
```diff
@@ -272,10 +272,16 @@ def build_env_for_part(self, part, root_part=True):
             env += self.build_env_for_part(dep_part, root_part=False)
 
         if root_part:
+            # this has to come before any {}/usr/bin
+            env += _create_pkg_config_override(
+                part.bindir, part.installdir, stagedir)
             env += part.env(part.installdir)
             env += _runtime_env(stagedir)
             env += _runtime_env(part.installdir)
+            env += _runtime_env(part.ubuntu_unpackdir)
```
sergiusens: @kyrofa python plugins fail to work, I'm thinking it might be due to this, did it work for you?

elopio:
retest this please

elopio:
retest this please

sergiusens:
Thanks @elopio

sergiusens:
@elopio final call is on you :wink:

elopio:
flashing...

sergiusens:
@elopio how is it doing?

elopio:
@sergiusens slooooow... Running in the snappy classic in rpi2. All unit tests passed. The integration tests have been running for an hour:

(classic)ubuntu@localhost:~/snapcraft$ ./runtests.sh integration
.................[sudo] password for ubuntu: 
EEEE.........[sudo] password for ubuntu: 
E....[sudo] password for ubuntu: 
E..

sergiusens:
@elopio hmm, maybe we can fix the sudo issue for armhf this cycle :-)

elopio:
@sergiusens I can take a look at that.
The tests that could run, ran successfully. The tests that need sudo failed. 
:+1: from me to land this. I'll see if I can get the failed tests to pass. If they uncover an unlikely error, we can still fix it before the release.

PR #338, opened 2016-02-25
Catch a failure to detach for the log

sergiusens:
A failure to detach should be a noop just like the checked
for conditions.

LP: #1549831


elopio:
seems harmless to me. +1.

elopio:
Merge with master, please.

PR #339, opened 2016-02-25
Try to create the travis lxc image adding a remote.

elopio:


elopio:
Closing, the problem is solved by updating the grep command.
Maybe later we'll want to change to use this lxd remote.

PR #340, opened 2016-02-25
travis: update the command to wait for the network

elopio:
LP: #1549880


sergiusens:
This is already a :+1: IMO :smile:

PR #343, opened 2016-02-26
lxd should be a suggestion, not a direct dep

sergiusens:
LP: #1550381


elopio:
yes, good. :+1:

PR #346, opened 2016-02-27
CMake plugin: Support installing via prefix.

kyrofa:
Currenty the CMake plugin uses an empty CMAKE_INSTALL_PREFIX and installs into the install directory via utilizing DESTDIR. This PR resolves LP: [#1544630](https://bugs.launchpad.net/snapcraft/+bug/1544630) by adding the ability to install via actually using CMAKE_INSTALL_PREFIX.

Also ups the CMake plugin coverage to 100%.

### DIFF
--- a/snapcraft/plugins/cmake.py
+++ b/snapcraft/plugins/cmake.py
@@ -28,6 +28,9 @@ Additionally, this plugin uses the following plugin-specific keywords:
     - configflags:
       (list of strings)
       configure flags to pass to the build using the common cmake semantics.
+    - install-via:
+      (enum, 'destdir' or 'prefix'; default 'destdir')
+      Whether to install via DESTDIR or CMAKE_INSTALL_PREFIX
 """
 
 import os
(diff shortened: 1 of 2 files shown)


sergiusens:
Looks good; `install-via` doesn't feel like the right keyword though. Maybe we need to sleep on it for creative thought to magically come up with something (it might as well be `install-via`).

It would be good to have some integration tests too.

kyrofa:
Yeah, I'm not a huge fan either but I wasn't coming up with anything better. I'll work on the integration tests-- mull this over.

kyrofa:
Okay so this is a little embarrassing but I made this change to the wrong plugin-- it needs to be in autotools. After a bit of research I'm not actually sure CMake needs this change at all. I'll close this (at least for now).

fenollp:
Hi there, pretty sure this is still happening. Had to set the install prefix to /bin (and ignore $DESTDIR): https://github.com/fenollp/voidstar/pull/62/files#diff-5fde7a6d86053f0e1d88c0a2a238941fR28

Not sure what the best way to fix this is.
https://bugs.launchpad.net/snapcraft/+bug/1544630 is marked as `invalid` for some reason...

PR #350, opened 2016-02-29
Clear error when specifying a non existing command

sergiusens:
When a command is supposed to be in PATH, make sure
the error is clear.

LP: #1550496

### DIFF
--- a/snapcraft/meta.py
+++ b/snapcraft/meta.py
@@ -19,6 +19,7 @@ import logging
 import re
 import shlex
 import shutil
+import subprocess
 import tempfile
 
 import yaml
@@ -49,6 +50,10 @@ _OPTIONAL_PACKAGE_KEYS = [
 _OPTIONAL_HOOKS = [
     'config',
 ]
+
+
+class CommandError(Exception):
+    pass
 
 
 def create(config_data):
@@ -198,17 +203,8 @@ def _wrap_exe(command, basename=None):
 
     wrapexec = '$SNAP/{}'.format(execparts[0])
     if not os.path.exists(exepath) and '/' not in execparts[0]:
-        # If it doesn't exist it might be in the path
-        logger.debug('Checking to see if {!r} is in the $PATH'.format(
-            execparts[0]))
-        with tempfile.NamedTemporaryFile('w+') as tempf:
-            script = ('#!/bin/sh\n' +
-                      '{}\n'.format(common.assemble_env()) +
-                      'which "{}"\n'.format(execparts[0]))
-            tempf.write(script)
-            tempf.flush()
-            common.run(['/bin/sh', tempf.name], cwd=snap_dir)
-            wrapexec = execparts[0]
+        _find_bin(execparts[0], snap_dir)
+        wrapexec = execparts[0]
     else:
         with open(exepath, 'rb') as exefile:
             # If the file has a she-bang, the path might be pointing to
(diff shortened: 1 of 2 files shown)


snapcraft/tests/test_meta.py
```diff
@@ -549,3 +549,13 @@ def test_exe_is_in_path(self, run_mock):
             wrapper_contents = wrapper_file.read()
 
         self.assertEqual(expected, wrapper_contents)
+
+    def test_command_does_not_exist(self):
```
kyrofa: Can you add a test for a file that's simply not executable as well?

kyrofa:
I'd like one more test in there, but other than that this looks great.

snapcraft/meta.py
```diff
@@ -222,10 +218,28 @@ def _wrap_exe(command, basename=None):
     return os.path.relpath(wrappath, snap_dir)
 
 
+def _find_bin(binary, basedir):
+    # If it doesn't exist it might be in the path
+    logger.debug('Checking to see if {!r} is in the $PATH'.format(binary))
```
elopio: s/Checking to see if/Checking if

kyrofa: s/Checking if/Checking that/ in that case

sergiusens:
@kyrofa done

kyrofa:
Looks perfect! :+1:

sergiusens:
@kyrofa just for the kicks I've added a sunny day scenario as well

snapcraft/meta.py
sergiusens: meh, old code :-P

snapcraft/meta.py
```diff
@@ -222,10 +218,28 @@ def _wrap_exe(command, basename=None):
...
         for k in [k for k in ('command', 'stop-command') if k in apps[app]]:
-            apps[app][k] = _wrap_exe(
-                apps[app][k], '{}-{}'.format(k, app))
-
+            try:
+                apps[app][k] = _wrap_exe(apps[app][k], '{}-{}'.format(k, app))
+            except CommandError as e:
+                raise EnvironmentError(
+                    'The specified command {!r} defined in {!r} does '
+                    'not exist'.format(str(e), app))
```
elopio: So, should it be: does not exist or is not executable ?

kyrofa: I guess that's a good question. Are the exit codes 126 and 127?

snapcraft/tests/test_meta.py
```diff
@@ -549,3 +549,48 @@ def test_exe_is_in_path(self, run_mock):
...
+        os.makedirs(os.path.dirname(cmd_path))
+        open(cmd_path, 'w').close()
+
+        with self.assertRaises(EnvironmentError) as raised:
+            meta._wrap_apps(apps)
+        self.assertEqual("The specified command 'command-not-executable' "
+                         "defined in 'app1' does not exist",
+                         str(raised.exception))
+
+    def test_command_found(self):
```
kyrofa: Oh yeah. That's probably a good idea :stuck_out_tongue: .

snapcraft/meta.py
sergiusens: If it is not executable it inherently does not exist from a "command" PoV.

sergiusens: @kyrofa and @elopio 

From the manpage
```
       1      if  one  or  more  specified commands is nonexistent or not exe-
              cutable
```

I'll make the change, too bad since travis already ran :weary: 
To be fair, people using the copy plugin get what they deserve! This would not happen when using a proper build system :wink:

PR #351, opened 2016-02-29
Autotools plugin: Support installing via prefix.

kyrofa:
Currently autotools installs via DESTDIR, but not all autotools projects work that way. This PR fixes LP: [#1544629](https://bugs.launchpad.net/snapcraft/+bug/1544629) by adding a config option that enables the prefix to be used instead.


elopio:
:+1: from me. I don't like install-via, but I can't think of anything better.
I think that install-to: prefix or install-to: destdir reads a little better, but I'm ok with anything Kyle decides.

sergiusens:
The problem with `install-to` is that both put the files in the same location, what changes is the mechanism

sergiusens:
:+1:

PR #353, opened 2016-02-29
Skip the BasePlugin from plugin search

sergiusens:
This is to support the case of a plugin importing baseplugin
as a symbol instead of the module.

LP: #1547681


kyrofa:
Yeah looks fine to me :+1:

elopio:
lgtm.

PR #355, opened 2016-03-01
Expose the invalid build-package on error

sergiusens:
The error that results from using an invalid build-package
is not very specific. This exposes the KeyError raised by apt.

LP: #1547672

### DIFF
--- a/integration_tests/test_snap.py
+++ b/integration_tests/test_snap.py
@@ -131,8 +131,8 @@ class SnapTestCase(integration_tests.TestCase):
         exception = self.assertRaises(
             subprocess.CalledProcessError, self.run_snapcraft, 'snap')
         expected = (
-            'Could not find all the "build-packages" required in '
-            'snapcraft.yaml\n')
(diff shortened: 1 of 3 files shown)


kyrofa:
Yeah, much better! :+1:

integration_tests/test_snap.py
```diff
@@ -131,8 +131,8 @@ def test_error_with_unexistent_build_package(self):
         exception = self.assertRaises(
             subprocess.CalledProcessError, self.run_snapcraft, 'snap')
         expected = (
-            'Could not find all the "build-packages" required in '
-            'snapcraft.yaml\n')
+            "Could not find a required package in 'build-packages': "
+            '"The cache has\nno package named \'inexistent-package\'"\n')
```
kyrofa: The newline in here is throwing me. Why is it here and not in the other test below?

sergiusens: `dedent` usage in `snapcraft.main`

PR #357, opened 2016-03-01
Support parallel builds (enabled by default).

kyrofa:
Several plugins within Snapcraft support parallel builds, but Snapcraft currently doesn't take advantage of this and builds them with a single job. This PR resolves LP: [#1551417](https://bugs.launchpad.net/snapcraft/+bug/1551417) by adding support for parallel builds, and enabling them by default in the autotools, cmake, and make plugins. It also adds an option to disable parallel builds via the `--no-parallel-build` option for the case of running in a RAM-limited environment.

### DIFF
--- a/snapcraft/main.py
+++ b/snapcraft/main.py
@@ -23,11 +23,14 @@ Usage:
   snapcraft [options]
 
 Options:
-  -h --help        show this help message and exit
-  -v --version     show program version and exit
-  -V --verbose     print additional information about command execution
-  -d --debug       print debug information while executing (including
-                   backtraces)
+  -h --help              show this help message and exit
+  -v --version           show program version and exit
+  -V --verbose           print additional information about command execution
+  -d --debug             print debug information while executing (including
+                         backtraces)
+  --no-parallel-build    use only a single build job per part (the default
+                         number of jobs per part is equal to the number of
+                         CPUs)
 
 The available commands are:
   list-parts   List available parts which are like “source packages” for snaps.
@@ -43,7 +46,9 @@ The available lifecycle commands are:
   clean        Remove content - cleans downloads, builds or install artifacts.
   cleanbuild   Create a snap using a clean environment managed by lxd.
   pull         Download or retrieve artifacts defined for a part.
-  build        Build artifacts defined for a part.
+  build        Build artifacts defined for a part. Build systems capable of
+               running parallel build jobs will do so unless
+               --no-parallel-build is specified.
   stage        Stage the part's built artifacts into the common staging area.
   strip        Final copy and preparation for the snap.
   snap         Create a snap.
@@ -66,6 +71,7 @@ from docopt import docopt
 from snapcraft import (
     log,
     commands,
+    common,
 )
 
 logger = logging.getLogger(__name__)
@@ -109,6 +115,8 @@ def main():
         log_level = logging.DEBUG
 
     log.configure(log_level=log_level)
+
+    common.set_enable_parallel_builds(not args['--no-parallel-build'])
 
     try:
         commands.load(cmd).main(argv=args['ARGS'])
--- a/snapcraft/common.py
+++ b/snapcraft/common.py
@@ -17,6 +17,8 @@
 # Data/methods shared between plugins and snapcraft
 
 import glob
+import logging
+import multiprocessing
 import os
 import platform
 import subprocess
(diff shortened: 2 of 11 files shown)


snapcraft/main.py
```diff
@@ -23,11 +23,13 @@
...
-  -V --verbose     print additional information about command execution
-  -d --debug       print debug information while executing (including
-                   backtraces)
+  -h --help              show this help message and exit
+  -v --version           show program version and exit
+  -V --verbose           print additional information about command execution
+  -d --debug             print debug information while executing (including
+                         backtraces)
+  --no-parallel-build    use only a single build job per part (defaults to
+                         number of CPUs)
```
elopio: This default is confusing. A little. I think I need more words, but I can't find a good way to say it.
Something like: defaults to use the number of CPUs to build parts in parallel.
The problem with this one is that it sounds like you will be building many parts at the same time, instead of building one part at a time.

kyrofa: Hmm, how about "the default number of jobs per part is equal to the number of CPUs"?

elopio: yes, sounds better, IMO.

snapcraft/main.py
```diff
@@ -110,6 +113,8 @@ def main():
 
     log.configure(log_level=log_level)
 
+    common.set_enable_parallel_builds(not args['--no-parallel-build'])
+
     try:
         commands.load(cmd).main(argv=args['ARGS'])
```
elopio: I hate this common and its global variables. Could you delay using it? In here you can pass --no-parallel-build to the build main in argv, which is the only who cares about it. Would that make sense, or just make it worse?

snapcraft/main.py
kyrofa: Done.

snapcraft/main.py
kyrofa: I agree with you, and I thought about that... but it honestly seemed inconsistent with how we were doing other things. I'm worried that doing that here would fragment the codebase's consistency, which may be worse than disliking how we solve an overall problem. Thoughts?

elopio:
lgtm. I just hate the common and globals, but I don't have a great solution for that.
If we could reduce the common usage passing arguments at least in a few places, I think that would be better.

What about adding some docs? It would me nice to mention that not all the plugins will use multiple CPUs.

snapcraft/main.py
elopio: Sure, that's a great point. Let's just keep thinking about how to get rid of all the globals.

kyrofa: I'm with you there.

kyrofa:
@elopio I'm always on board with more docs, though in this case I'm not sure where to put them. This is such a transparent change I don't think it'll really change anyone's life other than making things faster (for some plugins). What did you have in mind?

elopio:
I thought about docs/snapcraft-usage. That file IMO needs some more content to explain some of the flags.
I also thought about the help itself, maybe in https://github.com/ubuntu-core/snapcraft/pull/357/files#diff-07ab19acf782b5f8bd05c71b8d04f963L46 Mention that by default some build systems will run in parallel.
Any of them sound good to you?

snapcraft/main.py
sergiusens: It would need a rename, but this should live in the `yaml` module which loads the `project` (there, that's a new name) and keeps the state of it (state which does not correspond to an individual plugin).

kyrofa: Yeah that sounds promising. Would we then pass around the `project` to each command?

kyrofa:
@elopio alright, how does https://github.com/ubuntu-core/snapcraft/pull/357/files#diff-07ab19acf782b5f8bd05c71b8d04f963R49 look?

snapcraft/main.py
sergiusens: so `<command>.main(project)` would work in most cases.

I correct myself; the `yaml` module already handle the plugins as it is part of the whole tree anyways.

This would potentially allow for parallelizing unit test run.

kyrofa: Yeah, that sounds nice. LP: [#1551849](https://bugs.launchpad.net/snapcraft/+bug/1551849).

elopio:
Perfect. Thanks.

elopio:
retest this please

PR #358, opened 2016-03-01
Set a sane umask to create a proper snap

sergiusens:
This seems to solve the bug, I'm not sure how to add tests for this though (and not even sure it is worth it).


kyrofa:
This looks sensible :+1: . Agreed that testing would be wonky.

PR #359, opened 2016-03-01
Do not print which's output to stdout

sergiusens:
This will print to stderr if there is a reason to, but that is a good thing.

LP: #1551939


kyrofa:
Ahhh, yes thank you! :+1:

PR #360, opened 2016-03-02
Set a sane umask to create a proper snap

sergiusens:
LP: #1515394

### DIFF
--- a/snapcraft/main.py
+++ b/snapcraft/main.py
@@ -62,6 +62,7 @@ http://developer.ubuntu.com/snappy/snapcraft
 """
 
 import logging
+import os
 import pkg_resources
 import sys
 import textwrap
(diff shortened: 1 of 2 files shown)


kyrofa:
Yeah I like it :+1:

sergiusens:
This will be super naggy though

kyrofa:
How so? Won't most people have 0022 or 0002

sergiusens:
@kyrofa Naggy in the sense it will be spat out after every command that's run.

kyrofa:
@sergiusens ah. I don't see a way around that, though, unless you touch, say, `parts/.umask_warned` or something. But honestly, enough is printed with every command I'm more worried this warning will be lost completely.

jdstrand:
LGTM, thanks!

PR #361, opened 2016-03-02
Support for a daemon's restart-condition

sergiusens:
Just makes it valid syntax in snapcraft.yaml

### DIFF
--- a/schema/snapcraft.yaml
+++ b/schema/snapcraft.yaml
@@ -168,6 +168,14 @@ properties:
             enum:
               - simple
               - forking
+          restart-condition:
+              type: string
+              enum:
+                - on-success
+                - on-failure
+                - on-abnormal
+                - on-abort
+                - always
           slots:
             type: array
             minitems: 1
--- a/snapcraft/tests/test_yaml.py
+++ b/snapcraft/tests/test_yaml.py
@@ -734,10 +734,48 @@ class TestValidation(tests.TestCase):
             'service3': {
                 'command': 'binary3',
                 'daemon': 'forking',
+            },
+            'service4': {
+                'command': 'binary4',
+                'daemon': 'simple',
+                'restart-condition': 'always',
             }
         }
 
         snapcraft.yaml._validate_snapcraft_yaml(self.data)
+
+    def test_valid_restart_conditions(self):
+        self.data['apps'] = {
+            'service1': {
+                'command': 'binary1',
+                'daemon': 'simple',
+            }
+        }
+        valid_conditions = ['always', 'on-success', 'on-failure',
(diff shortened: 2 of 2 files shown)


sergiusens:
retest this please

schema/snapcraft.yaml
```diff
@@ -168,6 +168,15 @@ properties:
             enum:
               - simple
               - forking
+          restart-condition:
+              type: string
+              enum:
+                - never
```
kyrofa: Should this be "no"?

snapcraft/tests/test_yaml.py
```diff
@@ -734,6 +734,11 @@ def test_valid_app_daemons(self):
             'service3': {
                 'command': 'binary3',
                 'daemon': 'forking',
+            },
+            'service4': {
+                'command': 'binary4',
+                'daemon': 'simple',
+                'restart-condition': 'always',
```
kyrofa: Worth adding tests for the other conditions?

kyrofa:
Couple minor comments, but looks good :+1:

schema/snapcraft.yaml
sergiusens: The policy, as for `daemon` is to bring in the exact `systemd` names

snapcraft/tests/test_yaml.py
sergiusens: Sure

schema/snapcraft.yaml
kyrofa: Which at least according to [this](https://www.freedesktop.org/software/systemd/man/systemd.service.html) is "no," not "never." Right? Or am I reading something wrong?

snapcraft/tests/test_yaml.py
kyrofa: Not sure how important the watchdog test is... I guess it's possible to be added later. Maybe we don't want the brittleness there.

schema/snapcraft.yaml
sergiusens: I grabbed them from here https://github.com/ubuntu-core/snappy/blob/master/docs/meta.md
So more research is in order

schema/snapcraft.yaml
```diff
@@ -168,6 +168,14 @@ properties:
...
               - simple
               - forking
+          restart-condition:
+              type: string
+              enum:
+                - on-success
+                - on-failure
+                - on-abnormal
+                - on-abort
+                - always
```
kyrofa: Completely removing the "no?"

sergiusens: Yes for now https://github.com/ubuntu-core/snappy/pull/567

PR #362, opened 2016-03-02
Remove the useles decode done in tests.

sergiusens:
The examples tests shell out and capture output but this output
is not used so we should not decode it.

LP: #1552403


kyrofa:
Yup :+1:

PR #364, opened 2016-03-03
Create Plainbox Provider plugin

jocave:
lp: #1552369

It would be useful if this plugin could be released on to the 1.x series as well.

### DIFF
--- a/integration_tests/snaps/simple-plainbox-provider/2016.com.example:simple/manage.py
+++ b/integration_tests/snaps/simple-plainbox-provider/2016.com.example:simple/manage.py
@@ -1,0 +1,11 @@
+#!/usr/bin/env python3
+from plainbox.provider_manager import setup, N_
+
+
+setup(
+    name='plainbox-provider-simple',
+    namespace='2016.com.example',
+    version="1.0",
+    description=N_("A really simple Plainbox provider"),
+    gettext_domain="2016_com_example_simple",
+)
--- a/integration_tests/snaps/simple-plainbox-provider/snapcraft.yaml
+++ b/integration_tests/snaps/simple-plainbox-provider/snapcraft.yaml
@@ -1,0 +1,11 @@
+name: test-package
+version: 0.1
+summary: Create a snap of a plainbox provider
+description: |
+    Create a snap of a very simple plainbox that could
+    then be used by a checkbox application
+
+parts:
+    simple-plainbox-provider:
+        plugin: plainbox-provider
+        source: ./2016.com.example:simple
--- a/snapcraft/plugins/plainbox_provider.py
+++ b/snapcraft/plugins/plainbox_provider.py
@@ -1,0 +1,80 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2016 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+""" Create parts containing a Plainbox test collection known as a provider.
+
+Plainbox is a toolkit consisting of python3 library, development tools,
+documentation and examples. It is targeted at developers working on testing or
+certification applications and authors creating tests for such applications.
+More information: http://plainbox.readthedocs.org/en/latest/
+
+To find out more about authoring a plainbox provider, see the following
+documentation: http://plainbox.readthedocs.org/en/latest/author/providers.html
+
+This plugin uses the common plugin keywords as well as those for "sources".
+For more information check the 'plugins' topic for the former and the
+'sources' topic for the latter.
+
+Plugin specific keywords:
+    - checkbox-dev-ppa:
+      (boolean)
+      Whether to extend the source list to include the Checkbox Development
+      PPA https://launchpad.net/~checkbox-dev/+archive/ubuntu/ppa. Defaults to
+      true as features to support Snappy Ubuntu Core have yet to land in the
+      distro repositories.
+"""
+
+import snapcraft
+
+
+class PlainboxProviderPlugin(snapcraft.BasePlugin):
+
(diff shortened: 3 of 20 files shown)


snappy-m-o:

snappy-m-o:

snapcraft/plugins/plainbox_provider.py
```diff
@@ -0,0 +1,57 @@
...
+deb http://${prefix}.ubuntu.com/${suffix}/ ${release} main restricted
+deb http://${prefix}.ubuntu.com/${suffix}/ ${release}-updates main restricted
+deb http://${prefix}.ubuntu.com/${suffix}/ ${release} universe
+deb http://${prefix}.ubuntu.com/${suffix}/ ${release}-updates universe
+deb http://${prefix}.ubuntu.com/${suffix}/ ${release} multiverse
+deb http://${prefix}.ubuntu.com/${suffix}/ ${release}-updates multiverse
+deb http://${security}.ubuntu.com/${suffix} ${release}-security main restricted
+deb http://${security}.ubuntu.com/${suffix} ${release}-security universe
+deb http://${security}.ubuntu.com/${suffix} ${release}-security multiverse
+deb http://ppa.launchpad.net/checkbox-dev/ppa/ubuntu ${release} main
```
zyga: I'd use stable PPA here really or make this customizable with an explicit knob.

integration_tests/snaps/simple-plainbox-provider/2016.com.example:simple/manage.py
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env python3
+from plainbox.provider_manager import setup, N_
+
+
+setup(
+    name='2016.com.example:simple',
```
zyga: I'd use ``plainbox-provider-example`` here with an explicit ``namespace="2016.com.example"``

jocave:
I've updated the code to make it possible to disable the Checkbox Development PPA. Also, switched to using an explicit namespace (kept the provider name "Simple" though").

elopio:
OK to test

elopio:
retest this please

jocave:
retest this please

elopio:
retest this please

elopio:
add to whitelist

elopio:
retest this please

snapcraft/plugins/plainbox_provider.py
```diff
@@ -0,0 +1,76 @@
...
+            'type': 'boolean',
+            'default': 'true',
+        }
+        return schema
+
+    def __init__(self, name, options):
+        super().__init__(name, options)
+        self.build_packages.extend(['python3-plainbox', 'intltool'])
+        if self.options.checkbox_dev_ppa:
+            self._PLUGIN_STAGE_SOURCES = self._DEV_PPA_PLUGIN_STAGE_SOURCES
```
elopio: this _PLUGIN_STAGE_SOURCES is not a constant, so it shouldn't be all capital letters.
_plugin_stage_sources.

elopio: sorry, not from your branch. After you merge I'll propose something to change it.

For you, it seems safer to overwrite the PLUGIN_STAGE_SOURCES @property, instead of dealing with the internal attribute.

elopio:
:+1: from me.
Can you please remove the icon? It's not needed, and I think it's better to keep the test snaps with only the required bits.

integration_tests/snaps/simple-plainbox-provider/snapcraft.yaml
```diff
@@ -0,0 +1,12 @@
+name: test-package
+version: 0.1
+summary: Create a snap of a plainbox provider
+description: |
+    Create a snap of a very simple plainbox that could
+    then be used by a checkbox application
+icon: icon.png
```
sergiusens: remove this entry please

sergiusens:
`git rm integration_tests/snaps/simple-plainbox-provider/icon.png`

jocave:
Pushed commit with icon file removed from simple-plainbox-provider integration test.

sergiusens:
`git mv integration_tests/test_plainbox_provider_plugin.py integration_tests/test_plugin_plainbox_provider.py`

snapcraft/plugins/plainbox_provider.py
```diff
@@ -0,0 +1,76 @@
...
+deb http://ppa.launchpad.net/checkbox-dev/ppa/ubuntu ${release} main
+'''
+
+    @classmethod
+    def schema(cls):
+        schema = super().schema()
+        schema['properties']['checkbox-dev-ppa'] = {
+            'type': 'boolean',
+            'default': 'true',
+        }
```
sergiusens: since you aren't overriding `properties` it seems you are making use of the common `source` options, can you add that piece of information to the doc string?

snapcraft/plugins/plainbox_provider.py
```diff
@@ -0,0 +1,76 @@
...
+        return schema
+
+    def __init__(self, name, options):
+        super().__init__(name, options)
+        self.build_packages.extend(['python3-plainbox', 'intltool'])
```
sergiusens: `build-packages` are installed onto the host from the hosts defined `sources.list[.d]`; in the documentation above you mention that plainbox features are not completely in the archive. How does adding `python3-plainbox` here play into this?

snapcraft/plugins/plainbox_provider.py
sergiusens: Yes, don't take advantage of internal implementation details which are subject to change :wink:

sergiusens:
Thanks for this, I have a bunch of open questions in there though.

Also, does this rely on python provided by the os snap? It is fine from my PoV, just want to make sure.

sergiusens:
Given the previous comment I would like to see an example and an example_test (which would run on a snappy system)

elopio:
The example test would be so nice, indeed.
@jocave take a look at the examples_tests dir. Give me a shout if you need a hand with that one.

jocave:
Thank you for the reviews, I will put an example together and try and address the questions ASAP.

jocave:
I have added integration tests to cover a set of what I think are expected use cases for this plugin. One of these fails with a python import error for module 'plainbox'. This is provided by the package python3-plainbox which is explicitly added to the build-package list by the plugin.

I would like to know how to guarantee that 'build-package' works as advertised.

sergiusens:
@jocave build-packages will just land on the host and be used to build the system, there is additional magic to crawl everything ELF and bring in any .so it needs not provided in the snap/part area.

stage-packages however, which is what you use in your examples are a bit tricky, specially when it comes to python as you will need to define a `PYTHONPATH` for this.

Given this still needs a few twists and turns I am going to close it for now. Feel free to reopen when ready.

PR #365, opened 2016-03-03
Slots are of an interface and not of a type

sergiusens:
LP: 1549427


zyga:
+1

I'm sorry about not catching this earlier. Looks good and matches the Interfaces spec.

kyrofa:
:+1:

PR #366, opened 2016-03-03
New plugin: kernel

sergiusens:
Includes 2 plugins with an example for each.

Jointly developed by Alexander Sack and Sergio Schvezov

LP: #1552168

### DIFF
--- a/examples/96boards-kernel/snapcraft.yaml
+++ b/examples/96boards-kernel/snapcraft.yaml
@@ -1,0 +1,31 @@
+name: 96boards-kernel
+version: 4.4.0
+summary: A 96boards kernel built from source
+description: This is the reference kernel from 96boards
+type: kernel
+
+parts:
+  kernel:
+    plugin: kernel
+    source: https://github.com/96boards/linux.git
+    source-type: git
+    kdefconfig: [defconfig, distro.config]
+    kconfigs:
+      - CONFIG_LOCALVERSION="-96boards"
+      - CONFIG_DEBUG_INFO=n
+      - CONFIG_SQUASHFS=m
+    kernel-initrd-modules:
+      - squashfs
+    kernel-initrd-firmware:
+      - lib/firmware/e100/d101m_ucode.bin
+      - lib/firmware/e100/d101s_ucode.bin
+      - lib/firmware/e100/d102e_ucode.bin
+    kernel-image-target: Image
+    kernel-image-path: arch/arm64/boot/Image
+    kernel-device-tree:
+      - qcom/apq8016-sbc
+      - qcom/msm8916-mtp
+  410c-firmware:
+    plugin: tar-content
+    source: firmware.tar # from developer.qualcomm.com v1.2
+    destination: lib/firmware
--- a/snapcraft/common.py
+++ b/snapcraft/common.py
@@ -34,7 +34,9 @@ _DEFAULT_PLUGINDIR = '/usr/share/snapcraft/plugins'
 _plugindir = _DEFAULT_PLUGINDIR
 _DEFAULT_SCHEMADIR = '/usr/share/snapcraft/schema'
 _schemadir = _DEFAULT_SCHEMADIR
-_arch = None
+
+host_machine = platform.machine()
+target_machine = host_machine
 
 env = []
 
@@ -68,25 +70,36 @@ def run_output(cmd, **kwargs):
                                        **kwargs).decode('utf8').strip()
 
 
-_DEB_TRANSLATIONS = {
+_ARCH_TRANSLATIONS = {
     'armv7l': {
-        'arch': 'armhf',
+        'kernel': 'arm',
+        'deb': 'armhf',
+        'cross-compiler-prefix': 'arm-linux-gnueabihf-',
+        'cross-build-packages': ['gcc-arm-linux-gnueabihf'],
         'triplet': 'arm-linux-gnueabihf',
     },
     'aarch64': {
-        'arch': 'arm64',
+        'kernel': 'arm64',
+        'deb': 'arm64',
+        'cross-compiler-prefix': 'aarch64-linux-gnu-',
+        'cross-build-packages': ['gcc-aarch64-linux-gnu'],
         'triplet': 'aarch64-linux-gnu',
     },
     'i686': {
-        'arch': 'i386',
+        'kernel': 'x86',
+        'deb': 'i386',
         'triplet': 'i386-linux-gnu',
     },
     'ppc64le': {
-        'arch': 'ppc64el',
+        'kernel': 'powerpc',
+        'deb': 'ppc64el',
+        'cross-compiler-prefix': 'gcc-powerpc64-linux-gnu',
+        'cross-build-packages': ['gcc-powerpc64-linux-gnu'],
         'triplet': 'powerpc64le-linux-gnu',
     },
     'x86_64': {
-        'arch': 'amd64',
+        'kernel': 'x86',
+        'deb': 'amd64',
         'triplet': 'x86_64-linux-gnu',
     }
 }
@@ -98,20 +111,38 @@ class PlatformError(Exception):
         super().__init__(
             '{0} is not supported, please log a bug at '
             'https://bugs.launchpad.net/snapcraft/+filebug?'
-            'field.title=please+add+support+for+{0}'.format(
-                platform.machine()))
+            'field.title=please+add+support+for+{0}'.format(target_machine))
+
+
+def get_machine_info(machine):
+    try:
+        return _ARCH_TRANSLATIONS[machine]
+    except KeyError:
+        raise PlatformError()
+
+
+def set_target_machine(deb_arch):
+    global target_machine
+    for machine in _ARCH_TRANSLATIONS:
+        if _ARCH_TRANSLATIONS[machine].get('deb', '') == deb_arch:
+            logger.info('Setting target machine to {!r}'.format(machine))
+            target_machine = machine
+            return
+
+    raise EnvironmentError(
+        'Cannot set machine from deb_arch {!r}'.format(deb_arch))
 
 
 def get_arch():
     try:
-        return _DEB_TRANSLATIONS[platform.machine()]['arch']
+        return _ARCH_TRANSLATIONS[target_machine]['deb']
     except KeyError:
         raise PlatformError()
 
 
 def get_arch_triplet():
     try:
-        return _DEB_TRANSLATIONS[platform.machine()]['triplet']
+        return _ARCH_TRANSLATIONS[target_machine]['triplet']
     except KeyError:
         raise PlatformError()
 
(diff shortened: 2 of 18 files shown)


snapcraft/common.py
```diff
@@ -68,25 +68,30 @@ def run_output(cmd, **kwargs):
                                        **kwargs).decode('utf8').strip()
 
 
-_DEB_TRANSLATIONS = {
+_ARCH_TRANSLATIONS = {
     'armv7l': {
-        'arch': 'armhf',
+        'kernel': 'arm',
+        'deb': 'armhf',
```
kyrofa: Eww

snapcraft/common.py
```diff
@@ -98,20 +103,36 @@ def __init__(self):
         super().__init__(
             '{0} is not supported, please log a bug at '
             'https://bugs.launchpad.net/snapcraft/+filebug?'
-            'field.title=please+add+support+for+{0}'.format(
-                platform.machine()))
+            'field.title=please+add+support+for+{0}'.format(_machine))
```
kyrofa: I know you didn't add the title in this PR, but I wanted to voice a small concern. We're introducing a dependency here on Launchpad's URL format. Do we know that will remain stable? What happens in a few years when this code is still used but Launchpad renames `title` to be `summary` for example? We'll still be printing this but people will get a 404.

snapcraft/plugins/kbuild.py
```diff
@@ -0,0 +1,181 @@
...
+If kconfigfile is provided this plugin will use the provided config file
+wholesale as the starting point instead of make $kdefconfig. In case user
+configures both a kdefconfig as well as kconfigfile, kconfigfile approach will
+be used.
+
+This plugin is based on the snapcraft.BasePlugin and supports the properties
+provided by that plus the following kbuild specific options with semantics as
+explained above:
+
+    - kdefconfig:
```
kyrofa: I suggest reversing the order of listing the variables and the above description. Much of the above description mention the below variables by name but the reader has no context as to what those variables are until they get to the bottom.

snapcraft/plugins/kbuild.py
```diff
@@ -0,0 +1,181 @@
...
+      (filepath)
+      path to file to use as base configuration. If provided this option wins
+      over kdefconfig. default: None
+
+    - kconfigs
+      (list of strings)
+      explicit list of configs to force; this will override the configs that
+      were set as base through kdefconfig and kconfigfile and dependent configs
+      will be fixed using the defaults encoded in the kbuild config
+      definitions.  If you dont want default for one or more implicit configs
```
kyrofa: s/dont/don't/

snapcraft/plugins/kbuild.py
```diff
@@ -0,0 +1,181 @@
...
+
+        with open(config_path, 'r') as f:
+            config = (
+                '{config_override}\n\n{config}\n{config_override}\n'.format(
+                    config_override=config, config=f.read()))
+
+        # note that prepending and appending the overrides seems
+        # only way to convince all kbuild versions to pick up the
+        # configs during oldconfig in .config
+        config = '{}\n\n{}'.format(config, '\n'.join(self.options.kconfigs))
```
kyrofa: I'm not sure I understand what's going on here. Let's say `self.options.kconfigs = [A]` and the `config_path` contents is `B`. Lines 142-144 would result in:

```
A

B
A
```

Right? Then line 149 would result in:

```
A

B
A

A
```

What am I missing?

snapcraft/plugins/kbuild.py
```diff
@@ -0,0 +1,181 @@
...
+
+        with open(config_path, 'w') as f:
+            f.write(config)
+
+    def do_remake_config(self):
+        # update config to include kconfig amendments using oldconfig
+        cmd = 'yes "" | make {} oldconfig'.format(
+            ' '.join(self.options.build_parameters))
```
kyrofa: Referring to the [subprocess security guidelines](https://docs.python.org/3/library/subprocess.html#security-considerations), consider using `shlex.quote()` here to escape user-supplied input due to the use of `shell=True` below.

snapcraft/plugins/kernel.py
```diff
@@ -0,0 +1,297 @@
...
+            'type': 'array',
+            'minitems': 0,
+            'uniqueItems': True,
+            'items': {
+                'type': 'string',
+            },
+            'default': [],
+        }
+
+        schema['properties']['kernel-device-tree'] = {
```
kyrofa: No docs for this option, it seems?

snapcraft/plugins/kbuild.py
sergiusens: Hah, I thought I wasn't doing `A B A` but it turns out I was doing the header part while doing the read and the footer while doing the write. I'll fix this to do it all in one location.

snapcraft/common.py
sergiusens: What, how is that possible? :smile:

If everyone agreed we'd be jobless :wink:

snapcraft/common.py
sergiusens: This is stable. If it comes to it, it would need to be changed like any other link we have.

snapcraft/plugins/kbuild.py
sergiusens: ack

sergiusens:
@kyrofa I would be pleased for a rereview of this :-)

unit test writing took me the entire day!

examples/96boards-kernel/snapcraft.yaml
```diff
@@ -0,0 +1,31 @@
...
+      - lib/firmware/e100/d101s_ucode.bin
+      - lib/firmware/e100/d102e_ucode.bin
+    kernel-image-target: Image
+    kernel-image-path: arch/arm64/boot/Image
+    kernel-device-tree:
+      - qcom/apq8016-sbc
+      - qcom/msm8916-mtp
+  410c-firmware:
+    plugin: tar-content
+    source: firmware.tar # from developer.qualcomm.com v1.2
```
kyrofa: I haven't quite woken up yet, but where is this coming from to make the example succeed?

snapcraft/common.py
```diff
@@ -68,25 +70,32 @@ def run_output(cmd, **kwargs):
...
+        'kernel': 'arm',
+        'deb': 'armhf',
         'triplet': 'arm-linux-gnueabihf',
     },
     'aarch64': {
-        'arch': 'arm64',
+        'kernel': 'arm64',
+        'deb': 'arm64',
+        'cross-compiler-prefix': 'aarch64-linux-gnu-',
+        'cross-build-packages': ['gcc-aarch64-linux-gnu'],
```
kyrofa: Should we add these for other supported architectures as well?

sergiusens: we should, yes.

examples/96boards-kernel/snapcraft.yaml
sergiusens: from developer.qualcomm.com; you need to accept a license agreement.

sergiusens: FWIW anyone building such qualcomm kernel would know they need to do this so I didn't explain much. This will also be accompanied by a blog post (kernel plugin is subject to change).

kyrofa: Ah, okay. I wish we could have a more complete example is all, I guess. One that a dev can just run instead of needing to hunt these down.

snapcraft/plugins/kernel.py
```diff
@@ -0,0 +1,302 @@
...
+from snapcraft import storeapi
+from snapcraft.common import get_machine_info, target_machine
+from snapcraft.config import load_config
+from snapcraft.plugins import kbuild
+
+logger = logging.getLogger(__name__)
+
+
+_compression_command = {
+    'gz': 'gzip',
```
kyrofa: Don't the docs above state that xz is also an option here?

snapcraft/plugins/kernel.py
```diff
@@ -0,0 +1,302 @@
...
+}
+
+
+class KernelPlugin(kbuild.KBuildPlugin):
+
+    @classmethod
+    def schema(cls):
+        schema = super().schema()
+
+        schema['properties']['kernel-image-target'] = {
```
kyrofa: Type, or target?

snapcraft/plugins/kernel.py
```diff
@@ -0,0 +1,302 @@
...
+            'type': 'array',
+            'minitems': 1,
+            'uniqueItems': True,
+            'items': {
+                'type': 'string',
+            },
+            'default': [],
+        }
+
+        schema['properties']['kernel-device-trees'] = {
```
kyrofa: Docs diverge here as well, plural versus singular.

snapcraft/plugins/kernel.py
```diff
@@ -0,0 +1,302 @@
...
+            'items': {
+                'type': 'string',
+            },
+            'default': [],
+        }
+
+        schema['properties']['kernel-initrd-compression'] = {
+            'type': 'string',
+            'default': 'gz',
+            'enum': ['gz'],
```
kyrofa: The docs say this can be one of none, gz, bz2, xz. Looks like it's only gz?

kyrofa:
Other than some documentation issues this looks good to me, but I must admit it does go beyond my kernel experience. You may want ogra or ppisati to check this out.

snapcraft/plugins/kernel.py
sergiusens: Ok, updated the docs

snapcraft/plugins/kernel.py
sergiusens: Updated the docs

snapcraft/plugins/kernel.py
sergiusens: Updated the docs

snapcraft/plugins/kernel.py
sergiusens: Updated the docs

sergiusens:
I'm going to ignore those example errors:

```
|-- sergiusens@lindon:~/source/github.com/ubuntu-core/snapcraft (feature/1552168/kernel-plugin) --|
$ python3 -m examples_tests --filter busybox --ip 10.0.3.145 
Tests running...
Cleaning up for part "busybox"
Cleaning up staging area
Cleaning up snapping area
busybox_1.0_amd64.snap                                                                                      100%  468KB 468.0KB/s   00:00    

Ran 14 tests in 53.939s
OK
```

and 

```
|-- sergiusens@lindon:~/source/github.com/ubuntu-core/snapcraft (feature/1552168/kernel-plugin) --|
$ python3 -m examples_tests --filter libpipeline --ip 10.0.3.145 
Tests running...
Cleaning up for part "libpipeline"
Cleaning up for part "pipelinetest"
pipelinetest_1.0_amd64.snap                                                                                 100%   52KB  52.0KB/s   00:00    

Ran 14 tests in 75.111s
OK
```

PR #368, opened 2016-03-08
examples_tests: update the expected result of the libpipeline test

elopio:
The libpipeline examples tests the pipe using ls. Now that the snap is
run in an empty working directory, the result of ls is empty.

LP: #1554637


elopio:
restest this please

sergiusens:
:+1:

PR #369, opened 2016-03-09
Swap plugs and slot functionality for interfaces.

sergiusens:
LP: #1549427


zyga:
Looks good, checked against snappy master (eye-balling)

PR #376, opened 2016-03-17
Fix run()'s shell quoting

lool:
"$@" is the proper way to expand original positional args; $* will perform an extra IFS split.

NB: I realize that I'm not following proper contribution rules and that I have other branches to update, but I didn't want to lose this one  :-)


snappy-m-o:

snappy-m-o:

sergiusens:
ok to test

sergiusens:
Closing in favor of #473

PR #377, opened 2016-03-17
Correct local source pulling

sergiusens:
isdir evaluates to true as it follows symlinks by default and
later fails when doing an rmdir, so we invert the check and we
also specifically check for a link or fail in a controlled way
if not for healthier user output.

This also fixes the case where you could not pull twice with the
"is a symlink error".

LP: #1558446


zyga:
+1, thanks for fixing this

PR #378, opened 2016-03-17
New plugin: kbuild

sergiusens:
LP: #1558623

### DIFF
--- a/examples/busybox/snapcraft.yaml
+++ b/examples/busybox/snapcraft.yaml
@@ -1,0 +1,35 @@
+name: busybox
+version: 1.0
+summary: busybox example to showcase the snapcraft kbuild plugin
+description:
+   this is a snapcraft example package that contains the output of the
+   snapcraft busybox example.
+
+apps:
+  ls:
+    command: ls
+    plugs: [rw-all]
+  touch:
+    command: touch
+    plugs: [rw-all]
+
+plugs:
+  rw-all:
+    interface: old-security
+    security-override:
+      read-paths:
+        - /
+      write-paths:
+        - /
+
+parts:
+  busybox:
+    plugin: kbuild
+    source: https://github.com/asac/busybox.git
+    source-type: git
+    kconfigs:
+     - CONFIG_FEATURE_SUID=n
+     - CONFIG_HTTPD=n
+    snap:
+     -  bin/*
+
--- a/examples_tests/test_busybox.py
+++ b/examples_tests/test_busybox.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright (C) 2015 Canonical Ltd
+# Copyright (C) 2016 Canonical Ltd
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License version 3 as
@@ -14,34 +14,17 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import io
-from unittest import mock
-
-from snapcraft import tests
-from snapcraft.commands import list_plugins
+import examples_tests
 
 
-class ListPluginsCommandTestCase(tests.TestCase):
+class BusyBoxTestCase(examples_tests.ExampleTestCase):
 
-    @mock.patch('sys.stdout', new_callable=io.StringIO)
-    def test_list_plugins(self, mock_stdout):
-        expected_list = '''ant
-autotools
-catkin
-cmake
-copy
-go
-jdk
-kbuild
-make
-maven
-nil
-nodejs
-python2
-python3
-qml
-scons
-tar-content
-'''
-        list_plugins.main()
-        self.assertEqual(mock_stdout.getvalue(), expected_list)
+    example_dir = 'busybox'
+
+    def test_busybox(self):
+        self.build_snap(self.example_dir)
+        self.install_snap(self.example_dir, 'busybox', '1.0')
+        # TODO once PWD issues are sorted make use of busybox.touch
+        # for a more complete test run.
+        self.assert_command_in_snappy_testbed(
+            '/snaps/bin/busybox.ls', '')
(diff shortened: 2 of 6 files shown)


elopio:
retest this please

snapcraft/plugins/kbuild.py
```diff
@@ -0,0 +1,179 @@
...
+"""The kbuild plugin is used for building kbuild based projects as snapcraft
+parts.
+
+This plugin is based on the snapcraft.BasePlugin and supports the properties
+provided by that plus the following kbuild specific options with semantics as
+explained above:
+
+    - kdefconfig:
+      (list of kdefconfigs)
+      defconfig target to use to see base configuration. default: "defconfig"
```
elopio: this is not super clear. "to use to see" sounds bad, but I'm not sure how to fix it because I don't understand what it means "to see". Do you have any possible rewrite in mind?

snapcraft/plugins/kbuild.py
```diff
@@ -0,0 +1,179 @@
...
+    def build(self):
+        super().build()
+
+        config_path = os.path.join(self.builddir, '.config')
+
+        self.do_base_config(config_path)
+        self.do_patch_config(config_path)
+        self.do_remake_config()
+        self.do_build()
+        self.do_install()
```
elopio: any reason for not making all these do_* methods internal?

sergiusens: the kernel plugin uses them

snapcraft/plugins/kbuild.py
sergiusens: it isn't clear, I'll rephrase it

examples/busybox/snapcraft.yaml
```diff
@@ -0,0 +1,34 @@
...
+   this is a snapcraft example package that contains the output of the
+   snapcraft busybox example.
+
+apps:
+  ls:
+    command: ls
+    plugs: [old-security]
+  touch:
+    command: touch
+    plugs: [old-security]
```
kyrofa: Just a nit, but can we name this something representative of what it is? Maybe `read-and-write-all` or something?

examples/busybox/snapcraft.yaml
```diff
@@ -0,0 +1,34 @@
...
+version: 1.0
+summary: busybox example to showcase the snapcraft kbuild plugin
+description:
+   this is a snapcraft example package that contains the output of the
+   snapcraft busybox example.
+
+apps:
+  ls:
+    command: ls
+    plugs: [old-security]
```
kyrofa: Just a nit, but can we name this something more representative of what it is? `read-and-write-all` maybe? I'm worried this will get confused with the `old-security` interface.

examples/busybox/snapcraft.yaml
```diff
@@ -0,0 +1,34 @@
...
+    security-override:
+      read-paths:
+        - /
+      write-paths:
+        - /
+
+parts:
+  busybox:
+    plugin: kbuild
+    source: https://github.com/asac/busybox.git
```
kyrofa: Would it be a better idea to pull this repo into the ubuntu-core group so it doesn't belong to any one person (and be accidentally deleted)?

examples/busybox/snapcraft.yaml
sergiusens: I was avoiding writing `interface: old-security` in case that changed again :smile: 

I don't mind otherwise :wink:

kyrofa: Hahaha, fair point!

examples/busybox/snapcraft.yaml
sergiusens: That I'll leave for @didrocks for when he migrates to demos. This is not the only project having this issue.

snapcraft/plugins/kbuild.py
```diff
@@ -0,0 +1,169 @@
...
+
+    - kconfigs
+      (list of strings)
+      explicit list of configs to force; this will override the configs that
+      were set as base through kdefconfig and kconfigfile and dependent configs
+      will be fixed using the defaults encoded in the kbuild config
+      definitions.  If you don't want default for one or more implicit configs
+      coming out of these, just add them to this list as well.
+
+    - build-parameters:
```
kyrofa: This isn't actually an option.

kyrofa:
Few tiny comments-- looks great! :+1:

snapcraft/plugins/kbuild.py
sergiusens: Right I removed it as I only added it for cross compile in the kernel :-)

examples_tests/test_busybox.py
```diff
@@ -0,0 +1,30 @@
...
+
+class BusyBoxTestCase(examples_tests.ExampleTestCase):
+
+    example_dir = 'busybox'
+
+    def test_busybox(self):
+        self.build_snap(self.example_dir)
+        self.install_snap(self.example_dir, 'busybox', '1.0')
+        self.assert_command_in_snappy_testbed(
+            'sudo /snaps/bin/busybox.touch touched', '')
```
kyrofa: Why not touch `$SNAP_USER_DATA/touched` here?

elopio:
I like this. I have no idea about kbuild, and I can follow the code and the tests nicely, even the weirdness like preppend + append. So :+1: from me. The busybox tests gives me confidence, but as I said, I haven't seen kbuild before, so we need a review from somebody who knows about that. If @kyrofa knows plenty about this stuff, we are good to go.

kyrofa:
I've built my share of kernels, but there were a few things in here that were a little odd to me-- the list of `kdefconfigs` for instance (I only ever use one... not sure what using multiple is supposed to do). The prepend+append is a little weird, but I figure that came from some testing, so no big deal. We might want a third review, just in case.

sergiusens:
@kyrofa @elopio the prepend and append is in a comment in there :smile: and it is due to the fact that some versions of `make oldconfig` take the first entry of a config as valid and ignore the others while other versions take the latter entries as overrides.

multiple kdefconfigs was added to support ricmm when building Ubuntu kernels.

kyrofa:
> @kyrofa @elopio the prepend and append is in a comment in there :smile: and it is due to the fact that some versions of make oldconfig take the first entry of a config as valid and ignore the others while other versions take the latter entries as overrides.

Right, the comment is why I figured you tested that and I didn't need to question it. I don't see a downsize anyway.

sergiusens:
@kyrofa ubuntu has many defconfigs that get appended, turns out this make target appends all of them together.

PR #379, opened 2016-03-18
Support downloading snaps

sergiusens:
This is implemented to support kernel building

LP: #1558970


kyrofa:
:+1: looks good to me!

ogra1:
the initrd.img files comes originally from a deb in the archive. is it really necessary to pull the whole rootfs snap just to extract a 2MB file from it instead of just pulling the .deb package from the archive ?

sergiusens:
Just for the record as I mentioned this to @ogra1 on irc; we plan to remove this as soon as the generic initrd work is complete on the snappy side and it learns to pick it up

PR #382, opened 2016-03-18
Support cleaning strip step.

kyrofa:
This PR makes progress on LP: [#1537786](https://bugs.launchpad.net/snapcraft/+bug/1537786) toward implementing per-step clean rules by adding one for the `strip` step. The others are currently placeholders and will be implemented in future PRs.

### DIFF
--- a/snapcraft/pluginhandler.py
+++ b/snapcraft/pluginhandler.py
@@ -128,12 +128,14 @@ class PluginHandler:
     def should_step_run(self, step, force=False):
         return force or self.is_dirty(step)
 
-    def mark_done(self, step):
+    def mark_done(self, step, state=None):
+        if not state:
+            state = {}
+
         index = common.COMMAND_ORDER.index(step)
 
-        # Currently only creating the step file. TODO: This will be a YAML file
-        # holding step-specific state information.
-        open(self._step_state_file(step), 'w').close()
+        with open(self._step_state_file(step), 'w') as f:
+            f.write(yaml.dump(state))
 
         # We know we've only just completed this step, so make sure any later
         # steps don't have a saved state.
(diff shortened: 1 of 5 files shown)


snapcraft/pluginhandler.py
```diff
@@ -128,12 +128,11 @@ def is_dirty(self, step):
     def should_step_run(self, step, force=False):
         return force or self.is_dirty(step)
 
-    def mark_done(self, step):
+    def mark_done(self, step, state={}):
```
sergiusens: `state=None` and then inside check 

```python
if not state:
    state = {}
```

snapcraft/pluginhandler.py
```diff
@@ -226,15 +241,46 @@ def strip(self, force=False):
...
+
+        self.mark_done('strip', state)
+
+    def clean_strip(self):
+        logger.info('Cleaning up step "strip" for part "{}"'.format(self.name))
+
+        with open(self._step_state_file('strip'), 'r') as f:
+            state = yaml.load(f.read())
+
+        _clean_migrated_files(state['snap-files'], state['snap-directories'],
```
sergiusens: before passing this in, you need to load all parts and check if they migrated the same file to remove it from the wipe list

sergiusens:
Looks good, just comment about dict and filesets

snapcraft/pluginhandler.py
kyrofa: Ah, my python knowledge just increased! Thanks for the link :smiley: .

snapcraft/pluginhandler.py
kyrofa: Oh right, when writing this I assumed that would imply they were dependent parts, but I forgot that multiple parts can supply the same file as long as the contents don't change.

kyrofa:
This PR needs to land to an intermediate branch. Closing.

PR #391, opened 2016-03-22
Snaps of type os should not squash with -all-root

sergiusens:
When creating os snaps we want to preserve user permissions.

LP: #1557513


kyrofa:
Looks good! :+1:

PR #388, opened 2016-03-22
busybox example: update the expected results

elopio:
The next ubuntu-core release changes the directory where the command
is run. This lets us use busybox.touch and changes the expected
results of busybox.ls

LP: #1560255

PR #392, opened 2016-03-22
busybox example: update the expected results

sergiusens:
The next ubuntu-core release changes the directory where the command
is run. This lets us use busybox.touch and changes the expected
results of busybox.ls

LP: #1560255


sergiusens:
Opening #388 again instead of this

PR #393, opened 2016-03-22
Support architecture all packages when snapping

sergiusens:
When snapping a directory we want to support architecture all
by omission of mentioning the architecture.

LP: #1560481


kyrofa:
Looks good! :+1:

PR #394, opened 2016-03-22
Support cleaning strip step.

kyrofa:
This PR makes progress on LP: [#1537786](https://bugs.launchpad.net/snapcraft/+bug/1537786) by adding the ability to clean only a part's stripped files. This takes into account the possibility for parts to share files, and such files are not removed.

Of specific note: command wrappers and `snap/meta/` are not cleaned unless all parts have been cleaned. Ideally we'd be able to remove wrappers associated with a given part, but there isn't an elegant way to determine that association. Fortunately they're re-generated when stripped again, but still not ideal.


integration_tests/test_clean_strip_step.py
```diff
@@ -0,0 +1,55 @@
...
+
+        bindir = os.path.join(project_dir, 'snap', 'bin')
+        self.assertThat(os.path.join(bindir, 'file1'), FileExists())
+        self.assertThat(os.path.join(bindir, 'file2'), FileExists())
+
+        self.run_snapcraft(['clean', 'part1', '--step=strip'], project_dir)
+        self.assertThat(os.path.join(bindir, 'file1'), Not(FileExists()))
+        self.assertThat(os.path.join(bindir, 'file2'), FileExists())
+        self.assertThat(os.path.join(project_dir, 'stage'), DirExists())
+        self.assertThat(os.path.join(project_dir, 'parts'), DirExists())
```
elopio: /me bows.

snapcraft/pluginhandler.py
```diff
@@ -301,7 +359,21 @@ def _new_clean(self, project_staged_state=None,
...
+                              step)
+        except MissingState:
+            # If one of the step cleaning rules is missing state, it must be
+            # running on the output of an old Snapcraft. In that case, if we
+            # were specifically asked to clean that step we need to fail.
+            # Otherwise, just clean like the old Snapcraft did, and blow away
+            # the entire part directory.
+            if step:
+                raise
+
```
elopio: no need for a warning or debug in here? if !step, it could mean that we have the output of an old snapcraft, or that there is a bug we missed. logging something could be useful.

snapcraft/tests/test_pluginhandler.py
```diff
@@ -340,9 +341,93 @@ def test_stage_state(self):
...
+
+        self.handler.mark_done('build')
+        self.handler.stage()
         self.handler.strip()
 
         self.assertEqual('strip', self.handler.last_step())
+        with open(self.handler._step_state_file('strip'), 'r') as f:
+            state = yaml.load(f)
+
+        self.assertTrue(state, 'Expected strip to save state YAML')
```
elopio: how does this work? state is not True, because it is a pluginhandler.StripState.
seems like crazy python, where everything not False is considered to be True, which makes it not so clear in this case. What would return yaml.load if the state file was not saved? I think it would be a little clearer if we assert state is different from that.

elopio:
lgtm. Thanks for the integration tests. I think a good test to add in there is to make sure that we can clean the strip step, and then strip or snap again.

snapcraft/pluginhandler.py
kyrofa: Well, running on old Snapcraft output will probably be a common case. I can't think of anything to print that wouldn't make users suspicious that something is wrong when it's not.

snapcraft/tests/test_pluginhandler.py
kyrofa: Ah, this is a carry-over from before I had state classes, so this line is redundant anyway.

kyrofa:
> Thanks for the integration tests. I think a good test to add in there is to make sure that we can clean the strip step, and then strip or snap again.

Great idea-- done!

snapcraft/commands/clean.py
```diff
@@ -53,22 +53,47 @@ def _reverse_dependency_tree(config, part_name):
...
+
+def _remove_directory_if_empty(directory):
+    if os.path.isdir(directory) and not os.listdir(directory):
+        os.rmdir(directory)
+
+
+def _cleanup_common_directories(config):
+    _remove_directory_if_empty(common.get_partsdir())
+    _remove_directory_if_empty(common.get_stagedir())
+    _remove_directory_if_empty(common.get_snapdir())
```
sergiusens: +1

snapcraft/commands/clean.py
```diff
@@ -53,22 +53,47 @@ def _reverse_dependency_tree(config, part_name):
...
+    max_index = -1
+    for part in config.all_parts:
+        step = part.last_step()
+        if step:
+            index = common.COMMAND_ORDER.index(step)
+            if index > max_index:
+                max_index = index
+
+    # If no parts have been staged, remove staging area.
+    should_remove_stagedir = max_index < common.COMMAND_ORDER.index('stage')
```
sergiusens: clever

sergiusens:
:+1:

PR #398, opened 2016-03-23
Support cleaning build step.

kyrofa:
This PR makes progress on LP: [#1537786](https://bugs.launchpad.net/snapcraft/+bug/1537786) by adding the ability to clean only a part's built files. This involves adding a new clean_build() method to the plugin API as there are a few plugins that have more involved build steps than the base implementation's. This commit also reimplements `clean_build()` for the Go plugin.


snapcraft/pluginhandler.py
```diff
@@ -223,11 +223,13 @@ def get_state(self, step):
     def _step_state_file(self, step):
         return os.path.join(self.statedir, step)
 
-    def _setup_stage_packages(self):
-        ubuntu = repo.Ubuntu(
+    def _fetch_stage_packages(self):
+        self.ubuntu = repo.Ubuntu(
```
sergiusens: This should be done in `__init__` with a conditional

```python
if self.stage_packages:
    self.ubuntu = repo.Ubuntu(...)
```

or maybe regardless as some plugins might do crazy stuff with `stage_packages` after `__init__` :wink:

snapcraft/pluginhandler.py
```diff
@@ -255,13 +258,22 @@ def build(self, force=False):
             return
         self.makedirs()
         self.notify_stage('Building')
+        if self.code.stage_packages:
+            # These were already fetched and unpacked in pull(), but unpack
+            # again here in case the build was cleaned.
+            self._unpack_stage_packages()
```
sergiusens: Right, and this is why, `self.ubuntu` won't exist here

sergiusens:
Looks good, added some comments and another one here; you probably want to `clean_build` override the `nodejs` plugin as well.

snapcraft/pluginhandler.py
kyrofa: Fixed.

snapcraft/pluginhandler.py
kyrofa: Good catch, not sure what I was thinking there. I ended up extracting it into a property instead that initializes itself on first use (putting it in `__init__` would make it fetch sources every time a step was run, which wouldn't be ideal).

kyrofa:
> you probably want to `clean_build` override the nodejs plugin as well.

I looked at that plugin, but it doesn't seem to use anything other than the `builddir` and `installdir` in `build()`. It does in `pull()`, but that's for the next PR.

integration_tests/test_clean_build_step.py
```diff
@@ -0,0 +1,166 @@
...
+            partdir = os.path.join(self.partsdir, part)
+            self.parts[part] = {
+                'partdir': partdir,
+                'sourcedir': os.path.join(partdir, 'src'),
+                'builddir': os.path.join(partdir, 'build'),
+                'installdir': os.path.join(partdir, 'install'),
+                'bindir': os.path.join(partdir, 'install', 'bin'),
+            }
+
+    def verify_files_exist(self):
```
elopio: s/verify/assert

integration_tests/test_clean_build_step.py
```diff
@@ -0,0 +1,166 @@
...
+        self.verify_files_exist()
+
+        self.run_snapcraft(['clean', '--step=build'], self.project_dir)
+
+        for part_name, part in self.parts.items():
+            self.assertThat(part['builddir'], Not(DirExists()))
+            self.assertThat(part['installdir'], Not(DirExists()))
+            self.assertThat(part['sourcedir'], DirExists())
+
+        # Now try to build again
```
elopio: cool.

snapcraft/__init__.py
```diff
@@ -222,6 +223,20 @@ def build(self):
             ignore=lambda d, s: common.SNAPCRAFT_FILES
             if d is self.sourcedir else [])
 
+    def clean_build(self):
```
elopio: we had the concept on clean build (building in lxc) before this, which makes it confusing.
what about clean_build_step?

elopio:
:+1: 
I left a nitpicking comment. And I left a question about clean build (build in an lxc) and clean build (clean the build step).

snapcraft/__init__.py
kyrofa: Yeah it's a bit unfortunate. But the function calls are `pull()`, `build()`, `stage()` and `strip()`, no *_step. So I feel like I need to match up the clean_* functions the same way.

integration_tests/test_clean_build_step.py
kyrofa: Fixed.

kyrofa:
retest this please.

kyrofa:
retest this please.

sergiusens:
:+1:

PR #399, opened 2016-03-23
Fixed assets go into a setup dir

sergiusens:
Instead of declaring fixed assets we put them into a `setup` dir.
Implementation for gui (including icon) and license provided.

LP: 1560969

### DIFF
--- a/snapcraft/meta.py
+++ b/snapcraft/meta.py
@@ -100,13 +100,39 @@ def _setup_assets(meta_dir, config_data):
         _setup_config_hook(hooks_dir, config_data['config'])
 
     if 'license' in config_data:
+        logger.warning("DEPRECATED: 'license' defined in snapcraft.yaml")
         license_path = os.path.join(meta_dir, 'license.txt')
-        shutil.copyfile(config_data['license'], license_path)
+        os.link(config_data['license'], license_path)
 
     if 'icon' in config_data:
+        logger.warning("DEPRECATED: 'icon' defined in snapcraft.yaml")
         icon_ext = config_data['icon'].split(os.path.extsep)[1]
-        icon_path = os.path.join(meta_dir, 'icon.{}'.format(icon_ext))
-        shutil.copyfile(config_data['icon'], icon_path)
+        icon_dir = os.path.join(meta_dir, 'gui')
+        icon_path = os.path.join(icon_dir, 'icon.{}'.format(icon_ext))
+        os.mkdir(icon_dir)
+        os.link(config_data['icon'], icon_path)
+
+    _setup_from_setup(meta_dir)
+
+
+def _setup_from_setup(meta_dir):
+    setup_dir = 'setup'
+    if not os.path.exists(setup_dir):
+        return
+
+    gui_src = os.path.join(setup_dir, 'gui')
+    gui_dst = os.path.join(meta_dir, 'gui')
+    if os.path.exists(gui_dst) and os.path.exists(gui_src):
+        shutil.rmtree(gui_dst)
+    if os.path.exists(gui_src):
+        shutil.copytree(gui_src, gui_dst)
+
+    license_src = os.path.join(setup_dir, 'license.txt')
+    license_dst = os.path.join(meta_dir, 'license.txt')
+    if os.path.exists(license_dst) and os.path.exists(license_src):
+        os.unlink(license_dst)
+    if os.path.exists(license_src):
+        os.link(license_src, license_dst)
 
 
 def _setup_config_hook(hooks_dir, config):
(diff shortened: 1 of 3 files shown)


sergiusens:
retest this please

elopio:
retest this please

snapcraft/meta.py
```diff
@@ -100,13 +100,39 @@ def _setup_assets(meta_dir, config_data):
         _setup_config_hook(hooks_dir, config_data['config'])
 
     if 'license' in config_data:
+        logger.warning("DEPRECATED: 'license' defined in snapcraft.yaml")
         license_path = os.path.join(meta_dir, 'license.txt')
-        shutil.copyfile(config_data['license'], license_path)
+        os.link(config_data['license'], license_path)
 
     if 'icon' in config_data:
+        logger.warning("DEPRECATED: 'icon' defined in snapcraft.yaml")
```
kyrofa: Ah, I didn't realize this was actually going away. So this will become store-only?

kyrofa:
:+1: from me, though we may want a snappy review.

elopio:
the branch is ok, but it's missing some docs explaining the use of the new directory.
In the PITA side of things, I don't think "setup" is a good name, because it is not setting up anything. Something like "assets" would make more sense in my opinion. Of course, I don't want to start a new discussion about terminology if this term was already discussed and comes from one of the specifications.

kyrofa:
Good point @elopio, "assets" does seem more appropriate.

sergiusens:
@elopio @kyrofa this term comes from specifications. There's also an email thread where this was discussed.

kyrofa:
@sergiusens alright, that works.

snapcraft/meta.py
```diff
@@ -100,13 +100,39 @@ def _setup_assets(meta_dir, config_data):
...
+    gui_src = os.path.join(setup_dir, 'gui')
+    gui_dst = os.path.join(meta_dir, 'gui')
+    if os.path.exists(gui_dst) and os.path.exists(gui_src):
+        shutil.rmtree(gui_dst)
+    if os.path.exists(gui_src):
+        shutil.copytree(gui_src, gui_dst)
+
+    license_src = os.path.join(setup_dir, 'license.txt')
+    license_dst = os.path.join(meta_dir, 'license.txt')
+    if os.path.exists(license_dst) and os.path.exists(license_src):
```
zyga: This could be moved under the subsequent test ``if os.path.exists(license_src)`` and simplified to just ``if os.path.exists(license_dst)``, that is easier to read for me.

snapcraft/meta.py
```diff
@@ -100,13 +100,39 @@ def _setup_assets(meta_dir, config_data):
...
+
+
+def _setup_from_setup(meta_dir):
+    setup_dir = 'setup'
+    if not os.path.exists(setup_dir):
+        return
+
+    gui_src = os.path.join(setup_dir, 'gui')
+    gui_dst = os.path.join(meta_dir, 'gui')
+    if os.path.exists(gui_dst) and os.path.exists(gui_src):
```
zyga: Same as below for license

snapcraft/meta.py
```diff
@@ -100,13 +100,39 @@ def _setup_assets(meta_dir, config_data):
...
-        shutil.copyfile(config_data['icon'], icon_path)
+        icon_dir = os.path.join(meta_dir, 'gui')
+        icon_path = os.path.join(icon_dir, 'icon.{}'.format(icon_ext))
+        os.mkdir(icon_dir)
+        os.link(config_data['icon'], icon_path)
+
+    _setup_from_setup(meta_dir)
+
+
+def _setup_from_setup(meta_dir):
```
zyga: This looks okay but I'd like @mvo5 to ack it as well. I'm not 100% sure of the design of the setup directory.

zyga:
+1 but please get an ack from @mvo5 as I'm not 100% sure about setup/

snapcraft/meta.py
sergiusens: right, thanks for the suggestion, it feels like personal taste; I don't like nesting ifs; but your solution would be more performant.

sergiusens:
@elopio @kyrofa @zyga https://lists.ubuntu.com/archives/snappy-devel/2016-March/001598.html

@zyga I don't think @mvo5 was part of the origin directory; I just want to make sure someone from the snappy team acks I'm putting stuff in the right target directory :wink:

snapcraft/meta.py
zyga: Don't feel blocked on this, I just found it easier to read this way.

sergiusens:
I've already spotted many places where docs are lacking or plain 15.04 incorrect; I'll work on a general branch for that.

PR #401, opened 2016-03-24
Support per-step cleaning.

kyrofa:
This PR resolves LP: [#1537786](https://bugs.launchpad.net/snapcraft/+bug/1537786) by adding more fine-grained clean functionality, allowing one to clean each step per part rather than all steps.


sergiusens:
I fetched and did some local testing, I :+1:

PR #409, opened 2016-03-29
Explain the new way to provide snap icons.

sergiusens:
LP: #1561327


docs/metadata.md
```diff
@@ -105,6 +105,32 @@ To find out more about config command, have a look at the [config section]
...
+Some metadata is provided in the form of conventions, such as license files,
+icons and desktop files among others. For these fixed files to make it into
+your final snap they need to be in a `setup` directory at the same level of
+your `snapcraft.yaml`.
+
+### Snap icon
+
+To provide an icon for your snap, while for command line applications might
+not seem as important it is rather important for discoverability from
+management interfaces such as store fronts like webdm.
```
kyrofa: This could flow a bit better. Perhaps something like "Providing an icon for your snap is important, even for command-line applications, if for nothing else than discoverability from management[...]"

snapcraft/meta.py
```diff
@@ -105,7 +105,11 @@ def _setup_assets(meta_dir, config_data):
         os.link(config_data['license'], license_path)
 
     if 'icon' in config_data:
-        logger.warning("DEPRECATED: 'icon' defined in snapcraft.yaml")
+        # TODO: use developer.ubuntu.com once it has updated documentation.
```
kyrofa: Is there an issue for this so we can track it?

sergiusens: Not that I know of.

docs/metadata.md
sergiusens: Fixed

elopio:
:+1:

PR #410, opened 2016-03-29
docs/plugins.md: Update for new plugin API.

kyrofa:
This PR fixes LP: [#1561524](https://bugs.launchpad.net/snapcraft/+bug/1561524) by updating `docs/plugins.md` for the new plugin API. Which isn't really new anymore, but the doc never got updated to reflect.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -3,48 +3,50 @@
 Is Snapcraft missing support for your preferred build-system? Here's how you
 would add it!
 
-Snapcraft can be extended by adding new part plugins. Let's add support for
-running a custom build-tool named "Crafty". The goal of this plugin is to
-pass a configurable target to the `crafty sometarget` command inside the
-`src/crafty` source directory.
+Snapcraft can be extended by adding new part plugins, which are written in
+Python. Let's add support for running a custom build-tool named "Crafty". The
+goal of this plugin is to pass a configurable target to the `crafty sometarget`
+command inside the `src/crafty` source directory.
 
-Relative to the directory of `snapcraft.yaml`, create 
-`parts/plugins/x-crafty.yaml` with:
-
-```yaml
-options:
-    source:
-        required: true
-    source-type:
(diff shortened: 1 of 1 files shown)


docs/plugins.md
```diff
@@ -3,48 +3,49 @@
...
-    source-type:
-    source-tag:
-    source-branch:
-    crafty-target:
-        required: true
-```
-
-and `parts/plugins/x_crafty.py` with:
+Relative to the directory of `snapcraft.yaml`, create
+`parts/plugins/crafty.py` with:
```
sergiusens: it should still be `x-crafty`

docs/plugins.md
```diff
@@ -3,48 +3,49 @@
...
 
 ```python
 import snapcraft
 
 
-class XCraftyPlugin(snapcraft.BasePlugin):
+class CraftyPlugin(snapcraft.BasePlugin):
```
sergiusens: this without the x is fine.

docs/plugins.md
kyrofa: That just overrides any plugin named "crafty"' shipped in Snapcraft, right? I figured it would be best to simply discuss how plugins are written and then mention (lines 38-39) that the x_ would make it take precedence over in-Snapcraft plugins.

By the way, is there any reason not to simply take local plugins over Snapcraft plugins and get rid of the whole x_ thing?

sergiusens: `x_` because if someone gets an official plugin in place with that same name it would break your project.

kyrofa: Done. Though it seems to actually be `x-`, not `x_`.

kyrofa:
retest this please

kyrofa:
retest this please

sergiusens:
is this stuck?

kyrofa:
No, I started it again a few minutes ago as it died from the quota limit.

PR #413, opened 2016-03-30
add the option to upload an existing snap

elopio:
LP: #1563760

### DIFF
--- a/snapcraft/commands/upload.py
+++ b/snapcraft/commands/upload.py
@@ -20,7 +20,7 @@ snapcraft upload
 Upload snap package to the Ubuntu Store.
 
 Usage:
-  upload [options]
+  upload [options] SNAP_FILE
 
 Options:
   -h --help             show this help message and exit.
@@ -29,12 +29,12 @@ Options:
 
 import logging
 import os
+import subprocess
+import tempfile
 
+import yaml
 from docopt import docopt
 
-import snapcraft.yaml
-from snapcraft.commands import snap
-from snapcraft.common import format_snap_name
 from snapcraft.config import load_config
 from snapcraft.storeapi import upload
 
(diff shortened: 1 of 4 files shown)


snapcraft/commands/upload.py
```diff
@@ -42,20 +45,42 @@
...
+        snap_filename = args['SNAP_FILE']
+        if not os.path.exists(snap_filename):
+            raise FileNotFoundError(snap_filename)
+        else:
+            snap_name = _get_name_from_snap_file(snap_filename)
+            logger.info('Uploading existing {}.'.format(snap_filename))
+    else:
+        # make sure the full lifecycle is executed
+        yaml_config = snapcraft.yaml.load_config()
+        snap_filename = format_snap_name(yaml_config.data)
```
sergiusens: hmm, since you did the unsquashfs work I'd like to ask of you to check if the names match (the snap and snapcraft.yaml) :-)

snapcraft/tests/test_commands_upload.py
```diff
@@ -135,3 +135,41 @@ def test_just_upload_if_snap_file_exists(self):
...
+
+        temp_dir = upload.tempfile.TemporaryDirectory
+
+        def create_patch(method, return_value):
+            patcher = mock.patch.object(temp_dir, method)
+            mock_method = patcher.start()
+            mock_method.return_value = return_value
+            self.addCleanup(patcher.stop)
+
+        create_patch('__enter__', os.getcwd())
```
fgimenez: Adding `create_patch('_cleanup', None)` before the first `create_patch` call prevents the `ResourceWarning: Implicitly cleaning up <TemporaryDirectory...` warning message

elopio: That's great, thanks Federico!

snapcraft/commands/upload.py
```diff
@@ -42,20 +45,42 @@
...
+                'Snap {} not found. Running snap step to create it.'.format(
+                    snap_filename))
+            snap.main(argv=argv)
 
-    if not os.path.exists(snap_filename):
-        logger.info(
-            'Snap {} not found. Running snap step to create it.'.format(
-                snap_filename))
-        snap.main(argv=argv)
+        snap_name = yaml_config.data['name']
```
kyrofa: Why do this at all? Why not always use the unpacked snap's information rather than the in-tree YAML's, since it's possible to be out of date?

elopio: I think that's what I'm fixing with Sergio's comment. I'm currently failing if the in-tree yaml is name different than the snap.yaml. Let me know if you would prefer otherwise. A new rev is coming...

kyrofa:
With the addition of this capability, I actually see no added benefit from the `snapcraft upload` command (i.e. with no SNAP_FILE). I say scrap the in-tree YAML parsing altogether, make SNAP_FILE a required parameter, and remove upload from the lifecycle. More of a do-one-thing-and-do-it-well type of thing I guess. Just my two cents.

elopio:
I pushed @sergiusens change, with a test.
@kyrofa it's the same to me. I can just get rid of the old code to link upload with snap. You two agree on what would you prefer and I'll do it.

elopio:
retest this please

elopio:
@kyrofa Sergio agreed. I will delete the old code and leave just the new.

kyrofa:
@elopio awesome. Note that'll touch the upload unit tests as well.

PR #417, opened 2016-03-31
kbuild: always run defconf with -j1

sergiusens:
LP: #1564191


snapcraft/plugins/kbuild.py
```diff
@@ -110,8 +110,10 @@ def do_base_config(self, config_path):
         # if kconfigfile is provided use that
         # otherwise use defconfig to seed the base config
         if self.options.kconfigfile is None:
-            # first run defconfig for setting baseline
-            self.run(self.make_cmd + self.options.kdefconfig)
+            # we need to run this with -j1, unit tests are a good defense here.
+            make_cmd = self.make_cmd.copy()
+            make_cmd[1] = '-j1'
+            self.run(make_cmd + self.options.kdefconfig)
```
kyrofa: This seems a bit fragile. Why not refactor `self.make_cmd` into a function that accepts the number of build jobs as a parameter?

sergiusens: I'd argue it's ugly, not fragile. But not against a make_function. This will have a power play on cross compiling though.

kyrofa:
Alright I'm good with this :+1:

kyrofa:
retest this please

sergiusens:
Examples passed, adt failed due to scalingstack (known to pass in the past though).

PR #416, opened 2016-03-31
Copy plugin: Add support for sources.

kyrofa:
Currently the copy plugin can only be used for local assets and pulling down build- or stage-packages. This PR resolves LP: [#1559154](https://bugs.launchpad.net/snapcraft/+bug/1559154) by adding the ability to also use it for remote sources with no build system.

PR #420, opened 2016-04-02
Extends copy plugin to support  source and destination

didrocks:
The copy plugin now supports "source" and can either copy the whole content
(no "files" or "destination" parameter), or filter through one of those
keywords.
It also have special tar file support, putting in line with other plugins.

Add also help for copy plugin, add a get_source_type() public handler and support autodetection for uncompressed "tar".

Finally, add unit and integration tests, with test cases for those.


didrocks:
(Yeah, I know there are 3 commits and not one, I can push the first 2 in another PR if you wish as it's about the source support itself)

didrocks:
The example tests (executed in autopkgtests) seems to have been quite flacky for a while (previous runs didn't pass). I've retested this webcam_ui example in the branch (it shouldn't be impacted by those changes anyway), and it passes successfully.

sergiusens:
Autopackage tests seem to be in a bad state. Thanks for working on this! Here's the thing, @kyrofa worked on the same thing in #416. I'll let you guys figure it out :wink:

kyrofa:
@didrocks and I have discussed. I will be taking a few things from this and integrating into #416, so I'll close this to avoid confusion.

sergiusens:
El 04/04/16 a las 10:21, Kyle Fazzari escribió:
> @didrocks <https://github.com/didrocks> and I have discussed. I will be
> taking a few things from this and integrating into #416
> <https://github.com/ubuntu-core/snapcraft/pull/416>, so I'll close this
> to avoid confusion.

Great, because I saw good stuff in his PR :smile:

PR #421, opened 2016-04-04
Suggest a full clean if step clean is missing state.

kyrofa:
Currently if attempting to run a per-step clean on a tree made dirty by a version of Snapcraft that didn't support per-step cleaning, Snapcraft suggests trying to run that step again. This isn't possible since the step has already been run, so it actually requires a full clean. This PR resolves LP: [#1561328](https://bugs.launchpad.net/snapcraft/+bug/1561328) by updating the suggestions to reflect this fact.


elopio:
Thank you! :+1:

PR #427, opened 2016-04-05
VCS tools are test dependencies…

vilagithub:
...they should be installed to run the tests or the tests should be skipped.

This allows tests to be run concurrently (apt-get can't be run concurrently
;) and brings the time to run both unit and integration tests from 5'47 to 1'43 with:

  PYTHONPATH=`pwd` ols-run-tests -m integration_tests -m snapcraft -c 8

which is nice ;)

No tests were hurt during the production of this patch ;)

### DIFF
--- a/debian/tests/control
+++ b/debian/tests/control
@@ -5,6 +5,9 @@ Depends: @builddeps@
 Tests: integrationtests examplestests
 Restrictions: allow-stderr, isolation-container, rw-build-tree
 Depends: @,
+         bzr,
+         git,
+         mercurial,
          lxd,
          python3-pep8,
          pyflakes,
--- a/integration_tests/test_bzr_source.py
+++ b/integration_tests/test_bzr_source.py
@@ -16,14 +16,19 @@
 
 import os
 import subprocess
+import shutil
 
 import integration_tests
 
 
 class BzrSourceTestCase(integration_tests.TestCase):
 
+    def setUp(self):
+        super().setUp()
+        if shutil.which('bzr') is None:
+            self.skipTest('bzr is not installed')
+
     def _init_and_config_bzr(self):
-        subprocess.check_call(['sudo', 'apt-get', 'install', '-y', 'bzr'])
         subprocess.check_call(
             ['bzr', 'init', '.'],
             stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
--- a/integration_tests/test_git_source.py
+++ b/integration_tests/test_git_source.py
@@ -16,14 +16,19 @@
 
 import os
 import subprocess
+import shutil
 
 import integration_tests
 
 
 class GitSourceTestCase(integration_tests.TestCase):
 
+    def setUp(self):
+        super().setUp()
+        if shutil.which('git') is None:
+            self.skipTest('git is not installed')
+
     def _init_and_config_git(self):
-        subprocess.check_call(['sudo', 'apt-get', 'install', '-y', 'git'])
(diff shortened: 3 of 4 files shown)


snappy-m-o:

snappy-m-o:

elopio:
OK to test.

sergiusens:
this is missing a bug number

elopio:
lgtm. Maybe we can add in runtests.sh a check for all the requirements and fail if they are missng so we can also run the tests with that script, not just with adt-run.

elopio:
also note that some plugins will execute apt-get to install their build packages. So this is good, but doesn't remove all the executions of apt-get in the code.
A nice improvement would be to have retries on snapcraft/repo.py in case somebody else has the apt lock.

vilagithub:
> this is missing a bug number

Right, I wasn't sure how to phrase it nor if it was worth fixing so I thought about gathering feedback with that PR first ;)

vilagithub:
> note that some plugins will execute apt-get to install their build packages. So this is good, but doesn't remove all the executions of apt-get in the code.

Right, this was focused on tests only and only on tests that broke isolation and block concurrency.

> A nice improvement would be to have retries on snapcraft/repo.py in case somebody else has the apt lock.

I'm not sure I follow. As a user I don't expect to get meaningful results by running apt-get install concurrently (even through different commands).

For tests though, I would argue that if 'sudo apt-get install' needs to be used in a test this should happen in some vm where passwordless sudo access is provided.

Developers running tests are unlikely to provide passwordless sudo access and some would (rightly) freak out if running tests prompt for the root password ;)

*I* do my snapcraft work inside a container where passwordless sudo access is provided but I don't think I'm in the majority there...

vilagithub:
Found https://bugs.launchpad.net/snapcraft/+bug/1540523 which looks like a good match.
@sergiusens, do you agree ?

sergiusens:
El 06/04/16 a las 10:59, vilagithub escribió:
> Found https://bugs.launchpad.net/snapcraft/+bug/1540523 which looks like
> a good match.
> @sergiusens <https://github.com/sergiusens>, do you agree ?

Could be, that bug is much broader though (implementing requirements.txt and requirements-test.txt for dependencies as well).

vilagithub:
> Could be, that bug is much broader though (implementing requirements.txt and requirements-test.txt for dependencies as well).

Ha right, yeah, will file a more focused one then.

vilagithub:
https://bugs.launchpad.net/snapcraft/+bug/1566882

vilagithub:
Neither http://162.213.35.179:8080/job/github-snapcraft-autopkgtest-cloud/402/console nor http://162.213.35.179:8080/job/github-snapcraft-examples-tests-cloud/492/testReport/junit/test_webcam_webui/WebcamWebUITestCase/test_webcam_webui/ makes much sense to my uneducated eyes...

Should I try to reproduce locally (pointers welcome) or are those known spurious issues ?

sergiusens:
you will need to click on "Update branch".

Ping one of us when you do so we know not to land anything in between.

Thanks

vilagithub:
From http://162.213.35.179:8080/job/github-snapcraft-autopkgtest-cloud/429/console :

Finding IP address succeeded: 10.42.61.202
Waiting until ssh becomes available
Timed out waiting for ssh. Aborting! Console log:

is probably the relevant error: autopkgtest fails to setup the slave

vilagithub:
Rebased and test passing \o/

PR #430, opened 2016-04-06
Make geoip optional and local sources a default

sergiusens:
LP: #1536691


snapcraft/_options.py
```diff
@@ -0,0 +1,41 @@
...
+
+    @use_geoip.setter
+    def use_geoip(self, setting):
+        self.__use_geoip = setting
+
+    def __init__(self):
+        self.__use_geoip = False
+
+
+_project = _Project()
```
kyrofa: Not a huge fan of introducing more global state. Would it be dreadful to pass this around?

snapcraft/_options.py
```diff
@@ -0,0 +1,41 @@
...
+
+class _Project:
+
+    @property
+    def use_geoip(self):
+        return self.__use_geoip
+
+    @use_geoip.setter
+    def use_geoip(self, setting):
+        self.__use_geoip = setting
```
kyrofa: What is the benefit of this encapsulation?

sergiusens: El miércoles, 6 de abril de 2016 14h'52:17 ART, Kyle Fazzari 
>> +# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
>> +# GNU General Public License for more details.
>> +#
>> +# You should have received a copy of the GNU General Public License
>> +# along with this program.  If not, see <http://www.gnu.org/licenses/>.
>> +
>> +
>> +class _Project:
>> +
>> +    @property
>> +    def use_geoip(self):
>> +        return self.__use_geoip
>> +
>> +    @use_geoip.setter
>> +    def use_geoip(self, setting):
>> +        self.__use_geoip = setting
> 
> What is the benefit of this encapsulation?

By doing it this way we can change the implementation internally without 
worrying about the external interface. Also no one would be able to reach 
this.


-- 

snapcraft/_options.py
sergiusens: El miércoles, 6 de abril de 2016 14h'51:44 ART, Kyle Fazzari 
>> +
>> +class _Project:
>> +
>> +    @property
>> +    def use_geoip(self):
>> +        return self.__use_geoip
>> +
>> +    @use_geoip.setter
>> +    def use_geoip(self, setting):
>> +        self.__use_geoip = setting
>> +
>> +    def __init__(self):
>> +        self.__use_geoip = False
>> +
>> +
>> +_project = _Project()
> 
> Not a huge fan of introducing more global state. Would it be 
> dreadful to pass this around?

I don't see a way around this.Also this was supposed to be step one to 
encapsulate everything in `common` to this `_Project` class.

This looks pretty similar to how the logger package in python is 
implemented, except we'd be using a single element instead of N.

Also, if we pass this around I'd like to pass all of these in `common` 
around. It will be huge as there will always be more. Thoughts?


-- 

snapcraft/_options.py
kyrofa: Yeah that's good API design-- is it considered best-practice to do this for all properties (e.g. like it is in C++)?

sergiusens: I do feel we will do this for the arch ones specifically.

Another thing I was contemplating was making the values set immutable and only allow them to be set on class instantiation. Thoughts?

kyrofa: > Another thing I was contemplating was making the values set immutable and only allow them to be set on class instantiation. Thoughts?

I'm not sure I understand what you mean by "values set" here. You mean the values of this class? e.g. so plugins can't change these settings for the entire project?

sergiusens:
Also fixes LP: #1561068

kyrofa:
Yeah this looks good :+1:

snapcraft/_options.py
sergiusens: Exactly that. But we can dive into that when we tackle the common rework

kyrofa: Yeah :+1: on both the immutable and tackling when reworking common.

sergiusens:
Darn...

in examples
```
[1;32mEverything passed[0m

Build step 'Execute shell' marked build as failure
Recording test results
Setting status of 18a8a2d8082cfc3d7fdb6fe0c70d0cbc1d22c4b7 to FAILURE with url http://162.213.35.179:8080/job/github-snapcraft-examples-tests-cloud/527/ and message: ' '
```

and in autopackage tests:
```
ERROR: test_tomcat_maven_webapp.TomcatMavenWebappTestCase.test_tomcat_maven_webapp
output: {{{b"DEPRECATED: The 'tar-content' plugin's functionality has been replaced by the 'copy' plugin, and it will soon be removed.\nPreparing to pull local-files \nPulling local-files \nPreparing to pull tomcat \nPulling tomcat \nunexpected http status code when downloading 503\n"}}}

  File "/tmp/adt-run.J49D6j/build.Ojg/real-tree/examples_tests/test_tomcat_maven_webapp.py", line 25, in test_tomcat_maven_webapp
  File "/tmp/adt-run.J49D6j/build.Ojg/real-tree/examples_tests/__init__.py", line 131, in build_snap
    stderr=subprocess.STDOUT)
    **kwargs).stdout
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['/tmp/adt-run.J49D6j/build.Ojg/real-tree/bin/snapcraft', 'snap']' returned non-zero exit status 1
```

PR #431, opened 2016-04-06
Lifecycle: Check to see if prerequisite is already built.

kyrofa:
Currently, if one specifies a part to be built, the lifecycle checks to see if that part has any dependencies. If so, it immediately errors out if those dependencies are not in the list to be built. This PR fixes LP: [#1537790](https://bugs.launchpad.net/snapcraft/+bug/1537790) by checking that the dependencies are not already built before erroring.


snapcraft/tests/test_lifecycle.py
```diff
@@ -17,24 +17,35 @@
...
 version: 0
-summary: test stage
-description: if the build is succesful the state file will be updated
-icon: icon.png
+summary: test
+description: test
+"""
 
-parts:
+        super().make_snapcraft_yaml(yaml + parts)
```
sergiusens: maybe format is better here?

snapcraft/tests/test_lifecycle.py
```diff
@@ -42,7 +53,6 @@ def test_exception_when_dependency_is_required(self):
     after:
       - part1
 """)
-        open('icon.png', 'w').close()
```
sergiusens: :100:

sergiusens:
one minor comment, awesome work!

snapcraft/tests/test_lifecycle.py
kyrofa: Yeah it lets me handle the type a little better, too-- done.

sergiusens:
retest this please

PR #433, opened 2016-04-07
Converge all commands into one

sergiusens:
LP: #1567058
LP: #1567041


sergiusens:
I am also fixing bug #1567041 here

snapcraft/__init__.py
```diff
@@ -115,7 +115,11 @@
 import os
 import shutil
 
+from snapcraft._help import topic_help  # noqa
+from snapcraft._store import login, logout, upload  # noqa
 from snapcraft import common
+from snapcraft import lifecycle  # noqa
+from snapcraft import plugins  # noqa
```
kyrofa: Why would these issue warnings?

snapcraft/__init__.py
```diff
@@ -115,7 +115,11 @@
 import os
 import shutil
 
+from snapcraft._help import topic_help  # noqa
+from snapcraft._store import login, logout, upload  # noqa
```
kyrofa: Elsewhere we use a format like:

```python
from snapcraft._store import (
    login,
    logout,
    upload
)
```
Should we be consistent?

sergiusens: Those are imported but not used, so I'd like to call them out explicitly.
It's defacto done like this for a package `__init__.py`

snapcraft/__init__.py
sergiusens: Because they are imported but not used, but by doing this you can just `import snapcraft` and get `snapcraft.plugins` and `snapcraft.lifecycle` already imported as well :smile:

Just like you get `os.path` when importing `os`

snapcraft/__init__.py
kyrofa: Ah, which answers my other question there, too :smiley: .

snapcraft/__init__.py
kyrofa: Thanks for explaining!

snapcraft/__init__.py
sergiusens: I can give you a personal note here too, I don't like the way tupling looks either :stuck_out_tongue:

kyrofa: Let's say you imported 15 things here. How do you break the line?

snapcraft/lifecycle.py
```diff
@@ -25,11 +31,31 @@
...
+    if os.path.exists('snapcraft.yaml'):
+        raise EnvironmentError('snapcraft.yaml already exists!')
+    yaml = _TEMPLATE_YAML.strip()
+    with open('snapcraft.yaml', mode='w+') as f:
+        f.write(yaml)
+    logger.info('Created snapcraft.yaml.')
+
+
 def execute(step, part_names=None):
     """Exectute until step in the lifecycle.
```
kyrofa: Just noticed this says "Exectute." Might as well fix, eh?

snapcraft/main.py
```diff
@@ -19,82 +19,87 @@
...
+                                        (including backtraces)
+  --no-parallel-build                   use only a single build job per part
+                                        (the default number of jobs per part is
+                                        equal to the number of CPUs)
+  --target-arch ARCH                    EXPERIMENTAL: sets the target
+                                        architecture. Very few plugins support
+                                        this.
+  -s <step>, --step <step>              only clean the specified step and those
+                                        that depend upon it. <step> can be one
+                                        of: pull, build, stage or strip.
```
kyrofa: Still a relative docopt newbie, but should `--step` be here if it's in the Usage section above? Doesn't it make it a global option, when it should really be specific to `clean`?

snapcraft/main.py
```diff
@@ -19,82 +19,87 @@
...
 Usage:
-  snapcraft [--version | --help] [options] COMMAND [ARGS ...]
   snapcraft [options]
+  snapcraft [options] init
+  snapcraft [options] pull [<part> ...]
+  snapcraft [options] build [<part> ...]
+  snapcraft [options] stage [<part> ...]
+  snapcraft [options] strip [<part> ...]
+  snapcraft [options] clean [<part> ...] [--step <step>]
+  snapcraft [options] snap [<directory> --output <snap-file>]
```
kyrofa: Does this syntax mean that `--output` is required when specifying `<directory>`, and not possible to specify when _not_ specifying `<directory>`? Should this instead be `snapcraft [options] snap [<directory>] [--output <snap-file>]` ?

snapcraft/__init__.py
sergiusens: With tupling, but it would also be an indication that we are doing something wrong IMHO

snapcraft/main.py
```diff
@@ -19,82 +19,87 @@
...
+                                        (the default number of jobs per part is
+                                        equal to the number of CPUs)
+  --target-arch ARCH                    EXPERIMENTAL: sets the target
+                                        architecture. Very few plugins support
+                                        this.
+  -s <step>, --step <step>              only clean the specified step and those
+                                        that depend upon it. <step> can be one
+                                        of: pull, build, stage or strip.
+  -o <snap-file>, --output <snap-file>  used in case you want to rename the
+                                        snap.
```
kyrofa: Same question about the global option versus step-specific options.

snapcraft/main.py
```diff
@@ -19,82 +19,87 @@
 snapcraft
 
 Usage:
-  snapcraft [--version | --help] [options] COMMAND [ARGS ...]
   snapcraft [options]
+  snapcraft [options] init
+  snapcraft [options] pull [<part> ...]
+  snapcraft [options] build [<part> ...]
```
kyrofa: Now that these are all together, should `--no-parallel-build` be an option specific to build?

snapcraft/main.py
sergiusens: It doesn't really matter. I can break down the header (Options) into "Options for cleaning", "Options for building"
and then just Options.

*Meaning of [options] shortcut slightly changed. Previously it meant "any known option". Now it means "any option not in usage-pattern". This avoids the situation when an option is allowed to be repeated unintentionally.*

from the first lines of https://github.com/docopt/docopt

snapcraft/main.py
sergiusens: yeah, probably (and also to the `snapcraft` with no sub-command)

snapcraft/lifecycle.py
sergiusens: with what? "Executes"? Sorry I keep getting lost with these.

kyrofa: Just remove the extra T: s/Exectute/Execute/

snapcraft/main.py
```diff
@@ -123,13 +130,53 @@ def main():
...
+    elif argless_command:
+        argless_command()
+    elif args['clean']:
+        snapcraft.lifecycle.clean(args['<part>'], args['--step'])
+    elif args['upload']:
+        snapcraft.upload(args['<snap-file>'])
+    elif args['help']:
+        snapcraft.topic_help(args['<topic>'] or args['<plugin>'],
+                             args['--devel'], args['topics'])
+    else:  # snap of default:
```
kyrofa: s/snap of default/snap by default/

sergiusens: I wanted to say "or" but "by" is better :-)

snapcraft/main.py
```diff
@@ -19,82 +19,93 @@
...
-  snapcraft [--version | --help] [options] COMMAND [ARGS ...]
-  snapcraft [options]
+  snapcraft [options] [--no-parallel-build]
+  snapcraft [options] init
+  snapcraft [options] pull [<part> ...]
+  snapcraft [options] build [<part> ...] [--no-parallel-build]
+  snapcraft [options] stage [<part> ...]
+  snapcraft [options] strip [<part> ...]
+  snapcraft [options] clean [<part> ...] [--step <step>]
+  snapcraft [options] snap [<directory> --output <snap-file>]
```
kyrofa: Still got this.

snapcraft/main.py
sergiusens: It is totally valid POSIX to do it this way, you can look at the last bullet here:
https://github.com/docopt/docopt#option-descriptions-format

Also, the tests pass and we have tests for -o and <directory>. If you think it would look nicer to put them in individual brackets that's fine, but look at `git pull`:
```
SYNOPSIS
    git pull [options] [<repository> [<refspec>...]]
```

snapcraft/main.py
sergiusens: It is totally valid POSIX to do it this way, you can look at the last bullet here:
https://github.com/docopt/docopt#option-descriptions-format

Also, the tests pass and we have tests for -o and <directory>. If you think it would look nicer to put them in individual brackets that's fine, but look at `git pull`:
```
SYNOPSIS
    git pull [options] [<repository> [<refspec>...]]
```

kyrofa: Ah, then I'm good with this. My intuition misled me :smiley: .

kyrofa:
I'm a little sad to see the commands go, but I understand the need. :+1:

sergiusens:
El 07/04/16 a las 13:24, Kyle Fazzari escribió:
> I'm a little sad to see the commands go, but I understand the need. :+1:

I feel the same. But it was an invitation to a problem as those commands aren't necessarily independent.

PR #434, opened 2016-04-11
Remove global state attributes from snapcraft.common

sergiusens:


sergiusens:
@kyrofa and @elopio should I add the same for `get_snapdir`, `get_stagedir`, `get_plugindir` and `get_partsdir`

sergiusens:
same for `plugindir`, `schemadir` and `librariesdir` but those are more complicated.

sergiusens:
@elopio example tests fail straight out and for some reason I can't connect to the VPN :disappointed:

sergiusens:
retest this please

snapcraft/_options.py
```diff
@@ -14,16 +14,120 @@
...
+            self.__target_machine = self.__host_machine
+        else:
+            self.__target_machine = _find_machine(target_deb_arch)
+        self.__machine_info = _ARCH_TRANSLATIONS[self.__target_machine]
+
+
+def _find_machine(deb_arch):
+    for machine in _ARCH_TRANSLATIONS:
+        if _ARCH_TRANSLATIONS[machine].get('deb', '') == deb_arch:
+            logger.info('Setting target machine to {!r}'.format(machine))
```
kyrofa: This function isn't actually setting anything-- this `logger.info()` probably belongs in `_set_machine()`, no?

snapcraft/common.py
```diff
@@ -203,25 +119,19 @@ def get_schemadir():
...
-    if get_enable_parallel_builds():
-        try:
-            build_count = multiprocessing.cpu_count()
-        except NotImplementedError:
-            logger.warning('Unable to determine CPU count; disabling parallel '
-                           'builds')
-
-    return build_count
+    raise EnvironmentError(
+        "This plugin is outdated, use 'options.project.parallel_build_count'")
```
kyrofa: Thanks for not just breaking completely here :100: .

snapcraft/tests/__init__.py
```diff
@@ -49,13 +50,16 @@ def setUp(self):
...
-                        common.get_enable_parallel_builds())
         self.addCleanup(common.reset_env)
         common.set_schemadir(os.path.join(__file__,
                              '..', '..', '..', 'schema'))
         self.useFixture(fixtures.FakeLogger(level=logging.ERROR))
 
+        patcher = mock.patch('multiprocessing.cpu_count')
+        self.cpu_count = patcher.start()
+        self.cpu_count.return_value = 2
+        self.addCleanup(patcher.stop)
```
kyrofa: Yeah good idea. I should have done that in the first place.

snapcraft/plugins/kernel.py
```diff
@@ -142,16 +141,14 @@ def __init__(self, name, options):
         self.make_install_targets.extend(self._get_fw_install_targets())
 
         self.os_snap = os.path.join(self.sourcedir, 'os.snap')
-        self._target_arch = get_machine_info(target_machine)
 
-    def set_target_machine(self, machine):
-        self._target_arch = get_machine_info(machine)
+    def set_target_machine(self):
```
kyrofa: A "set" function with no parameter seems odd. Is this something that can be done in the initializer now?

snapcraft/pluginhandler.py
```diff
@@ -169,12 +169,15 @@ def _load_code(self, plugin_name, properties, part_schema):
 
         plugin = _get_plugin(module)
         options = _make_options(part_schema, properties, plugin.schema())
+        # For backwards compatibility we add the project options as an option
+        setattr(options, 'project', self._project_options)
```
kyrofa: `options` currently have a pretty well-defined meaning: "options from the YAML." I like not breaking API in this way, but I don't like diluting `options`. Particularly because: what happens when a local plugin declares `project` as a property in the schema?

Would it be better to actually set a `__project` attribute on the plugin instance and add a getter method on the base plugin (or something similar)?

kyrofa:
Check the title, too :wink: .

snapcraft/_options.py
sergiusens: yes, this was a refactor problem :-)

snapcraft/pluginhandler.py
sergiusens: This will need to go to the base plugin and it will break badly unless I add `**kwargs`, `*args`. What is your take?

snapcraft/plugins/kernel.py
sergiusens: I thought about this, but if we make plugins take the charge on this I can add something to manage this. The problem is plugins that don't support cross compiling.

snapcraft/pluginhandler.py
kyrofa: I'm not quite sure what you mean-- where are the `**kwargs` required?

snapcraft/plugins/kernel.py
kyrofa: > The problem is plugins that don't support cross compiling.

You mean as of right now if one tries cross-compiling with plugins that don't support it you get a good error? That's an excellent point.

snapcraft/__init__.py
```diff
@@ -288,8 +289,8 @@ def env(self, root):
         """
         return []
 
-    def set_target_machine(self):
-        """Set the target compilation architecture to machine."""
```
kyrofa: Can we leave this function, continue taking the `machine` parameter, deprecate it, and simply have it call `enable_cross_compilation()`?

sergiusens: El 11/04/16 a las 14:27, Kyle Fazzari escribió:
> In snapcraft/__init__.py
> <https://github.com/ubuntu-core/snapcraft/pull/434#discussion_r59245215>:
>
> > @@ -288,8 +289,8 @@ def env(self, root):
> >          """
> >          return []
> >  
> > -    def set_target_machine(self):
> > -        """Set the target compilation architecture to machine."""
>
> Can we leave this function, continue taking the |machine| parameter,
> deprecate it, and simply have it call |enable_cross_compilation()|?
While not hard, the target-arch thing was marked API unstable, so not really compelled to do it. Counter thoughts?

snapcraft/pluginhandler.py
```diff
@@ -169,15 +169,26 @@ def _load_code(self, plugin_name, properties, part_schema):
...
+            logger.warning('DEPRECATED: this plugin needs updating to use '
+                           'project options')
+            self.code = plugin(self.name, options)
+            # This is for plugins that don't inherit from BasePlugin
+            if not hasattr(self.code, 'project'):
+                setattr(self.code, 'project', self._project_options)
+            # This is for plugins that inherit from BasePlugin but don't have
+            # project in init.
+            if not self.code.project:
+                self.code.project = self._project_options
```
kyrofa: Thanks for the comments here, very clear. We can probably make that deprecation a little more helpful-- how about something like:

```python
logger.warning('DEPRECATED: the plugin used by part {!r} needs to be updated to accept project options in its initializer. See <link to plugins.md> for more information'.format(self.name)
```

And we can add that third param to the `docs/plugins.md` document.

snapcraft/__init__.py
kyrofa: If people were warned regarding the instability then I'm good with it.

kyrofa:
Looks great! :+1:

PR #436, opened 2016-04-11
Clean command: Don't automatically clean reverse dependencies.

kyrofa:
This PR resolves LP: [#1568004](https://bugs.launchpad.net/snapcraft/+bug/1568004) by making cleaning more consistent with the way parts are built (i.e. requiring that all interdependent parts be explicitly specified).

### DIFF
--- a/snapcraft/lifecycle.py
+++ b/snapcraft/lifecycle.py
@@ -226,18 +226,70 @@ def _reverse_dependency_tree(config, part_name):
     return dependents
 
 
-def _clean_part_and_all_dependents(config, part, staged_state, stripped_state,
-                                   step):
-    # Clean the part in question
-    part.clean(staged_state, stripped_state, step)
-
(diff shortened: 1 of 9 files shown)


snapcraft/lifecycle.py
```diff
@@ -226,19 +226,71 @@ def _reverse_dependency_tree(config, part_name):
...
     dependent_parts = {p for p in config.all_parts
                        if p.name in dependents}
     for dependent_part in dependent_parts:
         dependent_part.clean(staged_state, stripped_state, step)
 
+    # Finally, clean the part in question
+    config.get_part(part_name).clean(staged_state, stripped_state, step)
+
+
+def _humanize_list(items):
```
sergiusens: if monkeys ever read or if an extra terrestrial sees this they will feel left out :wink:

PR #440, opened 2016-04-12
Library crawling: Only warn if ldd fails.

kyrofa:
This PR fixes LP: [#1569280](https://bugs.launchpad.net/snapcraft/+bug/1569280) by catching `ldd` failures and warning about them instead of allowing them to cascade. It also skips crawling object (*.o) files, as its unnecessary.


snapcraft/pluginhandler.py
```diff
@@ -674,7 +674,10 @@ def _find_dependencies(workdir):
     fs_encoding = sys.getfilesystemencoding()
 
     for root, dirs, files in os.walk(workdir.encode(fs_encoding)):
-        for entry in itertools.chain(files, dirs):
+        # Filter out object (*.o) files-- we only care about binaries.
+        entries = (entry for entry in itertools.chain(files, dirs)
+                   if not entry.endswith(b'.o'))
```
sergiusens: :100:

PR #442, opened 2016-04-13
Automatically re-strip if dirty.

kyrofa:
Currently Snapcraft doesn't notice YAML changes if a given step has already run-- it simply says that it already ran and halts execution. This PR makes progress on LP: [#1477904](https://bugs.launchpad.net/snapcraft/+bug/1477904) by changing that for the strip step, where now if one changes the `snap` keyword in the YAML, strip will notice that it's out of date and automatically clean itself and re-run.

### DIFF
--- a/snapcraft/internal/states/_state.py
+++ b/snapcraft/internal/states/_state.py
@@ -16,30 +16,24 @@
 
 import yaml
 
-from snapcraft.internal.states._state import State
 
-
-def _strip_state_constructor(loader, node):
-    parameters = loader.construct_mapping(node)
-    return StripState(**parameters)
-
-yaml.add_constructor(u'!StripState', _strip_state_constructor)
-
-
-class StripState(State):
-    yaml_tag = u'!StripState'
-
+class State(yaml.YAMLObject):
     @classmethod
     def properties_of_interest(cls, options):
-        return {'snap': getattr(options, 'snap', ['*']) or ['*']}
+        return {}
 
-    def __init__(self, files, directories, dependency_paths=None,
-                 options=None):
-        super().__init__(options)
+    def __init__(self, options):
+        self.properties = self.properties_of_interest(options)
 
-        self.files = files
-        self.directories = directories
-        self.dependency_paths = set()
+    def __repr__(self):
+        items = sorted(self.__dict__.items())
+        strings = (': '.join((key, repr(value))) for key, value in items)
+        representation = ', '.join(strings)
 
-        if dependency_paths:
-            self.dependency_paths = dependency_paths
+        return '{}({})'.format(self.__class__.__name__, representation)
+
+    def __eq__(self, other):
+        if type(other) is type(self):
+            return self.__dict__ == other.__dict__
+
+        return False
--- a/snapcraft/internal/states/_state.py
+++ b/snapcraft/internal/states/_state.py
@@ -20,7 +20,12 @@ import yaml
 class State(yaml.YAMLObject):
     @classmethod
     def properties_of_interest(cls, options):
-        return {}
+        """Extract the properties concerning this step from the options.
+
+        Note that these options come from the YAML for a given part.
+        """
+
+        raise NotImplementedError
 
     def __init__(self, options):
         self.properties = self.properties_of_interest(options)
--- a/snapcraft/pluginhandler.py
+++ b/snapcraft/pluginhandler.py
@@ -586,6 +586,13 @@ def _migratable_filesets(fileset, srcdir):
                      not os.path.islink(os.path.join(srcdir, x))])
     snap_files = snap_files - snap_dirs
 
+    # Make sure we also obtain the parent directories of files
+    for snap_file in snap_files:
+        dirname = os.path.dirname(snap_file)
+        while dirname:
+            snap_dirs.add(dirname)
+            dirname = os.path.dirname(dirname)
+
     return snap_files, snap_dirs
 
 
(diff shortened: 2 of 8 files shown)


snapcraft/pluginhandler.py
```diff
@@ -599,18 +612,29 @@ def _migrate_files(snap_files, snap_dirs, srcdir, dstdir, missing_ok=False,
         common.link_or_copy(src, dst, follow_symlinks=follow_symlinks)
 
 
+def _remove_empty_tree(directory, barrier):
+    """Remove an empty directory tree up to (not including) a barrier path."""
+
+    while (directory != barrier and os.path.isdir(directory) and
```
sergiusens: this feels like a dangerous function.

kyrofa: There's of course the `listdir()` and the `rmdir` versus `rmtree`, but yeah it makes you blink. I'm not sure there's a better way to handle this though-- I found the problem with [this test case](https://github.com/ubuntu-core/snapcraft/pull/442/files/23c36316a7762b89e1809580c0bab05efbd154bc#diff-b308bc3f4f6ea8bc23d67863c6b31a33R803).

sergiusens:
Looks good, just a bit afraid of that function that walks upwards :-)

snapcraft/pluginhandler.py
```diff
@@ -634,7 +630,8 @@ def _clean_migrated_files(snap_files, snap_dirs, directory):
 
     for snap_dir in snap_dirs:
         migrated_directory = os.path.join(directory, snap_dir)
-        _remove_empty_tree(migrated_directory, directory)
+        if not os.listdir(migrated_directory):
```
sergiusens: if this is `True` shouldn't it log a warning?

snapcraft/tests/test_pluginhandler.py
```diff
@@ -199,7 +199,7 @@ def test_migrate_snap_files_already_exists(self):
         with open('install/foo', 'w') as f:
             f.write('installed')
 
-        files, dirs = pluginhandler._migratable_filesets('*', 'install')
+        files, dirs = pluginhandler._migratable_filesets(['*'], 'install')
```
sergiusens: I wonder how we got away with this.

snapcraft/pluginhandler.py
kyrofa: No, other parts might have files in there, too. Right?

snapcraft/tests/test_pluginhandler.py
kyrofa: I was hoping you wouldn't notice that :stuck_out_tongue: .

sergiusens:
:+1:

kyrofa:
retest this please

elopio:
retest this please

snapcraft/internal/states/_state.py
```diff
@@ -0,0 +1,39 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import yaml
+
+
+class State(yaml.YAMLObject):
+    @classmethod
+    def properties_of_interest(cls, options):
+        return {}
```
elopio: if this classmethod is only intended to be extended by subclasses, I think a good practices is to raise NotImplementedError.

https://docs.python.org/3/library/exceptions.html#NotImplementedError

snapcraft/pluginhandler.py
```diff
@@ -268,17 +278,20 @@ def build(self, force=False):
...
                                       '(already clean)')
             return
 
-        self.notify_part_progress('Cleaning build for')
+        self.notify_part_progress('Cleaning build for', hint)
         self.code.clean_build()
         self.mark_cleaned('build')
 
+    def fileset_for(self, step):
+        return getattr(self.code.options, step, ['*']) or ['*']
```
elopio: is this fileset_for being called?

elopio:
I'm having troubles understanding this, so I think you are missing a little more docs in the code.

the properties are of interest to who?
The idea is that you initialize the instance with the properties of interest of the class. Then something can change the instance properties, and that will mean that the step is dirty. Is that right? If so, what is it that changes the properties?

sergiusens:
@elopio what changes the properties is someone editing `snapcraft.yaml` changing the `snap` keyword which would result in having the `strip` step dirty and needing to redo it. That key for a `part` is reflected in `PluginHandler(...).code.options.snap`

snapcraft/pluginhandler.py
sergiusens: @elopio good catch, this should probably replace
https://github.com/ubuntu-core/snapcraft/pull/442/files#diff-2185a0b0812f4f8b12703640a5d79517R297

snapcraft/internal/states/_state.py
kyrofa: Yeah that would effectively force subclasses to implement it rather than provide sensible defaults, but I'm okay with it as all states probably will.

snapcraft/pluginhandler.py
kyrofa: Ah, thanks for that-- I was refactoring this before I realized I was going to do it wrong. Forgot to remove it.

kyrofa:
> I'm having troubles understanding this, so I think you are missing a little more docs in the code.

That's fair! I'll try documenting the states a bit more.

> the properties are of interest to who?

The step. Essentially it's saying "what YAML properties do you find interesting enough to consider yourself dirty if they change?"

> Then something can change the instance properties, and that will mean that the step is dirty. Is that right? If so, what is it that changes the properties?

Correct. The properties of interest come from the YAML, so if one changes the YAML one changes the properties of interest (as long as they change a property of interest, otherwise it's uninteresting and the step isn't dirty). The dirty calculation happens [here](https://github.com/ubuntu-core/snapcraft/pull/442/files#diff-2185a0b0812f4f8b12703640a5d79517R191).

snapcraft/internal/states/_state.py
kyrofa: Fixed.

snapcraft/pluginhandler.py
kyrofa: Fixed.

PR #443, opened 2016-04-13
Automatically re-stage if dirty.

kyrofa:
Currently Snapcraft only notices YAML changes for the strip step. For all other steps, if the step has already run it simply says so and bails. This PR makes progress on LP: [#1477904](https://bugs.launchpad.net/snapcraft/+bug/1477904) by changing that for the stage step, where now if one changes the `stage` keyword in the YAML, stage will notice that it's out of date and automatically clean itself (along with the strip step since it depends upon stage) and re-run.

### DIFF
--- a/snapcraft/pluginhandler.py
+++ b/snapcraft/pluginhandler.py
@@ -256,10 +256,10 @@ class PluginHandler:
         self.mark_done('pull')
 
     def clean_pull(self, hint=''):
-        state_file = self._step_state_file('pull')
-        if not os.path.isfile(state_file):
+        if self.is_clean('pull'):
+            hint = (hint + ' (already clean)').strip()
             self.notify_part_progress('Skipping cleaning pulled source for',
-                                      '(already clean)')
+                                      hint)
             return
 
         self.notify_part_progress('Cleaning pulled source for', hint)
@@ -284,10 +284,10 @@ class PluginHandler:
         self.mark_done('build')
 
     def clean_build(self, hint=''):
-        state_file = self._step_state_file('build')
-        if not os.path.isfile(state_file):
+        if self.is_clean('build'):
+            hint = (hint + ' (already clean)').strip()
             self.notify_part_progress('Skipping cleaning build for',
-                                      '(already clean)')
+                                      hint)
             return
 
         self.notify_part_progress('Cleaning build for', hint)
@@ -331,19 +331,18 @@ class PluginHandler:
         # dependencies here too
 
         self.mark_done('stage', internal.states.StageState(
-            snap_files, snap_dirs))
+            snap_files, snap_dirs, self.code.options))
 
     def clean_stage(self, project_staged_state, hint=''):
-        state_file = self._step_state_file('stage')
-        if not os.path.isfile(state_file):
+        if self.is_clean('stage'):
+            hint = (hint + ' (already clean)').strip()
             self.notify_part_progress('Skipping cleaning staging area for',
-                                      '(already clean)')
+                                      hint)
             return
 
         self.notify_part_progress('Cleaning staging area for', hint)
 
-        with open(state_file, 'r') as f:
-            state = yaml.load(f.read())
+        state = self.get_state('stage')
 
         try:
             self._clean_shared_area(self.stagedir, state,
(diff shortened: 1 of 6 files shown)


snapcraft/pluginhandler.py
```diff
@@ -256,10 +256,10 @@ def pull(self, force=False):
         self.mark_done('pull')
 
     def clean_pull(self, hint=''):
-        state_file = self._step_state_file('pull')
-        if not os.path.isfile(state_file):
+        if self.is_clean('pull'):
+            hint = (hint + ' (already clean)').strip()
```
sergiusens: `format` is better here.

snapcraft/pluginhandler.py
```diff
@@ -284,10 +284,10 @@ def build(self, force=False):
         self.mark_done('build')
 
     def clean_build(self, hint=''):
-        state_file = self._step_state_file('build')
-        if not os.path.isfile(state_file):
+        if self.is_clean('build'):
+            hint = (hint + ' (already clean)').strip()
```
sergiusens: `format` ditto

snapcraft/pluginhandler.py
```diff
@@ -331,19 +331,18 @@ def stage(self, force=False):
...
 
         self.mark_done('stage', internal.states.StageState(
-            snap_files, snap_dirs))
+            snap_files, snap_dirs, self.code.options))
 
     def clean_stage(self, project_staged_state, hint=''):
-        state_file = self._step_state_file('stage')
-        if not os.path.isfile(state_file):
+        if self.is_clean('stage'):
+            hint = (hint + ' (already clean)').strip()
```
sergiusens: `format`

snapcraft/pluginhandler.py
```diff
@@ -396,16 +395,15 @@ def strip(self, force=False):
             snap_files, snap_dirs, dependency_paths, self.code.options))
 
     def clean_strip(self, project_stripped_state, hint=''):
-        state_file = self._step_state_file('strip')
-        if not os.path.isfile(state_file):
+        if self.is_clean('strip'):
+            hint = (hint + ' (already clean)').strip()
```
sergiusens: `format`

snapcraft/tests/test_lifecycle.py
```diff
@@ -169,6 +172,68 @@ def test_dirty_strip_restrips(self):
             'Stripping part1 \n',
             self.fake_logger.output)
 
+    def test_dirty_stage_restages(self):
```
sergiusens: Can we have the same test but with 2 parts?

snapcraft/pluginhandler.py
kyrofa: Done.

snapcraft/pluginhandler.py
kyrofa: Done.

snapcraft/pluginhandler.py
kyrofa: Done.

snapcraft/pluginhandler.py
kyrofa: Done.

snapcraft/tests/test_lifecycle.py
kyrofa: I updated the strip and stage tests here to have multiple parts, with two tests each-- where one part is dirty (with the other built) and where both parts are dirty.

snapcraft/pluginhandler.py
```diff
@@ -296,7 +296,7 @@ def clean_build(self, hint=''):
 
     def migratable_fileset_for(self, step):
         plugin_fileset = self.code.snap_fileset()
-        fileset = getattr(self.code.options, step, ['*']) or ['*']
+        fileset = (getattr(self.code.options, step, ['*']) or ['*']).copy()
         fileset.extend(plugin_fileset)
```
sergiusens: Ah, nice catch

kyrofa: Can't really take the credit, I'm afraid. Good example tests!

sergiusens: Yeah I was talking to autopackage tests, not you @kyrofa :grinning:

kyrofa: :cry:

kyrofa: At least the CI that **is** working is helpful!

PR #444, opened 2016-04-13
Make meta a class

sergiusens:
This is work to enable easier handling of the global `common.get_Xdir()` calls 

LP: #1551849

### DIFF
--- a/snapcraft/internal/meta.py
+++ b/snapcraft/internal/meta.py
@@ -60,7 +60,7 @@ class CommandError(Exception):
     pass
 
 
-def create_snap_packaging(config_data):
+def create_snap_packaging(config_data, snap_dir, parts_dir):
     """Create snap.yaml and related assets in meta.
     Create  the meta directory and provision it with snap.yaml
     in the snap dir using information from config_data.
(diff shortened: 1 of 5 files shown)


snapcraft/internal/meta.py
```diff
@@ -0,0 +1,259 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
```
kyrofa: Should this be `_meta.py`?

snapcraft/internal/meta.py
```diff
@@ -0,0 +1,259 @@
...
+    :return: meta_dir.
+    """
+    packaging = _SnapPackaging(config_data, snap_dir, parts_dir)
+    packaging.write_snap_yaml()
+    packaging.setup_assets()
+
+    return packaging.meta_dir
+
+
+class _SnapPackaging:
```
kyrofa: I'm not sure I completely understand the benefit of having a class do this versus the standalone functions. Am I missing something?

snapcraft/internal/meta.py
sergiusens: It is just like `states` (although a package); as long as it is in `internal` it should be fine.

snapcraft/internal/meta.py
sergiusens: I don't need to pass around `snap_dir` and `parts_dir` which if `common.get` weren't used all over would really complicate the API.

kyrofa: Ah, okay, I gotcha.

sergiusens: I also plan to get rid of `common.assemble_env` this could be done in a much better way.

kyrofa:
Looks good! :+1:

PR #445, opened 2016-04-13
Move common directories to ProjectOptions

sergiusens:
LP: #1551849


kyrofa:
I think it deserves two: :+1: :+1:

PR #447, opened 2016-04-14
Detect dirty build step.

kyrofa:
Currently Snapcraft notices YAML changes only for the strip and stage steps. For the other steps, if they've already run, it simply says so and bails. This PR makes progress on LP: [#1477904](https://bugs.launchpad.net/snapcraft/+bug/1477904) by changing that for the build step, where now if one changes any part of the YAML that the part's plugin says is associated with building, build will notice that it's out of date and error out asking the user to clean the build step to continue.

The build step is not automatically cleaned like the strip and stage steps because builds can take a very long time and we don't want to blow one away without the user explicitly asking to do so.

### DIFF
--- a/snapcraft/__init__.py
+++ b/snapcraft/__init__.py
@@ -163,7 +163,8 @@ class BasePlugin:
             },
             'required': [
                 'source',
-            ]
+            ],
+            'build-properties': []
         }
 
     @property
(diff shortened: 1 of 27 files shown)


sergiusens:
:+1: for some reason I thought we were going to rebuild automatically, but I like this and we can improve on it for such case

PR #448, opened 2016-04-14
Detect dirty pull step.

kyrofa:
Currently Snapcraft notices YAML changes for every step except the pull step.This PR finishes up LP: [#1477904](https://bugs.launchpad.net/snapcraft/+bug/1477904) by adding this functionality for the pull step, where now if one changes any part of the YAML that the part's plugin says is associated with pulling, pull will notice that it's out of date and error out asking the user to clean the pull step to continue.

The pull step is not automatically cleaned like the strip and stage steps because pulls can take a very long time (and require network) and we don't want to blow one away without the user explicitly asking to do so.

### DIFF
--- a/snapcraft/__init__.py
+++ b/snapcraft/__init__.py
@@ -164,6 +164,8 @@ class BasePlugin:
             'required': [
                 'source',
             ],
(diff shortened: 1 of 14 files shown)


sergiusens:
:+1:

PR #451, opened 2016-04-14
Only add cross build packages when cross compiling.

sergiusens:
LP: #1570414


kyrofa:
Looks good and tested successfully on armhf :+1:

PR #455, opened 2016-04-14
Nil plugin: Make purpose more clear in documentation.

kyrofa:
Some users were under the impression that nil could pull sources, when it can't. This PR resolves LP: [#1549691](https://bugs.launchpad.net/snapcraft/+bug/1549691) by hopefully making that more clear, and giving an example of what it CAN do.

### DIFF
--- a/snapcraft/tests/test_pluginhandler.py
+++ b/snapcraft/tests/test_pluginhandler.py
@@ -1345,7 +1345,7 @@ class CleanTestCase(tests.TestCase):
             with self.subTest(key=key):
                 self.clear_common_directories()
 
-                schema = {'snap': {'type': 'array'}}
+                schema = {'stage': {'type': 'array'}}
                 properties = {'stage': value['fileset']}
 
                 handler = pluginhandler.load_plugin(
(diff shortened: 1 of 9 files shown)


sergiusens:
By also failing on using keywords not available to the part we are golden here.

kyrofa:
@sergiusens I thought that was something done in the new validator, but it doesn't seem so. Did I misunderstand that?

kyrofa:
> I thought that was something done in the new validator, but it doesn't seem so. Did I misunderstand that?

It was, but not in the nil plugin, so fixed that here as well.

snapcraft/tests/test_pluginhandler.py
```diff
@@ -1345,7 +1345,7 @@ def test_clean_stage(self):
             with self.subTest(key=key):
                 self.clear_common_directories()
 
-                schema = {'snap': {'type': 'array'}}
+                schema = {'stage': {'type': 'array'}}
```
sergiusens: what's this about?

kyrofa: A copy-paste error in a test that was only discovered as a result of the new `additionalProperties: false`.

sergiusens:
Looks rather good; a test in would be good to capture the fact that source can't be used with nil; an integration test is fine if it is easier

kyrofa:
> a test in would be good to capture the fact that source can't be used with nil; an integration test is fine if it is easier

Done. I did end up doing an integration test as the validation is handled in the pluginhandler but it depends on the nil plugin-- the test couldn't logically go in either place.

PR #457, opened 2016-04-14
kernel plugin: fix decompressio of lzma initrd (and others)

sergiusens:


snapcraft/plugins/kernel.py
```diff
@@ -179,9 +181,23 @@ def _unpack_generic_initrd(self):
...
+                temp_dir, 'squashfs-root', initrd_path)
+            cmd = shlex.split('file -L --mime-type {}'.format(tmp_initrd_path))
+            result = subprocess.check_output(cmd).decode(
+                sys.getfilesystemencoding())
+            mime_type = result.split()[-1]
+            logger.debug('initrd mime_type: {} {}'.format(
+                tmp_initrd_path, mime_type))
+            decompressor = 'gzip'
+            if 'gzip' in mime_type:
+                decompressor = 'gzip'
```
elopio: this if is not required because it doesn't change the value of the decompressor var.

kyrofa:
Closing in favor of #463.

PR #459, opened 2016-04-15
Remove LXD container once we've finished a build

morphis:
There is no need to keep a container if we take a different name
each time we run a build anyway. Container construction is fast
and therefor we can just drop the container once we're done with it.
    
LP: #1570708

### DIFF
--- a/snapcraft/lxd.py
+++ b/snapcraft/lxd.py
@@ -35,11 +35,11 @@ _PROXY_KEYS = ['http_proxy', 'https_proxy', 'no_proxy', 'ftp_proxy']
 
 class Cleanbuilder:
 
-    def __init__(self, snap_output, project, deb_arch,
+    def __init__(self, snap_output, tar_filename, project_options,
                  server=_DEFAULT_IMAGE_SERVER):
         self._snap_output = snap_output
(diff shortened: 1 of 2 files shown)


snappy-m-o:

snappy-m-o:

morphis:
Stacked on top of https://github.com/ubuntu-core/snapcraft/pull/458

snapcraft/lxd.py
```diff
@@ -68,6 +68,7 @@ def _create_container(self):
             yield
         finally:
             check_call(['lxc', 'stop', self._container_name])
+            check_call(['lxc', 'delete', self._container_name])
```
sergiusens: It is strange that this is needed as the container is ephemeral so should be destroyed on `stop`

morphis:
@sergiusens It could be that I saw left over container just because I didn't fixed the problem PR 458 fixes. Let me check that.

sergiusens:
I'll close this for now; my testing shows that they are indeed destroyed, but they have to be started in the first place which was failing and with the current logic this wouldn't really cut it either.

Thanks for the PR!

PR #463, opened 2016-04-15
Kernel plugin: Support lzma/xz-compressed initrd.

kyrofa:
This is another attempt at fixing LP: [#1569337](https://bugs.launchpad.net/snapcraft/+bug/1569337) which uses the magic library instead of shelling out to `file`. It also adds some tests.

### DIFF
--- a/snapcraft/plugins/kernel.py
+++ b/snapcraft/plugins/kernel.py
@@ -51,13 +51,12 @@ The following kernel specific options are provided by this plugin:
       list of device trees to build, the format is <device-tree-name>.dts.
 """
 
-import logging
 import glob
+import logging
+import magic
 import os
-import shlex
 import shutil
 import subprocess
-import sys
 import tempfile
 
 from snapcraft import storeapi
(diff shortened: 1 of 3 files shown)


snapcraft/plugins/kernel.py
```diff
@@ -179,9 +180,33 @@ def _unpack_generic_initrd(self):
...
+
+            # Support gzip (can be gzip or x-gzip)
+            if 'gzip' in mime_type:
+                decompressor = 'gzip'
+            # Support lzma/xz
+            elif any(x in mime_type for x in ('x-xz', 'x-lzma')):
+                decompressor = 'xz'
+            else:
+                raise RuntimeError(
+                    'initrd file type is unsupported: {!r}'.format(mime_type))
```
elopio: great!

snapcraft/plugins/kernel.py
```diff
@@ -179,9 +180,33 @@ def _unpack_generic_initrd(self):
...
+            mime_type = mime_detector.file(tmp_initrd_path)
+            if not mime_type:
+                raise RuntimeError(
+                    'Unable to determine mime type for {!r}: {}'.format(
+                        tmp_initrd_path, os.strerror(mime_detector.errno())))
+            logger.debug('initrd mime_type: {} {}'.format(
+                tmp_initrd_path, mime_type))
+
+            # Support gzip (can be gzip or x-gzip)
+            if 'gzip' in mime_type:
```
elopio: I think this check should be explicit for gzip or x-gzip.
There could be a case where we find foo-gzip, and we should fail with the unsupported exception.

snapcraft/plugins/kernel.py
elopio: oh, almost great. You are missing a test for this case :)

elopio:
:+1: with two buts.

snapcraft/plugins/kernel.py
kyrofa: Gah, you're too fast. Just pushed it up.

snapcraft/plugins/kernel.py
kyrofa: Alright, agreed.

kyrofa: Done.

kyrofa:
retest this please

PR #464, opened 2016-04-15
autopkgtests: run the tests using the installed package

elopio:
LP: #1570992

### DIFF
--- a/runtests.sh
+++ b/runtests.sh
@@ -77,7 +77,7 @@ run_integration(){
         export SNAPCRAFT=snapcraft-coverage
     fi
 
-    python3 -m unittest discover -b -v -s integration_tests
+    SNAPCRAFT_FROM_SOURCE=1 python3 -m unittest discover -b -v -s integration_tests
 }
 
 run_examples(){
@@ -86,7 +86,7 @@ run_examples(){
         export SNAPCRAFT=snapcraft-coverage
     fi
 
-    python3 -m examples_tests "$@"
+    SNAPCRAFT_FROM_SOURCE=1 python3 -m examples_tests "$@"
 }
 
 parseargs "$@"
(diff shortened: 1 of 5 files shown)


runtests.sh
```diff
@@ -77,7 +77,7 @@ run_integration(){
         export SNAPCRAFT=snapcraft-coverage
     fi
 
-    python3 -m unittest discover -b -v -s integration_tests
+    SNAPCRAFT_FROM_SOURCE=1 python3 -m unittest discover -b -v -s integration_tests
```
kyrofa: Will this break the ability to run the tests directly while developing? (e.g. `python3 -m unittest snapcraft.tests.foo`)

elopio: well, not break, but you will have to add the env var to your statement. What we could do is to add an arg to ./runtests integration so you can pass snapcraft.tests.foo. So my idea is that you would use ./runtests if you want to check the source, and adt-run if you want to check the installed.

I didn't want to rely on whether it is installed or not, because I have it installed and sometimes I still want to run it from source, son snapcraft --version or even just "which" are not useful, IMO.

I could have done it the other way, defining SNAPCRAFT_FROM_INSTALLED on debian/tests, but that sounded less intuitive for me.

kyrofa: > What we could do is to add an arg to ./runtests integration so you can pass snapcraft.tests.foo. So my idea is that you would use ./runtests if you want to check the source, and adt-run if you want to check the installed.

Yeah I do this often enough I'd have to add that variable to my .bashrc. It'd be better if runtests accepted the test module, or we did `SNAPCRAFT_FROM_INSTALLED` instead (since that's only done on CI).

elopio: ok, I will invert the logic. And please open a bug so I don't forget to add the argument handling to ./runtests integration.

kyrofa:
Yeah I like this :+1: . It needs to be rebased though.

PR #466, opened 2016-04-17
Don't expose broken or fragile APIs to plugins

sergiusens:
LP: #1571264


sergiusens:
I've exploratory tested this to death; I hope either @kyrofa or @elopio approve :-)

snapcraft/common.py
```diff
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright (C) 2015, 2016 Canonical Ltd
+# Copyright (C) 2015 Canonical Ltd
```
kyrofa: This should probably still be 2016, no?

sergiusens: @kyrofa true, it should be only 2016 though as it is completely new

kyrofa:
Yeah this looks great :+1:

PR #472, opened 2016-04-18
Macaroons

vilagithub:
Add macaroon support to login, register and upload and download.

Replace mocked tests with tests against the staging servers under snapcraft/tests/store_tests. This proved to speed up development and integration between snapcraft and the sca/sso/cpi/updown servers.

Running against staging servers is opt-in and governed by setting the TEST_USER_PASSWORD. Responses are recorded and can then be used without requiring access to the servers (this address the coverage issues in the previous iterations).

Package available in https://code.launchpad.net/~vila/+archive/ubuntu/ppa/+packages?field.name_filter=snapcraft&field.status_filter=published&field.series_filter= to help test/review. I'll try to keep it up to date with the content of this PR as much as possible (i.e. clicking teh recipe button ;)


snappy-m-o:

snappy-m-o:

vilagithub:
regarding code coverage, I don't know why the tests are skipped but I can reproduce locally:

snapcraft/storeapi/__init__                    150     62     44     23    56%
snapcraft/storeapi/_upload                     104     78     22     21    21%
snapcraft/_store                                68     55     18     18    15%


when the tests are not skipped though:

snapcraft/storeapi/__init__                    150     25     44     11    81%
snapcraft/storeapi/_upload                     104      6     22      5    91%
snapcraft/_store                                68     55     18     18    15%


So snapcraft/_store receives the same score which is expected since I replaced mocked tests with tests exercising the real code.

Now integration_tests without the password gives:

snapcraft/_store                                68     68     18     18     0%
snapcraft/storeapi/__init__                    150    150     44     44     0%
snapcraft/storeapi/_upload                     104    104     22     22     0%

meh, that's bogus, fixed with:

--- a/bin/snapcraft-coverage
+++ b/bin/snapcraft-coverage

@@ -37,7 +37,6 @@ if __name__ == '__main__':
     try:
         snapcraft.internal.dirs.setup_dirs()
         snapcraft.main.main()
-    except:
+    finally:
         cov.stop()
         cov.save()
-        raise

Without this patch, since all command succeed without raising an exception, the coverage is /never/ collected.

snapcraft/_store                                68     44     18     16    30%
snapcraft/storeapi/__init__                    150    135     44     41     9%
snapcraft/storeapi/_upload                     104    104     22     22     0%

and with the password:

snapcraft/_store                                68     18     18      9    69%
snapcraft/storeapi/__init__                    150     86     44     26    42%
snapcraft/storeapi/_upload                     104     40     22     10    60%

vilagithub:
So, combining unit and integration tests locally:
snapcraft/_store                                68      5     18      9    84%
snapcraft/storeapi/__init__                    150     25     44     11    81%
snapcraft/storeapi/_upload                     104      9     22      8    87%

vilagithub:
Hmpf, now that integration tests doesn't do coverage anymore... coverage stats... doesn't reflect reality.

sergiusens:
@vilagithub hey, I thought @elopio talked to you; we were on an adventure since a month or two ago to not mingle integration test coverage with unittests. When it comes to coverage, we really only care about unit tests coverage.

snapcraft/storeapi/__init__.py
```diff
@@ -13,9 +13,285 @@
...
 from .info import get_info  # noqa
-from ._login import login  # noqa
-from ._download import download  # noqa
-from ._upload import upload  # noqa
+
+
+logger = logging.getLogger(__name__)
+
+
+class InvalidCredentials(Exception):
```
elopio: according to pep8, exceptions should have the "Error" suffix.

snapcraft/storeapi/__init__.py
```diff
@@ -13,9 +13,285 @@
...
+    pass
+
+
+class StoreError(Exception):
```
elopio: according to pep8, exceptions should have the "Error" suffix.

snapcraft/storeapi/__init__.py
```diff
@@ -13,9 +13,285 @@
...
+
+    def __init__(self, **kwargs):
+        for key, value in kwargs.items():
+            setattr(self, key, value)
+
+    def __str__(self):
+        return self.fmt.format([], **self.__dict__)
+
+
+class SnapNotFound(StoreError):
```
elopio: according to pep8, exceptions should have the "Error" suffix.

snapcraft/storeapi/__init__.py
```diff
@@ -13,9 +13,285 @@
...
+
+    fmt = 'The  "{name}" for {arch} was not found in {channel}.'
+
+    def __init__(self, name, channel, arch):
+        super().__init__(name=name, channel=channel, arch=arch)
+
+
+class SHAMismatch(StoreError):
```
elopio: according to pep8, exceptions should have the "Error" suffix.

snapcraft/storeapi/__init__.py
```diff
@@ -13,9 +13,285 @@
...
+        file_sum = hashlib.sha512()
+        with open(path, 'rb') as f:
+            for file_chunk in iter(
+                    lambda: f.read(file_sum.block_size * 128), b''):
+                file_sum.update(file_chunk)
+        return expected_sha512 == file_sum.hexdigest()
+
+    # Low level helpers
+
+    def post(self, path, data, headers=None):
```
elopio: Please prepend the internal helpers with _.
s/post/_post

snapcraft/storeapi/__init__.py
```diff
@@ -13,9 +13,285 @@
...
+
+class SHAMismatch(StoreError):
+
+    fmt = 'SHA512 checksum for {path} is not {expected_sha}.'
+
+    def __init__(self, path, expected_sha):
+        super().__init__(path=path, expected_sha=expected_sha)
+
+
+def macaroon_auth(conf, acl):
```
elopio: This one is only used in this file. Please mark it as such with the _ in front.

snapcraft/storeapi/__init__.py
vilagithub: Fixed.

snapcraft/storeapi/__init__.py
vilagithub: Fixed.

snapcraft/storeapi/__init__.py
vilagithub: I started to do so, and got confused half way: requests.Session define post and get, ssoclient.V2ApiClient defines post and get, every session wrapper defines post and.

Changing that rule just for the session wrappers defined here makes the code a constant puzzle: should I use get or _get ? _post ? Is that different from post ?

Otherwise, what's your definition of an internal helper as opposed to a "private" (even in the python lose sense) helper ?

snapcraft/storeapi/__init__.py
vilagithub: Fixed.

But again what's the rule ?

elopio: If the func is used only in this file, we start prepending the _. That helps figuring out which is the right entry point for a test, and gives us a lot of freedom when we are refactoring stuff. Once we need to access the func from outside, we make sure that it's clean enough for us to support it for the foreseeable future. So we delay that discussion about how ready is the code for long term consumption until there's an external consumer.

integration_tests/__init__.py
```diff
@@ -104,3 +106,10 @@ def logout(self):
         expected = ('Clearing credentials for Ubuntu One SSO.\n'
                     'Credentials cleared.\n')
         self.assertEqual(expected, output)
+
+    def register_name(self, name):
+        process = pexpect.spawn(self.snapcraft_command,
+                                ['register-name', name])
```
elopio: Do we need to call the command register-name? Wouldn't it be clear enough to call it just register?

integration_tests/__init__.py
```diff
@@ -78,8 +78,10 @@ def get_output_ignoring_non_zero_exit(self, binary, cwd):
             output = exception.output
         return output
 
-    def login(self, email='u1test+snapcraft@canonical.com',
+    def login(self, email=None,
               password=None, expect_success=True):
+        email = email or os.getenv('TEST_USER_EMAIL',
+                                   'u1test+snapcraft@canonical.com')
```
elopio: thanks!

integration_tests/test_register.py
```diff
@@ -0,0 +1,51 @@
...
+class RegisterTestCase(integration_tests.TestCase):
+
+    def setUp(self):
+        super().setUp()
+        self.useFixture(fixture_setup.StagingStore())
+
+    # FIXME: The store doesn't provide a way to unregister a name *and*
+    # registrations are rate-limited for a given user. We work around that by
+    # creating unique names and assuming we only run against staging or local
+    # dev instances -- vila 2016-04-08
```
elopio: Is there a bug reported to allow un-registering in the store?

snapcraft/config.py
```diff
@@ -14,68 +14,75 @@
...
-    location = urlparse(api_endpoint).netloc
-    parser.remove_section(location)
 
-    with open(filename, 'w') as fd:
-        parser.write(fd)
+class Config(object):
+    """Hold configuration options in sections.
+
+    There can be two sections for the sso related credentials: production and
+    staging. This is gouverned by the UBUNTU_SSO_API_ROOT_URL environment
```
elopio: typo: s/gouverned/governed

snapcraft/storeapi/__init__.py
```diff
@@ -13,9 +13,282 @@
...
+        auth = _macaroon_auth(self.conf, 'package_upload')
+        response = self.post('register-name/',
+                             data=json.dumps(data),
+                             headers={'Authorization': auth,
+                                      'Content-Type': 'application/json'})
+        if not response.ok:
+            # if (response['errors'] == ['Authorization Required']
+            #     and (response.headers['WWW-Authenticate']
+            #          == "Macaroon needs_refresh=1":
+            # Refresh the discharge macaroon and retry
```
elopio: Should this be a bug instead of commented code?

snapcraft/storeapi/__init__.py
```diff
@@ -13,9 +13,282 @@
...
+        self.cpi = CPIClient(self.conf)
+        logger.info('Getting details for {}'.format(snap_name))
+        package = self.cpi.search_package(snap_name, channel, arch)
+        if package is None:
+            raise SnapNotFoundError(snap_name, channel, arch)
+        return self.download_snap(snap_name, channel, arch,
+                                  download_path, package['download_url'],
+                                  package['download_sha512'])
+
+    def get_macaroon(self, acl):
```
elopio: This one is internal, right? _get_macaroon if so.

snapcraft/storeapi/__init__.py
```diff
@@ -13,9 +13,282 @@
...
+        response = self.post('../../api/2.0/acl/{}/'.format(acl),
+                             data=json.dumps({}),
+                             headers={'Content-Type': 'application/json'})
+        if response.ok:
+            return response.json()['macaroon'], None
+        else:
+            return None, response.text
+
+    def get_discharges(self, email, password, one_time_password, macaroons):
```
elopio: _get_discharges

snapcraft/storeapi/__init__.py
```diff
@@ -13,9 +13,282 @@
...
+        try:
+            response = self.sso.session.post(
+                '/tokens/discharge', data=data,
+                headers={'Content-Type': 'application/json',
+                         'Accept': 'application/json'})
+            return dict(response.content['discharge_macaroons']), None
+        except sso.ApiException as err:
+            return None, err.body
+        except sso.UnexpectedApiError as err:
+            return None, err.json_body
```
elopio: This looks like go. I'm liking it a lot in go, but in python should we raise the exception instead? It's more common to see the caller doing try catch instead of if error.

snapcraft/storeapi/__init__.py
```diff
@@ -13,9 +13,282 @@
...
+            # FIXME: Cough, we may want to buffer here (and a progress bar
+            # would be nice) -- vila 2016-04-26
+            f.write(download.content)
+        if self.matching_sha512(download_path, expected_sha512):
+            logger.info('Successfully downloaded {} at {}'.format(
+                name, download_path))
+        else:
+            raise SHAMismatchError(download_path, expected_sha512)
+
+    def matching_sha512(self, path, expected_sha512):
```
elopio: _matching_sha512

snapcraft/storeapi/__init__.py
```diff
@@ -13,9 +13,282 @@
...
+        if self.session is not None:
+            self.session.close()
+        if self.sso.session is not None:
+            self.sso.session.close()
+        if self.updown is not None:
+            self.updown.close()
+
+
+# FIXME: Refactor as a base for all clients using requests.Session() (sca,
+# updown, cpi) with (conf, root_url) to build from -- vila 2016-04-27
```
elopio: Please report the bug so we add it to one of our next sprints.

snapcraft/storeapi/__init__.py
elopio: I'd still like our clients to handle a higher level api. _get, _post, search, and download.
Our clients encapsulate request.Session's methods, and offer no direct way to do a different request.
But I don't have a strong opinion with the get and post methods. Just an opinion. With many of the other methods that are used only inside the client class, I do have a strong opinion about marking them internal.

elopio: about internal versus private, internal has one underscore, private has two. I dislike the private mangling, so I never use it, and I think one underscore conveys all the information: it might change so don't use it outside of this file. Feel free to disagree, we can define a different convention for the next version of the API.

snapcraft/storeapi/__init__.py
```diff
@@ -13,9 +13,282 @@
...
+                '/tokens/discharge', data=data,
+                headers={'Content-Type': 'application/json',
+                         'Accept': 'application/json'})
+            return dict(response.content['discharge_macaroons']), None
+        except sso.ApiException as err:
+            return None, err.body
+        except sso.UnexpectedApiError as err:
+            return None, err.json_body
+
+    def upload_snap(self, name, data):
```
elopio: I think upload, upload_snap, download and download_snap need docstrings.

snapcraft/storeapi/__init__.py
```diff
@@ -13,9 +13,282 @@
...
+        data['name'] = name
+        headers = {'Authorization':
+                   _macaroon_auth(self.conf, 'package_upload')}
+        response = self.post('snap-upload/', data=data, headers=headers)
+        return response
+
+    def download_snap(self, name, channel, arch, download_path,
```
elopio: This looks internal too. So _download_snap.

snapcraft/storeapi/__init__.py
```diff
@@ -13,9 +13,282 @@
...
+
+    def __init__(self, conf):
+        self.conf = conf
+        self.root_url = os.environ.get('UBUNTU_STORE_SEARCH_ROOT_URL',
+                                       UBUNTU_STORE_SEARCH_ROOT_URL)
+        if self.conf.get('package_access') is None:
+            raise InvalidCredentialsError()
+        self.session = requests.Session()
+
+    def search_package(self, snap_name, channel, arch):
```
elopio: what about search instead of search package?
There's a long aweful discussion about how to call this things? snaps, apps, packages... We can just remove the word in here.

snapcraft/storeapi/_upload.py
```diff
@@ -45,52 +41,7 @@ def _update_progress_bar(progress_bar, maximum_value, monitor):
...
-
-    if app_url:
-        logger.info('Please check out the application at: %s\n',
-                    app_url)
-
-    return success
-
-
-def upload_files(binary_filename, config=None):
+def upload_files(binary_filename, session):
```
elopio: I was wondering some weeks ago why is this called upload_files instead of upload_file.
@sergiusens would you consider that an api breakage?

snapcraft/storeapi/common.py
```diff
@@ -19,45 +19,41 @@
...
-        )
-    except KeyError:
-        session = None
-    return session
-
-
-def store_api_call(path, session=None, method='GET', data=None):
-    """Issue a request for a particular endpoint of the MyApps API."""
-    result = {'success': False, 'errors': [], 'data': None}
+def store_raw_api_call(path, session=None, method='GET', data=None,
```
elopio: _store_raw_api_call

snapcraft/storeapi/__init__.py
```diff
@@ -13,9 +13,282 @@
...
+        for key, value in kwargs.items():
+            setattr(self, key, value)
+
+    def __str__(self):
+        return self.fmt.format([], **self.__dict__)
+
+
+class SnapNotFoundError(StoreError):
+
+    fmt = 'The  "{name}" for {arch} was not found in {channel}.'
```
elopio: There's a bug in here, and this line has no test. ;)
There's an extra space before name.

I'm sure I commented this yesterday, but it seems to have disappeared. Forgive me if you get it twice.

integration_tests/__init__.py
vilagithub: This has been defined by Martin and Mark, I wouldn't change it without feedback.
I think the rationale is that /names/ matters in the snappy world, they are defined by series and can change owner from one series to the others, there can be disputes, revocations and so on.
So, register-name carry that, a bare register will be less clear. But that's just my understanding.

integration_tests/__init__.py
vilagithub: De nada but what are you thanking me for ? Using the same name default account for tests ?

integration_tests/test_register.py
vilagithub: No. And I don't think there will be one implemented in the foreseeable future (see above for some complexity associated with snap names).
I put this FIXME there to explain why the issue is tricky from a test pov, this is the boundary blocking other type of tests by having to use a staging server.
The long term plan would be to have more tests around register-name against a dev server where we can more easily use scratch DBs and don't have to care about cleaning nor having to use shared resources with uuid which a pita for test writers.

But until we get there, tests against staging cover most of our needs.

snapcraft/config.py
vilagithub: Ha thanks, this was refactored from existing code.
Fixed.

snapcraft/storeapi/__init__.py
vilagithub: This is tricky ;-)
As mentioned outside this MP, there are other pieces to complete the macaroons implementation (mostly publish and here, macaroon refresh).
I did my best to limit this MP to the minimal set of features allowing snapcraft to switch from OAuth to macaroons.
Sooo, it's hard to file bugs until this MP has landed since they would refer to stuff that is not yet part of the project.
On that topic, I have the same dilemma about two issues I fixed for the integration coverage since they are against code that is not part of the project anymore...

snapcraft/storeapi/__init__.py
vilagithub: Right.
I think there is a tension here around where the private start and where it ends.
snapcraft API with the store is defined in '_store.py' there is a leading underscore there.
If storeapi is seen as a library, there are other use cases where get_macaroon is public.
Can we, until things become clearer, consider that '_store.py' carries the internal semantic ?

snapcraft/storeapi/__init__.py
vilagithub: This one is even more likely to be used elsewhere.

snapcraft/storeapi/__init__.py
vilagithub: Full agreement.
I've started moving away from that pattern in this MP but I'm not there yet. Can we postpone this ?

snapcraft/storeapi/__init__.py
vilagithub: See above ;)

snapcraft/storeapi/__init__.py
vilagithub: Yes.
This is also something I'm planning to do at the module, class and method levels once the refactoring settle.
The internal layering has received a fair amount of shaking already and will receive some more as the missing pieces are added.
I've tried to use names matching the server side to make discussions with the server devs easier, there are still a few remaining changes being implemented there so I'd rather postpone that documenting step (it's on my TODO so it won't be lost).

snapcraft/storeapi/__init__.py
vilagithub: See above, indeed it's internal behind _store. It's public in storeapi though.

snapcraft/storeapi/__init__.py
vilagithub: Yes and no ;)
At this level it's package for CPI but it should become snap pretty soon.
(note search_package but snap_name in the parameters).

snapcraft/storeapi/_upload.py
vilagithub: Hehe, this comes from clicks and doesn't really make sense anymore.
I refrained from changing them though as I tried to minimize the changes.
This introduces the biggest changes, but more followup MPs can refine (I'm thinking about the simplification and bugfix for the progress bar, again a bug hard to file at this point but I already have a fix in a WIP).

snapcraft/storeapi/common.py
vilagithub: This will die soon.

snapcraft/storeapi/__init__.py
vilagithub: Extra space fixed.
There is a test (with a bug in its name, fixed:  snapcraft.tests.store_tests.test_download.TestDownload_store.test_download_not_found ;) but that string is not exposed to the user as the exception is caught and re-raised as RuntimeError in _store as seems to be the practice in snapcraft.

integration_tests/__init__.py
elopio: For de-hardcoding the email.

snapcraft/storeapi/__init__.py
elopio: even on files with _something.py, I find it really useful to mark the functions that are only called from inside that file with a _. I prefer to use _func until it's clear that it will be needed outside, not the other way around.

snapcraft/storeapi/__init__.py
elopio: when it's used elsewhere, it's simple to remove the _.

snapcraft/tests/store_tests/test_download.py
```diff
@@ -0,0 +1,150 @@
...
+        self.assertRaises(storeapi.InvalidCredentialsError,
+                          self.download, 'basic')
+
+
+class TestSearchPackage(store_tests.TestCase):
+
+    def setUp(self):
+        super().setUp()
+        self.addCleanup(self.logout)
+        self.login()
```
elopio: We have anonymous searches, right?

snapcraft/storeapi/__init__.py
```diff
@@ -13,9 +13,282 @@
...
+        logger.info('Getting details for {}'.format(snap_name))
+        package = self.cpi.search_package(snap_name, channel, arch)
+        if package is None:
+            raise SnapNotFoundError(snap_name, channel, arch)
+        return self.download_snap(snap_name, channel, arch,
+                                  download_path, package['download_url'],
+                                  package['download_sha512'])
+
+    def get_macaroon(self, acl):
+        response = self.post('../../api/2.0/acl/{}/'.format(acl),
```
elopio: I find it weird that you are using the ../..
could you put a comment here?

snapcraft/tests/store_tests/test_download.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def assertLog(self, expected):
+        self.assertEqual(expected, self.logger.output)
+
+    def test_download_unknwon_package(self):
```
elopio: typo: unknwon

snapcraft/storeapi/__init__.py
```diff
@@ -13,9 +13,282 @@
...
+            return None
+        else:
+            return embedded['clickindex:package'][0]
+
+    def get(self, path, headers=None, params=None):
+        if headers is None:
+            headers = {}
+        headers.update({'Authorization':
+                        _macaroon_auth(self.conf, 'package_access')})
+        url = parse.urljoin(self.root_url, path)
```
elopio: This is weird. You are calling get twice, one for the search where this parse.urljoin is clear. But also for the download, where you have path=https://public.apps.staging.ubuntu.com/download-snap/{...}.snap and root_url=https://search.apps.staging.ubuntu.com/. This works just because urljoin returns the path if it can't join them, but it's not clear at all.
I think you should have two methods.

snapcraft/tests/store_tests/test_download.py
```diff
@@ -0,0 +1,150 @@
...
+            _store.download, 'gloo', 'bee', 'downloaded.snap', 'amd64')
+        self.assertEqual(
+            'Snap gloo for amd64 cannot be found in the bee channel',
+            str(exc))
+
+    def test_download_mismatch(self):
+        self.login()
+        self.addCleanup(self.logout)
+        self.addCleanup(setattr, storeapi.SCAClient, 'download',
+                        storeapi.SCAClient.download)
```
elopio: I would do this with python.mock.patch, of course ;)

snapcraft/storeapi/constants.py
```diff
@@ -16,8 +16,12 @@
 from __future__ import absolute_import, unicode_literals
 
 
+# FIXME: snapcraft targets the '16' series, hardcode it until more choices
+# become available server side -- vila 2016-04-22
+DEFAULT_SERIES = '16'
 SCAN_STATUS_POLL_DELAY = 5
 SCAN_STATUS_POLL_RETRIES = 5
 UBUNTU_SSO_API_ROOT_URL = 'https://login.ubuntu.com/api/v2/'
 UBUNTU_STORE_API_ROOT_URL = 'https://myapps.developer.ubuntu.com/dev/api/'
```
elopio: does this dev/api mean "development unstable API"?

integration_tests/__init__.py
vilagithub: Ha right.

This is a first step to explicit the assumptions made in the test suite against the staging server *content*.
Things like:
- those tests can't be run against production,
- the user must have registered the 'aseemble' and 'basic' names,
- the user must have register the 'femto' name and have published one upload
- may be others
Which are not currently documented.

snapcraft/storeapi/__init__.py
vilagithub: Nope, it's part of the design to allow such uses, the download url can (and is) redirected to a CDN on a different host and requests.Session provides support for that.

snapcraft/storeapi/constants.py
vilagithub: Care to elaborate on that ?

snapcraft/tests/store_tests/test_download.py
vilagithub: AIUI, anonymous searches is a workaround until macaroons are supported, so starting with the MP, no more anonymous searches. This specific search is for download anyway which was previously implemented with OAuth and as such was already requiring login.

vilagithub:
> Coverage decreased (-3.6%) to 92.584% 

as mentioned elsewhere this is caused by running with TEST_USER_PASSWORD set and not taking integration coverage into account.

Fixing the bugs in integration coverage as in https://github.com/vilagithub/snapcraft/tree/coverage and getting coverage for both unit and integration tests this branch says:

Name                                         Stmts   Miss Branch BrMiss  Cover

<...>
snapcraft/_store                                78      0     18      1    99%
<...>
snapcraft/storeapi/__init__                    166      3     45      2    98%
snapcraft/storeapi/_upload                     104      1     22      3    97%
snapcraft/storeapi/common                       57      0     24      1    99%
snapcraft/storeapi/compat                        1      0      0      0   100%
snapcraft/storeapi/constants                     8      0      0      0   100%
snapcraft/storeapi/info                          4      0      0      0   100%
<...>
TOTAL                                         9140    199   1454     97    97%

snapcraft/storeapi/constants.py
elopio: I'm wondering why does this url have the word "dev", and why it doesn't include a version.

vilagithub: dev stands for devportal, I think it goes back to the days where you were in the u1 team and I wasn't. I don't know why the API is not versioned. I would guess because, whether or not it's versioned, it's supported and versioning other APIs didn't prove to be that useful to maintain backwards compatibility.

snapcraft/tests/store_tests/test_download.py
vilagithub: fixed

integration_tests/__init__.py
kyrofa: Note that during the sprint this changed to simply "register."

snappy-m-o:

snappy-m-o:

integration_tests/__init__.py
vilagithub: Fixed.

sergiusens:
thank you for your original work on this, I am closing it as we have finished the landing from multiple small pieces. Feel free to take a peek and comment here or there!

Kudos.

PR #473, opened 2016-04-19
Fix shell quoting

sergiusens:
"$@" is the proper way to expand original positional args; $* will
perform an extra IFS split.

LP: #1572129

Signed-off-by: Loïc Minier <loic.minier@ubuntu.com>


zyga:
Looks good, thank you!

kyrofa:
Looks good to me :+1:

PR #480, opened 2016-04-21
Make upload more robust by ignoring suprious errors while polling the…

vilagithub:
… scan status.

While the snap is scanned on the store, the status is polled until completion.

If an error occurs while acquiring the status, ignore the errors.

Those errors are rare and the next retry generally succeeds so ignoring those errors is harmless.

If they persist until the last retry, something else may be going on and this will still be reported.

LP: #1572963

### DIFF
--- a/snapcraft/storeapi/_upload.py
+++ b/snapcraft/storeapi/_upload.py
@@ -20,13 +20,13 @@ import functools
 import time
 import os
 
+import requests
 from concurrent.futures import ThreadPoolExecutor
 from progressbar import (ProgressBar, Percentage, Bar, AnimatedMarker)
 from requests_toolbelt import (MultipartEncoder, MultipartEncoderMonitor)
 
 from .common import (
     get_oauth_session,
-    is_scan_completed,
     retry,
 )
 from .compat import open, quote_plus, urljoin
@@ -318,6 +318,30 @@ def get_post_files(metadata=None):
     return files
 
 
+def is_scan_completed(response):
+    """Return True if the response indicates the scan process completed."""
+    if response is None:
+        # To cope with spurious connection failures lacking a proper response:
+        # either we'll retry and succeed or we fail for all retries and report
+        # an error.
+        return False
+    if response.ok:
+        return response.json().get('completed', False)
+    return False
+
+
+def get_scan_status(session, url):
+    try:
+        resp = session.get(url)
+        return resp
+    except (requests.ConnectionError, requests.HTTPError):
+        # Something went wrong and we couldn't acquire the status. Upper
+        # level (is_scan_completed) will deal with the None response
+        # meaning we don't know the status. This avoid a spurious
+        # connection error breaking an upload for a wrong reason.
+        return None
+
+
 def get_scan_data(session, status_url):
     """Return metadata about the state of the upload scan process."""
     # initial retry after 5 seconds
@@ -328,7 +352,7 @@ def get_scan_data(session, status_url):
            delay=SCAN_STATUS_POLL_DELAY,
            backoff=1)
     def get_status():
-        return session.get(status_url)
+        return get_scan_status(session, status_url)
 
     response, aborted = get_status()
 
--- a/snapcraft/storeapi/common.py
+++ b/snapcraft/storeapi/common.py
@@ -67,13 +67,6 @@ def store_api_call(path, session=None, method='GET', data=None):
     return result
 
 
-def is_scan_completed(response):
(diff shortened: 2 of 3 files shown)


snappy-m-o:

snappy-m-o:

snapcraft/storeapi/_upload.py
```diff
@@ -318,6 +318,30 @@ def get_post_files(metadata=None):
...
+    if response.ok:
+        return response.json().get('completed', False)
+    return False
+
+
+def get_scan_status(session, url):
+    try:
+        resp = session.get(url)
+        return resp
+    except (requests.ConnectionError, requests.HTTPError):
```
sergiusens: What was HTTPError, will we loop forever if the resource does not exist?

vilagithub: > What was HTTPError

I don't remember encountering it in this particular case but the code involved can trigger it. I can remove it if you prefer.

> will we loop forever if the resource does not exist?

No. The caller retries a fixed number of times and will report the last error.

snapcraft/storeapi/_upload.py
```diff
@@ -318,6 +318,30 @@ def get_post_files(metadata=None):
...
+        # Something went wrong and we couldn't acquire the status. Upper
+        # level (is_scan_completed) will deal with the None response
+        # meaning we don't know the status. This avoid a spurious
+        # connection error breaking an upload for a wrong reason.
+        return None
```
elopio: According to [1], we should return an exception instead of None. That makes sense because the name of the exception will make it's meaning clearer than None.

[1] https://www.goodreads.com/book/show/23020812-effective-python

snapcraft/storeapi/_upload.py
```diff
@@ -318,6 +318,30 @@ def get_post_files(metadata=None):
...
+        # To cope with spurious connection failures lacking a proper response:
+        # either we'll retry and succeed or we fail for all retries and report
+        # an error.
+        return False
+    if response.ok:
+        return response.json().get('completed', False)
+    return False
+
+
+def get_scan_status(session, url):
```
elopio: I think I would name these helper funcions with a leading underscore, to make it clear that they are called only from higher level functions.

elopio:
this looks good to me. It's good that you moved the func out of common, because it was called only once.
I would prefer not to return None to signal the problem.

snapcraft/storeapi/_upload.py
vilagithub: > we should return an exception instead of None.

No, that's the bug. The caller (intermediate, hidden, out of our reach) can't catch it nor act properly. Returning None tells the caller: nah, not completed yet.

snapcraft/storeapi/_upload.py
vilagithub: There is a refactor in progress in another branch, I'd rather address that there ?

elopio: works for me.

snapcraft/storeapi/_upload.py
elopio: I see, @retry doesn't let you handle the exception. The comments make this clear, so I'm ok with this.

elopio:
retest this please

vilagithub:
From http://162.213.35.179:8080/job/github-snapcraft-autopkgtest-cloud/597/console
adt-run [16:32:03]: version 3.20.3
...
adt-run [17:02:03]: ERROR: testbed failure: cannot send to testbed: ['BrokenPipeError: [Errno 32] Broken pipe\n']

So 30 minutes hanging on the instance creation. Re-test ?
3.20.4 is out for xenial but I don't think that will fix that issue.

From http://162.213.35.179:8080/job/github-snapcraft-examples-tests-cloud/697/console

2016/04/27 16:31:55 *** Creating a Snappy m1.medium instance for release rolling, channel edge and image type custom ***

goroutine 1 [running]:
log.Panicf(0x5b1d20, 0x36, 0xc820039e58, 0x3, 0x3)
	/usr/lib/go/src/log/log.go:327 +0xd8
main.main()
	/build/snappy-tests-job-LsZciy/snappy-tests-job-1.0.0ubuntu1/obj-x86_64-linux-gnu/src/launchpad.net/snappy-tests-job/cmd/snappy-cloud-client/main.go:68 +0xb93

sergiusens:
retest this

elopio:
OK to test

kyrofa:
Looks good to me :+1:

PR #485, opened 2016-04-22
Examples tests: use systemctl instead of the removed snap service

elopio:


elopio:
Requires #484

examples_tests/__init__.py
```diff
@@ -178,9 +181,7 @@ def run_command_in_snappy_testbed(self, command):
...
         if not config.get('skip-install', False):
             output = self.run_command_in_snappy_testbed(
-                ['sudo', 'snap', 'service', 'status', snap])
-            expected = (
-                'Snap\t+Service\t+State\n'
-                '{}\t+{}\t+enabled; loaded; active \(running\)\n'.format(
-                    snap, service))
-            self.assertThat(output, MatchesRegex(expected))
+                ['systemctl', '--no-pager', 'status',
+                 'snap.{}.{}'.format(snap, service)])
```
kyrofa: Huh, this file format must have changed in 2.0, eh?

elopio:
retest this please

examples_tests/__init__.py
elopio: This is the unit name, and yes, it changed. I don't remember how was it named before; but that shouldn't matter, all of these details should be hidden behind a nice snap command. snap service was one of the commands I liked the most, I hope they bring it back.

kyrofa: Yeah, I just hope it gains the ability to access specific services within a snap.

kyrofa:
I'm not a systemd expert, but this looks good to me :+1: .

sergiusens:
@elopio please create a bug with tasks for `snapcraft` and `snappy` so we can be in sync for when it comes back

PR #487, opened 2016-04-25
Update the busybox test to use the snap data path.

elopio:
LP: #1574901

### DIFF
--- a/examples_tests/test_busybox.py
+++ b/examples_tests/test_busybox.py
@@ -25,7 +25,8 @@ class BusyBoxTestCase(examples_tests.ExampleTestCase):
         self.build_snap(self.example_dir)
         self.install_snap(self.example_dir, 'busybox', '1.0')
         # Run the binary once to create the data dir.
-        self.run_command_in_snappy_testbed('/snap/bin/busybox.touch')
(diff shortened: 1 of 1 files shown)


examples_tests/test_busybox.py
```diff
@@ -25,9 +25,12 @@ def test_busybox(self):
         self.build_snap(self.example_dir)
         self.install_snap(self.example_dir, 'busybox', '1.0')
         self.assert_command_in_snappy_testbed(
-            ['/snap/bin/busybox.touch', 'busybox.test'], '')
+            ['/snap/bin/busybox.touch',
+             'snap/busybox/$(ls snap/busybox/)/busybox.test'],
```
sergiusens: this is so weird

sergiusens: grep one entry? Should this be `$(ls ~/snap/busybox)`? Maybe quoted

elopio: I agree that it's weird, but it was the shortest way I found to get the dir.
The ~ doesn't hurt. I added that.
I think that quotes are not needed because of how python passes the args, but I don't know that much about bash to understand all the details.

zyga:
Looks good

PR #496, opened 2016-05-03
Don't fail if there is no system library list

sergiusens:
LP: #1577750

### DIFF
--- a/snapcraft/internal/libraries.py
+++ b/snapcraft/internal/libraries.py
@@ -71,7 +71,8 @@ def _get_system_libs():
     lib_path = os.path.join(common.get_librariesdir(), release)
 
     if not os.path.exists(lib_path):
-        logger.debug('No libraries to exclude from this release')
+        logger.warning('No libraries to exclude from this release')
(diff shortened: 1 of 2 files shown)


kyrofa:
This looks good to me, though I'm trying to decide the use-case for this. Specifically, if someone is building a snap targeting a release that doesn't exist, it's not going to work, right? Why plow ahead and build that snap instead of erroring out? Maybe my complaint is more with the bug: it doesn't really describe why what's happening right now is a problem.

kyrofa:
Oh, because it's coming from the running system, I see.

sergiusens:
@kyrofa if it is of any use although you seem to have figured it out https://objectstorage.prodstack4-5.canonical.com/v1/AUTH_77e2ada1e7a84929a74ba3b87153c0ac/autopkgtest-yakkety/yakkety/amd64/s/snapcraft/20160503_064449@/log.gz :-)

kyrofa:
Alright, combined with discussion on IRC, :+1: .

PR #497, opened 2016-05-03
Catkin plugin: Enforce C.UTF-8 locale.

kyrofa:
ROS apps run fine on Ubuntu Core, but throw boost locale errors on Xenial desktop. This is due to LP: [#1576411](https://bugs.launchpad.net/ubuntu/+source/snapd/+bug/1576411), where the desktop locales aren't available to snaps even though the locale environment is set for the snaps. This PR is a quick and dirty fix for LP: [#1577557](https://bugs.launchpad.net/snapcraft/+bug/1577557) that simply sets `LC_ALL=C.UTF-8` for snaps with Catkin parts. This PR can be reverted once LP: [#1576411](https://bugs.launchpad.net/ubuntu/+source/snapd/+bug/1576411) is actually fixed.


sergiusens:
Question, if a part, another part adds support to the snap for locales, will we be able to get out of the C locale?

kyrofa:
Good question... I guess it depends on who sets the environment last huh? So yeah, other parts that set a locale can bust ROS, and Catkin parts can bust other parts that want a locale. What's the best way forward here?

snapcraft/plugins/catkin.py
```diff
@@ -139,6 +139,11 @@ def env(self, root):
             # and use $ROS_HOME to determine where to put them.
             'ROS_HOME=$SNAP_USER_DATA/ros',
 
+            # ROS seems to bust with locale errors if used on anything other
+            # than C or C.UTF-8, so we'll set that here for a consistent
+            # experience across Ubuntu Core and Desktop environments.
```
elopio: is there a ROS bug for this?

elopio:
seems ugly, but the comment states that already. We are not dealing with locales yet, and we can come back to this change in case of problem, So if this makes it work, I'm ok with the change.

snapcraft/plugins/catkin.py
kyrofa: A [boost one](https://svn.boost.org/trac/boost/ticket/4688) actually. They claim it should be fixed in a newer version, but of course, that doesn't help us here.

kyrofa:
I've re-opened this PR. The bug which is the root cause of this issue (LP: [#1576411](https://bugs.launchpad.net/ubuntu/+source/snapd/+bug/1576411)) has sat idle for some time, and we're now [getting questions](https://askubuntu.com/questions/796021/ros-with-snappy) from people hitting this, so we need to work around it until the bug is fixed in snapd.

kyrofa:
retest this please

kyrofa:
Thanks for babysitting @sergiusens :smile: .

PR #499, opened 2016-05-04
Create the 'partial' directory for apt.

sergiusens:
LP: #1578007


sergiusens:
@mvo5 fwiw this is the solution I found to the problem, it seems 'partial' isn't created for us anymore.

sergiusens:
Works on `yakkety` (tested in a container), also works on `xenial`

kyrofa:
Blech. But :+1: .

sergiusens:
Merging this for now even though it would be an `apt` bug to unblock us.

PR #501, opened 2016-05-12
Add support for confinement property.

kyrofa:
Snaps need to be able to specify if they require devmode or if they can be run confined. This will allow for snapd to provide reasonable errors if one tries to install a snap that cannot run successfully under
confinement.

This PR resolves LP: [#1580819](https://bugs.launchpad.net/snapcraft/+bug/1580819) by adding the "confinement" property, which must be one of "strict" or "devmode". This property is optional, and defaults to "strict" if unspecified (to maintain backward-compatibility with current behavior). However, `snapcraft init` generates a YAML that sets "confinement" to "devmode" since we expect most snaps to start out needing that.

### DIFF
--- a/snapcraft/_schema.py
+++ b/snapcraft/_schema.py
@@ -41,17 +41,25 @@ class Validator:
         self._load_schema()
 
     @property
+    def schema(self):
+        """Return all schema properties."""
+
+        return self._schema['properties'].copy()
+
+    @property
     def part_schema(self):
-        sub = self.schema['properties']['parts']['patternProperties']
+        """Return part-specific schema properties."""
+
+        sub = self.schema['parts']['patternProperties']
         properties = sub['^(?!plugins$)[a-z0-9][a-z0-9+-]*$']['properties']
-        return properties.copy()
+        return properties
 
     def _load_schema(self):
         schema_file = os.path.abspath(os.path.join(
             common.get_schemadir(), 'snapcraft.yaml'))
         try:
             with open(schema_file) as fp:
-                self.schema = yaml.load(fp)
+                self._schema = yaml.load(fp)
         except FileNotFoundError:
             raise SnapcraftSchemaError(
                 'snapcraft validation file is missing from installation path')
@@ -60,7 +68,7 @@ class Validator:
         format_check = jsonschema.FormatChecker()
         try:
             jsonschema.validate(
-                self._snapcraft, self.schema, format_checker=format_check)
+                self._snapcraft, self._schema, format_checker=format_check)
         except jsonschema.ValidationError as e:
             messages = [e.message]
             if e.path:
--- a/snapcraft/_schema.py
+++ b/snapcraft/_schema.py
@@ -42,7 +42,7 @@ class Validator:
 
     @property
     def part_schema(self):
-        sub = self._schema['properties']['parts']['patternProperties']
+        sub = self.schema['properties']['parts']['patternProperties']
         properties = sub['^(?!plugins$)[a-z0-9][a-z0-9+-]*$']['properties']
         return properties.copy()
 
@@ -51,7 +51,7 @@ class Validator:
             common.get_schemadir(), 'snapcraft.yaml'))
         try:
             with open(schema_file) as fp:
-                self._schema = yaml.load(fp)
+                self.schema = yaml.load(fp)
         except FileNotFoundError:
             raise SnapcraftSchemaError(
                 'snapcraft validation file is missing from installation path')
@@ -60,7 +60,7 @@ class Validator:
         format_check = jsonschema.FormatChecker()
         try:
             jsonschema.validate(
-                self._snapcraft, self._schema, format_checker=format_check)
+                self._snapcraft, self.schema, format_checker=format_check)
         except jsonschema.ValidationError as e:
             messages = [e.message]
             if e.path:
--- a/snapcraft/internal/lifecycle.py
+++ b/snapcraft/internal/lifecycle.py
@@ -41,7 +41,7 @@ _TEMPLATE_YAML = r'''name: # the name of the snap
 version: # the version of the snap
 summary: # 79 char long summary
 description: # a longer description for the snap
-confinement: devmode # devmode means no confinement is supported
+confinement: devmode # devmode means the snap doesn't support confinement
 '''
 
 _STEPS_TO_AUTOMATICALLY_CLEAN_IF_DIRTY = {'stage', 'strip'}
--- a/snapcraft/internal/lifecycle.py
+++ b/snapcraft/internal/lifecycle.py
@@ -37,12 +37,13 @@ from snapcraft.internal import (
 logger = logging.getLogger(__name__)
 
 
-_TEMPLATE_YAML = r'''name: # the name of the snap
+_TEMPLATE_YAML = """name: # the name of the snap
 version: # the version of the snap
 summary: # 79 char long summary
 description: # a longer description for the snap
-confinement: devmode # devmode means the snap doesn't support confinement
-'''
+confinement: devmode # use "strict" to enforce system access only via \
+declared interfaces
+"""
 
 _STEPS_TO_AUTOMATICALLY_CLEAN_IF_DIRTY = {'stage', 'strip'}
 
--- a/snapcraft/internal/lifecycle.py
+++ b/snapcraft/internal/lifecycle.py
@@ -40,7 +40,8 @@ logger = logging.getLogger(__name__)
 _TEMPLATE_YAML = r'''name: # the name of the snap
 version: # the version of the snap
 summary: # 79 char long summary
-description: # A longer description for the snap
+description: # a longer description for the snap
+confinement: devmode # devmode means no confinement is supported
 '''
 
 _STEPS_TO_AUTOMATICALLY_CLEAN_IF_DIRTY = {'stage', 'strip'}
--- a/snapcraft/tests/test_commands_init.py
+++ b/snapcraft/tests/test_commands_init.py
@@ -41,7 +41,7 @@ class InitCommandTestCase(tests.TestCase):
 version: # the version of the snap
 summary: # 79 char long summary
 description: # a longer description for the snap
-confinement: devmode # devmode means no confinement is supported"""
+confinement: devmode # devmode means the snap doesn't support confinement"""
 
         main(['init'])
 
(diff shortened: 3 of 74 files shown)


kyrofa:
retest this please.

sergiusens:
Very good. So this is not going to be a list of attributes and this one of the tags?

niemeyer:
Hmm.. on second thoughts, "enabled" should probably be "strict", so it's more clear this is a set of options. Otherwise "confinement: enabled" feels a lot like "confinement: true". Let me get Mark's ack.

kyrofa:
> Very good. So this is not going to be a list of attributes and this one of the tags?

Actually that's a good question. I made some assumptions about that, but now that you mention it I'm reminded of the "grade" attribute as well (or whatever it's called). @niemeyer should this be a list of attributes instead of global properties?

kyrofa:
> Let me get Mark's ack.

Easy to change, let me know.

niemeyer:
I think this is just a list of possibilities. Either it's confined, or it's in devmode. The other option I can think of is for it to be enforced, which would mean the snap cannot be put it devmode. These are all exclusive with each other.

niemeyer:
Confirmed, "strict" and "devmode" as values.

kyrofa:
> Confirmed, "strict" and "devmode" as values.

Done.

snapcraft/internal/lifecycle.py
```diff
@@ -40,7 +40,8 @@
 _TEMPLATE_YAML = r'''name: # the name of the snap
 version: # the version of the snap
 summary: # 79 char long summary
-description: # A longer description for the snap
+description: # a longer description for the snap
+confinement: devmode # devmode means the snap doesn't support confinement
```
elopio: what about: devmode means the snap will not be confined. Better or worse?

snapcraft/tests/test_commands_init.py
```diff
@@ -37,6 +37,16 @@ def test_init_must_write_snapcraft_yaml(self):
...
+description: # a longer description for the snap
+confinement: devmode # devmode means the snap doesn't support confinement"""
+
         main(['init'])
 
         self.assertEqual('Created snapcraft.yaml.\n', fake_logger.output)
+
+        # Verify the generated yaml
+        with open('snapcraft.yaml', 'r') as f:
+            self.assertEqual(f.read(), expected_yaml)
```
elopio: nice, thank you.

snapcraft/tests/test_meta.py
```diff
@@ -37,6 +37,7 @@ def setUp(self):
             'version': '1.0',
             'description': 'my description',
             'summary': 'my summary',
+            'confinement': 'enabled',
```
elopio: strict

snapcraft/tests/test_meta.py
```diff
@@ -54,6 +55,7 @@ def test_create_meta(self):
             y = yaml.load(f)
 
         expected = {'architectures': ['amd64'],
+                    'confinement': 'enabled',
```
elopio: strict

snapcraft/tests/test_meta.py
```diff
@@ -258,6 +260,7 @@ def test_create_meta_with_app(self):
             'summary': 'my summary',
             'name': 'my-package',
             'version': '1.0',
+            'confinement': 'enabled',
```
elopio: strict

snapcraft/internal/lifecycle.py
elopio: after a quick talk, I prefer yours better and love sprints for the speed of sound discussions.

snapcraft/tests/test_yaml.py
```diff
@@ -426,13 +441,96 @@ def test_invalid_yaml_missing_description(self, mock_loadPlugin):
...
+
+        self.assertEqual(
+            raised.exception.message,
+            "'confinement' is a required property")
+
+    @unittest.mock.patch('snapcraft.internal.yaml.Config.load_plugin')
+    def test_invalid_yaml_invalid_confinement_types(self, mock_loadPlugin):
+        invalid_confinement_types = [
+            'foo',
+            'enabled-',
```
elopio: strict- ?

snapcraft/tests/test_meta.py
kyrofa: Fixed.

snapcraft/tests/test_meta.py
kyrofa: Fixed.

snapcraft/tests/test_meta.py
kyrofa: Fixed.

snapcraft/tests/test_yaml.py
kyrofa: Ah yes thank you.

snapcraft/internal/lifecycle.py
kyrofa: > love sprints for the speed of sound discussions.

No kidding!

elopio:
Thanks kyle, I'm :+1: with a couple of minor corrections.

This change will make all the existing snapcraft files out there to start failing. Can you please add to the commit description that it is backwards incompatible so @sergiusens can mention that in his release notes too?

kyrofa:
> This change will make all the existing snapcraft files out there to start failing.

Thanks for bringing that up, @elopio. @niemeyer do you see any way around that while still adding the functionality Mark wanted?

sergiusens:
@niemeyer can `confinement: strict` be implicit? If not we will break all existing `snapcraft.yaml` (which isn't that bad) but also break existing snaps on the store.

elopio:
I forgot to mention that you have to update docs/snapcraft-syntax.

niemeyer:
It is necessarily implicit as that's what exists today, and we don't have the field.

That said, snapcraft shouldn't treat it as implicit. It should reject a snapcraft.yaml that does not inform the confinment choice, and give an advice about it. This ensures it's a conscious decision and that there's awareness about the option.

kyrofa:
> I forgot to mention that you have to update docs/snapcraft-syntax.

Agh, good catch. I actually need to update all the docs to specify this.

kyrofa:
> I forgot to mention that you have to update docs/snapcraft-syntax.

Alright, done.

elopio:
:+1: the docs changes. Thanks.

sergiusens:
@niemeyer that's fine, but this is a concious backwards compatibility breakage which we would need to communicate.

snapcraft/tests/test_commands_init.py
```diff
@@ -37,6 +37,16 @@ def test_init_must_write_snapcraft_yaml(self):
         fake_logger = fixtures.FakeLogger(level=logging.INFO)
         self.useFixture(fake_logger)
 
+        expected_yaml = """name: # the name of the snap
+version: # the version of the snap
+summary: # 79 char long summary
+description: # a longer description for the snap
+confinement: devmode # devmode means the snap doesn't support confinement"""
```
niemeyer: Suggested wording:

`confinement: devmode # replace by "strict" to enforce system access only via declared interfaces`

This gives a hint about both what it does and how to move forward.

kyrofa: > This gives a hint about both what it does and how to move forward.

That says what "strict" does, but does it describe what "devmode" does? It may presuppose knowledge that there are only two options that are mutually exclusive.

niemeyer: It doesn't presupposes, but it implies. That's a common trick to provide more information in less space. The latter information you mention is not provided in the original version or the suggested one, and I think that's okay. We can't document it all in a single line.

kyrofa: Alright, done.

niemeyer: Thanks!

kyrofa:
> [...] this is a concious backwards compatibility breakage which we would need to communicate.

After the discussion today with Mark, the latest commit makes this optional, and defaults to "strict." Note that Snapcraft will pop a hint about that default, and always specify the confinement in the resulting `snap.yaml`.

snapcraft/_schema.py
```diff
@@ -42,7 +42,7 @@ def __init__(self, snapcraft_yaml=None):
 
     @property
     def part_schema(self):
-        sub = self._schema['properties']['parts']['patternProperties']
+        sub = self.schema['properties']['parts']['patternProperties']
```
elopio: can you explain why is this change in here?
I find it weird that we will have an attribute part_schema and another schema.

snapcraft/tests/test_yaml.py
```diff
@@ -426,13 +441,99 @@ def test_invalid_yaml_missing_description(self, mock_loadPlugin):
             "'description' is a required property")
 
     @unittest.mock.patch('snapcraft.internal.yaml.Config.load_plugin')
+    def test_invalid_yaml_missing_confinement(self, mock_loadPlugin):
```
elopio: it is no longer invalid. Something like test_yaml_missing_confinement_must_log_warning.

elopio:
I see the need to make the internal _schema public, but it will be confusing with the existing part_schema attribute. If there is no better way to do it, then I'm ok with doing it but some docstrings would be nice.

Other than that and a tiny test rename, :+1:

kyrofa:
> I see the need to make the internal _schema public, but it will be confusing with the existing part_schema attribute.

Good catch, I didn't actually notice the `part_schema` property. We do indeed need to publicize the overall schema properties so we have a more general way of doing this, but I think it still makes sense to have the `part_schema` since getting there involves needing to know the regex, which is schema-specific. So I added a new property to get the overall schema and made `part_schema` use it. I also documented both. Does that look better?

kyrofa:
retest this please

kyrofa:
retest this please

sergiusens:
retest this please

sergiusens:
@kyrofa can you click on update branch here?

kyrofa:
> can you click on update branch here?

It says it's up to date for me :stuck_out_tongue: .

snapcraft/_schema.py
```diff
@@ -41,10 +41,18 @@ def __init__(self, snapcraft_yaml=None):
...
+        return self._schema['properties'].copy()
+
+    @property
     def part_schema(self):
-        sub = self._schema['properties']['parts']['patternProperties']
+        """Return part-specific schema properties."""
+
+        sub = self.schema['parts']['patternProperties']
         properties = sub['^(?!plugins$)[a-z0-9][a-z0-9+-]*$']['properties']
-        return properties.copy()
```
sergiusens: Is removing `.copy()` here really needed? There might be side effects we don't know about and I know I added it for a reason.

sergiusens: ah. `self.schema`

sergiusens:
retest this please

PR #502, opened 2016-05-12
Add support for epoch property.

kyrofa:
In order to support step upgrades, snaps need to be able to specify upgrade paths. This PR resolves LP: [#1581113](https://bugs.launchpad.net/snapcraft/+bug/1581113) by implementing that via a new "epoch" property specified in the `snap.yaml`. Valid epochs are positive integers (e.g. `0`, `1`, etc.) or positive integers followed by an asterisk (e.g. `1*`, `2*`, etc.).

`0` is epoch 0; `1*` is the upgrade path from 0 to 1; `1` is epoch 1 (and so on).

### DIFF
--- a/snapcraft/tests/test_yaml.py
+++ b/snapcraft/tests/test_yaml.py
@@ -16,6 +16,7 @@
 
 import logging
 import os
+import re
 import subprocess
 import sys
 import tempfile
@@ -451,11 +452,54 @@ parts:
     @unittest.mock.patch('snapcraft.internal.yaml.Config.load_plugin')
     def test_yaml_valid_epochs(self, mock_loadPlugin):
         valid_epochs = [
-            0,
-            '1*',
-            1,
-            '400*',
-            1234,
+            {
+                'yaml': 0,
+                'expected': 0,
+            },
+            {
+                'yaml': '"0"',
+                'expected': '0',
+            },
+            {
+                'yaml': '1*',
+                'expected': '1*',
+            },
+            {
+                'yaml': '"1*"',
+                'expected': '1*',
+            },
+            {
+                'yaml': 1,
+                'expected': 1,
+            },
+            {
+                'yaml': '"1"',
+                'expected': '1',
+            },
+            {
+                'yaml': '400*',
+                'expected': '400*',
+            },
+            {
+                'yaml': '"400*"',
+                'expected': '400*',
+            },
+            {
+                'yaml': 1234,
+                'expected': 1234,
+            },
+            {
+                'yaml': '"1234"',
+                'expected': '1234',
+            },
+            {
+                'yaml': '0001',
+                'expected': 1,
+            },
+            {
+                'yaml': 1.0,
+                'expected': 1,
+            },
         ]
 
         fake_logger = fixtures.FakeLogger(level=logging.ERROR)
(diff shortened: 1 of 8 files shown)


snapcraft/tests/test_yaml.py
```diff
@@ -449,6 +449,79 @@ def test_tab_in_yaml(self, mock_loadPlugin):
             'on line 2 of snapcraft.yaml')
 
     @unittest.mock.patch('snapcraft.internal.yaml.Config.load_plugin')
+    def test_yaml_valid_epochs(self, mock_loadPlugin):
+        valid_epochs = [
+            0,
+            '1*',
+            1,
+            '400*',
+            1234,
```
elopio: what about 00001, valid or invalid?

elopio:
I'm +1, with one question. Also, please update docs/snapcraft-syntax.md

snapcraft/tests/test_yaml.py
kyrofa: Valid since YAML will parse it as 1, but difficult to test as `0001` isn't a valid number in Python.

kyrofa:
> Also, please update docs/snapcraft-syntax.md

Good catch, thanks-- done.

kyrofa:
retest this please.

kyrofa:
retest this please

snapcraft/tests/test_yaml.py
sergiusens: 00001 could be an integration test, but I fear we would be then testing the yaml parser instead :wink:

sergiusens: how about `'00001'`? Is that valid? I am asking without having looked at how this is spec'ed

kyrofa: You mean the string "00001"? That's rejected, to avoid having to validate the number separately from the asterisk in the string. I've updated the tests quite a bit, adding both a test for 01 and "01".

sergiusens:
you can make this always a string by implementing `JSONDecoder` so it loads as a `str` straight up

kyrofa:
> you can make this always a string by implementing `JSONDecoder` so it loads as a str straight up.

I've instead updated this to validate with a custom format, where I validate via strings only and can provide a nicer-looking error message. Let me know what you think.

snapcraft/tests/test_yaml.py
```diff
@@ -656,22 +653,11 @@ def test_invalid_yaml_invalid_epochs(self, mock_loadPlugin):
...
-                else:
-                    self.assertRegex(
-                        raised.exception.message,
-                        "The 'epoch' property does not match the required "
-                        "schema:.*does not match '{}'".format(
-                            re.escape('^(?:0|[1-9][0-9]*[*]?)$')))
+                self.assertRegex(
+                    raised.exception.message,
+                    "The 'epoch' property does not match the required "
+                    "schema:.*is not a 'epoch' \(epochs are positive integers "
```
sergiusens: an epoch?

kyrofa: It's actually hard-coded in jsonschema to 'is not a'. Terrible, eh? But I wanted to use the format checkers the way they intended...

kyrofa: I'm thinking about contributing 'is not a valid' upstream.

kyrofa:
retest this please

sergiusens:
retest this please :cry:

sergiusens:
retest this please

elopio:
Both tests passed, but jenkins failed to communicate the result back to github.
I'm merging.

PR #503, opened 2016-05-12
Reject plug/slot definitions at part level

zyga:
This patch removes incorrect schema entries that allow arbitrary plug
and slot definitions at a part level. Those definitions are ignored by
snapcraft but can cause potential confusion for users that read the
schema.

Fixes: https://bugs.launchpad.net/snapcraft/+bug/1581166
Signed-off-by: Zygmunt Krynicki <zygmunt.krynicki@canonical.com>

### DIFF
--- a/schema/snapcraft.yaml
+++ b/schema/snapcraft.yaml
@@ -163,12 +163,6 @@ properties:
             items:
               type: string
             default: ['*']
-          plugs:
-            type: object
-            description: used to connect to interfaces
-          slots:
-            type: object
-            description: used to offer interfaces
   plugs:
     type: object
   slots:


kyrofa:
Thanks @zyga-- can you please change your commit message to use `LP: #<bug>` instead of `Fixes` using the URL?

kyrofa:
@zyga mind also adding a regression test? Beyond that this looks good to me, but I'm not sure why this was there initially so I invoke @sergiusens.

sergiusens:
I indeed :+1: this, we discussed it together. But a unit test would be nice; in `snapcraft.test_yaml`

sergiusens:
Closing in favor of #514

PR #505, opened 2016-05-21
Clean up the ROS example test.

elopio:


elopio:
retest this please

elopio:
This is related to https://bugs.launchpad.net/snapcraft/+bug/1544790
It makes the examples tests pass in the cloud. However in order to reenable them in autopkgtests we first need a proxy exception. I'm requesting it...

examples_tests/test_ros.py
```diff
@@ -0,0 +1,36 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2015, 2016 Canonical Ltd
```
sergiusens: this should be just 2016

elopio: This is code from 2015, just moved to a separate file now. So I think it should have the two years.

elopio: This is code from 2015, just moved to a separate file now. So I think it should have the two years.

elopio: This is code from 2015, just moved to a separate file now. So I think it should have the two years.

kyrofa:
Closed in favor of #515

PR #509, opened 2016-05-23
Add node-engine to snapcraft.yaml

earnubs:
Allow the developer to set the version of node they want, the current plugin uses 4.2.2, LTS is 4.4.4 and higher versions enable more of ES2015 http://node.green/

### DIFF
--- a/snapcraft/plugins/nodejs.py
+++ b/snapcraft/plugins/nodejs.py
@@ -72,21 +72,18 @@ class NodePlugin(snapcraft.BasePlugin):
 
         if 'required' in schema:
             del schema['required']
-        
-        # Inform Snapcraft of the properties associated with building. If these
-        # change in the YAML Snapcraft will consider the build step dirty.
-        schema['build-properties'].append(['node-packages', 'node-engine'])
 
         # Inform Snapcraft of the properties associated with building. If these
         # change in the YAML Snapcraft will consider the build step dirty.
-        schema['build-properties'].append('node-packages')
+        schema['build-properties'].append(['node-packages', 'node-engine'])
 
         return schema
 
     def __init__(self, name, options, project):
         super().__init__(name, options, project)
         self._npm_dir = os.path.join(self.partdir, 'npm')
-        self._nodejs_tar = sources.Tar(_get_nodejs_release(), self._npm_dir)
+        self._nodejs_tar = sources.Tar(_get_nodejs_release(
+            self.options.node_engine), self._npm_dir)
 
     def pull(self):
         super().pull()
@@ -110,15 +107,15 @@ class NodePlugin(snapcraft.BasePlugin):
             self.run(['npm', 'install', '-g'])
 
 
-def _get_nodejs_base():
+def _get_nodejs_base(node_engine):
     machine = platform.machine()
     if machine not in _NODEJS_ARCHES:
         raise EnvironmentError('architecture not supported ({})'.format(
             machine))
-    return _NODEJS_BASE.format(version=_NODEJS_VERSION,
+    return _NODEJS_BASE.format(version=node_engine,
                                arch=_NODEJS_ARCHES[machine])
 
 
-def _get_nodejs_release():
-    return _NODEJS_TMPL.format(version=_NODEJS_VERSION,
-                               base=_get_nodejs_base())
+def _get_nodejs_release(node_engine):
+    return _NODEJS_TMPL.format(version=node_engine,
+                               base=_get_nodejs_base(node_engine))
--- a/snapcraft/plugins/nodejs.py
+++ b/snapcraft/plugins/nodejs.py
@@ -78,7 +78,8 @@ class NodePlugin(snapcraft.BasePlugin):
 
         # Inform Snapcraft of the properties associated with building. If these
         # change in the YAML Snapcraft will consider the build step dirty.
-        schema['build-properties'].append(['node-packages', 'node-engine'])
+        schema['build-properties'].append('node-packages')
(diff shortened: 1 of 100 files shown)


snapcraft/plugins/nodejs.py
```diff
@@ -64,6 +63,9 @@ def schema(cls):
             },
             'default': [],
         }
+        schema['properties']['node-engine'] = {
+            'type': 'string'
```
sergiusens: maybe add a default to `4.4.4` here?

snapcraft/plugins/nodejs.py
```diff
@@ -72,8 +74,12 @@ def schema(cls):
 
     def __init__(self, name, options):
         super().__init__(name, options)
-        self._nodejs_tar = sources.Tar(_get_nodejs_release(),
-                                       os.path.join(self.partdir, 'npm'))
+        try:
+            self._nodejs_tar = sources.Tar(
+                    _get_nodejs_release(self.options.node_engine),
+                    os.path.join(self.partdir, 'npm'))
+        except AttributeError:
```
sergiusens: by adding a default there is no need for this exception dance ;-)

snapcraft/plugins/nodejs.py
```diff
@@ -72,8 +74,12 @@ def schema(cls):
...
+            raise AttributeError('node-engine not defined in snapcraft.yaml')
```
kyrofa: This probably means it needs to be required in the schema instead of deleting the `required` key on line 71.

snapcraft/plugins/nodejs.py
```diff
@@ -64,6 +63,9 @@ def schema(cls):
             },
             'default': [],
         }
+        schema['properties']['node-engine'] = {
```
sergiusens: this needs to be explained in the docstring above

snapcraft/plugins/nodejs.py
earnubs: Originally I did havethe default, but I'm not sure it's exactly what you want because you're almost always going to have a default that is out of date, as such I thought it better to make the developer take full control of this, and no default...

snapcraft/plugins/nodejs.py
earnubs: agreed

snapcraft/plugins/nodejs.py
earnubs: not sure what you mean, did I delete the ```required``` key?

snapcraft/plugins/nodejs.py
earnubs: Done.

snapcraft/plugins/nodejs.py
kyrofa: Not _you_, but the plugin currently deletes it on line 71. If you add the `node-engine` option to the `required` key, snapcraft will catch it when validating the YAML instead of your needing to do it here. See the copy plugin for an example of what you'll probably need to do.

earnubs: Thanks!

snapcraft/plugins/nodejs.py
```diff
@@ -61,25 +64,23 @@ def schema(cls):
...
         schema['properties']['node-engine'] = {
             'type': 'string'
         }
-
-        if 'required' in schema:
-            del schema['required']
+        # `node-engine` keyword is required, but `node-packages` is not.
+        schema['required'].append('node-engine')
+        schema['required'].remove('source')
+        schema['properties']['source']['default'] = '.'
```
sergiusens: Can we leave this one for later? `default` is a contentious topic and would be better applied across the board

sergiusens:
Thanks for this, looks good. Mind rebasing or mergin master in to get rid of whatever the conflict github mentions is?

snappy-m-o:

snapcraft/plugins/nodejs.py
sergiusens: Yes, but this breaks backwards compatibility so we now are stuck with it. We update snapcraft often enough that we can always keep it using the best choice available

snapcraft/plugins/nodejs.py
```diff
@@ -62,18 +64,23 @@ def schema(cls):
...
-
-        if 'required' in schema:
-            del schema['required']
+        schema['properties']['node-engine'] = {
+            'type': 'string'
+        }
+        # `node-engine` keyword is required, but `node-packages` is not.
+        schema['required'].append('node-engine')
+        schema['required'].remove('source')
+        schema['properties']['source']['default'] = '.'
```
kyrofa: I doubt you want this line here (defaulting the source to '.'). Also, the comment doesn't seem quite right. Perhaps:

```python
# `node-engine` keyword is required, but `source` is not.
schema['required'].append('node-engine')
schema['required'].remove('source')
```

sergiusens:
OK, I am :+1: with this but the diff looks super weird

kyrofa:
@earnubs mind rebasing please? I think that's why the diff looks odd, and that was our mistake, not yours-- sorry about that :unamused: .

earnubs:
@kyrofa for reasons I don't understand I cannot rebase this without getting stuck in an endless merge conflict loop :)

I've merged master and that seems to have cleared up the PR diff, but if you're not happy I'll submit a new PR (rather than figure out what's going on with this rebase).

kyrofa:
@earnubs ah! That works just fine, thank you :smile: .

sergiusens:
ok to test

sergiusens:
ok to test

PR #510, opened 2016-05-23
Use the in ubuntu-core python3

sergiusens:
LP: #1582513

### DIFF
--- a/snapcraft/plugins/python3.py
+++ b/snapcraft/plugins/python3.py
@@ -69,16 +69,18 @@ class Python3Plugin(snapcraft.BasePlugin):
 
     def __init__(self, name, options, project):
         super().__init__(name, options, project)
-        self.stage_packages.extend([
+        self.build_packages.extend([
             'python3-dev',
+            'python3-pip',
             'python3-pkg-resources',
             'python3-setuptools',
         ])
 
     def env(self, root):
         return [
-            'PYTHONPATH={}'.format(os.path.join(
+            'PYTHONPATH={0}'.format(os.path.join(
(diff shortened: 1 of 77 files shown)


snapcraft/plugins/python3.py
```diff
@@ -69,16 +69,18 @@ def schema(cls):
...
             'python3-dev',
+            'python3-pip',
             'python3-pkg-resources',
             'python3-setuptools',
         ])
 
     def env(self, root):
         return [
-            'PYTHONPATH={}'.format(os.path.join(
+            'PYTHONPATH={0}'.format(os.path.join(
```
kyrofa: Is there a reason to use `{0}` instead of `{}` here?

sergiusens: left over from when I tried to use `:` with multiple entries, so no :-)

sergiusens:
retest this please

sergiusens:
all green :100:

kyrofa:
> all green :100: 

Can you really celebrate that when your tests look like [this](https://github.com/ubuntu-core/snapcraft/pull/510/files#diff-e047afa9d05eda5d5cf933f0e2dc11dcR22)? :stuck_out_tongue:

sergiusens:
@kyrofa I celebrate the fact that the infrastructure work and there was no regression! :wink:

kyrofa:
Hahaha

elopio:
alright, this seems to simplify the code a lot. Do we need a review from a python/pip expert?

You need to readd the tests for py3, and rebase with master.

PR #511, opened 2016-05-24
Run the integration tests against a local fake server when the user p…

elopio:
…assword is not in the environment.

LP: #1585023

PR #512, opened 2016-05-24
Move the login and logout methods to a client.

elopio:


elopio:
This is the first split of the macaroons branch. It just moves the simple login and logout methods to a client object. Requires #511.

elopio:
Not yet ready, closing for now.

PR #513, opened 2016-05-24
Add snapcraft examples to scaffold getting started examples

didrocks:
Adds unit and integration tests for it, reusing and making more generic all demos tests helper.

Note that first commit is on another PR (move examples to demos): https://github.com/ubuntu-core/snapcraft/pull/506 which needs to be merged first for review clarity.

This one will then be followed up by one PR changing the url for eamples 05- to point to this repository, and as well for disabling the manual run_demos() additional run when the "example" target is selected once CI changes the runner script.


didrocks:
Rebased on latest master changes

didrocks:
mock doesn't behave the same between python 3.4 (travis CI) and 3.5 (xenial), seems we have to patch builtins to be compatible in 3.4, gave it a shot for Travis CI.

didrocks:
ok, pushed and hoping this is the last time I had to rebase :)

So, for now, we are now shipping (as per demand) the binary in the snapcraft package (it needs to be installed alongside snapcraft). Ship them the scaffolding get started with it.

I still have hope to get it renamed from snapcraft-examples to "snapcraft learn", but that will be in another separated PR.

Do you mind merging this please? I'm away for a couple of days, FYI

kyrofa:
Note mark's comment on bug [#1586137](https://bugs.launchpad.net/snapcraft/+bug/1586137): "Let's call this "snapcraft examples" please, not snapcraft-examples."

demos_tests/__main__.py
```diff
@@ -67,7 +68,7 @@ def main():
     # them again.
     argv = [sys.argv[0]]
     argv.append('discover')
-    argv.append('demos_tests')
+    argv.append(os.path.dirname(__file__).split(os.path.sep)[-1])
```
elopio: I think you want here os.path.basename, instead of split and [-1]

examples/01-hello-world-cli/snapcraft.yaml
```diff
@@ -0,0 +1,16 @@
+name: hello-world-cli
+version: 0.1
+summary: Hello World simple C app
+description: |
+  This demo is intended to show how to build a C app
```
elopio: s/demo/example ?

snapcraft/main.py
```diff
@@ -116,6 +123,27 @@ def _get_version():
...
+        shutil.copytree(get_examplesdir(), dest_dir)
+    except FileExistsError:
+        # don't event try to copy if the dest exists already
+        if not os.path.isdir(dest_dir):
+            raise NotADirectoryError("{} is a file, can't be used as a "
+                                     "destination".format(dest_dir))
+        dest_dir = os.path.join(dest_dir, _SNAPCRAFT_TOUR_DIR)
+        shutil.copytree(get_examplesdir(), dest_dir)
+
+    print("Snapcraft tour initialized in {}.\n"
```
elopio: I think this should be logger.info, but we have some inconsistencies here in how to handle errors and messages in main. That's a bug for the next milestone.

elopio:
This is great didrocks. For the tests, what if instead of having a demos_tests and examples_tests, we put them all in a user_tests dir or snaps_tests dir?
For the point of view of our CI pipeline, the two serve the same purpose, so I see no reason why we would like to have them separated. I would prefer a single ./runtests command to run both.

demos_tests/__main__.py
didrocks: Oh, good catch!, I must have my head broken when writing this :)

examples/01-hello-world-cli/snapcraft.yaml
didrocks: Well, you can ignore the current examples directory, they are being renamed as "tour" and Mark is rewriting them.

snapcraft/main.py
didrocks: ok, I can rename and use logger.info if the output is the same.

didrocks:
I'm all for having just one single _tests directory. Let me try to play with this (have both under snaps_tests/ dir) and then, looping on the 2 subdirs to tests tour/ (replacing examples/) and demos/.

Do you still want the CI target to be "examples" or should be replace it with "snaps"?

I'll rebase on this, push Mark's current changes (so /!\ not to merge, and Mark did some changes in term of help text content in the same branch, so the diff will be a little bit larger…), rebasing on master and implementing the above change for pre-review.

elopio:
I'd like the ci target to have the same name of that new demos+tours tests directory.
We will have to change the jobs in jenkins, which is easy but requires syncronization or dealing with failed executions. So I'll just keep an eye on your proposals. Let me know if you need a hand dealing with the test suites.

didrocks:
Excellent! I'll push something with both targets "examples" and "snaps" first, running the some run_snaps tests (to ensure we always run something and I can see I didn't make a booboo ;)). We can even land it, and then, once CI has migrated (when most of current PR are rebased on master to change the targets), do another PR to remove the "examples" target.

I'll go for pushing the update today and will ping you :)

sergiusens:
If we land this tomorrow, can we update `.gitignore`?

The move from `examples` to `demos` is driving my `status` nuts :-P

didrocks:
Here we go, rebased on master and with changes to have demos_tests and tour_tests under the same directory.
Note that the command is backward-comptible:
`./runtests examples` and `./runtests snaps` both runs the same tests.

Seems that autopkgtest are down again.

Examples tests run both the old demos_tests and snaps_tests and the result is: "[1;32mEverything passed[0m". Unsure why the result doesn't reflect this.

(Note that the docstring changes are comming directly from Mark's branch as this one includes his current modifications: https://github.com/markshuttle/snapcraft/commits/tour)

bin/snapcraft-examples
```diff
@@ -0,0 +1,31 @@
...
+import os
+import sys
+
+# make running from snapshot easy
+topdir = os.path.abspath(os.path.join(__file__, '..', '..'))
+if os.path.exists(os.path.join(topdir, 'setup.py')):
+    sys.path = [topdir] + sys.path
+
+
+if __name__ == '__main__':
```
sergiusens: I thought this was supposed to be `snapcraft tour` (no `-` dash)

What is this for, I might be missing something?

setup.py
```diff
@@ -43,7 +43,7 @@
               'snapcraft.plugins',
               'snapcraft.storeapi'],
     package_data={'snapcraft.internal': ['manifest.txt']},
-    scripts=['bin/snapcraft'],
+    scripts=['bin/snapcraft', 'bin/snapcraft-examples'],
```
sergiusens: ditto, what is `snapcraft-examples`?

snaps_tests/demos_tests/test_ros.py
```diff
@@ -15,13 +15,13 @@
...
+import snaps_tests
 
 import os
 import subprocess
 
 
-class ROSTestCase(demos_tests.ExampleTestCase):
+class ROSTestCase(snaps_tests.SnapsTestCase):
 
     demo_dir = 'ros'
```
elopio: s/demo_dir/snap_content_dir

bin/snapcraft-examples
didrocks: It's me, this binary should be removed and is useless, I forgot to delete it. Pushing the removal right away

setup.py
didrocks: yeah, leftover as well (now fixed and pushed). Good catch!

snaps_tests/demos_tests/test_ros.py
didrocks: Good catch! Fixed

sergiusens:
demostests           FAIL non-zero exit status 1

sergiusens:
/usr/bin/python3: No module named demos_tests

elopio:
There is one error in one of the tests, but they are failing anyway. Skipped in #544

didrocks:
Seems like the forked version is now passing tests (I have no idea about the bad system calls though, seems like an apparmor issue when trying echo?)

didrocks:
retest this please

(seems the autopkgtest issue was a transient failure, connection dropped)

didrocks:
and everything pass! This branch should be good to merge, containing latest elopio's changes :)

PR #517, opened 2016-05-25
Release/2.9

sergiusens:


elopio:
:+1: 
There is no mention of the ROS example cleanup, but I think that's for the best as I mentioned that we might need to reopen the same bug after IS opens the proxy.

PR #518, opened 2016-05-25
Move the login and logout methods to a client.

elopio:


elopio:
This is the first split of the macaroons branch. It just moves the simple login and logout methods to a client object. Requires #511.

elopio:
retest this please

sergiusens:
@elopio just a reminder to update the branch :wink:

elopio:
The unit tests are passing locally and on jenkins, but failing on travis. And travis refuses to show the output. I'm retrying...

sergiusens:
Oh, this is a case of cyclic dependencies. I had the same problem before. Travis uses python3.4 I believe and this is not handled vey well.

snapcraft/tests/store/test_store_client.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+class LoginTestCase(tests.TestCase):
+
+    def setUp(self):
+        super().setUp()
+        self.useFixture(fixture_setup.FakeStore())
+        self.client = storeapi.StoreClient()
+
+    def test_login_successful(self):
```
sergiusens: is there a fake server response where we use the correct password but otp is required? (just want to test the response)

sergiusens:
:+1:

elopio:
restest this please

snapcraft/tests/store/test_store_client.py
elopio: yes, look down:
test_failed_login_with_wrong_one_time_password

That's only in unit tests though. We could add one for integration, even hitting the staging test. I'm not sure if it adds value, but it's easy to do. File a bug if you want it, and assign it to me.

sergiusens: I thought an incorrect otp was different than a missing one which is required

elopio: as far as I know, the answer is the same. Either an sso.ApiException or an sso.UnexpectedApiException. That is handled by the ssoclient. From our side, we just need to make sure that the success field is False.
Unless you mean something different and I misunderstood your comment.

PR #519, opened 2016-05-25
Improve the config handling.

elopio:


elopio:
I think this is the last of the possible independent refactors. The following branches will depend on each other so they can't be landed separately. This one depends on #518.

snapcraft/config.py
```diff
@@ -13,69 +13,73 @@
...
+        self.filename = None
+        self.load()
+
+    def _section_name(self):
+        # The only section we care about is the host from the SSO url
+        url = os.environ.get('UBUNTU_SSO_API_ROOT_URL',
+                             constants.UBUNTU_SSO_API_ROOT_URL)
+        return urllib.parse.urlparse(url).netloc
+
+    def get(self, option_name):
```
sergiusens: does this mean people need to re-login?

sergiusens:
fwiw it feels like `snapcraft.config` could be moved into the `storeapi`

snapcraft/config.py
elopio: no, the config format is the same.

PR #523, opened 2016-05-26
Support zip files as source

sergiusens:
LP: #1577062


snapcraft/internal/sources.py
```diff
@@ -80,6 +82,28 @@ def __init__(self, source, source_dir, source_tag=None,
...
 
+class FileBase(Base):
+
+    def pull(self):
+        if common.isurl(self.source):
+            self.download()
+        else:
+            shutil.copy2(self.source, self.source_dir)
+
+        self.provision(self.source_dir)
```
kyrofa: This doesn't seem to be implemented in this class. Should it be, and raise a NotImplemented or something?

snapcraft/internal/sources.py
```diff
@@ -80,6 +82,28 @@ def __init__(self, source, source_dir, source_tag=None,
...
+    def download(self):
+        req = requests.get(self.source, stream=True, allow_redirects=True)
+        if req.status_code is not 200:
+            raise EnvironmentError('unexpected http status code when '
+                                   'downloading {}'.format(req.status_code))
+
+        file = os.path.join(self.source_dir, os.path.basename(self.source))
+        with open(file, 'wb') as f:
+            for chunk in req.iter_content(1024):
+                f.write(chunk)
```
kyrofa: Not for this PR, but we really need a progress bar here. Perfect use-case for one.

snapcraft/internal/sources.py
sergiusens: I've been told lately it is not really necessary, it just needs to behave and act like a duct :-)

I am only saying this since it is tightly controlled code.

snapcraft/internal/sources.py
sergiusens: Good idea. I hope not, this already merged :weary:

kyrofa: :stuck_out_tongue:

snapcraft/internal/sources.py
kyrofa: Alrighty.

kyrofa:
You have my retroactive :+1:

PR #525, opened 2016-05-27
Support the assumes keyword

sergiusens:
LP: #1586429

### DIFF
--- a/docs/snapcraft-syntax.md
+++ b/docs/snapcraft-syntax.md
@@ -22,7 +22,8 @@ contain.
   this snap doesn't support running under confinement) or "strict" (i.e. full
   confinement supported via interfaces).
 * `assumes` (list of strings)
-  A feature tag list that the core needs to support for this snap to be run.
+  A list of features that must be supported by the core in order for this snap
+  to install.
 * `epoch` (string)
   The epoch to which this revision of the snap belongs. This is used to specify
   upgrade paths. For example, `0` is epoch 0; `1*` is the upgrade path from 0 to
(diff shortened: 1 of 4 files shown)


sergiusens:
@pedronis a sanity check from your side would be ideal

docs/snapcraft-syntax.md
```diff
@@ -21,6 +21,8 @@ contain.
   The type of confinement supported by the snap. Can be either "devmode" (i.e.
   this snap doesn't support running under confinement) or "strict" (i.e. full
   confinement supported via interfaces).
+* `assumes` (list of strings)
+  A feature tag list that the core needs to support for this snap to be run.
```
kyrofa: From the bug, this seems to be an install-time check. Maybe "A list of features that must be supported by the core in order for this snap to install." ?

sergiusens: El 27/05/16 a las 13:02, Kyle Fazzari escribió:
> In docs/snapcraft-syntax.md
> <https://github.com/ubuntu-core/snapcraft/pull/525#discussion_r64927675>:
>
> > @@ -21,6 +21,8 @@ contain.
> >    The type of confinement supported by the snap. Can be either "devmode" (i.e.
> >    this snap doesn't support running under confinement) or "strict" (i.e. full
> >    confinement supported via interfaces).
> > +* `assumes` (list of strings)
> > +  A feature tag list that the core needs to support for this snap to be run.
>
> From the bug, this seems to be an install-time check. Maybe "A list of
> features that must be supported by the core in order for this snap to
> install." ?
Good call

kyrofa:
Looks good! :+1:

PR #526, opened 2016-05-27
Bump our dependency to python 3.5

elopio:


sergiusens:
Yup, expected, bug bye

PR #528, opened 2016-05-27
Improve the template for snapcraft init

sergiusens:
LP: #1575581

### DIFF
--- a/integration_tests/test_snap.py
+++ b/integration_tests/test_snap.py
@@ -147,3 +147,10 @@ class SnapTestCase(integration_tests.TestCase):
         self.assertThat(
             os.path.join(project_dir, 'snap', 'usr', 'bin', 'nmtui'),
             Not(FileExists()))
+
+    def test_snap_from_snapcraft_init(self):
+        self.assertThat('snapcraft.yaml', Not(FileExists()))
+        self.run_snapcraft('init')
+        self.assertThat('snapcraft.yaml', FileExists())
+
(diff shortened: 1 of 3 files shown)


kyrofa:
Meh.... I'm so-so on this. Running "`snapcraft` to get started" will certainly _not_ do what the developer wants. I like the additions of the placeholder data in the global metadata, but I'm not a huge fan of the default part since it'll never do anything useful. I'm :+1: on the code though if we think it's necessary.

sergiusens:
@kyrofa imagine someone just getting started for the first time. Yeah it goes hand in hand with `snapcraft examples`

elopio:
I like this a lot. Every time I run snapcraft init I have to open an existing snapcraft.yaml to see fill it. This will make it a lot easier.

integration_tests/test_snap.py
```diff
@@ -147,3 +147,10 @@ def test_snap_with_exposed_files(self):
         self.assertThat(
             os.path.join(project_dir, 'snap', 'usr', 'bin', 'nmtui'),
             Not(FileExists()))
+
+    def test_snap_from_snapcraft_init(self):
+        self.assertThat('snapcraft.yaml', Not(FileExists()))
+        self.run_snapcraft('init')
+        self.assertThat('snapcraft.yaml', FileExists())
+
+        self.run_snapcraft('snap')
```
elopio: awesome! :tada:

PR #530, opened 2016-05-29
Moved the download to the store client.

elopio:
LP: #1586836


integration_tests/snaps/kernel-download/snapcraft.yaml
```diff
@@ -0,0 +1,10 @@
+name: kernel-download
+version: 0.1
+summary: Simple kernel snap to test download
+description: test
+type: kernel
```
sergiusens: you don't really need this to be a kernel snap, you just need to use the kernel plugin ;-)

snapcraft/storeapi/__init__.py
```diff
@@ -72,3 +133,89 @@ def login(self, email, password, token_name, one_time_password=None):
...
+
+        file_sum = hashlib.sha512()
+        with open(path, 'rb') as f:
+            for file_chunk in iter(
+                    lambda: f.read(file_sum.block_size * 128), b''):
+                file_sum.update(file_chunk)
+        return expected_sha512 == file_sum.hexdigest()
+
+
+class CPIClient(Client):
```
sergiusens: Not fond of this class name given that the name CPI is sort of something we don't want to expose outside the store

sergiusens:
Looks good, but let's just get rid of `CPIClient` as a class name

snapcraft/storeapi/__init__.py
elopio: IndexClient?

integration_tests/snaps/kernel-download/snapcraft.yaml
elopio: ok, removing it...

snapcraft/storeapi/__init__.py
sergiusens: El 30/05/16 a las 12:00, Leo Arias escribió:
> In snapcraft/storeapi/__init__.py
> <https://github.com/ubuntu-core/snapcraft/pull/530#discussion_r65082771>:
> 
>> +        else:
>> +            raise SHAMismatchError(download_path, expected_sha512)
>> +
>> +    def _is_downloaded(self, path, expected_sha512):
>> +        if not os.path.exists(path):
>> +            return False
>> +
>> +        file_sum = hashlib.sha512()
>> +        with open(path, 'rb') as f:
>> +            for file_chunk in iter(
>> +                    lambda: f.read(file_sum.block_size * 128), b''):
>> +                file_sum.update(file_chunk)
>> +        return expected_sha512 == file_sum.hexdigest()
>> +
>> +
>> +class CPIClient(Client):
> 
> IndexClient?

SnapIndexClient?

elopio: Done.

elopio:
retest this please

elopio:
retest this please

elopio:
retest this please

PR #531, opened 2016-05-30
Moved the upload to the store client.

elopio:
LP: #1586874


elopio:
Requires #530

.travis.yml
```diff
@@ -15,7 +15,7 @@ env:
   - TEST_SUITE=unit
 before_install:
   - sudo apt-get -qq update
-  - sudo apt-get install -y python3-coverage python3-distutils-extra
+  - sudo apt-get install -y python3-coverage python3-distutils-extra squashfs-tools
```
sergiusens: You probably want this in `debian/control`s `Build-Depends`

sergiusens: Or I don't understand why you need it here if we just use travis for unit tests.

snapcraft/_store.py
```diff
@@ -59,25 +54,42 @@ def logout():
...
+    logger.info('Uploading existing {}.'.format(snap_filename))
 
-        storeapi.upload(snap_filename, snap_name)
+    try:
+        store = storeapi.StoreClient()
+        result = store.upload(snap_filename)
+    except storeapi.InvalidCredentialsError:
+        logger.info('No valid credentials found.'
+                    ' Have you run "snapcraft login"?')
+        return
```
sergiusens: This should really `raise`, shouldn't it?

snapcraft/_store.py
```diff
@@ -59,25 +54,42 @@ def logout():
...
-    with tempfile.TemporaryDirectory() as temp_dir:
-        subprocess.check_call(
-            ['unsquashfs', '-d', os.path.join(temp_dir, 'squashfs-root'),
-             snap_path, '-e', os.path.join('meta', 'snap.yaml')])
-        with open(os.path.join(
-                temp_dir, 'squashfs-root', 'meta', 'snap.yaml')) as yaml_file:
-            snap_yaml = yaml.load(yaml_file)
+    # Print another newline to make sure the user sees the final result of the
+    # upload (success/failure).
+    logger.info('')
```
sergiusens: This should be a `print()`, it makes little sense to log an empty line if it is for ui purposes

.travis.yml
elopio: that sounds right. I wonder why the autopkgtests haven't failed.

snapcraft/_store.py
```diff
@@ -59,25 +54,42 @@ def logout():
...
 
-    return snap_yaml['name']
+    if success:
+        message = 'Application uploaded successfully'
+        if revision:
+            message += ' (as revision {})'.format(revision)
```
sergiusens: Why not `'{} (as revision {})'.format(message, revision)` ?

snapcraft/_store.py
```diff
@@ -59,25 +54,42 @@ def logout():
...
+        logger.info(message)
+    else:
+        logger.info('Upload did not complete.')
+
+    if errors:
+        logger.info('Some errors were detected:\n\n%s\n',
```
sergiusens: I thought we were going to use `format` for everything

snapcraft/_store.py
```diff
@@ -59,25 +54,42 @@ def logout():
...
+                    '\n'.join(str(error) for error in errors))
+
+    if app_url:
+        logger.info('Please check out the application at: %s\n',
```
sergiusens: same about `format`

snapcraft/storeapi/__init__.py
```diff
@@ -134,6 +137,36 @@ def logout(self):
...
+            raise InvalidCredentialsError()
+        data = _upload.upload_files(snap_filename, self.updown)
+        success = data.get('success', False)
+        if not success:
+            return data
+
+        result = _upload.upload_app(self.sca, snap_name, data)
+        return result
+
+    def _get_name_from_snap_file(self, snap_path):
```
sergiusens: this doesn't really need to be a class member, right?

.travis.yml
elopio: ah, I know. That is installed now by snapd. I added the build-depend.

snapcraft/_store.py
elopio: you are right, good catch. Also it should be error, not info. I made the fix, and added a test for the raise.

snapcraft/_store.py
```diff
@@ -60,9 +60,9 @@ def upload(snap_filename):
         store = storeapi.StoreClient()
         result = store.upload(snap_filename)
     except storeapi.InvalidCredentialsError:
-        logger.info('No valid credentials found.'
-                    ' Have you run "snapcraft login"?')
-        return
+        logger.error('No valid credentials found.'
+                     ' Have you run "snapcraft login"?')
+        raise
```
sergiusens: I was thinking more of a `raise EnvironmentError('No valid credentials found.'...)` but this works

snapcraft/_store.py
elopio: done.

snapcraft/_store.py
elopio: done.

snapcraft/_store.py
elopio: done.

snapcraft/_store.py
elopio: done.

snapcraft/_store.py
elopio: mmm, with that the error wouldn't be printed in red. But it makes sense, maybe we should review all the execptions that we are throwing back to main.

snapcraft/storeapi/__init__.py
elopio: moved up.

PR #532, opened 2016-05-30
Add macaroon support to login, upload and download.

elopio:
LP: #1586910


elopio:
Requires #531.

snapcraft/storeapi/__init__.py
```diff
@@ -208,6 +242,33 @@ def _is_downloaded(self, path, expected_sha512):
...
+        if one_time_password:
+            data['otp'] = one_time_password
+        response = self.post(
+            'tokens/discharge', data=json.dumps(data),
+            headers={'Content-Type': 'application/json',
+                     'Accept': 'application/json'})
+        if response.ok:
+            return response.json()['discharge_macaroon'], None
+        else:
+            return None, response.text
```
elopio: Instead of returning a tuple, this should raise an exception. For now I'm just copying vila's work, we can fix this later.

snapcraft/storeapi/macaroons.py
```diff
@@ -0,0 +1,263 @@
...
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+"""Code copied from pymacaroons until the package is available for xenial.
```
elopio: should we be working on packaging this?

sergiusens:
@elopio can you rebase please?

snapcraft/storeapi/macaroons.py
sergiusens: @elopio Update `debian/copyright` for now

snapcraft/tests/fake_servers.py
```diff
@@ -45,14 +46,15 @@ class FakeSSORequestHandler(BaseHTTPRequestHandler):
 
     def do_POST(self):
         parsed_path = urllib.parse.urlparse(self.path)
-        oauth_path = self._API_PATH + 'tokens/oauth'
-        if parsed_path.path.startswith(oauth_path):
-            self._handle_oauth()
+        tokens_discharge_path = self._API_PATH + 'tokens/discharge'
```
sergiusens: should this use a `urllib.parse` or equivalent?

snapcraft/tests/fake_servers.py
```diff
@@ -129,19 +130,39 @@ def __init__(self, server_address):
...
 
-    _DEV_API_PATH = '/dev/'
+    _DEV_API_PATH = '/dev/api/'
 
     def do_POST(self):
         parsed_path = urllib.parse.urlparse(self.path)
-        upload_path = self._DEV_API_PATH + 'click-package-upload/'
-        if parsed_path.path.startswith(upload_path):
+        acl_path = self._DEV_API_PATH + 'acl/'
+        upload_path = self._DEV_API_PATH + 'snap-upload/'
```
sergiusens: ditto wrt concat of url

sergiusens:
So now I am not so eager to add a message on config load to tell the use they need to relogin, this could be treated as a credential invalidation problem just fine.

snapcraft/tests/fake_servers.py
elopio: sure, why not. I just hate urljoin a little because when I miss the trailing / it does things I don't expect. But if the tests pass, all's good.

elopio:
Updated.
About the relogin, I'm ok with no message. Maybe a mention in the release email, and that's it. Thanks for the review.

snapcraft/storeapi/macaroons.py
cjwatson: I've already sorted out a backport; it's awaiting SRU team approval in xenial-proposed/NEW.

snapcraft/storeapi/__init__.py
```diff
@@ -89,6 +89,22 @@ def _get_name_from_snap_file(snap_path):
...
+    :return: A string suitable to use in an Authorization header.
+
+    """
+    root_macaroon_raw = conf.get('macaroon')
+    unbound_raw = conf.get('unbound_discharge')
+    root_macaroon = macaroons.Macaroon.deserialize(root_macaroon_raw)
+    unbound = macaroons.Macaroon.deserialize(unbound_raw)
+    bound = root_macaroon.prepare_for_request(unbound)
+    auth = 'Macaroon root={}, discharge={}'.format(root_macaroon_raw,
+                                                   bound.serialize())
```
cjwatson: This should check that it doesn't break framing.  See the [Launchpad version of this](http://bazaar.launchpad.net/~launchpad-pqm/launchpad/devel/view/head:/lib/lp/snappy/model/snapstoreclient.py#L55).

snapcraft/storeapi/__init__.py
```diff
@@ -280,15 +335,20 @@ def __init__(self, conf):
...
-            self.root_url, 'click-package-upload/{}/'.format(
-                urllib.parse.quote_plus(data['name'])))
-        return session.post(url, data)
-
-    def get(self, download_url):
-        session = common.get_oauth_session(config.Config())
-        return session.get(download_url)
+        auth = _macaroon_auth(self.conf)
+        response = self.post(
+            'snap-upload/', data=json.dumps(data),
```
cjwatson: The endpoint is now called snap-push ([bug report](https://bugs.launchpad.net/snapcraft/+bug/1587165)); there seems little point in adding new code that refers to the old endpoint.

snapcraft/storeapi/__init__.py
facundobatista: Yes, please

snapcraft/storeapi/__init__.py
```diff
@@ -280,15 +335,20 @@ def __init__(self, conf):
             'UBUNTU_STORE_API_ROOT_URL',
             constants.UBUNTU_STORE_API_ROOT_URL))
 
+    def get_macaroon(self, acls):
+        response = self.post(
+            'acl/',
+            data=json.dumps({'permissions': acls}),
+            headers={'Content-Type': 'application/json'})
```
facundobatista: Better if you also send    Accept: application/json   as it may be used in the future (same in other requests)

snapcraft/storeapi/macaroons.py
elopio: aweseme, thanks Colin. I will come back and delete this next week.

snapcraft/storeapi/__init__.py
elopio: ok, I added that to every request that was missing it.

snapcraft/storeapi/__init__.py
elopio: indeed.

snapcraft/storeapi/__init__.py
elopio: @facundobatista what about returning the errors in the response? Should that be an exception too?
Like:
if error:
    result['errors'] = error

Some of these statements come from pindonga's first implementation.

snapcraft/storeapi/__init__.py
elopio: I like a lot better the way that you inject the auth in the request. I hope this code gets out of snapcraft and goes into a shared tool, but if it remains here we can try to copy more things from you.

However there are a couple of details that I don't like about your implementation.
https://bazaar.launchpad.net/~launchpad-pqm/launchpad/devel/view/18086/lib/lp/snappy/model/snapstoreclient.py#L69
Here I think you shouldn't use assert. It can be disabled and you should use it only for internal consistency checks. The value seems to come from somewhere out of your control, so it should be guarded with if -> raise.
The previous check seems to be safe with the assert.

https://bazaar.launchpad.net/~launchpad-pqm/launchpad/devel/view/18086/lib/lp/snappy/tests/test_snapstoreclient.py#L98
Here you are testing only the root macaroon. The second argument, the discharge_macaroon will fail on deserialize, before it gets to the assert.

It made sense to me that any problem reading the macaroon is an InvalidCredentialsError, so I am  throwing an InvalidCredentialsError if deserialize fails with that binascii padding error.
Please let me know what do you think.

elopio:
Thanks for your reviews. The branch is updated.
This is basically a copy&paste from code I didn't write, so any suggestions to improve it are appreciated, don't hold back. We need to land this soon, but we will have time to clean it up during the following weeks. I already have some bugs to report about things I would like to change next week.

snapcraft/storeapi/__init__.py
cjwatson: I don't believe we ever disable asserts in Launchpad production, but I take your point.  I'll fix the issues you've raised; thanks!

A deserialize check is probably good enough in practice, although personally I think it's still worth a low-level paranoia check on specific allowed characters when embedding it in an Authorization header, but perhaps that's just me.

snapcraft/storeapi/__init__.py
facundobatista: It's ok to return the error in the HTTP response. 

The point is how you signal errors through the function layers for finally to store it in the response.

The sensible way to do this is using exceptions, not flagging every function call. I'm happy to provide more details about this if you want, just ping me on IRC.

elopio: @facundobatista I want to make better use of all our exceptions, and just have a couple of clues. When we start working on that I will ask for your input and review. Thanks!

snapcraft/storeapi/__init__.py
elopio: no, I think you are right. I was just trying to do the smallest implementation to get this in landed on time. Trying to implement your suggestion made it obvious that we are not doing a good job handling exceptions. I'm reporting a bug for the exceptions thing, and for the allowed characters check, to be fixed in one or two weeks. Thanks again.

elopio:
retest this please

PR #534, opened 2016-05-30
Wrap plugin list output content

didrocks:
Wrap plugin list output content up to 120 characters depending on terminal size.

Added tests to ensure if detection fails, we are using the 120 characters. as well. Fixes https://bugs.launchpad.net/snapcraft/+bug/1587057.

Note that it's using put cols as a subprocess. A couple of years ago, I had to look into how to get the terminal size on different platforms in a reliable way, and it seems it's the best one without duplicating a fragile logic which evolves over time as new terminals arise.

### DIFF
--- a/snapcraft/main.py
+++ b/snapcraft/main.py
@@ -106,7 +106,8 @@ from docopt import docopt
 
 import snapcraft
 from snapcraft.internal import lifecycle, log
-from snapcraft.internal.common import MAX_CHARACTERS_WRAP
+from snapcraft.internal.common import (format_output_in_columns,
+                                       MAX_CHARACTERS_WRAP)
 
 
 logger = logging.getLogger(__name__)
@@ -134,8 +135,8 @@ def _list_plugins():
             command = ['tput', 'cols']
             width = min(int(subprocess.check_output(command)), width)
 
-    print(textwrap.fill("  ".join(plugins), width=width,
-                        break_on_hyphens=False))
+    for line in format_output_in_columns(plugins, max_width=width):
+        print(line)
 
 
 def _get_project_options(args):
--- a/snapcraft/main.py
+++ b/snapcraft/main.py
@@ -94,16 +94,19 @@ For more help, visit the documentation:
 http://developer.ubuntu.com/snappy/snapcraft
 """
 
+from contextlib import suppress
 import logging
 import pkg_resources
 import pkgutil
 import sys
+import subprocess
 import textwrap
 
 from docopt import docopt
 
 import snapcraft
 from snapcraft.internal import lifecycle, log
+from snapcraft.internal.common import MAX_CHARACTERS_WRAP
 
 
 logger = logging.getLogger(__name__)
@@ -117,9 +120,22 @@ def _get_version():
 
 
 def _list_plugins():
+    plugins = []
     for importer, modname, is_package in pkgutil.iter_modules(
             snapcraft.plugins.__path__):
-        print(modname.replace('_', '-'))
+        plugins.append(modname.replace('_', '-'))
+
+    # we wrap the output depending on terminal size
+    width = MAX_CHARACTERS_WRAP
+    with suppress(OSError):
+        with suppress(subprocess.CalledProcessError):
+            # this is the only way to get current terminal size reliably
+            # without duplicating a bunch of logic
+            command = ['tput', 'cols']
+            width = min(int(subprocess.check_output(command)), width)
+
+    print(textwrap.fill("  ".join(plugins), width=width,
+                        break_on_hyphens=False))
 
 
 def _get_project_options(args):
--- a/snapcraft/main.py
+++ b/snapcraft/main.py
@@ -133,7 +133,9 @@ def _list_plugins():
             # this is the only way to get current terminal size reliably
             # without duplicating a bunch of logic
             command = ['tput', 'cols']
-            width = min(int(subprocess.check_output(command)), width)
+            candidate_width = \
+                subprocess.check_output(command, stderr=subprocess.DEVNULL)
+            width = min(int(candidate_width), width)
 
     for line in format_output_in_columns(plugins, max_width=width):
         print(line)
--- a/snapcraft/tests/test_commands_list_plugins.py
+++ b/snapcraft/tests/test_commands_list_plugins.py
@@ -26,8 +26,9 @@ class ListPluginsCommandTestCase(tests.TestCase):
 
     # plugin list when wrapper at MAX_CHARACTERS_WRAP
     default_plugin_output = (
-        'ant        catkin  copy  jdk     kernel  maven  nodejs   python3  tar-content\n'  # noqa
-        'autotools  cmake   go    kbuild  make    nil    python2  scons  \n')  # noqa
+        'ant        catkin  copy  jdk     kernel  maven  nodejs   python3  '
+        'tar-content\n'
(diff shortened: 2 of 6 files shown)


didrocks:
Note: I saw the bug report is more about column alignement than wrapping text. This can be used as a base. I'll look later this week if there is an easy way to reuse existing components for this (so, please do not merge yet)

snapcraft/main.py
```diff
@@ -117,9 +120,22 @@ def _get_version():
...
-        print(modname.replace('_', '-'))
+        plugins.append(modname.replace('_', '-'))
+
+    # we wrap the output depending on terminal size
+    width = MAX_CHARACTERS_WRAP
+    with suppress(OSError):
+        with suppress(subprocess.CalledProcessError):
+            # this is the only way to get current terminal size reliably
+            # without duplicating a bunch of logic
+            command = ['tput', 'cols']
```
sergiusens: Surprisingly I would of thought python had something built in for this

sergiusens:
Thanks for getting this started! As requested I will wait.

snapcraft/main.py
didrocks: Isn't it? :)
I remember to have been surprised by this as well a year ago, when looking for it, and after some digging into the web, that was the only way to get this done. I just did a quick look again yesterday, and nothing changed in between

didrocks:
I just rebased on latest master and added a couple of commits to now have the "by column" functionality (there is no built-in module to do that), but the logic was quite easy to implement with the required tests of course :)

sergiusens:
retest this please

snapcraft/tests/test_commands_list_plugins.py
```diff
@@ -23,25 +24,45 @@
...
+        self.assertEqual(mock_stdout.getvalue(), expected_output)
+
+    @mock.patch('sys.stdout', new_callable=io.StringIO)
+    @mock.patch('subprocess.check_output')
+    def test_list_plugins_error_invalid_terminal_size(self, mock_subprocess,
+                                                      mock_stdout):
+        def raise_error(x):
+            raise OSError()
+        mock_subprocess.side_effect = raise_error
+        main(['list-plugins'])
```
sergiusens: oh, there's a similar test in `integration_tests` reason for adt to fail most likely

`runtests.sh integration` ftw ;-)

Or `python3 -m unittest integration_tests ...`

sergiusens:
Nice tests!

didrocks:
Rebase on master to have correct coverage report (which should increase) and rerunning autopkgtests

snapcraft/tests/test_commands_list_plugins.py
```diff
@@ -23,25 +24,45 @@
 
 class ListPluginsCommandTestCase(tests.TestCase):
 
+    # plugin list when wrapper at MAX_CHARACTERS_WRAP
+    default_plugin_output = (
+        "ant        catkin  copy  jdk     kernel  maven  nodejs   python3  tar-content\n"  # noqa
+        "autotools  cmake   go    kbuild  make    nil    python2  scons  \n")  # noqa
```
elopio: For consistency, please use single quotes for strings.
And it hurts me seeing python code with more than 80 lines, I'm weird that way. Would it be too ugly to split the string before the \n?
'ant        catkin  copy  jdk     kernel  maven  nodejs   '
'python3  tar-content\n'
'autotools  cmake   go    kbuild  make    nil    '
'python2  scons  \n'

snapcraft/tests/test_commands_list_plugins.py
```diff
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright (C) 2015 Canonical Ltd
+# Copyright (C) 2015-2016 Canonical Ltd
```
elopio: legal nit: this should be 2015, 2016

snapcraft/tests/test_common.py
```diff
@@ -101,3 +101,45 @@ def test_arch_triplet_migration_message(self):
         self.assertEqual(
             str(raised.exception),
             "This plugin is outdated, use 'project.arch_triplet'")
+
+
+class FormatInColumnsTestCase(tests.TestCase):
+
+    elements_list = ["ant", "autotools", "catkin", "cmake", "copy", "go",
+                     "jdk", "kbuild", "kernel", "make", "maven", "nil",
+                     "nodejs", "python2", "python3", "scons", "tar-content"]
```
elopio: single quotes please.

snapcraft/tests/test_common.py
```diff
@@ -101,3 +101,45 @@ def test_arch_triplet_migration_message(self):
...
+
+    def test_format_output_in_columns_default(self):
+        '''Format output on 2 lines, with default max-width and space sep'''
```
elopio: double quotes for comments :p

snapcraft/tests/test_commands_list_plugins.py
didrocks: I'm a fan of 120 characters, but let's not open this debate here :p

Here, I used the #noqa stenza as the important part is to be able to check visually as well that we have a column output, making clear the test is about that. I have no strong opinion if you think it's still readable when being broken in multiple lines, just tell me :)

didrocks:
Just pushed the fixes (minus the 80 characters one, waiting on your feedback), thanks for the review!

didrocks:
The last commit should fix the autopkgtest failure (it does at least when running adt-run locally) :)

didrocks:
it seems that adt running in jenkins has other variables set that me running adt-run (with qemu) directly, do you know the difference?

The output is:
testtools.matchers._impl.MismatchError: 'ant\nautotools\ncatkin\ncmake\ncopy\ngo\njdk\nkbuild\nkernel\nmake\nmaven\nnil\nnodejs\npython2\npython3\nscons\ntar-content' not in 'ant        catkin  copy  jdk     kernel  maven  nodejs   python3  tar-content\nautotools  cmake   go    kbuild  make    nil    python2  scons  \n'

Which kinds of puzzle me as this is the old output (and it can only happen if tput cols returns less than 20). It seems to me that it's the older snapcraft version which runs, are we sure that the adt runner is building and running a new snapcraft package first (as we are using SNAPCRAFT_FROM_INSTALLED=1 using the "snapcraft" command)?

didrocks:
I fixed the second integration test, but it seems to me to be mostly a duplicate of the first one (running in a project directory though, so unsure if that influences, but there is no local plugin or plugin difference). Maybe time to drop it?

sergiusens:
El 01/06/16 a las 03:42, Didier Roche escribió:
> Just pushed the fixes (minus the 80 characters one, waiting on your
> feedback), thanks for the review!

elopio won this battle long ago :-P
I just keep QA happy fwiw ;-)

didrocks:
pushed commit to keep QA happy :)

elopio:
I saw your IRC message and I lol'd.

I'm happy :+1:

elopio:
Also +1 on dropping that test_main.py

snapcraft/main.py
sergiusens: Seems we need to add `ncurses-bin` to `debian/control`

sergiusens:
If the integration_test is useless, let's drop it!

If `ncurses-bin` is indeed necessary, please add, feel free to merge then or ping me and I'll merge.

sergiusens:
oh, this is an essential package, nvm

PR #536, opened 2016-05-30
Always set prefix in the autotools plugin

attente:
Some codebases use prefix, libdir, datadir, etc. to locate installed resources at runtime. Since those paths are accessed at runtime, the prefix always needs to match the final mounted location of the snap.

### DIFF
--- a/snapcraft/plugins/autotools.py
+++ b/snapcraft/plugins/autotools.py
@@ -32,7 +32,7 @@ In additon, this plugin uses the following plugin-specific keywords:
       (list of strings)
       configure flags to pass to the build such as those shown by running
       './configure --help'
-    - install-via:
+    - (deprecated) install-via:
       (enum, 'destdir' or 'prefix')
       Whether to install via DESTDIR or by using --prefix (default is
       'destdir')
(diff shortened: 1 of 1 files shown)


snappy-m-o:

snappy-m-o:

sergiusens:
ok to test

snapcraft/plugins/autotools.py
```diff
@@ -32,15 +32,15 @@
       (list of strings)
       configure flags to pass to the build such as those shown by running
       './configure --help'
-    - install-via:
+    - (deprecated) install-via:
```
kyrofa: Hmm. This doesn't look deprecated here, it looks completely removed :stuck_out_tongue: . This exists for the (surprisingly abundant) projects that don't respect DESTDIR; we need to make sure we continue to have a solution for them.

kyrofa:
This PR as it currently stands will break a lot of projects. I also don't like hard-coding `/snap/<name>/current` in here. If this functionality is indeed required then perhaps it should be another `install-via` enum rather than breaking the two existing ones.

sergiusens:
I'm closing this for now as it would break the existing contract with many builders.

PR #537, opened 2016-05-31
Add support for parsing the parts wiki.

josepht:
* Parse the parts wiki and store the results in a compressed YAML file.

LP: #1587583

### DIFF
--- a/bin/snapcraft-parser
+++ b/bin/snapcraft-parser
@@ -18,7 +18,7 @@
 import os
 import sys
 
-# make running from bzr snapshot easy
+# make running from source easy
 topdir = os.path.abspath(os.path.join(__file__, '..', '..'))
 if os.path.exists(os.path.join(topdir, 'setup.py')):
     sys.path = [topdir] + sys.path
--- a/bin/snapcraft-parser
+++ b/bin/snapcraft-parser
@@ -1,0 +1,29 @@
+#!/usr/bin/python3
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2016 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+import sys
+
+# make running from bzr snapshot easy
+topdir = os.path.abspath(os.path.join(__file__, '..', '..'))
+if os.path.exists(os.path.join(topdir, 'setup.py')):
+    sys.path = [topdir] + sys.path
+
+
+if __name__ == '__main__':
+    import snapcraft_parser.main
+    snapcraft_parser.main.main()
--- a/snapcraft_parser/main.py
+++ b/snapcraft_parser/main.py
@@ -79,8 +79,16 @@ def main(argv=None):
         sys.exit(textwrap.fill(str(e)))
 
 
+def _get_namespaced_partname(key, partname):
+    return "{key}{sep}{partname}".format(
+        key=key,
+        sep=PART_NAMESPACE_SEP,
+        partname=partname,
+    )
+
+
 def _update_after_parts(partname, after_parts):
-    return [_namespaced_partname(partname, x) for x in after_parts]
+    return [_get_namespaced_partname(partname, x) for x in after_parts]
 
 
 def _get_origin_data(origin_dir):
@@ -98,9 +106,9 @@ def _process_subparts(project_part, subparts, parts):
         source_part = parts.get(part)
         # TODO: get any dependent parts here as well using
         # 'project_part' to namespace them.
-        if source_part is not None:
-            parts_list[_namespaced_partname(project_part,
-                                            part)] = source_part
+        if source_part:
+            namespaced_part = _get_namespaced_partname(project_part, part)
+            parts_list[namespaced_part] = source_part
             after = source_part.get("after", [])
 
             if after:
@@ -130,7 +138,7 @@ def _process_index(output):
         project_part = data[key].get("project-part")
         subparts = data[key].get("parts", [])
 
-        if origin is not None:
+        if origin:
             # TODO: this should really be based on the origin uri not
             # the part name to avoid the situation where there are multiple
             # parts being pulled from the same repo branch.
@@ -146,7 +154,7 @@ def _process_index(output):
             options = Options()
             setattr(options, "source", origin)
 
-            if origin_type is not None:
+            if origin_type:
                 setattr(options, "source_type", origin_type)
 
             sources.get(origin_dir, None, options)
@@ -156,7 +164,7 @@ def _process_index(output):
             parts = origin_data.get('parts', {})
 
             source_part = parts.get(project_part)
-            if source_part is not None:
+            if source_part:
                 after = source_part.get('after', [])
 
                 if after:
@@ -182,7 +190,7 @@ def run(args):
         path = PARTS_FILE
 
     index = args.get('--index')
-    if index is not None:
+    if index:
         if "://" not in index:
             index = "%s%s" % ("file://", os.path.join(os.getcwd(), index))
         output = urllib.request.urlopen(index).read()
@@ -215,14 +223,6 @@ def _write_parts_list(path, master_parts_list):
     with open(path, "w") as fp:
         fp.write(yaml.dump(master_parts_list,
                             default_flow_style=False))
-
-
-def _namespaced_partname(key, partname):
-    return "{key}{sep}{partname}".format(
-        key=key,
-        sep=PART_NAMESPACE_SEP,
-        partname=partname,
-    )
 
 
 if __name__ == '__main__':  # pragma: no cover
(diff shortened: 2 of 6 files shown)


snappy-m-o:

snappy-m-o:

snapcraft_parser/main.py
```diff
@@ -0,0 +1,212 @@
...
+
+        sys.exit(textwrap.fill(str(e)))
+
+
+def _update_after_parts(partname, after_parts):
+    return [_namespaced_partname(partname, x) for x in after_parts]
+
+
+def _get_output():
+    # TODO: pull the part origin list from the wiki
```
sergiusens: maybe make this a command line switch (what input to give it) so we can pass it hand crafted files (to test easily) or actual urls (to test or the official one)

snapcraft_parser/main.py
```diff
@@ -0,0 +1,212 @@
...
+
+def _get_origin_data(origin_dir):
+    origin_data = {}
+    with open(os.path.join(origin_dir, 'snapcraft.yaml'), "r") as fp:
+        origin_data = yaml.load(fp)
+
+    return origin_data
+
+
+def run(args):
```
sergiusens: if we parametrize `_get_output` as part of `run` it would make the tests easier (less mocking).

snapcraft_parser/main.py
```diff
@@ -0,0 +1,212 @@
...
+from snapcraft.internal import log, sources
+
+
+logger = logging.getLogger(__name__)
+
+
+PART_NAMESPACE_SEP = '/'
+# TODO: make this a temporary directory that get's removed when finished
+BASE_DIR = "/tmp"
+PARTS_FILE = os.path.join(BASE_DIR, "snap-parts.yaml.gz")
```
sergiusens: maybe this should be `os.curdir` and overridable by providing `--output`

bin/snapcraft-parser
```diff
@@ -0,0 +1,29 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+import sys
+
+# make running from bzr snapshot easy
```
sergiusens: running from source (instead of bzr snapshot)

sergiusens:
One thing to not is that it would be good if the unit tests did not rely on network access. These are mostly fine for integration tests though

sergiusens:
ok to test

sergiusens:
I am seeing this when running myself:
```
$ ./bin/snapcraft-parser 
No revisions or tags to pull.                                                  
Part 'main/part2' is referenced by another part but not included in the master parts list
```

snapcraft_parser/main.py
```diff
@@ -0,0 +1,212 @@
...
+Options:
+  -h --help                             show this help message and exit
+  -p --print                            print the master parts list
+  -v --version                          show program version and exit
+  -d --debug                            print debug information while executing
+                                        (including backtraces)
+
+
+For more help, visit the documentation:
+http://developer.ubuntu.com/snappy/snapcraft
```
elopio: we don't have any information about parser in that URL.

snapcraft_parser/main.py
elopio: and if we will use a temp dir, it's better not to hardcode it. We can use python's tempfile.

snapcraft_parser/main.py
```diff
@@ -0,0 +1,212 @@
...
+        sys.exit(textwrap.fill(str(e)))
+
+
+def _update_after_parts(partname, after_parts):
+    return [_namespaced_partname(partname, x) for x in after_parts]
+
+
+def _get_output():
+    # TODO: pull the part origin list from the wiki
+    output = """
```
elopio: s/output/OUTPUT

snapcraft_parser/main.py
```diff
@@ -0,0 +1,212 @@
...
+                if source_part is not None:
+                    parts_list[_namespaced_partname(project_part,
+                                                    part)] = source_part
+
+        # after_parts should already be namespaced at this point.
+
+        if _valid_parts_list(parts_list, after_parts):
+            master_parts_list.update(parts_list)
+
+    _write_gzipped_parts_list(BASE_DIR, master_parts_list)
```
elopio: This run function is too long. Can you split it in smaller functions with nice names?

snapcraft_parser/main.py
josepht: The "--index" parameter has been added and removed the need for _get_output().

snapcraft_parser/main.py
josepht: Done. "--index"

snapcraft_parser/main.py
josepht: I've added the "--output" flag which allows the user to provide the location of the parts list and the default is now the current directory.

bin/snapcraft-parser
josepht: Fixed.

snapcraft_parser/main.py
josepht: Removed since this isn't really a script the average snapcraft user cares about.

snapcraft_parser/main.py
josepht: Removed in favor of the "--index" parameter.

snapcraft_parser/main.py
josepht: Done.  _process_subparts() feels a bit forced though.

josepht:
I think I've addressed all the review comments up to now.  Please have another look when you have a few minutes.

snapcraft_parser/main.py
```diff
@@ -0,0 +1,230 @@
...
+    return True
+
+def _write_gzipped_parts_list(path, master_parts_list):
+    logging.debug("Writing gzipped parts list to %s", path)
+    with gzip.open(path, "w") as fgp:
+        fgp.write(yaml.dump(master_parts_list,
+                            default_flow_style=False).encode("utf-8"))
+
+
+def _namespaced_partname(key, partname):
```
sergiusens: @elopio will probably mention the PEP about using a `get` in front of this method.

fwiw, I'd like this moved to right above the list comprehension that uses it for easier reading.

snapcraft_parser/main.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+def _process_subparts(project_part, subparts, parts):
+    after_parts = set()
+    parts_list = {}
+    for part in subparts:
+        source_part = parts.get(part)
+        # TODO: get any dependent parts here as well using
+        # 'project_part' to namespace them.
+        if source_part is not None:
```
sergiusens: this can be reduced to `if source_part:`

snapcraft_parser/main.py
```diff
@@ -0,0 +1,230 @@
...
+            parts_list[_namespaced_partname(project_part,
```
sergiusens: sorry for being dense here, but do you mind doing

```python
namespaced_part = _get_namespaced_partname(project_part, part)
part_list[namespaced_part] = source_part
```
?

snapcraft_parser/tests/test_parser.py
```diff
@@ -0,0 +1,284 @@
...
+                "main": {
+                    "source": "lp:something",
+                    "plugin": "copy",
+                    "files": ["file1", "file2"],
+                },
+            }
+        }
+
+        filename = "parts.gz"
+        try:
```
sergiusens: the `try` `except` here should not be needed, these tests are always run in a new tempdir for each test as part of the fixture setup.

snapcraft_parser/main.py
josepht: Done and done.

snapcraft_parser/main.py
josepht: Done.  In several other places as well.

snapcraft_parser/main.py
josepht: Not at all. Done.

snapcraft_parser/tests/test_parser.py
josepht: Removed.

snapcraft_parser/main.py
elopio: It will feel forced until you get used to the right way ;)
I'm joking, of course. When I finish my comments with a question mark, I'm honestly asking what do you think about my proposal. Feel free to disagree. If I have a strong opinion on something, I will let you know.

Personally, I like the function a little better now and if I were the one writing the code, I would try to make even more internal functions. Not a blocker, of course, just saying.

elopio:
Should this be added to the scripts in setup.py? I think that's what makes it available in the deb as an executable.

sergiusens:
@elopio @josepht well if you do that we now need to create a `debian/control` binary package for it ;-)

sergiusens:
something like

```diff
diff --git a/debian/control b/debian/control
index cd3f9cb..72c658b 100644
--- a/debian/control
+++ b/debian/control
@@ -56,6 +56,13 @@ Description: easily craft snaps
  natural workflow of an upstream to be extended with a simple release step
  into Snappy.
 
+Package: snapcraft-parser
+Architecture: all
+Depends: snapcraft (= ${binary:Version}), ${misc:Depends}, ${python3:Depends}
+Description: the cloud parser for shared snapcraft parts.
+  The parser is in charge of getting all the shared parts out there
+  indexed into a common cacheable file.
+
 Package: snapcraft-examples
 Architecture: all
 Depends: snapcraft (= ${binary:Version}), ${misc:Depends}, ${python3:Depends}
```

and a `debian/snapcraft-parser.install` file

josepht:
@elopio @sergiusens I've added the bits to debian/control and added a copy of debian/snapcraft.install as debian/snapcraft-parser.install.

sergiusens:
I created #545 to deal with a couple of packaging related bits (python and debian)

sergiusens:
Closing as #545 landed

PR #538, opened 2016-05-31
Support setting a gopath for a go project from vcs

sergiusens:
LP: #1583426

### DIFF
--- a/snapcraft/plugins/go.py
+++ b/snapcraft/plugins/go.py
@@ -31,6 +31,12 @@ Additionally, this plugin uses the following plugin-specific keywords:
       are pulled in automatically by `go get`.
       Packages that are not "main" will not cause an error, but would
       not be useful either.
+
+    - gopath:
+      (string)
+      When using a source entry, we sometimes need the code to live in a
+      certain location within the `GOPATH`. This has the format of a go
+      import path which would be go gettable.
 """
 
 import os
@@ -53,6 +59,10 @@ class GoPlugin(snapcraft.BasePlugin):
             },
             'default': [],
         }
+        schema['properties']['gopath'] = {
+            'type': 'string',
+            'default': ''
+        }
 
         if 'required' in schema:
             del schema['required']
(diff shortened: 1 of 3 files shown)


sergiusens:
To make the project defined in the bug work we need to add an extra key

`gopath: github.com/hashicorp/vault`

elopio:
lgtm, and it's awesome if it makes vault's yaml nicer.
But I still don't have it clear when I should use this value. Is it because I'm being dense?

sergiusens:
@elopio in the vault code, the `package main` has a `// $gopath` in it. it should not be needed in general

sergiusens:
@elopio I took this concept from the `dh_golang` helper. It could be used for every case and we can make it madatory but I don't want to break backwards compatibility

elopio:
huh, I think I understand now. :+1: 
I'll go and play a little more with vault.

snapcraft/plugins/go.py
```diff
@@ -31,6 +31,12 @@
       are pulled in automatically by `go get`.
       Packages that are not "main" will not cause an error, but would
       not be useful either.
+
+    - gopath:
+      (string)
+      When using a source entry, we sometimes need the code to live in a
+      certain location within the `GOPATH`. This has the format of a go
+      import path which would be go gettable.
```
kyrofa: `GOPATH` means something already, I feel like this might be confusing. Perhaps we should call it `importpath` or something similar.

snapcraft/plugins/go.py
```diff
@@ -31,14 +31,24 @@
       are pulled in automatically by `go get`.
       Packages that are not "main" will not cause an error, but would
       not be useful either.
+
+    - go-importpath:
+      (string)
+      This entry tells the checked out `source` to live within a certain path
+      within `GOPATH`.
+      This is not needed for `go-packages`.
```
kyrofa: Ah, yeah I like this much better. Good description as well.

PR #540, opened 2016-06-01
Add a ticker for snapping

sergiusens:
This replaces the mksquashfs output which is rather verbose.

LP: #1582955


snapcraft/internal/lifecycle.py
```diff
@@ -254,15 +256,35 @@ def snap(project_options, directory=None, output=None):
...
+            stderr=subprocess.STDOUT) as proc:
+        ret = None
+        if os.isatty(sys.stdout.fileno()):
+            ret = proc.poll()
+            while ret is None:
+                print('\033[0;32m\rSnapping {!r}\033[0;32m {}'.format(
+                    snap['name'], ticker[i]), end='')
+                i = (i+1) % len(ticker)
+                time.sleep(.2)
+                ret = proc.poll()
```
kyrofa: In other parts of the code (I'm thinking of the upload specifically) we used `progressbar` for this. Should we maintain consistency?

snapcraft/tests/test_commands_snap.py
```diff
@@ -40,16 +41,63 @@ class SnapCommandTestCase(tests.TestCase):
       plugin: nil
 """
 
+    def setUp(self):
+        super().setUp()
+
+        patcher = mock.patch('snapcraft.internal.lifecycle.Popen',
+                             new=mock.Mock(wraps=subprocess.Popen))
```
elopio: nice!

snapcraft/tests/test_commands_snap.py
```diff
@@ -40,16 +41,63 @@ class SnapCommandTestCase(tests.TestCase):
...
+        self.popen_mock = patcher.start()
```
elopio: we can start being obnoxious about terminology here now. If you call it self.popen_spy then it will be a little clearer in the test that you are keeping the original behaviour and just recording the calls.

snapcraft/tests/test_commands_snap.py
```diff
@@ -40,16 +41,63 @@ class SnapCommandTestCase(tests.TestCase):
...
+        self.addCleanup(patcher.stop)
+
+        patcher = mock.patch('sys.stdout.fileno')
```
elopio: I wouldn't mind a comment here about why we patch fileno.

snapcraft/tests/test_commands_snap.py
```diff
@@ -40,16 +41,63 @@ class SnapCommandTestCase(tests.TestCase):
...
+
+        self.assertTrue(os.path.exists(self.stage_dir),
+                        'Expected a stage directory')
+
+        self.verify_state('part1', self.state_dir, 'strip')
+
+        self.popen_mock.assert_called_once_with([
+            'mksquashfs', self.snap_dir, 'snap-test_1.0_amd64.snap',
+            '-noappend', '-comp', 'xz', '-no-xattrs', '-all-root'],
+            stderr=subprocess.STDOUT, stdout=subprocess.PIPE)
```
elopio: we get a .snap file at the end, right? We can assert that it exists.

elopio:
:+1:

sergiusens:
@elopio check now!

snapcraft/tests/__init__.py
```diff
@@ -36,7 +37,8 @@ def __init__(self, source=None, source_type=None, source_branch=None,
         self.source_subdir = source_subdir
 
 
-class TestCase(testscenarios.WithScenarios, fixtures.TestWithFixtures):
+class TestCase(testtools.TestCase, testscenarios.WithScenarios,
+               fixtures.TestWithFixtures):
```
elopio: you can remove TestWithFixtures. Testtools comes with a useFixture method.

PR #543, opened 2016-06-02
Rename strip to prime

sergiusens:
LP: #1582515

### DIFF
--- a/snapcraft/internal/states/_prime_state.py
+++ b/snapcraft/internal/states/_prime_state.py
@@ -19,15 +19,15 @@ import yaml
 from snapcraft.internal.states._state import State
 
 
-def _strip_state_constructor(loader, node):
+def _prime_state_constructor(loader, node):
     parameters = loader.construct_mapping(node)
-    return StripState(**parameters)
+    return PrimeState(**parameters)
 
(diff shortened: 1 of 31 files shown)


snapcraft/internal/states/_prime_state.py
```diff
@@ -19,15 +19,15 @@
...
+    return PrimeState(**parameters)
 
-yaml.add_constructor(u'!StripState', _strip_state_constructor)
+yaml.add_constructor(u'!PrimeState', _prime_state_constructor)
 
 
-class StripState(State):
-    yaml_tag = u'!StripState'
+class PrimeState(State):
+    yaml_tag = u'!PrimeState'
```
kyrofa: This will of course not work on a tree from an old snapcraft (it'll cause the "missing state" error to be raised), but I assume you know that.

kyrofa:
This isn't a small change, but it's pretty well covered by unit and integration tests, so I'm happy with it :+1: .

snapcraft/internal/states/_prime_state.py
sergiusens: yeah, this is what I mentioned in the standup. It would be as if it never existed, given that it is an easy step to get back on track I thought I'd let it be

PR #545, opened 2016-06-03
Feature/1587583/part parser

sergiusens:


snapcraft/internal/parser.py
```diff
@@ -204,45 +205,44 @@ class Options():
...
 
+
 def _write_parts_list(path, master_parts_list):
-    logging.debug("Writing parts list to %s", path)
-    with open(path, "w") as fp:
+    logging.debug('Writing parts list to {!r}'.format(path))
+    with open(path, 'w') as fp:
         fp.write(yaml.dump(master_parts_list,
-                            default_flow_style=False))
+                 default_flow_style=False))
```
josepht: Is this the right indentation?  I thought since it's part of the yaml.dump() call it should be indented with yaml.dump()'s parameters.

sergiusens: ah, either multiple of 4 or right nder the paren. static tests failed here and work now though

PR #551, opened 2016-06-06
Reenable get and set tourdir unit tests

didrocks:
The existing tests (also for plugin dir) were running order dependant.
Ensure we reset the state to default for all internal variables.

### DIFF
--- a/snapcraft/tests/test_common.py
+++ b/snapcraft/tests/test_common.py
@@ -23,6 +23,14 @@ from snapcraft import tests
 
 class CommonTestCase(tests.TestCase):
 
+    def setUp(self):
+        super().setUp()
+        # reset all default paths
+        common._plugindir = common._DEFAULT_PLUGINDIR
+        common._schemadir = common._DEFAULT_SCHEMADIR
+        common._librariesdir = common._DEFAULT_LIBRARIESDIR
+        common._tourdir = common._DEFAULT_TOURDIR
+
     def test_get_default_plugindir(self):
         self.assertEqual(
             '/usr/share/snapcraft/plugins', common.get_plugindir())
--- a/snapcraft/tests/test_common.py
+++ b/snapcraft/tests/test_common.py
@@ -25,7 +25,10 @@ class CommonTestCase(tests.TestCase):
 
     def setUp(self):
         super().setUp()
-        # reset all default paths
+        self.addCleanup(self._reset_default_paths)
+
+    def _reset_default_paths(self):
+        """Reset all default paths"""
         common._plugindir = common._DEFAULT_PLUGINDIR
         common._schemadir = common._DEFAULT_SCHEMADIR
         common._librariesdir = common._DEFAULT_LIBRARIESDIR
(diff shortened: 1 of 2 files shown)


elopio:
retest this please

snapcraft/tests/test_common.py
```diff
@@ -23,6 +23,14 @@
 
 class CommonTestCase(tests.TestCase):
 
+    def setUp(self):
+        super().setUp()
+        # reset all default paths
+        common._plugindir = common._DEFAULT_PLUGINDIR
+        common._schemadir = common._DEFAULT_SCHEMADIR
+        common._librariesdir = common._DEFAULT_LIBRARIESDIR
+        common._tourdir = common._DEFAULT_TOURDIR
```
elopio: Shouldn't this be reset in the tearDown instead?
You could do that nicely with something like:

self.addCleanup(self._reset_dirs), and move the four statements to the _reset_dirs method.

elopio:
https://bugs.launchpad.net/snapcraft/+bug/1589782

didrocks:
I always found better in all my projects to reset this in setUp() instead of tearDown(), as it's one of the requirements, but yeah, I'm fine with both :)

_reset_dirs isn't useful for now (first, it will need to be a dictionary, as each dir as a different default) and we only reset them in one single place.

Thoughts?

elopio:
_reset_dirs would be a method, a callable that addCleanup registers to be executed on tear down.

```
diff --git a/snapcraft/tests/test_common.py b/snapcraft/tests/test_common.py
index 38e43ec..229bbfb 100644
--- a/snapcraft/tests/test_common.py
+++ b/snapcraft/tests/test_common.py
@@ -25,7 +25,9 @@ class CommonTestCase(tests.TestCase):
 
     def setUp(self):
         super().setUp()
-        # reset all default paths
+        self.addCleanup(self._reset_default_paths)
+
+    def _reset_default_paths(self):
         common._plugindir = common._DEFAULT_PLUGINDIR
         common._schemadir = common._DEFAULT_SCHEMADIR
         common._librariesdir = common._DEFAULT_LIBRARIESDIR
```

didrocks:
Oh, I got what you mean now :)

Sure, sounds good, just pushed the change, (tests still passing, phew) thanks!

didrocks:
retest this please

snapcraft/tests/test_common.py
```diff
@@ -23,6 +23,17 @@
 
 class CommonTestCase(tests.TestCase):
 
+    def setUp(self):
+        super().setUp()
+        self.addCleanup(self._reset_default_paths)
+
+    def _reset_default_paths(self):
```
sergiusens: @elopio why not here? https://github.com/ubuntu-core/snapcraft/blob/master/snapcraft/tests/__init__.py#L50

elopio: sure, that works too. Inherit from the base test case, and add any paths it could be missing.

didrocks: Good catch! I have used that opportunity to replace the duplicate `    self.addCleanup(common.set_schemadir, common.get_schemadir())` with the tourdir cleanup. I added the librariedir cleanup that wasn't there too. Just pushed the new revision.

didrocks:
(just rebase on master for the sake of it with last comments)

PR #555, opened 2016-06-07
Make app names more restrictive.

kyrofa:
Until recently snapd was too permissive on app names, allowing them to potentially cause issues in the system (e.g. interfering with hook names, or resulting in invalid AppArmor profile names, etc.). This has been changed in snapd, and this PR resolves LP: [#1589613](https://bugs.launchpad.net/ubuntu/+source/snapd/+bug/1589613) by fixing it in snapcraft as well.

Note that this PR does not change the validation error when using an invalid app name, which is slightly dreadful, but requires a more in-depth refactor.

### DIFF
--- a/schema/snapcraft.yaml
+++ b/schema/snapcraft.yaml
@@ -81,7 +81,7 @@ properties:
     type: object
     additionalProperties: false
     patternProperties:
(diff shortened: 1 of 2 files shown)


sergiusens:
retest this please

PR #560, opened 2016-06-09
Add snapcraft-parser integration test.

josepht:
Add a "Happy Path" integration test for snapcraft-parser.
LP: #1590268

### DIFF
--- a/integration_tests/test_parser.py
+++ b/integration_tests/test_parser.py
@@ -17,21 +17,19 @@
 import os
 import subprocess
 
-from snapcraft.tests import TestCase
+import integration_tests
 
 
-class TestParser(TestCase):
-    '''Test bin/snapcraft-parser'''
+class TestParser(integration_tests.TestCase):
+    """Test bin/snapcraft-parser"""
 
     def test_parser_basic(self):
-        '''Test snapcraft-parser basic usage'''
-        src_path = os.path.join(os.path.dirname(os.path.abspath(__file__)),
-                                '..', 'bin', 'snapcraft-parser')
+        """Test snapcraft-parser basic usage"""
 
-        args = ['python3', src_path, '--index',
+        args = [self.snapcraft_parser_command, '--index',
                 'https://wiki.ubuntu.com/snapcraft/parts?action=raw',
                 '--output', 'parts.yaml']
-        with open(os.path.devnull, 'w') as devnull:
-            subprocess.check_call(args, stderr=devnull, stdout=devnull)
+        subprocess.check_call(args, stderr=subprocess.DEVNULL,
+                              stdout=subprocess.DEVNULL)
 
         self.assertTrue(os.path.exists('parts.yaml'))
--- a/integration_tests/test_parser.py
+++ b/integration_tests/test_parser.py
@@ -1,0 +1,37 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2016 Canonical Ltd
+#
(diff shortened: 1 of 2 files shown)


snappy-m-o:

snappy-m-o:

integration_tests/test_parser.py
```diff
@@ -0,0 +1,37 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+import subprocess
+
+from snapcraft.tests import TestCase
+
+
+class TestParser(TestCase):
+    '''Test bin/snapcraft-parser'''
```
elopio: For consistency, the comments are with three double quotes.

integration_tests/test_parser.py
```diff
@@ -0,0 +1,37 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+import subprocess
+
+from snapcraft.tests import TestCase
+
+
+class TestParser(TestCase):
```
elopio: This should inherit from integration_tests.TestCase

integration_tests/test_parser.py
```diff
@@ -0,0 +1,37 @@
...
+    '''Test bin/snapcraft-parser'''
+
+    def test_parser_basic(self):
+        '''Test snapcraft-parser basic usage'''
+        src_path = os.path.join(os.path.dirname(os.path.abspath(__file__)),
+                                '..', 'bin', 'snapcraft-parser')
```
elopio: We need this test also to run from the installed path in autopkgtests. Take a look at integration_tests/__init__.py and add a self.snapcraft_parser_command.

integration_tests/test_parser.py
```diff
@@ -0,0 +1,37 @@
...
+
+        args = ['python3', src_path, '--index',
```
elopio: No need for python3 there.

elopio:
Great, thanks.
I think I would like to get this to check parts of the output, but I don't like the current output. And I think I would like to check parts of the resulting file. Both things would fail if the index changes, so maybe it would be better to add a test later with a local index, and leave the one accessing the wiki simple. :+1: after the comments are fixed.

integration_tests/test_parser.py
```diff
@@ -0,0 +1,37 @@
...
+                'https://wiki.ubuntu.com/snapcraft/parts?action=raw',
+                '--output', 'parts.yaml']
+        with open(os.path.devnull, 'w') as devnull:
+            subprocess.check_call(args, stderr=devnull, stdout=devnull)
```
elopio: use subprocess.DEVNULL

integration_tests/test_parser.py
josepht: Fixed.

integration_tests/test_parser.py
josepht: Fixed.

integration_tests/test_parser.py
josepht: Fixed.

integration_tests/test_parser.py
josepht: Removed

josepht:
@elopio I've updated based on your comments.  Thanks for the review.

elopio:
:+1:

snappy-m-o:

snappy-m-o:

sergiusens:
OK to test

PR #565, opened 2016-06-10
Include 'maintainer' and 'description' in the parser output.

josepht:
LP: #1591199

### DIFF
--- a/snapcraft/internal/parser.py
+++ b/snapcraft/internal/parser.py
@@ -116,7 +116,8 @@ def _update_source(part, origin):
     return part
 
 
-def _process_subparts(project_part, subparts, parts, origin):
+def _process_subparts(project_part, subparts, parts, origin, maintainer,
+                      description):
     after_parts = set()
     parts_list = {}
     for part in subparts:
@@ -125,6 +126,8 @@ def _process_subparts(project_part, subparts, parts, origin):
         # 'project_part' to namespace them.
         if source_part:
             source_part = _update_source(source_part, origin)
+            source_part['maintainer'] = maintainer
+            source_part['description'] = description
 
             namespaced_part = _get_namespaced_partname(project_part, part)
             parts_list[namespaced_part] = source_part
@@ -157,6 +160,8 @@ def _process_index(output):
         origin_type = data.get('origin-type')
         project_part = data.get('project-part')
         subparts = data.get('parts', [])
+        maintainer = data.get('maintainer', '')
+        description = data.get('description', '')
 
         if origin:
             # TODO: this should really be based on the origin uri not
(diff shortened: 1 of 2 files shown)


snappy-m-o:

snappy-m-o:

snapcraft/tests/test_parser.py
```diff
@@ -408,3 +408,46 @@ def test_n_documents(
...
+                'main2': {
+                    'source': 'lp:project',
+                    'plugin': 'copy',
+                    'files': ['file1', 'file2'],
+                },
+            }
+        }
+        main(['--debug', '--index', TEST_OUTPUT_PATH])
+
+        part = _get_part("main")
```
elopio: single quotes. I should get a free beer for every one of these ;)

snapcraft/tests/test_parser.py
```diff
@@ -408,3 +408,46 @@ def test_n_documents(
...
+        self.assertEqual('John Doe <john.doe@example.com>', part['maintainer'])
+        self.assertEqual('example main', part['description'])
+
+        part = _get_part("main2")
```
elopio: single quotes.

snapcraft/tests/test_parser.py
```diff
@@ -408,3 +408,46 @@ def test_n_documents(
         main(['--debug', '--index', TEST_OUTPUT_PATH])
 
         self.assertEqual(2, _get_part_list_count())
+
+    @mock.patch('snapcraft.internal.parser._get_origin_data')
+    @mock.patch('snapcraft.internal.sources.get')
+    def test_maintaner_and_description_are_included(
```
elopio: I prefer when we have one thing per test. If you use an "and" in the test name, that's a clear indication that you have two tests. However, it might need too much duplication, or make the tests less clear if we move things to setup or functions. What do you think? Is there a nice way to split it in two?

elopio:
:+1:

snappy-m-o:

snappy-m-o:

snappy-m-o:

snappy-m-o:

snappy-m-o:

snappy-m-o:

snapcraft/tests/test_parser.py
josepht: Fixed.  I'll add a pre-commit hook so I don't go broke buying you beers. ;)

snapcraft/tests/test_parser.py
josepht: fixed.

snapcraft/tests/test_parser.py
josepht: I split the test into two.  There's a lot of duplication but most of it is the mocked output.

sergiusens:
ok to test

snapcraft/internal/parser.py
```diff
@@ -157,6 +160,8 @@ def _process_index(output):
         origin_type = data.get('origin-type')
```
sergiusens: all these gets should probably raise an exception so I guess just doing project_part = data['project_part'] and the same for the others is fine. We capture the exception in whatever is looping over the documents, log a warning and continue with the next document.

PR #566, opened 2016-06-10
Add qmake plugin.

kyrofa:
This PR resolves LP: [#1574774](https://bugs.launchpad.net/snapcraft/+bug/1574774) with a new qmake plugin. It also adds a qt4 and qt5 demo to make use of it.


kyrofa:
Hmm... main.py has apparently gotten complex.

sergiusens:
El 10/06/16 a las 17:15, Kyle Fazzari escribió:
> Hmm... main.py has apparently gotten complex.
With this branch?

snapcraft/plugins/qmake.py
```diff
@@ -0,0 +1,106 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2015 Canonical Ltd
```
elopio: 2016

snapcraft/plugins/qmake.py
```diff
@@ -0,0 +1,106 @@
...
+
+This plugin uses the common plugin keywords as well as those for "sources".
+For more information check the 'plugins' topic for the former and the
+'sources' topic for the latter.
+
+Additionally, this plugin uses the following plugin-specific keywords:
+
+    - options:
+      (list of strings)
+      additional options to pass to the qmake invocation.
```
elopio: qt-version too, right?

snapcraft/plugins/qmake.py
```diff
@@ -0,0 +1,106 @@
...
+
+        # Qt version must be specified
+        schema['required'].append('qt-version')
+
+        # Inform Snapcraft of the properties associated with building. If these
+        # change in the YAML Snapcraft will consider the build step dirty.
+        schema['build-properties'].append('options')
+
+        # Inform Snapcraft of the properties associated with pulling. If these
+        # change in the YAML Snapcraft will consider the pull step dirty.
```
elopio: I think you shouldn't duplicate the comment, just move the second statement right under the first.

snapcraft/plugins/qmake.py
```diff
@@ -0,0 +1,106 @@
...
+
+        self.run(['make', '-j{}'.format(
+            self.project.parallel_build_count)], env=env)
+
+        self.run(['make', 'install', 'INSTALL_ROOT=' + self.installdir],
+                 env=env)
+
+    def _build_environment(self):
+        env = os.environ.copy()
+        env['QT_SELECT'] = self.options.qt_version
```
elopio: cool, I think I'm going to like this.

snapcraft/tests/test_plugin_qmake.py
```diff
@@ -0,0 +1,169 @@
+
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2015 Canonical Ltd
```
elopio: 2016

snapcraft/tests/test_plugin_qmake.py
```diff
@@ -0,0 +1,169 @@
...
+                        'Expected "type" to be included in "options" '
+                        '"items"')
+
+        options_items_type = options_items['type']
+        self.assertEqual(options_items_type, 'string',
+                         'Expected "options" "items" "type" to be '
+                         '"string", but it was "{}"'
+                         .format(options_items_type))
+
+        # Check install-via property
```
elopio: s/install-via/qt-version

snapcraft/tests/test_plugin_qmake.py
```diff
@@ -0,0 +1,169 @@
...
+                       'INSTALL_ROOT={}'.format(plugin.installdir)],
+                      cwd=plugin.builddir, env=mock.ANY)])
+
+    def test_build_environment(self):
+        plugin = qmake.QmakePlugin('test-part', self.options,
+                                   self.project_options)
+        os.makedirs(plugin.builddir)
+        plugin.build()
+
+        expected = {'QT_SELECT': 'qt5'}
```
elopio: no test for qt4? You can make one test, with two subTests:
https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests

elopio:
I don't see a test for the unsupported qt version error.
And the demos could be tested for now just checking that you can get and install the snap. We can leave the execution for later, maybe report a bug saying that we need tests with xvfb.

kyrofa:
>> Hmm... main.py has apparently gotten complex.

> With this branch?

No, I'm getting that on master as well.

snapcraft/plugins/qmake.py
kyrofa: Fixed.

snapcraft/plugins/qmake.py
kyrofa: Ah yes, thank you. Fixed.

snapcraft/plugins/qmake.py
kyrofa: Yeah, a little verbose. Fixed.

snapcraft/tests/test_plugin_qmake.py
kyrofa: Fixed.

snapcraft/tests/test_plugin_qmake.py
kyrofa: Good catch, thank you. Fixed.

snapcraft/tests/test_plugin_qmake.py
kyrofa: Ah ha! I was totally intending to do this and apparently forgot, thank you. Fixed.

kyrofa:
> I don't see a test for the unsupported qt version error.

Good catch, I [added one](https://github.com/ubuntu-core/snapcraft/pull/566/commits/c462ec6da88a7fe005eb66d39bf07524276a86b2#diff-1ad30514a03c863399e52cc992396da5R118).

> And the demos could be tested for now just checking that you can get and install the snap. We can leave the execution for later, maybe report a bug saying that we need tests with xvfb.

Ah, thank you I forgot about actually adding them to demos tests-- done. We'll see if they work! Regarding the xvfb tests, I [created a bug](https://bugs.launchpad.net/snapcraft/+bug/1591539) for you :wink: .

kyrofa:
retest this please.

elopio:
testtools.matchers._impl.MismatchError: !=:
reference = '''\
ant        catkin  copy  jdk     kernel  maven  nodejs   python3  tar-content
autotools  cmake   go    kbuild  make    nil    python2  scons  
'''
actual    = '''\
ant        catkin  copy  jdk     kernel  maven  nodejs   python3  scons      
autotools  cmake   go    kbuild  make    nil    python2  qmake    tar-content
'''

kyrofa:
Should be fixed now. We have a lot of tests that test `list-plugins`!

elopio:
File a bug so we leave only one.

demos/qt4-text-editor/src/main.cpp
```diff
@@ -0,0 +1,59 @@
+/****************************************************************************
+**
+** Copyright (C) 2015 The Qt Company Ltd.
```
sergiusens: This will need an update in debian/copyright

sergiusens:
Please update the branch

kyrofa:
> Please update the branch

Done. If I have to update the `list-plugins` tests again I think I'll collapse.

demos/qt4-text-editor/src/main.cpp
kyrofa: Ooo, good catch.

kyrofa: Alright, take a look.

snapcraft/plugins/qmake.py
sergiusens: yeah, this will play nicely with the upcoming `build-environment` keyword.

elopio:
retest this please

kyrofa:
I just added support for a list of project files (Will Cooke needed it). Waiting to hear back regarding his testing.

snapcraft/internal/yaml.py
```diff
@@ -369,28 +369,32 @@ def _build_env(root, arch_triplet):
...
+        'CPATH', root, arch_triplet, prepend='', sep=':')
+    if include_path_for_env:
+        env.append(include_path_for_env)
     library_path = _get_library_paths('LDFLAGS', root, arch_triplet)
     if library_path:
         env.append(library_path)
+    library_path = _get_library_paths(
+        'LIBRARY_PATH', root, arch_triplet, prepend='', sep=':')
+    if library_path:
+        env.append(library_path)
```
kyrofa: @sergiusens @elopio may I draw your attention here. qmake only seems to respect these flags instead of the gcc ones. I had to put them here because I didn't want to duplicate the `_get_*_paths()` logic in the plugin, and that logic is not public. Is this okay? Or should we make those functions part of the public API?

lpotter:
Is there support for specifying arguments for CONFIG ? I am needing to be able to do this with a couple of snaps I am working on.

8none1:
Confirming that this qmake plugin has successfully built my project.  Thanks!

kyrofa:
Thanks @8none1!

@lpotter yes you should be able to do that via the `options` parameter, but please let me know if you run into problems.

snapcraft/plugins/qmake.py
```diff
@@ -0,0 +1,139 @@
...
+        self.run(['make', '-j{}'.format(
+            self.project.parallel_build_count)], env=env)
+
+        self.run(['make', 'install', 'INSTALL_ROOT=' + self.installdir],
+                 env=env)
+
+    def _extra_config(self):
+        extra_config = []
+
+        for root in [self.installdir, self.project.stage_dir]:
```
kyrofa: @sergiusens @elopio here's the side effect of moving this into the plugin: the plugin doesn't know whether or not it's a root part or a dependent part, so it needs to look in both installdir and stagedir for libs/includes. Do either of you see a downside to doing this?

kyrofa:
Alright, this is ready for another look.

snapcraft/plugins/qmake.py
sergiusens: should be mostly fine as this is what we do in the machinery already. Just need to make sure `installdir` is preferred over `stagedir`.

snapcraft/internal/common.py
```diff
@@ -255,3 +255,39 @@ def format_output_in_columns(elements_list, max_width=MAX_CHARACTERS_WRAP,
...
+
+
+def combine_paths(paths, prepend, separator):
+    paths = ['{}{}'.format(prepend, p) for p in paths]
+    return separator.join(paths)
+
+
+def format_path_variable(envvar, paths, prepend, separator):
+    if not paths:
+        return ''
```
sergiusens: So in the uses of `format_path_variable` I see you always check for `path`, so instead wouldn't it be better here to raise a `ValueError`?

sergiusens:
ok @kyrofa, good stuff, just a comment on a check going on.

snapcraft/plugins/qmake.py
kyrofa: Alright very good, I thought the same.

snapcraft/internal/common.py
kyrofa: Ah, yeah good catch. Honestly though I don't think we need to check it at all here-- if the caller asked to format an environment variable out of the paths, they should probably know that the paths are good, otherwise there's nothing reasonable to return. If you want ValueError raised though, that's fine too.

kyrofa: Meh... that's just asking for cryptic errors. I'll raise.

kyrofa: Done.

kyrofa:
> good stuff, just a comment on a check going on.

Very good-- fixed.

snapcraft/internal/common.py
sergiusens: El 24/06/16 a las 14:49, Kyle Fazzari escribió:
> In snapcraft/internal/common.py
> <https://github.com/ubuntu-core/snapcraft/pull/566#discussion_r68434984>:
> 
>> +        os.path.join(root, 'usr', 'lib'),
>> +        os.path.join(root, 'lib', arch_triplet),
>> +        os.path.join(root, 'usr', 'lib', arch_triplet),
>> +    ]
>> +
>> +    return [p for p in paths if os.path.exists(p)]
>> +
>> +
>> +def combine_paths(paths, prepend, separator):
>> +    paths = ['{}{}'.format(prepend, p) for p in paths]
>> +    return separator.join(paths)
>> +
>> +
>> +def format_path_variable(envvar, paths, prepend, separator):
>> +    if not paths:
>> +        return ''
> 
> Ah, yeah good catch. Honestly though I don't think we need to check it
> at all here-- if the caller asked to format an environment variable out
> of the paths, they should probably know that the paths are good,
> otherwise there's nothing reasonable to return. If you want ValueError
> raised though, that's fine too.

There's nothing wrong with raising a nice exception so the user of the
API can fix fast ;-)

PR #580, opened 2016-06-17
Don't copy libraries that are already in prime.

kyrofa:
Currently snapcraft is fooled by relative rpaths and copies linked libraries, even if they've already been primed. This PR fixes LP: [#1570895](https://bugs.launchpad.net/snapcraft/+bug/1570895) and stops that by checking each library path to ensure it hasn't already been primed before doing so.


integration_tests/test_rpath.py
```diff
@@ -0,0 +1,43 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2015 Canonical Ltd
```
sergiusens: 2016 is the current year. Stop living in the past :tongue:

kyrofa: This was terrible at the very beginning of the year. I kept feeling like it was 2015, so it was fine to see it. Then in, say, February, I started catching files that contained 2015 and putting 2016 in there. Now we're like halfway through the year, and I've apparently started assuming that all the copyrights say 2016 already :stuck_out_tongue: .

kyrofa: Fixed.

sergiusens:
ok to test

kyrofa:
retest this please.

PR #590, opened 2016-06-22
Integrate with new remote parts

sergiusens:


snapcraft/internal/parts.py
```diff
@@ -87,5 +87,44 @@ def _save_headers(self):
...
+        for key in ['description', 'maintainer']:
+            remote_part.pop(key)
+        return remote_part
+
+    def compose(self, part_name, properties):
+        """Return properties composed with the ones from part name in the wiki.
+        :param str part_name: The name of the part to query from the wiki
+        :param dict properties: The current set of properties
+        :return: Part properties from the wiki composed with the properties
+                 passed as a parameter. If there is no wiki part named name,
```
elopio: What about:
If there is no wiki part with the name passed as parameter, just the properties parameter will be returned.

snapcraft/internal/parts.py
```diff
@@ -87,5 +87,44 @@ def _save_headers(self):
...
+                 properties will be returned.
+        :rtype: dict
+        :raises KeyError: if the part named name is not found in the wiki.
```
elopio: I would just write: if the part is not found in the wiki.

snapcraft/internal/yaml.py
```diff
@@ -144,14 +144,12 @@ def _process_parts(self):
             plugin_name = properties.pop('plugin', None)
             if not plugin_name:
                 logger.info(
-                    'Searching the wiki to compose part "{}"'.format(
-                        part_name))
+                    'Searching in the remotes parts cache for part '
```
elopio: s/remotes/remote/

elopio:
I think the mosquitto example is failing here because it is not defined in the new wiki.
Also there is this integration test error: ======================================================================
ERROR: test_pull_wiki_part (test_wiki.WikiTestCase)
test_wiki.WikiTestCase.test_pull_wiki_part
testtools.testresult.real._StringException: output: {{{
Searching in the remotes parts cache for part 'curl'
Issues while validating snapcraft.yaml: the "plugin" keyword is missing for the "curl" part.
}}}

Seems legit failure.

snapcraft/internal/parts.py
sergiusens: I actually need to remove this as we a return a KeyError instead ;-)

snapcraft/internal/parts.py
sergiusens: done

snapcraft/internal/yaml.py
```diff
@@ -144,14 +143,19 @@ def _process_parts(self):
...
-                    # The wiki still supports using 'type' for snapcraft 1.x
-                    if 'type' in properties:
-                        del properties['type']
+                except KeyError:
+                    logger.error(
+                        'A part without a plugin defined will be searched for '
+                        'in the defined remote parts, {!r} is missing the '
+                        '`plugin` entry and is not defined in the current '
+                        'remote parts cache, try to run `snapcraft update` '
+                        'to refresh'.format(part_name))
```
elopio: This is hard to read, three commas in a sentence. I would remove the first sentence and just say: {!r} is not defined in the current remote parts cache, try to run `snapcraft update` to refresh. And maybe add a help entry were we explain more? So we can finish the message with: Type snapcraft help remote-parts for more information.
Just my opinion, I'm not sure my proposal is better.

snapcraft/internal/yaml.py
```diff
@@ -144,14 +143,19 @@ def _process_parts(self):
...
+                    'Searching in the remote parts cache for part '
+                    '{!r}'.format(part_name))
+                try:
+                    properties = self._remote_parts.compose(
+                        part_name, properties)
                     plugin_name = properties.pop('plugin', None)
-                    # The wiki still supports using 'type' for snapcraft 1.x
-                    if 'type' in properties:
-                        del properties['type']
+                except KeyError:
```
elopio: There is no test for this exception.

snapcraft/internal/yaml.py
```diff
@@ -144,14 +143,19 @@ def _process_parts(self):
...
+                    logger.error(
+                        'A part without a plugin defined will be searched for '
+                        'in the defined remote parts, {!r} is missing the '
+                        '`plugin` entry and is not defined in the current '
+                        'remote parts cache, try to run `snapcraft update` '
+                        'to refresh'.format(part_name))
 
             if not plugin_name:
                 raise PluginNotDefinedError(part_name)
```
elopio: this is also missing coverage. Seems closely related to your branch, but feel free to report it as an error to be fixed later.

snapcraft/internal/yaml.py
```diff
@@ -186,13 +190,16 @@ def _compute_part_dependencies(self):
...
                     if found:
-                        plugin_name = wiki_part.pop('plugin')
+                        plugin_name = remote_part.pop('plugin')
                         part.deps.append(self.load_plugin(
-                            dep, plugin_name, wiki_part))
+                            dep, plugin_name, remote_part))
                         self._part_names.append(dep)
+                    else:
+                        logger.info('Maybe {!r} is defined as a remote part, '
+                                    'run `snapcraft update` to refresh')
```
elopio: also missing a test. And should this be an error, shouldn't it?
This logic is so weird. I think this else should raise the exception.
And I don't have it clear if in that case, the following statement is needed.
Splitting smaller functions with nice names out of these loops would be nice, but again, maybe for a future refactor.

sergiusens: El 24/06/16 a las 03:04, Leo Arias escribió:
> In snapcraft/internal/yaml.py
> <https://github.com/ubuntu-core/snapcraft/pull/590#discussion_r68357000>:
>
> >                          self._part_names.append(dep)
> > +                    else:
> > +                        logger.info('Maybe {!r} is defined as a remote part, '
> > +                                    'run `snapcraft update` to refresh')
>
> also missing a test. And should this be an error, shouldn't it?
> This logic is so weird. I think this else should raise the exception.
> And I don't have it clear if in that case, the following statement is
> needed.
> Splitting smaller functions with nice names out of these loops would be
> nice, but again, maybe for a future refactor.

I couldn't agree more, tis is all horrible legacy code that needs some love (all of snapcraft.internal.yaml)

sergiusens: https://bugs.launchpad.net/snapcraft/+bug/1596044

elopio:
:+1:

PR #599, opened 2016-06-23
Allow parts without a 'source' entry.

josepht:
LP:#1595666


elopio:
lgtm. Maybe we can add a part without source to the wiki, so it will be parsed by the integration test.

snappy-m-o:

snappy-m-o:

sergiusens:
El 24/06/16 a las 03:32, Leo Arias escribió:
> lgtm. Maybe we can add a part without source to the wiki, so it will be
> parsed by the integration test.

We have one, well I worked around it, but when this lands I can revert that.

kyrofa:
ok to test

PR #602, opened 2016-06-24
Add a travis suite to run the staging store tests

elopio:
LP: #1596072

### DIFF
--- a/integration_tests/__init__.py
+++ b/integration_tests/__init__.py
@@ -62,7 +62,7 @@ class TestCase(testtools.TestCase):
             cwd = None
         try:
             return subprocess.check_output(
-                [self.snapcraft_command] + command, cwd=cwd,
+                [self.snapcraft_command, '-d'] + command, cwd=cwd,
                 stderr=subprocess.STDOUT, universal_newlines=True)
         except subprocess.CalledProcessError as e:
             self.addDetail('output', content.text_content(e.output))
--- a/integration_tests/__init__.py
+++ b/integration_tests/__init__.py
@@ -61,7 +61,7 @@ class TestCase(testtools.TestCase):
             cwd = None
         try:
             return subprocess.check_output(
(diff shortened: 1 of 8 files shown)


elopio:
skip ci

snappy-m-o:
Test jenkins

elopio:
This is good to go. As good as it gets at least. I launched the travis tests many many times, and in the end found that it will fail if the two executions sync and run a registration within 5 seconds. There's nothing we can do about it right now. We could sign up a different user for each test, but there is no way to accept the store agreement with a script.
The error happened like one every 10 times, so a retrigger would get us to green. And in addition to that, this will happen only when the branch is in snapcore/snapcraft, which we don't do often. And then the execution travis does for new pushes to master is a single one, and unlikely to sync up with another run against staging.

elopio:
retest this please

elopio:
retest this please

integration_tests/__init__.py
```diff
@@ -111,3 +112,9 @@ def logout(self):
         expected = ('Clearing credentials for Ubuntu One SSO.\n'
                     'Credentials cleared.\n')
         self.assertEqual(expected, output)
+
+    def register(self, snap_name):
+        self.run_snapcraft(['register', snap_name])
+        # sleep a few seconds to avoid hitting the store restriction on
+        # following registrations.
+        time.sleep(10)
```
celso-providelo: For the record, Store staging has 5 s rate-limit and production 60 s, a comment on the test would not hurt.

elopio:
@celso-providelo I didn't use the specific rate, and didn't comment much about it because you told me you were not comfortable making that info public :p 
I will have to come back and refactor this when I enable the tests to run in production. I can add more comments then.

celso-providelo:
@elopio slow-clapping for me, huh ?! As we discussed before, when this delay starts impacting the full-suite run we will have to find a better way (no rate-limit for specific users/groups).

PR #603, opened 2016-06-24
Run the travis tests on the ubuntu-core team against the staging store

elopio:
In order to do this, the store tests need to handle an empty password.
As travis only sets the encrypted environment variable for the team branches,
the personal PRs will run against the fake store.

LP: #1596072


elopio:
This PR is here just to show that #602 works also for personal branches, using the fake store. DO NOT MERGE.

PR #604, opened 2016-06-24
Update the remote parts cache before demo tests

sergiusens:
Make sure the cache is updated for the following demos:

- mosquitto
- qt4 text editor
- qt5 text editor

LP: #1596114


sergiusens:
snapstests           PASS

PR #607, opened 2016-06-25
Simplify the list plugins integration test

elopio:
This test has been unnecessarily painful to maintain, because every time
we add a new plugin, the format is changed. We don't need to check the format
anymore, and there is little risk of this breaking because the mechanism has been
extensively tested already.

LP: #1596114


sergiusens:
It was not this bug though. Let's take notice of that when creating the changelog

PR #608, opened 2016-06-25
Implement snapcraft search

sergiusens:
LP: #1596222

### DIFF
--- a/snapcraft/internal/common.py
+++ b/snapcraft/internal/common.py
@@ -213,7 +213,10 @@ def _search_and_replace_contents(file_path, search_pattern, replacement):
 
 
 def get_terminal_width(max_width=MAX_CHARACTERS_WRAP):
-    width = shutil.get_terminal_size().columns
+    if os.isatty(sys.stdout.fileno()):
+        width = shutil.get_terminal_size().columns
+    else:
+        width = max_width
     if max_width:
         width = min(max_width, width)
     return width
--- a/snapcraft/internal/common.py
+++ b/snapcraft/internal/common.py
@@ -218,6 +218,7 @@ def get_terminal_width(max_width=MAX_CHARACTERS_WRAP):
         width = min(max_width, width)
     return width
 
+
 def format_output_in_columns(elements_list, max_width=MAX_CHARACTERS_WRAP,
                              num_col_spaces=2):
     """Return a formatted list of strings ready to be printed line by line
--- a/snapcraft/internal/common.py
+++ b/snapcraft/internal/common.py
@@ -216,7 +216,7 @@ def get_terminal_width(max_width=MAX_CHARACTERS_WRAP):
     if os.isatty(sys.stdout.fileno()):
         width = shutil.get_terminal_size().columns
     else:
-        width = max_width
+        width = MAX_CHARACTERS_WRAP
     if max_width:
         width = min(max_width, width)
     return width
--- a/snapcraft/internal/common.py
+++ b/snapcraft/internal/common.py
@@ -212,6 +212,12 @@ def _search_and_replace_contents(file_path, search_pattern, replacement):
             f.write(replaced)
 
 
+def get_terminal_width(max_width=MAX_CHARACTERS_WRAP):
+    width = shutil.get_terminal_size().columns
+    if max_width:
+        width = min(max_width, width)
+    return width
+
 def format_output_in_columns(elements_list, max_width=MAX_CHARACTERS_WRAP,
                              num_col_spaces=2):
     """Return a formatted list of strings ready to be printed line by line
--- a/snapcraft/internal/parts.py
+++ b/snapcraft/internal/parts.py
@@ -125,18 +125,15 @@ class _RemoteParts(_Base):
         return remote_part
 
     def matches_for(self, part_match, max_len=0):
-        sm = difflib.SequenceMatcher(isjunk=None, autojunk=False)
-        sm.set_seq2(part_match)
+        matcher = difflib.SequenceMatcher(isjunk=None, autojunk=False)
+        matcher.set_seq2(part_match)
 
         matching_parts = {}
         for part_name in self._parts.keys():
-            sm.set_seq1(part_name)
-            add_part_name = sm.ratio() >= _MATCH_RATIO
-            sm.set_seq1(self._parts[part_name]['description'])
-            add_description = sm.ratio() >= _MATCH_RATIO
+            matcher.set_seq1(part_name)
+            add_part_name = matcher.ratio() >= _MATCH_RATIO
 
-            if add_part_name or add_description or part_match in part_name:
-                matching_parts[part_name] = self._parts[part_name]
+            if add_part_name or part_match in part_name:
                 matching_parts[part_name] = self._parts[part_name]
                 if len(part_name) > max_len:
                     max_len = len(part_name)
--- a/snapcraft/internal/parts.py
+++ b/snapcraft/internal/parts.py
@@ -133,7 +133,7 @@ class _RemoteParts(_Base):
             matcher.set_seq1(part_name)
             add_part_name = matcher.ratio() >= _MATCH_RATIO
 
-            if add_part_name or part_match in part_name:
+            if add_part_name or (part_match in part_name):
                 matching_parts[part_name] = self._parts[part_name]
                 if len(part_name) > max_len:
                     max_len = len(part_name)
--- a/snapcraft/internal/parts.py
+++ b/snapcraft/internal/parts.py
@@ -14,6 +14,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import difflib
 import logging
 import os
 import sys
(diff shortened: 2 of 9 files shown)


snapcraft/internal/common.py
```diff
@@ -212,6 +212,16 @@ def _search_and_replace_contents(file_path, search_pattern, replacement):
...
 
 
+def get_terminal_width(max_width=MAX_CHARACTERS_WRAP):
+    if os.isatty(sys.stdout.fileno()):
+        width = shutil.get_terminal_size().columns
+    else:
+        width = max_width
+    if max_width:
+        width = min(max_width, width)
+    return width
```
elopio: This seems wrong. 
You are calling: terminal_width = get_terminal_width(max_width=None). For not a tty, that will return None
Then you do: description_space = terminal_width - part_length - 5
So None - something - 5, will raise:
TypeError: unsupported operand type(s) for -: 'NoneType' and 'int'

snapcraft/internal/parts.py
```diff
@@ -117,6 +124,25 @@ def get_part(self, part_name, full=False):
                 remote_part.pop(key)
         return remote_part
 
+    def matches_for(self, part_match, max_len=0):
+        sm = difflib.SequenceMatcher(isjunk=None, autojunk=False)
+        sm.set_seq2(part_match)
```
elopio: using sm as a variable is a gopher thing. Something like "matcher" would be more pythonic.

snapcraft/internal/parts.py
```diff
@@ -117,6 +124,25 @@ def get_part(self, part_name, full=False):
...
+        matching_parts = {}
+        for part_name in self._parts.keys():
+            sm.set_seq1(part_name)
+            add_part_name = sm.ratio() >= _MATCH_RATIO
+            sm.set_seq1(self._parts[part_name]['description'])
+            add_description = sm.ratio() >= _MATCH_RATIO
+
+            if add_part_name or add_description or part_match in part_name:
+                matching_parts[part_name] = self._parts[part_name]
+                matching_parts[part_name] = self._parts[part_name]
```
elopio: why twice?

snapcraft/internal/parts.py
```diff
@@ -117,6 +124,25 @@ def get_part(self, part_name, full=False):
                 remote_part.pop(key)
         return remote_part
 
+    def matches_for(self, part_match, max_len=0):
```
elopio: _matches_for

snapcraft/internal/parts.py
```diff
@@ -151,5 +177,30 @@ def define(part_name):
...
+
+    terminal_width = get_terminal_width(max_width=None)
+    part_length = max(part_length, header_len)
+    # <space> + <space> + <description> + ... = 5
+    description_space = terminal_width - part_length - 5
+
+    if not matches:
+        # apt search does not return error, we probably shouldn't either.
+        logger.info('No matches found, try to run `snapcraft update` to '
+                    'refresh the remote parts cache.')
```
elopio: or maybe the part just doesn't exist. Should that be part of the message?
What about: to contribute new parts, read <some doc> ?

snapcraft/tests/test_commands_search.py
```diff
@@ -0,0 +1,73 @@
...
+                        new=tests.SilentProgressBar):
+            parts.update()
+
+        patcher = mock.patch('snapcraft.internal.parts.get_terminal_width')
+        self.mock_terminal = patcher.start()
+        self.mock_terminal.return_value = 80
+        self.addCleanup(patcher.stop)
+
+    @mock.patch('sys.stdout', new_callable=io.StringIO)
+    def test_searching_for_a_part_that_exists(self, mock_stdout):
```
elopio: you are missing a test for a match in description, not in name.
And for completenes, maybe a match in name, not in description.

snapcraft/internal/common.py
sergiusens: Ah, nice catch

snapcraft/internal/parts.py
sergiusens: The class is private, the method public, I don't see the reason for this one.

snapcraft/internal/parts.py
sergiusens: this is a vim problem

snapcraft/internal/parts.py
sergiusens: well it really is a search string, not really a part

snapcraft/tests/test_commands_search.py
sergiusens: I removed matching description for now

sergiusens:
@elopio ok, one more round!

snapcraft/internal/parts.py
kyrofa: This method is only ever used within this class though, no? Do you envision it being used by others?

elopio: This is how I see it: an internal class starts with _ and is used only in the module. An internal method starts with _ and is used only in the class. So internal classes can have internal methods. Of course, this is how I use internals in python, not a good practice I have seen everywhere. So we can have a discussion to define our rules for internals.

snapcraft/internal/parts.py
```diff
@@ -117,6 +124,22 @@ def get_part(self, part_name, full=False):
...
+    def matches_for(self, part_match, max_len=0):
+        matcher = difflib.SequenceMatcher(isjunk=None, autojunk=False)
+        matcher.set_seq2(part_match)
+
+        matching_parts = {}
+        for part_name in self._parts.keys():
+            matcher.set_seq1(part_name)
+            add_part_name = matcher.ratio() >= _MATCH_RATIO
+
+            if add_part_name or part_match in part_name:
```
elopio: Use parentheses here to make the rules of precedence clear:
if add_part_name or (part_match in part_name):

snapcraft/internal/parts.py
elopio: ok.

snapcraft/internal/parts.py
```diff
@@ -151,5 +174,30 @@ def define(part_name):
...
+        logger.info('No matches found, try to run `snapcraft update` to '
+                    'refresh the remote parts cache.')
+        return
+
+    print('{}  {}'.format(
+        _HEADER_PART_NAME.ljust(part_length, ' '), _HEADER_DESCRIPTION))
+    for part_key in matches.keys():
+        description = matches[part_key]['description']
+        if len(description) > description_space:
+            description = '{}...'.format(description[0:description_space])
```
kyrofa: I haven't thought through the difficulty of this, but would it be better to wrap instead? Looking something like this:

```
$ snapcraft search qt
PART NAME          DESCRIPTION
qt5conf            This sets up qt5.conf for projects using qml
                   and other qt5 components that need an
                   Ubuntu standard internal path setup by
                   default like the Ubuntu Core Apps.
mqtt-paho/python2  mqtt-paho for python.
mqtt-paho          mqtt-paho for python.
[...]
```

as opposed to:

```
$ snapcraft search qt
PART NAME          DESCRIPTION
qt5conf            This sets up qt5.conf for projects using...
mqtt-paho/python2  mqtt-paho for python.
mqtt-paho          mqtt-paho for python.
[...]
```

snapcraft/tests/test_commands_update.py
```diff
@@ -61,6 +61,12 @@ def test_update(self):
                 'description': 'test entry for part1',
                 'maintainer': 'none',
             },
+            'long-described-part': {
+                'plugin': 'go',
+                'source': 'http://source.tar.gz',
+                'description': 'this is a repetetive description ' * 3,
+                'maintainer': 'none',
+            },
```
elopio: where is this one used?

snapcraft/internal/parts.py
sergiusens: Sure, that's doable

snapcraft/tests/test_commands_update.py
sergiusens: when searching for everything

snapcraft/internal/parts.py
sergiusens: this is bad for filtering @kyrofa

kyrofa: Filtering? You mean e.g. piping to grep? So then the way to see the entire description is just to `snapcraft define` it?

snapcraft/tests/test_commands_update.py
elopio: ah, I see it now. It comes from the fake server.

elopio:
There's a typo: repetetive instead of repetitive. Looks good to me.

snapcraft/internal/parts.py
sergiusens: @elopio I am in agreement. An instance of _RemoteParts is created under the implementation of `search` and calls this method so it is not internal to the class https://github.com/ubuntu-core/snapcraft/pull/608/files#diff-ac0a8a66808d29ef0e343cb310093978R179

sergiusens: @kyrofa no its not https://github.com/ubuntu-core/snapcraft/pull/608/files#diff-ac0a8a66808d29ef0e343cb310093978R179

snapcraft/internal/parts.py
sergiusens: @kyrofa your idea is not bad, but a bug report would be better so if not `grepping` we present a different layout.

kyrofa: Yeah no problem, I don't have strong feelings here. I can make a bug if you'd like one.

kyrofa:
:+1: from me.

snapcraft/internal/parts.py
elopio: hum, that is more complex. It's like a sandwich of internal. I'll say you win. :D

PR #613, opened 2016-06-28
parser - Don't allow duplicate wiki entries.

josepht:
LP:#1597057

The first entry is retained and all duplicates are reported via a
warning and skipped.

### DIFF
--- a/snapcraft/tests/test_parser.py
+++ b/snapcraft/tests/test_parser.py
@@ -14,10 +14,12 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import logging
 import os
 from unittest import mock
 
 import requests
+import fixtures
 import yaml
 
 from snapcraft.internal.parser import (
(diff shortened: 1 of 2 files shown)


elopio:
lgtm

josepht:
retest this please

snapcraft/tests/test_parser.py
```diff
@@ -556,3 +556,35 @@ def test_missing_fields(self, mock_get, mock_get_origin_data):
...
+                    'files': ['file1', 'file2'],
+                },
+            }
+        }
+        main(['--debug', '--index', TEST_OUTPUT_PATH])
+
+        part = _get_part('main')
+        self.assertEqual('example main', part['description'])
+
+        self.assertEqual(1, _get_part_list_count())
```
elopio: You can also test that the warning is in the log. There are many tests using the FakeLogger fixture for that.

josepht: Thanks for the tip.  I've added a FakeLogger and log check.

sergiusens:
Landing as is @josepht but can this please be taken into account during the error code return values rework?

sergiusens:
oh, you updated the branch, now we need to wait :-)

PR #614, opened 2016-06-29
python3: use --root instead of --target

jocave:
This should mean that entry_points defined in the setup.py get created at the expected location as well as the modules installed in usr/lib/python3/dist-packages

### DIFF
--- a/snapcraft/plugins/python3.py
+++ b/snapcraft/plugins/python3.py
@@ -121,8 +121,8 @@ class Python3Plugin(snapcraft.BasePlugin):
         self.run(['python3', easy_install, '--prefix', prefix, 'pip'])
 
         pip3 = os.path.join(self.installdir, 'usr', 'bin', 'pip3')
-        pip_install = ['python3', pip3, 'install', '--target',
-                       site_packages_dir]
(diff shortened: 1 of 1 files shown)


snappy-m-o:

snappy-m-o:

sergiusens:
ok to test

sergiusens:
retest this please!!!!!!

sergiusens:
ok I am good with this code wise, I'll need to run manually unless maybe @didrocks wants to take a look and beat me to it

didrocks:
Conceptually, this is indeed a way better way of installing python based modules. I'm unsure what you meant with "I'll need to run manually", do you mean the examples tests? Is CI broken and can't run them at all anymore?

sergiusens:
El 01/07/16 a las 06:21, Didier Roche escribió:
> Conceptually, this is indeed a way better way of installing python based
> modules. I'm unsure what you meant with "I'll need to run manually", do
> you mean the examples tests? Is CI broken and can't run them at all anymore?
ci works, but I don't think we have any example that uses theses injection points.

sergiusens:
retest this please

PR #617, opened 2016-06-29
snapcraft update if there is no cache

sergiusens:
LP: #1596757

### DIFF
--- a/snapcraft/internal/parts.py
+++ b/snapcraft/internal/parts.py
@@ -111,11 +111,11 @@ class _RemoteParts(_Base):
     def __init__(self):
         super().__init__()
 
-        if os.path.exists(self.parts_yaml):
-            with open(self.parts_yaml) as parts_file:
-                self._parts = yaml.load(parts_file)
-        else:
-            self._parts = {}
+        if not os.path.exists(self.parts_yaml):
+            update()
+
+        with open(self.parts_yaml) as parts_file:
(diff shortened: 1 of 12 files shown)


sergiusens:
Heh, coverage decreased as I removed a bunch of test code that is not needed with this change

kyrofa:
Haha, good defense ;) . This looks good! :+1:

snapcraft/internal/parts.py
```diff
@@ -84,11 +84,11 @@ class _RemoteParts(_Base):
...
-        if os.path.exists(self.parts_yaml):
-            with open(self.parts_yaml) as parts_file:
-                self._parts = yaml.load(parts_file)
-        else:
-            self._parts = {}
+        if not os.path.exists(self.parts_yaml):
+            update()
+
+        with open(self.parts_yaml) as parts_file:
+            self._parts = yaml.load(parts_file)
```
kyrofa: Question here: will this simply result in a "file not found" error if the `part_yaml` still isn't here after the update?

sergiusens: How many ifs do you want :-P The previous statement makes it available (update). If that fails an exception will be raised already ;-)

kyrofa: Haha, I'm just trying to consider what happens in the case of a typo. File not found, or is that caught elsewhere?

PR #618, opened 2016-06-29
parser - Support .snapcraft.yaml files.

josepht:
When getting origin data use either snapcraft.yaml or .snapcraft.yaml.
Error if both or neither are present.

LP:#1597484
LP:#1597778


josepht:
retest this please

integration_tests/test_parser.py
```diff
@@ -33,3 +33,26 @@ def test_parser_basic(self):
                               stdout=subprocess.DEVNULL)
 
         self.assertTrue(os.path.exists('parts.yaml'))
+
+    def test_hidden_snapcraft_yaml(self):
+        """Test hidden .snapcraft.yaml file."""
+
+        args = [self.snapcraft_parser_command, '--index',
+                'http://people.canonical.com/~josepht/hidden_parts_wiki',
```
elopio: is there a way to make this a local file?

snapcraft/internal/parser.py
```diff
@@ -96,7 +101,21 @@ def _update_after_parts(partname, after_parts):
 
 def _get_origin_data(origin_dir):
     origin_data = {}
-    with open(os.path.join(origin_dir, 'snapcraft.yaml'), "r") as fp:
+    snapcraft_yaml_file = os.path.join(origin_dir, 'snapcraft.yaml')
+    hidden_snapcraft_yaml_file = os.path.join(origin_dir, '.snapcraft.yaml')
+    yaml_file = ''
```
elopio: I think it is not needed to initialize the yaml_file like this.

snapcraft/tests/test_parser.py
```diff
@@ -556,3 +561,41 @@ def test_missing_fields(self, mock_get, mock_get_origin_data):
...
+        origin = 'git@github.com:testuser/testproject.git'
+        origin_dir = _encode_origin(origin)
+
+        self.assertEqual('gitgithub.comtestusertestproject.git', origin_dir)
+
+    def test__encode_origin_lp(self):
+        origin = 'lp:~testuser/testproject/testbranch'
+        origin_dir = _encode_origin(origin)
+
+        self.assertEqual('lptestusertestprojecttestbranch', origin_dir)
```
elopio: This is not too bad, but ideally, we test only the public methods. In this case, all the tests should call run, not the internals. Did you try that approach? I understand it might require too much set up, or too many fakes. I'm not asking for it in this branch, but if the public method is not testable, that's a bug that should be reported.

josepht: I didn't try that approach because I foresaw it taking a lot of faking to get around reading the index and fetching the origins.

elopio: That's the thing. We should be able to easily inject a fake server or client with a canned or seeded response that will reply quickly. That will take a lot of time once, but then the following tests will be easier. We could also move some of these internal methods into a module, and make them the public api of that module, which parser consumes. That could be cheating if we are only doing that to make them look less private, but a combination of both could be nice, and we can leave a few whitebox tests for hard details. Again, I'm not blocking this branch because of this, just starting the discussion to find a way to make future tests a little less whitebox.

snappy-m-o:

snappy-m-o:

snapcraft/tests/test_parser.py
josepht: fwiw, PR#636 adds some fake servers for the basic happy-path integration test.

josepht:
retest this please

josepht:
retest this please

integration_tests/test_parser.py
```diff
@@ -36,3 +36,42 @@ def test_parser_basic(self):
                               stdout=subprocess.DEVNULL)
 
         self.assertTrue(os.path.exists('parts.yaml'))
+
+    def test_hidden_snapcraft_yaml(self):
+        """Test hidden .snapcraft.yaml file."""
+
+        print("JOE: pwd: %s" % os.getcwd())
```
sergiusens: this needs to be removed I guess

snapcraft/internal/parser.py
```diff
@@ -115,8 +124,25 @@ def _update_after_parts(partname, after_parts):
...
+            raise BadSnapcraftYAMLError(
+                'Origin has both "snapcraft.yaml" and ".snapcraft.yaml"')
+        else:
+            yaml_file = snapcraft_yaml_file
+    elif os.path.exists(hidden_snapcraft_yaml_file):
+        yaml_file = hidden_snapcraft_yaml_file
+
     try:
-        with open(os.path.join(origin_dir, 'snapcraft.yaml'), "r") as fp:
+        with open(yaml_file, "r") as fp:
```
sergiusens: single quotes ftw :-P

josepht: fixed

integration_tests/test_parser.py
josepht: removed

snapcraft/internal/parser.py
```diff
@@ -115,8 +124,25 @@ def _update_after_parts(partname, after_parts):
...
+            raise BadSnapcraftYAMLError(
+                'Origin has both "snapcraft.yaml" and ".snapcraft.yaml"')
+        else:
+            yaml_file = snapcraft_yaml_file
+    elif os.path.exists(hidden_snapcraft_yaml_file):
+        yaml_file = hidden_snapcraft_yaml_file
+
     try:
-        with open(os.path.join(origin_dir, 'snapcraft.yaml'), "r") as fp:
+        with open(yaml_file, 'r') as fp:
```
sergiusens: Actually one last nit here, 'r' is implicit so it can be just `with open(yaml_file) as fp:`

josepht: Removed 'r'.

PR #619, opened 2016-06-30
Add source-checksum option

tsimonq2:
This adds an optional `source-checksum` tag to ensure download integrity by checking the checksum.

## Features
 - It supports the following `source-type`s:
  - `source-type: tar`
  - `source-type: zip`
 - It supports the following checksums:
  - `md5`
  - `sha1`
  - `sha224`
  - `sha256`
  - `sha384`
  - `sha512`
 - You can use it in the following ways:
  - `source-checksum: HASH` - raw hash
  - `source-checksum: checksum.txt` - local file
  - `source-checksum: http(s)://example.com/checksum` - remote URL

## Failsafes
It has the following failsafes to ensure the code always works as intended:
 - Special exception for a checksum that doesn't match
 - Raises an exception for a checksum that is invalid, or incompatible with the checksums supported

## Testing coverage
I have the following testing in place:
 - Unit tests:
  - Testing of each checksum in raw form with a created file (tar/zip) that has a static checksum
  - Creation of a checksum file and checking the checksum with that local file
  - Pulling http://tsimonq2.net/misc/snapcraft-checksum-tar and http://tsimonq2.net/misc/snapcraft-checksum-zip respectively and checking the checksum of the static file with those
  - Checking for an invalid checksum
 - Integration tests:
  - Testing of each checksum in raw form with a created file (tar/zip) that has a static checksum in the `snapcraft.yaml` files for `simple-tar` and `simple-zip`

## Discrepancies between this specification and the source
If you find any difference between the code in this PR and the specification above, please comment, as that wasn't intended.


snappy-m-o:

snappy-m-o:

snapcraft/internal/sources.py
```diff
@@ -69,11 +75,12 @@
...
 import tempfile
 import zipfile
+import hashlib
+import urllib
 
 from snapcraft.internal import common
 
-
-logging.getLogger('urllib3').setLevel(logging.CRITICAL)
+logging.getLogger('urllib').setLevel(logging.CRITICAL)
```
sergiusens: why this change?

snapcraft/internal/sources.py
```diff
@@ -235,20 +265,54 @@ def pull(self):
...
+            if sha512 != source_checksum:
+                raise NonMatchingChecksum(
+                    "the checksum doesn't match the downloaded file")
+
+    def provision(self, dst, source_checksum, clean_target=True,
+                  keep_tarball=False):
         # TODO add unit tests.
         tarball = os.path.join(self.source_dir, os.path.basename(self.source))
 
+        self.pre_check_checksum(self, source_checksum, tarball)
```
sergiusens: this should be done after the download

snapcraft/internal/sources.py
```diff
@@ -307,10 +371,46 @@ def __init__(self, source, source_dir, source_tag=None,
         elif source_branch:
             raise IncompatibleOptionsError(
                 'can\'t specify a source-branch for a zip source')
-
-    def provision(self, dst, clean_target=True, keep_zip=False):
+        elif source_checksum:
+            raise IncompatibleOptionsError(
+                'can\'t specify source-checksum for a zip source right now')
+
+    def pre_check_checksum(self, source_checksum, zip):
```
sergiusens: this looks like duplicate code, please consider adding it to FileBase (which also needs its __init__ updated)

snapcraft/internal/sources.py
sergiusens: which is https://github.com/snapcore/snapcraft/pull/619/files#diff-e6205a80af76b4faec1d9241dc3b3b7cR119

sergiusens:
Please merge with master or rebase

sergiusens:
you also need to update https://github.com/snapcore/snapcraft/pull/619/files#diff-e6205a80af76b4faec1d9241dc3b3b7cR455

tsimonq2:
@sergiusens Thanks for taking the time to look at this. I updated my code with your suggestions. Let me know what you think.

@elopio Can you please take a look?

sergiusens:
From a quick look at the test failures it seems you still need to update all constructors for the sources used in the unit tests.

tsimonq2:
So right now, the Travis build has passed. My goal now is to do some more local testing and to get an integration test ready. I'll change the description accordingly.

snappy-m-o:

snappy-m-o:

snappy-m-o:

snappy-m-o:

snappy-m-o:

snappy-m-o:

tsimonq2:
I believe that all the intended testing and code is in place for this PR. Could this be evaluated please?

tsimonq2:
I also added sort of a specification to refer to in my initial comment.

tsimonq2:
Bump

snapcraft/tests/test_sources.py
```diff
@@ -68,6 +71,89 @@ def test_pull_tarball_must_download_to_sourcedir(self, mock_prov):
...
+                                    '2ada4d494bbc3018d7545b3c631dac1bfb787f81'
+                                    'e0b76530 checksum.tar')
+        source_checksum_write.close()
+        source_checksum = 'CHECKSUM'
+
+        sources.check_checksum_determine_format(
+            source_checksum, "checksum.tar")
+
+        # From remote file
+        source_checksum = 'http://tsimonq2.net/misc/snapcraft-checksum-tar'
```
sergiusens: this should use a fakeserver, check `snapcraft/tests/fake_servers.py`

snapcraft/internal/sources.py
```diff
@@ -435,3 +476,54 @@ def _get_source_type_from_uri(source, ignore_errors=False):
...
+            if source_checksum.index(" "):
+                source_checksum = source_checksum.split(" ", 1)[0]
+        finally:
+            filework.close()
+    check_checksum_determine_type_and_read_file(source_checksum, checkfile)
+
+
+def check_checksum_determine_type_and_read_file(source_checksum, checkfile):
+    if len(source_checksum) == 32:
+        chksum = hashlib.md5()
```
sergiusens: why the abreviation?

sergiusens:
Hi, looks good

One general comment is to use `checksum` instead of `chksum`

snapcraft/tests/test_sources.py
```diff
@@ -68,6 +71,89 @@ def test_pull_tarball_must_download_to_sourcedir(self, mock_prov):
         with open(os.path.join(dest_dir, tar_file_name), 'r') as tar_file:
             self.assertEqual('Test fake compressed file', tar_file.read())
 
+    def test_checksum(self):
```
sergiusens: what are you testing here, or in other words, what are your assertions to know it all worked as expected?

snapcraft/tests/test_sources.py
```diff
@@ -114,6 +200,89 @@ def test_extract_and_keep_zipfile(self, mock_zip):
         with open(zip_download, 'r') as zip_file:
             self.assertEqual('Test fake compressed file', zip_file.read())
 
+    def test_checksum_of_zip(self):
```
sergiusens: ditto on assertions

snapcraft/tests/test_sources.py
sergiusens: maybe I am confused by `check_checksum_determine_format` and a better method name is needed here.

snapcraft/internal/sources.py
```diff
@@ -435,3 +476,54 @@ def _get_source_type_from_uri(source, ignore_errors=False):
...
+            filework = open(filename, 'r')
+            source_checksum = filework.read()
+            if source_checksum.index(" "):
+                source_checksum = source_checksum.split(" ", 1)[0]
+        finally:
+            filework.close()
+    check_checksum_determine_type_and_read_file(source_checksum, checkfile)
+
+
+def check_checksum_determine_type_and_read_file(source_checksum, checkfile):
```
sergiusens: if this is only used by `check_checksum_determine_format` then prefix the method name with `_`

snapcraft/internal/sources.py
```diff
@@ -435,3 +476,54 @@ def _get_source_type_from_uri(source, ignore_errors=False):
...
+    elif len(source_checksum) == 56:
+        chksum = hashlib.sha224()
+    elif len(source_checksum) == 64:
+        chksum = hashlib.sha256()
+    elif len(source_checksum) == 96:
+        chksum = hashlib.sha384()
+    elif len(source_checksum) == 128:
+        chksum = hashlib.sha512()
+    else:
+        raise IncompatibleOptionsError("Invalid checksum format")
```
sergiusens: This could be a dictionary instead

```python
_HASH_FUNCTIONS = {
    32: hashlib.md5,
    40: hashlib.sha1,
    ...
    ...
}

try:
     checksum_function = _HASH_FUNCTIONS[len(source_checksum)]()
except KeyError:
    raise IncompatibleOptionsError("Invalid checksum format")

**Comment by sergiusens** [comment]
Another general comment,

you should name the method the sources classes `verify_checksum`; `verify_checksum` should call `_get_checksum` and return the checksum doing the http, file or checksum dance you do.

With that returned value, if you use the dict, just do the check there and finally check if they are a match or raise the exception.

This IMHO would read much simpler.

File: snapcraft/tests/test_sources.py (hunk 4 above)
Comment (tsimonq2) [inline snapcraft/tests/test_sources.py]: @sergiusens I'm a little confused as to the usage of this, where can I find an example of a test case that uses it?

File: snapcraft/internal/sources.py (hunk 5 above)
Comment (tsimonq2) [inline snapcraft/internal/sources.py]: I didn't have any specific reason, I'll change it

File: snapcraft/tests/test_sources.py (hunk 6 above)
Comment (tsimonq2) [inline snapcraft/tests/test_sources.py]: @sergiusens this function uses the various supported checksum formats and checks the checksum with them. This ensures that the checksum can be successfully checked with the raw checksums.

The only reason why I had to split it into separate functions was because the static tests were complaining about check_checksum being too complex.

**Comment by tsimonq2** [comment]
@sergiusens I made some changes that addresses your comments, let me know if you have more.

File: snapcraft/internal/sources.py [hunk 10]
```diff
@@ -435,3 +476,52 @@ def _get_source_type_from_uri(source, ignore_errors=False):
...
     return source_type
+
+
+def verify_checksum(source_checksum, checkfile):
+    if source_checksum.startswith('http'):
+        response = urllib.request.urlopen(source_checksum)
+        data = response.read()
+        source_checksum = data.decode('utf-8')
+        if (" " in source_checksum):
+            source_checksum = source_checksum.split(" ", 1)[0]
```
sergiusens: use single quotes here

snapcraft/internal/sources.py
```diff
@@ -435,3 +476,52 @@ def _get_source_type_from_uri(source, ignore_errors=False):
...
+        else:
+            print('No file name detected in the checksum file, perhaps an '
```
sergiusens: this should be an error, shouldn't it?

snapcraft/internal/sources.py
```diff
@@ -310,6 +346,10 @@ def __init__(self, source, source_dir, source_tag=None,
     def provision(self, dst, clean_target=True, keep_zip=False):
         zip = os.path.join(self.source_dir, os.path.basename(self.source))
 
+        if self.source_checksum:
+            verify_checksum(
+                self.source_checksum, zip)
```
sergiusens: these two lines certainly fit in one line, right?

snapcraft/internal/sources.py
```diff
@@ -310,6 +346,10 @@ def __init__(self, source, source_dir, source_tag=None,
     def provision(self, dst, clean_target=True, keep_zip=False):
         zip = os.path.join(self.source_dir, os.path.basename(self.source))
```
sergiusens: please use a different name than `zip`, `zip` is a python keyword and it might get confusing.

snapcraft/internal/sources.py
```diff
@@ -435,3 +476,52 @@ def _get_source_type_from_uri(source, ignore_errors=False):
...
+        else:
+            print('No file name detected in the checksum file, perhaps an '
+                  'invalid checksum file?')
+    if os.path.isfile(source_checksum):
+        filename = source_checksum
+        try:
+            filework = open(filename, 'r')
+            source_checksum = filework.read()
+            if (" " in source_checksum):
+                source_checksum = source_checksum.split(" ", 1)[0]
```
sergiusens: quotes

snapcraft/internal/sources.py
```diff
@@ -435,3 +476,52 @@ def _get_source_type_from_uri(source, ignore_errors=False):
...
+            else:
+                print('No file name detected in the checksum file, perhaps an '
```
sergiusens: should probably be an error.

snapcraft/internal/sources.py
```diff
@@ -435,3 +476,52 @@ def _get_source_type_from_uri(source, ignore_errors=False):
...
+
+    with open(checkfile, "rb") as f:
+        for chunk in iter(lambda: f.read(4096), b""):
+            checksum.update(chunk)
+
+    checksum = checksum.hexdigest()
+
+    if checksum != source_checksum:
+        raise ChecksumDoesNotMatch(
+            "the checksum ( "+source_checksum+" ) doesn't match the file"
```
sergiusens: can we stick to using `format` and single quotes? (no `+`)

snapcraft/internal/sources.py
```diff
@@ -435,3 +476,52 @@ def _get_source_type_from_uri(source, ignore_errors=False):
...
+        56: hashlib.sha224(),
+        64: hashlib.sha256(),
+        96: hashlib.sha384(),
+        128: hashlib.sha512()
+    }
+
+    try:
+        checksum = _HASH_FUNCTIONS[len(source_checksum)]
+    except KeyError:
+        raise IncompatibleOptionsError("Invalid checksum format")
```
sergiusens: single quotes please

snapcraft/tests/test_sources.py
```diff
@@ -68,6 +101,102 @@ def test_pull_tarball_must_download_to_sourcedir(self, mock_prov):
...
             self.assertEqual('Test fake compressed file', tar_file.read())
 
+    def test_checksum(self):
+        tar = tarfile.open("checksum.tar", "w")
+        tar.close()
+
+        # md5
+        source_checksum = '1276481102f218c981e0324180bafd9f'
+        sources.verify_checksum(
+            source_checksum, "checksum.tar")
```
sergiusens: these statements fit in the same line

snapcraft/internal/sources.py
tsimonq2: @sergiusens do you think it's **always** the case that the checksum file has a filename? If so, I'll correct it.

snapcraft/internal/sources.py
tsimonq2: I'm not sure what you mean here, @sergiusens

snapcraft/internal/sources.py
tsimonq2: @sergiusens This is applicable to the original source as well and should be dealt with in a separate PR, I'm just using what has already been given to me for variable names.

snapcraft/tests/test_sources.py
tsimonq2: Pushing a fix to a bunch of lines that can be fixed by this, thanks for pointing this out.

snapcraft/internal/sources.py
sergiusens: Should be something like 
```python
raise ChecksumDoesNotMatch(
   'the checksum {!r} doesn't match the file'.format(source_checksum)
```

sergiusens:
OK, one final checkup, in the bug sabdfl mentions we should support sha-3-384; how do we differentiate sha-2 and sha-3 with the autodetection?

snapcraft/internal/sources.py
tsimonq2: I can't use single quotes because `doesn't` has a quote.

kyrofa:
@tsimonq2 did you see @sergiusens's question?

sergiusens:
Closing due to inactivity

PR #620, opened 2016-06-30
Alter prefix for pkg-config to get correct values

sergiusens:
LP: #1595243


snapcraft/internal/repo.py
```diff
@@ -284,6 +285,17 @@ def _setup_apt_cache(rootdir, sources, project_options):
     return apt_cache, progress
 
 
+def fix_pkg_config(root, pkg_config_file):
+    """Opens a pkg_config_file and prefixes the prefix with root."""
+    pattern = re.compile('^prefix=(?P<prefix>.*)')
+    for line in fileinput.input(pkg_config_file, inplace=True):
```
kyrofa: Ah, I've not seen this before.

snapcraft/internal/pluginhandler.py
```diff
@@ -644,7 +651,7 @@ def _migratable_filesets(fileset, srcdir):
 
 
 def _migrate_files(snap_files, snap_dirs, srcdir, dstdir, missing_ok=False,
-                   follow_symlinks=False):
+                   follow_symlinks=False, fixup_func=None):
```
kyrofa: This doesn't make for a particularly intuitive function from its signature, but I guess there's little reason to walk `dstdir` again.

kyrofa: Nah, I take that back--  could see this being useful. I wonder if it's worth supporting multiple fixup_funcs.

kyrofa:
Yeah this is looking good. How do you plan on supporting `installdir` for the root? Also, could use some tests :wink: .

sergiusens:
El 01/07/16 a las 09:23, Kyle Fazzari escribió:
> Yeah this is looking good. How do you plan on supporting |installdir|
> for the root? Also, could use some tests 😉 .

Test will come when I figure out the installdir story :-)

snapcraft/internal/pluginhandler.py
didrocks: Don't you want to make it a lambda function taking one parameter? That way, it's clear from the signature that the function is of form:
```
def fun(dst):
  pass
```
As you don't have docstrings to specify it.

snapcraft/internal/pluginhandler.py
```diff
@@ -664,6 +671,8 @@ def _migrate_files(snap_files, snap_dirs, srcdir, dstdir, missing_ok=False,
             os.remove(dst)
 
         common.link_or_copy(src, dst, follow_symlinks=follow_symlinks)
+        if fixup_func:
```
didrocks: and so, with a lambda, you can just call the function and not caring :)

snapcraft/internal/repo.py
```diff
@@ -284,7 +285,26 @@ def _setup_apt_cache(rootdir, sources, project_options):
...
 
-def _fix_symlinks(debdir):
+def fix_pkg_config(root, pkg_config_file, prefix_trim=None):
+    """Opens a pkg_config_file and prefixes the prefix with root."""
+    pattern_trim = None
+    if prefix_trim:
+        pattern_trim = re.compile(
+            '^prefix={}(?P<prefix>.*)'.format(prefix_trim))
+    pattern = re.compile('^prefix=(?P<prefix>.*)')
+    for line in fileinput.input(pkg_config_file, inplace=True):
```
didrocks: I personnally tend to use fileinput as a contextmanager (easier to read, cleaner on opening/close files) as it's a more pythonish way of doing files handling in python3: https://docs.python.org/3/library/fileinput.html

snapcraft/internal/repo.py
```diff
@@ -299,18 +319,13 @@ def _fix_symlinks(debdir):
...
-                    if not _try_copy_local(path, target):
-                        continue
-                os.remove(path)
-                os.symlink(os.path.relpath(target, root), path)
+                _fix_symlink(path, debdir, root)
             elif os.path.exists(path):
                 _fix_filemode(path)
 
+            if path.endswith('.pc') and not os.path.islink(path):
+                fix_pkg_config(debdir, path)
```
didrocks: Shouldn't you call fixup_func() rather here? (need to match the number of parameters) and move it to common as you are duplicating the checks for pc files + symlinks?

snapcraft/internal/repo.py
```diff
@@ -326,6 +341,18 @@ def _fix_xml_tools(root):
                 format(root), xslt_config_path])
 
 
+def _fix_symlink(path, debdir, root):
+    target = os.path.join(debdir, os.readlink(path)[1:])
+    if _skip_link(os.readlink(path)):
+        logger.debug('Skipping {}'.format(target))
+        return
+    if not os.path.exists(target):
+        if not _try_copy_local(path, target):
```
didrocks: … `and` isn't of 2 nested if? ;)

snapcraft/internal/repo.py
sergiusens: This is the only comment I have not addressed as I really want to get rid of common

sergiusens:
the java example failed to build in examples tests. Unrelated IMO and it passed in adt.

kyrofa:
Yeah, this looks very good, :+1:

sergiusens:
tested on @mhall119 's pantheon and works fine

PR #621, opened 2016-07-01
Correctly check !# for commands in PATH

sergiusens:
Commands that were in PATH weren't hashbang checked. This
makes it always the case and makes the `command` always
relative to the $SNAP.

LP: #1597919

### DIFF
--- a/snapcraft/internal/meta.py
+++ b/snapcraft/internal/meta.py
@@ -217,15 +217,14 @@ class _SnapPackaging:
 
         wrapexec = '$SNAP/{}'.format(execparts[0])
         if not os.path.exists(exepath) and '/' not in execparts[0]:
-            _find_bin(execparts[0], self._snap_dir)
-            wrapexec = execparts[0]
-        else:
-            with open(exepath, 'rb') as exefile:
-                # If the file has a she-bang, the path might be pointing to
-                # the local 'parts' dir. Extract it so that _write_wrap_exe
-                # will have a chance to rewrite it.
-                if exefile.read(2) == b'#!':
-                    shebang = exefile.readline().strip().decode('utf-8')
+            exepath = _find_bin(execparts[0], self._snap_dir)
+
+        with open(exepath, 'rb') as exefile:
+            # If the file has a she-bang, the path might be pointing to
+            # the local 'parts' dir. Extract it so that _write_wrap_exe
+            # will have a chance to rewrite it.
+            if exefile.read(2) == b'#!':
+                shebang = exefile.readline().strip().decode('utf-8')
 
         self._write_wrap_exe(wrapexec, wrappath,
                              shebang=shebang, args=execparts[1:])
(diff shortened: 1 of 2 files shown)


sergiusens:
I think I want to rewrite the tests to actually run instead of mock it now.

evandandrea:
This doesn't seem to resolve the issue.

Both:
- git://git.launchpad.net/~thomir/+git/snapcraft-broken
- https://pastebin.canonical.com/160361

Create shebangs of the form `#!/home/ev/branches/snapcraft-broken/parts/testme/install/usr/bin/python3` when built using [hashbang-the-right-way](https://github.com/sergiusens/snapcraft/tree/bugfix/1597919/hashbang-the-right-way)

sergiusens:
El miércoles, 6 de julio de 2016 18h'50:38 ART, Evan Dandrea 
> This doesn't seem to resolve the issue.
> 
> Both:
> - git://git.launchpad.net/~thomir/+git/snapcraft-broken
> - https://pastebin.canonical.com/160361
> 
> Create shebangs of the form 
> `#!/home/ev/branches/snapcraft-broken/parts/testme/install/usr/bin/python3` 
> when built using 
> [hashbang-the-right-way](https://github.com/sergiusens/snapcraft/tree/bugfix/1597919/hashbang-the-right-way)

Right, what we do is ignore the shebang completely. To correct the shebang 
the right way we would need a fix in python's stdlibs.

> 
> ---
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/snapcore/snapcraft/pull/621#issuecomment-230918139


-- 

sergiusens:
@evandandrea here are some of the details http://stackoverflow.com/questions/1530702/dont-touch-my-shebang

we could of course add some support to the plugin to rollback to using `#!/usr/bin/env python[3]` but @bjornt  came up with a general solution of making the `command` prepend python3 or python depending on the shebang

evandandrea:
> @bjornt came up with a general solution of making the `command` prepend python3 or python depending on the shebang

Doesn't that assume python binaries in a snap won't call each other?

sergiusens:
El 07/07/16 a las 15:52, Evan Dandrea escribió:
>     @bjornt <https://github.com/bjornt> came up with a general solution
>     of making the |command| prepend python3 or python depending on the
>     shebang
> 
> Doesn't that assume python binaries in a snap won't call each other?

Yes it does; is this what is happening now?

evandandrea:
> Yes it does; is this what is happening now?

@sergiusens not in either of the two examples posted, but I don't think it unreasonable to expect that someone will do `subprocess.call(['other_script'])`

evandandrea:
> @sergiusens not in either of the two examples posted, but I don't think it unreasonable to expect that someone will do subprocess.call(['other_script'])

So until upstream python gets fixed, why not sed,#!.*python3$,#!/usr/bin/env python3,? We can put a bright yellow warning in the build output and offer a command line flag for --no-mangle.

kyrofa:
I'm going to close this in favor of #680, which rewrites shebangs instead of handling this at the command level. @evandandrea if you don't mind taking that for a spin, I'd appreciate it!

PR #623, opened 2016-07-01
Proper message when registering an already registered snap

sergiusens:
LP: #1596782


elopio:
lgtm. Just a small detail, the error I get when I try to register a name twice is:

The name 'u1test20160702' is already taken.

We can if needed rename snaps to ensure they match the expectations of most users. If you are the publisher most users expect for 'u1test20160702' then claim the name at 'https://myapps.developer.ubuntu.com/dev/click-apps/register-name/?series=16&name=u1test20160702'

That is not the URL to claim the name. That is to register another name, with a link to fill a form for claiming. So the message we are giving is a little inaccurate. This message would be accurate if the link is to the claim form: https://myapps.developer.ubuntu.com/dev/click-apps/register-name-dispute/?series=16&name=u1test20160702

I think the store should return this dispute url instead.

sergiusens:
@elopio I went with constructing one as the error message is pretty lame if not.

elopio:
great, with a link to the bug. :+1:

elopio:
Ran the tests with fake and staging servers. All good.
But I am not authorized to merge this, @sergiusens you go.

PR #624, opened 2016-07-02
Added maven-targets to maven plugin.

ZenHarbinger:
maven-targets is an object that allows a source-destination pair.
The default value is `target: jar`

This modification is useful if using a composite pom.xml that has many sub projects.  The parent pom.xml won't generate a 'target' directory.  This allows using multiple outputs of the parent pom in the snap.

Ex:
```yml
parts:
  algolink-mvn:
    plugin: maven
    maven-targets:
      algolink-gui/target: algolink-gui
      algolink-cli/target: algolink-cli
    source: .
```

### DIFF
--- a/snapcraft/plugins/maven.py
+++ b/snapcraft/plugins/maven.py
@@ -78,9 +78,14 @@ class MavenPlugin(snapcraft.plugins.jdk.JdkPlugin):
             'default': [],
         }
 
+        schema['properties']['maven-targets'] = {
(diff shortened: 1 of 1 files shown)


snappy-m-o:

snappy-m-o:

ZenHarbinger:
realized that copying the target directory carries baggage.
will figure out and re-submit.

PR #626, opened 2016-07-03
Added new maven-targets option to maven plugin.

ZenHarbinger:
Composite maven projects do not build to 'target'. This update allows the user to specify the child target to copy.

```yml
parts:
  algolink-mvn:
    plugin: maven
    source: .
    maven-targets:
      - algolink-gui
      - algolink-cli
```

By default, the basedir/target directory is searched, otherwise it uses the 'target' directory of each specified value.
The installdir for each child project is the name of the child project. If no values are specified, it reverts to the default of 'jar' or 'war' depending on the target type of the project.

### DIFF
--- a/snapcraft/plugins/maven.py
+++ b/snapcraft/plugins/maven.py
@@ -78,9 +78,20 @@ class MavenPlugin(snapcraft.plugins.jdk.JdkPlugin):
             'default': [],
         }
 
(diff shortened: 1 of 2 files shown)


snappy-m-o:

snappy-m-o:

PR #627, opened 2016-07-03
Added new maven-targets option to maven plugin.

ZenHarbinger:
Composite maven projects do not build to 'target'. This update allows the user to specify the child target to copy.

```yml
parts:
  algolink-mvn:
    plugin: maven
    source: .
    maven-targets:
      - algolink-gui
      - algolink-cli
```
By default, the basedir/target directory is searched, otherwise it uses the 'target' directory of each specified value.
The installdir for each child project is the name of the child project. If no values are specified, it reverts to the default of 'jar' or 'war' depending on the target type of the project.

### DIFF
--- a/snapcraft/plugins/maven.py
+++ b/snapcraft/plugins/maven.py
@@ -119,22 +119,19 @@ class MavenPlugin(snapcraft.plugins.jdk.JdkPlugin):
 
             jarfiles = glob.glob(os.path.join(src, '*.jar'))
             warfiles = glob.glob(os.path.join(src, '*.war'))
+            arfiles = glob.glob(os.path.join(src, '*.(j|w)ar'))
 
-            if len(f) > 0 and jarfiles:
+            if not (arfiles):
+                raise RuntimeError('could not find any built jar files for part')
+            elif len(f) > 0:
                 basedir = f
-                files = jarfiles
-            elif len(f) > 0 and warfiles:
-                basedir = f
-                files = warfiles
+                files = arfiles
             elif jarfiles:
                 basedir = 'jar'
                 files = jarfiles
-            elif warfiles:
+            else:
                 basedir = 'war'
                 files = jarfiles
-            else:
-                raise RuntimeError("could not find any"
-                                   "built jar files for part")
 
             targetdir = os.path.join(self.installdir, basedir)
             os.makedirs(targetdir, exist_ok=True)
--- a/snapcraft/plugins/maven.py
+++ b/snapcraft/plugins/maven.py
@@ -118,17 +118,16 @@ class MavenPlugin(snapcraft.plugins.jdk.JdkPlugin):
             src = os.path.join(self.builddir, f, 'target')
 
             jarfiles = glob.glob(os.path.join(src, '*.jar'))
-            warfiles = glob.glob(os.path.join(src, '*.war'))
             arfiles = glob.glob(os.path.join(src, '*.[jw]ar'))
 
             if not (arfiles):
-                raise RuntimeError('could not find any built jar files for part')
+                raise RuntimeError("could not find any"
+                                   "built jar files for part")
+            basedir = 'war'
             if len(f) > 0:
                 basedir = f
             elif jarfiles:
                 basedir = 'jar'
-            else:
-                basedir = 'war'
 
             targetdir = os.path.join(self.installdir, basedir)
             os.makedirs(targetdir, exist_ok=True)
--- a/snapcraft/plugins/maven.py
+++ b/snapcraft/plugins/maven.py
@@ -118,6 +118,7 @@ class MavenPlugin(snapcraft.plugins.jdk.JdkPlugin):
             src = os.path.join(self.builddir, f, 'target')
 
             jarfiles = glob.glob(os.path.join(src, '*.jar'))
+            warfiles = glob.glob(os.path.join(src, '*.war'))
             arfiles = glob.glob(os.path.join(src, '*.[jw]ar'))
 
             if not (arfiles):
--- a/snapcraft/plugins/maven.py
+++ b/snapcraft/plugins/maven.py
@@ -123,11 +123,8 @@ class MavenPlugin(snapcraft.plugins.jdk.JdkPlugin):
             if not (arfiles):
                 raise RuntimeError("could not find any"
                                    "built jar files for part")
-            basedir = 'war'
-            if len(f) > 0:
-                basedir = f
-            elif jarfiles:
-                basedir = 'jar'
+            basedir = 'jar' if jarfiles and len(f) == 0 else (
+                      'war' if warfiles and len(f) == 0 else f)
 
             targetdir = os.path.join(self.installdir, basedir)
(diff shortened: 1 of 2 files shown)


snappy-m-o:

snappy-m-o:

sergiusens:
ok to test

sergiusens:
:+1: from me

snappy-m-o:

snappy-m-o:

sergiusens:
ok to test

snapcraft/plugins/maven.py
```diff
@@ -103,18 +114,22 @@ def build(self):
...
+
+            jarfiles = glob.glob(os.path.join(src, '*.jar'))
+            warfiles = glob.glob(os.path.join(src, '*.war'))
+            arfiles = glob.glob(os.path.join(src, '*.[jw]ar'))
+
+            if not (arfiles):
+                raise RuntimeError("could not find any"
+                                   "built jar files for part")
+            basedir = 'jar' if jarfiles and len(f) == 0 else (
+                      'war' if warfiles and len(f) == 0 else f)
```
kyrofa: I personally think this is difficult to read. I'd prefer a normal if.

snapcraft/plugins/maven.py
```diff
@@ -103,18 +114,22 @@ def build(self):
...
+
+            targetdir = os.path.join(self.installdir, basedir)
+            os.makedirs(targetdir, exist_ok=True)
+            self.run(['cp', '-a'] + arfiles + [targetdir])
```
kyrofa: I'd prefer to use the python utilities for copying files rather than shelling out.

snapcraft/plugins/maven.py
ZenHarbinger: I get that, but then I couldn't pass the coverage and complexity.

snapcraft/plugins/maven.py
ZenHarbinger: Sure, that's just what was already there.

snapcraft/plugins/maven.py
kyrofa: Perhaps a refactor should be considered, then.

snapcraft/plugins/maven.py
sergiusens: El 07/07/16 a las 16:35, Kyle Fazzari escribió:
> In snapcraft/plugins/maven.py
> <https://github.com/snapcore/snapcraft/pull/627#discussion_r69972186>:
> 
>> +        for f in self.options.maven_targets:
>> +            src = os.path.join(self.builddir, f, 'target')
>> +
>> +            jarfiles = glob.glob(os.path.join(src, '*.jar'))
>> +            warfiles = glob.glob(os.path.join(src, '*.war'))
>> +            arfiles = glob.glob(os.path.join(src, '*.[jw]ar'))
>> +
>> +            if not (arfiles):
>> +                raise RuntimeError("could not find any"
>> +                                   "built jar files for part")
>> +            basedir = 'jar' if jarfiles and len(f) == 0 else (
>> +                      'war' if warfiles and len(f) == 0 else f)
>> +
>> +            targetdir = os.path.join(self.installdir, basedir)
>> +            os.makedirs(targetdir, exist_ok=True)
>> +            self.run(['cp', '-a'] + arfiles + [targetdir])
> 
> I'd prefer to use the python utilities for copying files rather than
> shelling out.

Ah, nice catch!

snapcraft/plugins/maven.py
kyrofa: Indeed, this isn't tested at all. Definitely missing some tests, there.

snapcraft/plugins/maven.py
sergiusens: El 07/07/16 a las 16:37, Matthew Aguirre escribió:
> In snapcraft/plugins/maven.py
> <https://github.com/snapcore/snapcraft/pull/627#discussion_r69972363>:
> 
>> +        for f in self.options.maven_targets:
>> +            src = os.path.join(self.builddir, f, 'target')
>> +
>> +            jarfiles = glob.glob(os.path.join(src, '*.jar'))
>> +            warfiles = glob.glob(os.path.join(src, '*.war'))
>> +            arfiles = glob.glob(os.path.join(src, '*.[jw]ar'))
>> +
>> +            if not (arfiles):
>> +                raise RuntimeError("could not find any"
>> +                                   "built jar files for part")
>> +            basedir = 'jar' if jarfiles and len(f) == 0 else (
>> +                      'war' if warfiles and len(f) == 0 else f)
>> +
>> +            targetdir = os.path.join(self.installdir, basedir)
>> +            os.makedirs(targetdir, exist_ok=True)
>> +            self.run(['cp', '-a'] + arfiles + [targetdir])
> 
> Sure, that's just what was already there.

@kyrofa pass the secret sauce on copytree link_or_copy ;-)

snapcraft/plugins/maven.py
ZenHarbinger: I have no idea how the dummy testing works, or how to set up a composite maven project for that. I'm willing to do it though.

snapcraft/plugins/maven.py
kyrofa: > @kyrofa pass the secret sauce on copytree link_or_copy ;-)

Ah, good idea.

Instead of copying this, you can hard-link them (and copy as a backup) using something like [this](https://github.com/snapcore/snapcraft/blob/master/snapcraft/internal/sources.py#L348).

snapcraft/plugins/maven.py
kyrofa: Yes please-- that's how we ensure this 1) works correctly, and 2) doesn't break in the future :smiley: .

ZenHarbinger: Where can I find any docs on this, it was pretty rough to understand what was there to begin with.

kyrofa:
This is looking good, thank you! A little more work to do. Also, please make sure that:

1. A bug exists for the feature you're trying to add, and
2. You sign the contributor license agreement.

More details about both of these items can be found in the [contribution guide](https://github.com/snapcore/snapcraft/blob/master/CONTRIBUTING.md).

snapcraft/plugins/maven.py
kyrofa: I'm afraid there are none. You have a good start-- you added the `maven_targets` to the `Options` class there, but you need to write some tests that actually make use of them and verify the results. [This test](https://github.com/snapcore/snapcraft/pull/627/files#diff-00b535e9137e4211b28c0c283f7b2291L78) is probably the simplest place to start, learn from, and build upon.

snapcraft/plugins/maven.py
ZenHarbinger: OK, I just checked in a fix for this, I will work on the tests later.
Use of: shutil.copy(); we don't want to copy the tree of the target directory because it would grab all the intermediary output too, not just the jar or war file.

ZenHarbinger:
Who is: Canonical Project Manager or contact for the contributor agreement?

PR #629, opened 2016-07-04
Proper message when registering an already registered snap

sergiusens:

PR #630, opened 2016-07-04
Report a nice error when registering too fast

sergiusens:
Depends on #629


integration_tests/test_store_register.py
```diff
@@ -28,18 +31,60 @@ def setUp(self):
...
+            'If you are the publisher most users expect for {0!r} '
+            'then please claim the name at').format(self.reserved_name)
+        self.assertThat(str(error.output), Contains(expected))
+        self.assertThat(str(error.output), Contains('register-name-dispute'))
+
+    def test_registrations_in_a_row_fail_if_too_fast(self):
+        # This test has a potential to fail if working off a slow
+        # network.
+        self.login(expect_success=True)
+        snap_name_1 = 'good snap'
```
elopio: This needs to be uuid'ed too. Otherwise it will fail with "already registered".

snapcraft/storeapi/errors.py
```diff
@@ -70,3 +73,50 @@ class StoreAuthenticationError(StoreError):
...
+        '{snap_name!r} then claim the name at {register_claim_url!r}')
+
+    __FMT_RESERVED = (
+        'The name {snap_name!r} is reserved.\n\n'
+        'If you are the publisher most users expect for '
+        '{snap_name!r} then please claim the name at {register_claim_url!r}')
+
+    __FMT_RETRY_WAIT = (
+        'You must wait {retry_after} minutes before trying to register '
+        'your next snap.')
```
elopio: I would report this in seconds, because according to Celso production requires a wait of 1 minute. So this will say: you must wait 1 minutes.

snapcraft/storeapi/errors.py
```diff
@@ -70,3 +73,50 @@ class StoreAuthenticationError(StoreError):
...
+        if response_json.get('status') == 409:
+            response_json['register_claim_url'] = self.__get_claim_url(
+                response_json.get('register_name_url', ''))
+            if response_json['code'] == 'already_registered':
+                self.fmt = self.__FMT_ALREADY_REGISTERED
+            if response_json['code'] == 'reserved_name':
+                self.fmt = self.__FMT_RESERVED
+        elif response_json.get('status') == 429:
+            if response_json['code'] == 'register_window':
+                print(response_json['retry_after']//60)
```
elopio: remove the print?

elopio:
lgtm, with a few details. The integration test is risky indeed, in staging. I'm happy if it passes ~10 times in a row.

integration_tests/test_store_register.py
sergiusens: Right

snapcraft/storeapi/errors.py
sergiusens: Godo point

snapcraft/storeapi/errors.py
sergiusens: Right, this was driving me insane

sergiusens:
@elopio so should we skip on integration?

elopio:
I think we can make it work, with:
snap_name_1 = 'goodsnap-{}'.format(uuid.uuid4().int)

sergiusens:
@elopio I mean, should we skip given the risk?

elopio:
Just hit re-run plenty of times, to get a sense if it will cause failures in the future.
It may not be required to skip.
(I hate and love git's live comments) :D

sergiusens:
@elopio, will merge with master, fix conflcits and address review

sergiusens:
@elopio I made a couple of more improvements here and there.

PR #631, opened 2016-07-05
Fix the store integration tests

elopio:
The tests were failing because we added a new one that attempts to register
two snaps too fast, in order to check the error the store sends for too many
attempts. After the second registration, we need to wait a little so the
following tests can register without getting the same error.

LP: #1599268

### DIFF
--- a/integration_tests/test_store_register.py
+++ b/integration_tests/test_store_register.py
@@ -80,9 +80,11 @@ class RegisterTestCase(integration_tests.TestCase):
 
         self.register(snap_name_1, wait=False)
 
+        # Wait after the registration attempt, so the following registrations
+        # don't get the error.
         error = self.assertRaises(
             subprocess.CalledProcessError,
-            self.register, snap_name_2, wait=False)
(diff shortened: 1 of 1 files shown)


integration_tests/test_store_register.py
```diff
@@ -80,9 +80,11 @@ def test_registrations_in_a_row_fail_if_too_fast(self):
 
         self.register(snap_name_1, wait=False)
 
+        # Wait after the registration attempt, so the following registrations
+        # don't get the error.
         error = self.assertRaises(
             subprocess.CalledProcessError,
-            self.register, snap_name_2, wait=False)
+            self.register, snap_name_2, wait=True)
```
sergiusens: this won't wait as the exception would be raised. We either need to catch and reraise in `<class>.register` or wait and the end of the test here (`addCleanup`)

elopio: damn, you are right. I'll addCleanup.

elopio:
Todo fríamente calculado ^_^ I just wanted to show red first, and then green ;)

PR #632, opened 2016-07-06
Refactor and update the logic that deals with the snap-push store endpoint

sergiusens:


snapcraft/storeapi/__init__.py
```diff
@@ -328,11 +316,90 @@ def register(self, snap_name, series):
...
+
+    def __init__(self, status_details_url):
+        self.__status_details_url = status_details_url
+
+    def track(self):
+        queue = Queue()
+        thread = Thread(target=self._update_status, args=(queue,))
+        thread.start()
+
+        widgets = ['Processing...', AnimatedMarker()]
```
kyrofa: Would it look better if there was a space between these two?

sergiusens: I don't know, is this a question or a statement? :-)

kyrofa: Haha, a question. But you can ignore it, this looks good to me :+1: .

sergiusens:
LP: #1572399 and LP: #1599953

PR #634, opened 2016-07-06
Run the parser integration tests with debug flag

elopio:


elopio:
Master already is calling the command with --debug. Closing...

PR #635, opened 2016-07-07
Fix parts cache handling

cjwatson:
requests sends "Accept-Encoding: gzip, deflate" by default which causes
parts.snapcraft.io not to send an ETag, so the logic for detecting when
the parts cache is already up to date didn't work.  Use Last-Modified /
If-Modified-Since instead of ETag / If-None-Match.


sergiusens:
Lovely, thanks. fwiw we also have a .gz to download because I really want Content-Length back too :smile:

snappy-m-o:

snappy-m-o:

PR #640, opened 2016-07-08
Setup the FakeParts store for all unit tests

sergiusens:
Now that we run snapcraft update by deftault to get an initial
remote parts cache, most (not all) test require the FakeParts
fixture to be setup.

LP: #1600273


kyrofa:
Looks good to me :+1:

elopio:
:+1:

PR #641, opened 2016-07-08
yield the correct exception when retries are out

sergiusens:
Small refactor to handle connection retries being
exhausted when trying to query status_details_url
from the snap-push store endpoint.

LP: #1600022

### DIFF
--- a/snapcraft/storeapi/__init__.py
+++ b/snapcraft/storeapi/__init__.py
@@ -395,11 +395,12 @@ class StatusTracker:
 
     def _get_status(self):
         connection_errors_allowed = 10
-        while connection_errors_allowed:
+        while True:
             try:
                 content = requests.get(self.__status_details_url).json()
             except (requests.ConnectionError, requests.HTTPError) as e:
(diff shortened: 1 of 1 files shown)


snapcraft/storeapi/__init__.py
```diff
@@ -395,11 +395,12 @@ def _update_status(self, queue):
 
     def _get_status(self):
         connection_errors_allowed = 10
-        while connection_errors_allowed:
+        while True:
             try:
                 content = requests.get(self.__status_details_url).json()
             except (requests.ConnectionError, requests.HTTPError) as e:
+                if not connection_errors_allowed:
+                    yield e
```
elopio: this reads so nice!

elopio:
:+1:

snapcraft/storeapi/__init__.py
sergiusens: Thanks, if I intend to use a variable or method in a conditional I try to make it readable

PR #644, opened 2016-07-09
Feature/ant plugin test

ZenHarbinger:
Updated the ant plugin to remove calls to self.run() for copy and mkdir -p
Use of shutil.copy() and os.makedirs.

Added ant test file.

https://bugs.launchpad.net/snapcraft/+bug/1600410

### DIFF
--- a/snapcraft/plugins/ant.py
+++ b/snapcraft/plugins/ant.py
@@ -27,6 +27,7 @@ For more information check the 'plugins' topic for the former and the
 import glob
 import logging
 import os
+import shutil
 
 import snapcraft
 import snapcraft.common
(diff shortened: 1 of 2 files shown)


snappy-m-o:

snappy-m-o:

snapcraft/plugins/ant.py
```diff
@@ -49,8 +50,9 @@ def build(self):
         if not files:
             raise RuntimeError('could not find any built jar files for part')
         jardir = os.path.join(self.installdir, 'jar')
-        self.run(['mkdir', '-p', jardir])
-        self.run(['cp', '-a'] + files + [jardir])
+        os.makedirs(jardir)
+        for f in files:
+            shutil.copy(f, jardir)
```
sergiusens: may I suggest using `os.link` here?

sergiusens:
Great conribution! I always like seeing tests!
I added a small minor comment, please tell me what you think, but it would reduce disk consumption if done.

ZenHarbinger:
If os.link is that much better, sure. I'll push a change up in a second.

ZenHarbinger:
Eh, I take it back, can we just copy since otherwise I have to get the file name so to create the link instead of just the target directory to copy to?

sergiusens:
ok to test

PR #648, opened 2016-07-12
Implemented `snapcraft release`

sergiusens:


integration_tests/test_store_release.py
```diff
@@ -0,0 +1,97 @@
...
+)
+
+import integration_tests
+
+from snapcraft.tests import fixture_setup
+
+
+class ReleaseTestCase(integration_tests.TestCase):
+
+    def _update_name_and_version(self, project_dir, name=None, version=None):
```
elopio: This could be moved to integration_tests.TestCase.

snapcraft/_store.py
```diff
@@ -111,6 +112,58 @@ def upload(snap_filename):
...
+def _get_text_for_opened_channels(opened_channels):
+    if len(opened_channels) == 1:
+        return 'The {!r} channel is now open.'.format(opened_channels[0])
+    else:
+        text = ''
+        for channel in opened_channels[:-1]:
+            if text:
+                text = '{}, {!r}'.format(text, channel)
+            else:
+                text = '{!r}'.format(channel)
```
elopio: text = ', '.join(['{!r}'.format(channel) for channel in opened_channels[:-1]])

snapcraft/_store.py
```diff
@@ -111,6 +112,58 @@ def upload(snap_filename):
...
+        text = '{} and {!r}'.format(text, opened_channels[-1])
+        return 'The {} channels are now open.'.format(text)
+
+
+def _get_text_for_channel(channel):
```
elopio: This should return a tuple of three items, not a list. So instead of [channel, version, revision], use (channel, version, revision)

snapcraft/_store.py
```diff
@@ -111,6 +112,58 @@ def upload(snap_filename):
...
+
+    if 'opened_channels' in channels:
+        logger.info(
+            _get_text_for_opened_channels(channels['opened_channels']))
+    channel_map = channels['channel_map']
+    parsed_channels = [_get_text_for_channel(c) for c in channel_map]
+    tabulated_channels = tabulate(parsed_channels,
+                                  headers=['Channel', 'Version', 'Revision'])
+    # This does not look good in green so we print instead
+    print('{}'.format(tabulated_channels))
```
elopio: Why not just print(tabulated_channels) ?

Using print will probably come back and bite us when we fix the bug about saving the logs to a file. At that point, I suppose we should log with debug level an untabulated version of the channels. I'm ok leaving this for later.

snapcraft/storeapi/errors.py
```diff
@@ -170,3 +170,31 @@ class StoreReviewError(StoreError):
     def __init__(self, result):
         self.fmt = self.__messages[result['code']]
         super().__init__()
+
+
+class StoreReleaseError(StoreError):
+
+    __FMT_NOT_REGISTERED = (
+        'Sorry, try `snapcraft register {snap_name}` before trying to '
+        'release or choose an existing revision.')
```
elopio: I'm not convinced about this message, mainly because of the second part. I think I would just remove "or choose an existing revision". Or, it could say "or choose an already registered snap name". I'm not sure if that's better though.

snapcraft/tests/fake_servers.py
```diff
@@ -288,7 +292,7 @@ def _handle_upload_request(self):
             int(self.headers['Content-Length'])).decode('utf8')
         data = json.loads(string_data)
         logger.debug(
-            'Handling registration request with content {}'.format(data))
+            'Handling upload request with content {}'.format(data))
```
elopio: nice catch.

elopio:
I'm like +0.8 here. I left a few comments for discussion.

snapcraft/_store.py
sergiusens: The `format` is a left over from when I was trying to not use `tabulate` so nice catch

snapcraft/storeapi/errors.py
sergiusens: Well the error shows up for both and I already requested OLS to provide better errors for when each scenario for a 404 happens

elopio:
Here's my pending +0.2. Thanks!

PR #654, opened 2016-07-13
Gradle option fix

ZenHarbinger:
The gradle plugin to snapcraft puts the options after the task, should be before when invoking the gradlew command.


snappy-m-o:

snappy-m-o:

PR #658, opened 2016-07-14
parser - Return non-zero code for wiki errors.

josepht:
LP:#1602394

### DIFF
--- a/snapcraft/internal/parser.py
+++ b/snapcraft/internal/parser.py
@@ -260,21 +260,16 @@ def _process_wiki_entry(entry, master_parts_list):
     try:
         data = yaml.load(entry)
     except ScannerError as e:
-        logger.warning(
-            'Bad wiki entry, possibly malformed YAML: {}'.format(e))
-        raise InvalidEntryError('Bad wiki entry: {}'.format(entry))
+        raise InvalidEntryError(
+            'Bad wiki entry, possibly malformed YAML for entry: {}'.format(e))
 
     part_name = data.get('project-part')
     if part_name is not None and part_name in master_parts_list:
-        logger.warning("Duplicate part found in wiki: {}".format(
-            part_name))
-        raise InvalidEntryError('duplicate entry: {}'.format(entry))
+        raise InvalidEntryError(
+            'Duplicate part found in the wiki: {} in entry {}'.format(
+                part_name, entry))
 
-    try:
-        parts_list, after_parts = _process_entry(data)
-    except InvalidEntryError as e:
-        logger.warning('Invalid wiki entry: {}'.format(e))
-        raise e
+    parts_list, after_parts = _process_entry(data)
 
     if is_valid_parts_list(parts_list, after_parts):
         master_parts_list.update(parts_list)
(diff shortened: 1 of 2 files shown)


josepht:
@sergiusens I can't add or remove labels.  Is that a power you can grant me?

snapcraft/internal/parser.py
```diff
@@ -267,15 +274,15 @@ def _process_index(output):
...
                 entry = ''
         else:
             entry = '\n'.join([entry, line])
 
     if entry:
-        _process_wiki_entry(entry, master_parts_list)
+        errors['wiki'] += _process_wiki_entry(entry, master_parts_list)
 
-    return master_parts_list
+    return (master_parts_list, errors)
```
elopio: I like that we are returning the number of errors, but in python we shouldn't return a tuple like in go.
Instead, if errors >  0, raise an exception. You can add an attribute to the WikiError class that's number_of_errors, so you can extract it later.

snapcraft/internal/parser.py
```diff
@@ -230,33 +233,37 @@ class Options:
...
         master_parts_list.update(parts_list)
 
+    return 0
+
 
 def _process_index(output):
     # XXX: This can't remain in memory if the list gets very large, but it
     # should be okay for now.
     master_parts_list = OrderedDict()
+    errors = dict(wiki=0)
```
elopio: I got lost with this. Do we have any other kind of errors? Why isn't it just a counter?

josepht: The thinking is that we'd have at least two classes of errors; 1. wiki entry errors (i.e. malformed yaml, missing fields, etc.) 2. snapcraft.yaml errors that make constructing the fully parsed part impossible.  For either one we want to be able to continue processing the subsequent entries.  I'll look into handling this with exceptions.

sergiusens:
Looks good, can you add a unit test showing this wiki parsing problem count?

sergiusens:
retest this please

snapcraft/internal/parser.py
```diff
@@ -260,24 +256,25 @@ class Options:
...
             'Bad wiki entry, possibly malformed YAML: {}'.format(e))
-        return
+        raise InvalidEntryError('Bad wiki entry: {}'.format(entry))
 
     part_name = data.get('project-part')
     if part_name is not None and part_name in master_parts_list:
         logger.warning("Duplicate part found in wiki: {}".format(
             part_name))
-        return
+        raise InvalidEntryError('duplicate entry: {}'.format(entry))
```
kyrofa: You capitalized the one above. Should this one be so as well?

kyrofa: Also, the warning and exception messages are redundant. Perhaps it would be better to _only_ raise here, and log in the catcher of the exception instead?

snapcraft/tests/test_parser.py
```diff
@@ -722,11 +727,6 @@ def test_carriage_returns(self, mock_get, mock_get_origin_data):
 origin: lp:snapcraft-parser-example
 description: example main
 project-part: main\r
----
-maintainer: Jim Doe <jim.doe@example.com>
```
kyrofa: Who the heck was Jim, anyway? John's brother?

PR #659, opened 2016-07-14
Support "never" as a daemon restart condition.

kyrofa:
This PR fixes LP: [#1595661](https://bugs.launchpad.net/ubuntu/+source/snapd/+bug/1595661) by allowing "never" as a restart condition. Note that "never" is not a valid systemd restart condition, but snapcore/snapd#1551 maps "never" into systemd's "no."


josepht:
LGTM.

elopio:
:+1:

PR #660, opened 2016-07-14
Fix python2 compile cache removal

SamYaple:
\_\_pycache\_\_ is python3. python2 generates .pyc files in the folders
next to the origin .py file.

There is a python3.5 related \_\_pycache\_\_ directory that is excluded
as well.

LP:#1603200


snappy-m-o:

snappy-m-o:

sergiusens:
ok to test

sergiusens:
Thanks for the fix!

PR #664, opened 2016-07-15
New plugin: Script (runs bash scripts)

monsterjamp:
Allows the user to use bash scripts during the build and pull stages.

Fixes LP#1604107
I've signed the CLA and my username on Launchpad is filjoa.


snappy-m-o:

snappy-m-o:

sergiusens:
Hi, thanks for your contribution. There are some simple fixes required like fixing the Copyright dates. It also feels like this plugin should be called `script` (run) instead of `bash`.

On the usefulness side, aside from the integration test is there any real world project that could use this already?

Last but not least, is there any benefit of not doing this in a custom plugin itself?

snapcraft/plugins/bash.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def __init__(self, name, options, project):
+        super().__init__(name, options, project)
+        self.build_packages.append('bash')
+
+    def pull(self):
+        super().pull()
+
+        if "pull" in self.options.stages:
+            command = [self.sourcedir + '/' + self.options.script]
```
sergiusens: please use `os.path.join` here.

monsterjamp:
I renamed the plugin to `script` and fixed the copyright dates.

I don't know about any real world projects that use bash scripts but I do have a project that I'm working on that would need to run a bash script in order to allow others to easily build. The project has a dependency that can only be obtained by going to the creator's website, logging in, downloading the dependency, and then manually installing the headers/libraries.

I'm not sure I completely understand your last question but having an existing bash plugin shipped with snapcraft means the user doesn't have to do any extra fiddling trying to get their project snapped. I haven't been on the IRC room for that long but I've already seen a few people asking how to run their bash scripts.

snapcraft/plugins/script.py
```diff
@@ -0,0 +1,106 @@
...
+This plugin uses the common plugin keywords as well as those for "sources".
+For more information check the 'plugins' topic for the former and the
+'sources' topic for the latter.
+Additionally, this plugin uses the following plugin-specific keyword:
+    - script-name:
+      (string)
+      Run the given script name.
+    - stages:
+      (list of strings)
+      Run the script in the given stages. (default: ['build'])
```
elopio: You could have a script for pull, and a different one for build. Or you could have to run two commands in one of those phases.

I'm not quite sure about this plugin, because you are trying to wrap a wild (and sometimes a little random :) thing like a custom build script into a generic workflow with fixed phases and arguments. So there will be a lot of cases that will need modifications on this plugin to work.

I think that instead of providing a plugin to run a script, we should provide better guidelines to let people make their own custom plugin that runs whatever they need to pull, and then whatever they need to build. In these past weeks I've seen all kinds of crazy things during build, I see no good way to make them behave in a similar way.

Grumpiness aside, welcome to the snapcraft! Your contribution is really appreciated. Thank you.

monsterjamp: I understand what you're saying, but I still think it wouldn't hurt to at least have a simple bash plugin to base custom plugins from. Also I would imagine that running one command for the building phase and maybe another command for the pulling phase covers the majority of the use cases for bash based project building.

kyrofa:
Yeah, I gotta say I agree with @elopio here. This is opening a can of worms: it feels like a one-size-fits-all solution to a problem that vastly differs from project to project. I see its features quickly bloating as people want capabilities for their specific problem, when we already have a solution for weird problems: local plugins (or just write a Makefile).

sergiusens:
Closing as we are discussing in #727

PR #666, opened 2016-07-18
Allow / in parts

sergiusens:
Subparts from remote parts are separated by a / so we now need
too support this scenario in the parts definition.

LP: #1602728

### DIFF
--- a/snapcraft/_baseplugin.py
+++ b/snapcraft/_baseplugin.py
@@ -85,6 +85,10 @@ class BasePlugin:
         self.project = project
         self.options = options
 
+        # The remote parts can have a '/' in them to separate the main project
+        # part with the subparts. This is rather unfortunate as it affects the
+        # the layout of parts inside the parts directory causing collisions
+        # between the main project part and its subparts.
         part_dir = name.replace('/', '\N{BIG SOLIDUS}')
         if project:
             self.partdir = os.path.join(project.parts_dir, part_dir)
--- a/snapcraft/_baseplugin.py
+++ b/snapcraft/_baseplugin.py
@@ -85,10 +85,11 @@ class BasePlugin:
         self.project = project
         self.options = options
 
+        part_dir = name.replace('/', '\N{BIG SOLIDUS}')
         if project:
-            self.partdir = os.path.join(project.parts_dir, self.name)
+            self.partdir = os.path.join(project.parts_dir, part_dir)
         else:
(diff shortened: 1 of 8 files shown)


kyrofa:
This is really weird :stuck_out_tongue: . But it works! :+1: once the tests pass.

kyrofa:
Failure on both:

```
FAIL: test_load_config_with_invalid_plugin_exits_with_error (snapcraft.tests.test_cmds.TestCommands)
  File "/tmp/adt-run.yY8AY2/build.qOh/real-tree/snapcraft/tests/test_cmds.py", line 58, in test_load_config_with_invalid_plugin_exits_with_error
    fake_logger.output)
AssertionError: 'Issue while loading plugin: unknown plugi[15 chars]st\n' != 'Starting new HTTP connection (1): localho[97 chars]st\n'
+ Starting new HTTP connection (1): localhost
+ "GET /parts.yaml HTTP/1.1" 200 1000
  Issue while loading plugin: unknown plugin: does-not-exist


```

snapcraft/_baseplugin.py
```diff
@@ -85,10 +85,11 @@ def __init__(self, name, options, project=None):
         self.project = project
         self.options = options
 
+        part_dir = name.replace('/', '\N{BIG SOLIDUS}')
```
elopio: why don't we just make it a subdir? I mean, keep the slash. Like partdir = os.path.join(project.parts_dir, *name.split('/'))

elopio: anyway, if you decide to keep the character that looks like slash but it's not, a comment is required here explaining why.

sergiusens: Sure. And I will explain why the subdir is not a good idea

elopio:
retest this please

sergiusens:
retest this please

ZenHarbinger:
This is causing me issues:
With the latest master, if I run snapcraft on the demos, I get the following error output:
`'^(?!plugins$)[a-z0-9][a-z0-9+-\/]*$'`
If I go in and take out the '\/' portion of the expression in _schema.py, it works again.

Am I missing something?

ZenHarbinger:
Blah, I figured it out, residual install from /usr/share/snapcraft that was not pointing to /usr/local/lib/python3/.../snapcraft/share.
Sorry!

PR #669, opened 2016-07-19
Add detection for an existing `.snapcraft.yaml` file before running `snapcraft init`

tsimonq2:
(This fixes bug [1604326](https://pad.lv/1604326))

This ensures that an extra `snapcraft.yaml` isn't created when a `.snapcraft.yaml` file already exists.


snappy-m-o:

snappy-m-o:

sergiusens:
Thanks for the quick fix.
I would appreciate a unit test for this though

tsimonq2:
@sergiusens this now has a unit test

PR #671, opened 2016-07-19
Add initial snapcraft manpage

tsimonq2:
This adds a manpage for Snapcraft, fixing [bug 1506764](https://pad.lv/1506764).


snappy-m-o:

snappy-m-o:

sergiusens:
thanks for this, but it would be good to have the manpage in troff/nroff (or if there is anything better, please use that) and have debhelper do the right thing.

tsimonq2:
@sergiusens that is what I did, I just used a separate directory and put an entry in the `snapcraft.install` file.

I since modified this to use the default debhelper method.

debian/manpages/snapcraft.1
```diff
@@ -0,0 +1,87 @@
...
+
+.SH PARTS ECOSYSTEM COMMANDS
+update
+    Updates the parts listing from the cloud.
+
+define
+    Shows the definition for the cloud part.
+
+search
+    Searches the remotes part cache for matching parts.
```
seb128: "the remotes part" should be "the remote parts" no?

tsimonq2: I literally just took this from `snapcraft help` so it should be corrected there as well.

debian/manpages/snapcraft.1
```diff
@@ -0,0 +1,87 @@
+.TH snapcraft 1 "19 July 2016" "version 2.13"
+.SH NAME
+Snapcraft - Snapcraft is a delightful packaging tool
+.SH SYNOPSIS
+snapcraft [options] <command>
+
+snapcraft (-h | --help)
+
+snapcraft --version
```
cjwatson: This needs a bunch of delicate formatting.  Something like:

    .B snapcraft
    .RI [ options ]
    .I command
    .br
    .B snapcraft
    .RB [\| \-h \||\| \-\-help \|]
    .br
    .B snapcraft
    .B \-\-version

debian/manpages/snapcraft.1
```diff
@@ -0,0 +1,87 @@
...
+.SH DESCRIPTION
+Snapcraft helps you assemble a whole project in a single tree out of many pieces. It can drive a very wide range of build and packaging systems, so that you can simply list all the upstream projects you want and have them built and installed together as a single tree.
```
cjwatson: It's generally good style in troff to start a new line after each piece of sentence-final punctuation.  The toolchain will use slightly better typography in some output formats if you do that.

debian/manpages/snapcraft.1
```diff
@@ -0,0 +1,87 @@
...
+.SH OPTIONS
+--help
+    Prints the available Snapcraft commands.
+
+--version
+    Prints the version of Snapcraft currently installed on your system.
```
cjwatson: Don't manually format definition lists with spaces this way (if you ever find yourself typing leading spaces on a line in troff, you're probably doing something wrong).  Do this instead:

    .TP
    .B \-\-help
    Prints the available Snapcraft commands.
    .TP
    .B \-\-version
    Prints the version of Snapcraft currently installed on your system.

The same goes for other such lists below.

debian/manpages/snapcraft.1
```diff
@@ -0,0 +1,87 @@
...
+    Updates the parts listing from the cloud.
+
+define
+    Shows the definition for the cloud part.
+
+search
+    Searches the remotes part cache for matching parts.
+
+.SH REPORTING BUGS
+To report a bug in Snapcraft, navigate to Launchpad: https://bugs.launchpad.net/snapcraft/+filebug
```
cjwatson: "Navigate" is a somewhat pedantic verb to use here.  I'd just say "go to Launchpad".

debian/snapcraft.manpages
```diff
@@ -0,0 +1 @@
+manpages/snapcraft.1
```
cjwatson: This is of course specific to the Debian package build.  If you can, I think it would be better to install the file in setup.py (perhaps in `data_files`), so that it can apply more easily to other distributions; if you do this right it shouldn't be necessary to touch `debian/` at all.

sergiusens:
We are looking into rewritting a full user friendly manpage for this so I don't think we want to get into this as is as it will be a retro fitting exercise otherwise.

sorry for the time invested in this.

PR #672, opened 2016-07-19
Capture the correct exception when not being able to decode json

sergiusens:


snapcraft/storeapi/errors.py
```diff
@@ -105,7 +105,7 @@ class StoreRegistrationError(StoreError):
     def __init__(self, snap_name, response=None):
         try:
             response_json = response.json()
-        except AttributeError:
+        except JSONDecodeError:
```
elopio: shouldn't you catch both?

elopio:
:+1:

PR #675, opened 2016-07-20
Allow godeps to fetch Go dependencies

stevenwilkin:
Addresses LP: [#1595981](https://bugs.launchpad.net/snapcraft/+bug/1595981)

This PR crudely allows [godeps](https://launchpad.net/godeps) to manage dependencies for Go projects, for example [snapweb](https://github.com/snapcore/snapweb).

One thing to note is that `go get` still gets run after `godeps` which is maybe redundant but this effort is the simplest way of closing the bug.


kyrofa:
Closing in favor of #691.

PR #676, opened 2016-07-20
Special handling for pc files for conflicts

sergiusens:
The .pc files can have a different prefix when coming from the the parts
installdir and messes up the conflict checker so we handle this case when
doing the check for .pc files.

To make this work correctly we need to skip os.linking when dealing with
.pc files.

LP: #1604472


elopio:
ship it!

kyrofa:
:+1:

PR #680, opened 2016-07-22
Rewrite shebangs generated by the python plugins.

kyrofa:
This PR resolves LP: [#1597919](https://bugs.launchpad.net/snapcraft/+bug/1597919) by walking through the install directories for python{2,3} parts, ensuring all python shebangs are using `env` instead of being paths.


sergiusens:
Looks great. I would like to see an integration test for the python related integration tests if possible (just extend them), bonus points if you add an LP: #XXXXX comment :wink:

snapcraft/plugins/python2.py
```diff
@@ -168,6 +169,11 @@ def build(self):
              '--prefix={}/usr'.format(self.installdir),
              ], cwd=self.builddir)
 
+        # Fix all shebangs to use the in-snap python.
+        common.replace_in_file(self.installdir, re.compile(r''),
+                               re.compile(r'#!.*python'),
```
sergiusens: Maybe we want to be clever here and not inadvertenly replace something we will regret?

A better match might be `r'#!{}.*python'.format(self.installdir)`, what do you think?

kyrofa:
> Looks great. I would like to see an integration test for the python related integration tests if possible (just extend them), bonus points if you add an LP: #XXXXX comment

Done. Give me those bonus points!

snapcraft/plugins/python2.py
kyrofa: I can't imagine a scenario where we don't want to use `#!/usr/bin/env python` instead of whatever they have. If they _don't_ use installdir in the shebang it'll still be broken, no?

sergiusens:
:100:

PR #684, opened 2016-07-25
Run tests from /tmp

sergiusens:
LP: #1606290


elopio:
:+1: if all the tests pass.

PR #687, opened 2016-07-26
Enable integration tests to run in the production store

elopio:
LP: #1606677


sergiusens:
kudos on the cleanup!

celso-providelo:
Thanks Leo,

The TestStore consolidation is much clearer and the register name intervals fit our configurations.

PR #688, opened 2016-07-26
Strip common path prefixes from linkname as well as name when extracting a tarfile

mhall119:
Stripping common path prefixes from the member name breaks hardlink name resolution unless you perform the same operation on the member linknames


snappy-m-o:

snappy-m-o:

kyrofa:
@mhall119 according to the integration tests, this seems to bust the nodejs and rust plugins.

kyrofa:
ok to test

PR #689, opened 2016-07-27
kernel plugin: kernel targets depending on debarch

sergiusens:
This allows using the same snapcraft.yaml to build kernels
for different arches as they have different image targets.

LP: #1604801


sergiusens:
retest this please

sergiusens:
@leannogasawara any updates on testing for this?

leannogasawara:
On Wed, Aug 3, 2016 at 6:21 PM, Sergio Schvezov <notifications@github.com>
wrote:

> @leannogasawara <https://github.com/leannogasawara> any updates on
> testing for this?
>

Yep, Paolo tested this last week:

<goog_1295843>
https://bugs.launchpad.net/snapcraft/+bug/1604801/comments/3

"With this patch i was able to snap, from the same kernel tree, kernels for
amd64, arm64 and armhf - looks good."

Thanks,
Leann

sergiusens:
thanks @leannogasawara :-)

snapcraft/plugins/kernel.py
```diff
@@ -23,8 +23,8 @@
 The following kernel specific options are provided by this plugin:
 
     - kernel-image-target:
-      (string; default: bzImage)
-      the kernel image make target to build; maps to make target.
+      (yaml object; default: bzImage)
```
elopio: Shouldn't this be: a string or a yaml object?
I think the help should explain that you can pass only one target, and how.

snapcraft/plugins/kernel.py
```diff
@@ -157,6 +153,23 @@ def enable_cross_compilation(self):
             self.project.kernel_arch))
         self.make_cmd.append('CROSS_COMPILE={}'.format(
             self.project.cross_compiler_prefix))
+        self._set_kernel_targets()
```
elopio: I'm confused here. Why do you call it again if you already called it in the init?

sergiusens: Backwards compatibility.

sergiusens: It depends on kernel arch from the debarch

snapcraft/plugins/kernel.py
sergiusens: Yeah probably so

snapcraft/plugins/kernel.py
```diff
@@ -77,7 +82,10 @@ def schema(cls):
         schema = super().schema()
 
         schema['properties']['kernel-image-target'] = {
-            'type': 'string',
+            'oneOf': [
+                {'type': 'string'},
+                {'type': 'object'},
+            ],
```
kyrofa: I didn't know you could do this! Would have made some earlier changes easier.

kyrofa:
I like it :+1:

PR #690, opened 2016-07-27
Preserve file ownership for 'os' snaps.

josepht:
LP:#1605903


snapcraft/internal/pluginhandler.py
```diff
@@ -650,15 +655,46 @@ def _migratable_filesets(fileset, srcdir):
...
+        dst = os.path.join(dstdir, directory)
+        os.makedirs(dst, exist_ok=True)
+        if keep_owner:
+            try:
+                st = os.stat(src, follow_symlinks=follow_symlinks)
+                uid = st.st_uid
+                gid = st.st_gid
+                os.chown(dst, uid, gid, follow_symlinks=follow_symlinks)
+            except OSError as e:
+                print("JOE: e: {}".format(e))
```
kyrofa: Heh.

kyrofa:
I'd like to see some unit and integration tests for this one.

snapcraft/internal/pluginhandler.py
```diff
@@ -650,15 +655,46 @@ def _migratable_filesets(fileset, srcdir):
...
-        os.makedirs(os.path.join(dstdir, directory), exist_ok=True)
+        src = os.path.join(srcdir, directory)
+        dst = os.path.join(dstdir, directory)
+        os.makedirs(dst, exist_ok=True)
+        if keep_owner:
+            try:
+                st = os.stat(src, follow_symlinks=follow_symlinks)
+                uid = st.st_uid
+                gid = st.st_gid
+                os.chown(dst, uid, gid, follow_symlinks=follow_symlinks)
```
kyrofa: I know ogra said this solved the issue, but I don't quite understand how. The LP builders run as root, which means the files are owned by root. How are `uid` and `gid` not root here?

josepht: There are build steps that set a few file ownerships such as /home/ubuntu

kyrofa: Ah, when building the rootfs? Okay, gotcha.

kyrofa: Honestly I can't imagine a scenario where we don't want this to be the case. I say we just do it all the time. Thoughts?

kyrofa: Would `shutil.copystat()` do what you needed here?

josepht: shutil.copystat() looks like just the thing.  Thanks.

josepht: Where we use '--all-root' for squashfs would override the behavior, but that might not really matter since we're being explicit about it.

kyrofa: Oh man, I forgot about that. Then by all means, do this all the time! That greatly simplifies things, no?

snapcraft/_options.py
```diff
@@ -137,13 +137,14 @@ def snap_dir(self):
         return os.path.join(self.__project_dir, 'prime')
 
     def __init__(self, use_geoip=False, parallel_builds=True,
-                 target_deb_arch=None):
+                 target_deb_arch=None, project_type="app"):
```
sergiusens: single quotes please

sergiusens: also call this `snap_type` instead of `project_type`

sergiusens:
Looks good, minor comments and as @kyrofa says, units and integrations would be nice.

ogra1:
it isnt only /home/ubuntu, many daemons are creating their own system user or group and make their workdirs owned by these users or groups with a chown call from a maintainer script at install time...
note that dpkg/apt always run as root when installing packages, so "the build server runs as root" doesnt really make any difference. 
this change is needed for the same reasons we run mksquashfs without --all-root, we need the system users and their ownerships to boot :)

snapcraft/_options.py
josepht: I'll actually remove this since we no longer retain ownership conditionally based on the snap type.

snapcraft/internal/pluginhandler.py
```diff
@@ -650,15 +650,34 @@ def _migratable_filesets(fileset, srcdir):
...
+            pass
+
+
 def _migrate_files(snap_files, snap_dirs, srcdir, dstdir, missing_ok=False,
                    follow_symlinks=False, fixup_func=lambda *args: None):
-    for directory in snap_dirs:
-        os.makedirs(os.path.join(dstdir, directory), exist_ok=True)
 
+    _migrate_files_directories(snap_dirs, srcdir, dstdir,
+                               follow_symlinks=follow_symlinks)
```
kyrofa: I don't see a benefit to extracting this into another function now that the conditional snap types are gone. Can we move it back in?

snapcraft/internal/pluginhandler.py
```diff
@@ -650,15 +650,34 @@ def _migratable_filesets(fileset, srcdir):
...
     for snap_file in snap_files:
         src = os.path.join(srcdir, snap_file)
         dst = os.path.join(dstdir, snap_file)
         os.makedirs(os.path.dirname(dst), exist_ok=True)
+        try:
+            shutil.copystat(src, dst, follow_symlinks=follow_symlinks)
+        except FileNotFoundError:
+            logger.info("src not found: {!r}".format(src))
+            pass
```
kyrofa: In what situation would this exception be thrown, and how can it be non-fatal?

snapcraft/internal/lifecycle.py
```diff
@@ -107,6 +107,7 @@ class _Executor:
 
     def __init__(self, config, project_options):
         self.config = config
+        project_options.project_type = config.data.get('type', '')
```
kyrofa: Is this still necessary?

snapcraft/internal/pluginhandler.py
josepht: There are 30+ tests that fail for files that aren't in src.  install/foo, install/bar, etc.  Do you think the tests are broken?  If so, I can chase that down and not catch that exception.  Does the same apply to the directory copystat above as well?

josepht: Nevermind, it was my fault.  I failed to call copystat on the directory only.

snapcraft/internal/lifecycle.py
josepht: No, removed.

sergiusens:
using sudo in a test tells me we probably just want to mock/fake this part with enough certainty that it will work.

elopio:
lgtm :+1:

sergiusens:
retest this please

PR #698, opened 2016-07-29
Add option disable-parallel for all plugins.

blakerouse:
Some projects do not support parallel building with autotools. For the maas-rack charm I need to disable parallel building just for specifically building isc-dhcp-server. This gives the ability for a developer to disable this option per part, allowing those parts that can build in parallel to do so.


snappy-m-o:

snappy-m-o:

kyrofa:
Does `--no-parallel-build` work for this use-case?

```
$ snapcraft -h
  <snip>
  snapcraft [options] build [<part> ...] [--no-parallel-build]
  <snip>

Options specific to building:
  --no-parallel-build                   use only a single build job per part
                                        (the default number of jobs per part is
                                        equal to the number of CPUs)

  <snip>
```

kyrofa:
I'm going to go ahead and close this as there's already a way to accomplish what you're trying to do. Please log a bug if you still think it's an issue and we can discuss there.

blakerouse:
Please think about re-opening this. Its rather important for a MAAS snap.

https://bugs.launchpad.net/snapcraft/+bug/1608994

snappy-m-o:

snappy-m-o:

sergiusens:
I've reopened as the bug makes sense. Some project intrinsincly do not support this.

kyrofa:
Agreed, thank you for the bug, @blakerouse. We should consider supporting this in a more generic manner rather than implementing it within each plugin, though.

blakerouse:
Supporting it more generically would be awesome.

blakerouse:
I have updated this to make it more generic. It now works across all plugins as it has been added to the BasePlugin.

sergiusens:
ok to test

sergiusens:
retest this please and DO NOT, I say DO NOT timeout

sergiusens:
\o/

PR #703, opened 2016-08-01
Update the completion list of commands

seb128:
Update the completion list of commands #703


snappy-m-o:

snappy-m-o:

sergiusens:
Thanks for this, do you btw know how to make commands that take filenames (snaps) to drt wrt completion?

PR #704, opened 2016-08-01
Slightly tweak the "no jar files" errors (lp: #1606352)

seb128:
https://bugs.launchpad.net/snapcraft/+bug/1606352

### DIFF
--- a/snapcraft/plugins/ant.py
+++ b/snapcraft/plugins/ant.py
@@ -47,7 +47,7 @@ class AntPlugin(snapcraft.plugins.jdk.JdkPlugin):
         self.run(['ant'])
         files = glob.glob(os.path.join(self.builddir, 'target', '*.jar'))
         if not files:
-            raise RuntimeError('could not find any built jar files for part')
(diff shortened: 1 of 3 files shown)


snappy-m-o:

snappy-m-o:

snapcraft/plugins/ant.py
```diff
@@ -47,7 +47,7 @@ def build(self):
         self.run(['ant'])
         files = glob.glob(os.path.join(self.builddir, 'target', '*.jar'))
         if not files:
-            raise RuntimeError('could not find any built jar files for part')
+            raise RuntimeError('Could not find any built jar files for part')
```
sergiusens: maybe even better and make it
```python
raise RuntimeError('Could not find any built jar files for part {!r}'.format(self.name))
```

sergiusens: same for all others.

sergiusens:
Thanks for the improvement to the plugin, I've asked for a minor addition, feel free to say no as this contribution is already good

PR #706, opened 2016-08-02
Added "camera" plug into the example

liu-xiao-guo:
"camera" plug is needed to get the example working, and we should do the manual connection like to make it working:

$ sudo snap connect webcam-webui:camera ubuntu-core:camera


snappy-m-o:

snappy-m-o:

PR #707, opened 2016-08-02
Use the proper requirements.txt path

sergiusens:
fix the implementation to look for requirements.txt in the
part's source dir.

LP: #1588855


kyrofa:
Nice, looks good :+1:

integration_tests/snaps/pip-requirements-file/snapcraft.yaml
```diff
@@ -11,7 +11,9 @@ confinement: strict
 parts:
   python2:
     plugin: python2
+    source: .
```
kyrofa: Hmm... should this be required, now?

kyrofa: Well no, I guess not since we have `python-packages`, but could we toss an obvious error if `requirements` is used without `source`?

sergiusens: @kyrofa I plan to solve than when I deal with the culprit for https://bugs.launchpad.net/snapcraft/+bug/1606894

kyrofa: Ah, very good.

PR #657, opened 2016-07-14
Add constraints to python2 plugin

SamYaple:
LP:#1603110

There is currently no support for constraints[1] files with the python2 plugin. Some projects make heavy use of constraints. OpenStack is one such project

[1] https://pip.pypa.io/en/stable/user_guide/#constraints-files

PR #708, opened 2016-08-03
Add support for use of pip constraints

javacruft:
Pip supports use of constraints to limit/fix the versions of
Python modules that should be installed; this is used extensively
within the OpenStack community to limit transient dependency
issues causing functional issues.

Add optional 'upper-constraints' support for Python 2 and 3
plugins.

### DIFF
--- a/snapcraft/plugins/python2.py
+++ b/snapcraft/plugins/python2.py
@@ -35,6 +35,10 @@ Additionally, this plugin uses the following plugin-specific keywords:
     - python-packages:
       (list)
       A list of dependencies to get from PyPi
+    - constraints:
+      (string)
+      path or url to a constraints file to constrain installed module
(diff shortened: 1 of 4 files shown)


snappy-m-o:

snappy-m-o:

javacruft:
Actually this might be nicer consumed as 'constraints' rather than upper-constraints which is OpenStack specific naming of a file - I'll update.

sergiusens:
Hi, thanks for your contibution but this PR is racy :-) In the sense that one already is in progress #657 
May I ask you to work with @SamYaple instead?

javacruft:
@sergiusens I should have looked first; I'll close this Pull as @SamYaple work looks pretty much done to me and does everything mine did!

PR #709, opened 2016-08-04
Improve the go plugins usability

sergiusens:
Allow full use of source keywords combined with go-packages
which has the side effect of building the select main packages
from those sources and falls back to fetching the dependencies
remotely through go get's default behavior.

LP: #1599328


snapcraft/plugins/go.py
```diff
@@ -122,30 +115,25 @@ def clean_pull(self):
...
-            if os.path.islink(local_path):
-                os.unlink(local_path)
-            os.symlink(self.sourcedir, local_path)
-        self._run(['go', 'get', '-t', '-d', './{}/...'.format(go_package)])
-
-    def _remote_pull(self):
-        for go_package in self.options.go_packages:
-            self._run(['go', 'get', '-t', '-d', go_package])
+            go_package = os.path.basename(os.path.abspath(self.options.source))
+            go_package = go_package.rstrip('.git')
```
kyrofa: Do we have a test case for the .git?

snapcraft/plugins/go.py
```diff
@@ -89,31 +92,21 @@ def __init__(self, name, options, project):
...
+            go_package = self._get_local_go_package()
+            go_package_path = os.path.join(self._gopath_src, go_package)
+            if os.path.islink(go_package_path):
+                os.unlink(go_package_path)
+            os.makedirs(os.path.dirname(go_package_path), exist_ok=True)
+            os.symlink(self.sourcedir, go_package_path)
+            self._run(['go', 'get', '-t', '-d', './{}/...'.format(go_package)])
+
+        for go_package in self.options.go_packages:
+            self._run(['go', 'get', '-t', '-d', go_package])
```
kyrofa: And `go get` won't hit the internet if that package already exists?

snapcraft/plugins/go.py
```diff
@@ -163,17 +151,25 @@ def clean_build(self):
...
+    def _build_environment(self):
+        env = os.environ.copy()
+        env['GOPATH'] = self._gopath
 
-    def _run(self, cmd, **kwargs):
-        cmd = ['env', 'GOPATH={}'.format(self._gopath)] + cmd
-        return self.run(cmd, cwd=self._gopath_src, **kwargs)
+        # Add $GOPATH/bin so godeps is actually callable.
+        env['PATH'] = '{}:{}'.format(
+            os.path.join(self._gopath, 'bin'), env.get('PATH', ''))
```
kyrofa: I'm not sure `$PATH` is necessary here. Godeps probably won't be there :wink: .

elopio:
What about adding build tags? https://bugs.launchpad.net/snapcraft/+bug/1609623
Future branch?

snapcraft/plugins/go.py
```diff
@@ -163,17 +151,25 @@ def clean_build(self):
...
+
+        include_paths = []
+        for root in [self.installdir, self.project.stage_dir]:
+            include_paths.extend(
+                common.get_library_paths(root, self.project.arch_triplet))
+
+        flags = common.combine_paths(include_paths, '-L', ' ')
+        env['CGO_LDFLAGS'] = '{} {} {}'.format(
+            env.get('CGO_LDFLAGS', ''), flags, env.get('LDFLAGS', ''))
```
kyrofa: Wow, this is just... amazingly slick code :+1: .

snapcraft/plugins/go.py
```diff
@@ -89,31 +92,21 @@ def __init__(self, name, options, project):
...
-                '-L{0}/usr/lib',
-                '-L{0}/lib/{1}',
-                '-L{0}/usr/lib/{1}',
-                '$LDFLAGS']).format(root, self.project.arch_triplet) + '"',
-        ]
-        return env
-
     def pull(self):
-        # use -d to only download (build will happen later)
-        # use -t to also get the test-deps
```
elopio: I find that this comment was useful.

snapcraft/tests/test_plugin_go.py
```diff
@@ -129,21 +113,19 @@ class Options:
...
 
         self.run_mock.assert_has_calls([
-            mock.call(['env', 'GOPATH={}'.format(plugin._gopath),
-                       'go', 'get', '-t', '-d', './dir/...'],
-                      cwd=plugin._gopath_src),
-            mock.call(['env', 'GOPATH={}'.format(plugin._gopath),
-                       'go', 'install', './dir/...'],
-                      cwd=plugin._gopath_src),
+            mock.call(['go', 'get', '-t', '-d', './dir/...'],
+                      cwd=plugin._gopath_src, env=mock.ANY),
```
kyrofa: Since you already have a test for this (`test_pull_local_sources()`) you might consider resetting the mock before you run build() and just asserting the build calls.

snapcraft/tests/test_plugin_go.py
```diff
@@ -161,12 +143,10 @@ class Options:
...
 
         self.run_mock.assert_has_calls([
-            mock.call(['env', 'GOPATH={}'.format(plugin._gopath),
-                       'go', 'get', '-t', '-d', plugin.options.go_packages[0]],
-                      cwd=plugin._gopath_src),
-            mock.call(['env', 'GOPATH={}'.format(plugin._gopath),
-                       'go', 'install', plugin.options.go_packages[0]],
-                      cwd=plugin._gopath_src),
+            mock.call(['go', 'get', '-t', '-d', plugin.options.go_packages[0]],
+                      cwd=plugin._gopath_src, env=mock.ANY),
```
kyrofa: Ditto.

snapcraft/plugins/go.py
elopio: \o/ I was going crazy with that .git in the directory.

snapcraft/plugins/go.py
```diff
@@ -122,30 +115,25 @@ def clean_pull(self):
...
         super().build()
-        if self.options.source is not None:
-            self._local_build()
-        self._remote_build()
+
+        for go_package in self.options.go_packages:
+            self._run(['go', 'install', go_package])
+        if not self.options.go_packages:
+            self._run(['go', 'install',
+                       './{}/...'.format(self._get_local_go_package())])
```
elopio: I saw one project that needed 'go install ./{}', no ...
Could we handle that somehow?
https://github.com/ubuntu/snappy-playpen/pull/173/files#diff-2f36489c45dda5ec3ff6f5e0f8358e5dR38

elopio:
It's clearer, that's for sure. So this will help us solving the rest of the go troubles.
Before my +1, can you please add an integration test for the mulitple main thing?

snapcraft/plugins/go.py
kyrofa: It actually NEEDS it? i.e. it doesn't work without the `...`? Why would that be?

snapcraft/plugins/go.py
sergiusens: ;-)

sergiusens:
@elopio new feature ;-)

snapcraft/plugins/go.py
sergiusens: @kyrofa correct as we are not using `-u` ... `go get -h` is your friend ;-)

```
The -u flag instructs get to use the network to update the named packages
and their dependencies.  By default, get uses the network to check out
missing packages but does not use it to look for updates to existing packages.
```

snapcraft/plugins/go.py
sergiusens: truly true

snapcraft/plugins/go.py
sergiusens: I removed this

snapcraft/plugins/go.py
sergiusens: ok, made it better even

integration_tests/snaps/multimple-main-go-packages/main1/main.go
```diff
@@ -0,0 +1,7 @@
+package main
```
kyrofa: "multimple?" /me runs to see if you made the same typo in the integration test so I can point it out.

kyrofa: Hey wait, where's the test?

snapcraft/plugins/go.py
elopio: Because there is a windows vendored dependency inside that is missing the buildtag. It tries to compile it and fails.

sergiusens: El 04/08/16 a las 19:46, Leo Arias escribió:
> In snapcraft/plugins/go.py
> <https://github.com/snapcore/snapcraft/pull/709#discussion_r73616018>:
>
> >  
> >      def build(self):
> >          super().build()
> > -        if self.options.source is not None:
> > -            self._local_build()
> > -        self._remote_build()
> > +
> > +        for go_package in self.options.go_packages:
> > +            self._run(['go', 'install', go_package])
> > +        if not self.options.go_packages:
> > +            self._run(['go', 'install',
> > +                       './{}/...'.format(self._get_local_go_package())])
>
> Because there is a windows vendored dependency inside that is missing
> the buildtag. It tries to compile it and fails.

Yeah, but build tags would solve this as it would having a:
os_linux.go
os_windows.go
os_macos.go

in the same package :-)

sergiusens:
BTW, I can build lxd now which is what drove most of these changes

snapcraft/plugins/go.py
elopio: But wouldn't there be some spurious weird empty spaces if the env.gets return ''?
`' '.join(filter(None, [env.get('CGO_LDFLAGS'), flags, env.get('LDFLAGS')]))`

Of course, I don't really care. If you prefer yours, I'm ok.

elopio:
:+1:

PR #711, opened 2016-08-04
Add make-install-var field to the make plugin.

blakerouse:
Each project is different on the install flag for the Makefile. Allow changing the make-install-var when building with snapcraft.

### DIFF
--- a/snapcraft/plugins/make.py
+++ b/snapcraft/plugins/make.py
@@ -32,6 +32,10 @@ Additionally, this plugin uses the following plugin-specific keyword:
     - make-parameters:
       (list of strings)
       Pass the given parameters to the make command.
+
+    - make-install-var:
+      (string; default: DESTDIR)
+      Use the variable as the install target.
 """
 
 import snapcraft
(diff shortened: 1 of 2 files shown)


snappy-m-o:

snappy-m-o:

sergiusens:
looks good, ok to test

sergiusens:
thanks btw!

snapcraft/tests/test_plugin_make.py
```diff
@@ -31,6 +31,7 @@ def setUp(self):
         class Options:
             makefile = None
             make_parameters = []
+            make_install_var = "DESTDIR"
```
kyrofa: Single quotes here, please.

snapcraft/plugins/make.py
```diff
@@ -32,6 +32,10 @@
     - make-parameters:
       (list of strings)
       Pass the given parameters to the make command.
+
+    - make-install-var:
+      (string; default: DESTDIR)
+      Use the variable as the install target.
```
kyrofa: This is a little confusing-- "install" **is** the target. Perhaps something like "Use this variable to redirect the installation into the snap" or something like that.

blakerouse:
Fixed the requested changes.

PR #713, opened 2016-08-08
rust plugin: mock downloads in unit tests

sergiusens:
LP: #1610996


elopio:
:+1:

PR #715, opened 2016-08-08
Add artifacts option to make plugin.

jhobbs:
Some makefiles don't have an install target. This option allows specifying a list of artifacts to copy from the source directory to the install directory after running 'make'.

If the list of artifacts is not empty, the 'make install' step will be skipped and the artifacts will be copied by the plugin.

### DIFF
--- a/snapcraft/plugins/make.py
+++ b/snapcraft/plugins/make.py
@@ -28,7 +28,7 @@ Additionally, this plugin uses the following plugin-specific keyword:
     - artifacts:
       (list)
       Link/copy the given files from the make output to the snap
-      intstallation directory. If specified, the 'make install'
+      installation directory. If specified, the 'make install'
       step will be skipped.
 
     - makefile:
--- a/snapcraft/plugins/make.py
+++ b/snapcraft/plugins/make.py
@@ -25,6 +25,12 @@ For more information check the 'plugins' topic for the former and the
 
 Additionally, this plugin uses the following plugin-specific keyword:
 
+    - artifacts:
+      (list)
+      Link/copy the given files from the make output to the snap
+      intstallation directory. If specified, the 'make install'
+      step will be skipped.
+
     - makefile:
       (string)
       Use the given file as the makefile.
@@ -38,6 +44,7 @@ Additionally, this plugin uses the following plugin-specific keyword:
       Use this variable to redirect the installation into the snap.
 """
 
+import os
 import snapcraft
 import snapcraft.common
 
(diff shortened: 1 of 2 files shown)


snappy-m-o:

snappy-m-o:

snapcraft/plugins/make.py
```diff
@@ -87,5 +96,10 @@ def build(self):
             command.extend(self.options.make_parameters)
 
         self.run(command + ['-j{}'.format(self.project.parallel_build_count)])
-        self.run(command + [
-            'install', self.options.make_install_var + '=' + self.installdir])
+        if len(self.options.artifacts):
+            for artifact in self.options.artifacts:
+                self.run(['cp', artifact, self.installdir])
```
sergiusens: we have a link_or_copy function in `common`. Mind using that instead?

snapcraft/plugins/make.py
```diff
@@ -87,5 +96,10 @@ def build(self):
             command.extend(self.options.make_parameters)
 
         self.run(command + ['-j{}'.format(self.project.parallel_build_count)])
-        self.run(command + [
-            'install', self.options.make_install_var + '=' + self.installdir])
+        if len(self.options.artifacts):
```
sergiusens: This would be more pythonic

```python
if self.options.artifacts:
```

snapcraft/tests/test_plugin_make.py
```diff
@@ -141,3 +142,19 @@ def test_build_install_var(self, run_mock):
...
+    def test_build_artifacts(self, run_mock):
+        self.options.artifacts = ['artifact1', 'artifact2']
+        plugin = make.MakePlugin('test-part', self.options,
+                                 self.project_options)
+        os.makedirs(plugin.sourcedir)
+
+        plugin.build()
+
+        self.assertEqual(3, run_mock.call_count)
+        run_mock.assert_has_calls([
```
sergiusens: once `link_or_copy` is used we would be able to assert that the file exists instead

sergiusens:
Thanks for this.

I am guessing that using `artifacts` invalidates the whole `make install` logic. It would be nice to have the initial docstring for the module explain this option and explain it invalidates the install target.

snapcraft/plugins/make.py
jhobbs: Ok, no problem.

snapcraft/plugins/make.py
jhobbs: Agreed, will change.

snapcraft/tests/test_plugin_make.py
jhobbs: Well, the file doesn't exist. Everything is being mocked out, which makes sense - the goal of this unit test isn't to test link_or_copy. I'll change the unit test to make sure link_or_copy is being called.

jhobbs:
@sergiusens,

Thanks for the review. I believe I've addressed all your comments - please have another look.

snapcraft/plugins/make.py
```diff
@@ -25,6 +25,12 @@
 
 Additionally, this plugin uses the following plugin-specific keyword:
 
+    - artifacts:
+      (list)
+      Link/copy the given files from the make output to the snap
+      intstallation directory. If specified, the 'make install'
```
blakerouse: s/intstallation/installation

jhobbs: Thanks Blake, fixed.

sergiusens:
OK to test

PR #719, opened 2016-08-10
support dest-subdir on dump plugin

ycheng:
add "dest-subdir" option for dump plugin.

it's useful for target like tomcat, that we might want then to install to /opt/tomcat directory.


snappy-m-o:

snappy-m-o:

sergiusens:
Thanks for your contribution but if you really need destdir you should use the `organize` core primitive.

Please reopen if this is not the case. Improvements to how `organize` is handled are welcome btw.

PR #720, opened 2016-08-10
Start the fake parts server only in the tests that need it

elopio:
Load the remote parts in the config lazily.

LP: #1611828

### DIFF
--- a/snapcraft/internal/yaml.py
+++ b/snapcraft/internal/yaml.py
@@ -109,10 +109,6 @@ class PluginNotDefinedError(Exception):
 
 class Config:
 
-    @property
-    def part_names(self):
-        return self._part_names
-
     def __init__(self, project_options=None):
         if project_options is None:
             project_options = snapcraft.ProjectOptions()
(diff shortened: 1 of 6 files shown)


sergiusens:
retest this please

snapcraft/internal/yaml.py
```diff
@@ -132,10 +128,18 @@ def __init__(self, project_options=None):
...
 
+    @property
+    def part_names(self):
+        return self._part_names
+
+    @property
+    def _remote_parts(self):
+        if getattr(self, '_remote_parts_attr', None) is None:
+            self._remote_parts_attr = parts.get_remote_parts()
+        return self._remote_parts_attr
```
kyrofa: Should this be `self.__remote_parts_attr` (note the double underscore)?

elopio: well, not really, I think. The double underscore does mangling, but there's always a way to access the attribute from outside. And from inside it works the same as one underscore. So I don't like it too much. I prefer just to rely on our own style in which one underscore means internal, and we try to respect it as much as possible.

kyrofa:
Oh gosh yes :+1: . I'm merging this because I can't stand running the tests again without it.

PR #721, opened 2016-08-11
Set owner/group for directories in stage and prime

josepht:
shutil.copystat doesn't preserve owner/group so do it manually with
os.stat() and os.chown()

LP:#1605903

### DIFF
--- a/snapcraft/internal/pluginhandler.py
+++ b/snapcraft/internal/pluginhandler.py
@@ -662,6 +662,7 @@ def _migrate_files(snap_files, snap_dirs, srcdir, dstdir, missing_ok=False,
         gid = dir_stat.st_gid
         os.makedirs(dst, exist_ok=True)
         os.chown(dst, uid, gid, follow_symlinks=follow_symlinks)
+        shutil.copystat(src, dst, follow_symlinks=follow_symlinks)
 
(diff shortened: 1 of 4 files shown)


kyrofa:
Hmm... the history looks odd here. Does it need a rebase?

snapcraft/internal/pluginhandler.py
```diff
@@ -656,15 +656,22 @@ def _migrate_files(snap_files, snap_dirs, srcdir, dstdir, missing_ok=False,
     for directory in snap_dirs:
         src = os.path.join(srcdir, directory)
         dst = os.path.join(dstdir, directory)
+
+        dir_stat = os.stat(src)
+        uid = dir_stat.st_uid
+        gid = dir_stat.st_gid
         os.makedirs(dst, exist_ok=True)
-        shutil.copystat(src, dst, follow_symlinks=follow_symlinks)
+        os.chown(dst, uid, gid, follow_symlinks=follow_symlinks)
```
kyrofa: So now instead of copying the mode we're copying the owner info. Which is fine. But is there any chance we want to do both?

josepht: Excellent point. I think we should.

PR #723, opened 2016-08-11
Set owner/group for directories in stage and prime

josepht:
shutil.copystat doesn't preserve owner/group so do it manually with
os.stat() and os.chown()

LP:#1605903

### DIFF
--- a/snapcraft/internal/pluginhandler.py
+++ b/snapcraft/internal/pluginhandler.py
@@ -657,7 +657,7 @@ def _migrate_files(snap_files, snap_dirs, srcdir, dstdir, missing_ok=False,
         src = os.path.join(srcdir, directory)
         dst = os.path.join(dstdir, directory)
 
-        dir_stat = os.stat(src)
+        dir_stat = os.stat(src, follow_symlinks=follow_symlinks)
         uid = dir_stat.st_uid
         gid = dir_stat.st_gid
         os.makedirs(dst, exist_ok=True)
@@ -668,12 +668,15 @@ def _migrate_files(snap_files, snap_dirs, srcdir, dstdir, missing_ok=False,
         src = os.path.join(srcdir, snap_file)
         dst = os.path.join(dstdir, snap_file)
 
-        dir_stat = os.stat(srcdir)
+        dir_stat = os.stat(os.path.dirname(src),
+                           follow_symlinks=follow_symlinks)
         uid = dir_stat.st_uid
         gid = dir_stat.st_gid
-        os.makedirs(dstdir, exist_ok=True)
-        shutil.copystat(srcdir, dstdir, follow_symlinks=follow_symlinks)
-        os.chown(dstdir, uid, gid, follow_symlinks=follow_symlinks)
+        os.makedirs(os.path.dirname(dst), exist_ok=True)
+        shutil.copystat(os.path.dirname(src), os.path.dirname(dst),
+                        follow_symlinks=follow_symlinks)
+        os.chown(os.path.dirname(dst), uid, gid,
+                 follow_symlinks=follow_symlinks)
 
         if missing_ok and not os.path.exists(src):
             continue
--- a/snapcraft/internal/pluginhandler.py
+++ b/snapcraft/internal/pluginhandler.py
@@ -662,6 +662,7 @@ def _migrate_files(snap_files, snap_dirs, srcdir, dstdir, missing_ok=False,
         gid = dir_stat.st_gid
         os.makedirs(dst, exist_ok=True)
         os.chown(dst, uid, gid, follow_symlinks=follow_symlinks)
+        shutil.copystat(src, dst, follow_symlinks=follow_symlinks)
 
     for snap_file in snap_files:
         src = os.path.join(srcdir, snap_file)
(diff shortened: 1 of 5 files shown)


snapcraft/internal/pluginhandler.py
```diff
@@ -656,15 +656,24 @@ def _migrate_files(snap_files, snap_dirs, srcdir, dstdir, missing_ok=False,
...
-        os.makedirs(os.path.dirname(dst), exist_ok=True)
-        shutil.copystat(os.path.dirname(src), os.path.dirname(dst),
-                        follow_symlinks=follow_symlinks)
+
+        dir_stat = os.stat(srcdir)
+        uid = dir_stat.st_uid
+        gid = dir_stat.st_gid
+        os.makedirs(dstdir, exist_ok=True)
+        shutil.copystat(srcdir, dstdir, follow_symlinks=follow_symlinks)
+        os.chown(dstdir, uid, gid, follow_symlinks=follow_symlinks)
```
kyrofa: This section is operating upon `srcdir` and `dstdir`, when perhaps it should be operating upon `src` and `dst`.

josepht: src and dst are files which we're not copying here.  Though if snap_file
has directory path components I'd need to use os.path.dirname(src) I guess.

josepht: I've updated the branch to operate on the proper directories.

snapcraft/internal/pluginhandler.py
```diff
@@ -656,15 +656,27 @@ def _migrate_files(snap_files, snap_dirs, srcdir, dstdir, missing_ok=False,
     for directory in snap_dirs:
         src = os.path.join(srcdir, directory)
         dst = os.path.join(dstdir, directory)
+
+        dir_stat = os.stat(src, follow_symlinks=follow_symlinks)
+        uid = dir_stat.st_uid
+        gid = dir_stat.st_gid
         os.makedirs(dst, exist_ok=True)
+        os.chown(dst, uid, gid, follow_symlinks=follow_symlinks)
         shutil.copystat(src, dst, follow_symlinks=follow_symlinks)
```
kyrofa: Looking good. Though lines 660-665 look ripe for extraction into a function that can also be used on 671-679, no?

josepht: Done.

kyrofa:
:+1: from me, thanks @josepht! Nice tests, too. @elopio can you please take a look?

josepht:
I think this is failing due to insufficient permissions.  Which might be problematic in general since only root can make a file owned by a different user.  I guess we can warn on failure of os.chown() but keep going maybe?

kyrofa:
Ah, yeah that seems like a good compromise.

snapcraft/tests/test_pluginhandler.py
```diff
@@ -287,8 +288,72 @@ def test_migrate_files_preserves_ownership(self, copystat_mock):
...
-        copystat_mock.assert_called_with('install', 'stage',
-                                         follow_symlinks=True)
+        self.assertTrue(chown_mock.called)
+
+    @patch('os.chown')
+    def test_migrate_files_chown_permissions(self, chown_mock):
+        os.makedirs('install')
+        os.makedirs('stage')
+
+        chown_mock.side_effect = PermissionError("No no no")
```
kyrofa: Good test, thank you.

snapcraft/internal/pluginhandler.py
```diff
@@ -650,21 +650,34 @@ def _migratable_filesets(fileset, srcdir):
...
 
+def _create_dirs(srcdir, dstdir, follow_symlinks=False):
+    dir_stat = os.stat(srcdir, follow_symlinks=follow_symlinks)
+    uid = dir_stat.st_uid
+    gid = dir_stat.st_gid
+    os.makedirs(dstdir, exist_ok=True)
+    try:
+        os.chown(dstdir, uid, gid, follow_symlinks=follow_symlinks)
+    except PermissionError as e:
+        logger.warning(e)
```
kyrofa: Will this actually be informative? Or does it only say "permission denied"? If it's the latter, it might be worth adding the directory name to the warning (e.g. "unable to chown {dstdir}: {error}").

snapcraft/internal/pluginhandler.py
```diff
@@ -650,21 +650,34 @@ def _migratable_filesets(fileset, srcdir):
     return snap_files, snap_dirs
 
 
+def _create_dirs(srcdir, dstdir, follow_symlinks=False):
+    dir_stat = os.stat(srcdir, follow_symlinks=follow_symlinks)
+    uid = dir_stat.st_uid
+    gid = dir_stat.st_gid
+    os.makedirs(dstdir, exist_ok=True)
+    try:
+        os.chown(dstdir, uid, gid, follow_symlinks=follow_symlinks)
```
kyrofa: GitHub won't let me comment quite where I'd like to, but I just realized: if the `link_or_copy` call on line 696 ends up copying instead of linking, it will also need a `chown`. Right?

elopio:
@josepht looks good, but I don't understand what happened here. I ran the manual test to check for ownership, and build the os snap too. Did I miss something when trying this?

josepht:
@elopio I'm not sure how it worked but I can't think of anything you might've missed.

elopio:
I ran the manual test in 2.14 again and it fails. Clearly my mistake when running it before the release, sorry about that.

I ran it again from this branch, and when I run prime I get the warning. When I run sudo prime, I get the ownership preserved: http://pastebin.com/g5iBkrRF

@josepht can you please update the manual test to run sudo prime ?

PR #724, opened 2016-08-11
In travis, install the static checkers with pip.

elopio:
Remove them from debian tests because they are not needed anymore.

LP: #1610263


elopio:
retest this please

PR #726, opened 2016-08-12
Implement basic form of `snapcraft register-key`

cjwatson:
This adds a basic register-key command that can be used to register keys
from your snap keyring.

We don't yet have all the prescribed error modes in place, and nor are
we able to tell which keys are already enabled.  That requires some more
work on the store side.


snappy-m-o:

snappy-m-o:

cjwatson:
Note that https://code.launchpad.net/~cjwatson/snap-assertions-service/public-key-headers-endpoint/+merge/302612 and https://code.launchpad.net/~cjwatson/software-center-agent/less-terrible-account-key/+merge/302789 need to be deployed before we can go ahead with this, but it should be possible to review this in the meantime to speed things up.

sergiusens:
Seems travis needs some deps installed 
```
  Running setup.py install for pygpgme ... error
    Complete output from command /usr/bin/python3 -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-c3o2kata/pygpgme/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-iwyk7pry-record/install-record.txt --single-version-externally-managed --compile:
    running install
    running build
    running build_py
    creating build
    creating build/lib.linux-x86_64-3.5
    creating build/lib.linux-x86_64-3.5/gpgme
    copying gpgme/editutil.py -> build/lib.linux-x86_64-3.5/gpgme
    copying gpgme/__init__.py -> build/lib.linux-x86_64-3.5/gpgme
    running build_ext
    building 'gpgme._gpgme' extension
    creating build/temp.linux-x86_64-3.5
    creating build/temp.linux-x86_64-3.5/src
    x86_64-linux-gnu-gcc -pthread -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -g -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -fPIC -I/usr/include/python3.5m -c src/gpgme.c -o build/temp.linux-x86_64-3.5/src/gpgme.o
    In file included from src/gpgme.c:22:0:
    src/pygpgme.h:24:19: fatal error: gpgme.h: No such file or directory
    compilation terminated.
    error: command 'x86_64-linux-gnu-gcc' failed with exit status 1
```

cjwatson:
Thanks.  I'll fix that in a bit - the store interface is moving quite quickly here so I want to get that settled down first.

sergiusens:
El 16/08/16 a las 13:10, Colin Watson escribió:
> Thanks. I'll fix that in a bit - the store interface is moving quite
> quickly here so I want to get that settled down first.

Indeed, not a stranger to that story :-)

snapcraft/_store.py
```diff
@@ -73,6 +75,84 @@ def logout():
     logger.info('Credentials cleared.')
 
 
+def _get_usable_secret_keys(query=None):
+    ctx = gpgme.Context()
```
elopio: we usually try to avoid variable names like this. Specially for somebody like me who doesn't speak great English, ctx is less readable than context.

snapcraft/_store.py
```diff
@@ -73,6 +75,84 @@ def logout():
...
+    for key in ctx.keylist(query, True):
+        if not key.subkeys or not key.uids or key.disabled or not key.can_sign:
```
elopio: could you group these four conditions in a function with a nice name explaining what this check is about?
like if not _is_usable_key(key), and then in _is_usable_key a comment about what makes a key usable.

snapcraft/_store.py
```diff
@@ -73,6 +75,84 @@ def logout():
...
+            continue
+        mainkey = key.subkeys[0]
+        if (mainkey.expired or mainkey.revoked or mainkey.disabled or
+                mainkey.invalid):
```
elopio: Here too, it would be nice to group the statement in a function. _is_mainkey_usable? I'm not sure about the name.

snapcraft/_store.py
```diff
@@ -73,6 +75,84 @@ def logout():
...
+            continue
+        # We're currently only interested in RSA >= 4096-bit.
+        if mainkey.pubkey_algo != gpgme.PK_RSA or mainkey.length < 4096:
```
elopio: I would move this to a function too, but I tend to write way too many functions :p I won't complain about this one, just saying...

snapcraft/_store.py
```diff
@@ -73,6 +75,84 @@ def logout():
...
+            continue
+        yield key
+
+
+def _display_fingerprint(fingerprint):
```
elopio: Maybe display is not the right word here, because it's not displaying anything. What about _get_formatted_fingerprint or _format_fingerprint ?

snapcraft/_store.py
```diff
@@ -73,6 +75,84 @@ def logout():
...
+    assert len(fingerprint) == 40
```
elopio: I don't like assertions. I think it's better to make an if and raise the exception explicitly. The "assert" statement can be disabled and the "if" can't, but that's not likely to happen. My main problem with assertions is that they are intended to discover errors from the programmer, things that should never happen. And for me, that's what unit tests are for. So if you already have a test that makes sure this condition can't happen, the statement is not useful. And if this can happen during the normal operation of the program, that's what exceptions are for, and they are clearer.

snapcraft/_store.py
```diff
@@ -73,6 +75,84 @@ def logout():
...
+    chunks = [fingerprint[i:i + 4] for i in range(0, 40, 4)]
+    return '{}  {}'.format(' '.join(chunks[:5]), ' '.join(chunks[5:]))
+
+
+_display_key_algorithm = {
+    gpgme.PK_RSA: 'R',
+}
+
+
+def _display_key(key):
```
elopio: Here again, display is not the right verb. get or format seem better.

snapcraft/_store.py
```diff
@@ -73,6 +75,84 @@ def logout():
...
+
+
+def register_key(query):
+    keys = list(_get_usable_secret_keys(query=query))
+    if not keys:
+        raise RuntimeError(
+            'You have no usable GPG secret keys matching "{}".'.format(query))
+    key = _select_key(keys)
+    fingerprint = key.subkeys[0].fpr
+    ctx = gpgme.Context()
```
elopio: s/ctx/context
On the line above I was wondering what fpr was, then I saw you assigned it to "fingerprint", that's lovely :D

snapcraft/_store.py
```diff
@@ -73,6 +75,84 @@ def logout():
...
+    key_data = io.BytesIO()
+    ctx.export(fingerprint.encode('ascii'), key_data)
+    logger.info('Registering GPG key ...')
+    store = storeapi.StoreClient()
+    try:
+        store.register_key(key_data.getvalue())
+    except storeapi.errors.InvalidCredentialsError:
```
elopio: @sergiusens this feels like a decorator, right? We are duplicating it in many places, we could just tag the functions that require authentication instead.

snapcraft/_store.py
```diff
@@ -73,6 +75,84 @@ def logout():
...
+        logger.error('No valid credentials found.'
+                     ' Have you run "snapcraft login"?')
+        raise
+    logger.info(
+        'Done. The GPG key {} will be expected for signing your '
+        'assertions.'.format(_display_fingerprint(fingerprint)))
```
elopio: "expected" sounds confusing to me. What do you think about: "The GPG key ... will be used to sign your assertions"? I'm not sure it's better though.

elopio:
This is great, thanks for sharing this @cjwatson !

How hard would it be to write an integration test for the happy path? We are using pyexpect, which makes it easy to handle the input. But I'm wondering about creating the gpg key to be selected. We would have to patch the ~/.gnupg path to use a temp file and maybe we can insert a fake key in there.

snapcraft/_store.py
sergiusens: El 16/08/16 a las 23:01, Leo Arias escribió:
> In snapcraft/_store.py
> <https://github.com/snapcore/snapcraft/pull/726#discussion_r75052594>:
> 
>> +
>> +def register_key(query):
>> +    keys = list(_get_usable_secret_keys(query=query))
>> +    if not keys:
>> +        raise RuntimeError(
>> +            'You have no usable GPG secret keys matching "{}".'.format(query))
>> +    key = _select_key(keys)
>> +    fingerprint = key.subkeys[0].fpr
>> +    ctx = gpgme.Context()
>> +    key_data = io.BytesIO()
>> +    ctx.export(fingerprint.encode('ascii'), key_data)
>> +    logger.info('Registering GPG key ...')
>> +    store = storeapi.StoreClient()
>> +    try:
>> +        store.register_key(key_data.getvalue())
>> +    except storeapi.errors.InvalidCredentialsError:
> 
> @sergiusens <https://github.com/sergiusens> this feels like a decorator,
> right? We are duplicating it in many places, we could just tag the
> functions that require authentication instead.

Maybe, let's look into that after this lands.

cjwatson:
I've refactored this to use the new snap key management interfaces (https://docs.google.com/document/d/1Bkn9sLgXqkGsgV1eBWO935z5jZJiJQCL24muqw2zD7U).  I know that Travis is currently failing; I need to get it to install snapd, but I need to wait for a new enough version to land in xenial-updates first.

At least a basic integration test should be feasible, and I'm working on that now.

sergiusens:
https://bugs.launchpad.net/snapcraft/+bug/1612730

cjwatson:
retest this please

josepht:
Colin said 'retest this please'

sergiusens:
ok to test

integration_tests/test_store_register_key.py
```diff
@@ -0,0 +1,46 @@
...
+        temp_keys_dir = os.path.join(self.path, '.snap', 'gnupg')
+        shutil.copytree(keys_dir, temp_keys_dir)
+        self.useFixture(fixtures.EnvironmentVariable(
+            'SNAP_GNUPG_HOME', temp_keys_dir))
+
+    def test_successful_key_registration(self):
+        if os.getenv('TEST_STORE', 'fake') != 'fake':
+            self.skipTest(
+                'Cannot register test keys against staging/production until '
+                'we have a way to delete them again.')
```
elopio: is there a bug to track this?

integration_tests/__init__.py
```diff
@@ -132,6 +132,38 @@ def register(self, snap_name, wait=True):
...
+    def register_key(self, key_name, email=None, password=None,
+                     expect_success=True):
+        email = email or self.test_store.user_email
+        password = password or self.test_store.user_password
+
+        process = pexpect.spawn(
+            self.snapcraft_command, ['register-key', key_name])
+
+        process.expect_exact(
+            'Enter your Ubuntu One SSO credentials.\r\n'
```
evandandrea: "Please enter your Ubuntu One credentials."

elopio:
When I run register-key without a key, I get:

./bin/snapcraft register-key
'NoneType' object is not iterable

snapcraft/storeapi/errors.py
```diff
@@ -75,6 +75,38 @@ def __init__(self, message):
         super().__init__(message=message)
 
 
+class StoreAccountInformationError(StoreError):
+
+    fmt = 'Error fetching account information from store: {error}'
+
+    def __init__(self, response):
+        error = '%d %s' % (response.status_code, response.reason)
```
elopio: can you please use format instead?

snapcraft/storeapi/errors.py
```diff
@@ -75,6 +75,38 @@ def __init__(self, message):
...
+            pass
+        super().__init__(error=error)
+
+
+class StoreKeyRegistrationError(StoreError):
+
+    fmt = 'Key registration failed: {error}'
+
+    def __init__(self, response):
+        error = '%d %s' % (response.status_code, response.reason)
```
elopio: please use format.

integration_tests/test_store_register_key.py
cjwatson: I've filed https://bugs.launchpad.net/bugs/1621441 and added a reference here.

cjwatson:
The register-key `'NoneType' object is not iterable` error is really a snapd bug, and I've sent a fix for that as https://github.com/snapcore/snapd/pull/1869.  I've also made a change on this side to tolerate the peculiar output.

PR #728, opened 2016-08-12
Use link_or_copy instead of just hard-links for deb cache.

kyrofa:
This PR fixes LP: [#1612813](https://bugs.launchpad.net/snapcraft/+bug/1612813), making the cache work across filesystem boundaries.


elopio:
makes sense. :+1:

PR #729, opened 2016-08-13
Dump plugin: Don't remove install directory.

kyrofa:
Removing the install directory prevents the usage of stage-packages. It was done because `shutil.copytree` simply can't deal with merging two directories, but the Dump plugin needs to be able to deal with that situation.

This PR fixes LP: [#1611827](https://bugs.launchpad.net/snapcraft/+bug/1611827) by taking a lightly modified version of the  `_linktree()` logic from the Copy plugin and putting it into the Dump plugin. This has the added benefit of not copying broken symlinks into the snap as well.


snapcraft/plugins/dump.py
```diff
@@ -35,7 +35,58 @@ class DumpPlugin(snapcraft.BasePlugin):
...
     def build(self):
         super().build()
-        if os.path.exists(self.installdir):
-            os.rmdir(self.installdir)
-        shutil.copytree(self.builddir, self.installdir,
-                        copy_function=snapcraft.common.link_or_copy)
+        _linktree(self.builddir, self.installdir, self.installdir)
+
+
+def _link_or_copy(source, destination, boundary):
```
elopio: a nit from the clean code book. Define the functions in this order:

_link_tree
_link_or_copy
_create_similar_directory

That way you can read it from top to bottom, with less vertical jumps.

snapcraft/plugins/dump.py
```diff
@@ -35,7 +35,58 @@ class DumpPlugin(snapcraft.BasePlugin):
...
+    except FileNotFoundError:
+        raise FileNotFoundError('{!r} is a broken symlink pointing outside the snap'.format(source))
+
+
+def _create_similar_directory(source, destination):
+    os.makedirs(destination, exist_ok=True)
+    shutil.copystat(source, destination, follow_symlinks=False)
+
+
+def _linktree(source_tree, destination_tree, boundary):
```
elopio: _link_tree.

elopio: on the first read, the meaning of the boundary argument is not clear. A comment would be nice.

elopio:
lgtm, with a few pita comments. :+1: 
Thanks for adding the regression test.

snapcraft/plugins/dump.py
kyrofa: Yeah good point. Though I did it in the order of calls, which flipped the last two.

snapcraft/plugins/dump.py
kyrofa: Both are done.

sergiusens:
Looks good! And history repeats it seems ;-)

I do have one comment though, it would be nice to put `_linktree()` in `common` and name it `link_or_copy_tree(...)` as we use it in many other places already.

Maybe it's also time to *NOT* use common for this but a new non internal `file_utils` module.

kyrofa:
> Maybe it's also time to NOT use common for this but a new non internal file_utils module.

Yeah, I'll do that. I'll make it take a `copy_function` param like `shutil.copytree` so we can continue doing the boundary checking without needing it in the generic version.

kyrofa:
> Maybe it's also time to NOT use common for this but a new non internal file_utils module.

Alright, done. Note that I moved all common file operations into `file_utils.py`, but continued importing them in `common.py` so as to not break API.

elopio:
static error: snapcraft/tests/test_common.py:18:1: F401 're' imported but unuseds

PR #731, opened 2016-08-15
Factor parts config into snapcraft/internal/parts.py

josepht:
LP:#1596044


sergiusens:
Hi there! A few questions:

- Is there a step 2 to this?
- how about renaming `snapcraft.internal.yaml` to something like `snapcraft.internal.project_loader`? What do you think @kyrofa?

sergiusens:
snapcraft/internal/parts.py:26:1: F401 'snapcraft' imported but unused

josepht:
@sergiusens yes there'll be a part 2 that does the logic refactoring.  I'm happy to change the name of s.i.yaml to s.i.project_loader if @kyrofa agrees.

snapcraft/internal/yaml.py
```diff
@@ -109,15 +98,20 @@ def __init__(self, part):
 
 class Config:
 
+    @property
+    def part_names(self):
+        return self.parts_config._part_names
```
kyrofa: Hmm... you're exporting something that's hidden in a subclass. That seems a little odd.

kyrofa:
> how about renaming snapcraft.internal.yaml to something like snapcraft.internal.project_loader? What do you think @kyrofa?

Yeah, I agree, good name!

snapcraft/internal/yaml.py
josepht: Great catch.  I've updated it to export from PartConfig and access it via that in Config.

josepht:
I've re-named s.i.yaml to s.i.project_loader and fixed the part_name thing in Config()

snapcraft/internal/parts.py
```diff
@@ -129,6 +131,207 @@ def compose(self, part_name, properties):
...
+
+            if 'filesets' in properties:
+                del properties['filesets']
+
+            self.load_plugin(part_name, plugin_name, properties)
+
+        self._compute_part_dependencies()
+        self.all_parts = self._sort_parts()
+
+    def _compute_part_dependencies(self):
```
elopio: you could call this _compute_dependencies.

snapcraft/internal/parts.py
```diff
@@ -129,6 +131,207 @@ def compose(self, part_name, properties):
...
+                    break
+            if not top_part:
+                raise SnapcraftLogicError(
+                    'circular dependency chain found in parts definition')
+            sorted_parts = [top_part] + sorted_parts
+            self.all_parts.remove(top_part)
+
+        return sorted_parts
+
+    def part_prereqs(self, part_name):
```
elopio: I would call this get_prereqs.

snapcraft/internal/parts.py
```diff
@@ -129,6 +131,207 @@ def compose(self, part_name, properties):
...
+        """Returns a set with all of part_names' prerequisites."""
+        return set(self.after_requests.get(part_name, []))
+
+    def part_dependents(self, part_name):
```
elopio: I would call this get_dependents.

snapcraft/internal/parts.py
```diff
@@ -129,6 +131,207 @@ def compose(self, part_name, properties):
...
+        return dependents
+
+    def get_part(self, part_name):
+        for part in self.all_parts:
+            if part.name == part_name:
+                return part
+
+        return None
+
+    def validate_parts(self, part_names):
```
elopio: This can be called only validate.

snapcraft/internal/project_loader.py
```diff
@@ -128,188 +122,33 @@ def __init__(self, project_options=None):
...
-            properties['snap'] = _expand_filesets_for('snap', properties)
-
-            if 'filesets' in properties:
-                del properties['filesets']
-
-            self.load_plugin(part_name, plugin_name, properties)
-
-        self._compute_part_dependencies()
-        self.all_parts = self._sort_parts()
+        self.parts_config = parts.PartsConfig(self.data.get('parts', {}),
```
elopio: would it be weird to call this one only parts? So you call it like config.parts instead of config.parts_config ?

snapcraft/internal/lifecycle.py
```diff
@@ -320,7 +320,8 @@ def _clean_part_and_all_dependents(part_name, step, config, staged_state,
         dependent_part.clean(staged_state, primed_state, step)
 
     # Finally, clean the part in question
-    config.get_part(part_name).clean(staged_state, primed_state, step)
+    config.parts_config.get_part(part_name).clean(staged_state, primed_state,
+                                                  step)
```
elopio: This long chain of calls is a little worrying [1]. 
Is it right for the part to have the responsibility to clean itself?
If so, we can at least hide it making a clean_part method in parts_config.
Also this made me notice that the _Executor doesn't have a lot of use for config, only for parts_config. Why don't we store parts_config as an attribute?

[1] http://c2.com/cgi/wiki?TrainWreck

elopio:
I like this refactor. I've left some questions and suggestions about naming, because I feel that we are using the word "part" too much here. Feel free to disagree :)

sergiusens:
I +1 all of @elopio's name change suggestions

josepht:
@elopio @sergiusens I've updated the names according to Leo's suggestions.

PR #732, opened 2016-08-16
Remove store dispute logic

sergiusens:
The `-dispute` URL when registering is no longer valid as the register
and dispute URLs have been merged into one.

LP: #1613609


elopio:
+1, after the static error is fixed.
Why didn't we get notice about this change?

sergiusens:
@elopio I do not know :-/

PR #733, opened 2016-08-17
Add the go-buildtags property to the go plugin

elopio:
LP: #1609623


elopio:
retest this please

snapcraft/plugins/go.py
```diff
@@ -39,6 +39,10 @@
       This entry tells the checked out `source` to live within a certain path
       within `GOPATH`.
       This is not needed and does not affect `go-packages`.
+
+    - go-buildtags:
+      (string)
```
sergiusens: I am almost certain I feel this should be a list/array even if we eventually do a `' '.join(...)`

kyrofa: Agreed. That would make it consistent with the other plugins with similar options, as well.

kyrofa: Should we support `go-installflags` instead? So people can use `-asmflags` or `-ldflags` in addition to `-tags`?

elopio: so it would be like:
```
  go-installflags:
    - -asmflags
    - -ldflags
    - -tags=tag1,tag2
```
?

kyrofa: Okay. In that case, should we make sure we stay true to the command-line options? i.e. call this `go-tags`?

elopio: So:

go-buildtags: [tag1, tag2]

And that will turn into -tags=tag1,tag2. lgtm, I'll make the change.

sergiusens: El 17/08/16 a las 11:55, Kyle Fazzari escribió:
> In snapcraft/plugins/go.py
> <https://github.com/snapcore/snapcraft/pull/733#discussion_r75139056>:
> 
>> @@ -39,6 +39,10 @@
>>        This entry tells the checked out `source` to live within a certain path
>>        within `GOPATH`.
>>        This is not needed and does not affect `go-packages`.
>> +
>> +    - go-buildtags:
>> +      (string)
> 
> Okay. In that case, should we make sure we stay true to the command-line
> options? i.e. call this |go-tags|?

go-tags feels confusing to me, but if both of you prefer it fine by me ;-)

kyrofa: > go-tags feels confusing to me, but if both of you prefer it fine by me ;-)

Not necessarily, but if we ever intend on supporting other cli options for `go install`, naming some of them corresponding to the param in question and others different might make for a confusing experience.

sergiusens: El 17/08/16 a las 12:53, Kyle Fazzari escribió:
> In snapcraft/plugins/go.py
> <https://github.com/snapcore/snapcraft/pull/733#discussion_r75150618>:
> 
>> @@ -39,6 +39,10 @@
>>        This entry tells the checked out `source` to live within a certain path
>>        within `GOPATH`.
>>        This is not needed and does not affect `go-packages`.
>> +
>> +    - go-buildtags:
>> +      (string)
> 
>     go-tags feels confusing to me, but if both of you prefer it fine by
>     me ;-)
> 
> Not necessarily, but if we ever intend on supporting other cli options
> for |go install|, naming some of them corresponding to the param in
> question and others different might make for a confusing experience.

Right, but maybe ldflags and cflags end up being a core thing and not go
specific.

kyrofa: Yeah, I don't have strong feelings here, +1 either way.

PR #736, opened 2016-08-17
Disable internet access during unit tests.

elopio:


elopio:
I'm not sure if this will work, it's just a try.

.travis.yml
```diff
@@ -21,6 +21,8 @@ install:
   - sudo apt-get -qq update
   - sudo apt-get install -y python3-coverage
 script:
+  # Disable internet access during unit tests.
+  - if [ "$TEST_SUITE" = unit ]; then echo 0 | sudo tee /proc/sys/net/ipv4/ip_forward; fi
```
sergiusens: This needs to happen after installing the apt deps or we are not going to have those deps in :-)

Seems it can do the trick though

elopio: I got that. What I don't know is how to do some work, keep the container, and then do more work on the same one. It seems cleaner to disable the internet access from inside the container just when the tests start, but I don't know how to do that.
I'm inclined to wait for Federico to come back from holidays, he probably has a better idea.

sergiusens: El 19/08/16 a las 17:20, Leo Arias escribió:
> In .travis.yml
> <https://github.com/snapcore/snapcraft/pull/736#discussion_r75545648>:
> 
>> @@ -21,6 +21,8 @@ install:
>>    - sudo apt-get -qq update
>>    - sudo apt-get install -y python3-coverage
>>  script:
>> +  # Disable internet access during unit tests.
>> +  - if [ "$TEST_SUITE" = unit ]; then echo 0 | sudo tee /proc/sys/net/ipv4/ip_forward; fi
> 
> I got that. What I don't know is how to do some work, keep the
> container, and then do more work on the same one. It seems cleaner to
> disable the internet access from inside the container just when the
> tests start, but I don't know how to do that.
> I'm inclined to wait for Federico to come back from holidays, he
> probably has a better idea.

When you debuild locally there is something done there that causes the
same effect, we should probably go down that path instead.

sergiusens:
This will need to be reopened once it is ready

PR #738, opened 2016-08-18
nodjs, gulp plugins: use http_proxy to connect npm registry

m-shibata:
npm regigstry server doesn't support to connect via HTTPS proxy.
However if https_proxy env is set, npm command access to registry via
HTTPS proxy.  Avoid to this, if https_proxy env is set, then
nodejs/gulp plugin will change https-proxy property by "npm config".

LP: #1611372

### DIFF
--- a/snapcraft/plugins/gulp.py
+++ b/snapcraft/plugins/gulp.py
@@ -99,6 +99,14 @@ class GulpPlugin(snapcraft.BasePlugin):
         env = os.environ.copy()
         env['PATH'] = '{}:{}'.format(
             os.path.join(self._npm_dir, 'bin'), env['PATH'])
+
+        # make npm use http proxy to access registry via https
+        # ref. https://github.com/npm/npm/issues/2050
+        if 'http_proxy' in os.environ:
+            logger.info('Set http_proxy environment variable to "https-proxy"')
+            self.run(['npm', 'config', '-g', 'set', 'https-proxy',
+                     os.environ['http_proxy']], env=env)
+
         self.run(['npm', 'install', '-g', 'gulp-cli'], env=env)
         if os.path.exists(os.path.join(self.builddir, 'package.json')):
             self.run(['npm', 'install', '--only-development'], env=env)
--- a/snapcraft/plugins/gulp.py
+++ b/snapcraft/plugins/gulp.py
@@ -100,9 +100,11 @@ class GulpPlugin(snapcraft.BasePlugin):
         env['PATH'] = '{}:{}'.format(
             os.path.join(self._npm_dir, 'bin'), env['PATH'])
 
-        # make npm use http proxy to access registry via https
+        # Setting the "npm_config_https_proxy" environment variable with the
+        # value of "http_proxy".
         # ref. https://github.com/npm/npm/issues/2050
-        if 'http_proxy' in os.environ:
+        if ('http_proxy' in os.environ and
+                'npm_config_https_proxy' not in os.environ):
             logger.info('Set http_proxy environment variable to "https-proxy"')
             env['npm_config_https_proxy'] = os.environ['http_proxy']
 
--- a/snapcraft/plugins/gulp.py
+++ b/snapcraft/plugins/gulp.py
@@ -104,8 +104,7 @@ class GulpPlugin(snapcraft.BasePlugin):
         # ref. https://github.com/npm/npm/issues/2050
         if 'http_proxy' in os.environ:
             logger.info('Set http_proxy environment variable to "https-proxy"')
-            self.run(['npm', 'config', '-g', 'set', 'https-proxy',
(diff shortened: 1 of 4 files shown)


snappy-m-o:

snappy-m-o:

sergiusens:
Thanks for your contribution, I have a couple of questions which are going to be inline in the code.

snapcraft/plugins/gulp.py
```diff
@@ -99,6 +99,14 @@ def build(self):
         env = os.environ.copy()
         env['PATH'] = '{}:{}'.format(
             os.path.join(self._npm_dir, 'bin'), env['PATH'])
+
+        # make npm use http proxy to access registry via https
+        # ref. https://github.com/npm/npm/issues/2050
+        if 'http_proxy' in os.environ:
```
sergiusens: why don't you instead check for `https_proxy`? Or check for both and set both. What about `no_proxy`?

snapcraft/plugins/gulp.py
```diff
@@ -99,6 +99,14 @@ def build(self):
...
+            logger.info('Set http_proxy environment variable to "https-proxy"')
+            self.run(['npm', 'config', '-g', 'set', 'https-proxy',
```
sergiusens: instead of setting a global config why not setup the environment using node variables like explained here:
https://docs.npmjs.com/misc/config

sergiusens:
I also did a bit more research and it seems npm honors http_proxy, https_proxy and no_proxy. Can you confirm that this is not the case?

snapcraft/plugins/gulp.py
m-shibata: Only needs "`http_proxy`".

npm honors `http_proxy`, `https_proxy` and `no_proxy` environment variables. Normally if these variables are set, not need to use "`npm config`".  However if "`https_proxy`" is "https://proxy/", npm connect to registry server via HTTPS protocol and occur connection failed.

To workaround this, make npm use HTTP protocol by setting `https-proxy` to "http://proxy".

* `http_proxy`: use to access "http://registry.npmjs.org/"
* `https_proxy`: not use to access npm servers
* `no_proxy`: npm itself honors this environment variable

snapcraft/plugins/gulp.py
m-shibata: Oh, I didn't know that, sorry.

It seems that "`npm_config_https-proxy=...`" env is more useful and safely. I  will change to use it.

m-shibata:
@sergiusens

Thanks for your review, and pushed modified version, please recheck it.

> I also did a bit more research and it seems npm honors http_proxy, https_proxy and no_proxy. Can you confirm that this is not the case?

Yes, npm honors `https_proxy`. And just it cause the problem.

* Use case 1: A user no use http/https proxies.

`http_proxy`/`https_proxy` are empty. Doesn't care about proxy.

* Use case 2: A user only use http scheme for proxy server.

```
http_proxy="http://proxy.domain/"
https_proxy="http://proxy.domain/"
```

npm and other tools (like apt, wget) will work fine.

* Use case 3: A user use both http/https scheme for proxy servers.

```
http_proxy="http://proxy.domain/"
https_proxy="https://proxy.domain/"
```

It seems that npm parse scheme of `https_proxy` and connect to registry server
via proxy sever with HTTPS protocol (sorry for details are unknown).
Result of this, `SSL23_GET_SERVER_HELLO:unknown protocol:s23_clnt.c`
error will occured.

`apt` and `wget` have no error with the set of environment variables.

snapcraft/plugins/gulp.py
```diff
@@ -99,6 +99,13 @@ def build(self):
         env = os.environ.copy()
         env['PATH'] = '{}:{}'.format(
             os.path.join(self._npm_dir, 'bin'), env['PATH'])
+
+        # make npm use http proxy to access registry via https
+        # ref. https://github.com/npm/npm/issues/2050
+        if 'http_proxy' in os.environ:
+            logger.info('Set http_proxy environment variable to "https-proxy"')
```
sergiusens: Make this
Setting the "npm_config_https_proxy" environment variable with the value of "https_proxy"

snapcraft/plugins/gulp.py
sergiusens: maybe make the check

```python
if 'http_proxy' in os.environ and 'npm_config_https_proxy' not in os.environ:
```

to avoid an unwanted override

sergiusens:
looks good in general, just one minor nit

m-shibata:
@sergiusens 

sorry for very late reply. applied your pointed out, and rebased on upstream.

m-shibata:
Discussion on Launchpad, it was decided that the PR isn't merged.
https://bugs.launchpad.net/snapcraft/+bug/1611372

PR #739, opened 2016-08-18
Only show chown warnings with --debug.

josepht:
Use logger.debug() for chown warnings since they are expected to
fail for most use cases.

LP:#1614524

### DIFF
--- a/snapcraft/file_utils.py
+++ b/snapcraft/file_utils.py
@@ -68,7 +68,7 @@ def link_or_copy(source, destination, follow_symlinks=False):
         try:
             os.chown(destination, uid, gid, follow_symlinks=follow_symlinks)
         except PermissionError as e:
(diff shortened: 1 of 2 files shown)


snapcraft/file_utils.py
```diff
@@ -68,7 +68,7 @@ def link_or_copy(source, destination, follow_symlinks=False):
         try:
             os.chown(destination, uid, gid, follow_symlinks=follow_symlinks)
         except PermissionError as e:
-            logger.warning('unable to chown {destination}: {error}'.format(
+            logger.debug('unable to chown {destination}: {error}'.format(
```
sergiusens: Why is *unable* not capitalized here?

josepht: Fixed.

PR #742, opened 2016-08-18
Handle missing source type packages in the parser

josepht:
Improve error messages if a source type package (eg. bzr, git, etc)
is not installed and suggest the appropriate packages.

LP:#1611541


snapcraft/internal/sources.py
```diff
@@ -389,7 +389,13 @@ def get(sourcedir, builddir, options):
     handler_class = _get_source_handler(source_type, options.source)
     handler = handler_class(options.source, sourcedir, source_tag,
                             source_branch)
-    handler.pull()
+    try:
+        handler.pull()
+    except FileNotFoundError:
```
sergiusens: is this really a `FileNotFoundError`?

josepht: That's what I got when I tried it out.

josepht:
```
josepht@spruce:~$ python3
Python 3.5.2 (default, Jul  5 2016, 12:43:10) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import subprocess
>>> subprocess.call(['bzr', 'branch', 'lp:fake'])
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.5/subprocess.py", line 557, in call
    with Popen(*popenargs, **kwargs) as p:
  File "/usr/lib/python3.5/subprocess.py", line 947, in __init__
    restore_signals, start_new_session)
  File "/usr/lib/python3.5/subprocess.py", line 1551, in _execute_child
    raise child_exception_type(errno_num, err_msg)
FileNotFoundError: [Errno 2] No such file or directory: 'bzr'
>>> 
```

josepht: Though, this might mask errors for zip, tar, and local source types.

snapcraft/internal/sources.py
```diff
@@ -389,7 +389,15 @@ def get(sourcedir, builddir, options):
...
                             source_branch)
-    handler.pull()
+    e_msg = 'A required package is missing, please install these packages: {}'
+    try:
+        handler.pull()
+    except FileNotFoundError:
+        if type(handler) in [Bazaar, Git, Subversion, Mercurial]:
+            required_packages = get_required_packages(options)
+            logging.error(e_msg.format(required_packages))
+        else:
```
sergiusens: You might still need to raise an exception right? or else you break the logic of pull when those packages are not installed. You could raise a more specific exception in this case with that logging.error message instead of the logging.error call done here.

sergiusens: When I say logic for pull, I mean the behavior expected when executed in the context of the plugin lifecycle.

josepht: You're right.  I'll work on an update to fix that.

snapcraft/internal/parser.py
```diff
@@ -275,6 +275,22 @@ def _process_wiki_entry(entry, master_parts_list):
         master_parts_list.update(parts_list)
 
 
+def _handle_wiki_entry(entry, master_parts_list):
+    wiki_errors = 0
+    try:
+        _process_wiki_entry(entry, master_parts_list)
+    except InvalidEntryError as e:
+        logger.warning('Invalid wiki entry: {}'.format(e))
```
sergiusens: logger.warning('Invalid wiki entry: {!r}'.format(e))

snapcraft/internal/parser.py
```diff
@@ -275,6 +275,22 @@ def _process_wiki_entry(entry, master_parts_list):
...
+        wiki_errors = 1
+    except sources.MissingPackageError as e:
+        logger.warning(
+            'Missing package for source type: {}'.format(e))
```
sergiusens: `'Missing package for source type: {!r}'.format(e))`

snapcraft/internal/parser.py
```diff
@@ -275,6 +275,22 @@ def _process_wiki_entry(entry, master_parts_list):
...
+        # Not really a wiki error
```
sergiusens: Maybe tie this comment to a bug?

snapcraft/internal/sources.py
```diff
@@ -86,6 +86,10 @@ def __init__(self, message):
         self.message = message
 
 
+class MissingPackageError(Exception):
```
sergiusens: can you consolidate this in the new `snapcraft.errors` and inherit from the SnapcraftError exception in there?

snapcraft/internal/sources.py
```diff
@@ -389,7 +393,15 @@ def get(sourcedir, builddir, options):
     handler_class = _get_source_handler(source_type, options.source)
     handler = handler_class(options.source, sourcedir, source_tag,
                             source_branch)
-    handler.pull()
+    e_msg = 'A required package is missing, please install these packages: {}'
```
sergiusens: can you put this message in the `MissingPackageError` exception?

sergiusens:
ok, I am good!

snapcraft/internal/errors.py
```diff
@@ -60,3 +60,14 @@ def __init__(self, *, part_name, other_part_name, conflict_files):
         super().__init__(part_name=part_name,
                          other_part_name=other_part_name,
                          file_paths='\n'.join(sorted(spaced_conflict_files)))
+
+
+class MissingPackageError(SnapcraftError):
+
+    fmt = (
+        'A required package is missing, please install these packages:'
```
elopio: This is a little wrong, because it says:
"A requried package" -> singular
"install these packages:" -> plural

Maybe, we can say: "One or more packages are missing, please install:"

snapcraft/internal/parser.py
```diff
@@ -233,6 +233,22 @@ def _process_wiki_entry(entry, master_parts_list):
...
+    except InvalidEntryError as e:
+        logger.warning('Invalid wiki entry: {!r}'.format(e))
+        wiki_errors = 1
+    except sources.MissingPackageError as e:
+        logger.warning(e)
+        # TODO: support multiple types of errors, see
+        # https://bugs.launchpad.net/snapcraft/+bug/1616613
+        wiki_errors = 1
+
+    return wiki_errors
```
elopio: returning 1 or 0 to indicate error sounds like the job for an exception. You could just reraise it, and keep the try below:

try:
    _handle_wiki_entry
except SnapcraftError:
   wiki_errors += 1

snapcraft/internal/parser.py
```diff
@@ -233,6 +233,22 @@ def _process_wiki_entry(entry, master_parts_list):
         master_parts_list.update(parts_list)
 
 
+def _handle_wiki_entry(entry, master_parts_list):
```
elopio: you have handle_wiki_entry and process_wiki_entry. Just with the name it will be impossible to understand what each one does. I can't think of better names, so it would be good to put docstrings.

snapcraft/internal/sources.py
elopio: yes, this doesn't sound too good. To better identify it we would have to raise an exception of our own on the pull. Like which bzr || raise exception.
So at least make an XXX comment here with a link to the bug. Even better to fix it at once :)

elopio: This doesn't look good. To do it right, the check sould come from the source class.
Like which bzr || raise exception.

snapcraft/internal/parser.py
josepht: docstrings added.

snapcraft/internal/parser.py
josepht: Hrm, I'll take a look at this.  I'll bet I can do away with _handle_wiki_entry() using your suggestion.

snapcraft/internal/parser.py
sergiusens: ah @elopio will say docstrings use three double quotes `"""` :wink:

snapcraft/internal/parser.py
sergiusens: @josepht are you going to take an action about this?

josepht: Yes

josepht:
@elopio I've reworked the error reporting a bit as you suggested.

elopio:
:+1: Thanks Joe.

snapcraft/internal/errors.py
```diff
@@ -69,3 +69,25 @@ def __init__(self, *, part_name, other_part_name, conflict_files):
...
+
+    def __init__(self, required_packages):
+        super().__init__(required_packages=required_packages)
+
+
+class InvalidEntryError(SnapcraftError):
+
+    fmt = (
+        'Invalid wiki entry:'
+        ' {error!r}'
```
sergiusens: out of curiosity why isn't this all on the same line?

josepht: No reason other than "copy-paste-edit" without thinking :)

snapcraft/internal/errors.py
```diff
@@ -69,3 +69,22 @@ def __init__(self, *, part_name, other_part_name, conflict_files):
...
+    fmt = (
+        'One or more required packages are missing, please install:'
+        ' {required_packages!r}'
+    )
+
+    def __init__(self, required_packages):
+        super().__init__(required_packages=required_packages)
+
+
+class InvalidEntryError(SnapcraftError):
```
sergiusens: Last one. This should be `InvalidWikiEntryError`

josepht: Done.

PR #744, opened 2016-08-19
testing: only run the ros demo on xenial

sergiusens:
It is currently failing on yakkety due to gpg key errors with
the cache it is using.

LP: #1614476


josepht:
+1

elopio:
:+1:

PR #745, opened 2016-08-19
Actually return a non-zero error code on errors.

josepht:
LP:#1602394

### DIFF
--- a/snapcraft/internal/parser.py
+++ b/snapcraft/internal/parser.py
@@ -358,5 +358,5 @@ def _write_parts_list(path, master_parts_list):
                  default_flow_style=False))
 
 
-if __name__ == '__main__':  # pragma: no cover
-    sys.exit(main())        # pragma: no cover
+if __name__ == '__main__':        # pragma: no cover
+    sys.exit(1 if main() else 0)  # pragma: no cover
--- a/snapcraft/internal/parser.py
+++ b/snapcraft/internal/parser.py
@@ -34,6 +34,7 @@ import logging
 import os
 import pkg_resources
 import re
+import sys
 import urllib
 import yaml
 from yaml.scanner import ScannerError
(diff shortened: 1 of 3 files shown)


snapcraft/internal/parser.py
```diff
@@ -358,4 +359,4 @@ def _write_parts_list(path, master_parts_list):
 
 
 if __name__ == '__main__':  # pragma: no cover
-    main()                  # pragma: no cover
+    sys.exit(main())        # pragma: no cover
```
sergiusens: does dis mean the error code is 1 for one wiki error, 2 for two and so on? Is this what we want? Or do we want

```python
wiki_errrors = main()
if wiki_errors:
    sys.exit(1)
```

@elopio ?

josepht: Good catch.  You're right we want 1 or 0, maybe another in the future for things like network issues getting the index.

snapcraft/internal/parser.py
```diff
@@ -357,5 +358,5 @@ def _write_parts_list(path, master_parts_list):
                  default_flow_style=False))
 
 
-if __name__ == '__main__':  # pragma: no cover
-    main()                  # pragma: no cover
+if __name__ == '__main__':        # pragma: no cover
+    sys.exit(1 if main() else 0)  # pragma: no cover
```
sergiusens: I think I liked my version better. This feels harder to read.

josepht: Fixed.

elopio:
This wfm, but you say on the bug that you want to exit with 2 on origin errors.
If that's still the case, maybe you can open a different bug for that.

josepht:
I'll do that.  I don't think it's a very high priority.

elopio:
⟫ /home/elopio/workspace/canonical/snapcraft/bin/snapcraft-parser -i /tmp/test
Invalid wiki entry: Missing key in wiki entry: 'project-part'
1 wiki errors found!
⟫ echo $?
1

:shipit:

sergiusens:
+1 waiting for adt to finish to merge

PR #748, opened 2016-08-22
kernel plugin: vmlinuz -> kernel.img hard link

piso77:
Adhere to the new naming scheme decided in Heidelberg, and create a kernel.img copy of vmlinuz.


sergiusens:
Can you update the unit tests for this case?

snapcraft/plugins/kernel.py
```diff
@@ -311,6 +311,7 @@ def _copy_vmlinuz(self):
         os.link(src, dst)
         # why oh why?
         os.link(src, os.path.join(self.installdir, 'vmlinuz'))
```
sergiusens: do we still need this line?

kyrofa: And should they be link_or_copies?

sergiusens: that should come later if needed. This plugin is full of `os.link` calls ;-)

piso77:
Ok, i removed the vmlinuz hard link, the comment and i've pushed again - once this get in, we might rewrite to use symlinks instead of hard links.

PR #749, opened 2016-08-22
Allow registering private packages

sergiusens:
LP: #1602793


elopio:
:+1:

PR #750, opened 2016-08-22
storeapi: remove dependency on the 'success' attr

sergiusens:
LP: #1615786

### DIFF
--- a/snapcraft/tests/fake_servers.py
+++ b/snapcraft/tests/fake_servers.py
@@ -235,13 +235,18 @@ class FakeStoreUploadRequestHandler(BaseHTTPRequestHandler):
 
     def _handle_upload_request(self):
         logger.info('Handling upload request')
-        self.send_response(200)
-        self.send_header('Content-Type', 'application/octet-stream')
+        if 'UPDOWN_BROKEN' in os.environ:
+            response_code = 500
+            content_type = 'text/plain'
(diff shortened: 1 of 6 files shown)


sergiusens:
The work done here fixes LP: #1611116 as well

sergiusens:
retest this please

snapcraft/tests/fake_servers.py
```diff
@@ -235,14 +235,18 @@ def do_POST(self):
 
     def _handle_upload_request(self):
         logger.info('Handling upload request')
-        self.send_response(200)
-        self.send_header('Content-Type', 'application/octet-stream')
+        if 'UPDOWN_BROKEN' in os.environ:
```
kyrofa: Why do this via environment variables instead of setting something on the fixture?

kyrofa:
This looks pretty good, one question there.

snapcraft/tests/fake_servers.py
sergiusens: yeah, bug for @elopio to be able to change the behavior of the fixture

kyrofa: Done, LP: [#1616117](https://bugs.launchpad.net/snapcraft/+bug/1616117).

kyrofa:
:+1: from me.

PR #752, opened 2016-08-23
Ant properties, build target, and destination directory options

evandandrea:
Add options for setting ant properties, an ant build target, and the destination directory for ant build artifacts (point the right property at the snapcraft installdir). This change preserves compatibility with the old method of assuming jars will end up in target/.

The motivation was the Cassandra snap. It currently uses a [custom plugin](https://github.com/evandandrea/cassandra-snap/blob/e7f88807c9c0887f3df3dc5a7ca15b196b0cd604/parts/plugins/x-cassandra.py), which can be replaced with the following two settings to the ant plugin:
```
ant-dest-property: dist.dir
ant-build-target: artifacts
```

I'm completely open to changing the name of these options or altering their behaviour.

### DIFF
--- a/integration_tests/snaps/simple-ant/snapcraft.yaml
+++ b/integration_tests/snaps/simple-ant/snapcraft.yaml
@@ -1,0 +1,17 @@
+name: simple-ant
+version: 0
+summary: test ant builds
+description: |
+ Test ant builds and options.
+confinement: strict
+
+parts:
+  a:
+    plugin: ant
+    source: my-app
+    ant-properties:
+      foo: 'bar'
+    ant-dest-property: 'dist.dir'
+    ant-build-target: 'artifacts'
+    snap:
+      - foo.jar
--- a/integration_tests/snaps/simple-ant/snapcraft.yaml
+++ b/integration_tests/snaps/simple-ant/snapcraft.yaml
@@ -12,6 +12,7 @@ parts:
     ant-properties:
       foo: 'bar'
     ant-dest-property: 'dist.dir'
-    ant-build-target: 'artifacts'
+    ant-build-targets:
+      - artifacts
     snap:
       - foo.jar
--- a/integration_tests/snaps/simple-ant/snapcraft.yaml
+++ b/integration_tests/snaps/simple-ant/snapcraft.yaml
@@ -10,9 +10,6 @@ parts:
     plugin: ant
     source: my-app
     ant-properties:
-      foo: 'bar'
-    ant-dest-property: 'dist.dir'
+      dist.dir: 'target'
     ant-build-targets:
       - artifacts
-    snap:
-      - foo.jar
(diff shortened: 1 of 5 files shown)


snappy-m-o:

snappy-m-o:

snapcraft/plugins/ant.py
```diff
@@ -38,21 +53,52 @@
...
+    def schema(cls):
+        schema = super().schema()
+        schema['properties']['ant-properties'] = {
+            'type': 'object',
+            'default': {},
+        }
+        schema['properties']['ant-build-target'] = {
+            'type': 'string',
+        }
+        schema['properties']['ant-dest-property'] = {
```
sergiusens: is this need only because we don't support using something like `$SNAPCRAFT_PART_INSTALLDIR` in the yaml?

evandandrea: Correct

sergiusens: The others are prefect, just wondering if `ant-dest-property` was just to get non exposed information from the plugin.

In any case, https://github.com/snapcore/snapcraft/pull/757 (variable naming pending a discussion)

snapcraft/plugins/ant.py
```diff
@@ -38,21 +53,52 @@
...
 
+    @classmethod
+    def schema(cls):
+        schema = super().schema()
+        schema['properties']['ant-properties'] = {
+            'type': 'object',
+            'default': {},
+        }
+        schema['properties']['ant-build-target'] = {
+            'type': 'string',
```
sergiusens: is it possible to have more than one target?

sergiusens: or should it be (either by multiple calls from a list or in a oneliner through a ' '.join(...))

snapcraft/plugins/ant.py
```diff
@@ -22,6 +22,21 @@
...
+    - properties:
+      (object)
+      A dictionary of key-value pairs. Set the following properties when
+      running ant.
+
+    - targets:
+      (list of strings)
+      Run the given ant targets.
+
+    - dest-property:
```
sergiusens: I guess this now needs to be removed as well.

sergiusens:
ok to test

integration_tests/snaps/simple-ant/snapcraft.yaml
```diff
@@ -0,0 +1,15 @@
+name: simple-ant
+version: 0
+summary: test ant builds
+description: |
+ Test ant builds and options.
+confinement: strict
```
sergiusens: mind adding `grade: ...` here?

sergiusens:
ok, last but not least, can you update the branch with master and add a unit test for getting the plugin's schema class method? You can boilerplate from the other plugin tests to get this done quicker.

sergiusens:
```
snapcraft/tests/test_plugin_ant.py:53:13: E128 continuation line under-indented for visual indent

snapcraft/tests/test_plugin_ant.py:54:13: E128 continuation line under-indented for visual indent

snapcraft/tests/test_plugin_ant.py:57:13: E128 continuation line under-indented for visual indent

snapcraft/tests/test_plugin_ant.py:58:13: E128 continuation line under-indented for visual indent
```

PR #754, opened 2016-08-24
Use in plugin code to remove unwanted files.

sergiusens:
Do not spread similar logic everywhere and consolidate
on methods provided by snapcraft.file_utils.

LP: #1616464


integration_tests/test_python_plugin.py
```diff
@@ -83,6 +83,16 @@ def test_build_rewrites_shebangs(self):
...
 
+    def test_build_does_not_keep_pyc_files_in_install(self):
+        project_dir = 'python-entry-point'
+        self.run_snapcraft('stage', project_dir)
+
+        pyc_files = []
+        for _, _, files in os.walk(os.path.join(project_dir, 'stage')):
+            pyc_files.extend([f for f in files if f.endswith('pyc')])
+
+        self.assertEqual(pyc_files, [])
```
elopio: :+1:

snapcraft/plugins/python2.py
```diff
@@ -190,34 +191,24 @@ def build(self):
...
-                               re.compile(r'#!.*python'),
-                               r'#!/usr/bin/env python')
+        file_utils.replace_in_file(self.installdir, re.compile(r''),
+                                   re.compile(r'#!.*python'),
+                                   r'#!/usr/bin/env python')
 
     def snap_fileset(self):
         fileset = super().snap_fileset()
         fileset.append('-usr/bin/pip*')
-        fileset.append('-usr/lib/python*/dist-packages/easy-install.pth')
```
elopio: Why is there no need to remove pth anymore?

snapcraft/plugins/python3.py
```diff
@@ -166,10 +167,22 @@ def build(self):
              '--prefix={}/usr'.format(self.installdir),
              '--root={}'.format(self.installdir)], cwd=self.builddir)
 
+        def remove_func(dirpath, files):
+            print(dirpath, files)
```
elopio: remove the print.

snapcraft/plugins/python3.py
```diff
@@ -166,10 +167,22 @@ def build(self):
              '--prefix={}/usr'.format(self.installdir),
              '--root={}'.format(self.installdir)], cwd=self.builddir)
 
+        def remove_func(dirpath, files):
```
elopio: A comment explaining what you are removing would be nice.

snapcraft/plugins/python2.py
sergiusens: El 24/08/16 a las 15:08, Leo Arias escribió:
> In snapcraft/plugins/python2.py
> <https://github.com/snapcore/snapcraft/pull/754#discussion_r76108816>:
> 
>>  
>>      def snap_fileset(self):
>>          fileset = super().snap_fileset()
>>          fileset.append('-usr/bin/pip*')
>> -        fileset.append('-usr/lib/python*/dist-packages/easy-install.pth')
> 
> Why is there no need to remove pth anymore?

This is an oversight and we should actually remove all .pth files :-)

snapcraft/plugins/python3.py
```diff
@@ -166,10 +167,22 @@ def build(self):
...
+            print(dirpath, files)
+            endswith_pyc = (x.endswith('.pyc') for x in files)
+            return dirpath.endswith('__pycache__') and all(endswith_pyc)
```
elopio: It would be safer to use os.path.basename(dirpath) == '__pycache__'.
Not that this works only if dirpath doesn't end with a slash, but that's the same assumption in your code.

snapcraft/plugins/python3.py
```diff
@@ -166,10 +167,22 @@ def build(self):
...
+        file_utils.replace_in_file(self.installdir, re.compile(r''),
+                                   re.compile(r'#!.*python'),
+                                   r'#!/usr/bin/env python')
+
+    def snap_fileset(self):
+        fileset = super().snap_fileset()
+        fileset.append('-usr/bin/pip*')
+        return fileset
 
     @property
```
elopio: I think the properties go between the __init__ and the methods. But I'm not sure where did I see this guideline.

snapcraft/plugins/python3.py
```diff
@@ -80,6 +80,7 @@ def __init__(self, name, options, project):
             'python3-pkg-resources',
             'python3-setuptools',
         ])
+        self._pip_dir = os.path.join(self.partdir, 'pip')
```
elopio: This is not used.

integration_tests/test_python_plugin.py
```diff
@@ -83,6 +83,16 @@ def test_build_rewrites_shebangs(self):
         self.assertEqual('#!/usr/bin/env python2', python2_shebang)
         self.assertEqual('#!/usr/bin/env python3', python3_shebang)
 
+    def test_build_does_not_keep_pyc_files_in_install(self):
+        project_dir = 'python-entry-point'
```
elopio: This is a project with both py2 and py3 parts, right?
The name "python-entry-point" does not tell the purpose of the snap in this test. Maybe create a new one? If you rename it, then it will not be clear for the entry point tests.

sergiusens: @elopio I think I'll use `pip-requirements-list`, too much copy paste snaps will be bad in the long run.

snapcraft/plugins/python3.py
sergiusens: yes

snapcraft/plugins/python3.py
sergiusens: are you sure? :-)

snapcraft/plugins/python3.py
sergiusens: I can make another PR for that, I don't think it is mandatory though

snapcraft/file_utils.py
```diff
@@ -150,3 +150,18 @@ def _search_and_replace_contents(file_path, search_pattern, replacement):
             f.seek(0)
             f.truncate()
             f.write(replaced)
+
+
+def remove_files(directory, remove_pattern_func=lambda file_name: True):
+    """Remove files from directory if remove_pattern_func returns True."""
+    for root, _, files in os.walk(directory):
+        for file_name in files:
+            if remove_pattern_func(file_name):
```
kyrofa: Do you think there's a benefit to passing the entire path to the `remove_pattern_func()`?

snapcraft/plugins/python2.py
```diff
@@ -187,37 +188,31 @@ def build(self):
...
-        fileset.append('-usr/lib/python*/*.pyc')
-        fileset.append('-usr/lib/python*/*/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*/*/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*/*/*/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*/*/*/*/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*/*/*/*/*/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*/*/*/*/*/*/*.pyc')
```
kyrofa: Yessssssssssssssssss :100: . I hated that.

sergiusens:
oh mosquitto :-/ restest this please

snapcraft/file_utils.py
sergiusens: I could not think of any. Would there be?

snapcraft/plugins/python2.py
sergiusens: the xmas tree is gone :-)

snapcraft/file_utils.py
kyrofa: What if I wanted to remove `foo/bar.pyc` but _not_ `foo/baz/bar.pyc`? I know you're not using it that way, but it might make a function in `file_utils` more generically useful.

sergiusens:
restest this please

snapcraft/plugins/python3.py
```diff
@@ -181,6 +177,8 @@ def remove_func(dirpath, files):
     def snap_fileset(self):
         fileset = super().snap_fileset()
         fileset.append('-usr/bin/pip*')
+        # TODO figure out how this is created in the first place.
```
elopio: will you file a bug for this one?

elopio:
This still looks good to me, too.

sergiusens:
Closing in favor of #765

PR #758, opened 2016-08-25
Add an integration test for parts with filesets

jocave:
Use a part from the parts service which uses filesets

### DIFF
--- a/integration_tests/test_parts.py
+++ b/integration_tests/test_parts.py
@@ -61,3 +61,39 @@ class PartsTestCase(integration_tests.TestCase):
             "  source-type: tar\n")
 
         self.assertEqual(expected, output)
+
+
+class PartsWithFilesetsTestCase(integration_tests.TestCase):
+
+    def test_part_with_fileset(self):
+        self.run_snapcraft('update')
+
+        output = self.run_snapcraft(['define', 'simple-make-filesets'])
+
+        expected = (
+            "Maintainer: 'Jonathan Cave <jonathan.cave@canonical.com>'\n"
+            "Description: 'The filesets test from the integration test suite.'"
+            "\n\n"
+            "simple-make-filesets:\n"
+            "  filesets:\n"
+            "    files:\n"
+            "    - share/file1\n"
+            "    - share/file2\n"
+            "  organize:\n"
+            "    dir1: new/dir1\n"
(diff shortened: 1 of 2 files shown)


snappy-m-o:

snappy-m-o:

integration_tests/test_parts.py
```diff
@@ -61,3 +61,39 @@ def test_curl_exists_and_properly_defined(self):
...
+
+
+class PartsWithFilesetsTestCase(integration_tests.TestCase):
+
+    def test_part_with_fileset(self):
+        self.run_snapcraft('update')
+
+        output = self.run_snapcraft(['define', 'simple-make-filesets'])
+
+        expected = (
```
sergiusens: for easier reading might consider using `"""` here?
Like
```python
    expected = ("""Maintainer: ...
Description: ...
...
""")
```

jocave:
@sergiusens Correct - I didn't add the part there as from what I could see the integration tests were getting their parts from the real server. Is the idea to have the part in fake_servers as well even though this test wouldn't use it?

Sorry for not having a fix for the functionality that this test exercises. I had a look at the parts code today, but found it tricky to follow.

sergiusens:
El viernes, 26 de agosto de 2016 10h'01:47 ART, Jonathan Cave 
> @sergiusens Correct - I didn't add the part there as from what 
> I could see the integration tests were getting their parts from 
> the real server. Is the idea to have the part in fake_servers as 
> well even though this test wouldn't use it?

You are indeed correct, I was thinking about the parts parser service which 
is all faked.

> Sorry for not having a fix for the functionality that this test 
> exercises. I had a look at the parts code today, but found it 
> tricky to follow.

That's fine, the parts code and lifecycle are very complex in themselves. 
There is a much needed refactor needed there, Joe has worked on it for a 
bit but it is still some ways ahead.


-- 

integration_tests/test_parts.py
sergiusens: nevermind this comment, the rest of this file has this format

sergiusens:
OK, revisiting this, I think `define` is fine an what needs fixing is the use of the fileset during execution

sergiusens:
err, ignore me

sergiusens:
#799 has this test in.
thanks

PR #762, opened 2016-08-26
node plugin: run build in pull phase to download dependencies.

smoser:
This runs the build in the pull phase to download all needed dependencies.
There is no way to 'npm install --download-only'.

Additionally, we add the --cache-min=Infinity which is required.
Without it, the run in the build phase will try to reach the npm registry.

Lastly, instead of '-g', use the more self-documenting long form '--global'.

### DIFF
--- a/snapcraft/plugins/nodejs.py
+++ b/snapcraft/plugins/nodejs.py
@@ -96,6 +96,8 @@ class NodePlugin(snapcraft.BasePlugin):
         super().pull()
         os.makedirs(self._npm_dir, exist_ok=True)
         self._nodejs_tar.download()
+        # do the install in the pull phase to download all dependencies.
+        self._npm_install()
 
     def clean_pull(self):
         super().clean_pull()
@@ -106,12 +108,16 @@ class NodePlugin(snapcraft.BasePlugin):
 
     def build(self):
         super().build()
+        self._npm_install()
+
+    def _npm_install(self):
         self._nodejs_tar.provision(
             self.installdir, clean_target=False, keep_tarball=True)
+        npm_install = ['npm', '--cache-min=Infinity', 'install', '--global']
         for pkg in self.options.node_packages:
-            self.run(['npm', 'install', '-g', pkg])
+            self.run(npm_install + [pkg])
         if os.path.exists(os.path.join(self.builddir, 'package.json')):
-            self.run(['npm', 'install', '-g'])
+            self.run(npm_install)
 
 
 def _get_nodejs_base(node_engine):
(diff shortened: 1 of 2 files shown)


snappy-m-o:

snappy-m-o:

smoser:
I dont know if this is exactly what sergio was looking for when he suggested this, but I have verfified that after a successful npm install, the second doees not need network access.  Tested that in a container like this:

    #!/bin/bash
    set -e
    set -o pipefail
    
    LOG="$PWD/my.log"
    cleanup() { network_up; }
    network_up() { sudo ifup eth0; }
    network_down() { sudo ifdown eth0; }
    msg() { echo "$@" >> "$LOG"; echo "$@" 1>&2; }
    run() {
        local envcmd="" n=$SECONDS
        msg "[start]" "${@}"
        "$@"
        r=$?
        msg "[finished ret=$r took=$(($SECONDS-n))]" "$@"
        return $r
    }
    trap cleanup EXIT
    npm_install=( npm --cache-min=Infinity install azure-cli )
    
    # do it as would be in the pull section
    network_up
    rm -Rf pull
    mkdir pull
    msg pull network-up
    ( cd pull && run "${npm_install[@]}" ) 2>&1 | tee pull.log
    
    # now with network down
    msg build network-down
    network_down
    rm -Rf build
    mkdir build
    ( cd build && run "${npm_install[@]}" ) 2>&1 | tee build.log

snapcraft/plugins/nodejs.py
```diff
@@ -96,6 +96,8 @@ def pull(self):
         super().pull()
         os.makedirs(self._npm_dir, exist_ok=True)
         self._nodejs_tar.download()
+        # we build in 'pull' phase to download all dependencies
+        self.build()
```
elopio: This for sure solves the problem, but I'm concerned because I doubt that build was designed to be called from pull. There could be many unintended consequences, and the clean_pull is now a little worse for sure. We now can remove the build completely, and do just pull. That sounds weird. I'm not sure, this is a huge thing for a friday evening :)

sergiusens: El viernes, 26 de agosto de 2016 16h'32:09 ART, Leo Arias 
>> @@ -96,6 +96,8 @@ def pull(self):
>>          super().pull()
>>          os.makedirs(self._npm_dir, exist_ok=True)
>>          self._nodejs_tar.download()
>> +        # we build in 'pull' phase to download all dependencies
>> +        self.build()
> 
> This for sure solves the problem, but I'm concerned because I 
> doubt that build was designed to be called from pull. There 
> could be many unintended consequences, and the clean_pull is now 
> a little worse for sure. We now can remove the build completely, 
> and do just pull. That sounds weird. I'm not sure, this is a 
> huge thing for a friday evening :)
> 

This is exactly what the python plugins do.
I would call an internal function though.



-- 

smoser:
I updated to to address sergiusens' comment.
Separated the build into _npm_install and call it from build and pull.

sergiusens:
ok to test

PR #765, opened 2016-08-29
Use a recursive iglob for filesets

sergiusens:
This adds support for `**` in filesets for parts.

LP: #1616464


elopio:
retest this please

snapcraft/plugins/autotools.py
```diff
@@ -124,11 +124,7 @@ def build(self):
...
-        self._remove_la_files()
-
-    def _remove_la_files(self):
-        for root, _, files in os.walk(self.installdir):
-            for file_name in files:
-                if file_name.endswith('.la'):
-                    os.unlink(os.path.join(root, file_name))
+    def snap_fileset(self):
+        fileset = super().snap_fileset()
+        fileset.append('-**/*.la')
```
elopio: Please copy the comment:
# Remove .la files which don't work when they are moved around

snapcraft/plugins/python2.py
```diff
@@ -206,18 +206,8 @@ def build(self):
...
-        fileset.append('-usr/lib/python*/*/*/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*/*/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*/*/*/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*/*/*/*/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*/*/*/*/*/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*/*/*/*/*/*/*.pyc')
+        fileset.append('-**/*.pth')
+        fileset.append('-**/*.pyc')
```
elopio: A coment on why are you removing them would be nice here too.

snapcraft/plugins/python3.py
```diff
@@ -184,14 +184,6 @@ def python_version(self):
...
-        fileset.append('-usr/lib/python*/*/*/__pycache__/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/__pycache__/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*/__pycache__/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*/*/__pycache__/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*/*/*/__pycache__/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*/*/*/*/__pycache__/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*/*/*/*/*/__pycache__/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*/*/*/*/*/*/__pycache__/*.pyc')
+        fileset.append('-**/*.pth')
+        fileset.append('-**/__pycache__')
```
elopio: and here.

elopio:
:+1:
I worry about @kyrofa, he was so happy about the other branch.

sergiusens:
El 30/08/16 a las 00:49, Leo Arias escribió:
>
> 👍
> I worry about @kyrofa <https://github.com/kyrofa>, he was so happy 
> about the other branch.
>

He said he was going to be happy with this one as well ;-)

integration_tests/test_python_plugin.py
```diff
@@ -83,6 +83,24 @@ def test_build_rewrites_shebangs(self):
...
+        self.run_snapcraft('stage', project_dir)
+
+        pyc_files = []
+        pth_files = []
+        for _, _, files in os.walk(os.path.join(project_dir, 'stage')):
+            pyc_files.extend([f for f in files if f.endswith('pyc')])
+            pth_files.extend([f for f in files if f.endswith('pth')])
+
+        self.assertEqual([], pyc_files)
+        self.assertEqual([], pth_files)
```
kyrofa: If this fails, it might be much easier to digest an error based on the length of the lists rather than directly comparing their contents (it'll print the lists, right?).

snapcraft/internal/pluginhandler.py
```diff
@@ -756,7 +756,8 @@ def _generate_include_set(directory, includes):
     include_files = set()
     for include in includes:
         if '*' in include:
-            matches = glob.glob(os.path.join(directory, include))
+            pattern = os.path.join(directory, include)
+            matches = iglob(pattern, recursive=True)
```
kyrofa: `glob.glob` also accepts the `recursive` param. Is the switch to `iglob` performance-related?

snapcraft/plugins/python2.py
```diff
@@ -206,18 +206,11 @@ def build(self):
...
-        fileset.append('-usr/lib/python*/*/*/*/*/*/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*/*/*/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*/*/*/*/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*/*/*/*/*/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*/*/*/*/*/*/*.pyc')
+        # .pth files are only read from the built-in site-packages directory.
+        # We use PYTHONPATH for everything so not needed.
+        fileset.append('-**/*.pth')
+        # This is a major cause of inter part conflict.
+        fileset.append('-**/*.pyc')
```
kyrofa: This actually pleases me _more_ than the previous PR, as I don't feel like we're fighting snapcraft here.

snapcraft/plugins/python3.py
```diff
@@ -184,14 +184,10 @@ def python_version(self):
...
-        fileset.append('-usr/lib/python*/*/*/*/*/*/*/__pycache__/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*/*/*/*/__pycache__/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*/*/*/*/*/__pycache__/*.pyc')
-        fileset.append('-usr/lib/python*/*/*/*/*/*/*/*/*/*/__pycache__/*.pyc')
+        # .pth files are only read from the built-in site-packages directory.
+        # We use PYTHONPATH for everything so not needed.
+        fileset.append('-**/*.pth')
+        # Holds all the .pyc files. It is a major cause of inter part
+        # conflict.
+        fileset.append('-**/__pycache__')
```
kyrofa: Yes, lovely indeed. Thank you for the comments as well.

kyrofa:
Very nice! I made one ignorable suggestion, :+1: from me.

snapcraft/internal/pluginhandler.py
sergiusens: @kyrofa it should be. We go over the list only when making it a set on the line below

kyrofa: Which means you evaluate them all immediately anyway, right? I don't see the difference between this and `glob` for our usage anyway. That said, I don't have a problem with `iglob`, I was just curious for the reasoning behind the switch.

SamYaple:
So this is actually a problem. Specifically with the .pth removal. There are several projects that I know of that have the .pth to (ab)use python importing. Here is a line from the dogpile.cache pth file that allows it to import from site-packages/dogpile/cache without having site-pacakges/dogpile/__init__.py

import sys, types, os;p = os.path.join(sys._getframe(1).f_locals['sitedir'], *('dogpile',));ie = os.path.exists(os.path.join(p,'__init__.py'));m = not ie and sys.modules.setdefault('dogpile', types.ModuleType('dogpile'));mp = (m or []) and m.__dict__.setdefault('__path__',[]);(p not in mp) and mp.append(p)

there are many such projects that do things like this unfortunately.

sergiusens:
El 22/09/16 a las 13:44, Sam Yaple escribió:
>
> So this is actually a problem. Specifically with the .pth removal. 
> There are several projects that I know of that have the .pth to 
> (ab)use python importing. Here is a line from the dogpile.cache pth 
> file that allows it to import from site-packages/dogpile/cache without 
> having site-pacakges/dogpile/*init*.py
>
> import sys, types, os;p = 
> os.path.join(sys./getframe(1).f_locals['sitedir'], *('dogpile',));ie = 
> os.path.exists(os.path.join(p,'__init_/.py'));m = not ie and 
> sys.modules.setdefault('dogpile', types.ModuleType('dogpile'));mp = (m 
> or []) and m.*dict*.setdefault('*path*',[]);(p not in mp) and mp.append(p)
>
> there are many such projects that do things like this unfortunately.
>

In your snapcraft.yaml, can you try to explicitly add it?

SamYaple:
Thats a silly thing to suggest. This was rather complicated to even figure out why it was broken. Even though I can add this file specifically in, this will lead many people to believe that python does not work well with snapcraft.

Indeed, looking through the IRC logs show several cases where people are complaining about it not being able to import a package after this patch merged, I would bet at least some of those are this exact issue.

Why is removing the pth file something that is even needed?

sergiusens:
El 22/09/16 a las 14:25, Sam Yaple escribió:
>
> Thats a silly thing to suggest. This was rather complicated to even 
> figure out why it was broken. Even though I can add this file 
> specifically in, this will lead many people to believe that python 
> does not work well with snapcraft.
>

I am not saying we should keep it this way! :-)

> Indeed, looking through the IRC logs show several cases where people 
> are complaining about it not being able to import a package after this 
> patch merged, I would bet at least some of those are this exact issue.
>
> Why is removing the pth file something that is even needed?
>

I think this was just oversight

SamYaple:
Submitting a patch :)

Not a big deal in the grand scheme of things. Shame i just missed 2.18 window though

sergiusens:
El 22/09/16 a las 14:29, Sam Yaple escribió:
>
> Submitting a patch :)
>
> Not a big deal in the grand scheme of things. Shame i just missed 2.18 
> window though
>

don't worry, I am preparing 2.18.1 right now.

PR #768, opened 2016-08-30
Update kernel meta to the latest spec.

sergiusens:
LP: #1618019


elopio:
could we have a manual test to check the kernel build?

sergiusens:
@elopio can we have a new bug for that as it is totally unrelated to the current bug?

PR #770, opened 2016-08-30
HACKING.md: list dependencies to install with pip

sergiusens:
Also describe using `PYTHONHOME` as an alternative mechanism.


elopio:
:+1:

PR #771, opened 2016-08-31
Python plugin improvements

sergiusens:
The unit tests still need updating, what do people think about this approach?

The demos still work btw.

### DIFF
--- a/snapcraft/_baseplugin.py
+++ b/snapcraft/_baseplugin.py
@@ -240,6 +240,7 @@ class BasePlugin:
         if not cwd:
             cwd = self.builddir
         print(' '.join(cmd))
+        print('working directory', cwd)
         os.makedirs(cwd, exist_ok=True)
         return common.run(cmd, cwd=cwd, **kwargs)
 
--- a/snapcraft/_baseplugin.py
+++ b/snapcraft/_baseplugin.py
@@ -239,8 +239,7 @@ class BasePlugin:
     def run(self, cmd, cwd=None, **kwargs):
         if not cwd:
             cwd = self.builddir
-        if True:
-            print(' '.join(cmd))
+        print(' '.join(cmd))
         os.makedirs(cwd, exist_ok=True)
         return common.run(cmd, cwd=cwd, **kwargs)
 
--- a/snapcraft/_baseplugin.py
+++ b/snapcraft/_baseplugin.py
@@ -240,7 +240,6 @@ class BasePlugin:
         if not cwd:
             cwd = self.builddir
         print(' '.join(cmd))
-        print('working directory', cwd)
         os.makedirs(cwd, exist_ok=True)
         return common.run(cmd, cwd=cwd, **kwargs)
 
--- a/snapcraft/_baseplugin.py
+++ b/snapcraft/_baseplugin.py
@@ -237,7 +237,7 @@ class BasePlugin:
 
     # Helpers
     def run(self, cmd, cwd=None, **kwargs):
-        if cwd is None:
+        if not cwd:
             cwd = self.builddir
         if True:
             print(' '.join(cmd))
@@ -245,9 +245,7 @@ class BasePlugin:
         return common.run(cmd, cwd=cwd, **kwargs)
 
     def run_output(self, cmd, cwd=None, **kwargs):
-        if cwd is None:
+        if not cwd:
             cwd = self.builddir
-        if True:
-            print(' '.join(cmd))
         os.makedirs(cwd, exist_ok=True)
         return common.run_output(cmd, cwd=cwd, **kwargs)
--- a/snapcraft/plugins/python.py
+++ b/snapcraft/plugins/python.py
@@ -16,8 +16,7 @@
 
 """The python plugin can be used for python 2 or 3 based parts.
 
-The python plugin can be used for python projects where you would
-want to do:
+It can be used for python projects where you would want to do:
 
     - import python modules with a requirements.txt
     - build a python project that has a setup.py
--- a/snapcraft/plugins/python.py
+++ b/snapcraft/plugins/python.py
@@ -114,7 +114,8 @@ class PythonPlugin(snapcraft.BasePlugin):
                 'python-setuptools',
             ]
         else:
-            return []
+            raise AssertionError('python-version has a value of {!r}'.format(
+                self.options.python_version))
 
     @property
     def plugin_stage_packages(self):
@@ -123,7 +124,8 @@ class PythonPlugin(snapcraft.BasePlugin):
         elif self.options.python_version == 'python2':
             return ['python2']
         else:
-            return []
+            raise AssertionError('python-version has a value of {!r}'.format(
+                self.options.python_version))
 
     @property
     def system_pip_command(self):
--- a/snapcraft/plugins/python.py
+++ b/snapcraft/plugins/python.py
@@ -113,9 +113,6 @@ class PythonPlugin(snapcraft.BasePlugin):
                 'python-pkg-resources',
                 'python-setuptools',
             ]
-        else:
-            raise AssertionError('python-version has a value of {!r}'.format(
-                self.options.python_version))
 
     @property
     def plugin_stage_packages(self):
@@ -123,9 +120,6 @@ class PythonPlugin(snapcraft.BasePlugin):
             return ['python3']
         elif self.options.python_version == 'python2':
             return ['python']
-        else:
-            raise AssertionError('python-version has a value of {!r}'.format(
-                self.options.python_version))
 
     @property
     def system_pip_command(self):
--- a/snapcraft/plugins/python.py
+++ b/snapcraft/plugins/python.py
@@ -206,7 +206,8 @@ class PythonPlugin(snapcraft.BasePlugin):
             self.run(pip_install + self.options.python_packages, env=env)
 
         if os.path.exists(setup):
-            self.run(pip_install + ['.'], cwd=self.sourcedir, env=env)
+            cwd = os.path.dirname(setup)
+            self.run(pip_install + ['.'], cwd=cwd, env=env)
 
     def build(self):
         super().build()
--- a/snapcraft/plugins/python.py
+++ b/snapcraft/plugins/python.py
@@ -47,11 +47,12 @@ Additionally, this plugin uses the following plugin-specific keywords:
 
 import os
 import re
+import stat
 import subprocess
 from glob import glob
 
 import snapcraft
-from snapcraft import common
+from snapcraft import file_utils
 
 
 class PythonPlugin(snapcraft.BasePlugin):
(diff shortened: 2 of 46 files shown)


snapcraft/plugins/python2.py
```diff
@@ -40,182 +40,35 @@
...
 
-import snapcraft
-from snapcraft import common
+from snapcraft.plugins import python3
 
 logger = logging.getLogger(__name__)
 
 
-class Python2Plugin(snapcraft.BasePlugin):
+class Python2Plugin(python3.Python3Plugin):
```
elopio: this is not a great idea, conceptually. Python3 is not a parent of python2, so inheritance could cause problems in the future. A better approach might be to make a common BasePythonPlugin, and inherit both Python2 and Python3 plugins from it.

kyrofa: +1, agreed.

sergiusens: and where would `BasePythonPlugin` live? in the `python3` module or the `python2` one?

elopio: neither. In a common module.

sergiusens: El 01/09/16 a las 11:04, Leo Arias escribió:
>
> In snapcraft/plugins/python2.py 
> <https://github.com/snapcore/snapcraft/pull/771#discussion_r77179576>:
>
> >
> >  logger = logging.getLogger(__name__)
> >
> >
> > -class Python2Plugin(snapcraft.BasePlugin):
> > +class Python2Plugin(python3.Python3Plugin):
>
> neither. In a common module.
>

I don't know, this gives the plugins we write special treatment, this 
module would have to be a plugin as well and we risk having a useless 
plugin like the jdk one.

snapcraft/_baseplugin.py
```diff
@@ -237,17 +237,15 @@ def parallel_build_count(self):
 
     # Helpers
     def run(self, cmd, cwd=None, **kwargs):
-        if cwd is None:
+        if not cwd:
             cwd = self.builddir
         if True:
```
elopio: you can just remove the if True, and leave the print for now.

snapcraft/plugins/python.py
```diff
@@ -0,0 +1,235 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""The python plugin can be used for python 2 or 3 based parts.
+
+The python plugin can be used for python projects where you would
+want to do:
```
elopio: s/The python plugin/It

snapcraft/plugins/python.py
```diff
@@ -0,0 +1,235 @@
...
+            ]
+        elif self.options.python_version == 'python2':
+            return [
+                'python-dev',
+                'python-pip',
+                'python-pkg-resources',
+                'python-setuptools',
+            ]
+        else:
+            return []
```
elopio: raise an exception? This should never happen.
Now this is a good use for assert :p

snapcraft/plugins/python.py
```diff
@@ -0,0 +1,235 @@
...
+            },
+            'default': [],
+        }
+        schema['properties']['process-dependency-links'] = {
+            'type': 'boolean',
+            'default': False,
+        }
+        schema['properties']['python-version'] = {
+            'type': 'string',
+            'default': 'python3',
```
elopio: is there a way to enforce the two valid values?

sergiusens: El 01/09/16 a las 12:56, Leo Arias escribió:
>
> In snapcraft/plugins/python.py 
> <https://github.com/snapcore/snapcraft/pull/771#discussion_r77202263>:
>
> > +        schema['properties']['python-packages'] = {
> > +            'type': 'array',
> > +            'minitems': 1,
> > +            'uniqueItems': True,
> > +            'items': {
> > +                'type': 'string'
> > +            },
> > +            'default': [],
> > +        }
> > +        schema['properties']['process-dependency-links'] = {
> > +            'type': 'boolean',
> > +            'default': False,
> > +        }
> > +        schema['properties']['python-version'] = {
> > +            'type': 'string',
> > +            'default': 'python3',
>
> is there a way to enforce the two valid values?
>

yes I added an enum

snapcraft/plugins/python.py
```diff
@@ -0,0 +1,235 @@
...
+        schema.pop('required')
+
+        # Inform Snapcraft of the properties associated with pulling. If these
+        # change in the YAML Snapcraft will consider the pull step dirty.
+        schema['pull-properties'].extend([
+            'requirements',
+            'constraints',
+            'python-packages',
+            'python-version',
+        ])
```
elopio: What about the build properties? 
I'm not sure how does this work. If the pull properties are changed, and snapcraft build is executed, will it run pull again?

snapcraft/plugins/python.py
```diff
@@ -0,0 +1,236 @@
...
+            return []
+
+    @property
+    def plugin_stage_packages(self):
+        if self.options.python_version == 'python3':
+            return ['python3']
+        elif self.options.python_version == 'python2':
+            return ['python2']
+        else:
+            return []
```
elopio: this too should never happen

snapcraft/_baseplugin.py
sergiusens: yeah, sounds good

snapcraft/plugins/python3.py
```diff
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright (C) 2015 Canonical Ltd
+# Copyright (C) 2015-2016 Canonical Ltd
```
elopio: 2015, 2016

snapcraft/plugins/python3.py
```diff
@@ -16,6 +16,8 @@
 
 """The python3 plugin can be used for python 3 based parts.
 
+This plugin is DEPRECATED in favor of the python plugin.
```
elopio: print a deprecation warning when it's used?

snapcraft/plugins/python.py
sergiusens: @kyrofa why was `build-properties` not originally added here?

snapcraft/plugins/python2.py
```diff
@@ -31,191 +33,31 @@
...
-            'constraints',
-            'python-packages'
-        ])
+        del schema['properties']['python-version']
+        schema['pull-properties'].remove('python-version')
 
         return schema
 
     def __init__(self, name, options, project):
+        setattr(options, 'python_version', 'python2')
```
elopio: why setattr instead of options.python_version = ... ?

elopio:
oh, yes! /me approves.

snapcraft/plugins/python2.py
sergiusens: Because it doesn't exist, options gets created from the plugin's schema

snapcraft/plugins/python3.py
sergiusens: El 01/09/16 a las 13:01, Leo Arias escribió:
>
> In snapcraft/plugins/python3.py 
> <https://github.com/snapcore/snapcraft/pull/771#discussion_r77203058>:
>
> > @@ -1,6 +1,6 @@
> >  # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
> >  #
> > -# Copyright (C) 2015 Canonical Ltd
> > +# Copyright (C) 2015-2016 Canonical Ltd
>
> 2015, 2016
>
>
we already had this discussion ;-)

elopio: yes, and as far as I remember, I won ^_^

snapcraft/plugins/python.py
```diff
@@ -0,0 +1,237 @@
...
+                self.options.python_version))
+
+    @property
+    def plugin_stage_packages(self):
+        if self.options.python_version == 'python3':
+            return ['python3']
+        elif self.options.python_version == 'python2':
+            return ['python']
+        else:
+            raise AssertionError('python-version has a value of {!r}'.format(
```
SamYaple: no reason to do this. just don't pop 'required' from the options schema. enforce the requirement of python_version (and specific values with enum so we dont have to validate it either)

elopio: yup, and this could be a dictionary. return self._python_stage_packages[self.options.python_version]

Then an exception would be raised anyway if there is a programmer error somewhere.

elopio:
I think this is very good. I left some minor ignorable comments, so you'll have my +1...
after you add an integration test running the prime command on a  snap with "plugin: python" :laughing:

SamYaple:
You beat me to it! Glad you also like the approach of python2 and python3 inheriting a 'python' plugin for deprecation. I could think of no other way to do it

snapcraft/plugins/python.py
sergiusens: the problem is on inheritance. Adding the schema makes the option visible to the python2 and oython3 plugins

sergiusens: IOW this is here more for the children than the parent itself

sergiusens:
LP: #1616032

snapcraft/plugins/python.py
SamYaple: Hmmm. I suppose schema could be added to the children too if that a problem.

I just don't like the idea of checking python versions in multiple places. Perhaps abstract the else to another function so we don't have to change exceptions in multiple places should this need to be handled differently in the future? (I'm seeing potential uses for python3.4 vs python3.5 maybe). I just don't like violating DRY in this way.

sergiusens: Seems like something that can be decorated.

sergiusens:
LP: #1616407

sergiusens:
LP: #1606894

sergiusens:
LP: #1602079

sergiusens:
retest this please

sergiusens:
LP: #1594809

sergiusens:
retest this please

sergiusens:
LP: #1613898

sergiusens:
LP: #1607297

snapcraft/plugins/python.py
```diff
@@ -235,3 +245,16 @@ def snap_fileset(self):
         elif self.options.python_version == 'python2':
             fileset.append('-**/*.pyc')
         return fileset
+
+
+def _replicate_owner_mode(path):
```
elopio: would you mind explaining in a docstring what is this doing?

SamYaple:
Merge merge merge :+1: 

Patchset looks good. I have been using it exclusively for a few days now. Possibly some docs/testing stuff that is needed, but functionally it is solid.

elopio:
when trying to snap integration_tests/snaps/pip-requirements-list, it results in a conflict with bin/jsonschema. This is known, right? It's part of the change you commented in the mailing list about entry points?

snapcraft/plugins/python.py
```diff
@@ -0,0 +1,254 @@
...
+
+    def __init__(self, name, options, project):
+        super().__init__(name, options, project)
+        self.build_packages.extend(self.plugin_build_packages)
+        self.stage_packages.extend(self.plugin_stage_packages)
+
+    def env(self, root):
+        return [
+            'PYTHONUSERBASE={}'.format(root),
+            'PYTHONHOME={}'.format(os.path.join(root, 'usr'))
```
elopio: Take a look, this is where I see the problem:
```
$ ls prime/usr/lib/python2.7/dist-packages/
README
$ ls prime/usr/lib/python2.7/site-packages
ls: cannot access 'prime/usr/lib/python2.7/site-packages': No such file or directory
$ ls prime/lib/python2.7/site-packages/
argparse-1.2.2.dist-info  easy_install.py   functools32-3.2.3.post2.dist-info  pip                  setuptools                   wheel-0.29.0.dist-info
argparse.py               easy_install.pyc  jsonschema                         pip-8.1.2.dist-info  setuptools-26.1.1.dist-info
argparse.pyc              functools32       jsonschema-2.5.1.dist-info         pkg_resources        wheel
```

So there's nothing in usr, all the installed packages end in lib. What I don't yet understand is why when I try with jsonschema, it finds the module anyway. But when I try with my matrix snap, it doesn't. :pensive:

sergiusens:
El 07/09/16 a las 00:32, Leo Arias escribió:
>
> when trying to snap integration_tests/snaps/pip-requirements-list, it 
> results in a conflict with bin/jsonschema. This is known, right? It's 
> part of the change you commented in the mailing list about entry points?
>

There is a fileset the takes this into account, integration tests pass, 
so I don't see what the problem is. This means entry point work now btw!

sergiusens:
LP: #1619104

PR #773, opened 2016-09-01
parser - Handle project name and version variables

josepht:
Replace $SNAPCRAFT_PROJECT_VERSION and $SNAPCRAFT_PROJECT_NAME
with their respective values from the origin's snapcraft.yaml
in remote parts definitions.

LP:#1619418

### DIFF
--- a/snapcraft/internal/parser.py
+++ b/snapcraft/internal/parser.py
@@ -134,7 +134,24 @@ def _update_source(part, origin):
     return part
 
 
-def _process_entry_parts(entry_parts, parts, origin, maintainer, description):
+def _replace_variables(item, name, version):
+    if isinstance(item, dict):
+        for key, value in item.items():
+            item[key] = _replace_variables(value, name, version)
+    elif isinstance(item, list):
+        for index in range(len(item)):
+            item[index] = _replace_variables(item[index], name, version)
+    elif isinstance(item, str):
+        if name:
+            item = item.replace('$SNAPCRAFT_PROJECT_NAME', name)
+        if version:
+            item = item.replace('$SNAPCRAFT_PROJECT_VERSION', version)
+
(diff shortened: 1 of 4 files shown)


snapcraft/internal/parser.py
```diff
@@ -134,7 +134,24 @@ def _update_source(part, origin):
     return part
 
 
-def _process_entry_parts(entry_parts, parts, origin, maintainer, description):
+def _replace_variables(item, name, version):
```
sergiusens: this looks a lot like `project_loader`'s `replace_attr`. Maybe generalize it a bit to define a list of replacement tuples such that you can pass something like this

```python
[
    ('$SNAPCRAFT_PROJECT_NAME', name),
    ('$SNAPCRAFT_PROJECT_VERSION', version),
]
```

and then in the function for the `str` instances:
```python
for replacement in replacements:
    attr.replace(replacement[0], replacement[1])
```

josepht: Done!

snapcraft/internal/parser.py
```diff
@@ -134,7 +134,23 @@ def _update_source(part, origin):
     return part
 
 
-def _process_entry_parts(entry_parts, parts, origin, maintainer, description):
+def _replace_variables(item, replacements):
```
sergiusens: Hmm, there might have been a miscomm somewhere :)

The general idea was to remove the `_` from https://github.com/snapcore/snapcraft/blob/master/snapcraft/internal/project_loader.py#L201 and adapt that ;-)

sergiusens:
retest this please

snapcraft/internal/parser.py
josepht: Updated.

PR #776, opened 2016-09-07
Updated the contents of the snapcraft init template

elopio:
LP: #1620428

### DIFF
--- a/snapcraft/internal/lifecycle.py
+++ b/snapcraft/internal/lifecycle.py
@@ -41,7 +41,7 @@ logger = logging.getLogger(__name__)
 
 
 _TEMPLATE_YAML = """name: my-snap-name # you probably want to 'snapcraft register <name>'
-version: "0.1" # just for humans, typically '1.2+git' or '1.3.2'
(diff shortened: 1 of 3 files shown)


snapcraft/internal/lifecycle.py
```diff
@@ -40,23 +40,23 @@
...
 
-_TEMPLATE_YAML = """name: my-snap  # the name of the snap
-version: 0  # the version of the snap
-summary: This is my-snap's summary  # 79 char long summary
-description: This is my-snap's description  # a longer description for the snap
-confinement: devmode  # use "strict" to enforce system access only via \
-declared interfaces
-grade: devel # use "stable" to assert the snap quality
+_TEMPLATE_YAML = """name: my-snap-name # you probably want to 'snapcraft register <name>'
+version: "0.1" # just for humans, typically '1.2+git' or '1.3.2'
```
sergiusens: Wasn't `version: "0.1"` supposed to have single quotes according to the last bug comment from Mark?

elopio: I think you are right. Pushed.

elopio:
retest this please

PR #777, opened 2016-09-08
go plugin: don't put debugging symbols in generated binaries

teknoraver:
Go binaries should never stripped or they can behave badly,
but it's possible to avoid debugging symbol generation at build time.
The binaries are roughly half the size

### DIFF
--- a/snapcraft/plugins/go.py
+++ b/snapcraft/plugins/go.py
@@ -144,12 +144,12 @@ class GoPlugin(snapcraft.BasePlugin):
         super().build()
 
         for go_package in self.options.go_packages:
-            self._run(['go', 'install', go_package])
+            self._run(['go', 'install', '--ldflags=-s -w', go_package])
         if not self.options.go_packages:
             tags = []
             if self.options.go_buildtags:
                 tags = ['-tags={}'.format(','.join(self.options.go_buildtags))]
-            self._run(['go', 'install'] + tags +
+            self._run(['go', 'install', '--ldflags=-s -w'] + tags +
                       ['./{}/...'.format(self._get_local_go_package())])
 
         install_bin_path = os.path.join(self.installdir, 'bin')
--- a/snapcraft/tests/test_plugin_go.py
+++ b/snapcraft/tests/test_plugin_go.py
@@ -227,7 +227,7 @@ class GoPluginTestCase(tests.TestCase):
         plugin.build()
 
         self.run_mock.assert_called_once_with(
-            ['go', 'install', './dir/...'],
+            ['go', 'install', '--ldflags=-s -w', './dir/...'],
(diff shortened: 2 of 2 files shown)


snappy-m-o:

snappy-m-o:

elopio:
Hello @teknoraver. Thanks a lot for your contribution!
When we talked about buildtags in #733, we also discussed a little about the other arguments.
Maybe this can be part of the schema too, with a default value that can be overwritten by the projects. @sergiusens do you agree?

Also, for every change we require tests. Please take a look at snapcraft/tests and integration_tests.

sergiusens:
ok to test

teknoraver:
@elopio I've changed the unit testing accordingly, indeed the automated testing are just passed.
What do you think about making this the default behaviour and add a go-buildtags named "debug" or wathever to strip the symbols?

btw, the space saved is more or less 50%:
```
$ go build -o snapd-debug github.com/snapcore/snapd/cmd/snapd
$ go build -ldflags='-s -w' -o snapd-release github.com/snapcore/snapd/cmd/snapd
$ ll snapd-debug snapd-release
-rwxrwxr-x 1 matteo matteo  13M set 16 02:50 snapd-debug
-rwxrwxr-x 1 matteo matteo 7,7M set 16 02:51 snapd-release
```

sergiusens:
@teknoraver we strip when doing debuilds though. I am just a bit afraid that this will move away from our goal of having `stage` with debugging symbols and `prime` with lean and mean binaries

kyrofa:
Note that it sounds like `strip` will work against go binaries. Note also that at least in my use-cases, I always configure my projects to build as a release instead of a debug build (e.g. cmake). Though I'd probably stop doing that if I knew snapcraft stripped in order to prime.

teknoraver:
You're right, it seems that Go support stripping of binaries since version X, and the resulting file is the same (I used `strip -s -R.comment`)
```
$ ls -l
-rwxrwxr-x 1 matteo matteo 15661976 ott 18 10:43 snapd-full
-rwxrwxr-x 1 matteo matteo 10596344 ott 18 10:46 snapd-full-strip
-rwxrwxr-x 1 matteo matteo 10596456 ott 18 10:43 snapd-nodbg
-rwxrwxr-x 1 matteo matteo 10596344 ott 18 10:44 snapd-nodbg-strip
$ md5sum *
e02c01dfb1b3c24edc9ac0b45c82f7f5  snapd-full
7f689651b8504e92d1a23608fb04e0e9  snapd-full-strip
308f9b8169fb7fc1e4bbd2ff13712504  snapd-nodbg
7f689651b8504e92d1a23608fb04e0e9  snapd-nodbg-strip
```

PR #778, opened 2016-09-08
Use --force-dangerous when testing the installation of snaps

elopio:
LP: #1621535

### DIFF
--- a/debian/tests/control
+++ b/debian/tests/control
@@ -22,4 +22,4 @@ Depends: @,
          python3-fixtures,
          python3-pexpect,
          python3-testscenarios,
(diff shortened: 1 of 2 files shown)


ogra1:
--force-dangerous is deprecated (that was fast, eh ?) it is just called --dangerous now

elopio:
@ogra1 fast, but didn't make it into xenial-updates yet. We will need another similar branch next week.

kyrofa:
@ogra1 there's deprecated, and then there's just completely removed :stuck_out_tongue: .

sergiusens:
:cry:

PR #779, opened 2016-09-08
Do not compile pyc files when installing with pip

SamYaple:
LP: #1621540

### DIFF
--- a/snapcraft/plugins/python.py
+++ b/snapcraft/plugins/python.py
@@ -153,7 +153,7 @@ class PythonPlugin(snapcraft.BasePlugin):
         env['PYTHONUSERBASE'] = self.installdir
 
         subprocess.check_call([
-            self.system_pip_command, 'install', '--user',
+            self.system_pip_command, 'install', '--user', '--no-compile',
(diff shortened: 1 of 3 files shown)


snappy-m-o:

snappy-m-o:

sergiusens:
Interesting, I thought when we talked that `--no-compile` did not do the trick, mind removing the fileset in the plugin that removes `pyc` and `__pycache__` if this indeed does work?

SamYaple:
Yea there were still some pyc files that remains, im unsure why just yet. But the --no-compile option still speed up build time and effectively stopped creating most of the pyc files (and that is the documented purpose of --no-compile)

sergiusens:
Try to setup the env with https://docs.python.org/3/using/cmdline.html#envvar-PYTHONDONTWRITEBYTECODE

sergiusens:
ok to test

SamYaple:
Yea that ENV var fixes it for me.

sergiusens:
@SamYaple ok, so different branch then :-)

PR #780, opened 2016-09-08
Organize without removing files on target

sergiusens:
LP: #1599709

### DIFF
--- a/snapcraft/internal/pluginhandler.py
+++ b/snapcraft/internal/pluginhandler.py
@@ -671,6 +671,7 @@ def _migrate_files(snap_files, snap_dirs, srcdir, dstdir, missing_ok=False,
 
         fixup_func(dst)
 
+
 def _organize_filesets(fileset, base_dir):
     for key in fileset:
         src = os.path.join(base_dir, key)
@@ -690,6 +691,7 @@ def _organize_filesets(fileset, base_dir):
         else:
             os.makedirs(os.path.dirname(dst), exist_ok=True)
             shutil.move(src, dst)
+
 
 def _clean_migrated_files(snap_files, snap_dirs, directory):
     for snap_file in snap_files:
--- a/snapcraft/internal/pluginhandler.py
+++ b/snapcraft/internal/pluginhandler.py
@@ -678,8 +678,9 @@ def _organize_filesets(fileset, base_dir):
 
         if os.path.isfile(dst):
             raise EnvironmentError(
-                'Stepping over existing file for organization %r',
-                os.path.relpath(dst, base_dir))
+                'Trying to organize file {!r} to {!r}, '
+                'but it already exists'.format(
+                    key, os.path.relpath(dst, base_dir)))
 
         if os.path.isdir(src):
             file_utils.link_or_copy_tree(src, dst)
--- a/snapcraft/internal/pluginhandler.py
+++ b/snapcraft/internal/pluginhandler.py
@@ -324,23 +324,9 @@ class PluginHandler:
         return _migratable_filesets(fileset, self.code.installdir)
 
     def _organize(self):
-        organize_fileset = getattr(self.code.options, 'organize', {}) or {}
+        fileset = getattr(self.code.options, 'organize', {}) or {}
 
-        for key in organize_fileset:
-            src = os.path.join(self.code.installdir, key)
-            dst = os.path.join(self.code.installdir, organize_fileset[key])
-
-            os.makedirs(os.path.dirname(dst), exist_ok=True)
-
-            if os.path.exists(dst):
-                logger.warning(
-                    'Stepping over existing file for organization %r',
-                    os.path.relpath(dst, self.code.installdir))
-                if os.path.isdir(dst):
-                    shutil.rmtree(dst)
-                else:
-                    os.remove(dst)
-            shutil.move(src, dst)
+        _organize_filesets(fileset.copy(), self.code.installdir)
 
     def stage(self, force=False):
         self.makedirs()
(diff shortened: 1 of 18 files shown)


snapcraft/internal/pluginhandler.py
```diff
@@ -324,23 +324,9 @@ def migratable_fileset_for(self, step):
...
-            if os.path.exists(dst):
-                logger.warning(
-                    'Stepping over existing file for organization %r',
-                    os.path.relpath(dst, self.code.installdir))
-                if os.path.isdir(dst):
-                    shutil.rmtree(dst)
-                else:
-                    os.remove(dst)
-            shutil.move(src, dst)
+        _organize_filesets(fileset.copy(), self.code.installdir)
```
kyrofa: Why `.copy()` here?

snapcraft/internal/pluginhandler.py
```diff
@@ -685,6 +671,24 @@ def _migrate_files(snap_files, snap_dirs, srcdir, dstdir, missing_ok=False,
...
 
+def _organize_filesets(fileset, base_dir):
+    for key in fileset:
+        src = os.path.join(base_dir, key)
+        dst = os.path.join(base_dir, fileset[key])
+
+        if os.path.isfile(dst):
+            raise EnvironmentError(
+                'Stepping over existing file for organization %r',
+                os.path.relpath(dst, base_dir))
```
kyrofa: This message implies that this is just a warning and things will continue, but it really seems to be an error.

snapcraft/internal/pluginhandler.py
```diff
@@ -685,6 +671,24 @@ def _migrate_files(snap_files, snap_dirs, srcdir, dstdir, missing_ok=False,
...
+
+        if os.path.isdir(src):
+            file_utils.link_or_copy_tree(src, dst)
+            # TODO create alternate organization location to avoid
+            # deletions.
```
kyrofa: Yeah, and it'd be ideal if file-overwriting was caught by the conflict detector instead of us having to do it here.

kyrofa: Scratch that, it will, but this needs to be within the same part.

snapcraft/internal/pluginhandler.py
kyrofa: Maybe something like "Trying to organize file {} to {}, but it already exists"

kyrofa:
Approach looks good, couple comments.

snapcraft/internal/pluginhandler.py
sergiusens: Because I am calling a function outside of the class and being extra careful, that's all

snapcraft/internal/pluginhandler.py
sergiusens: What text do you recommend?

snapcraft/internal/pluginhandler.py
kyrofa: But you don't rely on `fileset` being immutable here, do you? You don't use it again, the function exits here.

snapcraft/internal/pluginhandler.py
```diff
@@ -324,23 +324,9 @@ def migratable_fileset_for(self, step):
         return _migratable_filesets(fileset, self.code.installdir)
 
     def _organize(self):
-        organize_fileset = getattr(self.code.options, 'organize', {}) or {}
+        fileset = getattr(self.code.options, 'organize', {}) or {}
```
kyrofa: Isn't `fileset = getattr(self.code.options, 'organize', {})` enough? The `or {}` is redundant, no?

snapcraft/internal/pluginhandler.py
kyrofa: Something like my second comment in this thread :wink: .

snapcraft/internal/pluginhandler.py
sergiusens: this was the origin of the problems, `self.code.options.organize` is what gets out.

snapcraft/internal/pluginhandler.py
sergiusens: This predates the days we had a jsonschema, so it *was* necessary, I am a little hesitant to remove it now though.

snapcraft/internal/pluginhandler.py
kyrofa: Ah, okay same page.

snapcraft/internal/pluginhandler.py
sergiusens: I would also remove `filesets`, `snap`, `prime`, `stage`, `source`, `source-.*` and `organize` out of options and into the pluginhandler itself.

sergiusens: There you go [LP: #1621601](https://bugs.launchpad.net/snapcraft/+bug/1621601)

snapcraft/internal/pluginhandler.py
```diff
@@ -685,6 +671,25 @@ def _migrate_files(snap_files, snap_dirs, srcdir, dstdir, missing_ok=False,
...
+def _organize_filesets(fileset, base_dir):
+    for key in fileset:
+        src = os.path.join(base_dir, key)
+        dst = os.path.join(base_dir, fileset[key])
+
+        if os.path.isfile(dst):
+            raise EnvironmentError(
+                'Trying to organize file {!r} to {!r}, '
+                'but it already exists'.format(
+                    key, os.path.relpath(dst, base_dir)))
```
kyrofa: Good call on keeping it relative.

kyrofa:
I present you with my :+1:

PR #781, opened 2016-09-08
Workaround bzr with auth proxies

sergiusens:
We just disable it as a workaround, mostly for building
with launchpad.

LP: #1606203

### DIFF
--- a/snapcraft/internal/sources.py
+++ b/snapcraft/internal/sources.py
@@ -78,6 +78,7 @@ from snapcraft import file_utils
 from snapcraft.internal.indicators import download_requests_stream
 
 
+logger = logging.getLogger(__name__)
 logging.getLogger('urllib3').setLevel(logging.CRITICAL)
 
 
@@ -137,6 +138,29 @@ class Bazaar(Base):
             raise IncompatibleOptionsError(
                 'can\'t specify a source-branch for a bzr source')
 
+    def _get_environment(self):
+        # A workaround for LP: #1606203
+        env = os.environ.copy()
+
+        proxies = dict(
+            http_proxy=env.get('http_proxy', ''),
+            https_proxy=env.get('https_proxy', ''),
+        )
+
+        proxies = {key: proxies[key] for key in proxies if proxies[key]}
+
+        for key in proxies:
+            if '@' in proxies[key]:
+                logger.warning(
+                    'Removing the {!r} environment variable as '
+                    'authenticated proxies and bzr will not '
+                    'work'.format(key))
+                proxies[key] = ''
+
+        env.update(proxies)
+
+        return env
+
     def pull(self):
         tag_opts = []
         if self.source_tag:
@@ -149,7 +173,7 @@ class Bazaar(Base):
             cmd = ['bzr', 'branch'] + tag_opts + \
                   [self.source, self.source_dir]
 
-        subprocess.check_call(cmd)
+        subprocess.check_call(cmd, env=self._get_environment())
 
 
 class Git(Base):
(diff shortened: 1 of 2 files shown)


kyrofa:
:+1: yick.

snapcraft/internal/sources.py
```diff
@@ -137,6 +138,29 @@ def __init__(self, source, source_dir, source_tag=None,
...
+
+        proxies = {key: proxies[key] for key in proxies if proxies[key]}
+
+        for key in proxies:
+            if '@' in proxies[key]:
+                logger.warning(
+                    'Removing the {!r} environment variable as '
+                    'authenticated proxies and bzr will not '
+                    'work'.format(key))
+                proxies[key] = ''
```
elopio: I don't understand why it got so complicated. What is that second assignment you are doing to proxies?
One simple loop should be enough.

```
>>> proxies = dict(
...   http_proxy='test@test',
...   https_proxy='test@test',
... )
>>> for key in proxies.keys():
...   if '@' in proxies[key]:
...     proxies[key] = ''
... 
>>> proxies
{'https_proxy': '', 'http_proxy': ''}
```

sergiusens: El 09/09/16 a las 00:22, Leo Arias escribió:
>
> In snapcraft/internal/sources.py 
> <https://github.com/snapcore/snapcraft/pull/781#discussion_r78124084>:
>
> > +        env = os.environ.copy()
> > +
> > +        proxies = dict(
> > +            http_proxy=env.get('http_proxy', ''),
> > +            https_proxy=env.get('https_proxy', ''),
> > +        )
> > +
> > +        proxies = {key: proxies[key] for key in proxies if proxies[key]}
> > +
> > +        for key in proxies:
> > +            if '@' in proxies[key]:
> > +                logger.warning(
> > +                    'Removing the {!r} environment variable as '
> > +                    'authenticated proxies and bzr will not '
> > +                    'work'.format(key))
> > +                proxies[key] = ''
>
> I don't understand why it got so complicated. What is that second 
> assignment you are doing to proxies?
> One simple loop should be enough.
>
> |>>> proxies = dict( ... http_proxy='test@test', ... 
> https_proxy='test@test', ... ) >>> for key in proxies.keys(): ... if 
> '@' in proxies[key]: ... proxies[key] = '' ... >>> proxies 
> {'https_proxy': '', 'http_proxy': ''}|

|What if one of those entries don't exist?|||

elopio: then env.get will return '', if '@' will be false and the env var will remain empty.
If you would like for the key to remain unset, then maybe:

```
patched_vars = {}
for var in ('http_proxy', 'https_proxy'):
    if '@' in env.get(var, ''):
        patched_vars[var] = ''
env.update(patched_vars)
```

vilagithub:
Hmmm, sorry to be late in the game I got pinged only yesterday about https://bugs.launchpad.net/snapcraft/+bug/1606203  :-/

None of the log files are available anymore so I can't be 100% sure.

But it seems to me the issue here is caused by the launchpad pluging not using launchpadlib and introducing a bug in the proxy handling for a launchpad very specific feature (checking the related packaging branch) which is not supported by launchpad anymore anyway ;)

Long story short, if there is a way to reproduce, adding '-Olaunchpad.packaging_verbosity=off' to the 'bzr pull' command line should avoid the issue.

Your patch here will break all uses of authenticated proxies, which are, AFAIK, properly supported for http and https.

Alternatively, if this issue occurs only on launchpad builders 'no_proxy=launchpad.net' may be enough ?

sergiusens:
Fix in bazaar, no need for this ugly hack!

PR #784, opened 2016-09-09
Support globbing for organize

sergiusens:
LP: #1616459

### DIFF
--- a/snapcraft/internal/pluginhandler.py
+++ b/snapcraft/internal/pluginhandler.py
@@ -673,24 +673,26 @@ def _migrate_files(snap_files, snap_dirs, srcdir, dstdir, missing_ok=False,
 
 
 def _organize_filesets(fileset, base_dir):
-    for key in fileset:
+    for key in sorted(fileset, key=lambda x: ['*' in x, x]):
         src = os.path.join(base_dir, key)
         dst = os.path.join(base_dir, fileset[key])
 
-        if os.path.isfile(dst):
-            raise EnvironmentError(
-                'Trying to organize file {!r} to {!r}, '
-                'but it already exists'.format(
-                    key, os.path.relpath(dst, base_dir)))
+        sources = iglob(src, recursive=True)
 
-        if os.path.isdir(src):
-            file_utils.link_or_copy_tree(src, dst)
-            # TODO create alternate organization location to avoid
-            # deletions.
-            shutil.rmtree(src)
-        else:
-            os.makedirs(os.path.dirname(dst), exist_ok=True)
-            shutil.move(src, dst)
+        for src in sources:
+            if os.path.isdir(src) and '*' not in key:
+                file_utils.link_or_copy_tree(src, dst)
+                # TODO create alternate organization location to avoid
+                # deletions.
+                shutil.rmtree(src)
+            elif os.path.isfile(dst):
+                raise EnvironmentError(
+                    'Trying to organize file {key!r} to {dst!r}, '
+                    'but {dst!r} already exists'.format(
+                        key=key, dst=os.path.relpath(dst, base_dir)))
+            else:
+                os.makedirs(os.path.dirname(dst), exist_ok=True)
+                shutil.move(src, dst)
 
 
 def _clean_migrated_files(snap_files, snap_dirs, directory):
--- a/snapcraft/tests/test_pluginhandler.py
+++ b/snapcraft/tests/test_pluginhandler.py
@@ -671,7 +671,54 @@ class OrganizeTestCase(tests.TestCase):
             setup_files=['foo', 'bar'],
             organize_set={'foo': 'bar'},
             expected=EnvironmentError,
-        ))
+        )),
+        ('*_for_files', dict(
+            setup_dirs=[],
+            setup_files=['foo.conf', 'bar.conf'],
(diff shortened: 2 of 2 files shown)


josepht:
Looks good to me.

snapcraft/internal/pluginhandler.py
```diff
@@ -673,24 +673,26 @@ def _migrate_files(snap_files, snap_dirs, srcdir, dstdir, missing_ok=False,
 
 
 def _organize_filesets(fileset, base_dir):
-    for key in fileset:
+    for key in sorted(fileset, key=lambda x: ['*' in x, x]):
```
kyrofa: Forgive my ignorance, but what is this lambda doing?

snapcraft/internal/pluginhandler.py
```diff
@@ -673,24 +673,26 @@ def _migrate_files(snap_files, snap_dirs, srcdir, dstdir, missing_ok=False,
...
+                # deletions.
+                shutil.rmtree(src)
+            elif os.path.isfile(dst):
+                raise EnvironmentError(
+                    'Trying to organize file {key!r} to {dst!r}, '
+                    'but {dst!r} already exists'.format(
+                        key=key, dst=os.path.relpath(dst, base_dir)))
+            else:
+                os.makedirs(os.path.dirname(dst), exist_ok=True)
+                shutil.move(src, dst)
```
kyrofa: This may leave empty parent directories lying around (something that is currently a problem). Could be fixed with a while loop looking for empty parent directories and stopping at the installdir.

snapcraft/internal/pluginhandler.py
kyrofa: Ah ha, nevermind.

snapcraft/internal/pluginhandler.py
sergiusens: El 09/09/16 a las 15:49, Kyle Fazzari escribió:
>
> In snapcraft/internal/pluginhandler.py 
> <https://github.com/snapcore/snapcraft/pull/784#discussion_r78230329>:
>
> > +        sources = iglob(src, recursive=True)
> > +
> > +        for src in sources:
> > +            if os.path.isdir(src) and '*' not in key:
> > +                file_utils.link_or_copy_tree(src, dst)
> > +                # TODO create alternate organization location to avoid
> > +                # deletions.
> > +                shutil.rmtree(src)
> > +            elif os.path.isfile(dst):
> > +                raise EnvironmentError(
> > +                    'Trying to organize file {key!r} to {dst!r}, '
> > +                    'but {dst!r} already exists'.format(
> > +                        key=key, dst=os.path.relpath(dst, base_dir)))
> > +            else:
> > +                os.makedirs(os.path.dirname(dst), exist_ok=True)
> > +                shutil.move(src, dst)
>
> This may leave empty parent directories lying around (something that 
> is currently a problem). Could be fixed with a while loop looking for 
> empty parent directories and stopping at the installdir.
>

Aren't you afraid we might remove a legit directory? Wouldn't filesets 
do the job here?

kyrofa: I can't imagine a scenario where the empty parent directory would be desired, but I easily run into not wanting it (and indeed have). Say I have the following layout:

```
src/foo/configs/my.conf
src/foo/scripts/start-my-thing
```

And my YAML:

```yaml
# [...]

parts:
    my-part:
        plugin: dump
        source: src/foo
        organize:
            configs/my.conf: my-conf-dir/my.conf
            scripts/start-my-thing: bin/start-my-thing
```

As things are now, I need to filter out the configs and scripts folder using the stage/snap keywords, when I feel like that's a problem we can solve automatically.

kyrofa: To be fair, not something that needs to be solved here. But if we agree that it's something we can solve, I'll log a bug for it.

kyrofa:
Looks good other than the empty parent directories, which doesn't need to be solved here.

snapcraft/internal/pluginhandler.py
sergiusens: El 09/09/16 a las 16:04, Kyle Fazzari escribió:
>
> In snapcraft/internal/pluginhandler.py 
> <https://github.com/snapcore/snapcraft/pull/784#discussion_r78232388>:
>
> > +        sources = iglob(src, recursive=True)
> > +
> > +        for src in sources:
> > +            if os.path.isdir(src) and '*' not in key:
> > +                file_utils.link_or_copy_tree(src, dst)
> > +                # TODO create alternate organization location to avoid
> > +                # deletions.
> > +                shutil.rmtree(src)
> > +            elif os.path.isfile(dst):
> > +                raise EnvironmentError(
> > +                    'Trying to organize file {key!r} to {dst!r}, '
> > +                    'but {dst!r} already exists'.format(
> > +                        key=key, dst=os.path.relpath(dst, base_dir)))
> > +            else:
> > +                os.makedirs(os.path.dirname(dst), exist_ok=True)
> > +                shutil.move(src, dst)
>
> To be fair, not something that needs to be solved here. But if we 
> agree that it's something we can solve, I'll log a bug for it.
>

Please do!

kyrofa: Done, LP: [#1621980](https://bugs.launchpad.net/snapcraft/+bug/1621980).

PR #785, opened 2016-09-09
Check if option is url for pip

SamYaple:
Requirements and Constraints file can be a local file or a url. Add
a simple check to use the appropriate path

LP: #1621543

### DIFF
--- a/snapcraft/plugins/python.py
+++ b/snapcraft/plugins/python.py
@@ -53,6 +53,7 @@ from glob import glob
 
 import snapcraft
 from snapcraft import file_utils
+from snapcraft.common import isurl
 
 
 class PythonPlugin(snapcraft.BasePlugin):
@@ -165,10 +166,12 @@ class PythonPlugin(snapcraft.BasePlugin):
                        '--disable-pip-version-check']
 
         if self.options.constraints:
-            pip_install = pip_install + [
-                '--constraint',
-                os.path.join(self.sourcedir, self.options.constraints),
-            ]
+            if isurl(self.options.constraints):
+                constraints = self.options.constraints
+            else:
+                constraints = os.path.join(self.sourcedir,
+                                           self.options.constraints)
+            pip_install = pip_install + ['--constraint', constraints]
 
         if self.options.process_dependency_links:
             pip_install.append('--process-dependency-links')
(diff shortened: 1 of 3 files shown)


snappy-m-o:

snappy-m-o:

snapcraft/plugins/python.py
```diff
@@ -205,11 +210,7 @@ def _run_pip(self, setup):
             self.run(pip_install + ['.'], cwd=cwd, env=env)
 
     def _fix_permissions(self):
-        for root, dirs, files in os.walk(self.installdir):
-            for filename in files:
-                _replicate_owner_mode(os.path.join(root, filename))
-            for dirname in dirs:
-                _replicate_owner_mode(os.path.join(root, dirname))
```
SamYaple: not sure why this got removed. probably bad rebase. will address

snapcraft/plugins/python.py
```diff
@@ -252,3 +253,6 @@ def _replicate_owner_mode(path):
     if file_mode & stat.S_IRUSR:
         new_mode |= stat.S_IRUSR | stat.S_IRGRP | stat.S_IROTH
     os.chmod(path, new_mode)
+
+def is_str_url(s):
```
sergiusens: instead of this `from snapcraft.common import isurl` and use that

snapcraft/plugins/python.py
```diff
@@ -192,10 +195,12 @@ def _run_pip(self, setup):
         env = self._get_build_env()
 
         if self.options.requirements:
-            self.run(pip_install + [
-                '--requirement',
-                os.path.join(self.sourcedir, self.options.requirements),
-            ], env=env)
+            if is_str_url(self.options.requirements):
```
sergiusens: instead of this `from snapcraft.common import isurl` and use that

snapcraft/plugins/python.py
```diff
@@ -165,10 +166,12 @@ def _get_pip_command(self):
                        '--disable-pip-version-check']
 
         if self.options.constraints:
-            pip_install = pip_install + [
-                '--constraint',
-                os.path.join(self.sourcedir, self.options.constraints),
-            ]
+            if is_str_url(self.options.constraints):
```
sergiusens: instead of this `from snapcraft.common import isurl` and use that

snapcraft/plugins/python.py
```diff
@@ -45,11 +45,12 @@
       The python version to use. Valid options are: python2 and python3
 """
 
+from glob import glob
 import os
 import re
 import stat
 import subprocess
-from glob import glob
+import urlparse
```
sergiusens: instead of this `from snapcraft.common import isurl` and use that

snapcraft/plugins/python.py
SamYaple: ah nice. already a built in. will replace it

snapcraft/plugins/python.py
```diff
@@ -45,14 +45,15 @@
       The python version to use. Valid options are: python2 and python3
 """
 
+from glob import glob
 import os
 import re
 import stat
 import subprocess
-from glob import glob
```
sergiusens: why was this swapped to the top?

sergiusens:
ok to test

sergiusens:
Looks good except for the unnecessary move of the glob import

snapcraft/plugins/python.py
SamYaple: this was not intentional. not sure why it was moved. fixing

PR #790, opened 2016-09-10
Reduces download time of `git clone` fetching just a single branch.

ehbello:
This pull request speed up the pull time of a git repository only downloading the commits of related branch thanks to --single-branch switch of Git.

Signed-off-by: Enrique Hernández Bello <ehbello@gmail.com>

### DIFF
--- a/snapcraft/internal/sources.py
+++ b/snapcraft/internal/sources.py
@@ -181,7 +181,7 @@ class Git(Base):
         else:
             branch_opts = []
             if self.source_tag or self.source_branch:
-                branch_opts = ['--branch',
+                branch_opts = ['--single-branch', '--branch',
                                self.source_tag or self.source_branch]
             subprocess.check_call(['git', 'clone', '--depth', '1',
                                   '--recursive'] + branch_opts +
--- a/snapcraft/tests/test_sources.py
+++ b/snapcraft/tests/test_sources.py
@@ -196,16 +196,16 @@ class TestGit(SourceTestCase):
         git.pull()
 
         self.mock_run.assert_called_once_with(
-            ['git', 'clone', '--depth', '1', '--recursive', '--branch',
-             'my-branch', 'git://my-source', 'source_dir'])
+            ['git', 'clone', '--depth', '1', '--recursive', '--single-branch',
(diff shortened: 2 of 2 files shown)


elopio:
this makes sense to me. :+1:

kyrofa:
Thanks for the contribution @ehbello! However, we already use `--depth=1`, in which case I'm afraid using `--single-branch` adds nothing extra. I'm going to go ahead and close this, but please feel free to re-open if you disagree.

Thanks again!

elopio:
Hey @kyrofa, but we are going to make --depth optional. Take a look at https://bugs.launchpad.net/snapcraft/+bug/1617783

kyrofa:
@elopio yeah, but talking to @sergiusens, we're going to make it opt-out, not opt-in. Also, I suspect --single-branch will have the same (or similar) limitation as --depth with regards to that bug. If not we can perhaps use it when the user requests to opt out of --depth, but then we might continue playing whack-a-mole with our efficiency gains getting in the way of other weird build systems.

elopio:
sabdfl said the default must be without depth. So if you want to make it opt-out, there's some convincing to do :) 
I'm happy with it being opt-in because the semantics of the alternative are weirder. If the default is depth=1, how do you express an infinite depth? Of course, the depth is never too big, but having to use something like depth=10, or depth=100 feels wrong.

sergiusens:
El 19/09/16 a las 15:46, Leo Arias escribió:
>
> sabdfl said the default must be without depth. So if you want to make 
> it opt-out, there's some convincing to do :)
> I'm happy with it being opt-in because the semantics of the 
> alternative are weirder. If the default is depth=1, how do you express 
> an infinite depth? Of course, the depth is never too big, but having 
> to use something like depth=10, or depth=100 feels wrong.
>

`--single-branch` will bring in the same set of problems as using depth

 From the git manual:

            Create a shallow clone with a history truncated to the 
specified number of
            commits. Implies --single-branch unless --no-single-branch 
is given to fetch
            the histories near the tips of all branches.

We wouldn't be providing a depth=n entry but rather a source-shallow: 
true|false option

I will point this out in the bug.

PR #791, opened 2016-09-10
Enable crosscompilation for dump and copy plugins

ehbello:
This pull request enables crosscompilation for dump and copy plugins to be able to use them, for example, providing blobs after build a kernel snap for a different architecture.

### DIFF
--- a/snapcraft/tests/test_plugin_copy.py
+++ b/snapcraft/tests/test_plugin_copy.py
@@ -332,6 +332,10 @@ class TestCopyPlugin(TestCase):
                 with open(destination, 'r') as f:
                     self.assertEqual(f.read(), symlink['expected_contents'])
 
+    def test_copy_enable_cross_compilation(self):
+        c = CopyPlugin('copy', self.mock_options, self.project_options)
+        c.enable_cross_compilation()
+
 
 class TestRecursivelyLink(TestCase):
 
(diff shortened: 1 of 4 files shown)


snapcraft/tests/test_plugin_copy.py
```diff
@@ -332,6 +332,10 @@ def test_copy_symlinks_that_should_be_followed(self):
                 with open(destination, 'r') as f:
                     self.assertEqual(f.read(), symlink['expected_contents'])
 
+    def test_copy_enable_cross_compilation(self):
+        c = CopyPlugin('copy', self.mock_options, self.project_options)
+        c.enable_cross_compilation()
```
sergiusens: this test does not verify anything. Maybe call `.build()` after or add a code comment about the expected behavior

snapcraft/tests/test_plugin_dump.py
```diff
@@ -194,3 +194,7 @@ def test_dump_broken_symlink(self):
             str(raised.exception),
             '{!r} is a broken symlink pointing outside the snap'.format(
                 os.path.join(plugin.builddir, 'bad_relative')))
+
+    def test_dump_enable_cross_compilation(self):
+        plugin = DumpPlugin('dump', self.options, self.project_options)
+        plugin.enable_cross_compilation()
```
sergiusens: ditto

snapcraft/tests/test_plugin_copy.py
ehbello: This test verifies that this plugin does not stop when the target architecture is different from host architecture. Both plugins do the same on all architectures. The other features already have their own tests.

ehbello: ... so this test, simply checks that 'enable_cross_compilation' method don't raise any exception.

sergiusens:
ok to test

sergiusens:
LP: #1624544

PR #792, opened 2016-09-10
Fix `git clone` when transport is http and it does not support --depth

ehbello:
This pull request remove ``--depth=1`` of HTTP transport for Git due there are repositories as git.kernel.org that does not support that.

Signed-off-by: Enrique Hernández Bello <ehbello@gmail.com>

### DIFF
--- a/snapcraft/internal/sources.py
+++ b/snapcraft/internal/sources.py
@@ -183,9 +183,14 @@ class Git(Base):
             if self.source_tag or self.source_branch:
                 branch_opts = ['--branch',
                                self.source_tag or self.source_branch]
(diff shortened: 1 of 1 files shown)


elopio:
Please take a look at https://bugs.launchpad.net/snapcraft/+bug/1617783
Instead of just removing it, we want to make it optional.
Also, every change we land requires unit tests, take a look in snapcraft/tests.

Thank you!

sergiusens:
We are removing `--depth=1` in general and making it **opt-in** in #814 
Thank you for your time and sorry for taking so long to get back to you.

PR #793, opened 2016-09-12
Unify python plugin

SamYaple:


snapcraft/plugins/python.py
```diff
@@ -45,14 +45,15 @@
       The python version to use. Valid options are: python2 and python3
```
SamYaple: ignore this whole commit, its only here to allow me to work on, and test, this change without an ugly rebase when this patch merges

SamYaple:
I think it is funny and want to point this out: https://coveralls.io/builds/7852962

All edited files are 100% coveraged covered, but the overall percentage of coverage has fallen. This check would pass if snapcraft was 100% coveraged. thats pretty funny to me

sergiusens:
It is funny, I am not a fan and we can ask @elopio (and he will say its to find dead test code most likely). But the fact that you unified tests means the total lines of code covered has reduced making the 100% mark harder to reach while still having the same amount of untested code.

SamYaple:
Eh, I could always add in some filler code to make up some more lines...... :)

I am going to be working on a few more tests for the plugin in the future. Do proper unit tests against each function and mock all the rest. As it is right now, we are doing _ok_ with the unit tests, but I feel they are more busy work to get coveralls to pass rather than useful.

I don't plan on doing that for this commit though.

Once https://github.com/snapcore/snapcraft/pull/785 merges, I will rebase this and we can discuss what to do here.

snappy-m-o:

snappy-m-o:

SamYaple:
@elopio Over all coverage percentage has dropped due to the deduplication of the test suites, while code coverage has actually risen (in this case, they python plugins are now 100% coveraged). I believe that means we need to force merge this and ignore coveralls.

Thoughts?

sergiusens:
oh, I intend to ignore it indeed!

snapcraft/tests/test_plugin_python.py
```diff
@@ -0,0 +1,249 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2015 Canonical Ltd
```
sergiusens: date needs fixing here

SamYaple:
Heh. ok.

Well the plugin is coming along quite nicely I think. It has (i believe) every feature I want now. I am goign to iron out the test suite a bit more and refactor a few sections, but functionally its solid for me now.

snapcraft/tests/test_plugin_python.py
SamYaple: how do you want it fixed? 2015-2016, or 2015, 2016

sergiusens: I personally prefer 2015-2016 as then it is just a one digit change to maintain, but I don't mind either.

SamYaple: i do as well. it shall be so

sergiusens:
Is there a bug for this?

SamYaple:
Not that I know of. Is that a hard requirement?

sergiusens:
LP: #1623289

SamYaple:
thanks

PR #795, opened 2016-09-13
WIP: snap-build generation and pushing support

caio1982:


snapcraft/_store.py
```diff
@@ -184,6 +183,30 @@ def register(snap_name, is_private=False):
...
+    store = storeapi.StoreClient()
+    account_info = store.get_account_information()
+
+    snap_yaml = _get_data_from_snap_file(snap_filename)
+    snap_name = snap_yaml['name']
+    grade = snap_yaml['grade']
+
+    with _requires_login():
+        store.sign_build(
+            account_info, snap_name, snap_filename, grade, key_name, local)
```
cprov: The specific part of the procedure that requires Store login is `push_snap_build()`. The way it is implemented, it would force users to login on the store even if `--local` is passed.
Can you call push_snap_build() here, only if --local is not set ?

snapcraft/storeapi/__init__.py
```diff
@@ -387,6 +399,63 @@ def snap_release(self, snap_name, revision, channels):
 
         return response_json
 
+    def generate_snap_build(self, account_info, snap_name,
```
cprov: This method doesn't belong to SCA layer, IMO. It lands and sign an assertion on disk and never talk to the Store.

snapcraft/storeapi/__init__.py
```diff
@@ -387,6 +399,63 @@ def snap_release(self, snap_name, revision, channels):
...
+        logger.info('Assertion {} saved to disk.'.format(assertion_file))
+
+    def push_snap_build(self, account_info, snap_name, snap_filename):
+        assertion_file = sub(r"(.*).snap$", r"\1.snap-build", snap_filename)
+        try:
+            snaps = account_info['snaps'][constants.DEFAULT_SERIES]
+            snap_id = snaps[snap_name]['snap-id']
+        except KeyError as err:
+            msg = 'Could not get all account information needed'
+            raise snapcraft.internal.meta.CommandError(msg)
```
cprov: This block is repeated. I'd make it `push_snap_build(self, snap_id, assertion)` and extract the inputs on the callsite.

snapcraft/tests/store/test_store_client.py
```diff
@@ -153,6 +153,15 @@ def test_download_with_invalid_credentials_raises_exception(self):
...
                 'test-snap', 'test-channel', download_path)
 
+class SignBuildTestCase(tests.TestCase):
+    def setUp(self):
+        super().setUp()
+        self.useFixture(fixture_setup.FakeStore())
+        self.client = storeapi.Storeclient()
+
+    def test_sign_build(self):
+        pass
```
cprov: uhm, needs implementation.

snapcraft/tests/test_commands_sign_build.py
```diff
@@ -0,0 +1,66 @@
...
+        mock_get_account_information.return_value = {'account_id': 'abcd'}
+
+        main(['init'])
+        main([])
+        #self.assertEqual("XXXXX", self.fake_logger.output)
+
+        # create a sample snap and then sign its build
+        main(['sign-build', 'my-snap-name_0.1_amd64.snap'])
+
+        self.assertEqual("XXXXX", self.fake_logger.output)
```
cprov: ditto.

snapcraft/_store.py
caio1982: I could but giving it another look a few lines above store.get_account_information() also needs a store login, and it is used by both generate and sign, even if --local is set (because we need the snap ID to call `snap sign-build`) :-(

snapcraft/tests/store/test_store_client.py
caio1982: Yes, tests are pretty much absent right now...

snapcraft/storeapi/__init__.py
caio1982: I will try to simplify it, ack.

snapcraft/storeapi/__init__.py
caio1982: See the need of the account info inside generate_snap_build(), though I will see how I can refactor it.

snapcraft/_store.py
cprov: Right, account-info requires login, which means it can't ever create and sign a snap-build w/o talking to the Store. This is less than ideal and will have to fixed eventually (caching it locally) and therefore, should not 'leak' to `generate_snap_build()` or `push_snap_build()`. Let's try to solve this here and pass specific info down.

sergiusens:
El 21/09/16 a las 11:32, Celso Providelo escribió:
>
> *@cprov* commented on this pull request.
>
>
> In snapcraft/_store.py <https://github.com/snapcore/snapcraft/pull/795>:
>
> > +    if not os.path.exists(snap_filename):
> +        raise FileNotFoundError(
> +            'The file {!r} does not exist.'.format(snap_filename))
> +
> +    # we pass account_info to sign_build so we call it only once
> +    # for both generating/signing and pushing the assertion data
> +    store = storeapi.StoreClient()
> +    account_info = store.get_account_information()
> +
> +    snap_yaml = _get_data_from_snap_file(snap_filename)
> +    snap_name = snap_yaml['name']
> +    grade = snap_yaml['grade']
> +
> +    with _requires_login():
> +        store.sign_build(
> +            account_info, snap_name, snap_filename, grade, key_name, local)
>
> Right, account-info requires login, which means it can't ever create 
> and sign a snap-build w/o talking to the Store. This is less than 
> ideal and will have to fixed eventually (caching it locally) and 
> therefore, should not 'leak' to |generate_snap_build()| or 
> |push_snap_build()|. Let's try to solve this here and pass specific 
> info down.
>

+1, this was a sad oversight from us.

snapcraft/_store.py
cjwatson: get_account_information shouldn't require a *recent* login, though; the credentials are cached.

snapcraft/_store.py
```diff
@@ -190,6 +190,41 @@ def register(snap_name, is_private=False):
...
+
+    store = storeapi.StoreClient()
+    with _requires_login():
+        try:
+            info = store.get_account_information()
+            authority_id = info['account_id']
+            snap_id = info['snaps'][snap_series][snap_name]['snap-id']
+        except KeyError:
+            raise RuntimeError(
+                'Your account lacks information to assert this snap.')
```
cjwatson: In various places in this PR you talk about asserting a snap, asserting a built snap file, and asserting a build.  I think it would be worth settling on a consistent vocabulary.

snapcraft/tests/test_commands_sign_build.py
```diff
@@ -0,0 +1,99 @@
...
+                              mock_get_account_information,
+                              mock_register_key):
+        account_info = {'account_id': 'abcd',
+                        'account_keys': [
+                            {
+                                'name': 'default',
+                                'public-key-sha3-384': (
+                                    get_sample_key('default')['sha3-384']),
+                            },
+                        ],
```
cjwatson: I'd be inclined to just omit `account_keys` from this sample input data, since it's quite verbose and you don't need it.

snapcraft/tests/test_commands_sign_build.py
```diff
@@ -0,0 +1,99 @@
...
+                            'basic': {
+                                'snap-id': 'snap-id',
+                                }
+                            }
+                        }}
+
+        mock_installed.return_value = True
+        mock_input.side_effect = ['sample.person@canonical.com', '123456']
+        mock_getpass.return_value = 'secret'
+        mock_check_output.side_effect = mock_snap_output
```
cjwatson: Do you actually need this?  `mock_snap_output` only handles `snap keys --json` and `snap export-key ...`, neither of which seems relevant to `sign-build`.  I guess that's just because you're registering a key below, but that also seems somewhat irrelevant in a unit test of `sign-build`'s behaviour.

It will probably simplify your life if you mock `snap sign-build` independently (i.e. by writing your own version of `mock_snap_output` rather than modifying the one in `snapcraft.tests.test_commands_register_key`), and make your unit test independent of key registration.

integration_tests/test_store_sign_build.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def test_successful_sign_build_local(self):
+        project_dir = 'basic'
+        self.addCleanup(self.logout)
+        self.login()
+
+        self.run_snapcraft('snap', project_dir)
+        os.chdir(project_dir)
+        self.assertThat(self.snap_path, FileExists())
+        self.assertEqual(0, self.register_key('default'))
```
cjwatson: This shouldn't be needed in the `--local` case.

snapcraft/storeapi/__init__.py
```diff
@@ -432,6 +449,42 @@ def snap_release(self, snap_name, revision, channels):
...
         return response_json
 
+    def generate_snap_build(self, authority_id, snap_id, snap_name,
+                            snap_filename, grade, key_name, assertion):
+        with open(assertion, 'w+') as outfile:
+            try:
+                # snap-digest is calculated by snap sign-build itself
+                # snap-size is calculated by snap sign-build itself
+                # timestamp is taken by snap sign-build itself
+                cmd = ['snap', 'sign-build',
```
cjwatson: You'll need to update a few package relationships to refer to `snapd (>= 2.15~)` rather than `snapd (>= 2.14.2~)` in order to use this: at least the one in `debian/control` and the first one in `debian/tests/control`, and possibly the second in that latter file as well.

snapcraft/tests/test_commands_sign_build.py
caio1982: Yeah, I figured that after checking it with @cprov. I was under the impression I really needed to get register-key to work first, since I was afraid of mocking too much of `sign-build` and ending up with a bogus test. Thanks for the rest of the review btw!

snapcraft/_store.py
```diff
@@ -190,6 +189,73 @@ def register(snap_name, is_private=False):
...
+        except KeyError:
+            raise RuntimeError(
+                ('Your account lacks information to sign '
+                 'build assertions for this snap.'))
+
+    snap_build_path = snap_filename + '-build'
+    # XXX cprov 20160922: the local cache/check is needed here
+    # to figure out if the snap-build on disk (named by version)
+    # corresponding to the existing snap blob.
+    if not os.path.isfile(snap_build_path):
```
cprov: This point has to be better addressed, the current code will push the wrong assertion if the snap was just rebuilt. Without processing the assertion and checking sha3-384, it can only be done by caching the assertion on `~/.config/snapcraft/` as Mark suggested. Even with the local cache, things might get confusing on 3rd-part builds.

snapcraft/tests/fake_servers.py
```diff
@@ -378,6 +382,38 @@ def _handle_acl_request(self, permission):
...
+        if snap_build == 'test-not-implemented':
+            self.send_response(501)
+            self.send_header('Content-Type', 'text/plain')
+            content = b'Not Implemented'
+        elif snap_build == 'test-invalid-data':
+            self.send_response(400)
+            self.send_header('Content-Type', 'application/json')
+            error = {
+                'error_list': [
+                    # XXX cprov 20160922: fix endpoint error handler.
```
cprov: This needs fix server side, the current error on 400 is super dry.

snapcraft/_store.py
```diff
@@ -32,7 +32,7 @@
 logger = logging.getLogger(__name__)
 
 
-def _get_name_from_snap_file(snap_path):
+def _get_data_from_snap_file(snap_path):
```
cprov: This change affects push & sign-build, so it's covered indirectly, however I think it deserves a unittest (and dropping the undescore)

integration_tests/test_store_sign_build.py
cprov: Also, please merge https://github.com/caio1982/snapcraft/pull/4

cprov:
@elopio, @sergiusens this PR is superseded by https://github.com/snapcore/snapcraft/pull/831.

Caio is on leave this week, I am assuming this change.

PR #799, opened 2016-09-14
Load project information in one location

sergiusens:
Make sure the full project can be resolved before loading
the plugins from the parts

LP: #1622435

### DIFF
--- a/integration_tests/test_parts.py
+++ b/integration_tests/test_parts.py
@@ -61,3 +61,39 @@ class PartsTestCase(integration_tests.TestCase):
             "  - -share\n")
 
         self.assertEqual(expected, output)
+
+
+class PartsWithFilesetsTestCase(integration_tests.TestCase):
+
+    def test_part_with_fileset(self):
+        self.run_snapcraft('update')
+
+        output = self.run_snapcraft(['define', 'simple-make-filesets'])
+
+        expected = (
+            "Maintainer: 'Jonathan Cave <jonathan.cave@canonical.com>'\n"
+            "Description: 'The filesets test from the integration test suite.'"
+            "\n\n"
+            "simple-make-filesets:\n"
+            "  filesets:\n"
+            "    files:\n"
+            "    - share/file1\n"
+            "    - share/file2\n"
+            "  organize:\n"
+            "    dir1: new/dir1\n"
+            "    dir2: new/dir2\n"
+            "    file1: share/file1\n"
+            "    file2: share/file2\n"
+            "  plugin: make\n"
+            "  snap:\n"
+            "  - -new/dir1\n"
+            "  source: https://github.com/jocave/simple-make-filesets.git\n"
+            "  stage:\n"
+            "  - $files\n"
+            "  - new/dir1\n"
+            "  - new/dir2\n")
+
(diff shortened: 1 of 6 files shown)


snapcraft/internal/project_loader.py
```diff
@@ -351,3 +390,22 @@ def _ensure_grade_default(yaml_data, schema):
...
+    fileset_for_step = properties.get(step, {})
+    new_step_set = []
+
+    for item in fileset_for_step:
+        if item.startswith('$'):
+            try:
+                new_step_set.extend(filesets[item[1:]])
+            except KeyError:
+                raise SnapcraftLogicError(
+                    '\'{}\' referred to in the \'{}\' fileset but it is not '
```
josepht: I know this was copied over, but should we replace \'{}\' with {!r}?

sergiusens: El 14/09/16 a las 10:14, Joe Talbott escribió:
>
> In snapcraft/internal/project_loader.py 
> <https://github.com/snapcore/snapcraft/pull/799#discussion_r78745088>:
>
> > @@ -351,3 +390,22 @@ def _ensure_grade_default(yaml_data, schema):
> >          logger.warning('"grade" property not specified: defaulting '
> >                         'to "stable"')
> >          yaml_data['grade'] = schema['grade']['default']
> > +
> > +
> > +def _expand_filesets_for(step, properties):
> > +    filesets = properties.get('filesets', {})
> > +    fileset_for_step = properties.get(step, {})
> > +    new_step_set = []
> > +
> > +    for item in fileset_for_step:
> > +        if item.startswith('$'):
> > +            try:
> > +                new_step_set.extend(filesets[item[1:]])
> > +            except KeyError:
> > +                raise SnapcraftLogicError(
> > +                    '\'{}\' referred to in the \'{}\' fileset but it is not '
>
> I know this was copied over, but should we replace \'{}\' with {!r}?
>
Yeah, good point.

josepht:
Looks good to me.

integration_tests/test_parts.py
```diff
@@ -61,3 +61,39 @@ def test_curl_exists_and_properly_defined(self):
...
+            "  - new/dir1\n"
+            "  - new/dir2\n"
+            "  snap:\n"
+            "  - -new/dir1\n"
+            "  organize:\n"
+            "    file1: share/file1\n"
+            "    file2: share/file2\n"
+            "    dir1: new/dir1\n"
+            "    dir2: new/dir2\n"
+            "  source: https://github.com/jocave/simple-make-filesets.git\n")
```
kyrofa: This seems super fragile, no? One change on the wiki and our tests break.

sergiusens: Yes, as most of the wiki tests.

sergiusens: but as an integration test we need to talk to the wiki; same is valid for talking to the store, one bad upload and many tests would break (/me looks at hello-world)

snapcraft/internal/project_loader.py
sergiusens: I am going to ignore this one for now and rework these logic errors as a whole into `snapcraft.internal.errors`

PR #800, opened 2016-09-14
LP: #1623509 nodejs: fix to install dev-deps in build phase

jonathon-love:
this PR makes the nodejs plugin download the devDependencies from package.json during the build phase.

https://bugs.launchpad.net/snapcraft/+bug/1623509

i have submitted the contributor agreement as well.

### DIFF
--- a/snapcraft/plugins/nodejs.py
+++ b/snapcraft/plugins/nodejs.py
@@ -113,11 +113,12 @@ class NodePlugin(snapcraft.BasePlugin):
     def _npm_install(self):
         self._nodejs_tar.provision(
             self.installdir, clean_target=False, keep_tarball=True)
-        npm_install = ['npm', '--cache-min=Infinity', 'install', '--global']
+        npm_install = ['npm', '--cache-min=Infinity', 'install']
         for pkg in self.options.node_packages:
-            self.run(npm_install + [pkg])
+            self.run(npm_install + ['--global'] + [pkg])
         if os.path.exists(os.path.join(self.builddir, 'package.json')):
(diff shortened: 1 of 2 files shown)


snapcraft/plugins/nodejs.py
```diff
@@ -117,7 +117,8 @@ def _npm_install(self):
         for pkg in self.options.node_packages:
             self.run(npm_install + [pkg])
         if os.path.exists(os.path.join(self.builddir, 'package.json')):
-            self.run(npm_install)
+            self.run(['npm', '--cache-min=Infinity', 'install'])
```
sergiusens: Maybe make 
```python
npm_install = ['npm', '--cache-min=Infinity', 'install']
```

Then

```python
self.run(npm_install +  ['--global'] + [pkg])
...
self.run(npm_install)
self.run(npm_instal +  ['--global'])
```

jonathon-love:
done!

snapcraft/plugins/nodejs.py
jonathon-love: done!

sergiusens:
OK to test

jonathon-love:
hi,

is there a trick to accessing the results of those (failing) 3 tests? clicking 'Details' just times out.

perhaps it is a temporary outage(?), i will try again later.

with thanks

sergiusens:
You cannot see them as they run on our special infra. But the failures are not related to your PR

jonathon-love:
righto. so this doesn't require any more attention from me?

PR #801, opened 2016-09-14
In the busybox test, use the last installed data dir

elopio:
LP: #1623621


elopio:
@sergiusens yes, jenkins approves. 

FAILED (failures=1)

We are down to one: https://bugs.launchpad.net/snapcraft/+bug/1623623

PR #803, opened 2016-09-14
Do not run the bootstrap directory as a script (autotools plugin).

evandandrea:
Sometimes an autotools-using project has a bootstrap directory, such as Basho's fork of Erlang (https://github.com/basho/otp) as used in Riak.


snapcraft/plugins/autotools.py
```diff
@@ -98,6 +98,8 @@ def build(self):
                 path = os.path.join(self.builddir, script)
                 if not os.path.exists(path):
                     continue
+                if os.path.isdir(path):
```
sergiusens: I would just put this in the `if` above but not going to be dramatic about it ;-)

sergiusens:
ok to test

PR #805, opened 2016-09-16
Refresh discharge macaroon if necessary

cjwatson:
"WWW-Authenticate: Macaroon needs_refresh=1" tells us that we need to
ask SSO to refresh our discharge macaroon, so do that.

LP: #1621171

### DIFF
--- a/snapcraft/storeapi/__init__.py
+++ b/snapcraft/storeapi/__init__.py
@@ -357,8 +357,8 @@ class SCAClient(Client):
     def _is_needs_refresh_response(response):
         return (
             response.status_code == requests.codes.unauthorized and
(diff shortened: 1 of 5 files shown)


sergiusens:
ok to test

snapcraft/storeapi/__init__.py
```diff
@@ -149,14 +149,27 @@ def logout(self):
         self.conf.clear()
         self.conf.save()
 
+    def _refresh_if_necessary(self, func, *args, **kwargs):
```
sergiusens: not a fan of this name; but can't think anything which wouldn't be a full blown sentence.

PR #807, opened 2016-09-16
Fix parts integration tests

cjwatson:
Insulate the parts integration tests against trivial differences in YAML
dump ordering of dictionaries.

### DIFF
--- a/integration_tests/test_parts.py
+++ b/integration_tests/test_parts.py
@@ -16,6 +16,9 @@
 
 import os
 
+from testtools.matchers import StartsWith
+import yaml
+
 import integration_tests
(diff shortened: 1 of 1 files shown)


josepht:
This is a result of our using OrderedDict for everything.  I had to carefully craft expected parts to get them to match.  I think for most tests other than those explicitly testing the ordering we could convert the OrderedDicts to regular dicts and compare those.

cjwatson:
Yeah, but in this case we're comparing the serialised output.  I don't think the integration test actually conceptually cares about the ordering, so it might as well just load and compare dictionaries.

PR #810, opened 2016-09-18
nodejs plugin: Add mechanism to run “npm run” commands.

RAOF:
Some complex projects don't install with “npm install”; rather they use “npm run” commands to drive their build, assembly, and packaging.

This support is required to build a snap from (at least) https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-

### DIFF
--- a/integration_tests/snaps/nodejs-with-run-commands/snapcraft.yaml
+++ b/integration_tests/snaps/nodejs-with-run-commands/snapcraft.yaml
@@ -10,4 +10,6 @@ parts:
     plugin: nodejs
     source: .
     npm-run:
-        - run-command
+        - run-command-one
+        - run-command-two
+        - run-command-three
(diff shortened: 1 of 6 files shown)


elopio:
\o/ Thanks @RAOF. This is needed for snapping vector.im, and I think also the rocket guys were doing something weird because of this.

integration_tests/snaps/nodejs-with-run-commands/snapcraft.yaml
```diff
@@ -0,0 +1,15 @@
+name: test-package
+version: 0.1
+summary: A simple nodejs project.
+description: |
+  Proves that it is possible to build a snap using local nodejs sources.
```
elopio: can you please update this description to say something about the run tasks?

snapcraft/plugins/nodejs.py
```diff
@@ -31,6 +31,9 @@
     - node-engine:
       (string)
       The version of nodejs you want the snap to run on.
+    - npm-run:
+      (list)
+      A list of targets to `npm run`
```
elopio: Would it be useful to explain here that the run tasks are executed after npm install? I'm a n00b, so I don't know if that's obvious.

sergiusens:
ok to test

PR #812, opened 2016-09-19
New plugin: jhbuild

attente:


kyrofa:
Hey @attente! You could probably use some test coverage, here :wink: .

attente:
Ah, of course. On the way :)

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,1035 @@
...
+JHBuild is a tool used to build components of the GNOME ecosystem from version
+control. See the JHBuild manual for more details:
+https://developer.gnome.org/jhbuild/stable/.
+
+The plugin can be customized with the following keys:
+
+  - modules (required): a list of modules to include in the part, modules that
+                        should be skipped should be prefixed with '-'
+  - moduleset: the module set JHBuild should use (default: gnome-world)
+  - snap-name (required): the name of the snap this part is included in
```
elopio: This will always be $SNAPCRAFT_PROJECT_NAME. We probably can drop this attribute.

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,1035 @@
...
+  - snap-revision: the revision of the snap this part is included in
```
elopio: same as above, for $SNAPCRAFT_PROJECT_VERSION.

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,1035 @@
...
+                   (default: current)
+  - container-name: the name of the LXC container used for building
+                    (default: snapcraft-<snap name>-<part name>)
+  - distribution: the distribution the LXC container is based off of
+                  (default: host distribution)
+  - release: the release of the distribution the LXC container is based off of
+             (default: host release)
+  - architecture: the architecture of the LXC container
+                  (default: host architecture)
```
elopio: I like that you can select the image you want, but I think this should be more general, not just for this plugin.
We had https://bugs.launchpad.net/snapcraft/+bug/1597952 for this.

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,1035 @@
...
+  - container-packages: a list of build packages the LXC container should be
+                        bootstrapped with
+  - debmirror: the mirror to use for downloading .deb packages
+  - jhtarballs: the directory on the host containing source tarballs
+  - jhmirror: the directory on the host containing DVCS mirror repositories
+  - jhcheckout: the directory on the host container DVCS checkout repositories
+  - ccache: the directory on the host to use for ccache caching
+  - disable-pull: True to skip the pull stage (default: False)
+  - disable-build: True to skip the build stage (default: False)
```
elopio: why would you like to disable pull or build?

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,1035 @@
...
+
+  - Building WebKit requires a lot of time and computing power. If WebKit is a
+    dependency of your JHBuild module, it can be skipped by adding '-WebKit' to
+    the list of modules, and adding 'libwebkit2gtk-4.0-dev' to the list of
+    container-packages.
+
+  - Using the debmirror option in combination with a local apt-cacher-ng
+    installation can prevent repeated downloading of .deb packages. Use the IP
+    address of your lxcbr0 interface with the port number apt-cacher-ng uses.
+    For example, debmirror: http://10.0.3.1:3142
```
elopio: https://bugs.launchpad.net/snapcraft/+bug/1602085
Here again, I think it should be general, not only for this plugin.

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,1035 @@
...
+
+  - Specify directories on your host for jhtarballs and jhmirror to prevent
+    repeated downloading of the JHBuild module sources. It is best to reserve
+    directories on your local machine that can be used for all snaps you might
+    want to build.
```
elopio: We have caching for debs. This makes me wonder if we should offer a simple and generic caching solution for all plugins. It might be a mess, though.

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,1035 @@
...
+
+  - If you are debugging your snap, but don't need to rebuild the sources (for
+    example, to determine the list of plugs required under confinement), use
+    the disable-pull, disable-build, and disable-jhupdate options.
```
elopio: Ah, I see. But the problem here IMO is that if you want to debug your snap, you usually do it outside of the lxc cleanbuild. So you run pull once, and the next time it will be skipped.
I think we need some cool design to allow us to keep and reuse a lxc machine for cleanbuild, so we could do something similar.

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,1035 @@
...
+
+  - To add a shell for debugging your snap, add an extra app whose command is:
+    jhbuild -f $SNAP/etc/jhbuildrc shell. You might also want to add gdb,
+    strace, etc. to your list of stage-packages for debugging purposes.
```
elopio: I must say that this section for tips and tricks is great.

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,1035 @@
...
+STAGE_PACKAGES = [
+    'python',
+    'ttf-ubuntu-font-family',
+    'adwaita-icon-theme-full',
+]
+
+# build dependencies that JHBuild doesn't know about
+MODULE_EXCEPTIONS = {
+    'vala':                                                     ['valac'],
+    'gst-plugins-bad':                                          ['libgl1-mesa-dev'],
```
elopio: I don't think it's common in python to use spaces like this, to format a table. I would prefer just one space between the : and the element. You might even get pep8 errors for this, I'm not sure.

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,1035 @@
...
+
+        self.stage_packages += STAGE_PACKAGES
+
+    @property
+    def modules(self):
+        return [module for module in self.options.modules if not module.startswith('-')]
+
+    @property
+    def skipped_modules(self):
+        return [module[1:] for module in self.options.modules if module.startswith('-')]
```
elopio: I might be slow, but why are you removing the first element?

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,1035 @@
...
+        """
+        return self.options.container_name or 'snapcraft-%s-%s' % (self.options.snap_name, self.name)
+
+    @property
+    def src_jhtarballs(self):
+        """Get the absolute path to the JHBuild tarballs on the host.
+
+        /<path to snapcraft.yaml>/parts/<part name>/src/jhtarballs
+        """
+        return self.options.jhtarballs or os.path.join(self.sourcedir, 'jhtarballs')
```
elopio: wouldn't this be better in the init?

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,1035 @@
...
+        return os.path.join(self.prefix, 'etc', 'jhbuildrc')
+
+    # TODO: use custom /etc/fonts/fonts.conf
+    @property
+    def xdgdatahome(self):
+        """Get the absolute path to XDG_DATA_HOME in the build container.
+
+        /snap/<snap name>/<snap revision>/usr/share
+        """
+        return os.path.join(self.prefix, 'usr', 'share')
```
elopio: We usually define the attributes in the init, not as properties. Because then the statement is executed only once, not every time the attribute is accessed.

snapcraft/plugins/jhbuild.py
attente: Is there a way to obtain that value in the plugin though?

snapcraft/plugins/jhbuild.py
attente: Same question as above.

snapcraft/plugins/jhbuild.py
attente: It's just to remove the initial '-' that's present for the skipped modules, like '-Webkit' -> 'WebKit'.

snapcraft/plugins/jhbuild.py
attente: Sure.

snapcraft/plugins/jhbuild.py
attente: Sure.

snapcraft/plugins/jhbuild.py
attente: Right, I'm just not sure how to make the LXC container grab the debs that Snapcraft already caches. If there's a way, I'd gladly prefer to use that instead.

snapcraft/plugins/jhbuild.py
attente: No problem.

attente:
Superceded by https://github.com/snapcore/snapcraft/pull/924 which doesn't require the use of LXC any more.

PR #813, opened 2016-09-19
"snapcraft validate" and "snapcraft gated" commands

ralsina:
Implements snapcraft validate and gated commands, according to discussions and specs.


runtests.sh
```diff
@@ -49,7 +49,7 @@ python3 -m coverage && coverage="true"
 
 run_static_tests(){
     SRC_PATHS="bin snapcraft integration_tests snaps_tests"
-    python3 -m flake8 --max-complexity=10 $SRC_PATHS
+    python3 -m flake8 --max-complexity=12 $SRC_PATHS
```
elopio: hey, that's cheating! :)

snapcraft/_store.py
```diff
@@ -281,3 +282,83 @@ def download(snap_name, channel, download_path, arch):
...
+    details_url = search_results['_links']['self']['href']
+    details = store.cpi.get(details_url).json()
+    if details.get('result') == 'error':
+        raise RuntimeError(
+            'Error obtaining details for snap {}: {}'.format(
+                name, ', '.join(details['errors'])))
+    return details
+
+
+def gated(snap_name):
```
elopio: I would call this: print_gated_snaps.

sergiusens: or `show_gated`

snapcraft/_store.py
```diff
@@ -281,3 +282,82 @@ def download(snap_name, channel, download_path, arch):
         raise RuntimeError(
             'Failed to download {} at {} (mismatched SHA)'.format(
                 snap_name, download_path))
+
+
+def _get_snap_data(name, store):
+    search_results = store.cpi.search_package(name, 'stable', None)
+    if search_results is None:
```
sergiusens: is there a scenario where another zero value would make `if not search_results` not a desirable way to write this?

snapcraft/_store.py
```diff
@@ -281,3 +282,82 @@ def download(snap_name, channel, download_path, arch):
...
+            'approved-snap-id': approved_data['snap_id'],
+            'approved-snap-revision': rev,
+            'timestamp': datetime.datetime.utcnow().isoformat() + 'Z',
+            'valid': "true"
+        }
+        if revoke:
+            assertion['valid'] = "false"
+
+        cmdline = ['snap', 'sign']
+        if key is not None:
```
sergiusens: `if key:` reads better here, is there a possibility for it to be a zero value?

snapcraft/storeapi/__init__.py
```diff
@@ -437,6 +446,41 @@ def snap_release(self, snap_name, revision, channels):
...
+            'snaps/{}/validations'.format(snap_id), data=json.dumps(data),
+            headers={'Authorization': auth,
+                     'Content-Type': 'application/json',
+                     'Accept': 'application/json'})
+        if not response.ok:
+            raise errors.StoreValidationError(snap_id, response)
+        try:
+            response_json = response.json()
+        except (AttributeError, JSONDecodeError):
+            raise errors.StoreValidationError(snap_id, response)
```
sergiusens: what would this error look like?

runtests.sh
ralsina: Sorry, that was just temporary while I figure out how to make main have less than a 12 complexity with 12 commands, adding a XXX comment :-)

snapcraft/_store.py
ralsina: I was going with the convention of functions being called the same as the command. If you guys like that better, no problem for me.

snapcraft/_store.py
ralsina: No such scenario. I sorta like this better since None is what the called function returns, but am ok with changing it :-)

snapcraft/_store.py
ralsina: There is a remote chance of someone doing something like --key='' I guess but that ends up being None. Changing it, since empty strings are invalid key names anyway.

snapcraft/storeapi/__init__.py
ralsina: Honestly, not very good, like

```
Expecting value: line 1 column 1 (char 0)
```

I'd change it to a generic "Invalid response from server" maybe?

snapcraft/_store.py
sergiusens: fair enough from me.

snapcraft/_store.py
sergiusens: let's change it please. Thanks

snapcraft/storeapi/__init__.py
sergiusens: Yes please, the json decode error is just not friendly for a user.

What would help with debugging this later is to add a logger.debug line with `response` somewhere in the message. Like `logger.debug('Invalid response from the server when doing blah: {} {}'.format(response.code, response.text))`

snapcraft/_store.py
ralsina: Changed

snapcraft/storeapi/__init__.py
ralsina: Done.

runtests.sh
ralsina: Done by adding a specific flake8 comment to disable this check in the function that was failing. 

Rationale is: I can make the complexity "go away" by doing things like a dictionary dispatch but IRL it's replacing a straightforward if/elif chain with something much less obvious.

snapcraft/_store.py
```diff
@@ -281,3 +283,95 @@ def download(snap_name, channel, download_path, arch):
...
+                name, ', '.join(details['errors'])))
+    return details
+
+
+def gated(snap_name):
+    """Print list of snaps gated by snap_name."""
+    store = storeapi.StoreClient()
+    # Get data for the gating snap
+    with _requires_login():
+        snap_data = _get_snap_data(snap_name, store)
```
cprov: @ralsina let's drop this search-based query by the new snap-details endpoint caller:

```
store.cpi.get_package(self, snap_name, channel, arch):
````
https://github.com/snapcore/snapcraft/blob/master/snapcraft/storeapi/__init__.py#L291

Comment (ralsina) [inline snapcraft/_store.py]: ack, replacing.

Comment (ralsina) [inline snapcraft/_store.py]: Done.

File: integration_tests/__init__.py [hunk 7]
```diff
@@ -186,3 +186,27 @@ def update_name_and_version(self, project_dir, name=None, version=None):
             else:
                 print(line)
         return updated_project_dir
+
+    def gated(self, snap_name, expected_validations, expected_error=None):
```
cprov: Roberto, expected_validations has to be optional as well, otherwise you end up passing a and expected result that is completely irrelevant in the testing scenario.

integration_tests/test_validations.py
```diff
@@ -0,0 +1,80 @@
...
+        self.assertEqual(0, self.gated('ubuntu-core', [
+            ('snap-1', '3'),
+            ('snap-2', '5'),
+        ]))
+
+    def test_gated_no_login_failure(self):
+        self.assertEqual(1, self.gated('ubuntu-core', [
+            ('snap-1', '3'),
+            ('snap-2', '5'),
+        ], expected_error='Have you run "snapcraft login'))
```
cprov: Like this call, the expected_validations is not relevant here, only the expected_error

snapcraft/_store.py
```diff
@@ -281,3 +283,92 @@ def download(snap_name, channel, download_path, arch):
         raise RuntimeError(
             'Failed to download {} at {} (mismatched SHA)'.format(
                 snap_name, download_path))
+
+
+def _get_snap_data(name, store):
+    try:
+        details = store.cpi.get_package(name, 'stable', None)
```
cprov: get_package() is not prepared for receiving arch=None, it's pushing a bogus header for CPI waiting for a disaster to happen ... arch is not important for snap-id lookups, so adjust get_package to suppress 'X-Ubuntu-Architecture' is arch is None (similar tweak on errors.SnapNotFound)
Also, I don't see much value in this wrapper if SnapNotFound message is sane, no need to turn it into RuntimeError, 2 different things.

snapcraft/_store.py
```diff
@@ -281,3 +283,92 @@ def download(snap_name, channel, download_path, arch):
...
+        raise RuntimeError()
+
+    store = storeapi.StoreClient()
+
+    # Get data for the gating snap
+    with _requires_login():
+        snap_data = _get_snap_data(snap_name, store)
+    release = str(snap_data['release'][0])
+    snap_id = snap_data['snap_id']
+    developer_id = snap_data['developer_id']
```
cprov: "developer_id" will be the account-id of the uploader of the latest revision available in stable channel, because of package-collaborating (sharing) it might not be the  "publisher", there is no "publisher_id" in CPI. More importantly, it's not the currently authenticated "account-id", so I think it will be less ambiguous (little more expensive) if you simply fetch account-info.account-id.

snapcraft/storeapi/__init__.py
```diff
@@ -296,7 +305,8 @@ def get_package(self, snap_name, channel, arch):
         }
         params = {
             'channel': channel,
-            'fields': 'status,download_url,download_sha512',
+            'fields': 'status,download_url,download_sha512,snap_id,'
+                      'developer_id,release',
```
cprov: as commented above, "developer_id" is not what you are looking for for validation assertions, it's always signed by the publisher.

integration_tests/test_validations.py
```diff
@@ -0,0 +1,80 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
```
ralsina: I'll split it.

ralsina: Done

integration_tests/test_validations.py
ralsina: Fixed

snapcraft/_store.py
ralsina: Will do!

snapcraft/_store.py
ralsina: Ack, will do.

integration_tests/__init__.py
ralsina: Done

snapcraft/storeapi/__init__.py
ralsina: Done.

snapcraft/_store.py
```diff
@@ -272,12 +274,93 @@ def download(snap_name, channel, download_path, arch):
...
-            'Snap {name} for {arch} cannot be found'
-            ' in the {channel} channel'.format(name=snap_name, arch=arch,
-                                               channel=channel))
     except storeapi.errors.SHAMismatchError:
         raise RuntimeError(
             'Failed to download {} at {} (mismatched SHA)'.format(
                 snap_name, download_path))
+
+
+def _get_snap_data(name, store):
```
sergiusens: Is there a reason for this method?

snapcraft/storeapi/errors.py
```diff
@@ -37,10 +37,14 @@ class StoreError(SnapcraftError):
 
 class SnapNotFoundError(StoreError):
 
-    fmt = 'The "{name}" for {arch} was not found in {channel}.'
+    fmt = 'Snap "{name}" for {arch} cannot be found in the {channel} channel.'
```
sergiusens: Mind using `!r` here?

snapcraft/storeapi/errors.py
```diff
@@ -37,10 +37,14 @@ class StoreError(SnapcraftError):
...
 
     def __init__(self, name, channel, arch):
-        super().__init__(name=name, channel=channel, arch=arch)
+        if arch:
+            super().__init__(name=name, channel=channel, arch=arch)
+        else:
+            self.fmt = 'Snap {name} was not found in the {channel} channel.'
```
sergiusens: Same `!r`

integration_tests/test_gated.py
```diff
@@ -0,0 +1,46 @@
...
+
+import os
+
+import integration_tests
+
+
+class GatedTestCase(integration_tests.StoreTestCase):
+
+    def setUp(self):
+        if os.getenv('TEST_STORE') in ['staging', 'production']:
```
cprov: The check below seems to be more effective and the pattern used in other places.
```
if os.getenv('TEST_STORE', 'fake') != 'fake':
```

integration_tests/test_gated.py
```diff
@@ -0,0 +1,46 @@
...
+                          'available in real stores.')
+        super().setUp()
+
+    def test_gated_success(self):
+        self.addCleanup(self.logout)
+        self.login()
+        self.assertEqual(0, self.gated('ubuntu-core', [
+            ('snap-1', '3'),
+            ('snap-2', '5'),
+        ]))
```
cprov: this indentation is a little odd ... What about the following:

```
    validations = [('snap-1', '3'), ('snap-2', '5')]
    exit_code = self.gated('ubuntu-code', validations)
    self.assertEqual(0, exit_code)
```
(repeated in several tests below)

snapcraft/_store.py
```diff
@@ -272,12 +274,93 @@ def download(snap_name, channel, download_path, arch):
     try:
         with _requires_login():
             store.download(snap_name, channel, download_path, arch)
-    except storeapi.errors.SnapNotFoundError:
-        raise RuntimeError(
-            'Snap {name} for {arch} cannot be found'
-            ' in the {channel} channel'.format(name=snap_name, arch=arch,
-                                               channel=channel))
```
cprov: this code is gone because SnapNotFoundError will be raised to top-level exception handler with equivalent info, right ?

snapcraft/_store.py
```diff
@@ -272,12 +274,93 @@ def download(snap_name, channel, download_path, arch):
...
     except storeapi.errors.SHAMismatchError:
         raise RuntimeError(
             'Failed to download {} at {} (mismatched SHA)'.format(
                 snap_name, download_path))
+
+
+def _get_snap_data(name, store):
+    details = store.cpi.get_package(name, 'stable', None)
+    return details
```
cprov: I don't understand what this function buys us. Can't callsites call store.cpi.get_package() directly ?
Also, you've made "arch" optional, no need to pass None.

snapcraft/storeapi/__init__.py
cprov: So, you don't need 'developer_id' and 'release' in the current code.

snapcraft/storeapi/errors.py
```diff
@@ -225,3 +229,19 @@ def __init__(self, snap_name, response):
...
+
+class StoreValidationError(StoreError):
+
+    fmt = 'Received error {status_code!r}: {text!r}'
+
+    def __init__(self, snap_id, response, message=None):
+        try:
+            response_json = response.json()
+            # XXX This error is insanely verbose, SCA needs to trim it down
+            response_json['text'] = response.json()['error_list'][0]['message']
```
cprov: mitigated by https://code.launchpad.net/~cprov/software-center-agent/sign-build-errors/+merge/306939

snapcraft/storeapi/__init__.py
```diff
@@ -432,6 +444,51 @@ def snap_release(self, snap_name, revision, channels):
...
+                     'Accept': 'application/json'})
+        if not response.ok:
+            raise errors.StoreValidationError(snap_id, response)
+        try:
+            response_json = response.json()
+        except JSONDecodeError:
+            message = ('Invalid response from the server when pushing '
+                       'validations: {} {}').format(
+                           response.status_code, response)
+            logger.debug(message)
```
cprov: this is hardly a debug message, it's an error. What about making sure the raised exception has enough information to guide the user and that's it.
If our servers are not returning proper JSON and we don't know already ... it's more like our problem.

snapcraft/tests/test_commands_gated.py
```diff
@@ -0,0 +1,64 @@
...
+                      self.fake_logger.output)
+
+    def test_gated_success(self):
+        self.client.login('dummy', 'test correct password')
+
+        main([self.command_name, 'ubuntu-core'])
+
+        expected_output = """Name      Approved
+snap-1           3
+snap-2           5"""
```
cprov: would textwrap.dedent() help you here ?

snapcraft/_store.py
ralsina: Just that it was common code between gated and validate.

snapcraft/storeapi/errors.py
ralsina: Sure!

snapcraft/storeapi/errors.py
ralsina: Sure! For all args?

snapcraft/_store.py
ralsina: Exactly.

snapcraft/_store.py
ralsina: Well, it was more complicated before :-) I agree it has become a bit too simple and can just be inlined.

integration_tests/test_gated.py
ralsina: Done.

snapcraft/storeapi/__init__.py
ralsina: This error usually will mean something like "we are getting HTML instead of JSON because we are pointing to the wrong URL" and the error message will say "Bad response". I think this was suggested by @sergiusens in the previous review, happy to change it if he's ok with it.

snapcraft/storeapi/errors.py
ralsina: Awesome. Want me to remove the XXX?

snapcraft/tests/test_commands_gated.py
ralsina: It will look nicer. Changing it.

snapcraft/storeapi/__init__.py
ralsina: Fixed.

.travis.yml
```diff
@@ -15,13 +15,13 @@ env:
...
 
 install:
   - sudo apt-get -qq update
   - sudo apt-get install -y python3-coverage
 script:
-  - docker run -e TEST_USER_PASSWORD=$TEST_USER_PASSWORD -v $(pwd):$(pwd) -t ubuntu:xenial sh -c "export LC_ALL=en_US.UTF-8 && locale-gen en_US.UTF-8 && apt update && cd $(pwd) && $DEPENDENCIES && ./runtests.sh $TEST_SUITE"
+  - docker run -e TEST_USER_EMAIL=$TEST_USER_EMAIL -e TEST_USER_PASSWORD=$TEST_USER_PASSWORD -e TEST_STORE=$TEST_STORE -v $(pwd):$(pwd) -t ubuntu:xenial sh -c "export LC_ALL=en_US.UTF-8 && locale-gen en_US.UTF-8 && apt update && cd $(pwd) && $DEPENDENCIES && ./runtests.sh $TEST_SUITE"
 after_success:
   - pip install coveralls
   - coveralls
```
elopio: There's something weird with your diff, this shouldn't be from your PR.

integration_tests/test_gated.py
```diff
@@ -0,0 +1,44 @@
...
+
+import integration_tests
+
+
+class GatedTestCase(integration_tests.StoreTestCase):
+
+    def setUp(self):
+        if os.getenv('TEST_STORE') != 'fake':
+            self.skipTest('Right combination of snaps and IDs is not '
+                          'available in real stores.')
```
elopio: So, we will never be able to test this against a real system? There is no way to seed some valid values there?

ralsina: The main problem is that you need the logged in user to match against the publisher of several snaps in the store. I suppose we can make it work by creating a user account, uploading snaps for it and sharing the password between testers?

snapcraft/_store.py
```diff
@@ -272,12 +274,88 @@ def download(snap_name, channel, download_path, arch):
...
+        if snap_sign.returncode != 0:
+            err = err.decode('ascii', errors='replace')
+            raise RuntimeError('Error signing assertion: {!s}'.format(err))
+
+        # Save assertion to a properly named file
+        fname = '{}-{}-r{}.assertion'.format(snap_name, gated_name, rev)
+        with open(fname, 'wb') as f:
+            f.write(assertion)
+
+        store.push_validation(snap_id, assertion)
```
elopio: This function is too long, can you split it? You are splitting it with comments, it seems it could be nice to have one smaller function per comment.

snapcraft/main.py
```diff
@@ -41,6 +41,9 @@
   snapcraft [options] list-plugins
   snapcraft [options] tour [<directory>]
   snapcraft [options] update
+  snapcraft [options] gated <snap-name>
+  snapcraft [options] validate <snap-name> [--key-name=<key-name>] """ \
+"""[<snap-revision> ...]
```
elopio: #noqa here wouldn't hurt, it could be nicer to read.

snapcraft/main.py
```diff
@@ -289,11 +292,13 @@ def _run_clean(args, project_options):
...
     commands = (
         'list-keys', 'keys', 'register-key',
-        'register', 'upload', 'release', 'push')
+        'register', 'upload', 'release', 'push', 'validate', 'gated')
     return any(args.get(command) for command in commands)
 
 
-def _run_store_command(args):
+# This function's complexity is correlated to the number of
+# commands, no point in checking that.
```
elopio: well, actually ;)
I think it could be nice to split the commands that handle keys from the commands that handle snap uploads and releases. Not a big deal though, just if you agree.

elopio:
Thanks Roberto, this looks pretty good. I've left some questions, I'm just worried about the lack of integration tests against the real server.
And please don't forget to leave your comments and suggestions :) If there's something you hated while implementing these features, let us know.

snapcraft/_store.py
ralsina: Sure! I'll do it tomorrow morning.

ralsina:
Not hated, but docopt is weird once you have these many commands (I know sergio is removing it)

.travis.yml
ralsina: I did something weird trying to collapse the commits I guess. I'll revert this one.

sergiusens:
El martes, 27 de septiembre de 2016 20h'19:10 ART, Roberto Alsina 
> Not hated, but docopt is weird once you have these many 
> commands (I know sergio is removing it)

I am sort of settled on click. Not sure if due to everything I read or the 
irony 🙃




-- 

snapcraft/_store.py
ralsina: I split it in 3, there is a largish chunk in the loop that's hard to split without writing functions with many args, which are a separate readability problem. Let me know if you like the current split or want me to split finer.

ralsina:
Ok, I think I covered all suggestions by @cprov and @elopio, so waiting for final reviews. Thanks for the effort reviewing this guys :-D

integration_tests/test_gated.py
```diff
@@ -0,0 +1,44 @@
...
+
+import os
+
+import integration_tests
+
+
+class GatedTestCase(integration_tests.StoreTestCase):
+
+    def setUp(self):
+        if os.getenv('TEST_STORE') != 'fake':
```
cprov: I would use `os.getenv('TEST_STORE', 'fake')` just to be on the safe-side in case TEST_STORE is not defined.

integration_tests/test_validate.py
```diff
@@ -0,0 +1,61 @@
...
+
+import fixtures
+
+import integration_tests
+
+
+class ValidateTestCase(integration_tests.StoreTestCase):
+
+    def setUp(self):
+        if os.getenv('TEST_STORE') != 'fake':
```
cprov: ditto.

snapcraft/_store.py
```diff
@@ -272,12 +274,99 @@ def download(snap_name, channel, download_path, arch):
...
             'Failed to download {} at {} (mismatched SHA)'.format(
                 snap_name, download_path))
+
+
+def gated(snap_name):
+    """Print list of snaps gated by snap_name."""
+    store = storeapi.StoreClient()
+    # Get data for the gating snap
+    with _requires_login():
+        snap_data = store.cpi.get_package(snap_name, 'stable', None)
```
cprov: I think we should resolve the snap-id via account-info (logged users accessible snaps) instead of hitting CPI and require an active publication in 'stable'.

Also 'None' is not needed, arch is optional

snapcraft/_store.py
```diff
@@ -272,12 +274,99 @@ def download(snap_name, channel, download_path, arch):
...
+    """Generate, sign and upload validation assertions."""
+
+    # Check validations format
+    _check_validations(validations)
+
+    store = storeapi.StoreClient()
+
+    # Get data for the gating snap
+    with _requires_login():
+        snap_data = store.cpi.get_package(snap_name, 'stable', None)
```
cprov: Again, snap-id comes from account-info and 'None' is obsolete.

snapcraft/_store.py
```diff
@@ -272,12 +274,99 @@ def download(snap_name, channel, download_path, arch):
...
+
+    # Need the ID of the logged in user.
+    with _requires_login():
+        account_info = store.get_account_information()
+    authority_id = account_info['account_id']
+
+    # Then, for each requested validation, generate assertion
+    for validation in validations:
+        gated_name, rev = validation.split('=', 1)
+        approved_data = store.cpi.get_package(gated_name, 'stable', None)
```
cprov: Right, for the validated snaps (3rd-part most likely) we hit CPI and require stable publications __ for now, because I wonder if we would even set things up while all snaps are in unstable channels, but I am sure we can sort this out later.
None is not needed.

snapcraft/_store.py
cprov: It looks better now, thanks.

snapcraft/storeapi/__init__.py
cprov: ok, shows additional data if '-d' is given, I am sold.

snapcraft/storeapi/errors.py
cprov: Only if it fixes your problem.

integration_tests/test_gated.py
ralsina: getenv returns None for undefined vars...

snapcraft/_store.py
ralsina: Argh, forgot to remove it when I made it optional :-P
I'll look into resolving via account-info now.

snapcraft/_store.py
ralsina: This only affects automatic updates, right? do we even do those for other channels?

integration_tests/test_gated.py
ralsina: Doh. Of course someone may run this manually :-)

integration_tests/test_validate.py
ralsina: done

integration_tests/test_gated.py
ralsina: done

snapcraft/_store.py
ralsina: Done

snapcraft/_store.py
ralsina: Done

snapcraft/_store.py
cprov: You are right, the happy path is enable refresh-control, no validation nothing gets updated even if the controlled snaps reach 'stable'. Once there one overrides refresh-control and installs the not-yet-validation snaps, run tests, then create the corresponding validations.
Thanks for clarifying this.

integration_tests/test_gated.py
elopio: Yes, that sounds "good" to me. We have u1test+snapcraft@canonical.com, I can give you the password.

snapcraft/storeapi/__init__.py
sergiusens: @cprov I need this for the odd case of having issues in production where I can just tell people to add --debug :-)

snapcraft/main.py
```diff
@@ -35,13 +35,15 @@
...
-  snapcraft [options] sign-build <snap-file> [--key-name <key-name>] [--local]
+  snapcraft [options] sign-build <snap-file> [--key-name=<key-name>] [--local]
   snapcraft [options] upload <snap-file>
   snapcraft [options] push <snap-file> [--release <channels>]
   snapcraft [options] release <snap-name> <revision> <channel>
   snapcraft [options] list-plugins
   snapcraft [options] tour [<directory>]
   snapcraft [options] update
+  snapcraft [options] gated <snap-name>
+  snapcraft [options] validate <snap-name> [<snap-revision> ...] [--key-name=<key-name>]
```
ralsina: Better arg-name ... maybe "validation"?

Yes, it's always snap-name=revision

And it's most definitely not optional, there must be at least one. How does one tell docopt "one or more of these"?

Yes, calling "snapcraft validate telegram-sergiusens" currently does nothing. I can add a check for "at least one validation" if there's no way to tell docopt to do it.

ralsina: @sergiusens I just pushed a change where it now is <validation>... which renames it, and makes it required at least once.

sergiusens: http://try.docopt.org/?doc=snapcraft%0D%0A%0D%0AUsage%3A%0D%0A++snapcraft+validate+%3Csnap-name%3E+%28--keys+%3Cone%3E+%7C+%3Ctwo%3E%29&argv=validate+snap+--keys+1

sergiusens:
ok to test

PR #814, opened 2016-09-20
Make source-depth a parameter and opt-in

sergiusens:
LP: #1617783

### DIFF
--- a/snapcraft/internal/sources.py
+++ b/snapcraft/internal/sources.py
@@ -33,6 +33,11 @@ code for that part, and how to unpack it if necessary.
     control system or compression algorithm. The source-type key can tell
     snapcraft exactly how to treat that content.
 
+  - source-depth: <integer>
+
+    By default clones or branches with full history, specifying a depth
+    will truncate the history to the specified number of commits.
+
   - source-branch: <branch-name>
 
     Snapcraft will checkout a specific branch from the source tree. This
@@ -90,11 +95,12 @@ class IncompatibleOptionsError(Exception):
 class Base:
 
     def __init__(self, source, source_dir, source_tag=None,
-                 source_branch=None):
+                 source_branch=None, source_depth=None):
         self.source = source
         self.source_dir = source_dir
         self.source_tag = source_tag
         self.source_branch = source_branch
+        self.source_depth = source_depth
 
 
 class FileBase(Base):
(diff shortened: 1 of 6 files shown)


kalikiana:
What's the rationale for changing the default behavior? It seems to me like I'd want to add source-depths to all my snaps now because it saves time and disk space.

snapcraft/internal/sources.py
```diff
@@ -119,8 +125,9 @@ def download(self):
 class Script(FileBase):
 
     def __init__(self, source, source_dir, source_tag=None,
-                 source_branch=None):
-        super().__init__(source, source_dir, source_tag, source_branch)
+                 source_branch=None, source_depth=None):
+        super().__init__(source, source_dir, source_tag,
+                         source_branch, source_depth)
```
josepht: Can we call it 'git-source-depth'? Also items 19-21 in Effective Python might give you some ideas for simplifying the base class `__init__()` by popping the accepted kwargs and throwing an exception if there are any leftover options

sergiusens: All this sounds good, but the bug mentions `source-depth` and this same modus operandi we are doing applies for `source-tag` when using local sources or tarballs

I will leave the popping for when we move the logic to the pluginhandler

sergiusens:
El 20/09/16 a las 11:26, Christian Dywan escribió:
>
> What's the rationale for changing the default behavior? It seems to me 
> like I'd want to add source-depths to all my snaps now because it 
> saves time and disk space.
>
Refer to the bug please for discussing the rationale.

sergiusens:
@kalikiana in case you missed it, refer to the bug please for discussion of the rationale

snapcraft/internal/sources.py
```diff
@@ -33,6 +33,11 @@
     control system or compression algorithm. The source-type key can tell
     snapcraft exactly how to treat that content.
 
+  - source-depth: <integer>
+
+    By default clones or branches with full history, specifying a depth
+    will truncate the history to the specified number of commits.
```
elopio: What about mentioning that usually only one branch is needed?

elopio:
lgtm

kyrofa:
I'm not a fan of this change, but :+1: to the code.

PR #817, opened 2016-09-20
lxd: use built-in image streams.

sergiusens:
Instead of setting up our own remote for each cleanbuild
invocation use the provided built-in image streams from
lxd.

LP: #1625787


elopio:
Good. Recently I looked at my remotes and I had like 15 pet names. I like this better.

PR #821, opened 2016-09-22
Make copies of remote parts to avoid ordering issues

sergiusens:


elopio:
:+1: But also here, like Joe, I don't understand the bug. Can you please describe it more in launchpad and add some comments to the test?

sergiusens:
@elopio @josepht explained!

PR #822, opened 2016-09-22
Don't filter .pth files in python plugin

SamYaple:
The .pth files are needed by some projects to help with importing. We
currently filter them out, breaking imports for those projects.

LP: #1626668

### DIFF
--- a/snapcraft/plugins/python.py
+++ b/snapcraft/plugins/python.py
@@ -234,9 +234,6 @@ class PythonPlugin(snapcraft.BasePlugin):
         fileset.append('-bin/pip*')
         fileset.append('-bin/easy_install*')
         fileset.append('-bin/wheel')
-        # .pth files are only read from the built-in site-packages directory.
-        # We use PYTHONPATH for everything so not needed.
-        fileset.append('-**/*.pth')
         # Holds all the .pyc files. It is a major cause of inter part
         # conflict.
         fileset.append('-**/__pycache__')
--- a/snapcraft/tests/test_plugin_python.py
+++ b/snapcraft/tests/test_plugin_python.py
@@ -175,7 +175,6 @@ class PythonPluginTestCase(tests.TestCase):
             '-bin/pip*',
             '-bin/easy_install*',
             '-bin/wheel',
-            '-**/*.pth',
             '-**/__pycache__',
             '-**/*.pyc',
         ]


SamYaple:
Heh. Coverage falling again. what a suprise

sergiusens:
it is so little though

elopio:
Hey @SamYaple, thanks again for your code! Do you have a simple project that fails because of this?

SamYaple:
@elopio it failed for me with dogpile.cache, it will for you too. There are others as well that this type of import would break.

I commented on it in this patch https://github.com/snapcore/snapcraft/pull/765

Im trying to write a quick simple project that will break like this and publish it to pypi so we can write a test against it for the future

elopio:
@SamYaple yes, that would be awesome. If it's a dummy simple project, we can put it in our integration suite. If it's more complex, we can put it in our demos. ^_^

sergiusens:
El 22/09/16 a las 15:18, Leo Arias escribió:
>
> @SamYaple <https://github.com/SamYaple> yes, that would be awesome. If 
> it's a dummy simple project, we can put it in our integration suite. 
> If it's more complex, we can put it in our demos. ^_^
>

The reason I want it our demos is to be able to run it from the snap.

sergiusens:
ok to test

SamYaple:
Lets go ahead and merge this. I have filed a tech debt bug to add an appropriate regression test for this.

https://bugs.launchpad.net/snapcraft/+bug/1626839

I dont want to hold up this fix for that test though.

sergiusens:
I tried to this http://paste.ubuntu.com/23220251/ on an unpatched snapcraft and was able to do this http://paste.ubuntu.com/23220246/ (I am not sure how to expose the current .pth bug)

SamYaple:
http://paste.ubuntu.com/23220481/

For dogpile.cache, it doesn't have the issue on python3 and version 0.6.x do not have the silly .pth usage.

I'd rather not be using dogpile.cache specifically for testing, but it is the easiest one to demonstrate at teh moment. I am working on a simple module to reproduce this.

sergiusens:
@SamYaple I am not questioning it doesn't happen, I was just not seeing it :-)

SamYaple:
Well I can't reproduce it on python3 yet. I would like to have an example of that for reference sake. But that is what the other bug will track. I feel like this is a regression that could end up happening again.

PR #823, opened 2016-09-22
plainbox-provider plugin: rewrite python shebangs

sergiusens:
LP: #1626587

### DIFF
--- a/snapcraft/plugins/plainbox_provider.py
+++ b/snapcraft/plugins/plainbox_provider.py
@@ -52,5 +52,5 @@ class PlainboxProviderPlugin(snapcraft.BasePlugin):
 
         # Fix all shebangs to use the in-snap python.
         file_utils.replace_in_file(self.installdir, re.compile(r''),
-                                   re.compile(r'#!.*python'),
+                                   re.compile(r'^#!.*python'),
                                    r'#!/usr/bin/env python')
--- a/snapcraft/plugins/plainbox_provider.py
+++ b/snapcraft/plugins/plainbox_provider.py
@@ -29,7 +29,10 @@ For more information check the 'plugins' topic for the former and the
 'sources' topic for the latter.
 """
 
+import re
+
 import snapcraft
+from snapcraft import file_utils
 
 
 class PlainboxProviderPlugin(snapcraft.BasePlugin):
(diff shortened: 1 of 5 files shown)


snapcraft/plugins/plainbox_provider.py
```diff
@@ -46,3 +49,8 @@ def build(self):
             "python3", "manage.py", "install", "--layout=relocatable",
             "--prefix=/providers/{}".format(self.name),
             "--root={}".format(self.installdir)])
+
+        # Fix all shebangs to use the in-snap python.
+        file_utils.replace_in_file(self.installdir, re.compile(r''),
+                                   re.compile(r'#!.*python'),
```
josepht: I'd like to see this be r'^#!.*python' or something similar so that in the rare case that that string is in a file it won't get replaced.  May need to take leading spaces into account as well, IDK what posix says about shebang lines.

snapcraft/plugins/plainbox_provider.py
```diff
@@ -46,3 +49,8 @@ def build(self):
...
+                                   r'#!/usr/bin/env python')
```
elopio: Looks good, but it's missing a unit test. More than one when you implement the regexp suggestion.

sergiusens:
El jueves, 22 de septiembre de 2016 17h'18:49 ART, Leo Arias 
> elopio requested changes on this pull request.
> 
> 
> 
>> @@ -46,3 +49,8 @@ def build(self):
>              "python3", "manage.py", "install", "--layout=relocatable",
>              "--prefix=/providers/{}".format(self.name),
>              "--root={}".format(self.installdir)])
> +
> +        # Fix all shebangs to use the in-snap python.
> +        file_utils.replace_in_file(self.installdir, re.compile(r''),
> +                                   re.compile(r'#!.*python'),
> +                                   r'#!/usr/bin/env python')
> 
> Looks good, but it's missing a unit test. More than one when 
> you implement the regexp suggestion.

Yeah, there are no units for this plug in.



-- 

snapcraft/plugins/plainbox_provider.py
sergiusens: DONE

snapcraft/plugins/plainbox_provider.py
sergiusens: DONE!

PR #824, opened 2016-09-23
Add `snapcraft create-key`

cjwatson:
Wrapping `snap create-key` means that we don't confuse users by
expecting them to use two similar-but-different commands during the key
creation and registration workflow.  In addition, this augments
`snap create-key` by failing if a key with the same name is already
registered with the store, since a later registration attempt would fail
in that case.

### DIFF
--- a/debian/snapcraft-completion
+++ b/debian/snapcraft-completion
@@ -4,7 +4,7 @@ _snapcraft()
     COMPREPLY=()
     cur="${COMP_WORDS[COMP_CWORD]}"
     prev="${COMP_WORDS[COMP_CWORD-1]}"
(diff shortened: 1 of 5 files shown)


cjwatson:
I haven't included an integration test here since the same store interactions are mostly tested elsewhere, and since I didn't want to get into having to ensure a suitable entropy source in test environments.

cjwatson:
I've made a suggestion in that document matching what I've done here, if somebody would like to review and accept that.  I've also applied your suggested tweak to the create_key name guard, thanks.

emgee:
fwiw, I was just going through the model assertion creation and needed to create and then register a key. Using `snap create-key` followed by `snapcraft register-key` is a strange UX.

As Colin points out, devs really need one tool for this to avoid confusion ... and snapcraft seems to be that tool.

PR #826, opened 2016-09-23
Do not depend on Content-Length when Content-Encoding is gzip

tachyons:
LP: #1611776

### DIFF
--- a/snapcraft/internal/indicators.py
+++ b/snapcraft/internal/indicators.py
@@ -30,7 +30,11 @@ def download_requests_stream(request_stream, destination, message=None):
     if not message:
         message = 'Downloading {!r}'.format(os.path.basename(destination))
 
-    total_length = int(request_stream.headers.get('Content-Length', '0'))
+    if request_stream.headers.get('Content-Encoding') == 'gzip' :
+        total_length = len(request_stream.content)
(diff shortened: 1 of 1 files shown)


snapcraft/internal/indicators.py
```diff
@@ -30,7 +30,11 @@ def download_requests_stream(request_stream, destination, message=None):
     if not message:
         message = 'Downloading {!r}'.format(os.path.basename(destination))
 
-    total_length = int(request_stream.headers.get('Content-Length', '0'))
+    if (request_stream.headers.get('Content-Encoding') == 'gzip'):
```
sergiusens: Why not make this the behavior for any `Content-Encoding`?

Such as:
```python
if request_stream.headers.get('Content-Encoding', ''):
```

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding

tachyons: Yeah you are right , actual length can vary in any of these cases

tachyons:
I checked my solution with yaml file provided by the op . It is working fine . But my knowledge in python is limited and I never wrote test cases in python before . So closing this pull request so that someone else can do that . Thanks

PR #827, opened 2016-09-23
Support setting build targets in the maven plugin. Make the maven plu…

evandandrea:
- Support setting build targets in the maven plugin.
- Make the maven plugin match the ant plugin for setting java properties.

We could standardise around key-value pairs for options or a list of -Dprop=val, -Dprop2=val, … cmake, autotools, gradle, etc use the former approach, so maybe I should flip this PR to migrate the ant plugin to a list of string options?


sergiusens:
@evandandrea the `make` based ones use lists since order is important, not sure how it affects ant and maven.

kyrofa:
This PR has conflicts.

kyrofa:
This has been sitting for a while. I'm going to go ahead and close it. Please re-propose once you're ready to continue.

PR #829, opened 2016-09-26
sources: fix type when calling with depth

sergiusens:
LP: #1627772


integration_tests/test_git_source.py
```diff
@@ -116,3 +116,18 @@ def test_pull_git_branch(self):
...
+
+        self._init_and_config_git()
+        subprocess.check_call(
+            ['git', 'commit', '-m', '"1"', '--allow-empty'],
+            stdout=subprocess.DEVNULL)
+        subprocess.check_call(
+            ['git', 'commit', '-m', '"2"', '--allow-empty'],
+            stdout=subprocess.DEVNULL)
+
+        self.run_snapcraft('pull', project_dir)
```
elopio: This is good to make sure that it won't crash, but it won't check that only one commit.
Can you please add a check for git rev-list --all --count == 1 ?
Trying it locally, I see: warning: --depth is ignored in local clones; use file:// instead.
So we are probably also missing a file:// in front of the path if it's local.

sergiusens:
El 26/09/16 a las 17:18, Leo Arias escribió:
> This is good to make sure that it won't crash, but it won't check that 
> only one commit.
> Can you please add a check for git rev-list --all --count == 1 ?
> Trying it locally, I see: warning: --depth is ignored in local clones; 
> use file:// instead.

Even so, I don't want to test git and its behavior. This is only a 
gimmick on our tests as this is a very unusual scenario.

> So we are probably also missing a file:// in front of the path if it's 
> local.

file://. ?? How do you suppose this would go?

PR #830, opened 2016-09-26
python plugin: only replace proper shebangs.

sergiusens:
Only replace shebangs at the beginning of a file.

LP: #1626587


elopio:
With this patch, I could build the plainbox snap that was previously giving conflict errors.

PR #836, opened 2016-09-28
Deal with 404 responses from the store's snap status endpoint.

thomir:
This PR helps resolve https://bugs.launchpad.net/snapcraft/+bug/1627865

There's still not enough error checking code there - JSONDecodeErrors are still possible, and should be handled correctly, but that's an issue for another time. I didn't find any tests for this code - if I'm missing them, please let me know and I'll add some.

### DIFF
--- a/snapcraft/storeapi/__init__.py
+++ b/snapcraft/storeapi/__init__.py
@@ -494,7 +494,14 @@ class StatusTracker:
         connection_errors_allowed = 10
         while True:
             try:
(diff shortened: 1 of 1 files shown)


elopio:
Hey Thomi. We have some tests for push in snapcraft/tests/store/test_store_client.py
They are using a fake server with a hard-coded return values to get some coverage. Maybe you can get the fake server to return a 404.

kyrofa:
Closing due to inactivity.

PR #839, opened 2016-09-29
Fixing 'sign-build' integration tests.

cprov:
Even for building the local assertion, the command actually checks if the
account logged in has upload permission on the targeted snap (resolving
name to snap-id via account-info).

The test now builds an random snap, register the name in the store and
only then signs the snap. The fake server was extended to record registered
names and return them in the account-info endpoint.

### DIFF
--- a/integration_tests/test_store_sign_build.py
+++ b/integration_tests/test_store_sign_build.py
@@ -17,7 +17,7 @@
 import os
 import fixtures
 import shutil
-import integration_tests
+import uuid
 
 from testtools.matchers import (
     Contains,
@@ -25,6 +25,8 @@ from testtools.matchers import (
     FileExists,
     Not,
 )
+
+import integration_tests
 
 
 class SignBuildTestCase(integration_tests.StoreTestCase):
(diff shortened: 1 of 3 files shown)


integration_tests/test_store_sign_build.py
```diff
@@ -26,6 +26,8 @@
     Not,
 )
 
+import integration_tests
```
sergiusens: Thanks for doing the right thing 😃

integration_tests/test_store_sign_build.py
```diff
@@ -51,17 +53,27 @@ def test_unsuccessful_sign_build_no_login(self):
...
+        snap_path = '{}_{}_{}.snap'.format(name, version, 'all')
+        self.assertThat(snap_path, FileExists())
+        self.register(name)
+
+        status = self.sign_build(snap_path, local=True)
         self.assertEqual(0, status)
-        self.assertThat(self.snap_build_path, FileExists())
-        self.assertThat(self.snap_build_path,
+
+        snap_build_path = '{}-build'.format(snap_path)
```
sergiusens: Was this even not in existence before?

snapcraft/tests/fake_servers.py
```diff
@@ -483,13 +490,15 @@ def _handle_registration_request(self):
...
+            self._handle_successful_registration(snap_name)
 
-    def _handle_successful_registration(self):
+    def _handle_successful_registration(self, name):
+        self.server.registered_names.append(name)
         self.send_response(201)
         self.send_header('Content-Type', 'application/json')
         self.end_headers()
         response = {'snap_id': 'test-snap-id'}
+        self.server.regitered_name
```
sergiusens: Typo

sergiusens: Also what is the intention of this line?

integration_tests/test_store_sign_build.py
cprov: It was, the test works if the account logged in was the publisher of 'basic' name, but this is impossible in staging SCA. See https://travis-ci.org/snapcore/snapcraft/builds/163606844

snapcraft/tests/fake_servers.py
cprov: The *intention* was to leave no doubts this was a 2am PR ;-) typo removed.

PR #840, opened 2016-09-29
Catkin plugin: build with in-snap python.

kyrofa:
Our test cases all had python installed on the system, so the fact that the plugin used it went unnoticed. This PR fixes LP: [#1628990](https://bugs.launchpad.net/snapcraft/+bug/1628990) by using the in-snap python for build-time as well as run-time. It also pulls in gcc, libc-dev, and make as build packages.

### DIFF
--- a/snapcraft/plugins/catkin.py
+++ b/snapcraft/plugins/catkin.py
@@ -43,6 +43,7 @@ import subprocess
 import snapcraft
 from snapcraft import (
     common,
+    file_utils,
     repo,
 )
 
@@ -302,9 +303,9 @@ deb http://${security}.ubuntu.com/${suffix} trusty-security main universe
             return '"' + ';'.join(paths) + '"'
 
         # Looking for any path-like string
-        common.replace_in_file(self.rosdir, re.compile(r'.*Config.cmake$'),
-                               re.compile(r'"(.*?/.*?)"'),
-                               rewrite_paths)
+        file_utils.replace_in_file(self.rosdir, re.compile(r'.*Config.cmake$'),
+                                   re.compile(r'"(.*?/.*?)"'),
+                                   rewrite_paths)
 
     def _finish_build(self):
         self._use_in_snap_python()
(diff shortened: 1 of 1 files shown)


snapcraft/plugins/catkin.py
```diff
@@ -277,6 +283,22 @@ def build(self):
         self._finish_build()
 
     def _prepare_build(self):
+        # Fix all shebangs to use the in-snap python.
+        common.replace_in_file(self.rosdir, re.compile(r''),
+                               re.compile(r'#!.*python'),
```
josepht: should probably start with '^'

snapcraft/plugins/catkin.py
```diff
@@ -277,6 +283,22 @@ def build(self):
...
+                               r'#!/usr/bin/env python')
+
+        # Also replace the python usage in 10.ros.sh to use the in-snap python.
+        ros10_file = os.path.join(self.rosdir,
+                                  'etc/catkin/profile.d/10.ros.sh')
+        if os.path.isfile(ros10_file):
```
josepht: I don't know if we care, but this is racey.  I think the SOP is to wrap the `open()` in a `try: ... except:`

snapcraft/plugins/catkin.py
```diff
@@ -277,6 +283,22 @@ def build(self):
         self._finish_build()
 
     def _prepare_build(self):
+        # Fix all shebangs to use the in-snap python.
+        common.replace_in_file(self.rosdir, re.compile(r''),
```
sergiusens: use `fileutils` instead

snapcraft/plugins/catkin.py
sergiusens: Agreed

snapcraft/plugins/catkin.py
kyrofa: Of course we care! I blindly copied that from elsewhere. I've refactored a bit to fix both places.

snapcraft/plugins/catkin.py
kyrofa: Indeed, done.

snapcraft/plugins/catkin.py
kyrofa: Good catch, thanks! Done, and fixed one other place in this plugin as well.

snapcraft/plugins/catkin.py
kyrofa: Ah, this was tricky. I missed the 'r+' on the mode-- we open for read/write. We need to know if it exists before we try, or we'll create it (and we don't want to if it's not already there). I've waffled back on this.

PR #844, opened 2016-09-30
Add `snapcraft history`.

maxiberta:
Add `snapcraft history` subcommand following discussed specs (note this is very similar to proposed `snapcraft status`).

LP: #1629339


snapcraft/_store.py
```diff
@@ -348,6 +354,24 @@ def download(snap_name, channel, download_path, arch):
...
+
+    parsed_revisions = [
+        (rev['revision'], rev['timestamp'], rev['arch'], rev['version'],
+         _get_text_for_current_channels(
+            rev['channels'], rev['current_channels']))
+        for rev in history]
+    tabulated_revisions = tabulate(
+        parsed_revisions,
+        headers=['Rev.', 'Uploaded', 'Arch', 'Version', 'Channels'],
+        tablefmt="plain")
```
sergiusens: oh, we have in a guy in our team that told us to always use single quotes unless it is a docstring ;-)

snapcraft/storeapi/__init__.py
```diff
@@ -194,6 +194,30 @@ def release(self, snap_name, revision, channels):
...
+    def get_snap_history(self, snap_name, series=None, arch=None):
+        account_info = self.get_account_information()
+        snap_id = None
+        try:
+            if series:
+                snap_id = account_info['snaps'][series][snap_name]['snap-id']
+            else:
+                for key, value in account_info['snaps'].items():
+                    if snap_name in value:
+                        snap_id = value[snap_name]['snap-id']
```
josepht: Will there be only one?

maxiberta: It's possible that there are multiple matches if the snap name is registered in different series (`account_info['snaps']` is a dict with series name as keys), but it should be ok as they should all share the same snap id. I'm adding a break once found so we keep the 1st one.

snapcraft/_store.py
maxiberta: Heh, no problem. Even in cases such as:

```self.assertIn("Snap 'mysnap' was not found.", str(error.output))``` ?

maxiberta:
@josepht sorry, you mentioned generators above but I'm not sure exactly what you mean.

josepht:
Sorry, I meant comprehensions.

sergiusens:
@maxiberta maybe he meant nested list comprehensions which to be honest I had to read 3 times to get a feel of what it is doing.

I personally often prefer to write 2, one in the form of a generator `(` and `)` and the next in either form

snapcraft/storeapi/__init__.py
maxiberta: This snap id to snap name resolution stuff could be moved into its own method/function as it'll be used by `get_snap_status()` as well.

maxiberta:
Fixed integration tests on the staging store; and --series now defaults to DEFAULT_SERIES.

PR #845, opened 2016-09-30
Add `snapcraft status`.

maxiberta:
Add `snapcraft status` subcommand following discussed specs (note this is very similar to proposed `snapcraft history`).

LP: #1629339


sergiusens:
ok to test

snapcraft/_store.py
```diff
@@ -348,6 +355,23 @@ def download(snap_name, channel, download_path, arch):
...
+        status = store.get_snap_status(snap_name, series, arch)
+
+    parsed_channels = [
+        channel
+        for arch, channel_map in sorted(status.items())
+        for channel in _format_channel_map(channel_map, arch)]
+
+    tabulated_channels = tabulate(
+        parsed_channels, headers=['Arch', 'Channel', 'Version', 'Revision'],
+        tablefmt="plain")
```
sergiusens: oh, we have in a guy in our team that told us to always use single quotes unless it is a docstring ;-)

snapcraft/storeapi/__init__.py
```diff
@@ -194,6 +194,31 @@ def release(self, snap_name, revision, channels):
...
+        account_info = self.get_account_information()
+        snap_id = None
+        try:
+            if series:
+                snap_id = account_info['snaps'][series][snap_name]['snap-id']
+            else:
+                snap_id = None
+                for key, value in account_info['snaps'].items():
+                    if snap_name in value:
+                        snap_id = value[snap_name]['snap-id']
```
maxiberta: It's possible if the snap name is registered in different series (`account_info['snaps']` is a dict with series name as keys), but it should be ok as they should all share the same snap id. I'm adding a `break` once found so we keep the 1st one.

snapcraft/storeapi/errors.py
```diff
@@ -261,3 +273,24 @@ def __init__(self, response):
             pass
 
         super().__init__(error=error)
+
+
+class StoreSnapStatusError(StoreError):
+
+    fmt = (
+        'Error fetching status of snap id "{snap_id}" for {arch} '
+        'in {series} series: {error}.')
```
kyrofa: Should probably use `!r` here for consistency.

maxiberta: Done! Thanks

kyrofa: Thank you!

snapcraft/storeapi/__init__.py
maxiberta: Ok, I'm setting the default value of `series` to `constants.DEFAULT_SERIES` as discussed with @cprov. As a result this block of code will be made much simpler.

sergiusens:
err `s/history/status/`

maxiberta:
Updated with fixed integration tests on the staging store, and `--series` now defaults to `DEFAULT_SERIES`.

snapcraft/storeapi/__init__.py
```diff
@@ -503,6 +521,28 @@ def push_snap_build(self, snap_id, snap_build):
         if not response.ok:
             raise errors.StoreSnapBuildError(response)
 
+    def snap_status(self, snap_id, series, arch):
+        qs = {}
+        if series:
+            qs['series'] = series
+        if arch:
+            qs['arch'] = arch
+        url = 'snaps/' + snap_id + '/status'
```
elopio: Nevermind. If we ever run this in windows, the path will be wrong. Hardcoding the / is the right thing to do.

sergiusens:
El jueves, 6 de octubre de 2016 03h'28:31 ART, Leo Arias 
> elopio commented on this pull request.
> 
> 
> 
>> @@ -503,6 +521,28 @@ def push_snap_build(self, snap_id, snap_build):
>          if not response.ok:
>              raise errors.StoreSnapBuildError(response)
>  
> +    def snap_status(self, snap_id, series, arch):
> +        qs = {}
> +        if series:
> +            qs['series'] = series
> +        if arch:
> +            qs['arch'] = arch
> +        url = 'snaps/' + snap_id + '/status'
> 
> Nevermind. If we ever run this in windows, the path will be 
> wrong. Hardcoding the / is the right thing to do.

If anything, it would be urljoin which would be needed.

> 



-- 

maxiberta:
The merge of `snapcraft history` caused some conflicts. I'm fixing asap.

PR #848, opened 2016-10-04
`sign-build` to prompt users for key selection

cprov:
This way we can handle cases when the keyring is not setup, the selected key name is not present or selected key is not registered (LP #1629966).

Also prevent `snap sign-build` failures to create stray/empty '-build' files (LP #1629984).

### DIFF
--- a/snapcraft/_store.py
+++ b/snapcraft/_store.py
@@ -152,7 +152,7 @@ def list_keys():
     tabulated_keys = tabulate(
         [('*' if key['sha3-384'] in enabled_keys else '-',
           key['name'], key['sha3-384'],
-          '' if key['sha3-384'] in enabled_keys else '(not enabled)')
+          '' if key['sha3-384'] in enabled_keys else '(not registered)')
          for key in keys],
         headers=["", "Name", "SHA3-384 fingerprint", ""],
         tablefmt="plain")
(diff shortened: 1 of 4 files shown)


kyrofa:
Having the bugref in the title makes the resulting commit a little yucky. You already have it in the PR description which is perfect!

kyrofa:
Could you also please make the bugrefs in the commit conform to the [contribution guide](https://github.com/snapcore/snapcraft/blob/master/CONTRIBUTING.md)? i.e. on a line all by themselves, so `gbp dch` can pick them up.

cprov:
@kyrofa, commit message conforms with the style guide, thanks for suggesting it.

snapcraft/_store.py
```diff
@@ -259,26 +266,39 @@ def sign_build(snap_filename, key_name='default', local=False):
...
+        key = _maybe_prompt_for_key(key_name)
+        if not local:
+            is_registered = [
+                a for a in account_info['account_keys']
+                if a['public-key-sha3-384'] == key['sha3-384']
+            ]
+            if not is_registered:
+                raise RuntimeError(
+                    'The key "{}" is not registered in the Store.\n'
+                    'Please register it with `snapcraft register-key {}` '
```
sergiusens: `{!r}` instead of `"{}"`

sergiusens: In both instances of `{}`

cprov: Good point, fixed in the next revision.

PR #855, opened 2016-10-05
tools: script to talk to the staging servers

sergiusens:
This is a sourceable script which would make it easy to work
against the staging servers.

LP: #1630083

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -49,14 +49,13 @@ Snapcraft has the ability to upload snaps for publication in the Snappy Store.
 If you're working on a feature that requires you to interact with the store, you
 might want to use the staging server instead of the production store. To do
 that, make sure you have an account on the
-[staging server](https://login.staging.ubuntu.com), then set the following
-environment variables:
+[staging server](https://login.staging.ubuntu.com), then run:
 
-    UBUNTU_STORE_API_ROOT_URL='https://myapps.developer.staging.ubuntu.com/dev/api/'
-    UBUNTU_STORE_SEARCH_ROOT_URL=https://search.apps.staging.ubuntu.com/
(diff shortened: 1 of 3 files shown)


kyrofa:
Though I wonder if it would be even cooler if we made this a subshell, then one can simply exit out of it. That seems to be how other similar tools work.

sergiusens:
@kyrofa I was going for a python venv xp, android does it this way too now that I think of it, I don't mind either way.

kyrofa:
No strong opinions here either, +1 either way. Curious to see what the others think.

sergiusens:
@kyrofa I don't mind, this is my idea of a *vacation* from a bigger refactor going on :-P

HACKING.md
```diff
@@ -49,14 +49,13 @@ Snapcraft has the ability to upload snaps for publication in the Snappy Store.
...
+[staging server](https://login.staging.ubuntu.com), then run:
 
-    UBUNTU_STORE_API_ROOT_URL='https://myapps.developer.staging.ubuntu.com/dev/api/'
-    UBUNTU_STORE_SEARCH_ROOT_URL=https://search.apps.staging.ubuntu.com/
-    UBUNTU_STORE_UPLOAD_ROOT_URL='https://upload.apps.staging.ubuntu.com/'
-    UBUNTU_SSO_API_ROOT_URL='https://login.staging.ubuntu.com/api/v2/'
+    source tools/staging_env.sh
 
+You will see a prompt indicating that you are going to be talking to the staging
+server. Once you are done with working with the staging servers you can run
```
elopio: You should remove one "with":
Once you are done _ working with the staging...

PR #856, opened 2016-10-05
Call organize() after building.

josepht:
Add an integration test to verify that no conflicts are found.

LP:#1628098


integration_tests/snaps/no-conflicts/snapcraft.yaml
```diff
@@ -0,0 +1,19 @@
+name: test-package
+version: 0.1
+summary: one line summary
+description: a longer description
+icon: icon.png
+confinement: strict
```
sergiusens: add `grade`

josepht:
A couple of questions:

1. Do you want me to move the `_organize()` call back to `stage()`?
2. Am I correct in my understanding that your concern is the case where `organize` might create a conflict?  I.e. `part1` uses `organize` to move `bin/test-a` to `bin/test` and `part2` has a `bin/test` as well?

sergiusens:
@josepht your integration test covers item `2` perfectly, yes to `1`

We want to make `organize` smarter by having organize not work on install dir but on stage dir instead; So this is the dance.

1. `install` is populated as mandated by the plugin code.
2. we check for conflicts for the `install`->`stage` transition:
    1. We grab the organize list for the part
    1. We create the migratable filesets list taking into account the organize list
3. we migrate files from `install` to `stage` using `src, dst` tuples `src` coming from `install` and `dst` being `stage` with applied organization.

josepht:
@sergiusens so `_organize()` can go away if `migratable_filteset_for()` is going to handle the logic of mapping filepaths based on the `organize` option.  Right?

sergiusens:
El 06/10/16 a las 10:48, Joe Talbott escribió:
>
> @sergiusens <https://github.com/sergiusens> so |_organize()| can go 
> away if |migratable_filteset_for()| is going to handle the logic of 
> mapping filepaths based on the |organize| option. Right?
>

Correct.

josepht:
@sergiusens what do you think about adding the organize mappings to `_migrate_fileset()` so that it takes an array of tuples/arrays/dicts/whatever that includes the install path name and stage path name?  Or did you have something else in mind?

josepht:
@sergiusens Converting filesets to tuples of ($path, $new_path or None) is proving to be a lot of work and I feel like adding a bunch of `snap_file[0]` and ` snap_file[1]` is making the code less readable.  Do you have any suggestions?

josepht:
Superseded by #897

PR #858, opened 2016-10-06
python plugin: allow usage of bzr

sergiusens:
LP: #1631057


blakerouse:
I have tested this fix with my MAAS snap and it fixes the issue. Thanks for the quick fix.

PR #859, opened 2016-10-07
Set GOBIN in go plugin build environment.

tasdomas:
Fixes #1553215. If the user's environment has GOBIN set, it will interfere
with the go plugin, forcing the end result executables to end up there instead of
the expected location.


sergiusens:
I see you have not signed the CLA, signing it for this is not really needed but I do recommend doing so, read more about it on CONTRIBUTING.md

PR #861, opened 2016-10-07
python plugin: install from wheel for local setup.py

sergiusens:
This will avoid a rebuild and dtrt wrt data_files.

LP: #1631071

### DIFF
--- a/snapcraft/plugins/python.py
+++ b/snapcraft/plugins/python.py
@@ -227,7 +227,13 @@ class PythonPlugin(snapcraft.BasePlugin):
             wheels = []
             for command in commands:
                 wheels.extend(pip.wheel(**command))
-            pip.install(wheels)
+
+            installed = pip.list(self.run_output)
+            wheel_names = [os.path.basename(w).split('-')[0] for w in wheels]
+            # we want to avoid installing what is already provided in
+            # stage-packages
+            need_install = [k for k in wheel_names if k not in installed]
+            pip.install(need_install)
 
     def _fix_permissions(self):
         for root, dirs, files in os.walk(self.installdir):
@@ -280,6 +286,22 @@ class _Pip:
 
         if dependency_links:
             self._extra_pip_args.append('--process-dependency-links')
+
+    def list(self, exec_func=None):
+        """Return a dict of installed python packages with versions."""
+        if not exec_func:
+            exec_func = self._exec_func
+        cmd = [self._runnable, 'list']
+
+        output = exec_func(cmd, env=self._env)
+        package_listing = {}
+        version_regex = re.compile('\((.+)\)')
+        for line in output.splitlines():
+            line = line.split()
+            m = version_regex.search(line[1])
+            package_listing[line[0]] = m.group(1)
+
+        return package_listing
 
     def wheel(self, args, **kwargs):
         cmd = [
--- a/snapcraft/plugins/python.py
+++ b/snapcraft/plugins/python.py
@@ -199,6 +199,8 @@ class PythonPlugin(snapcraft.BasePlugin):
                    constraints=constraints,
                    dependency_links=self.options.process_dependency_links)
 
+        wheels = []
+
         if self.options.requirements:
             if isurl(self.options.requirements):
                 requirements = self.options.requirements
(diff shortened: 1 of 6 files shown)


blakerouse:
This is not working for me, I get the following when a dependency uses data_files.

    copying doc/index.rst -> /usr/share/doc/curtin
    error: could not create '/usr/share/doc/curtin/index.rst': Permission denied
    
    ----------------------------------------
Command "/usr/bin/python3 -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-1c8u3k0k/curtin/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-oup3eg2d-record/install-record.txt --single-version-externally-managed --no-compile --user --prefix=" failed with error code 1 in /tmp/pip-build-1c8u3k0k/curtin/
Command '['/bin/sh', '/tmp/tmpavrrdjm4', 'pip', 'install', '--user', '--no-compile', '--disable-pip-version-check', '--no-index', '--find-links', '/home/blake/Code/maas/maas-snap-early/parts/maas/packages', '--requirement', '/home/blake/Code/maas/maas-snap-early/parts/maas/src/requirements.txt']' returned non-zero exit status 1

snapcraft/tests/__init__.py
```diff
@@ -26,6 +26,12 @@
 from snapcraft.tests import fixture_setup
 
 
+class ContainsList(list):
+
+        def __eq__(self, other):
+            return all([i[0] in i[0] for i in zip(self, other)])
```
josepht: I have no idea what this is doing.  Well I can gather from the name what it's trying to do but the `i[0] in i[0]` part seems strange to me.

```
>>> s = ['0', '1', '2']
>>> o = ['1', '2', '3']
>>> all([i[0] in i[0] for i in zip(s, o)])
True
>>> all([i[0] in i[1] for i in zip(s, o)])
False
>>> [i[0] in i[0] for i in zip(s, o)]
[True, True, True]
>>> [(i, i[0] in i[0]) for i in zip(s, o)]
[(('0', '1'), True), (('1', '2'), True), (('2', '3'), True)]
>>> t = ['a', 'b', 'c']
>>> [(i, i[0] in i[0]) for i in zip(s, t)]
[(('0', 'a'), True), (('1', 'b'), True), (('2', 'c'), True)]
```

sergiusens: doh, should be i[0] in i[1]

snapcraft/tests/__init__.py
```diff
@@ -26,6 +26,12 @@
 from snapcraft.tests import fixture_setup
 
 
+class ContainsList(list):
```
josepht: This should have a couple of tests I think. ;-)

blakerouse:
So it does work for the python project but it does not work when a dependency in a requirements.txt file defines a data_files.

What is also weird is the location the files are placed. I would expect that it be placed into parts/part-name/install/ but instead it installs at parts/part-name/install/usr/lib/python3/site-packages/. I know you mentioned this is a limitation of setuptools, but that is very strange and not what developer will expected. This issue can be worked around by using organize, but that will require organize also for dependencies, that will require the developer to know more than they might want about their dependencies.

blakerouse:
The last commit did fix it for dependencies, but its still all under site-packages.

blakerouse:
Using organize with this patch works very well. Only thing that is not working now is cleanbuild as snapcraft does not copy itself into the container it uses the one for the archive.

elopio:
retest this please

snapcraft/plugins/python.py
```diff
@@ -277,17 +287,44 @@ def __init__(self, *, exec_func, runnable, package_dir, env,
         if dependency_links:
             self._extra_pip_args.append('--process-dependency-links')
 
+    def list(self, exec_func=None):
```
elopio: coming late to the party, but this name is not good. I would expect something closer to the docstring, like get_installed_python_packages. Also, why is this not marked as internal with the _ in front?

sergiusens: It replicates the `pip list` command, just like `install`, `download` and `wheel`. It would be inconsistent to change

elopio: Ah, I see. I didn't notice this was inside the Pip class. It is good, I'm sorry.

elopio:
I ran the integration tests in a yakkety lxc, and the snaps tests in a yakkety kvm. All pass.

sergiusens:
retest this please

blakerouse:
I tested this and it still works correctly. The integration test in this branch tests this well as it has the exact issues I was hitting with the MAAS snap.

elopio:
retest this please

PR #863, opened 2016-10-11
history/status alignment fixes.

cprov:
Also benefiting of RFC3339 (snapd/SAS) timestamps from the Store.

LP: #1632155


elopio:
ok to test

PR #865, opened 2016-10-12
Add `init-plugin` 

jaymell:
Hi: This PR adds 'init-plugin' (LP# 1629910), which creates a plugin in the appropriate place after running:

```snapcraft init-plugin <plugin-name>```

It includes test cases. A couple of comments:

* I tried to put a short example in the comments that provides some clues as to how to build a plugin, but it's also somewhat contrived, and could arguably be pared down a bit and/or have more relevant information. Any thoughts?

* Code -- would it be preferable for this code to go in a separate file rather than cluttering up main.py?

Any other suggestions would be most welcome -- I'd really like for this to be useful and maintainable.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -23,7 +23,7 @@ class CraftyPlugin(snapcraft.BasePlugin):
 
         # Add another option for our custom target.
         schema['properties']['crafty-target'] = {
-            type: 'string'
+            'type': 'string'
         }
 
         # Our custom target is required.
@@ -32,6 +32,9 @@ class CraftyPlugin(snapcraft.BasePlugin):
         return schema
 
     def build(self):
+        # Use base implementation to set up the 'build' directory.
+        super().build()
+        # Run custom build command.
         return self.run(['crafty', self.options.crafty_target])
 ```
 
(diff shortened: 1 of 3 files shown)


jaymell:
I was working on an update to this to store the custom plugin code in a file under, say, /usr/share/snapcraft/custom_plugin/ rather than storing it directly in code. I'd originally thought it might be useful to dynamically generate a list of methods provided by BasePlugin to place in the plugin comments, but since doing so doesn't seem particularly helpful, I suppose there's not much benefit to storing the code as a string.

Perhaps that would make this PR more palatable -- or maybe I should just close it? :)

elopio:
Hey @jaymell. Thanks for your contribution, and sorry for the late answer.
We left this unreviewed because this was a topic we had to discuss on the sprint last week. We should have commented here about that.

During the meeting we agreed to clean up the base plugin, and then generate a template like you did. There's still work to do, but we will appreciate your comments once the PR is ready. Just subscribe to the bug in launchpad and you will get a notification.

pura vida

jaymell:
This has been interesting problem to investigate -- will be excited to see the end product. Cheers!

PR #866, opened 2016-10-13
Login with option to agree to terms of service and human friendly errors

psivaa:
Storeside changes to introduce graceful failure messages and an option to sign developer agreement via snapcraft CLI. Also includes some refactoring in the surrounding area to move messages/errors/warnings into 'storeapi/constants.py'

LP: #1619193


psivaa:
The static checks and the tests have passed now. The relevant change needed in the server side has also landed. So if the review is OK, this can land.

snapcraft/storeapi/errors.py
```diff
@@ -87,20 +87,32 @@ class StoreMacaroonNeedsRefreshError(StoreError):
     fmt = 'Authentication macaroon needs to be refreshed.'
 
 
+class DeveloperAgreementSignError(StoreError):
+
+    fmt = 'Developer agreement sign error: {message}'
```
sergiusens: This sentence needs to be more human friendly

snapcraft/_store.py
```diff
@@ -50,6 +50,59 @@ def _get_data_from_snap_file(snap_path):
...
+    return False
+
+
+def _get_url_from_error(error):
+    if error.extra:
+        return error.extra[0].get('url')
+    return None
+
+
+def _check_dev_agreement_and_namespace_statuses(store):
```
sergiusens: Can we just raise exceptions if the conditions are not met. The exception would carry a friendly error message we would raise to the user just like we do for all others

snapcraft/_store.py
```diff
@@ -50,6 +50,59 @@ def _get_data_from_snap_file(snap_path):
...
+    except storeapi.errors.StoreAccountInformationError as e:
+        if storeapi.constants.MISSING_NAMESPACE in e.error:
+            # A precaution if store does not return new style error.
+            url = (_get_url_from_error(e) or
+                   storeapi.constants.UBUNTU_STORE_ACCOUNT_URL)
+            return _fail_login(
+                storeapi.constants.NAMESPACE_ERROR.format(url))
+        else:
+            raise
+    return True
```
sergiusens: let's use exeptions instead

snapcraft/_store.py
```diff
@@ -59,19 +112,23 @@ def _login(store, acls=None, save=True):
...
                 email, password, one_time_password=one_time_password,
                 acls=acls, save=save)
-    except (storeapi.errors.InvalidCredentialsError,
-            storeapi.errors.StoreAuthenticationError):
-        print()
-        logger.info('Login failed.')
-        return False
+
+        # Continue if agreement and namespace conditions are met.
+        if not _check_dev_agreement_and_namespace_statuses(store):
```
sergiusens: just deal with the exception in the exception catching logic a couple lines below

snapcraft/_store.py
```diff
@@ -50,6 +50,59 @@ def _get_data_from_snap_file(snap_path):
...
+            url = (_get_url_from_error(e) or
+                   storeapi.constants.UBUNTU_STORE_ACCOUNT_URL)
+            return _fail_login(
+                storeapi.constants.NAMESPACE_ERROR.format(url))
+        else:
+            raise
+    return True
+
+
 def _login(store, acls=None, save=True):
```
sergiusens: I don't mind if this just raises an exception that is friendly enough and this `return _login(store)` is refactored a bit.

This may not be a task for you and you may not want to do it, but we have a lot of double logic with exceptions and if True/False around.

snapcraft/storeapi/__init__.py
```diff
@@ -609,6 +612,20 @@ def close_channels(self, snap_id, channel_names):
...
+        data = {'latest_tos_accepted': latest_tos_accepted}
+        response = self.post(
+            'agreement/', data=json.dumps(data),
+            headers={'Authorization': auth,
+                     'Content-Type': 'application/json',
+                     'Accept': 'application/json'})
+
+        if not response.ok:
+            raise errors.DeveloperAgreementSignError(
+                'Failed to sign developer ToS.')
```
sergiusens: Hmm, `response.ok` not being ok may not be an indication of failing to sign the developer agreement and a 503 instead

snapcraft/tests/fake_servers.py
```diff
@@ -669,6 +676,28 @@ def _handle_release_request(self):
 
         self.wfile.write(data)
 
+    def _handle_sign_request(self):
+        string_data = self.rfile.read(
+            int(self.headers['Content-Length'])).decode('utf8')
+        data = json.loads(string_data)
+        if data['latest_tos_accepted'] is not True:
+            self.send_response(400)
```
sergiusens: Can we add a 500 case?

kyrofa:
Would you mind please updating the PR title to be more descriptive? The bug reference should be in the PR description and the commit, not the title.

snapcraft/_store.py
psivaa: Agreed and only use exception here.

snapcraft/_store.py
psivaa: I personally don't mind doing this. But moving the exception handling to `login()` for refactoring `return _login(store)` will imho shift the complexity to `login`. So i don't feel there is much gain here. But if you still insist on it, i can happily do that :)

snapcraft/storeapi/__init__.py
psivaa: Right. The error message was just an indicator. Now use an exception to use the response itself.

snapcraft/storeapi/errors.py
psivaa: Fixed

snapcraft/tests/fake_servers.py
psivaa: done.

psivaa:
Thanks for the reviews. I've addressed them and replied inline. Would appreciate a re-review please. 
Thanks

snapcraft/_store.py
```diff
@@ -59,19 +112,24 @@ def _login(store, acls=None, save=True):
...
+        _check_dev_agreement_and_namespace_statuses(store)
+
+    except storeapi.errors.InvalidCredentialsError:
+        return _fail_login(storeapi.constants.INVALID_CREDENTIALS)
+    except storeapi.errors.StoreAuthenticationError:
+        return _fail_login(storeapi.constants.AUTHENTICATION_ERROR)
+    except storeapi.errors.StoreAccountInformationError:
+        return _fail_login(storeapi.constants.ACCOUNT_INFORMATION_ERROR)
+    except storeapi.errors.NeedTermsSignedError as e:
+        return _fail_login(e.message)
```
josepht: Any reason to not include this in storeapi.constants as well?

snapcraft/storeapi/constants.py
```diff
@@ -25,3 +25,29 @@
...
+MISSING_AGREEMENT = 'Developer has not signed agreement.'
+MISSING_NAMESPACE = 'Developer profile is missing short namespace.'
+AGREEMENT_ERROR = (
+    'You must agree to the developer terms and conditions to upload snaps.')
+NAMESPACE_ERROR = (
+    'You need to set a username. It will appear in the developer field '
+    'alongside the other details for your snap. Please visit {} and login '
+    'again.')
+AGREEMENT_INPUT_MSG = (
+    'Do you agree to the developer terms and conditions. ({})? [y/N]')
```
josepht: I'm not a fan of mixing formatting in the constants.  It's not a blocker for me just an opinion.

snapcraft/storeapi/errors.py
```diff
@@ -87,20 +87,45 @@ class StoreMacaroonNeedsRefreshError(StoreError):
     fmt = 'Authentication macaroon needs to be refreshed.'
 
 
+class DeveloperAgreementSignError(StoreError):
+
+    fmt = (
+        'There was an error whilst signing developer agreement.\n'
```
josepht: s/whilst/while/

snapcraft/storeapi/constants.py
psivaa: Yea, I agree that its not perfect.  But I think keeping them here will make the actual logic more readable.

snapcraft/_store.py
psivaa: The `NeedTermsSignedError` will contain different messages based on the error in `_check_dev_agreement_and_namespace_statuses`.

snapcraft/storeapi/errors.py
psivaa: Fixed. Thanks

snapcraft/storeapi/__init__.py
```diff
@@ -286,6 +286,9 @@ def push_validation(self, snap_id, assertion):
     def get_validations(self, snap_id):
         return self.sca.get_validations(snap_id)
 
+    def sign_developer_agreement(self, latest_tos_accepted=True):
```
sergiusens: shouldn't `latest_tos_accepted` be `False` by default?

snapcraft/storeapi/__init__.py
```diff
@@ -618,6 +621,19 @@ def close_channels(self, snap_id, channel_names):
                                    response.content))
             raise errors.StoreChannelClosingError(response)
 
+    def sign_developer_agreement(self, latest_tos_accepted=True):
```
sergiusens: ditto on `latest_tos_accepted`

snapcraft/storeapi/__init__.py
psivaa: Yes, fair point. Thanks. Changed.

snapcraft/storeapi/__init__.py
psivaa: Changed. Thanks

psivaa:
Thanks for the comments. Pushed the fix for them.

PR #868, opened 2016-10-13
Parametrize call args for pluginhandler

sergiusens:
LP: #1633163

Signed-off-by:` Sergio Schvezov <sergio.schvezov@ubuntu.com>


sergiusens:
retest this please

PR #871, opened 2016-10-23
Remove XXX comment from stage-packages

nhandler:
This comment is currently visible to the public at http://snapcraft.io/docs/reference/plugins/common . There is no need for it anymore. This commit removes it and adds a few additional details.


kyrofa:
Might I suggest the following (which includes changes to `build-packages`:

    - build-packages: [deb, deb, deb...]

    A list of Ubuntu packages to install on the build host before building
    the part. The files from these packages typically will not go into the
    final snap unless they contain libraries that are direct dependencies of
    binaries within the snap (in which case they'll be discovered via `ldd`),
    or they are explicitly described in stage-packages.
    
    - stage-packages: [deb, deb, deb...]

    A list of Ubuntu packages to be downloaded and unpacked to join the part
    before it's built. Note that these packages are not installed on the host.
    Like the rest of the part, all files from these packages will make it into
    the final snap unless filtered out via the `snap` keyword.

sergiusens:
Closing in favor of #913 due to lack of activity.

PR #872, opened 2016-10-23
Add some further bash-completion

cwayne18:
Autocomplete the plugins after 'help'
Autocomplete directory names after 'snap' or 'tour'
Autocomplete all '.snap' files after 'gated, upload, push, sign-build, register' or 'validate'

### DIFF
--- a/debian/snapcraft-completion
+++ b/debian/snapcraft-completion
@@ -1,12 +1,33 @@
 _snapcraft()
 {
     local cur prev opts
+    _init_completion -s || return
+
     COMPREPLY=()
     cur="${COMP_WORDS[COMP_CWORD]}"
     prev="${COMP_WORDS[COMP_CWORD-1]}"
     opts="help init list-plugins login logout list-keys keys create-key register-key register tour push release clean cleanbuild pull build sign-build stage prime snap update define search gated validate history status close"
 
+    case "$prev" in
+    help)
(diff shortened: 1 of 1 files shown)


elopio:
Hey @cwayne18, I have a question and since you are our autocomplete expert... How would autocomplete work if snapcraft is not installed as a deb? For example, if it comes from a snap, or pip.

cwayne18:
@elopio I suppose for pip we could include the bash-completion file in data_files in setup.py, as for snaps, not entirely sure the best way to go about it yet

sergiusens:
Sorry for taking so long to get to this. `data_files` is a rather dead concept in the `pip` world, if you use it, it would land in `site-packages` going forward.

For snapcraft as a snap we need to have some form of feature work with the snappy core team.

debian/snapcraft-completion
```diff
@@ -1,12 +1,33 @@
...
+    help)
+        plugins=$(snapcraft list-plugins)
+        COMPREPLY=( $(compgen -W "$plugins" -- $cur))
+        return 0
+        ;;
+    snap | tour)
+        _filedir -d
+        return 0
+        ;;
+    gated | upload | push | sign-build | register | validate)
```
sergiusens: gated register and validate shouldn't use snap-file names.

PR #875, opened 2016-10-27
Add the test upstream rules

elopio:


elopio:
I'm merging this because it was already reviewed by pitti. I'm just recovering it after updating the debian branch.

PR #876, opened 2016-10-27
repo: allow for architecture-specific stage-packages

stgraber:
This makes it possible to instruct snapcraft to only install some
packages on a given architecture or set of architectures by making a part specific to a particular architecture.

LP: #1637282

Signed-off-by: Stéphane Graber stgraber@ubuntu.com


kyrofa:
Thanks for the PR @stgraber. I'd like to wait on @sergiusens to yay/nay the design, but I'll point out that tests are missing.

kyrofa:
Note that in the previous sprint we did discuss arch-specific options in the YAML, and this may fall under the scope of that work.

stgraber:
Got sick of waiting for that Docker container to download bits from the UK, looks like using the US mirror makes things a lot more likely to succeed in time :)

stgraber:
And now travis and coverails are happy.

sergiusens:
During last weeks sprint we wanted to introduce a core functionality that would allow you to do:

```yaml
key-attribute-name:
     arch-1:
         <key-attribute-values>
     arch-2:
         <key-attribute-values>
```

I guess this removes the facility when dealing with `stage-packages` and `build-packages`. The language you propose is also familiar to people dealing with packaging, although does not translate well into other things, so maybe the alternative to our big yaml map we could do:

```yaml
part1:
    architectures: [amd64, armhf]
    source: https://my-source.tar.xz
    plugin: autotools
part2:
    architectures: [ppcel64]
    source: https://my-source.tar.xz
    plugin: autotools
    configflags:
        - --disable-weird-feature-for-ppcel64
```

sergiusens:
I forgot to mention, the latter would fit better with this current PR even though there might be some overlap here and there.

sergiusens:
For kernel build targets we already have:

```yaml
kernel-image-target:
  amd64: bzImage
  armhf: Image
```

stgraber:
I'll update the branch to add a "architectures" field to parts that contains a list of architectures to build the part for, that seems reasonable (will require us to define a separate part just for criu, but that's no big deal).

stgraber:
@sergiusens ok, that's it re-implemented as an architecture list against parts. Confirmed to work for my scenario and tests look happy too.

sergiusens:
The proposal is to do it like this:

```yaml
stage-packages: [package1/arch1, package1/arch2, package2, package3/arch1]
```

There is no typo in package# here.

This avoids double lists which look confusing in the one liner syntax

The per part architecture has been nacked.

stgraber:
@sergiusens so the potential problem with this approach is that the "name/xyz" syntax already exists in apt to request a package to come from a particular pocket, like "lxd/xenial-proposed" which will likely confuse some users and also prevent you from actually implementing the "name/pocket" syntax down the road if there's a use for it.

sergiusens:
El 03/11/16 a las 17:40, Stéphane Graber escribió:
>
> @sergiusens <https://github.com/sergiusens> so the potential problem 
> with this approach is that the "name/xyz" syntax already exists in apt 
> to request a package to come from a particular pocket, like 
> "lxd/xenial-proposed" which will likely confuse some users and also 
> prevent you from actually implementing the "name/pocket" syntax down 
> the road if there's a use for it.
>

You are indeed correct!

I can think of using `:` then as it is "canonical" syntax from apt

CCing Gustavo

stgraber:
Yeah, I was going to suggest using a colon which is the existing deb syntax.

stgraber:
Going to dig out my previous implementation and change it to support "package:arch"

stgraber:
And done, hopefully CI will be happy :)

stgraber:
And looks like we're all good again. @sergiusens

PR #877, opened 2016-10-28
Add new "source-commit" field for VCS sources

stgraber:
This is very much a work in progress. It's an attempt to fix https://bugs.launchpad.net/snapcraft/+bug/1632126 by introducing a new source-commit field to snapcraft.yaml

This pull request is primarily there to get clean test results.


stgraber:
Will try to figure out some tests to make coverails happy tomorrow (and also actually test that the change works for my use case).

stgraber:
Looks like this is ready for review, been confirmed to work for me and has tests.

sergiusens:
Looks good.
ok to test

It would be nice to also see the integration tests to follow this, here's the git one https://github.com/snapcore/snapcraft/blob/master/integration_tests/test_git_source.py

stgraber:
Ok, I'll look at updating the existing integration tests to cover this change too.

stgraber:
Hmm, not sure how I'd do it for the git case though since I don't know the git hash ahead of time...

I should be able to add tests for bzr and mercurial though.

stgraber:
Using "HEAD^" as the commit string for the git test, that's not ideal but should still be good enough for the integration test.

stgraber:
And looks like all tests passed with that trick. So @sergiusens that's the branch now with integration tests for bzr, git and hg.

snapcraft/tests/test_base_plugin.py
```diff
@@ -172,34 +172,52 @@ class GetSourceTestCase(tests.TestCase):
...
+            'error': 'source-commit'})
     ]
 
     @unittest.mock.patch('snapcraft.internal.sources._check_for_package')
     def test_get_source_with_branch_must_raise_error(self, mock_check):
         mock_check.side_effect = None
         options = tests.MockOptions('lp:this', self.source_type,
-                                    self.source_branch, self.source_tag)
+                                    self.source_branch, self.source_tag,
+                                    source_commit=self.source_commit)
```
kyrofa: The inconsistency here is bothering me. I suggest we either make all the arguments positional or keyword.

kyrofa:
I have a few comments, but this looks very good. Thanks @stgraber!

.travis.yml
```diff
@@ -21,7 +21,7 @@ install:
   - sudo apt-get -qq update
   - sudo apt-get install -y python3-coverage
 script:
-  - docker run -e TEST_USER_EMAIL=$TEST_USER_EMAIL -e TEST_USER_PASSWORD=$TEST_USER_PASSWORD -e TEST_STORE=$TEST_STORE -v $(pwd):$(pwd) -t ubuntu:xenial sh -c "export LC_ALL=en_US.UTF-8 && locale-gen en_US.UTF-8 && apt update && cd $(pwd) && $DEPENDENCIES && ./runtests.sh $TEST_SUITE"
+  - docker run -e TEST_USER_EMAIL=$TEST_USER_EMAIL -e TEST_USER_PASSWORD=$TEST_USER_PASSWORD -e TEST_STORE=$TEST_STORE -v $(pwd):$(pwd) -t ubuntu:xenial sh -c "export LC_ALL=en_US.UTF-8 && locale-gen en_US.UTF-8 && sed -i s/archive.ubuntu.com/us.archive.ubuntu.com/g /etc/apt/sources.list && apt update && cd $(pwd) && $DEPENDENCIES && ./runtests.sh $TEST_SUITE"
```
stgraber: Makes the testsuite run about 10min faster for me :)

The Docker images defaults to the UK archive mirror but Travis runs everything in GCE on the US east coast, so the US archive mirror is just a few ms away.

The US mirror pool is an official mirror pool maintained by Canonical and so is guaranteed to have the exact same content at the exact same time as the UK pool.

snapcraft/tests/test_base_plugin.py
sergiusens: fwiw, I was ok with how this was, but don't sweat about it, this source class it is under refactor.

PR #878, opened 2016-10-28
Added a fix for cases where modpbrobe append options to the line, ie.…

croepha:
… the module path is not the last field, it is always the second...


kyrofa:
This will need @sergiusens to take a look to verify the fix itself, but it's failing both static and unit tests. Please take a look.

croepha:
Yes, it looks like in need to delete some blank lines... sorry about that, I'm a bit busy, but I'll try to fix this up soon.

sergiusens:
@croepha any updates on this?

croepha:
@sergiusens  I removed the lines, `./runtests.sh` still complains about things that I didn't touch though

sergiusens:
hm, tests related to the code you touched do fail:

```

ERROR: test_build_with_kconfigfile_and_modules (snapcraft.tests.test_plugin_kernel.KernelPluginTestCase)

  File "/home/travis/build/snapcore/snapcraft/snapcraft/tests/test_plugin_kernel.py", line 672, in test_build_with_kconfigfile_and_modules

    plugin.build()

  File "/home/travis/build/snapcore/snapcraft/snapcraft/plugins/kbuild.py", line 169, in build

    self.do_install()

  File "/home/travis/build/snapcore/snapcraft/snapcraft/plugins/kernel.py", line 348, in do_install

    self._make_initrd()

  File "/home/travis/build/snapcore/snapcraft/snapcraft/plugins/kernel.py", line 253, in _make_initrd

    for src in set(_.split()[1] for _ in modprobe_outs):

  File "/home/travis/build/snapcore/snapcraft/snapcraft/plugins/kernel.py", line 253, in <genexpr>

    for src in set(_.split()[1] for _ in modprobe_outs):

IndexError: list index out of range

```

croepha:
So, im not sure if I am interpreting the test properly, but it looks like its trying to pass `modprobe` output as `/tmp/tmp14ecfxht/parts/test-part/install/lib/modules/4.4.2/some-module.ko`  (i.e. just one field.. but I haven't seen `modprobe` output that format... Its format looks like this: ```
builtin squashfs
insmod /lib/modules/4.4.0-45-generic/kernel/drivers/gpu/drm/ttm/ttm.ko
insmod /lib/modules/4.4.0-45-generic/kernel/drivers/gpu/drm/vmwgfx/vmwgfx.ko enable_fbdev=1
```
I could add a conditional to detect cases where there is just one field, and that would make the test pass, but what if there were module arguments in that case? There wouldn't be a way to distinguish the arguments from the module path... 
I would say either the test is broken, or I'm not understanding what is going on here...

croepha:
it just occurred to me that I could just search for first and second fields for a `.ko` extension, but to me this seem like trying to solve a simple problem by adding a lot of complexity...

sergiusens:
@lool hey, while you are at it, mind confirming this solution? I can fix the tests.

lool:
The approach looks good, looking at the proposed changes (BTW there's a floating newline still) and the modprobe output:
modprobe -n --show-depends -d parts/kernel/build/initrd-staging/ -S 3.18.25-OpenNetworkLinux squashfs
insmod /mnt/onl-snap/parts/kernel/build/initrd-staging//lib/modules/3.18.25-OpenNetworkLinux/kernel/fs/squashfs/squashfs.ko

However I didn't debug why the tests aren't passing.

kyrofa:
Where does this PR stand?

croepha:
Im working on other things at the moment, so I don't have time to check to see if these changes still work with the current code base, and I'm not sure the best way to proceed.  If it were up to just me, I would modify the tests so that my code passes.

sergiusens:
Rebased and fixed tests in #1208

PR #879, opened 2016-11-01
The latest icon definition is deprecated

liu-xiao-guo:
"icon" is not supported in the latest snapcraft. The correct way is to define it in "setup/gui" like what is described in https://github.com/snapcore/snapcraft/blob/master/docs/your-first-snap.md


sergiusens:
As mentioned in the bug replies, can we keep this about fixing the typos and documentation update but not remove the icon entry, it *should* be working.

liu-xiao-guo:
I think it should be fine. Thanks

sergiusens:
ok, feel free to open the PR again once the change on `docs/snapcraft-syntax.md` has been reverted.

PR #880, opened 2016-11-01
Remove license concepts

sergiusens:
LP: #1638303


schema/snapcraft.yaml
```diff
@@ -57,16 +57,6 @@ properties:
...
-  license:
-    type: string
-    description: path to a license file.
-  license-agreement:
-    type: string
-    description: action to take on licensing.
-    enum:
-      - explicit
-  license-version:
-    description: license version (used to re-trigger an agreement action).
```
kyrofa: Won't removing this from the schema cause the `snapcraft.yaml`s out there that are using this to actually break? What if we left them along with the deprecation warning, but rephrased the deprecation warning to say "this doesn't actually do anything anymore"?

sergiusens: Yeah, what a horrible world to live in :-/

kyrofa: On the bright side, once we can deliver as a snap we can jump major version numbers without worrying about SRUs.

sergiusens: I don't think we need a warning, this should be a flat out breakage as if not it would create the expectation that it works when it clearly doesn't

PR #883, opened 2016-11-01
python plugin: wheel and install in the proper order

sergiusens:
LP: #1638389


kyrofa:
Although you're failing static tests.

PR #884, opened 2016-11-02
Implement API packaging plugin as requested in #1638508

bzoltan1:
We have discussed with Sergio/Colin on the sprint about the need of API packages.

The conclusion was that since we do not want -dev snaps in the store we must provide an alternative way to publish API packages.

The story is that an upstream content interface plug developer (for example Qt or Gtk)  produces boththe runtime interface and the development interfaces. The runtime interfaces go with the snap package. This PR contains the simple implementation of a snapcraft plugin (development) what creates a tar.gz of the stage space. This tar.gz can be published as the development interfac counterpart of the runtime interfaces.

### DIFF
--- a/integration_tests/snaps/simple-development/snapcraft.yaml
+++ b/integration_tests/snaps/simple-development/snapcraft.yaml
@@ -1,0 +1,22 @@
+name: libapitest
+version: 0.1
+summary: Testing api packaging
+description: Testing api packaging
+confinement: strict
+
+build-packages: [gcc, libc6-dev]
+
+parts:
+  project:
+    plugin: make
+    source: .
+
+  api:
+    plugin: development
+    source: .
+    exclude:  ['Makefile','*.c','*.o', '*tar.gz', '*.snap']
+    extension: api
+    after:
+      - project
+ 
+
--- a/integration_tests/test_development_plugin.py
+++ b/integration_tests/test_development_plugin.py
@@ -1,0 +1,49 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2015, 2016 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+
+from testtools.matchers import (
+    DirExists,
+    FileExists
+)
+
+import integration_tests
+
+
+class DevelopmentPluginTestCase(integration_tests.TestCase):
+
+    def test_stage_developer_plugin(self):
+        project_dir = 'simple-development'
+        self.run_snapcraft('stage', project_dir)
+
+        expected_files = [
+            'libapitest.c',
+            'libapitest.h',
+            'libapitest.o',
+            'libapitest.so',
+            'libapitest.so.1',
+            'libapitest.so.1.0.1',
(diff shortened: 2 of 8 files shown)


snapcraft/plugins/development.py
```diff
@@ -0,0 +1,92 @@
...
+        pass
+
+    def build(self):
+        super().build()
+        self.archive_name = os.getcwd() +\
+            "/" +\
+            snapcraft.internal.load_config(self.project).data['name'] +\
+            "-" +\
+            self.options.extension +\
+            ".tar.gz"
```
elopio: You don't need \ here. finishing the line in + will just work.

snapcraft/plugins/development.py
```diff
@@ -0,0 +1,92 @@
...
+        for i, p in enumerate(self.options.exclude):
+            self.options.exclude[i] = "--exclude={}".format(p.strip())
+
+        if os.path.isdir(self.project.stage_dir):
+            self.run(['/bin/tar',
+                      'czf',
+                      self.archive_name] +
+                     self.options.exclude +
+                     ['-C',
+                      '%s' % self.project.stage_dir, '.'])
```
elopio: This means that by the time the this part runs, the stage dir needs to be full with the contents from the other parts.
The order in which the parts run is not deterministic. This means that this part needs to have the after keyword including all the other parts. If you forget one, or add a new part and don't update this one, you will get the wrong tar.

Maybe we could hack the after keyword on the init of this plugin, but I'm not sure it's a good idea.
Maybe a better idea would be to add a new step on the lifecycle, that tars stage after all the parts ran?
@kyrofa is on holidays and @sergiusens is in a spring, so we should wait a little for their reviews.

snapcraft/tests/test_plugin_development.py
```diff
@@ -0,0 +1,159 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2015 Canonical Ltd
```
elopio: 2016

integration_tests/test_development_plugin.py
```diff
@@ -0,0 +1,49 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2015, 2016 Canonical Ltd
```
elopio: Just  2016.

integration_tests/snaps/simple-development/snapcraft.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+  api:
+    plugin: development
+    source: .
+    exclude:  ['Makefile','*.c','*.o', '*tar.gz', '*.snap']
+    extension: api
+    after:
+      - project
+ 
+
```
elopio: Please remove the extra newlines at the end.

integration_tests/test_development_plugin.py
```diff
@@ -0,0 +1,49 @@
...
+            'libapitest.so',
+            'libapitest.so.1',
+            'libapitest.so.1.0.1',
+            'main.c',
+            'Makefile',
+        ]
+        for expected_file in expected_files:
+            self.assertThat(
+                os.path.join(project_dir, 'stage', expected_file),
+                FileExists())
```
elopio: Instead of this, I would untar libapitest-api.tar.gz and verify that it contains the expected files

elopio:
ok to test

elopio:
You have static errors:
snapcraft/tests/test_plugin_development.py:31:1: F811 redefinition of unused 'development' from line 26
snapcraft/plugins/development.py:36:1: F401 'subprocess' imported but unused
integration_tests/test_development_plugin.py:19:1: F401 'testtools.matchers.DirExists' imported but unused

You can ./runtests.sh static.

sergiusens:
I am closing this until we have an initial discussion about this with @evandandrea, you and me.

snapcraft/plugins/development.py
bzoltan1: the `./runtests.sh static ` was giving 'E901 SyntaxError: invalid syntax' for the line when I have removed the \

I found that \ thing strange too, but only this way keeps both pep and the test happy

PR #886, opened 2016-11-03
Support for gadget snaps

sergiusens:
LP: #1638824


ogra1:
it might make sense to mention the path in the error message, nothing makes clear that gadget.yaml needs to live in a meta/ subdir.

sergiusens:
@ogra1 it says clearly, "in the root of your snapcraft project"

PR #887, opened 2016-11-03
Cache apt related files (config, packages) in 'apt' parent directory.

squidsoup:
LP: #1635611

Apt related cache now lives in `$HOME/.cache/snapcraft/apt/sources-list-digest/[apt, packages]` allowing for new top-level cache dirs for snapcraft unrelated to apt.


sergiusens:
thanks for this!

ok to test

PR #888, opened 2016-11-04
Always respect go-buildtags

stgraber:
There is no mention in documentation that go-buildtags should only apply
to building from the local tree and since those tags are global to the
build anyway, it's best to always pass them.

One need for this change is the LXD snap.

Signed-off-by: Stéphane Graber <stgraber@ubuntu.com>


stgraber:
Before:
```
root@snapcraft:~/snappy-16# ldd prime/bin/lxd | grep sqlite3
root@snapcraft:~/snappy-16#
```

After:
```
root@snapcraft:~/snappy-16# ldd prime/bin/lxd | grep sqlite3
        libsqlite3.so.0 => /usr/lib/x86_64-linux-gnu/libsqlite3.so.0 (0x00007f2297c28000)
root@snapcraft:~/snappy-16#
```

stgraber:
This tiny change makes LXD about 20x faster by not using the bundled libsqlite3 but instead causing Go to link against the system one.

stgraber:
https://github.com/lxc/lxd-pkg-ubuntu/blob/snappy-16/snapcraft.yaml is the snapcraft.yaml which showed us this issue.

PR #889, opened 2016-11-07
cache: snap revision caching on 'push'

squidsoup:
This feature will only be enabled with the `DELTA_UPLOADS_EXPERIMENTAL` environment variable set.

I was a little uncertain where `cache_snap()` should live; `repo` seemed like the most appropriate place, but perhaps you could provide some guidance there. Other thoughts are perhaps this should be SnapCache (similar to AptCache), rather than a solitary func, as we'll be providing pruning in a following branch (see discussion on Rocketchat)?

LP: #1639672

### DIFF
--- a/snapcraft/_store.py
+++ b/snapcraft/_store.py
@@ -30,7 +30,10 @@ from tabulate import tabulate
 import yaml
 
 from snapcraft import storeapi
-from snapcraft.internal import repo
+from snapcraft.internal import (
+    cache,
+    repo,
+)
 
 
 logger = logging.getLogger(__name__)
@@ -393,6 +396,10 @@ def push(snap_filename, release_channels=None):
     else:
         logger.info('Uploaded {!r}'.format(snap_name))
     tracker.raise_for_code()
+
+    if os.environ.get('DELTA_UPLOADS_EXPERIMENTAL'):
+        snap_cache = cache.SnapCache()
+        snap_cache.cache(snap_filename, result['revision'])
 
     if release_channels:
         release(snap_name, result['revision'], release_channels)
--- a/snapcraft/internal/cache/__init__.py
+++ b/snapcraft/internal/cache/__init__.py
@@ -14,6 +14,4 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-from snapcraft.internal import cache             # noqa
-from snapcraft.internal import states            # noqa
-from snapcraft.internal.project_loader import load_config  # noqa
+from ._snap import SnapCache  # noqa
--- a/snapcraft/internal/cache/_cache.py
+++ b/snapcraft/internal/cache/_cache.py
@@ -1,0 +1,45 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2016 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+
+from xdg import BaseDirectory
+
+import snapcraft
+
+
+class SnapcraftCache:
+    """Generic cache base class.
+
+    This class is responsible for cache location, notification and pruning.
+    """
+    def __init__(self):
+        self.cache_root = os.path.join(
+            BaseDirectory.xdg_cache_home, 'snapcraft')
+
+    def cache(self):
+        raise NotImplementedError
+
+    def prune(self):
+        raise NotImplementedError
+
+
+class SnapcraftProjectCache(SnapcraftCache):
+    """Project specific cache"""
+    def __init__(self):
+        super().__init__()
+        self.project_cache_root = os.path.join(
+            self.cache_root, snapcraft.internal.load_config().data['name'])
--- a/snapcraft/tests/test_commands_push.py
+++ b/snapcraft/tests/test_commands_push.py
@@ -22,11 +22,13 @@ from unittest import mock
 
 import docopt
 import fixtures
+from xdg import BaseDirectory
 
 from snapcraft import (
     storeapi,
     tests
 )
+from snapcraft.internal.cache._snap import _rewrite_snap_filename_with_revision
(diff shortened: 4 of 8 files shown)


squidsoup:
Actually, if you can hold off reviewing this, I need to make some changes (namely, revision should be added to the cached snap's filename, and revisions should be cached in a project/snap-name parent directory).

squidsoup:
@seawaywen yep, that seems sensible.

squidsoup:
@sergiusens has provided some direction for this feature, and this branch will need refactoring, but I'll leave the PR open.

squidsoup:
@sergiusens have updated the branch - is that roughly what you had in mind?

snapcraft/internal/cache.py
```diff
@@ -0,0 +1,80 @@
...
+class SnapcraftCache:
+    """Generic cache base class.
+
+    This class is responsible for cache location, notification and pruning.
+    """
+    def __init__(self):
+        self.cache_root = os.path.join(
+            BaseDirectory.xdg_cache_home, 'snapcraft')
+        self.project_cache_root = os.path.join(
+            self.cache_root, snapcraft.internal.load_config().data['name'])
```
sergiusens: Can this be moved to a `SnapcraftProjectCache` inheriting from this class? This would make moving the apt caching logic cleaner (as it is a global cache).

snapcraft/internal/cache.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def cache(self):
+        pass
```
sergiusens: any benefits of `pass` instead of a `NotImplementedError` here?

snapcraft/internal/cache.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def prune(self):
+        pass
```
sergiusens: We will probably have a base class defined `prune` here so this is :+1:

squidsoup:
@sergiusens great thanks, have moved the project root cache to SnapcraftProjectCache. We also now raise NotImplemented (I took you too literally when you asked for noops on the card :)).

snapcraft/_store.py
```diff
@@ -336,6 +339,10 @@ def push(snap_filename, release_channels=None):
         logger.info('Uploaded {!r}'.format(snap_name))
     tracker.raise_for_code()
 
+    if os.environ.get('DELTA_UPLOADS_EXPERIMENTAL') and 'revision' in result:
```
sergiusens: This may be defensive programming but isn't `revision` always supposed to be in the result and if it isn't it is because it did not pass review for which `tracker.raise_for_code` would have raised an exception by then.

snapcraft/tests/test_commands_push.py
```diff
@@ -245,3 +247,80 @@ def test_push_and_release_a_snap_to_N_channels(self):
         mock_upload.assert_called_once_with('my-snap-name', snap_file)
         mock_release.assert_called_once_with('my-snap-name', 9,
                                              ['edge', 'beta', 'candidate'])
+
+    def test_push_revision_cached_experimental_deltas_enabled(self):
```
sergiusens: How about using test scenarios for this and the following test?

snapcraft/_store.py
squidsoup: It was, but I agree, probably unnessary, thanks.

snapcraft/tests/test_commands_push.py
squidsoup: Yep certainly, I'll have a look at that this morning.

squidsoup: @sergiusens PR updated using test scenarios (thanks for that, that's much nicer!).

snapcraft/internal/cache/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from snapcraft.internal.cache._cache import rewrite_snap_filename_with_revision  # noqa
```
sergiusens: Not sure about this one, and you can weigh in, would this have a better life in the `fileutils` module?

snapcraft/internal/cache/_cache.py
```diff
@@ -0,0 +1,88 @@
...
+        cached_snap_path = os.path.join(snap_cache_dir, cached_snap)
+        try:
+            shutil.copyfile(snap_filename, cached_snap_path)
+        except OSError:
+            logger.warning(
+                'Unable to cache snap {}.'.format(cached_snap))
+        return cached_snap_path
+
+
+def rewrite_snap_filename_with_revision(snap_file, revision):
```
sergiusens: it feels like this needs to be `_rewrite_snap_filename_with_revision`

snapcraft/internal/cache/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+from snapcraft.internal.cache._cache import SnapcraftCache  # noqa
+from snapcraft.internal.cache._cache import SnapcraftProjectCache  # noqa
+from snapcraft.internal.cache._cache import SnapCache  # noqa
```
sergiusens: It seems for now only this guy needs to be exported. If testing is the only reason, using the ugly names with _ seems better.

snapcraft/internal/cache/_cache.py
```diff
@@ -0,0 +1,88 @@
...
+
+class SnapcraftProjectCache(SnapcraftCache):
+    """Project specific cache"""
+    def __init__(self):
+        super().__init__()
+        self.project_cache_root = os.path.join(
+            self.cache_root, snapcraft.internal.load_config().data['name'])
+
+
+class SnapCache(SnapcraftProjectCache):
```
sergiusens: Can this live in its own module together with `rewrite_snap_filename_with_revision`?

snapcraft/internal/cache/_cache.py
squidsoup: @sergiusens thanks, have addressed those change requests.

snapcraft/internal/cache/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from snapcraft.internal.cache._snap import SnapCache  # noqa
```
sergiusens: is there a case here for relative imports? I only ask for this so it feels like a real package

PR #890, opened 2016-11-08
cli: implementing '[list-]registered' command

cprov:
List registered snaps along with their registration status, snap-id and visibility.

LP: #1640237


sergiusens:
Wait, did you check https://trello.com/c/tsuyBVm5/11-allow-seeing-all-the-registered-snaps-i-have-in-the-store ?

This still hasn't been signed off

cprov:
@sergiusens oh, didn't check the card on your board, but it's not far.

Right now, the Store doesn't know about the number of Installs, but I can expose 'Since' and 'Price' on the API and adjust current 'Private' to 'Visibility'.

Where do we stand on exposing 'Snap-id' (arguably long & noisy) ? Also what about registered names in dispute or pending approval ?

sergiusens:
During the last sprint we said we might use `list-registered|registered`. Can you update the UX document as well please?

elopio:
lgtm. I only seems that you are missing the case when no snaps have been registered.

cprov:
@sergiusens UX doc [updated](https://docs.google.com/document/d/1HhJWEBj13xmVSva9MEllr36_CSdJaKRAY-7-oBBGeZs/edit#heading=h.ommeigotreoj) and command renamed to the original suggestion ([list-]registered). The is a WIP MP for the store exposing "Price" and another will follow to expose "Installs" (slightly more complicated).

@elopio good point, missing test added.

snapcraft/_store.py
```diff
@@ -163,6 +164,28 @@ def _requires_login():
...
+         'private' if info['private'] else 'public')
+        for name, info in account_info['snaps'].get(series, {}).items()]
+
+    if not snaps:
+        print('There are no registered snaps for series {!r}.'.format(series))
+        return
+
+    tabulated_snaps = tabulate(
+        sorted(snaps, key=operator.itemgetter(0)),
+        headers=["Name", "Status", "Snap-Id", "Visibility"],
```
sergiusens: If `Snap-ID` goes through, then it should be `ID`, not `Id` ;-)

cprov:
@sergiusens dropped 'Snap-ID' & 'Status' and added 'Since', 'Price' and 'Notes'.

Dropping 'Status' is problematic because developer will not have a way to identify name disputes / review status, I am thinking about using 'Notes' to show this nuance.

sergiusens:
El 12/12/16 a las 13:44, Celso Providelo escribió:
>
> @sergiusens <https://github.com/sergiusens> dropped 'Snap-ID' & 
> 'Status' and added 'Since', 'Price' and 'Notes'.
>
> Dropping 'Status' is problematic because developer will not have a way 
> to identify name disputes / review status, I am thinking about using 
> 'Notes' to show this nuance.
>
Notes seems like a good location to add "tags" that people can query 
about further. If not we will have to come up with an infinite column 
listing for yes/no values :-)

sergiusens:
Looks like the fake servers need updating

snapcraft/main.py
```diff
@@ -94,6 +96,8 @@
   list-plugins List the available plugins that handle different types of part.
   login        Authenticate session against Ubuntu One SSO.
   logout       Clear session credentials.
+  list-registered List snap names registered or shared with you.
+  registered   Alias for list-snaps.
```
tsdgeos: Should this be "Alias for list-registered" ?

snapcraft/_store.py
```diff
@@ -163,6 +164,32 @@ def _requires_login():
...
+    ]
+
+    if not snaps:
+        print('There are no registered snaps for series {!r}.'.format(series))
+        return
+
+    tabulated_snaps = tabulate(
+        sorted(snaps, key=operator.itemgetter(0)),
+        headers=["Name", "Since", "Visibility", "Price", "Notes"],
+        tablefmt="plain")
```
elopio: I know you all think I have problems, I know. But single quotes please :)

cprov:
@elopio and @tsdgeos comments addressed, thanks for the feedback.

PR #891, opened 2016-11-08
repo: apt-mark new build-packages as automatically installed

3v1n0:
This allows to clean them up by just running `apt-get autoremove` once the snapcrafting is done, but in any case it's not breaking anything as it only applies to packages that were not previously installed.

A simpler solution for https://bugs.launchpad.net/snapcraft/+bug/1640143

### DIFF
--- a/integration_tests/__init__.py
+++ b/integration_tests/__init__.py
@@ -75,14 +75,15 @@ class TestCase(testtools.TestCase):
             self.addDetail('output', content.text_content(e.output))
             raise
 
-        deb_env = os.environ.copy()
-        deb_env.update({
-            'DEBIAN_FRONTEND': 'noninteractive',
-            'DEBCONF_NONINTERACTIVE_SEEN': 'true',
-        })
-        self.addCleanup(
-            subprocess.check_call, ['sudo', 'apt-get', 'autoremove', '-qq'],
-            stderr=subprocess.STDOUT, env=deb_env)
+        if not os.getenv('SNAPCRAFT_IGNORE_APT_AUTOREMOVE', False):
+            deb_env = os.environ.copy()
+            deb_env.update({
+                'DEBIAN_FRONTEND': 'noninteractive',
+                'DEBCONF_NONINTERACTIVE_SEEN': 'true',
+            })
+            self.addCleanup(
+                subprocess.check_call, 'sudo apt-get autoremove -qq'.split(),
+                stderr=subprocess.STDOUT, env=deb_env)
 
         return ret_val
 
--- a/integration_tests/__init__.py
+++ b/integration_tests/__init__.py
@@ -68,12 +68,23 @@ class TestCase(testtools.TestCase):
             cwd = os.path.join(self.path, yaml_dir)
 
         try:
-            return subprocess.check_output(
+            ret_val = subprocess.check_output(
                 [self.snapcraft_command, '-d'] + command, cwd=cwd,
                 stderr=subprocess.STDOUT, universal_newlines=True)
         except subprocess.CalledProcessError as e:
             self.addDetail('output', content.text_content(e.output))
             raise
+
+        deb_env = os.environ.copy()
+        deb_env.update({
+            'DEBIAN_FRONTEND': 'noninteractive',
+            'DEBCONF_NONINTERACTIVE_SEEN': 'true',
+        })
+        self.addCleanup(
+            subprocess.check_call, ['sudo', 'apt-get', 'autoremove', '-qq'],
+            stderr=subprocess.STDOUT, env=deb_env)
+
+        return ret_val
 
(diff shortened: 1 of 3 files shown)


sergiusens:
Nice PR, not sure you need `str` when using `format` though

sergiusens:
Can you update the integration tests for this in the base test class by adding a addCleanup of ap autoremove?

3v1n0:
> Nice PR, not sure you need str when using format though

In fact I was thinking the same (generally the error gets auto-converted), but I followed what it was done few lines above.

3v1n0:
Ok, tests added...

3v1n0:
Integration tests updated to...

debian/control
```diff
@@ -15,6 +15,7 @@ Build-Depends: bash-completion,
                python3-lxml,
                python3-magic,
                python3-petname,
+               python3-pexpect,
```
sergiusens: we don't run the integration tests on package build just in adt, reason for this not being needed

3v1n0:
@sergiusens ok, removed the debian bit and some double quotes (I've left them in the logger message for consistency to what it was done before)

3v1n0:
Ah, and do you also want not to use this behavior as default if an env variable (`SNAPCRAFT_IGNORE_APT_MARK_AUTO_INSTALLED` ?) is set?

sergiusens:
no, let's not add magic env vars for this please :-)

integration_tests/__init__.py
```diff
@@ -67,13 +67,42 @@ def run_snapcraft(self, command, project_dir=None, yaml_dir=None):
...
         try:
-            return subprocess.check_output(
+            snapcraft_output = subprocess.check_output(
                 [self.snapcraft_command, '-d'] + command, cwd=cwd,
                 stderr=subprocess.STDOUT, universal_newlines=True)
         except subprocess.CalledProcessError as e:
             self.addDetail('output', content.text_content(e.output))
             raise
 
+        if not os.getenv('SNAPCRAFT_IGNORE_APT_AUTOREMOVE', False):
```
sergiusens: can't we just always do the cleanup?

integration_tests/__init__.py
```diff
@@ -67,13 +67,42 @@ def run_snapcraft(self, command, project_dir=None, yaml_dir=None):
...
+                'apt-get autoremove output',
+                content.text_content(autoremove_output.decode('utf-8')))
+        except subprocess.CalledProcessError as e:
+            self.addDetail(
+                'apt-get autoremove error', content.text_content(str(e)))
+            self.addDetail(
+                'apt-get autoremove output',
+                content.text_content(e.output.decode('utf-8')))
+
+            if os.getenv('SNAPCRAFT_APT_AUTOREMOVE_CHECK_FAIL', False):
```
sergiusens: why is this needed?

snapcraft/internal/repo.py
```diff
@@ -98,6 +99,13 @@ def install_build_packages(packages):
                                'Dpkg::Progress-Fancy=1',
                                '--no-install-recommends', '-y',
                                'install'] + new_packages, env=env)
+        try:
+                subprocess.check_call(['sudo', 'apt-mark', 'auto'] +
+                                      new_packages, env=env)
```
elopio: You have four extra spaces here.

snapcraft/internal/repo.py
```diff
@@ -98,6 +99,13 @@ def install_build_packages(packages):
...
+        except subprocess.CalledProcessError as e:
+                logger.warning(
+                    "Impossible to mark packages as auto-installed: {}"
+                    .format(e))
```
elopio: And here, four extra spaces too.

snapcraft/internal/repo.py
```diff
@@ -98,6 +99,13 @@ def install_build_packages(packages):
                                'Dpkg::Progress-Fancy=1',
                                '--no-install-recommends', '-y',
                                'install'] + new_packages, env=env)
+        try:
+                subprocess.check_call(['sudo', 'apt-mark', 'auto'] +
+                                      new_packages, env=env)
+        except subprocess.CalledProcessError as e:
+                logger.warning(
+                    "Impossible to mark packages as auto-installed: {}"
```
elopio: Please use single quotes, for consistency :)

elopio:
I'm ok with the tests. Thanks @3v1n0 !

3v1n0:
@elopio  Fixed the spacing issue, let me know what I should do with the env variables and we should be all set.

sergiusens:
@3v1n0 sorry for causing conflicts, mind rebasing/merging with master?

3v1n0:
@sergiusens ok, done... And fixed also the wrong `is_dumb_terminal` call in repo.py (it was missing a `()`).

PR #892, opened 2016-11-08
Catch PermissionError when attempting to replace contents in a readonly file. (LP: #1640305)

larryprice:
I noticed this when using the python plugin to build parts for libertine. The python plugin attempts to open every file in the system with 'r+', and our readonly file would cause an exception to be thrown.

Also see https://bugs.launchpad.net/snapcraft/+bug/1640305


sergiusens:
staic and unit tests have failed, mind taking a look at them?

larryprice:
Thanks for the heads up... the test I added works locally, I'll investigate the issue with docker in the morning.

larryprice:
I removed the test I wrote for my new code. It always fails in docker because docker runs as root and will never get the PermissionDenied error when trying to write to a readonly file. I considered adding some mocking to force a test, but noticed that several of the except cases in file_utils.py were not covered for similar reasons. Let me know if I need to implement a unit test with mocking for this.

snapcraft/tests/test_file_utils.py
```diff
@@ -68,6 +68,25 @@ def test_replace_in_file(self):
...
+        file_info = {
+            'path': os.path.join('bin', 'readonly'),
+            'contents': '#!/foo/bar/baz/python',
+            'expected': '#!/foo/bar/baz/python',
+        }
+        with open(file_info['path'], 'w') as f:
+            f.write(file_info['contents'])
+
+        with mock.patch('snapcraft.file_utils.open') as mock_open:
+            mock_open.side_effect = PermissionError("")
```
kyrofa: I think this could be shortened a little:

```python
with mock.patch('snapcraft.file_utils.open', side_effect=PermissionError):
    # file_utils.replace...
```

snapcraft/file_utils.py
```diff
@@ -150,18 +150,22 @@ def _search_and_replace_contents(file_path, search_pattern, replacement):
...
+                return
+
+            replaced = search_pattern.sub(replacement, original)
+            if replaced != original:
+                f.seek(0)
+                f.truncate()
+                f.write(replaced)
+    except PermissionError:
+        # This file was readonly or otherwise uneditable.
+        return
```
kyrofa: I also wonder if this is what we want every time this function is called. Perhaps we want to leave such things up to the caller.

kyrofa: Nah, nevermind-- this is a protected member of this module.

larryprice:
I agree with both comments and have made the changes accordingly (as well as updated the branch).

snapcraft/file_utils.py
```diff
@@ -163,9 +163,9 @@ def _search_and_replace_contents(file_path, search_pattern, replacement):
                 f.seek(0)
                 f.truncate()
                 f.write(replaced)
-    except PermissionError:
-        # This file was readonly or otherwise uneditable.
-        return
+    except PermissionError as e:
+        logger.debug('Unable to open {path} for writing: {error}'.format(
```
kyrofa: This will be hidden unless snapcraft is run in debug mode-- I was thinking this would be a warning. Do you think this will happen often enough (and people will care little enough) to only be debug?

snapcraft/file_utils.py
```diff
@@ -150,18 +150,22 @@ def _search_and_replace_contents(file_path, search_pattern, replacement):
...
+                return
+
+            replaced = search_pattern.sub(replacement, original)
+            if replaced != original:
+                f.seek(0)
+                f.truncate()
+                f.write(replaced)
+    except PermissionError as e:
+        logger.debug('Unable to open {path} for writing: {error}'.format(
+            path=file_path, error=e))
```
elopio: Sorry for the late feedback here.
I don't like this huge try-catch.  I think that the statement that would throw PermissionError is f.truncate, isn't it?
If so, we can put a smaller try catch at the end of this code block.

Also, an alternative would be to check if the file is writable before trying to write it.

snapcraft/tests/test_file_utils.py
```diff
@@ -64,6 +64,25 @@ def test_replace_in_file(self):
...
+        file_info = {
+            'path': os.path.join('bin', 'readonly'),
+            'contents': '#!/foo/bar/baz/python',
+            'expected': '#!/foo/bar/baz/python',
+        }
+        with open(file_info['path'], 'w') as f:
+            f.write(file_info['contents'])
+
+        with mock.patch('snapcraft.file_utils.open',
+                        side_effect=PermissionError('')):
```
elopio: Instead of the mock, what about writing the file and then changing it's permission to readonly?

snapcraft/file_utils.py
kyrofa: I suspect it's actually the `open` using the `r+` flags, no?

larryprice: That's right - the throw PermissionError happens as soon as `open` is called with `r+`.

As for the large try-catch block, I believe that's the most pythonic way to handle this situation (asking forgiveness instead of permission). From what I've read, the try-catch will have little-to-no effect on performance unless the catch block is hit frequently, which shouldn't be the the case here (has only affected me with one or two files over the course of quite a few snap buidings).

snapcraft/tests/test_file_utils.py
larryprice: I tried that initially, but running the tests in docker causes that to fail because the docker user can write to any file it wants (regardless of permissions).

snapcraft/file_utils.py
larryprice: Agreed. I think that making it a warning is fair since this should happen very infrequently.

snapcraft/tests/test_file_utils.py
elopio: um, right, because they run as sudo. We should run as a normal user, but ugh docker.
A comment in the code explaining this would be appreciated.

snapcraft/file_utils.py
elopio: The problem with a big try-catch block is that you don't really control which statement is raising the exception. If it's raised in a statement that's not the one you think could raise it, it would be caught anyway and hide a bug.

In this case, I would break python's rule of not asking permission, or open the file twice - one for reading and one for writing in a try-catch. But those are not necessarily cleaner than yours, and as long as the try block doesn't grow with a more complex statement, the risk of a bug is not big. I'm changing my review to leave a +1, so you can keep your code if you like it better than my suggestions.

PR #893, opened 2016-11-09
Add a script to retry autopktests

elopio:
LP: #1640276

### DIFF
--- a/tools/retry_autopkgtest.sh
+++ b/tools/retry_autopkgtest.sh
@@ -46,9 +46,7 @@ echo "${SNAPCRAFT_AUTOPKGTEST_SECRET}" > "${temp_dir}/sec.txt"
 
 for testbed in "${testbeds[@]}"; do
     IFS=':' read -r release architecture <<< "$testbed"
-    if [ "${architecture}" == "" ]; then
-        architecture='amd64'
-    fi
+    [ -z "$architecture" ] && architecture='amd64'
     echo "Launching tests for the ${release} release in the ${architecture} architecture..."
     "${temp_dir}/retry-github-test" "https://api.github.com/repos/snapcore/snapcraft/pulls/${pr}" "https://autopkgtest.ubuntu.com/request.cgi?release=${release}&arch=${architecture}&package=snapcraft&ppa=snappy-dev%2Fsnapcraft-daily" "${temp_dir}/sec.txt"
 done
--- a/tools/retry_autopkgtest.sh
+++ b/tools/retry_autopkgtest.sh
@@ -9,7 +9,7 @@
 #   SNAPCRAFT_AUTOPKGTEST_SECRET: The secret to authenticate the test execution.
 
 if [ -z "${SNAPCRAFT_AUTOPKGTEST_SECRET}" ]; then
-    echo 'Set the secret to the environment variable $SNAPCRAFT_AUTOPKGTEST_SECRET.'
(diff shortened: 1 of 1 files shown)


tools/retry_autopkgtest.sh
```diff
@@ -0,0 +1,35 @@
...
+# Retry the autopkgtest run in a pull request.
+# Arguments:
+#   release:      The name of the Ubuntu release to test.
+#   architecture: The architecture to test.
+#   pr_id:        The identifier of the pull request to test.
+# Environment variables:
+#   SNAPCRAFT_AUTOPKGTEST_SECRET: The secret to authenticate the test execution.
+
+if [ -z "${SNAPCRAFT_AUTOPKGTEST_SECRET}" ]; then
+    echo 'Set the secret to the environment variable $SNAPCRAFT_AUTOPKGTEST_SECRET.'
```
sergiusens: this one should not have a `$`

tools/retry_autopkgtest.sh
```diff
@@ -0,0 +1,35 @@
...
+    exit 1
+fi
+
+if [ "$#" != 3 ]; then
+    echo "Usage: "$0" release architecture pr_id"
```
sergiusens: `$0 <release> <architecture> <PR_ID>`

Just `<PR>` would be better, I was thinking if I had to use something different than the number. `<PR#>` if you don't like `<PR>`

sergiusens: It would be nice to actually do it like this:

`$0 <PR> release:arch...`

So then we can call it once with

$0 889 zesty:amd64 yakkety:amd64 xenial:amd64

elopio: ahh, you are challenging my bash knowledge. Let me try.

elopio:
@sergiusens please review again, and let me know if it's what you expected.

I tested it launching a few tests in this same pr, it works.

tools/retry_autopkgtest.sh
```diff
@@ -0,0 +1,56 @@
...
+fi
+
+if [ "$#" -lt 1 ]; then
+    echo "Usage: "$0" <PR> [release[:architecture] ...]"
+    exit 1
+fi
+
+pr="$1"
+shift
+testbeds=( "$@" )
```
sergiusens: from examples above, this needs a default for xenial

tools/retry_autopkgtest.sh
```diff
@@ -0,0 +1,56 @@
...
+# Download the retry script.
+wget https://git.launchpad.net/~ubuntu-release/+git/autopkgtest-cloud/plain/tools/retry-github-test -O "${temp_dir}/retry-github-test"
+chmod +x "${temp_dir}/retry-github-test"
+
+# Save the secret to a file.
+echo "${SNAPCRAFT_AUTOPKGTEST_SECRET}" > "${temp_dir}/sec.txt"
+
+for testbed in "${testbeds[@]}"; do
+    IFS=':' read -r release architecture <<< "$testbed"
+    if [ "${architecture}" == "" ]; then
```
sergiusens: just a snirky one, this can also be written as
```sh
if [ -z "${architecture}" ]; then
```

Or the whole conditional replaced with
```sh
[ -z "$architecture" ] && architecture='amd64'
```

tools/retry_autopkgtest.sh
elopio: I forgot that one, sorry. Thanks for catching it.

PR #896, opened 2016-11-10
indicators: work with Content-Encoding set

sergiusens:
If Content-Encoding is set to a compressed stream we
currently crash badly, this handles it correctly.

LP: #1611776


elopio:
The integration test fails without your change, nice job there.
I think the complication to add a unit test is not worth it, so :+1:

PR #897, opened 2016-11-10
pluginhandler: use an alternate location to organize.

josepht:
Don't organize in the install dir.

Allow for detecting conflicts that would be caused by the 'organize' logic.

LP: 1628098


debian/snapcraft-completion
```diff
@@ -18,6 +18,16 @@ _snapcraft()
         _filedir -d
         return 0
         ;;
+    stage | pull | build | clean | prime)
```
kyrofa: 1. This isn't really related to this feature. Would you mind putting it in another PR?
2. I've actually wanted this for a while, nice job! However, it'll only work on dirty trees, right? What if we implemented a `snapcraft list-parts` command and used that here instead?

integration_tests/test_stage.py
```diff
@@ -46,3 +46,32 @@ def test_conflicts(self):
...
+    def test_no_conflicts(self):
+        project_dir = 'no-conflicts'
+        self.run_snapcraft('stage', project_dir)
+
+    def test_post_organize_conflicts(self):
+        project_dir = 'post-organize-conflicts'
+
+        exception = self.assertRaises(
+            subprocess.CalledProcessError,
+            self.run_snapcraft, 'stage', project_dir)
```
kyrofa: You're only using `stage` here, which means you can operate on individual parts. Since in the snaps you created the code being build is identical, I suggest you combine them into a single snap with multiple parts, two of which don't conflict and two of which do conflict. Otherwise we end up with two identical snaps except for a single organize key. Would that work?

debian/snapcraft-completion
elopio: +1. List parts will be useful on itself.

integration_tests/snaps/post-organize-conflicts/snapcraft.yaml
```diff
@@ -0,0 +1,19 @@
+name: test-package
+version: 0.1
+summary: one line summary
+description: a longer description
```
elopio: ditto.

snapcraft/file_utils.py
```diff
@@ -131,17 +131,22 @@ def create_similar_directory(source, destination, follow_symlinks=False):
...
 
-    shutil.copystat(source, destination, follow_symlinks=follow_symlinks)
+    if real_source:
+        try:
+            os.chown(destination, uid, gid, follow_symlinks=follow_symlinks)
+        except PermissionError as exception:
+            logger.debug('Unable to chown {}: {}'.format(
+                destination, exception))
+
+        shutil.copystat(source, destination, follow_symlinks=follow_symlinks)
```
elopio: A style question here, why are you using the real_source var? I might be missing something, but I think this would be the same:
```
os.makedirs(destination, exist_ok=True)
if os.path.exists(source):
    stat = os.stat(source, follow_symlinks=follow_symlinks)
    uid = stat.st_uid
    gid = stat.st_gid
    try:
        os.chown(destination, uid, gid, follow_symlinks=follow_symlinks)
        except PermissionError as exception:
        [...]
```

I also find it weird that if the source doesn't exist, this will just create the destination directory. Could you mention that in the docstring of the function?

snapcraft/internal/pluginhandler.py
```diff
@@ -804,27 +857,56 @@ def _migratable_filesets(fileset, srcdir):
...
+            dst = os.path.join(dstdir, directory)
 
         snapcraft.file_utils.create_similar_directory(src, dst)
+        if copy_tree:
+            file_utils.link_or_copy_tree(src, dst)
+
+
+def _can_skip(src, dst, missing_ok=False):
+    if os.path.isdir(src):
+        snapcraft.file_utils.create_similar_directory(
```
elopio: This makes me think that the name of the function is not right.
can_skip sounds like something that returns true or false, and that's it. This will create a directory as a side-effect, that I wouldn't expect.

debian/snapcraft-completion
kyrofa: @josepht reminder on this one.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -919,16 +985,37 @@ def _organize_filesets(fileset, base_dir):
...
+
 def _clean_migrated_files(snap_files, snap_dirs, directory):
     for snap_file in snap_files:
-        os.remove(os.path.join(directory, snap_file))
+        path = snap_file
+        if type(snap_file) == tuple:
+            path = snap_file[1]
+
+        # XXX: hack to remove directories should be fixed elsewhere
+        if os.path.isdir(os.path.join(directory, path)):
```
kyrofa: Remember that `os.path.isdir()` will return true for symlinks to directories. Will that cause problems?

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -453,13 +453,20 @@ def _get_fileset(self, option, default=None):
...
-        self._organize()
         snap_files, snap_dirs = self.migratable_fileset_for('stage')
+        organize_fileset = getattr(self.code.options, 'organize', {}) or {}
+
+        snap_files, new_snap_dirs = _organize_fileset(
+            snap_files, organize_fileset, self.code.installdir)
+        snap_dirs, extra_snap_dirs = _organize_fileset(
+            snap_dirs, organize_fileset, self.code.installdir)
+
+        snap_dirs |= new_snap_dirs | extra_snap_dirs
```
kyrofa: This chunk of code looks an _awful_ lot like `_get_staged_files()`, below. Should it be using that?

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -501,11 +508,41 @@ def clean_stage(self, project_staged_state, hint=''):
...
         snap_files, snap_dirs = self.migratable_fileset_for('prime')
-        _migrate_files(snap_files, snap_dirs, self.stagedir, self.snapdir)
+        organize_fileset = getattr(self.code.options, 'organize', {}) or {}
+
+        snap_files, new_snap_dirs = _organize_fileset(
+            snap_files, organize_fileset, self.code.installdir)
+        snap_dirs, extra_snap_dirs = _organize_fileset(
+            snap_dirs, organize_fileset, self.code.installdir)
+
+        snap_dirs |= new_snap_dirs | extra_snap_dirs
```
kyrofa: This also looks hauntingly like `_get_staged_files()`.

snapcraft/internal/pluginhandler/__init__.py
josepht: This instance is okay I think, because we're just adding the directory to the list that we clean later in code that I didn't modify.

kyrofa: That's exactly my concern though. Doesn't that mean that below you'll try to call `os.rmdir()` on a symlink? What removes the symlink itself?

snapcraft/tests/pluginhandler/test_pluginhandler.py
```diff
@@ -1821,6 +1821,54 @@ def test_clean_part_remaining_parts(self, mock_exists, mock_listdir,
...
+        # Create part1 and get it through the "build" step.
+        handler1 = mocks.loadplugin('part1')
+        handler1.makedirs()
+
+        bindir = os.path.join(handler1.code.installdir, 'bin')
+        os.makedirs(bindir)
+        path = os.path.join(bindir, '1')
+        os.mkdir(path)
+        symlink_path = os.path.join(bindir, '1-link')
+        os.symlink(path, symlink_path)
```
kyrofa: Note that this is a broken symlink. You probably want `os.symlink('1', symlink_path)`.

snapcraft/tests/pluginhandler/test_pluginhandler.py
```diff
@@ -1821,6 +1821,54 @@ def test_clean_part_remaining_parts(self, mock_exists, mock_listdir,
...
+
+        handler1.mark_done('build')
+        handler1.stage()
+
+        stage_symlink_path = os.path.join(handler1.stagedir, 'bin', '1-link')
+        stage_path = os.path.join(handler1.stagedir, 'bin', '1')
+        self.assertTrue(os.path.exists(stage_symlink_path))
+        self.assertTrue(os.path.exists(stage_path))
```
kyrofa: The assertion errors are way better if you kneel to testtools: `self.assertThat(stage_path, FileExists())`. This way you can even use `self.assertThat(stage_symlink_path, tests.LinkExists('1'))` that will verify that the link exists and is pointing to '1'.

snapcraft/tests/pluginhandler/test_pluginhandler.py
```diff
@@ -1821,6 +1821,54 @@ def test_clean_part_remaining_parts(self, mock_exists, mock_listdir,
...
+        # Create part1 and get it through the "build" step.
+        handler1 = mocks.loadplugin('part1')
+        handler1.makedirs()
+
+        bindir = os.path.join(handler1.code.installdir, 'bin')
+        os.makedirs(bindir)
+        path = os.path.join(bindir, '1')
+        open(path, 'w').close()
+        symlink_path = os.path.join(bindir, '1-link')
+        os.symlink(path, symlink_path)
```
kyrofa: Broken symlink here as well.

snapcraft/tests/pluginhandler/test_pluginhandler.py
```diff
@@ -1821,6 +1821,54 @@ def test_clean_part_remaining_parts(self, mock_exists, mock_listdir,
...
+
+        handler1.mark_done('build')
+        handler1.stage()
+
+        stage_symlink_path = os.path.join(handler1.stagedir, 'bin', '1-link')
+        stage_path = os.path.join(handler1.stagedir, 'bin', '1')
+        self.assertTrue(os.path.exists(stage_symlink_path))
+        self.assertTrue(os.path.exists(stage_path))
```
kyrofa: Ditto the recommendation to use testtools' matchers.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -508,8 +500,8 @@ def clean_stage(self, project_staged_state, hint=''):
 
         self.mark_cleaned('stage')
 
-    def _get_staged_files(self):
-        snap_files, snap_dirs = self.migratable_fileset_for('stage')
+    def _get_step_files(self, step):
```
kyrofa: Ah, this makes me much happier, thank you :grin: .

snapcraft/file_utils.py
```diff
@@ -139,16 +139,20 @@ def create_similar_directory(source, destination, follow_symlinks=False):
...
-    try:
-        os.chown(destination, uid, gid, follow_symlinks=follow_symlinks)
-    except PermissionError as exception:
-        logger.debug('Unable to chown {}: {}'.format(destination, exception))
+    if os.path.exists(source):
+        stat = os.stat(source, follow_symlinks=follow_symlinks)
+        uid = stat.st_uid
+        gid = stat.st_gid
+
+    if os.path.exists(source):
```
kyrofa: I can't quite determine why this is being checked twice instead of once. Am I missing something?

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -861,31 +882,60 @@ def _migratable_filesets(fileset, srcdir):
     return snap_files, snap_dirs
 
 
-def _migrate_files(snap_files, snap_dirs, srcdir, dstdir, missing_ok=False,
-                   follow_symlinks=False, fixup_func=lambda *args: None):
-
+def _migrate_dirs(snap_dirs, srcdir, dstdir, step='stage'):
```
kyrofa: How is `step` used, here?

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -861,31 +882,60 @@ def _migratable_filesets(fileset, srcdir):
...
+    # If the file is already here and it's a symlink, leave it alone.
+    if os.path.islink(dst):
+        return True
+
+    return False
+
+
+def _migrate_files(snap_files, snap_dirs, srcdir, dstdir, step='stage',
+                   missing_ok=False, follow_symlinks=False,
+                   fixup_func=lambda *args: None):
```
kyrofa: How is `step` used, here?

snapcraft/internal/pluginhandler/__init__.py
josepht: It's not.

snapcraft/file_utils.py
josepht: It shouldn't be.  I'll fix it.

snapcraft/internal/pluginhandler/__init__.py
josepht: It's not.

kyrofa:
(note the conflicts with master, here).

snapcraft/file_utils.py
```diff
@@ -139,16 +139,19 @@ def create_similar_directory(source, destination, follow_symlinks=False):
...
 
-    stat = os.stat(source, follow_symlinks=follow_symlinks)
-    uid = stat.st_uid
-    gid = stat.st_gid
-    os.makedirs(destination, exist_ok=True)
-    try:
-        os.chown(destination, uid, gid, follow_symlinks=follow_symlinks)
-    except PermissionError as exception:
-        logger.debug('Unable to chown {}: {}'.format(destination, exception))
+    if os.path.exists(source):
```
kyrofa: You didn't answer my question here: why would `create_similar_directory()` be called if the directory we want to imitate (`source`) doesn't exist?

josepht: Sorry I missed it.  I think you're right that the check isn't needed anyway.

kyrofa:
This means `organize` no longer needs to be a [build property](https://github.com/snapcore/snapcraft/blob/master/snapcraft/internal/states/_build_state.py#L35), right?

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -1059,38 +1131,157 @@ def _file_collides(file_this, file_other):
...
+    if matched:
+        fileset, dirs = _add_filepath_to_fileset(
+            filepath, srcdir, fileset, dirs)
+    else:
+        fileset.add(filepath)
+
+    return fileset, dirs
+
+
+def _organize_fileset(fileset_orig, organize_fileset, srcdir):
```
kyrofa: I'm having some serious problems understanding what's happening through here. Combined with the fact that the pluginhandler is already quite bloated, I'm starting to think this behavior would be better suited to being extracted into another module, perhaps a class. And documented very well.

sergiusens:
closing this for now. It needs a potential new approach

PR #900, opened 2016-11-12
ftp source support

3v1n0:
As requested on bug https://bugs.launchpad.net/snappy/+bug/1602323

This adds support for FTP sources, using urllib was the quickest choice, and with some refactor of the indicator module, adding a progressbar was possible too. If using ftplib or others is preferred let me know, but I went for the fastest and safer way.

In theory this path could be used also for http(s) downloads, but not to change what we've done so far, I've left that job to the nicer `requests` library. `requests` being an HTTP only library has no native suppot for this, while trying to use something like [requests-ftp](https://pypi.python.org/pypi/requests-ftp) I don't think is a choice (looking quite immature).

### DIFF
--- a/snapcraft/internal/indicators.py
+++ b/snapcraft/internal/indicators.py
@@ -17,7 +17,7 @@
 import os
 import sys
 
-from urllib.request import FancyURLopener
+from urllib.request import urlretrieve
 from progressbar import (
     AnimatedMarker,
     Bar,
@@ -85,7 +85,7 @@ def download_urllib_source(uri, destination, message=None):
             min(total_read, total_length) if total_length > 0 else total_read)
 
     progress_bar = None
-    FancyURLopener().retrieve(
+    urlretrieve(
         uri, destination, lambda n, s, l: progress_cb(progress_bar, n, s, l))
 
     if progress_bar:
--- a/snapcraft/internal/indicators.py
+++ b/snapcraft/internal/indicators.py
@@ -17,6 +17,7 @@
 import os
 import sys
 
+from urllib.request import FancyURLopener
 from progressbar import (
     AnimatedMarker,
     Bar,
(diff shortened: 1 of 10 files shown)


3v1n0:
This will probably need to be fixed once https://github.com/snapcore/snapcraft/pull/906 gets merged...

3v1n0:
Wondering why the integration test fails in autopkgtest with the `urllib.error.URLError: <urlopen error ftp error: TimeoutError(110, 'Connection timed out')>` error, while it works fine here and in travis...

3v1n0:
Ok the connection timeout issue was due to missing `ftp_proxy` in autopkgtest...

For it to pass, right now it was enough to reuse the `http_proxy`, unfortunately [changing it in debian/tests](https://github.com/snapcore/snapcraft/pull/900/commits/9cf814a1b36cebf62f9bf694a54cfbc68563f8b4) wasn't enough, so I went for this workaround: https://github.com/snapcore/snapcraft/pull/900/commits/eb078e848.

That can be removed once https://github.com/ubuntu-core/snappy-jenkins/pull/210 will be applied, I guess.

snapcraft/internal/indicators.py
```diff
@@ -64,6 +71,36 @@ def download_requests_stream(request_stream, destination, message=None):
     progress_bar.finish()
 
 
+class UrllibDownlaoder(object):
```
elopio: typo: Downlaoder/Downloader

snapcraft/internal/indicators.py
```diff
@@ -64,6 +71,36 @@ def download_requests_stream(request_stream, destination, message=None):
...
+        self.message = message
+        self.progress_bar = None
+
+    def download(self):
+        urlretrieve(self.uri, self.destination, self._progress_cb)
+
+        if self.progress_bar:
+            self.progress_bar.finish()
+
+    def _progress_cb(self, block_num, block_size, total_length):
```
elopio: I know I'm being a pita, but can you rename this to _progress_callback?
I have a hardtime with acronyms because my brain is not yet fast enough in english to understand them. I always have to spent a few minutes parsing it. That's the only thing I hate about go, and that's why I love that python code tends to be verbose :)

snapcraft/tests/__init__.py
```diff
@@ -109,6 +111,22 @@ def setUp(self):
         self.useFixture(fixture_setup.FakeParts())
 
 
+class FakeFileHttpServerBasedTestCase(TestCase):
```
elopio: The python style is to name it FakeFileHTTPServerBasedTestCase

elopio:
@3v1n0 that sounds like a thing to add to our autopkgtest setup. Can you talk to pitti about it?
I'm ok with the workaround, and thanks for adding this feature.

3v1n0:
Ok @elopio, changed the naming as suggested.

As for the `ftp_proxy` thing, isn't https://github.com/ubuntu-core/snappy-jenkins/pull/210 enough?!?

martinpitt:
Canonical datacenter's proxy does not allow ftp, sorry. You have to use http, or skip the test.

elopio:
@3v1n0 nop, sorry, I forgot to reply to that. Last week we stopped using that jenkins server, and instead we are now sending the PRs directly to the autopkgtest infrastructure: https://wiki.ubuntu.com/ProposedMigration/AutopkgtestInfrastructure#Integration_with_GitHub_and_GitLab_pull.2Fmerge_requests

3v1n0:
@martinpitt but... If I set the `ftp_proxy=$http_proxy` the test does pass... So why not just doing that by default?

3v1n0:
@sergiusens any moment for looking at this?

3v1n0:
Rebased on latest master

PR #901, opened 2016-11-13
Basic 'enable-ci' implementation.

cprov:
Exploratory work for automating travis CI setup.

It's driving "travis-cli" (sudo gem install travis), which could either be replaced by playing travis/GH APIs directly or, perhaps, snapping the ruby app (too crazy ?).

It also proposes per-project snapcraft configuration, where macaroons could be properly attenuated and relocated for each project. However, I am not entirely sold on this and the risk of committing credentials, even if they are attenuated.

Regarding macaroon attenuation, the Store support attenuation to one (or multiple) channels was recently added.


snapcraft/_store.py
```diff
@@ -100,6 +101,119 @@ def _requires_login():
         raise
 
 
+@contextmanager
+def _requires_command_success(command, not_found_err=None, failure_err=None):
```
elopio: This is similar to what Joe did for cloning repos. Search for MissingPackageError.
It's not the same, because this one actually tries the command, but maybe something can be shared?

elopio:
Thanks Celso. I just had a quick look before lunch.
It worries me a little to fill snapcraft with commands. Would it be a terrible idea to make this a script that we tell people to download and run? wdyt?

snapcraft/_store.py
cprov: Thanks for the pointer @elopio.
They could possibly be grouped into integration_utils.py (suggestion) but I am not sure about the gains at this point.
A small observation about the 'package' term being overloaded there. Joe's function should probably be called _check_command_exists() (according to `which` usage), it's a command/bin that is checked, not necessarily a package.

cprov:
@elopio, IMHO these set of (integration) commands belong to snapcraft, it's the de-facto tool for building snaps. Additional binaries and installations would confuse developers.

We cogitated having plugins for extending snapcraft itself in the past, which could somehow alleviate this concern, but it would also require embedded management (like parts) to make sense, and there we go into the rabbit hole ... We are not talking about tons of new code to justify separate shipping/install, the integration sub-commands will always be tiny and very specific, and more importantly, under upstream control (review/release process).

snapcraft/_store.py
sergiusens: It should also be using `shutil.which` ;-)

Agreed on renaming the exception as well.

sergiusens:
About filling the cli with these commands I am more for an `enable-ci` command with `--travis`, `--semaphore-ci`, `--circle-ci`, maybe defaulting to the most popular atm.

snapcraft/_store.py
```diff
@@ -100,6 +101,119 @@ def _requires_login():
...
+    EnvironmentError(
+        'Travis CLI (`travis version`) is not functional.\n'
+        'Make sure it works correctly in your system '
+        'before trying this again.'))
+@_requires_command_success(
+    'git status',
+    EnvironmentError(
+        'Git (`git`) is not available, this tool cannot verify '
+        'its prerequisites.\n'
+        'Please install it (e.g. `sudo apt-get install git`) '
```
elopio: nit: apt instead of apt-get

snapcraft/_store.py
```diff
@@ -100,6 +101,119 @@ def _requires_login():
...
+
+def _travis_encrypt_config(config_path):
+    """Encrypt given snapcraft config file for Travis jobs.
+
+    Depends on working `travis` CLI (boo!)
+    """
+    cmd = [
+        'travis', 'encrypt-file',
+        '--add', '--force', '--decrypt-to', '.snapcraft.cfg',
+        config_path, '.snapcraft.cfg.enc',
```
elopio: Evan's gist puts the encripted file in a travis variable, and it puts only a small secret in the .yaml.
I don't think there's a big security difference on either approach, but it could be easier if we don't have to add another file to the repository. Do you have a reason to prefer this approach?

snapcraft/_store.py
```diff
@@ -100,6 +101,119 @@ def _requires_login():
...
+    )
+
+    series = storeapi.constants.DEFAULT_SERIES
+    # XXX cprov 20160923: Using unresolved (name, series) caveat,
+    # instead of 'snap-id'. Mixed feelings on why it is supported.
+    # Needs caveat syntax for restricting store request origin (IP
+    # or reverse-dns).
+    packages = [{'name': snap_name, 'series': series}]
+
+    logger.info('Acquiring specific authorisation information ...')
```
elopio: nit: you use "authorization" below. I prefer it with z, but I'm not sure which one is more correct.
Also above you are using initialised a couple of times. I would spell that with a z too.

snapcraft/_store.py
```diff
@@ -100,6 +101,119 @@ def _requires_login():
...
+    store = storeapi.StoreClient()
+    if not _login(store, packages=packages, save=False):
+        raise RuntimeError('Cannot continue without logging in successfully.')
+
+    logger.info(
+        'Encrypting authorization for Travis and adjusting project to '
+        'automatically decrypt and setup it in testbeds.')
+
+    # XXX cprov: dumping a local config too ...
```
elopio: why don't you remove it after it's encrypted?

snapcraft/config.py
```diff
@@ -66,6 +70,18 @@ def is_empty(self):
...
+        # Local configurations (per project) are supposed to be static.
+        # That's why it's only checked for 'loading' and never written to.
+        # Essentially, all authentication-related changes, like login/logout
+        # or macaroon-refresh, will not be persisted for the next runs.
+        local_config = '.snapcraft.cfg'
+        if os.path.exists(local_config):
+            self.parser.read(local_config)
+            logger.warn(
+                'Using local configuration (`.snapcraft.cfg`), changes will '
+                'not be persisted.')
```
elopio: This needs tests :)

snapcraft/_store.py
cprov: fixed

snapcraft/_store.py
cprov: Our current macaroons don't fit into a travis encrypted variable size limit (128 bytes). Anyway, it turns out that encrypting file (.snapcraft.cfg) is actually cleaner.

snapcraft/_store.py
cprov: fixed, american spelling.

snapcraft/config.py
cprov: Tests added, along with a major refactor for isolating "integrations" implementations.

debian/snapcraft-completion
```diff
@@ -6,7 +6,7 @@ _snapcraft()
     COMPREPLY=()
     cur="${COMP_WORDS[COMP_CWORD]}"
     prev="${COMP_WORDS[COMP_CWORD-1]}"
-    opts="help init list-plugins login logout list-keys keys create-key register-key register tour push release clean cleanbuild pull build sign-build stage prime snap update define search gated validate history status close"
+    opts="help init list-plugins login logout list-keys keys create-key register-key register tour push release clean cleanbuild pull build sign-build stage prime snap update define search gated validate history status close enable-ci"
```
sergiusens: can you add an autocomplete for `travis` below?

snapcraft/__init__.py
```diff
@@ -251,6 +251,7 @@
 from snapcraft import plugins                       # noqa
 from snapcraft import sources                       # noqa
 from snapcraft import file_utils                    # noqa
+from snapcraft.integrations import enable_ci        # noqa
```
sergiusens: I am starting to wonder about all these... is this one needed here?

snapcraft/integrations/travis.py
```diff
@@ -0,0 +1,153 @@
...
+It will acquire properly attenuated Store credentials and encrypt it for
+use in your testbed ('.travis_snapcraft.cfg'), only Travis has the private
+key to decrypt it.
+
+Then it will adjust Travis configuration ('.travis.yml') with the commands
+to decrypt credentials and install and run `snapcraft` to release your snap
+(inside a ubuntu:xenial docker container) during the 'after_success' phase.
+See the example below:
+
+{{{
```
sergiusens: is this ubuntu wiki syntax? Maybe ```

snapcraft/integrations/travis.py
```diff
@@ -0,0 +1,153 @@
...
+        'Encrypting authorization for Travis and adjusting project to '
+        'automatically decrypt and use it during "after_success".')
+
+    with tempfile.NamedTemporaryFile(mode='w') as fd:
+        store.conf.parser.write(fd)
+        fd.flush()
+        _encrypt_config(fd.name)
+
+    with open('.travis.yml', 'r+') as fd:
+        travis_conf = yaml.load(fd)
```
sergiusens: this is taking advantage of our `OrderedDict` right? well I am more interested in the `dump`

snapcraft/integrations/travis.py
```diff
@@ -0,0 +1,153 @@
...
+            'docker run -v $(pwd):$(pwd) -t ubuntu:xenial '
+            'sh -c "apt-update -qq && apt install snapcraft -y && '
+            'cd $(pwd) && snapcraft && snapcraft push *.snap --release edge"')
+        fd.seek(0)
+        yaml.dump(travis_conf, fd, default_flow_style=False)
+
+    logger.info(
+        'Done. Now you just have to review and commit changes in your '
+        'Travis project (`.travis.yml`).\n'
+        'Also make sure you add the new ".travis_snapcraft.cfg" file.')
```
sergiusens: shouldn't this be a constant somewhere `.travis_snapcraft.cfg` ?

snapcraft/integrations/travis.py
cprov: converted to Markdown

snapcraft/integrations/travis.py
cprov: Yes, yaml parsers are configured globally for snapcraft, it's checked in a new test.

snapcraft/integrations/travis.py
cprov: good point, constants added.

snapcraft/integrations/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+            'Please select one of the supported integration systems: '
+            '{}.'.format(ci_system, ','.join(SUPPORTED_CI_SYSTEMS))
+        )
+
+    module = importlib.import_module(
+        'snapcraft.integrations.{}'.format(ci_system))
+
+    print(module.__doc__)
+
+    if input('Continue (y/N):') == 'y':
```
sergiusens: It feels we are mixing UI a lot; this can land as is, but what would the best place to put this in be?

I guess we will have a module in snapcraft that drives this instead of having the package do all the presentation work.

snapcraft/integrations/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+            cmd_list, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    except FileNotFoundError as err:
+        raise not_found_err
+    except subprocess.CalledProcessError as err:
+        raise failure_err
+    yield
+
+
+@contextmanager
+def requires_path_exists(path, error_message=None):
```
sergiusens: Can this me moved to file_utils?

snapcraft/integrations/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+    if error_message is None:
+        error_message = 'Required path does not exist: {}'.format(path)
+    if not os.path.exists(path):
+        raise EnvironmentError(error_message)
```
sergiusens: Can we have a snapcraft.errors exception used please?

My ultimate goal is to allow all primitive exceptions go through and raise a stack trace on the console to have an easier separation of exceptions used for user feedback and those which we just don't handle

snapcraft/integrations/travis.py
```diff
@@ -0,0 +1,164 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Snapcraft integration for Travis (CI).
+
+This command currently depends on working `travis` CLI environment and
```
sergiusens: I added some comments about this text on the UX document if you care to reflect here.

snapcraft/integrations/__init__.py
cprov: Yes, it makes sense to introduce a standard way of exposing commands preambles and continue/abort user interaction (associated with '-y' cmdline switch)

snapcraft/integrations/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Snapcraft integrations layer.
+
+Defines 'enable-ci' command infrastructure to support multiple integrations
+systems in an isolated form.
```
kyrofa: If the purpose of this module is to define the very specific `enable-ci` infrastructure, why name it something generic like "integrations"? Do we intend to grow this in the future?

snapcraft/integrations/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+        '`{}` not found.'.format(cmd_list[0]))
+    failure_err = failure_err or EnvironmentError(
+        '`{}` failed.'.format(command))
+    try:
+        subprocess.check_call(
+            cmd_list, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    except FileNotFoundError as err:
+        raise not_found_err
+    except subprocess.CalledProcessError as err:
+        raise failure_err
```
kyrofa: Is there a downside to swallowing the actual error message here?

snapcraft/integrations/tests/test_travis.py
```diff
@@ -0,0 +1,185 @@
...
+    def test_enable_missing_travis_cli(self, mock_check_call):
+        mock_check_call.side_effect = [FileNotFoundError()]
+
+        with self.assertRaises(EnvironmentError) as raised:
+            travis.enable()
+
+        self.assertEqual([
+            'Travis CLI (`travis`) is not available.',
+            'Please install it (e.g. `sudo gem install travis`) before '
+            'trying this command again.',
```
kyrofa: At least on a vanilla lxc image, ruby isn't even installed, so this will just result in `sudo: gem: command not found`. Are we assuming the users will understand what a gem is and what they need to do in order to install them, or should we mention that ruby needs to be installed first?

snapcraft/integrations/travis.py
```diff
@@ -0,0 +1,164 @@
...
+a previously initialised Travis project (`.travis.yml`).
+
+Make sure your Travis project is also configured to "Build pushes", this
+way every new push to `master` will result in a new snap revision on the
+Store.
```
kyrofa: Should this be "[...] new snap revision _in_ the Store" ?

kyrofa: Also, what happens if there are multiple branches? Will it build a new snap regardless of branch?

snapcraft/integrations/travis.py
```diff
@@ -0,0 +1,164 @@
...
+            'sh -c "apt-update -qq && apt install snapcraft -y && '
+            'cd $(pwd) && snapcraft && snapcraft push *.snap --release edge"; '
+            'fi')
+        fd.seek(0)
+        yaml.dump(travis_conf, fd, default_flow_style=False)
+
+    logger.info(
+        'Done. Now you just have to review and commit changes in your '
+        'Travis project (`{}`).\n'
+        'Also make sure you add the new `{}` file.'.format(
```
kyrofa: Nah, nevermind, reading it all together, I like what's here.

snapcraft/integrations/travis.py
cprov: yes, the preposition-donkey-hat is *mine* :-)

snapcraft/tests/test_commands_enable_ci.py
```diff
@@ -0,0 +1,104 @@
...
+            '    after_success:',
+            '    - openssl aes-256-cbc -K <travis-key> -iv <travis-iv>',
+            '      -in .travis_snapcraft.cfg -out .snapcraft.cfg -d',
+            '    - if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then',
+            '      docker run -v $(pwd):$(pwd) -t ubuntu:xenial',
+            '      sh -c "apt update -qq && apt install snapcraft -y && '
+            'cd $(pwd) &&',
+            '      snapcraft && snapcraft push *.snap --release edge";',
+            '      fi',
+            ''
```
kyrofa: It's nice being able to test that this works as expected, but I'm afraid that duplicating the large text here in its entirety will end up being difficult to maintain (i.e. fix a typo, have to change the test). Would the test be as robust if we used some regex to just make sure the right thing is printed, for example?

snapcraft/integrations/travis.py
cprov: the travis commands added are configured to run only on changes to master (` if [ "$TRAVIS_PULL_REQUEST" = "false" ];`)

kyrofa: But Travis will run on push to _any_ branch in the repo, not just master, unless limited by the user.

snapcraft/integrations/__init__.py
cprov: Sure, the only caveat is that it's currently `snapcraft.internal.errors`, the _internal_ bit makes it confusing during the transition.

snapcraft/integrations/travis.py
cprov: Adjusted.

snapcraft/integrations/travis.py
kyrofa: To be clear, I'm not sure you can completely solve this limitation without forcing the developer into a workflow (which I don't think we want to do). I wonder if [using the `deploy` step and making use of the `on` keyword](https://docs.travis-ci.com/user/deployment/#Conditional-Releases-with-on%3A) would make it easier for the dev to customize, though?

snapcraft/integrations/__init__.py
cprov: The error handling was redesigned according to @sergiusens request above. I think the idea is to give complete control about the error message to the callsites, so far, how it failed is not super important, we are only interested on the fact it failed. We might have other requirements in the future.

kyrofa: Works for me.

snapcraft/integrations/travis.py
```diff
@@ -0,0 +1,164 @@
...
+
+This command currently depends on working `travis` CLI environment and
+a previously initialised Travis project (`.travis.yml`).
+
+Make sure your Travis project is also configured to "Build pushes", this
+way every new push to `master` will result in a new snap revision on the
+Store.
+
+This operation will acquire properly attenuated Store credentials and
+encrypt it for use in your testbed (`.travis_snapcraft.cfg`), only Travis
```
elopio: s/encrypt it/encrypt them

snapcraft/integrations/travis.py
```diff
@@ -0,0 +1,164 @@
...
+has the private key to decrypt it and will be only available to branches
+of the same repository, not forks.
+
+Then it will adjust Travis configuration ('.travis.yml') with the commands
+to decrypt credentials and install and run `snapcraft` to release your snap
```
elopio: decrypt credentials, (comma instead of and), install and run.

snapcraft/integrations/travis.py
```diff
@@ -0,0 +1,164 @@
...
+(inside a ubuntu:xenial docker container) during the 'after_success' phase.
```
elopio: s/a ubuntu/an ubuntu. I think.

snapcraft/integrations/__init__.py
elopio: I think the input questions are usually followed by a space, like this:

`'Continue (y/N): '`

snapcraft/integrations/tests/test_travis.py
cprov: yes, `apt install ruby-dev ruby-ffi libffi-dev` added to the instructions.

snapcraft/integrations/__init__.py
cprov: yup, fixed.

snapcraft/integrations/travis.py
cprov: fixed

snapcraft/integrations/travis.py
cprov: fixed

snapcraft/integrations/travis.py
cprov: It's `if [ "$TRAVIS_BRANCH" = "master" ];`, per rocketchat discussion.

cprov: let me have another look on the deploy script provider.

kyrofa: @cprov I made that comment before we talked on Rocket. No need to restructure this unless you find you like that better.

elopio:
I have done a full exploratory on all the things I could imagine here, and I'm pretty happy with it.
A few important things:
- It will not work when .travis.yml and snapcraft.yaml are not in the same directory.
- A similar but maybe less important case: it will not work when the repository builds multiple snaps. Like https://github.com/elopio/hashicorp-snaps, which is intended to include all the projects from hashicorp. I don't think too much about this, because the travis config is too custom to automate. It's probably ok to not support this case.
- The frustration when you start from scratch and you don't have any of the requirements. Every time you run enable-ci and answer 'y', it tells you about one missing thing. So you have to do the same three or four times until you get everything right.

In general, nice work. I'm going to start using right now.

sergiusens:
El 22 nov. 2016 8:54 PM, "Leo Arias" <notifications@github.com> escribió:
>
> I have done a full exploratory on all the things I could imagine here,
and I'm pretty happy with it.
> A few important things:
>
> It will not work when .travis.yml and snapcraft.yaml are not in the same
directory.

Does it work with .snapcraft.yaml? Will there need to be adjustments for
when we move to .snapcraft/snapcraft.yaml?

cprov:
@sergiusens the current implementation uses `snapcraft.internal.load_config`, which presumably will be adjusted to support any of the available alternatives.

cprov:
@kyrofa it's using Travis 'deploy' phase now for building, pushing and releasing snaps. As you suggested it not only better in terms of semantics, but also clearer in terms of logic (replaced the bash if by built-in attributes). Thanks for pointing that out.

cprov:
@elopio for the cases you've pointed, I can only think of `snapcraft enable-ci raw` which would spit .snapcraft.cfg besides snapcraft.yaml and you would be on your own to securely store the attenuated credentials, make them available to testbeds and run the appropriated commands.
I don't think we gain much diving into the rabbit hole of figuring out all the various way a snap can be built nor forcing users to have a specific building setup. 
What do you say ?

snapcraft/integrations/__init__.py
cprov: Yes, semaphore-ci and potentially other suitable systems, on top of that the idea of having a `raw` integration, basically to acquire attenuated creds and let developer do what they want with them.

sergiusens:
I say we don't focus on `raw` right now. We already only support snapcraft.yaml's in the root for a reason.

snapcraft/integrations/travis.py
```diff
@@ -0,0 +1,183 @@
...
+container) during the 'deploy' phase.
+
+See the example below::
+
+    sudo: required
+    services:
+    - docker
+    after_success:
+    - openssl aes-256-cbc -K <travis-key> -iv <travis-iv>
+      -in .travis_snapcraft.cfg -out .snapcraft.cfg -d
```
elopio: This is wrong @cprov. You should put the .cfg in the config dir:
$ mkdir -p "${HOME}/.config/snapcraft"
... -out "${HOME}/.config/snapcraft/snapcraft.cfg"

snapcraft/integrations/travis.py
```diff
@@ -0,0 +1,183 @@
...
+    deploy:
+      skip_cleanup: true
+      provider: script
+      script: docker run -v $(pwd):$(pwd) -t ubuntu:xenial sh -c
```
elopio: You also need to mount the home here:
-v "${HOME}":/root

snapcraft/integrations/travis.py
cprov: @elopio the changes present only is this PR allows snapcraft to pick .snapcraft.cfg when it's present.

snapcraft/integrations/travis.py
cprov: @elopio that's not needed with .snapcraft.cfg

snapcraft/integrations/travis.py
elopio: This is wrong @cprov. You should put the .cfg in the config dir:
$ mkdir -p "${HOME}/.config/snapcraft"
... -out "${HOME}/.config/snapcraft/snapcraft.cfg"

elopio: One more detail, when this is done from a fork, the encripted env vars will not be set. So it prints:
iv undefined.
It would be nice to do the decrypt only in master, too.

cprov: The decrypting command is added by travis CLI and copes nicely with `iv` being undefined (I mean, understands there is nothing to do in that case). While we could adjust it manually I would rather not dispute travis control on that.

snapcraft/integrations/travis.py
cprov: @elopio here is an example for the transition period, current xenial snapcraft & running this PR locally. https://github.com/cprov/surl/blob/master/.travis.yml#L22

elopio: Ahh, I understand. This was failing for us because docker was using the snapcraft from the archive. Ok then, I'm changing my review back to green.

snapcraft/config.py
```diff
@@ -66,6 +70,18 @@ def is_empty(self):
         return True
 
     def load(self):
+        # Local configurations (per project) are supposed to be static.
+        # That's why it's only checked for 'loading' and never written to.
+        # Essentially, all authentication-related changes, like login/logout
+        # or macaroon-refresh, will not be persisted for the next runs.
+        local_config = '.snapcraft.cfg'
```
sergiusens: is it too much to ask to create a `.snapcraft` directory? How much of a burden is an extra file? We are planning on using this layout btw:
```
.plugins/
    snapcraft.yaml
    plugins/
        x_plugin1.py
        x_plugin2.py
        ...
```

cprov:
@sergiusens `.snapcraft/snapcraft.cfg` it is, thanks for the pointers.

PR #903, opened 2016-11-14
store: download without login

sergiusens:
LP: #1606903

### DIFF
--- a/snapcraft/storeapi/__init__.py
+++ b/snapcraft/storeapi/__init__.py
@@ -250,7 +250,7 @@ class StoreClient():
         package = self.cpi.get_package(snap_name, channel, arch)
         self._download_snap(
             snap_name, channel, arch, download_path,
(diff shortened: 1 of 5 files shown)


snapcraft/storeapi/__init__.py
```diff
@@ -372,7 +374,8 @@ def get_package(self, snap_name, channel, arch=None):
     def get(self, url, headers=None, params=None, stream=False):
         if headers is None:
             headers = {}
-        headers.update({'Authorization': _macaroon_auth(self.conf)})
+        with contextlib.suppress(errors.InvalidCredentialsError):
```
elopio: can you explain this supress?

sergiusens: We cannot assume all requests require a login, can we? We already have the with login guard on every call that requires one. If you are not logged in there is no need to _macaroon_auth

PR #904, opened 2016-11-15
cleanbuild: use pylxd instead of lxd command line

kalikiana:
https://bugs.launchpad.net/snapcraft/+bug/1641520


debian/control
```diff
@@ -17,6 +17,7 @@ Build-Depends: bash-completion,
                python3-petname,
                python3-pkg-resources,
                python3-progressbar,
+               python3-pylxd,
```
kyrofa: Does this also need to be in the binary package below?

snapcraft/internal/lxd.py
```diff
@@ -17,9 +17,14 @@
...
 import os
+import sys
 from contextlib import contextmanager
-from subprocess import check_call, CalledProcessError
+from subprocess import CalledProcessError
 from time import sleep
+import pylxd
+from six.moves.urllib import parse
+from ws4py.client import WebSocketBaseClient
+from ws4py.manager import WebSocketManager
```
kyrofa: Does this require python3-ws4py?

snapcraft/internal/lxd.py
```diff
@@ -17,9 +17,14 @@
 
 import logging
 import os
+import sys
 from contextlib import contextmanager
-from subprocess import check_call, CalledProcessError
+from subprocess import CalledProcessError
 from time import sleep
+import pylxd
+from six.moves.urllib import parse
```
kyrofa: Seeing six in python3-only code made me double-take. Would `urllib.parse` be better here?

snapcraft/internal/lxd.py
```diff
@@ -40,30 +72,104 @@ def __init__(self, snap_output, tar_filename, project_options):
...
+        logger.info('Pulling {} from {}'.format(dst, src))
+        with open(dst, 'wb') as destination_file:
+            data = self._container.files.get(src)
+            destination_file.write(data)
 
     def _container_run(self, cmd):
-        check_call(['lxc', 'exec', self._container_name, '--'] + cmd)
+        if not self._container:
+            raise CalledProcessError(-1, cmd)
+        print('Executing {}'.format(cmd))
```
kyrofa: Should this be using a logger?

snapcraft/internal/lxd.py
```diff
@@ -40,30 +72,104 @@ def __init__(self, snap_output, tar_filename, project_options):
...
+        })
+        self._fds = response.json()['metadata']['metadata']['fds']
+        operation_id = response.json()['operation'].split('/')[-1]
+        self._parsed = parse.urlparse(
+            self._client.api.operations[operation_id].websocket._api_endpoint)
+        self._manager = WebSocketManager()
+        self._pipe(sys.stdin), self._pipe(sys.stdout), self._pipe(sys.stderr)
+        self._manager.start()
+        while len(self._manager) > 0:
+            sleep(.1)
```
kyrofa: Is there a possibility this will wait forever? Perhaps we should implement a timeout.

snapcraft/internal/lxd.py
```diff
@@ -32,6 +37,33 @@
...
+            self.close()
+            return
+
+        self._manager.add(self)
+
+    def received_message(self, message):
+        if self._io == sys.stdin:
+            return
+
+        if len(message.data) == 0:
```
kyrofa: A simple `if not message.data` would likely be more pythonic.

snapcraft/internal/lxd.py
```diff
@@ -40,30 +72,104 @@ def __init__(self, snap_output, tar_filename, project_options):
...
+            fingerprint = self._get_fingerprint_by_name(
+                'ubuntu:xenial/{}'.format(self._project_options.deb_arch))
+            logger.info('Creating container {} with fingerprint {}'.format(
+                self._container_name, fingerprint[:12]))
+            self._container = self._client.containers.create({
+                'name': self._container_name,
+                'ephemeral': True,
+                'source': {
+                    'type': 'image',
+                    'fingerprint': fingerprint,
```
kyrofa: It seems the API can handle `'image': 'ubuntu/xenial'`. Can it not also handle the arch? Or is there another benefit to using the fingerprint instead?

snapcraft/internal/lxd.py
kyrofa: Ah, I see that's a dependency of pylxd. Should we depend on it directly?

snapcraft/internal/lxd.py
```diff
@@ -40,30 +72,104 @@ def __init__(self, snap_output, tar_filename, project_options):
...
+        with open(dst, 'wb') as destination_file:
+            data = self._container.files.get(src)
+            destination_file.write(data)
 
     def _container_run(self, cmd):
-        check_call(['lxc', 'exec', self._container_name, '--'] + cmd)
+        if not self._container:
+            raise CalledProcessError(-1, cmd)
+        print('Executing {}'.format(cmd))
+        # API call because self._container.execute doesn't expose IO streams
```
kyrofa: Can you explain this a little more? `execute()` blocks, so this makes me think that you simply want to be able to show scrolling output while the command executes within the container rather than blocking, but then you also pipe `stdin`. How is that used?

snapcraft/internal/lxd.py
```diff
@@ -40,30 +72,104 @@ def __init__(self, snap_output, tar_filename, project_options):
...
+        self._pipe(sys.stdin), self._pipe(sys.stdout), self._pipe(sys.stderr)
+        self._manager.start()
+        while len(self._manager) > 0:
+            sleep(.1)
+        response = self._client.api.operations[operation_id].get()
+        exit_status = response.json()['metadata']['metadata']['return']
+        if exit_status is not 0:
+            raise CalledProcessError(exit_status, cmd)
+
+    def _pipe(self, io):
```
kyrofa: Perhaps this should be called `_connect_pipe()`? Its return value is never used.

snapcraft/internal/lxd.py
```diff
@@ -40,30 +72,104 @@ def __init__(self, snap_output, tar_filename, project_options):
...
+            },
+            'wait-for-websocket': True,
+            'interactive': False,
+        })
+        self._fds = response.json()['metadata']['metadata']['fds']
+        operation_id = response.json()['operation'].split('/')[-1]
+        self._parsed = parse.urlparse(
+            self._client.api.operations[operation_id].websocket._api_endpoint)
+        self._manager = WebSocketManager()
+        self._pipe(sys.stdin), self._pipe(sys.stdout), self._pipe(sys.stderr)
```
kyrofa: Please put these statements on their own lines.

snapcraft/internal/lxd.py
```diff
@@ -40,30 +72,104 @@ def __init__(self, snap_output, tar_filename, project_options):
...
+
+    def _get_fingerprint_by_name(self, name):
+        for url in self._client.api.images.get().json()['metadata']:
+            fingerprint = url.split('/')[-1]
+            response = self._client.api.images[fingerprint].get()
+            props = response.json()['metadata']['properties']
+            image_name = '{}:{}/{}'.format(
+                props['os'], props['release'], props['architecture'])
+            if name == image_name:
+                return fingerprint
```
kyrofa: I'm curious why the API is being used directly here instead of iterating over `client.images.all()` and using its `properties` and `fingerprint` attributes?

kalikiana: Because that only returns the fingerprints - but we need to pick an image by name.

snapcraft/internal/lxd.py
kalikiana: I don't see where the API takes a name... fingerprint only takes a fingerprint.

snapcraft/internal/lxd.py
kalikiana: We wait til the pipes close, same as pylxd would out of the box. Different builds can take any amount of time, so a hard timeout would to my mind effectively prevent you from snapping certain things.

snapcraft/internal/lxd.py
kyrofa: Perhaps I misunderstand, but this is from the docs:

```python
config = {'name': 'my-container', 'source': {'type': 'image', 'image': 'ubuntu/trusty'}}
container = client.containers.create(config, wait=True)
```

Isn't `ubuntu/trusty` the name, there?

snapcraft/internal/lxd.py
kyrofa: Correct me if I'm wrong, but does this not do the same thing?

```python
def _get_fingerprint_by_name(self, name):
    for image in client.images.all():
        image_name = '{}:{}/{}'.format(
            image.properties['os'], image.properties['release'], image.architecture)
        if image_name == name:
            return image.fingerprint
```

snapcraft/internal/lxd.py
kalikiana: Those docs must be outdated. If you try that you'll get "Must specify one of alias, fingerprint or properties for init from image" (https://github.com/lxc/lxd/blob/master/lxd/containers_post.go#L122).

snapcraft/internal/lxd.py
kalikiana: Sure, it only returns a fingerprint, but it takes a name argument.

snapcraft/internal/lxd.py
kyrofa: Huh, indeed! Alright, sorry for the noise.

snapcraft/internal/lxd.py
kyrofa: Fair enough.

snapcraft/internal/lxd.py
kyrofa: Okay, thanks for the explanation.

snapcraft/internal/lxd.py
kyrofa: Ah okay, makes sense.

kyrofa:
This is looking good to me, but there's a problem in the tests:

```
Collecting python3-pylxd==2.0.5 (from -r requirements.txt (line 7))

  Could not find a version that satisfies the requirement python3-pylxd==2.0.5 (from -r requirements.txt (line 7)) (from versions: )

No matching distribution found for python3-pylxd==2.0.5 (from -r requirements.txt (line 7))
```

kyrofa:
Hmm, looks like this needs libssl-dev to build.

sergiusens:
This is what I get when I try to cleanbuild

```
sergiusens@snapcraft:~/snapcraft/demos/godd$ snapcraft cleanbuild --debug
Setting up part 'godd' with plugin 'go' and properties {'go-importpath': 'github.com/mvo5/godd', 'plugin': 'go', 'source-type': 'git', 'stage': [], 'source': 'https://github.com/mvo5/godd', 'build-packages': ['gcc', 'libgudev-1.0-dev'], 'snap': []}.
Use of pull-properties in the schema is deprecated.
Plugins should now implement get_pull_properties                                                            
Use of build-properties in the schema is deprecated.
Plugins should now implement get_build_properties                                                           
"GET /1.0 HTTP/1.1" 200 None
"GET /1.0/images HTTP/1.1" 200 67
Creating container snapcraft-undissemblingly-nonbeneficent-ela with fingerprint N/A
"POST /1.0/containers HTTP/1.1" 202 467
"GET /1.0/operations/63bc64e1-a0be-42cc-b856-0f7cb093fab4 HTTP/1.1" 200 398
"GET /1.0/operations/63bc64e1-a0be-42cc-b856-0f7cb093fab4/wait HTTP/1.1" 200 398
Stopping snapcraft-undissemblingly-nonbeneficent-ela
  File "/home/sergiusens/snapcraft/bin/snapcraft", line 31, in <module>
    snapcraft.main.main()
  File "/home/sergiusens/snapcraft/snapcraft/main.py", line 230, in main
    return run(args, project_options)
  File "/home/sergiusens/snapcraft/snapcraft/main.py", line 271, in run
    lifecycle.cleanbuild(project_options),
  File "/home/sergiusens/snapcraft/snapcraft/internal/lifecycle.py", line 288, in cleanbuild
    lxd.Cleanbuilder(snap_filename, tar_filename, project_options).execute()
  File "/home/sergiusens/snapcraft/snapcraft/internal/lxd.py", line 171, in execute
    with self._create_container():
  File "/usr/lib/python3.5/contextlib.py", line 59, in __enter__
    return next(self.gen)
  File "/home/sergiusens/snapcraft/snapcraft/internal/lxd.py", line 161, in _create_container
    }, wait=True)
  File "/home/sergiusens/.local/lib/python3.5/site-packages/pylxd/models/container.py", line 132, in create
    client.operations.wait_for_operation(response.json()['operation'])
  File "/home/sergiusens/.local/lib/python3.5/site-packages/pylxd/models/operation.py", line 29, in wait_for_operation
    operation.wait()
  File "/home/sergiusens/.local/lib/python3.5/site-packages/pylxd/models/operation.py", line 51, in wait
    raise exceptions.LXDAPIException(response)
pylxd.exceptions.LXDAPIException: not found
```

sergiusens:
This is what I get when I want to cleanbuild:
```
sergiusens@snapcraft:~/snapcraft/demos/godd$ snapcraft cleanbuild --debug
Setting up part 'godd' with plugin 'go' and properties {'source': 'https://github.com/mvo5/godd', 'go-importpath': 'github.com/mvo5/godd', 'build-packages': ['gcc', 'libgudev-1.0-dev'], 'plugin': 'go', 'snap': [], 'stage': [], 'source-type': 'git', 'prime': []}.
"GET /1.0 HTTP/1.1" 200 None
"GET /1.0/images HTTP/1.1" 200 67
Creating container snapcraft-inadvertently-anthracitic-kip with fingerprint N/A
"POST /1.0/containers HTTP/1.1" 202 463
"GET /1.0/operations/d0a7bdbc-cb8b-4309-a55b-8c2a360b70c0 HTTP/1.1" 200 394
"GET /1.0/operations/d0a7bdbc-cb8b-4309-a55b-8c2a360b70c0/wait HTTP/1.1" 200 394
Stopping snapcraft-inadvertently-anthracitic-kip
  File "/home/sergiusens/snapcraft/bin/snapcraft", line 31, in <module>
    snapcraft.main.main()
  File "/home/sergiusens/snapcraft/snapcraft/main.py", line 236, in main
    return run(args, project_options)
  File "/home/sergiusens/snapcraft/snapcraft/main.py", line 278, in run
    lifecycle.cleanbuild(project_options),
  File "/home/sergiusens/snapcraft/snapcraft/internal/lifecycle.py", line 288, in cleanbuild
    lxd.Cleanbuilder(snap_filename, tar_filename, project_options).execute()
  File "/home/sergiusens/snapcraft/snapcraft/internal/lxd.py", line 171, in execute
    with self._create_container():
  File "/usr/lib/python3.5/contextlib.py", line 59, in __enter__
    return next(self.gen)
  File "/home/sergiusens/snapcraft/snapcraft/internal/lxd.py", line 161, in _create_container
    }, wait=True)
  File "/home/sergiusens/.local/lib/python3.5/site-packages/pylxd/models/container.py", line 132, in create
    client.operations.wait_for_operation(response.json()['operation'])
  File "/home/sergiusens/.local/lib/python3.5/site-packages/pylxd/models/operation.py", line 29, in wait_for_operation
    operation.wait()
  File "/home/sergiusens/.local/lib/python3.5/site-packages/pylxd/models/operation.py", line 51, in wait
    raise exceptions.LXDAPIException(response)
pylxd.exceptions.LXDAPIException: not found
```

sergiusens:
Apparently I have no fingerprints available

snapcraft/internal/lxd.py
sergiusens:
```
curl -s --unix-socket /var/snap/lxd/common/lxd/unix.socket lxd/1.0/containers -X POST -d '{"name": "blah", "ephemeral": true, "source": {"type": "image", "mode": "pull", "server": "https://cloud-images.ubuntu.com/releases", "protocol": "simplestreams", "alias": "xenial"}}'
```

sergiusens:
closing due to lack of activity. Also pylxd in xenial differs greatly from the on in latter series of ubuntu

kalikiana:
Unfortunate to see this beeing closed, I had been waiting for some review/ help with the tests - the code has worked well for me.
Not sure how pylxd xenial/others is relevant, as far as I'm aware Xenial is the only relevant target for stable snaps anyway.

PR #905, opened 2016-11-15
Snap revision prune

seawaywen:
This prune function will remove all the cached revision files expect the latest revision under the snapcraft’s xdg cache(XDG_CACHE_HOME/snapcraft/{project_name}/revisions),

So far I'm uncertain where should this revision prune function be invoked. now it's called right after the revision cache function is triggered in "snap push".

It's the following branch for PR: #889

LP: #1641512

### DIFF
--- a/snapcraft/_store.py
+++ b/snapcraft/_store.py
@@ -400,6 +400,7 @@ def push(snap_filename, release_channels=None):
     if os.environ.get('DELTA_UPLOADS_EXPERIMENTAL'):
         snap_cache = cache.SnapCache()
         snap_cache.cache(snap_filename, result['revision'])
+        snap_cache.prune(keep_revision=result['revision'])
 
     if release_channels:
         release(snap_name, result['revision'], release_channels)
--- a/snapcraft/internal/cache/_cache.py
+++ b/snapcraft/internal/cache/_cache.py
@@ -33,7 +33,7 @@ class SnapcraftCache:
     def cache(self):
         raise NotImplementedError
 
-    def prune(self):
+    def prune(self, *args, **kwargs):
         raise NotImplementedError
 
 
@@ -41,5 +41,6 @@ class SnapcraftProjectCache(SnapcraftCache):
     """Project specific cache"""
     def __init__(self):
         super().__init__()
+        self.config = snapcraft.internal.load_config()
         self.project_cache_root = os.path.join(
-            self.cache_root, snapcraft.internal.load_config().data['name'])
+            self.cache_root, self.config.data['name'])
--- a/snapcraft/internal/cache/_snap.py
+++ b/snapcraft/internal/cache/_snap.py
@@ -16,9 +16,9 @@
 
 import logging
 import os
-import shutil
 
-from snapcraft.internal.cache._cache import SnapcraftProjectCache
+from snapcraft.file_utils import link_or_copy
+from ._cache import SnapcraftProjectCache
 
 
 logger = logging.getLogger(__name__)
@@ -28,6 +28,7 @@ class SnapCache(SnapcraftProjectCache):
     """Cache for snap revisions."""
     def __init__(self):
         super().__init__()
+        self.snap_cache_dir = self._setup_snap_cache()
 
     def _setup_snap_cache(self):
         snap_cache_path = os.path.join(self.project_cache_root, 'revisions')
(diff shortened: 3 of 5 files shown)


sergiusens:
Now that #889 is in can we rebase or merge with master?

seawaywen:
@sergiusens my source branch rebased from #889, the conflicts were all resolved, I think the simple merge should work.

snapcraft/_store.py
```diff
@@ -547,6 +548,14 @@ def history(snap_name, series, arch):
     print(tabulated_revisions)
 
 
+def latest_rev(snap_name, series=None, arch=None):
```
sergiusens: being pythonic and all and this is not a direct reflection of a cli command, can you call this `get_latest_rev` and maybe add a docstring? Also document the return result.

Is there a change for there not to be a latest rev? Maybe some safe guards for that if necessary

snapcraft/internal/cache/_snap.py
```diff
@@ -39,19 +41,43 @@ def cache(self, snap_filename, revision):
...
-        cached_snap_path = os.path.join(snap_cache_dir, cached_snap)
+        cached_snap_path = os.path.join(self.snap_cache_dir, cached_snap)
         try:
             shutil.copyfile(snap_filename, cached_snap_path)
         except OSError:
             logger.warning(
                 'Unable to cache snap {}.'.format(cached_snap))
         return cached_snap_path
 
+    def prune(self, revision=None):
```
sergiusens: Can this return a list of what was prunned so the caller can print it?

snapcraft/internal/cache/_cache.py
```diff
@@ -33,13 +33,14 @@ def __init__(self):
     def cache(self):
         raise NotImplementedError
 
-    def prune(self):
+    def prune(self, revision=None):
```
sergiusens: `revision` shouldn't be in `SnapcraftCache`, if necessary go with the `**kwargs, *args` trick here for the cases we need to call `super`

sergiusens:
BTW, thanks for the PR

snapcraft/_store.py
seawaywen: as we discussed, remove the *get_lastest_rev* from cache package, we won't have this issue then.

snapcraft/internal/cache/_snap.py
seawaywen: yes, as you suggested, the function will return the prunned list

snapcraft/internal/cache/_cache.py
seawaywen: declared in base class like: *def prune(self, *args, **kwargs)*,
and in subclass it looks like: *def prune(self, keep_revision, *args)

seawaywen:
@sergiusens I have updated the branch with some changes as we discussed yesterday. mainly get rid of the `get_latest_revision` in the cache package's business logic.

snapcraft/internal/cache/_snap.py
```diff
@@ -39,19 +40,47 @@ def cache(self, snap_filename, revision):
...
-        cached_snap_path = os.path.join(snap_cache_dir, cached_snap)
+        cached_snap_path = os.path.join(self.snap_cache_dir, cached_snap)
         try:
             shutil.copyfile(snap_filename, cached_snap_path)
         except OSError:
             logger.warning(
                 'Unable to cache snap {}.'.format(cached_snap))
         return cached_snap_path
 
+    def prune(self, keep_revision, *args):
```
sergiusens: this should be
```python
def prune(self, *, keep_revision):
```

snapcraft/internal/cache/_snap.py
```diff
@@ -39,19 +40,47 @@ def cache(self, snap_filename, revision):
...
+        """Prune the snap revisions beside the keep_revision in XDG cache.
+
+        :returns: pruned file paths list.
+        """
+        if keep_revision is None:
```
sergiusens: This check isn't needed, the one below does the job, if `None` is passed this is what an `int` conversion raises:

```python
int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

snapcraft/internal/cache/_snap.py
```diff
@@ -60,3 +89,22 @@ def _rewrite_snap_filename_with_revision(snap_file, revision):
...
+    split_name_parts = filename.split('_')
+    if len(split_name_parts) != 4:
+        logger.debug('The cached snap filename {} is invalid.'.format(
+            snap_filename))
+        return None
+
+    try:
+        rev = int(split_name_parts[-1])
+    except ValueError:
+        logger.debug('The cached snap filename has invalid revision.')
```
sergiusens: typo: has *an* invalid

snapcraft/internal/cache/_snap.py
```diff
@@ -39,19 +40,47 @@ def cache(self, snap_filename, revision):
...
         """
-        snap_cache_dir = self._setup_snap_cache()
-
         cached_snap = _rewrite_snap_filename_with_revision(
             snap_filename,
             revision)
-        cached_snap_path = os.path.join(snap_cache_dir, cached_snap)
+        cached_snap_path = os.path.join(self.snap_cache_dir, cached_snap)
         try:
             shutil.copyfile(snap_filename, cached_snap_path)
```
sergiusens: I just noticed this from the previous review; mind changing it to

```python
from snapcraft.file_utils import link_or_copy
...
...
...
try:
    link_or_copy(snap_filename, cached_snap_path)
...
```

snapcraft/internal/cache/_snap.py
seawaywen: changed the signature as above

snapcraft/internal/cache/_snap.py
seawaywen: Yes, you're right, thanks pointing out. removed the extra check.

snapcraft/internal/cache/_snap.py
seawaywen: update the typo

PR #906, opened 2016-11-15
indicators: support TERM=dumb

sergiusens:
LP: #1617224


snapcraft/internal/indicators.py
```diff
@@ -36,16 +36,22 @@ def download_requests_stream(request_stream, destination, message=None):
...
+        maxval = total_length
+    elif not total_length and is_dumb_terminal():
+        widgets = [message]
+        maxval = UnknownLength
     else:
-        progress_bar = ProgressBar(
-            widgets=[message, AnimatedMarker()],
-            maxval=UnknownLength)
+        widgets = [message, AnimatedMarker()]
+        maxval = UnknownLength
```
elopio: I find this slightly clearer, but not a lot:

```
maxval = total_length if total_length else UnknownLength
if is_dumb_terminal():
  widgets = [message]
elif maxval == UnknownLength:
  widgets = [message, AnimatedMarker()]
else:
  widgets = [message,
                   Bar(marker='=', left='[', right=']'),
                   ' ', Percentage()]
```

sergiusens: The reason I did it like this is because I went to a talk a while back about presenting these as tables of truth for much faster view of the possible combinations.

sergiusens: My take also separates what varies and what is set from that very distinctly

elopio: fair point. I still like mine better, but I don't have a reason, so you won :)

PR #908, opened 2016-11-16
Let Rust plugin fetch dependencies in pull

ChrisMacNaughton:
Pulling the dependencies in the pull stage allows
building Rust snaps on Launchpad builders,
which are offline during the build stage


sergiusens:
Thanks for the PR, seems the unit tests need updating though.
You can try locally by running

    ./runtests.sh unit

Also might as well run

    ./runtests.sh static

Before pushing again. Thanks again

ChrisMacNaughton:
It looks like the tests failed because travis couldn't update packages?

snapcraft/tests/test_plugin_rust.py
```diff
@@ -86,7 +86,7 @@ def test_pull(self, run_mock, script_mock):
 
         plugin.pull()
 
-        self.assertEqual(1, run_mock.call_count)
+        self.assertEqual(2, run_mock.call_count)
```
elopio: This test would require more than checking that rust was called twice. Please take a look at the next statement that checks the arguments of the first call, and do something similar to the second.

snapcraft/plugins/rust.py
```diff
@@ -54,6 +54,12 @@ def schema(cls):
 
     def __init__(self, name, options, project):
         super().__init__(name, options, project)
+        self.build_packages.extend([
+            'build-essential',
+            'git',
+            'curl',
+            'file',
```
ChrisMacNaughton: These are the things that were required for me to get rustc installed on one of the launchpad builders. Can't find specific documentation about it but that's where this list came from.

ChrisMacNaughton:
@elopio tests should be accurate now, and the integration rust project now pulls in a dependency :smile:

snapcraft/plugins/rust.py
```diff
@@ -54,6 +54,12 @@ def schema(cls):
 
     def __init__(self, name, options, project):
         super().__init__(name, options, project)
+        self.build_packages.extend([
+            'build-essential',
+            'git',
```
kyrofa: Git should be pulled in simply by using a git source. It shouldn't be required here.

snapcraft/plugins/rust.py
```diff
@@ -54,6 +54,12 @@ def schema(cls):
 
     def __init__(self, name, options, project):
         super().__init__(name, options, project)
+        self.build_packages.extend([
+            'build-essential',
```
kyrofa: This is a large meta package. Are you sure you need all of it?

snapcraft/plugins/rust.py
kyrofa: Unless the pull of rust eventually turns into a git clone. Is that the case?

ChrisMacNaughton: Pulling Rust shouldn't clone, but using Cargo for the build can

snaps_tests/demos_tests/test_downloader_with_wiki_parts.py
```diff
@@ -27,7 +27,8 @@ def test_downloader_with_wiki_parts(self):
         snap_path = self.build_snap(self.snap_content_dir)
         self.install_snap(snap_path, 'downloader', '1.0')
         expected = (
-            '.*<title>The leading OS for PC, tablet, phone and cloud '
+            '.*<title>The leading operating system for PCs, tablets, '
```
sergiusens: This should be reverted as it was fixed in master with a more permanent solution.

ChrisMacNaughton:
@sergiusens I've reverted that change, and it looks like that is the issue with the autopkgtests still.

PR #909, opened 2016-11-16
update dependencies for running the tests

ChrisMacNaughton:


ChrisMacNaughton:
Hold off on this, still need something

HACKING.md
```diff
@@ -83,7 +84,7 @@ e.g.; to update the list for 16.04,
 
 Install the needed dependencies.
 
-    sudo apt install build-essential python3-dev libapt-pkg-dev libsodium-dev gcc libffi-dev
+    sudo apt install libarchive13 python3-pip build-essential python3-dev libapt-pkg-dev libsodium-dev gcc libffi-dev squashfs-tools
```
elopio: what about using sudo apt build-dep -y ./ so we don't have to update it every time?
The downside is that it requires the src apt repos.

sergiusens: No, that won't be future proof either; that was the way it was before; it will also install a bunch of python dependencies we don't want

sergiusens:
I am closing this as #910 was a superset of this.
Thanks for the help.

elopio:
I was curious to see what @ChrisMacNaughton was still needing, you ruined the suspense @sergiusens!

sergiusens:
@elopio it was `squashfs-tools`

PR #910, opened 2016-11-16
Update HACKING.md

sergiusens:
Removed a very obsolete section about testing, added it to the end after installing and making sure the correct set of dependencies are in place.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -58,7 +58,7 @@ If you don't have python3 installed, you can use the one from the archives:
 
 Install the needed dependencies.
 
-    sudo apt install build-essential python3-dev libapt-pkg-dev libsodium-dev gcc libffi-dev libarchive13
+    sudo apt install build-essential python3-dev libapt-pkg-dev libsodium-dev gcc libffi-dev libarchive13 squashfs-tools
 
 If installing to `PYTHONHOME` run:
 
--- a/HACKING.md
+++ b/HACKING.md
@@ -4,35 +4,6 @@
 
 To see all the commands and options, run `snapcraft --help`.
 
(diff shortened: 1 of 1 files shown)


HACKING.md
```diff
@@ -81,15 +52,28 @@ e.g.; to update the list for 16.04,
...
+
+### Testing
+
+We assume you have run through the installation instructions, to run all the tests execute:
+
+    ./runtests.sh
+
+You can selectively run a selective group of tests like:
 
+    ./runtests.sh [static|unit|integration|snaps]
```
elopio: This will actually try to set up an all snaps image with ubuntu-device-flash. I think that the right thing to do now is to run the snaps tests locally:
https://github.com/snapcore/snapcraft/pull/911

HACKING.md
```diff
@@ -81,15 +52,28 @@ e.g.; to update the list for 16.04,
...
 ### Installing using pip
 
+If you don't have python3 installed, you can use the one from the archives:
+
+    sudo apt install python3-pip python3-setuptools python3-pkg-resources
+
 Install the needed dependencies.
 
-    sudo apt install build-essential python3-dev libapt-pkg-dev libsodium-dev gcc libffi-dev
+    sudo apt install build-essential python3-dev libapt-pkg-dev libsodium-dev gcc libffi-dev libarchive13
```
elopio: as I suggested in #909, would it be better to use sudo apt build-dep -y ./ ?

elopio:
+1, with one question.

HACKING.md
sergiusens: indeed

PR #913, opened 2016-11-17
help: update stage-packages and build-packages

sergiusens:
The documentation for stage-packages and build-packages
was outdated. Not taking into account the elf crawling logic
and basically incomplete in some other cases.

LP: #1642446

### DIFF
--- a/snapcraft/__init__.py
+++ b/snapcraft/__init__.py
@@ -139,13 +139,17 @@ of the choice of plugin.
   - build-packages: [deb, deb, deb...]
 
     A list of Ubuntu packages to install on the build host before building
-    the part. The files from these packages will not go into the final snap
-    unless they are also explicitly described in stage-packages.
+    the part. The files from these packages typically will not go into the
+    final snap unless they contain libraries that are direct dependencies of
+    binaries within the snap (in which case they'll be discovered via `ldd`),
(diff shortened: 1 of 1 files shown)


snapcraft/__init__.py
```diff
@@ -139,13 +139,17 @@
...
+    the part. The files from these packages typically will not go into the
+    final snap unless they contain libraries that are direct dependencies of
+    binaries within the snap (in which case they'll be discovered via `ldd`),
+    or they are explicitly described in stage-packages.
 
   - stage-packages: [deb, deb, deb...]
 
-    A list of Ubuntu packages must be unpacked in the `stage/` directory.
-    XXX before build? Before stage?
+    A list of Ubuntu packages to be downloaded and unpacked to join the part
```
elopio: I feel like "to join" is not the right thing to say here.
"unpacked into the part directory" ?

elopio:
+1, with one comment about "to join". Feel free to ignore it if you prefer it as is.

sergiusens:
El 17/11/16 a las 12:24, Leo Arias escribió:
>
> +1, with one comment about "to join". Feel free to ignore it if you 
> prefer it as is.
>
Those are Kyle's words, he is techinally the native speaker ;-)

PR #914, opened 2016-11-17
meta: icon is now dedeprecated

sergiusens:
The icon entry in snapcraft.yaml is now (again) the preferred
method to have an icon in a snap, overriding the one in
the `setup` directory if necessary too.

LP: #1610268


elopio:
integration test with icon in the yaml, please.

sergiusens:
@elopio we are full of those

PR #915, opened 2016-11-17
cache: cleanup logic to pass project name

sergiusens:
Instances shouldn't need to have knowledge of the full config. Cleanup the logic a bit.

LP: #1642773


squidsoup:
lgtm :+1:

PR #917, opened 2016-11-18
Fix unittests in armhf

elopio:
LP: #1643029


martinpitt:
I fixed our armhf test machines so that they say `armv7l` for uname -m instead of `armv8`, and retried the xenial/armhf test.

elopio:
Thanks @martinpitt. I'll take a look at the new errors.

elopio:
Closed in favor of #945, to split the integration and snaps tests to separate branches.

PR #918, opened 2016-11-18
create the deltas package/class with xdetal generation implementation

seawaywen:
Add a new package "deltas" and implement the XDelta based delta generation feature.

By specifying the source file and target file, it can generate the delta file base on xdelta algorithm. 
(Flexible to extend to use the other algorithms like *bsdiff* later)

LP: #1642822

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -58,7 +58,7 @@ If you don't have python3 installed, you can use the one from the archives:
 
 Install the needed dependencies.
 
-    sudo apt install build-essential python3-dev libapt-pkg-dev libsodium-dev gcc libffi-dev libarchive13 squashfs-tools
+    sudo apt install build-essential python3-dev libapt-pkg-dev libsodium-dev gcc libffi-dev libarchive13 squashfs-tools xdelta
 
 If installing to `PYTHONHOME` run:
 
--- a/snapcraft/internal/deltas/_deltas.py
+++ b/snapcraft/internal/deltas/_deltas.py
@@ -1,0 +1,230 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2016 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+import logging
+import os
+import subprocess
+import time
+
+from snapcraft import file_utils
+from snapcraft.internal.deltas.errors import (
+    DeltaFormatError,
+    DeltaFormatOptionError,
+    DeltaGenerationError,
+    DeltaToolError,
+)
+
+
+logger = logging.getLogger(__name__)
+
+
+delta_format_options = [
+    'xdelta'
+]
+
+
+class BaseDeltasGenerator:
+    """Class for delta generation
+
+    This class is responsible for the snap delta file generation
+    """
+
+    def __init__(self, *, source_path, target_path,
+                 delta_file_extname='delta', delta_format=None,
+                 delta_tool_path=None):
+
+        self.source_path = source_path
+        self.target_path = target_path
+        self.delta_format = delta_format
+        self.delta_file_extname = delta_file_extname
+        self.delta_tool_path = delta_tool_path
+
+        # some pre-checks
+        self._check_properties()
+        self._check_file_existence()
+        self._check_delta_gen_tool()
+
+    def _check_properties(self):
+        if not self.delta_format:
+            raise DeltaFormatError()
+        if not self.delta_tool_path:
+            raise DeltaToolError()
+        if self.delta_format not in delta_format_options:
+            raise DeltaFormatOptionError(
+                delta_format=self.delta_format,
+                format_options_list=delta_format_options)
+
+    def _check_file_existence(self):
+        if not os.path.exists(self.source_path):
+            raise ValueError(
+                'source file {!r} does not exist, '
+                'please specify a valid source file'.format(self.source_path))
+
+        if not os.path.exists(self.target_path):
+            raise ValueError(
+                'target file {!r} does not exist, '
+                'please specify a valid target file'.format(self.target_path))
+
+    def _check_delta_gen_tool(self):
+        """Check if the delta generation tool exists"""
+        if not file_utils.executable_exists(self.delta_tool_path):
+            raise DeltaToolError(delta_tool=self.delta_tool_path)
+
+    def find_unique_file_name(self, path_hint):
+        """Return a path on disk similar to 'path_hint' that does not exist.
+
+        This function can be used to ensure that 'path_hint' points to a file
+        on disk that does not exist. The returned filename may be a modified
+        version of 'path_hint' if 'path_hint' already exists.
+        """
+        target = path_hint
+        counter = 0
+        while os.path.exists(target):
+            target = "{}-{}".format(path_hint, counter)
+            counter += 1
+        return target
+
+    def _setup_std_output(self, delta_file):
+        """Helper to setup the stdout and stderr for subprocess"""
+        workdir = '/tmp/'
+        _, delta_name = os.path.split(delta_file)
+
+        stdout_path = self.find_unique_file_name(
+            os.path.join(workdir, '{}.out'.format(delta_name)))
+        stdout_file = open(stdout_path, 'wb')
+
+        stderr_path = self.find_unique_file_name(
+            os.path.join(workdir, '{}.err'.format(delta_name)))
+        stderr_file = open(stderr_path, 'wb')
+
+        return workdir, stdout_path, stdout_file, stderr_path, stderr_file
+
+    def _update_progress_indicator(self, proc, progress_indicator):
+        """Update the progress indicator"""
+        # the caller should start the progressbar outside
+        ret = None
+        count = 0
+        ret = proc.poll()
+        while ret is None:
+            if count >= progress_indicator.maxval:
+                progress_indicator.start()
+                count = 0
+            progress_indicator.update(count)
+            count += 1
+            time.sleep(.2)
+            ret = proc.poll()
+        print('')
+        # the caller should finish the progressbar outside
+
+    def make_delta(self, output_dir=None, progress_indicator=None,
+                   is_for_test=False):
+        """Call the delta generation tool to create the delta file.
+
+        returns: generated delta file path
+        """
+        logger.info('Generating {} delta for {}->{}.'.format(
+                self.delta_format, self.source_path, self.target_path))
+
+        if output_dir is not None:
+            # consider creating the delta file in the specified output_dir
+            # with generated filename.
+            if not os.path.exists(output_dir):
+                os.makedirs(output_dir, exist_ok=True)
+
+            _, _file_name = os.path.split(self.target_path)
+            full_filename = os.path.join(output_dir, _file_name)
+            delta_file = self.find_unique_file_name(
+                '{}.{}'.format(full_filename, self.delta_file_extname))
+        else:
+            # create the delta file under the target_path with
+            # the generated filename.
+            delta_file = self.find_unique_file_name(
+                '{}.{}'.format(self.target_path, self.delta_file_extname))
+
+        delta_cmd = self.get_delta_cmd(self.source_path,
+                                       self.target_path,
+                                       delta_file)
+
+        workdir, stdout_path, stdout_file, \
+            stderr_path, stderr_file = self._setup_std_output(delta_file)
+
+        proc = subprocess.Popen(
+            delta_cmd,
+            stdout=stdout_file,
+            stderr=stderr_file,
+            cwd=workdir
+        )
+
+        if progress_indicator:
+            self._update_progress_indicator(proc, progress_indicator)
+        else:
+            proc.wait()
+
+        stdout_file.close()
+        stderr_file.close()
+
+        if self.is_returncode_unexpected(proc):
+            _stdout = _stderr = ''
+            with open(stdout_path) as f:
+                _stdout = f.read()
+            with open(stderr_path) as f:
+                _stderr = f.read()
+
+            # cleanup the testcase std logs
+            if is_for_test:
+                os.remove(stdout_path)
+                os.remove(stderr_path)
+
+            raise DeltaGenerationError(
+                'Could not generate {} delta.\n'
+                'stdout log: {}\n'
+                'stdout: \n{}\n'
+                '---------'
+                'stderr log: {}\n'
+                'stderr: \n{}\n'
+                '---------'
+                'returncode: {}'.format(
+                    self.delta_format,
+                    stdout_path, _stdout,
+                    stderr_path, _stderr,
+                    proc.returncode
+                )
+            )
+
+        self.log_delta_file(delta_file)
+
+        # is used for log file cleanup in unittest
+        if is_for_test:
+            os.remove(stdout_path)
+            os.remove(stderr_path)
+
+        return delta_file
+
+    # ------------------------------------------------------
+    # the methods need to be implemented in subclass
+    # ------------------------------------------------------
+    def get_delta_cmd(self, source_path, target_path, delta_file):
+        """Get the delta generation command line"""
+        raise NotImplementedError
+
+    def is_returncode_unexpected(self, proc):
+        """Check if the subprocess return code is expected"""
+        raise NotImplementedError
+
+    def log_delta_file(self, delta_file):
+        """Log the delta generation result"""
+        pass
--- a/snapcraft/internal/deltas/_xdelta.py
+++ b/snapcraft/internal/deltas/_xdelta.py
@@ -1,0 +1,57 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2016 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+import logging
+import shutil
+import subprocess
+
+from snapcraft.internal.deltas import BaseDeltasGenerator
+
+
+logger = logging.getLogger(__name__)
+
+
+class XDeltaGenerator(BaseDeltasGenerator):
+
+    def __init__(self, *, source_path, target_path):
+        delta_format = 'xdelta'
+        delta_tool_path = shutil.which(delta_format)
+        super().__init__(source_path=source_path,
+                         target_path=target_path,
+                         delta_file_extname='xdelta',
+                         delta_format=delta_format,
+                         delta_tool_path=delta_tool_path)
+
+    def get_delta_cmd(self, source_path, target_path, delta_file):
+        return [
+            self.delta_tool_path,
+            'delta',
+            source_path,
+            target_path,
+            delta_file
+        ]
+
+    def is_returncode_unexpected(self, proc):
+        # Success is exiting with 0 or 1. Yes, really. I know.
+        # https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=212189
+        return proc.returncode not in (0, 1)
+
+    def log_delta_file(self, delta_file):
+        xdelta_output = subprocess.check_output(
+            [self.delta_tool_path, 'info', delta_file],
+            universal_newlines=True)
+        logger.debug('xdelta delta diff generation:\n{}'.format(xdelta_output))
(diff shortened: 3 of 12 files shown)


HACKING.md
```diff
@@ -58,7 +58,7 @@ If you don't have python3 installed, you can use the one from the archives:
 
 Install the needed dependencies.
 
-    sudo apt install build-essential python3-dev libapt-pkg-dev libsodium-dev gcc libffi-dev libarchive13 squashfs-tools
+    sudo apt install build-essential python3-dev libapt-pkg-dev libsodium-dev gcc libffi-dev libarchive13 squashfs-tools xdelta
```
sergiusens: Nice!

snapcraft/internal/deltas/_deltas.py
```diff
@@ -0,0 +1,219 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+delta_formats_options = [
+    'xdelta'
+]
+
+
+class DeltaGenerationFailed(Exception):
```
sergiusens: Can you put all these exceptions in an `errors` module inside the `deltas` package? And in `__init__.py` import errors.

snapcraft/internal/deltas/_xdelta.py
```diff
@@ -0,0 +1,48 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+import logging
+import subprocess
+
+from ._deltas import BaseDeltasGenerator
```
sergiusens: Can we achieve the full length of tests without private imports? I am not saying to make this public, just want to know if we can, refactoring becomes easier that way.

snapcraft/internal/deltas/_deltas.py
seawaywen: Yes, extract the exceptions into _errors module now.

snapcraft/internal/deltas/_xdelta.py
seawaywen: I'm not sure if I 100% understand the point, what I improved is to expose the needed the classes/function in the *snapcraft/internal/deltas/__init__.py*, then in the test class only * import snapcraft.internal*,  the use the class like: snapcraft.internal.deltas.XDeltaGenerator.

Correct me if I misunderstand you.

snapcraft/internal/deltas/_deltas.py
```diff
@@ -0,0 +1,210 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+import logging
+import os
+import subprocess
+
+from ._errors import (
```
sergiusens: This is a good idea but doesn't follow the style we have been going with.

To follow the style it would look more like

snapcraft.deltas.errors:
```python
class DeltaGenerationFailed(...):
   ...


class DeltaFormatIsNoneError(...):
   ...
...
...
```

In `snapcraft/deltas/__init__.py`:
```python
from . import errors
```

And in code that access this from outside the package:

```python
from snapcraft.internal.deltas.errors import DeltaGenerationFailed
...
except DeltaGenerationFailed:
```

Or:
```python
from snapcraft.internal import deltas

except deltas.errors.DeltaGenerationFailed:
```

Also, why `DeltaGenerationFailed` instead of `DeltaGenerationError`?

snapcraft/internal/deltas/_xdelta.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+class XDeltaGenerator(BaseDeltasGenerator):
+
+    delta_format = 'xdelta'
+    delta_file_extname = 'xdelta'
+    delta_tool_path = '/usr/bin/xdelta'
```
sergiusens: is there any value in changing this to
```python
shutil.which('xdelta')
```
?

sergiusens: When snapcraft is a snap this will certainly be a problem (the hardcoded path)

snapcraft/internal/deltas/_xdelta.py
sergiusens: Disregard this comment, it is in the wrong place and I was distracted :-)

snapcraft/internal/deltas/_deltas.py
seawaywen: made the changes by complying current code style.

snapcraft/internal/deltas/_xdelta.py
seawaywen: updated to use shutil.which

snapcraft/internal/deltas/_deltas.py
```diff
@@ -0,0 +1,210 @@
...
+    def log_delta_file(self, delta_file):
+        pass
+
+
+def executable_exists(path):
+    """Return True if 'path' exists and is readable and executable."""
+    return os.path.exists(path) and os.access(path, os.R_OK | os.X_OK)
+
+
+def which(name):
```
seawaywen: removed the own implemented which and use the shutil.which instead.

snapcraft/internal/deltas/_deltas.py
```diff
@@ -0,0 +1,199 @@
...
+        self._check_file_existence()
+        self._check_delta_gen_tool()
+
+    def _check_properties(self):
+        if self.delta_format is None:
+            raise DeltaFormatIsNoneError(
+                'delta_format must be set in subclass')
+        if self.delta_tool_path is None:
+            raise DeltaToolPathIsNoneError(
+                'delta_too_path must be set in subclass')
```
sergiusens: typo here

snapcraft/internal/deltas/_deltas.py
```diff
@@ -0,0 +1,199 @@
...
+        self.target_path = target_path
+        self.pre_check()
+
+    def pre_check(self):
+        self._check_properties()
+        self._check_file_existence()
+        self._check_delta_gen_tool()
+
+    def _check_properties(self):
+        if self.delta_format is None:
```
sergiusens: can the messages here be part of the exception? Also user friendly, specially in the case of tool paths' as that is user visible.

Look at snapcraft.storeapi.errors or snapcraft.errors for inspiration or the ""dogmatic"" way to do it

snapcraft/internal/deltas/_deltas.py
```diff
@@ -0,0 +1,199 @@
...
+        if not os.path.exists(self.target_path):
+            raise ValueError(
+                'target file {} not exist'.format(self.target_path))
+
+    def _check_delta_gen_tool(self):
+        """check if the delta generation tool exists"""
+        if not executable_exists(self.delta_tool_path):
+            delta_path = shutil.which(self.delta_format)
+            if not delta_path:
+                raise DeltaGenerationError(
```
sergiusens: Why this exception here and the `DeltaToolPathIsNoneError` in `_check_for_properties`?

snapcraft/internal/deltas/_deltas.py
```diff
@@ -0,0 +1,199 @@
...
+            os.path.join(workdir, '{}-out'.format(self.delta_format)))
+        stdout_file = open(stdout_path, 'wb')
+
+        stderr_path = self.find_unique_file_name(
+            os.path.join(workdir, '{}-err'.format(self.delta_format)))
+        stderr_file = open(stderr_path, 'wb')
+
+        return workdir, stdout_path, stdout_file, stderr_path, stderr_file
+
+    def make_delta(self, output_dir=None):
```
sergiusens: this needs a parameter to pass in an indicator (which the caller can pass in) as I bet this will take some time.

snapcraft/internal/deltas/_deltas.py
```diff
@@ -0,0 +1,199 @@
...
+    # the methods need to be implemented in subclass
+    # ------------------------------------------------------
+    def get_delta_cmd(self, source_path, target_path, delta_file):
+        raise NotImplementedError
+
+    def log_delta_file(self, delta_file):
+        pass
+
+
+def executable_exists(path):
```
sergiusens: can this go under `file_utils`?

snapcraft/internal/deltas/_xdelta.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+class XDeltaGenerator(BaseDeltasGenerator):
+
+    delta_format = 'xdelta'
+    delta_file_extname = 'xdelta'
+    delta_tool_path = shutil.which('xdelta')
```
sergiusens: isn't the preferred python way to do these initializations in __init__?

snapcraft/internal/deltas/_xdelta.py
```diff
@@ -0,0 +1,49 @@
...
+            self.delta_tool_path,
+            'delta',
+            source_path,
+            target_path,
+            delta_file
+        ]
+
+    def log_delta_file(self, delta_file):
+        logger.debug(
+            "xdelta delta diff generation:\n%s" %
```
sergiusens: single quotes please and use `format` instead of `%`

snapcraft/internal/deltas/errors.py
```diff
@@ -0,0 +1,31 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+class DeltaGenerationError(Exception):
+    """A Delta failed to generate."""
+
+
+class DeltaFormatIsNoneError(Exception):
```
sergiusens: Can this be called just `DeltaFormatError` ?

snapcraft/internal/deltas/errors.py
```diff
@@ -0,0 +1,31 @@
...
+    """A delta format must be set."""
+
+
+class DeltaToolPathIsNoneError(Exception):
```
sergiusens: Can this be called `DeltaToolError`?

snapcraft/internal/deltas/_deltas.py
seawaywen: change to use the same exception: DeltaToolPathIsNoneError

snapcraft/internal/deltas/_deltas.py
seawaywen: updated.

snapcraft/internal/deltas/_deltas.py
seawaywen: moved to file_utils file.

snapcraft/internal/deltas/errors.py
seawaywen: yes, updated.

snapcraft/internal/deltas/errors.py
seawaywen: updated.

snapcraft/internal/deltas/_xdelta.py
seawaywen: moved the properties into the **__init__**  as we discussed.

snapcraft/internal/deltas/_xdelta.py
seawaywen: updated.

snapcraft/internal/deltas/_deltas.py
seawaywen: As you suggested, move the error messages into the related exception class, reused the SnapcraftError  as the base class.

sergiusens: hmm, if you don't mind change this to
```python
if not self.delta_format:
```

snapcraft/internal/deltas/_deltas.py
```diff
@@ -0,0 +1,214 @@
...
+
+    def pre_check(self):
+        self._check_properties()
+        self._check_file_existence()
+        self._check_delta_gen_tool()
+
+    def _check_properties(self):
+        if self.delta_format is None:
+            raise DeltaFormatError()
+        if self.delta_tool_path is None:
```
sergiusens:
```python
if not self.delta_tool_path:
```

snapcraft/internal/deltas/_deltas.py
```diff
@@ -0,0 +1,214 @@
...
+            raise DeltaToolError()
+        if self.delta_format not in delta_formats_options:
+            raise DeltaFormatOptionError(
+                delta_format=self.delta_format,
+                format_options_list=delta_formats_options)
+
+    def _check_file_existence(self):
+        if not os.path.exists(self.source_path):
+            raise ValueError(
+                'source file {} not exist'.format(self.source_path))
```
sergiusens:
```python
The source {!r} does not exist
```

snapcraft/internal/deltas/_deltas.py
```diff
@@ -0,0 +1,214 @@
...
+
+        if not os.path.exists(self.target_path):
+            raise ValueError(
+                'target file {} not exist'.format(self.target_path))
```
sergiusens:
```python
The target file {!r} does not exist
```

snapcraft/internal/deltas/_deltas.py
```diff
@@ -0,0 +1,214 @@
...
+
+    def _check_delta_gen_tool(self):
+        """check if the delta generation tool exists"""
```
sergiusens: uppercase, *Check* instead of _check_

snapcraft/internal/deltas/_deltas.py
```diff
@@ -0,0 +1,214 @@
...
+    delta_format = None
+    delta_file_extname = 'delta'
+    delta_tool_path = None
+
+    def __init__(self, source_path, target_path):
+        self.source_path = source_path
+        self.target_path = target_path
+        self.pre_check()
+
+    def pre_check(self):
```
sergiusens: Since `pre_check` is only used in `__init__` maybe just get rid of it and make it part of `__init__`

snapcraft/internal/deltas/_deltas.py
sergiusens: Also, this might be a better func:

```python
def _get_delta_gen_tool(self, *, delta_format):
    """Return the path to the delta generation tool."""
    delta_tool_path = shutil.which(delta_format)
    if not delta_path:
        raise DeltaToolError(delta_tool=delta_format)

    return delta_tool_path
```

snapcraft/internal/deltas/_deltas.py
```diff
@@ -0,0 +1,214 @@
...
+    """Class for delta generation
+
+    This class is responsible for the snap delta file generation
+    """
+
+    delta_format = None
+    delta_file_extname = 'delta'
+    delta_tool_path = None
+
+    def __init__(self, source_path, target_path):
```
sergiusens: Rewrite this as 
```python
def __init__(self, *, source_path, target_path, delta_file_extname='delta', delta_format=None, delta_tool_path=None):
...
...
    self.delta_tool_path = self._get_delta_gen_tool(delta_format=delta_format)
```
```

File: snapcraft/internal/deltas/_xdelta.py [hunk 22]
```diff
@@ -0,0 +1,49 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class XDeltaGenerator(BaseDeltasGenerator):
+
+    def __init__(self, source_path, target_path):
+        self.delta_format = 'xdelta'
+        self.delta_file_extname = 'xdelta'
+        self.delta_tool_path = shutil.which('xdelta')
```
sergiusens: Instead of these three call `super` with the arguments, it should do the right thing

sergiusens:
This might also be easier to review if we split this in two.

snapcraft/internal/deltas/_deltas.py
seawaywen: update.

snapcraft/internal/deltas/_deltas.py
seawaywen: updated.

snapcraft/internal/deltas/_deltas.py
seawaywen: call the precheck()'s methods in `__init__` directly

snapcraft/internal/deltas/_deltas.py
seawaywen: I kept the old `_check_delta_gen_tool` name, since this function is used to check if the passed delta_tool_path is executable and existed. After some cleanups, the logic now should be more clear.

seawaywen:
according to your comments,
1.  I removed the class attributes, and put the parameters into __init__
2. did some cleanup for the delta tool generator. use the shutil to identify the delta tool path, pass it 
via super(), then check the executable and exists

snapcraft/internal/deltas/_deltas.py
```diff
@@ -0,0 +1,210 @@
...
+            self._update_progress_indicator(proc, progress_indicator)
+        else:
+            proc.wait()
+
+        stdout_file.close()
+        stderr_file.close()
+
+        # Success is exiting with 0 or 1. Yes, really. I know.
+        # https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=212189
+        if proc.returncode not in (0, 1):
```
thomir: Only xdelta returns 1 (in the case where src and target are the same, so the output delta is empty), yet it's being checked in the base class for all current and future delta algorithms.

sergiusens: Good catch

seawaywen: check the returncode in the subclass instead.

snapcraft/internal/deltas/_deltas.py
```diff
@@ -0,0 +1,211 @@
...
+    DeltaFormatOptionError,
+    DeltaGenerationError,
+    DeltaToolError,
+)
+
+
+logger = logging.getLogger(__name__)
+
+
+delta_formats_options = [
```
sergiusens: this should be named `delta_format_options` I think

snapcraft/internal/deltas/_deltas.py
```diff
@@ -0,0 +1,211 @@
...
+                format_options_list=delta_formats_options)
+
+    def _check_file_existence(self):
+        if not os.path.exists(self.source_path):
+            raise ValueError(
+                'source file {!r} not exist'.format(self.source_path))
+
+        if not os.path.exists(self.target_path):
+            raise ValueError(
+                'target file {!r} not exist'.format(self.target_path))
```
sergiusens: A user would see this so we need a user friendly sentence here, something like

'The target file {!r} does not exist' (and something on how to solve it).

snapcraft/internal/deltas/_deltas.py
```diff
@@ -0,0 +1,211 @@
...
+            raise DeltaToolError()
+        if self.delta_format not in delta_formats_options:
+            raise DeltaFormatOptionError(
+                delta_format=self.delta_format,
+                format_options_list=delta_formats_options)
+
+    def _check_file_existence(self):
+        if not os.path.exists(self.source_path):
+            raise ValueError(
+                'source file {!r} not exist'.format(self.source_path))
```
sergiusens: A user would see this so we need a user friendly sentence here, something like

'The source file {!r} does not exist' (and something on how to solve it).

snapcraft/internal/deltas/_deltas.py
```diff
@@ -0,0 +1,211 @@
...
+        if not file_utils.executable_exists(self.delta_tool_path):
+            raise DeltaToolError(delta_tool=self.delta_tool_path)
+
+    def find_unique_file_name(self, path_hint):
+        """Return a path on disk similar to 'path_hint' that does not exist.
+
+        This function can be used to ensure that 'path_hint' points to a file
+        on disk that does not exist. The returned filename may be a modified
+        version of 'path_hint' if 'path_hint' already exists.
+
```
sergiusens: Is it ok to remove this empty line?

snapcraft/internal/deltas/_deltas.py
```diff
@@ -0,0 +1,211 @@
...
+            stderr_path, stderr_file = self._setup_std_output(self.source_path)
+
+        proc = subprocess.Popen(
+            delta_cmd,
+            stdout=stdout_file,
+            stderr=stderr_file,
+            cwd=workdir
+        )
+
+        if progress_indicator is not None and isinstance(progress_indicator,
```
sergiusens: Can this be reformated to
```python
if progress_indicator:
    self._update_progress_indicator(proc, progress_indicator)
else:
    proc.wait()
```

The benefits here is that a wrong instance would not silently fail. We should also document the function telling it we expect a ProgressBar type indicator.

snapcraft/internal/deltas/_deltas.py
```diff
@@ -0,0 +1,211 @@
...
+                )
+            )
+
+        self.log_delta_file(delta_file)
+
+        return delta_file
+
+    # ------------------------------------------------------
+    # the methods need to be implemented in subclass
+    # ------------------------------------------------------
```
sergiusens: Can we add some docstrings to these "to-be implemented" methods?

snapcraft/internal/deltas/_deltas.py
seawaywen: glad to rename it.

snapcraft/internal/deltas/_deltas.py
seawaywen: removed.

snapcraft/internal/deltas/_deltas.py
seawaywen: Yes, added.

snapcraft/internal/deltas/_deltas.py
seawaywen: addressed the change.

snapcraft/internal/deltas/_deltas.py
seawaywen: updated to: 
The source file {!r} does not exist, please specify a valid source file.

snapcraft/internal/deltas/_deltas.py
seawaywen: updated to: 
The target file {!r} does not exist, please specify a valid target file.

seawaywen:
as we discussed , now it put delta stdout/stderr logs into /tmp, and if delta generation raise exception, then it will mention the log files with location.

PR #920, opened 2016-11-18
pluginhandler: ensure staged files are included in the prime step.

josepht:
If the 'prime' fileset is made up of '*/' or only includes '-XXX'
entries combine the 'stage' fileset as well.

LP:#1623279
LP:#1605164


josepht:
retest this please

elopio:
Hey Joe. We disconnected Jenkins from the PRs, so to retry you will have to use the script in the tools directory. I've done that for you.

elopio:
I think this requires integration tests. What do you think @josepht ?

elopio:
The error you are getting is because you need to rebase master. I have already fixed it.

snapcraft/internal/pluginhandler.py
```diff
@@ -322,6 +322,13 @@ def clean_build(self, hint=''):
     def migratable_fileset_for(self, step):
         plugin_fileset = self.code.snap_fileset()
         fileset = (getattr(self.code.options, step, ['*']) or ['*']).copy()
+        # If we're priming and we don't have an explicit set of files to prime
+        # include the files from the stage step
+        if (step == 'prime' or step == 'snap') and fileset == ['*']:
+            stage_fileset = (
+                getattr(self.code.options, 'stage', ['*']) or ['*']).copy()
```
kyrofa: We use this in a few different places, and it's a little tough to read. Can we extract this logic into a less compact helper function?

snapcraft/internal/pluginhandler.py
```diff
@@ -874,3 +881,23 @@ def check_for_collisions(parts):
...
+
+    to_combine = False
+    # combine if fileset_1 has a wildcard
+    # XXX: should this only be a single wildcard and possibly excludes?
+    if '*' in fileset_1:
+        to_combine = True
+        fileset_1.remove('*')
+
+    # combine if fileset_1 is only excludes
+    if set([x[0] for x in fileset_1]) == set('-'):
```
kyrofa: Would it be more clear (and maybe efficient) to create the set straight away? `{x[0] for x in fileset_1}`

snapcraft/tests/test_pluginhandler.py
```diff
@@ -2359,3 +2359,30 @@ def test_fail_to_load_magic_raises_exception(self, mock_magic):
...
+        expected_fileset = ['a', 'b']
+        combined_fileset = pluginhandler._combine_filesets(
+            fileset_1, fileset_2)
+        self.assertEqual(set(expected_fileset), set(combined_fileset))
+
+    def test__combine_filesets_implicit_wildcard(self):
+        fileset_1 = ['-a']
+        fileset_2 = ['a', 'b']
+
+        expected_fileset = ['a', '-a', 'b']
```
kyrofa: What does such a fileset mean? Will `a` be copied, or no?

josepht:
Ultimately no 'a' will not be copied.  The method that handles
includes/excludes will filter that out.

snapcraft/internal/pluginhandler.py
```diff
@@ -322,6 +322,13 @@ def clean_build(self, hint=''):
     def migratable_fileset_for(self, step):
         plugin_fileset = self.code.snap_fileset()
         fileset = (getattr(self.code.options, step, ['*']) or ['*']).copy()
+        # If we're priming and we don't have an explicit set of files to prime
+        # include the files from the stage step
+        if (step == 'prime' or step == 'snap') and fileset == ['*']:
```
kyrofa: I don't think the `snap` step ever reaches this function, does it?

josepht:
It's used during 'prime'.  I think I chose 'snap' because 'prime' is
deprecated.

snapcraft/internal/pluginhandler.py
kyrofa: `prime` isn't deprecated. Perhaps you were thinking `strip`?

josepht: Ah, you're right.  I've switched it to 'prime' to be consistent with the calling method.

snapcraft/internal/pluginhandler.py
```diff
@@ -874,3 +881,23 @@ def check_for_collisions(parts):
         # And add our files to the list
         parts_files[part.name] = {'files': part_files,
                                   'installdir': part.installdir}
+
+
+def _combine_filesets(fileset_1, fileset_2):
```
josepht: @sergiusens What are your thoughts on this?  I think @kyrofa has a good point and the logic shouldn't be too complex.

snapcraft/internal/pluginhandler.py
josepht: I'm not sure I understand your question.  Do you mean simply pulling the first set(...) out into a variable or something else?

kyrofa: I just meant changing this to `if {x[0] for x in fileset_1} == set('-')` .

josepht: Does that create a set()?

kyrofa: Indeed it does:

```python
>>> a = ['foo', 'bar', 'baz']
>>> {x[0] for x in a}
{'f', 'b'}
```

kyrofa: Note that you can do a hash that way as well:

```python
>>> a = ['foo', 'bar', 'baz']
>>> {x[0]: x for x in a}
{'f': 'foo', 'b': 'baz'}
```

snapcraft/internal/pluginhandler.py
```diff
@@ -417,13 +417,28 @@ def clean_build(self, hint=''):
 
     def migratable_fileset_for(self, step):
         plugin_fileset = self.code.snap_fileset()
-        fileset = (getattr(self.code.options, step, ['*']) or ['*']).copy()
+        if step == 'prime':
+            step = 'snap'
+        fileset = self._get_fileset(step).copy()
+        # If we're priming and we don't have an explicit set of files to prime
+        # include the files from the stage step
+        if step == 'prime' and fileset == ['*']:
```
elopio: In case you don't see the comment on the other branch, This is a condition that will never be True.
if step is 'prime', the if above will change it to 'snap'.

integration_tests/test_snap.py
```diff
@@ -169,3 +169,17 @@ def test_error_on_bad_yaml(self):
             "Issues while validating snapcraft.yaml: found character '\\t' "
             "that cannot start any token on line 13 of snapcraft.yaml",
             str(error.output))
+
+    def test_prime_includes_stage_fileset(self):
+        project_dir = 'prime-from-stage'
```
elopio: This project is nowhere. Missing add?
Also, I think this should be in test_prime.py

snapcraft/internal/pluginhandler.py
sergiusens: @josepth so @kyrofa is correct that if `stage` and `prime` contradict each other a conflict error should be raised.

integration_tests/test_snap.py
sergiusens: @josepht are you going to address this?

josepht: @sergiusens fixed.

integration_tests/snaps/prime-from-stage-excludes/a
```diff
@@ -0,0 +1 @@
+a
```
sergiusens: this file is echoed, right? So is it needed? Same for the other 2 below

integration_tests/test_prime.py
```diff
@@ -0,0 +1,60 @@
...
+    def test_prime_includes_stage_fileset(self):
+        project_dir = 'prime-from-stage'
+
+        self.run_snapcraft('prime', project_dir)
+        self.assertThat(
+            os.path.join(project_dir, 'prime', 'a'),
+            FileExists())
+        self.assertThat(
+            os.path.join(project_dir, 'prime', 'c'),
+            FileExists())
```
kyrofa: Haha, you choose 'a', 'b', and 'c' as the file names, but then check them out of order? You're killing my OCD here man :stuck_out_tongue: . Not enough to change it, just saying. You're mean.

integration_tests/snaps/prime-from-stage/snapcraft.yaml
```diff
@@ -0,0 +1,16 @@
+name: my-snap-name
```
kyrofa: Is there a reason these couldn't be the same snap, two parts?

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -1071,3 +1087,40 @@ def check_for_collisions(parts):
...
+    """Combine filesets if the first is an explicit or implicit wildcard."""
+
+    f1_excludes = set(_get_excludes(fileset_1))
+    f2_includes = set(_get_includes(fileset_2))
+
+    contradicting_fileset = set.intersection(f1_excludes, f2_includes)
+
+    if contradicting_fileset:
+        raise EnvironmentError('File conflicts: {key!r}'.format(
+                               key=contradicting_fileset))
```
kyrofa: 1) This isn't really an `EnvironmentError`, is it? Should we make a new Snapcraft error class for this?
2) This error is going to look like "File conflicts: {a, b}." I don't know how much that will inform a brand new snapcraft user, and we have the information at hand (in the caller if necessary) to make this a more informative/helpful message.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -426,13 +426,29 @@ def clean_build(self, hint=''):
 
     def migratable_fileset_for(self, step):
         plugin_fileset = self.code.snap_fileset()
-        fileset = (getattr(self.code.options, step, ['*']) or ['*']).copy()
+        fileset_step = step
+        if step == 'prime':
+            fileset_step = 'snap'
```
kyrofa: Now that #1006 has landed you should be able to just use 'prime'.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -1100,29 +1100,33 @@ def _get_excludes(fileset):
...
+def _combine_filesets(starting_fileset, modifying_fileset):
+    """
+    Combine filesets if modifying_fileset is an explicit or implicit
+    wildcard.
+    """
 
-    f1_excludes = set(_get_excludes(fileset_1))
-    f2_includes = set(_get_includes(fileset_2))
+    starting_excludes = set(_get_excludes(starting_fileset))
+    modifying_includes = set(_get_includes(modifying_fileset))
```
kyrofa: Good call on the renames, that seems more clear.

integration_tests/snaps/prime-from-stage-excludes/Makefile
```diff
@@ -0,0 +1,10 @@
+
```
kyrofa: Can this snap be removed, now?

josepht: Good catch.  I removed it.

snapcraft/internal/errors.py
```diff
@@ -39,6 +39,14 @@ class MissingState(Exception):
     pass
 
 
+class FileConflictError(SnapcraftError):
+
+    fmt = (
+        'The following files would create a conflict in the snap: '
+        '{fileset!r}'
```
kyrofa: This is still pretty generic. However, is this exception ever raised except in this very specific context? Must we keep it generic? Something like "The following files have been excluded by the 'stage' keyword, but included by the 'prime' keyword: {fileset!r}" tells the user how to solve the problem.

integration_tests/snaps/prime-from-stage-excludes/Makefile
kyrofa: Thanks for combining those. We just have so many snaps in there already!

snapcraft/internal/errors.py
josepht: I updated this as you've suggested.

PR #921, opened 2016-11-19
Add support for hooks.

kyrofa:
Hooks are now supported in snapd and the click reviewer tools. Time to finish off LP: [#1586465](https://bugs.launchpad.net/snapcraft/+bug/1586465) by supporting them in snapcraft as well. This is done declaratively via a `hooks` option in the snapcraft.yaml, for example:

    hooks:
      configure:
        command: run-my-hook
        plugs: [network]

This works exactly like apps, where `run-my-hook` should be a binary available within the snap. Then during the prime step, snapcraft will generate the hook (the configure hook in this example), which calls the command supplied.

### DIFF
--- a/snapcraft/internal/meta.py
+++ b/snapcraft/internal/meta.py
@@ -208,17 +208,8 @@ class _SnapPackaging:
         if os.path.exists(wrappath):
             os.remove(wrappath)
 
-        wrapexec = '$SNAP/{}'.format(execparts[0])
-        if not os.path.exists(exepath) and '/' not in execparts[0]:
-            _find_bin(execparts[0], self._snap_dir)
-            wrapexec = execparts[0]
-        else:
-            with open(exepath, 'rb') as exefile:
-                # If the file has a she-bang, the path might be pointing to
-                # the local 'parts' dir. Extract it so that _write_wrap_exe
-                # will have a chance to rewrite it.
-                if exefile.read(2) == b'#!':
-                    shebang = exefile.readline().strip().decode('utf-8')
+        wrapexec, shebang = _extract_binary_and_shebang(
+            execparts[0], self._snap_dir)
 
         self._write_wrap_exe(wrapexec, wrappath,
                              shebang=shebang, args=execparts[1:])
@@ -254,22 +245,28 @@ class _SnapPackaging:
 
         command_parts = shlex.split(command)
 
-        shebang = None
-        hook_exec = '$SNAP/{}'.format(command_parts[0])
-        exepath = os.path.join(self._snap_dir, command_parts[0])
-        if not os.path.exists(exepath) and '/' not in command_parts[0]:
-            _find_bin(command_parts[0], self._snap_dir)
-            hook_exec = command_parts[0]
-        else:
-            with open(exepath, 'rb') as f:
-                # If the file has a she-bang, the path might be pointing to
-                # the local 'parts' dir. Extract it so that _write_wrap_exe
-                # will have a chance to rewrite it.
-                if f.read(2) == b'#!':
-                    shebang = f.readline().strip().decode('utf-8')
+        hook_exec, shebang = _extract_binary_and_shebang(
+            command_parts[0], self._snap_dir)
 
         self._write_wrap_exe(hook_exec, hook_path,
                              shebang=shebang, args=command_parts[1:])
+
+
+def _extract_binary_and_shebang(binary, basedir):
+    shebang = None
+    exepath = os.path.join(basedir, binary)
+    if not os.path.exists(exepath) and '/' not in binary:
+        _find_bin(binary, basedir)
+    else:
+        binary = '$SNAP/{}'.format(binary)
+        with open(exepath, 'rb') as f:
+            # If the file has a she-bang, the path might be pointing to
+            # the local 'parts' dir. Extract it so that _write_wrap_exe
+            # will have a chance to rewrite it.
+            if f.read(2) == b'#!':
+                shebang = f.readline().strip().decode('utf-8')
+
+    return (binary, shebang)
 
 
 def _find_bin(binary, basedir):
--- a/snapcraft/internal/meta.py
+++ b/snapcraft/internal/meta.py
@@ -182,6 +182,20 @@ class _SnapPackaging:
 
         os.chmod(wrappath, 0o755)
 
+    def _wrap_apps(self, apps):
+        for app in apps:
+            cmds = (k for k in ('command', 'stop-command') if k in apps[app])
+            for k in cmds:
+                try:
+                    apps[app][k] = self._wrap_exe(
+                        apps[app][k], '{}-{}'.format(k, app))
+                except CommandError as e:
+                    raise EnvironmentError(
+                        'The specified command {!r} defined in the app {!r} '
+                        'does not exist or is not executable'.format(
+                            str(e), app))
+        return apps
+
     def _wrap_exe(self, command, basename=None):
         execparts = shlex.split(command)
         exepath = os.path.join(self._snap_dir, execparts[0])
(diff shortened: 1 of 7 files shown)


kyrofa:
Note that I'd like to make a snap test for this as well, but I need to talk to @elopio about that. I figure it wouldn't hurt to make the PR anyway-- I'll do that first thing Monday.

snapcraft/internal/meta.py
```diff
@@ -207,6 +211,33 @@ def _wrap_exe(self, command, basename=None):
...
+        if not os.path.exists(exepath) and '/' not in command_parts[0]:
+            _find_bin(command_parts[0], self._snap_dir)
+            hook_exec = command_parts[0]
+        else:
+            with open(exepath, 'rb') as f:
+                # If the file has a she-bang, the path might be pointing to
+                # the local 'parts' dir. Extract it so that _write_wrap_exe
+                # will have a chance to rewrite it.
+                if f.read(2) == b'#!':
+                    shebang = f.readline().strip().decode('utf-8')
```
elopio: can we refactor these lines into a function so they are not duplicated in wrap_exe?

snapcraft/internal/meta.py
```diff
@@ -221,6 +252,25 @@ def _wrap_apps(self, apps):
...
+    def _wrap_hooks(self, hooks):
+        for hook in hooks:
+            try:
+                command = hooks[hook]['command']
+
+                # The 'command' is for snapcraft only, not snapd. Remove it,
+                # and if that makes the dict empty, set it to None instead.
+                del hooks[hook]['command']
+                if len(hooks[hook]) == 0:
+                    hooks[hook] = None
```
kyrofa: I have a dict of dicts. I'm removing an item from a sub-dict, not the dict being traversed. Is that still not okay?

snapcraft/internal/meta.py
```diff
@@ -207,6 +211,33 @@ def _wrap_exe(self, command, basename=None):
 
         return os.path.relpath(wrappath, self._snap_dir)
 
+    def _generate_hook(self, hook, command):
```
kyrofa: Agreed, I just wanted to conform to the order already there for the binary wrappers. But reordering them as well isn't that much of a change, so I did that, too.

niemeyer:
Please hold off landing this until we discuss in more detail.

Offering a "command" looks very strange and unlike what we had been doing and talking about. Hooks are unlike applications in that this is not snap calling something that the user has. Hooks need to implement logic which is particular to snapd itself, and they will want to do that exchanging information with snapd via snapctl or the respective API.

As such, why are we requesting a "command" there? The command is the hook name, which should live in a hooks directory next to other hooks which people will learn can be expected there. Having everybody defining their hooks in arbitrary places and arbitrary arguments will miss the chance of defining a recognizable pattern across the whole ecosystem.

sergiusens:
@niemeyer maybe `command` is not the right name, but we did agree to use a declarative form in Bucharest. User feedback is that `by-location` or `by-contention` is not the best. We can have a follow-up discussion if need be.

kyrofa:
@niemeyer developers are already used to how apps work-- ship a binary, connect the dots with `command`, and away they go. This accomplishes the same goal for hooks while remaining familiar. The only difference is the `command` they specify isn't runnable by the user, but by snapd upon a particular event.

niemeyer:
@sergiusens One or another you'll have a command being fired which will be a procedural executable. So saying "we want declarative" doesn't really help much. A better mantra is "we want sanity". :)

I suggest instead defining a location where hooks will be found, with the required proper names which we'll want people to be aware of. We can allow people to tweak that location as well. Brainstorm strawman:
```
hooks:
    path: hooks/
    configure:
        plugs: [foo, bar]
```

If the hook doesn't require any plugs or slots, then it doesn't need to be declared. It's enough for it to simply exist inside the directory (that's how snapd works as well).

niemeyer:
@kyrofa Again, hooks are not apps. There are great reasons for apps to be organized the way they are today, and those reasons do not apply to the implementation of hooks. You're doing a disservice to people by forcing them to think how they want to name their hooks and which parameters they want to pass the hook, because hooks do not accept these, and have a known name.  I'll be writing documentation and telling people about how hooks work, and you'll need to talk to those people again to explain that actually snapcraft doesn't behave that way, but rather offers an inconvenient layer on top of snapd's semantics. Not really a great idea, IMO.

kyrofa:
> You're doing a disservice to people by forcing them to think how they want to name their hooks [...]

Is this not simply flexibility? Now people can name them as they see fit and they'll still work.

> [...] and which parameters they want to pass the hook, because hooks do not accept these

Just because hooks don't accept parameters at run-time doesn't mean developers may never want to use some in the hook itself. Let's say we had a mysql snap, and an `install` hook. All the developer wants to do is initialize mysql upon install. In that case, this PR enables something simple like this:

```yaml
hooks:
  install:
    command: mysqld --initialize-insecure --basedir="$SNAP" --datadir="$SNAP_DATA/mysql" 
```

Without this ability the developer would have to write a script to do the same thing. It's another hoop to jump through.

niemeyer:
> Is this not simply flexibility? Now people can name them as they see fit and they'll still work.

Compare these two alternatives: on one of them you can get into a directory and tell exactly which hooks the snap offers. On the other, each snaps has a different way to do things, with almost no benefit.

> Just because hooks don't accept parameters at run-time doesn't mean developers may never want to > use some in the hook itself. Let's say we had a mysql snap, and an `install` hook. All the developer
> wants to do is initialize mysql upon install. In that case, this PR enables something simple like this:

There's no install hook right now. It's called "configure" today, which will most often contain more than a single line, and often be written in other languages than just shell. So the apparent simplicity is not simple at all. Instead of telling people that they have a directory with hooks of well defined names running on a known environment (it's snapd's land), which they will often just copy from their preferred example project, you need to tell people that this only works for some cases and not for others, and that "mysqld might call snapctl" (what!?), and offer suggest schemas that may be followed or not based on personal preference. See the issue?

kyrofa:
> Compare these two alternatives: on one of them you can get into a directory and tell exactly which hooks the snap offers. On the other, each snaps has a different way to do things, with almost no benefit.

On one, you need to look in the directory of hooks as well as the YAML to determine their plugs (if any). On the other, you simply look at the `snapcraft.yaml`, where all hooks will be specified along with everything else, including any plugs they may or may not require. No need to look further.

> There's no install hook right now.

We have no idea what hooks will exist in the future. Just because the `configure` hook is currently all-encompassing doesn't mean they'll all be that way. The point is that this works for a number of use-cases.

niemeyer:
> We have no idea what hooks will exist in the future.

No, but we know that the ones we do have and the ones that are coming do not work that way, and we know that they have expected names, that they likely won't be alone, and that they support snapctl to be able to access configuration and other options. We are creating a convention for how to organize event handlers, instead of a convention for how to call commands. Hooks.. events.. not apps.

sergiusens:
@niemeyer just to be esoteric on corner cases how would this work with a JAVA based hook or a python2 one?

niemeyer:
It's not much different from a shell hook really. We'll want a Python library that can talk to snapd for snapctl-related activities, but otherwise it's a normal Python application doing event-related activities (health-checking, configuration, etc).

kyrofa:
> It's not much different from a shell hook really. We'll want a Python library that can talk to snapd for snapctl-related activities, but otherwise it's a normal Python application doing event-related activities (health-checking, configuration, etc).

It is though, because it relies on something contained within the snap (java or python) as opposed to something on the system (e.g. /bin/sh). There are many other examples (e.g. ROS). There's a lot of environmental setup that developers are used to Snapcraft handling for them by way of generating the app wrappers. This PR allows those expectations to continue to hold true for hooks.

niemeyer:
@kyrofa Please let that sync for a moment: if people cannot create *programs* to handle an event, the system is meaningless.

niemeyer:
Can we please have a call to discuss the design here?

sergiusens:
We have time for this PR as we need to wait at least for the latest snapd to make it to xenial-updates, yakkety-updates and zesty-updates. No rush

mardy:
In case you haven't reached a decision yet, one possible compromise would be to make the "command" key optional: if it's given, snapcraft will generate the hook script for you, and if it's not given, it should just check that a hook executable with the proper name is found, and include it in the snap.

morphis:
What is the status on this? Did any discussion already happen? We have hook support in Ubuntu Core since beginning of November and still can't use it as snapcraft is missing support for it and we can get our snaps reliable build and published otherwise by the CI infrastructure we have in place.

niemeyer:
No progress. We need to have a call to discuss this.

kyrofa:
An invite for such a call went out Tuesday, scheduled for this coming Wednesday.

kyrofa:
This has been redesigned and placed behind a prerequisite feature. This PR is no longer valid.

PR #923, opened 2016-11-22
pluginhandler: source management moved to the core

sergiusens:
LP: #1621601


snapcraft/sources.py
```diff
@@ -25,4 +25,3 @@
 from snapcraft.internal.sources import Tar         # noqa
 from snapcraft.internal.sources import Local       # noqa
 from snapcraft.internal.sources import Zip         # noqa
-from snapcraft.internal.sources import get         # noqa
```
elopio: This is breaking nextcloud:
https://github.com/nextcloud/nextcloud-snap/blob/master/parts/plugins/x-php.py#L91

elopio:
These work:
https://git.launchpad.net/~checkbox-dev/plainbox-provider-snappy/+git/packaging
https://github.com/docker-snap/docker
https://github.com/magne4000/quassel-webserver
https://github.com/conjure-up/conjure-up
https://github.com/juju-solutions/charm-pkg
https://github.com/getnikola/nikola/
https://github.com/RocketChat/Rocket.Chat/

Custom plugins:
https://github.com/elopio/syncthing/commit/25c955ccf70dd1a1fec79be61817843abb312766
https://github.com/elopio/syncthing-cli/blob/snapcraft/parts/plugins/x-go.py
https://github.com/elopio/tg

From the playpen:
mirage-os
kodi-stable
mpv
texworks
tyrant-unleashed-optimizer
ubuntukylin-icon-theme

sergiusens:
@elopio strange that someone would put source in `build-properties` instead of `pull-properties`!

elopio:
@sergiusens strange indeed. I wonder what that dude was thinking about. I doubt somebody else would have been using those properties, so it seems to be safe enough to break that behaviour. Maybe add a small note somewhere.

sergiusens:
@elopio I've created https://github.com/maas/meta-maas/pull/11 for the maas issue

kyrofa:
I pushed my update directly here, which should solve the IPFS issues. Note that I also made the `get_{pull,build}_properties()` methods class methods. If we don't like that let me know.

sergiusens:
@kyrofa I like these changes! And how wonderful how github allows anyone to push here now :wink:

I guess, well I hope that it is because you are a contributor to this branch as I merged some code from you, but still, it doesn't mean I want ALL your code :stuck_out_tongue:

kyrofa:
> I guess, well I hope that it is because you are a contributor to this branch as I merged some code from you, but still, it doesn't mean I want ALL your code :stuck_out_tongue: 

Haha! It's because you left the "Allow edits from maintainers" checkbox checked.

snapcraft/sources.py
elopio: Solved.

snapcraft/internal/parser.py
```diff
@@ -191,12 +191,11 @@ def _process_entry(data):
...
-    class Options:
-        source = origin
-        source_type = origin_type
-
-    options = Options()
-    sources.get(origin_dir, None, options)
+    source_handler = sources.get_source_handler(origin,
+                                                source_type=origin_type)
+    handler = source_handler(origin, source_dir=origin_dir)
+    repo.check_for_command(handler.command)
```
kyrofa: This doesn't quite feel like the right place to do this. Shouldn't `handler.pull()` be the one checking that it can be called instead of putting that responsibility on the caller?

snapcraft/plugins/go.py
```diff
@@ -111,7 +112,7 @@ def pull(self):
         super().pull()
         os.makedirs(self._gopath_src, exist_ok=True)
 
-        if self.options.source:
+        if glob('{}/**/*.go'.format(self.sourcedir), recursive=True):
```
kyrofa: Although... wouldn't having _anything_ in `self.sourcedir` mean that `source` was set? Why not just check to see if `self.sourcedir` is empty?

snapcraft/tests/test_plugin_go.py
```diff
@@ -299,7 +299,7 @@ class Options:
         plugin = go.GoPlugin('test-part', Options(), self.project_options)
 
         os.makedirs(plugin.options.source)
-        os.makedirs(plugin.sourcedir)
+        open(os.path.join(plugin.options.source, 'main.go'), 'w').close()
```
kyrofa: Should this be `plugin.sourcedir` instead of `plugin.options.source`?

snapcraft/tests/test_project_loader.py
```diff
@@ -282,18 +282,19 @@ def test_config_adds_vcs_packages_to_build_packages(self):
...
+                            s[1], c.parts.build_tools))
 
     def test_config_adds_vcs_packages_to_build_packages_from_types(self):
         scenarios = [
             ('git', 'git'),
             ('hg', 'mercurial'),
             ('mercurial', 'mercurial'),
             ('bzr', 'bzr'),
-            ('tar', 'tar'),
+            ('tar', ''),
```
kyrofa: Is this because tar is on the base image? Can we assume that'll always be the case? What if it's removed? What about other distros?

snapcraft/plugins/go.py
sergiusens: because it may sometimes not be empty and confuse the `go` plugin. Check the webcam-webui demo and see

snapcraft/tests/test_project_loader.py
sergiusens: No, it is because we use the `python` implementation

snapcraft/internal/parser.py
sergiusens: The problem is, the parser doesn't auto-install `build-packages` while snapcraft does and we don't need `repo.check_for_command` so tests for snapcraft require extra mocking I wasn't comfortable with just to `__init__` sources.

.vscode/launch.json
```diff
@@ -0,0 +1,120 @@
+{
```
kyrofa: Did you mean to commit this? Are we missing an ignore?

snapcraft/internal/pluginhandler.py
```diff
@@ -63,12 +64,18 @@ def ubuntu(self):
...
-    def __init__(self, *, plugin_name, part_name, part_properties,
-                 project_options, part_schema):
+    def __init__(self, *, plugin_name, part_name,
+                 part_properties, project_options, part_schema):
+        self._part_properties = self._setup_source(part_properties.copy())
         self.valid = False
         self.code = None
         self.config = {}
         self._name = part_name
+        # The remote parts can have a '/' in them to separate the main project
```
josepht: s/remote// :-)

snapcraft/tests/test_plugin_dump.py
```diff
@@ -114,41 +117,44 @@ def test_dump_symlinks(self):
...
         plugin = DumpPlugin('dump', self.options, self.project_options)
 
-        os.makedirs('src')
-        with open(os.path.join('src', 'file'), 'w') as f:
+        os.makedirs(os.path.join(plugin.builddir, 'src'))
+        with open(os.path.join(plugin.builddir, 'src', 'file'), 'w') as f:
             f.write('foo')
 
-        with open('unsnapped', 'w') as f:
+        with open(os.path.join('unsnapped'), 'w') as f:
```
josepht: Did you leave something out here?  Otherwise no need for `os.path.join()`.

snapcraft/internal/parser.py
kyrofa: Ah, I see, okay.

snapcraft/tests/test_project_loader.py
kyrofa: Ah, of course.

.vscode/launch.json
sergiusens: ah no. In a different PR though ;-)

sergiusens:
Thanks @kyrofa and @josepht I think I've addressed all your comments!
And, and @3v1n0 thank you for your unexpected review, fixed as well!

3v1n0:
@sergiusens ahah, I was just passing-by as I was curious to see how things were changing :-)

PR #926, opened 2016-11-23
sources: add current dir to ignore list if we're iterating on parent

3v1n0:
This avoids to iterate over and over our ignored folders when launching snapcraft from a sub-sub-folder of the sources.

Added integration tests. Rewritten unit tests.

Fixes <a href='http://pad.lv/1644350'>LP: #1644350</a>

### DIFF
--- a/integration_tests/snaps/local-source-subfolders/packaging/snap-package/yes-really-deep/snapcraft.yaml
+++ b/integration_tests/snaps/local-source-subfolders/packaging/snap-package/yes-really-deep/snapcraft.yaml
@@ -1,0 +1,12 @@
+name: lets-dump
+version: '0.1'
+summary: Let's dump parent folders
+description: Dumping up-folder
+
+grade: devel
+confinement: strict
+
+parts:
+  dump-part:
+    plugin: dump
+    source: ../../..
--- a/snapcraft/internal/sources.py
+++ b/snapcraft/internal/sources.py
@@ -529,21 +529,22 @@ class Local(Base):
         current_parent_folder = os.path.dirname(os.getcwd())
 
         def ignore(directory, files):
+            ignored = []
             if directory is source_abspath or \
                directory == current_parent_folder:
-                ignored = copy.copy(common.SNAPCRAFT_FILES)
                 relative_cwd = os.path.basename(os.getcwd())
                 if os.path.join(directory, relative_cwd) == os.getcwd():
                     # Source is a parent of the working directory.
                     # Do not recursively copy it into itself.
                     ignored.append(relative_cwd)
+
+            if directory is source_abspath:
+                ignored.extend(common.SNAPCRAFT_FILES)
                 snaps = glob.glob(os.path.join(directory, '*.snap'))
                 if snaps:
                     snaps = [os.path.basename(s) for s in snaps]
                     ignored += snaps
-                return ignored
-            else:
-                return []
+            return ignored
 
         shutil.copytree(source_abspath, self.source_dir,
                         copy_function=file_utils.link_or_copy, ignore=ignore)
--- a/snapcraft/internal/sources.py
+++ b/snapcraft/internal/sources.py
@@ -530,7 +530,7 @@ class Local(Base):
         source_abspath = os.path.abspath(self.source)
 
         def ignore(directory, files):
-            if directory is source_abspath or \
+            if directory == source_abspath or \
                directory == current_dir:
                 ignored = copy.copy(common.SNAPCRAFT_FILES)
                 snaps = glob.glob(os.path.join(directory, '*.snap'))
--- a/snapcraft/internal/sources.py
+++ b/snapcraft/internal/sources.py
@@ -68,6 +68,7 @@ cases you want to refer to the help text for the specific plugin.
 
 """
 
+import copy
 import glob
 import logging
 import os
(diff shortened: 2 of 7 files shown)


3v1n0:
I've semplified a little the PR, as I think we can just ignore the `SNAPCRAFT_FILES` on the current dir, without caring much about the parent and ignoring all the contents... Users might use other facilities such as `snap` or `stage` to filter unwanted data in case.

If the [other way](https://github.com/3v1n0/snapcraft/blob/2b2eec4cce6ffd3f5a2acadd47c4491501650a0c/snapcraft/internal/sources.py#L528) is preferred, though (it still might have some good points), let me know and I'll revert to that.

sergiusens:
thanks for this, I will review later; If you are wondering why I am not merging some of your branches it is because I want to get #923 out the door and not have to rebase that one as it will have conflicts.

I don't mind rebasing your branches after #923 gets into master though. So there should be no pain for you :wink:

3v1n0:
@sergiusens this still merges cleanly after #923 landing... So all yours ;-)

snapcraft/tests/test_sources.py
```diff
@@ -657,27 +657,51 @@ def test_init_with_source_depth_raises_exception(self):
...
-
-        self.assertTrue(
-            'subdir' not in os.listdir(os.path.join('subdir', 'foo')))
+        for snap_dir in [f for f in common.SNAPCRAFT_FILES if '.' not in f]:
+            cwd = os.getcwd()
+            os.chdir('subdir')
+            source_dir = os.path.join(snap_dir, 'foo')
+            local = sources.Local('..', source_dir)
+            local.pull()
+            os.chdir(cwd)
```
sergiusens: this should be a `self.addCleanup(os.chdir, cwd)` and this should be right after the `os.chdir`.
I still wonder if there's a fixture we could use instead, maybe we can generalize `TempCWD` in `snapcraft.tests.fixture_setup` (if you go down this path, make it a prereq PR as it is useful in itself).

integration_tests/snaps/local-source-subfolders/packaging/snap-package/yes-really-deep/snapcraft.yaml
```diff
@@ -0,0 +1,12 @@
...
+summary: Let's dump parent folders
+description: Dumping up-folder
+
+grade: devel
+confinement: strict
+
+parts:
+  dump-part:
+    plugin: dump
+    source: ../../..
```
sergiusens: this doesn't feel really really deep, one more level will though :blush: 

:smile:

snapcraft/tests/test_sources.py
elopio: This is inside the loop. But addCleanup would run after the test finishes. You can use test scenarios instead of a for inside the test, and then IMO things are clearer.
And I'm not sure if I'm missing something, but tests.TestCase is already using TempCWD, so you never have to worry about leaving the cwd in a different dir, the parent test will clean that up.

sergiusens:
El 30/11/16 a las 18:11, Leo Arias escribió:
>
> *@elopio* commented on this pull request.
>
>
> In snapcraft/tests/test_sources.py 
> <https://github.com/snapcore/snapcraft/pull/926>:
>
> >          snapcraft_files_before_pull = copy.copy(common.SNAPCRAFT_FILES)
>   
>           # Verify that the snapcraft root dir does not get copied into itself.
>           os.makedirs('subdir')
>   
> -        cwd = os.getcwd()
> -        os.chdir('subdir')
> -        local = sources.Local('..', 'foo')
> -        local.pull()
> -        os.chdir(cwd)
> -
> -        self.assertTrue(
> -            'subdir' not in os.listdir(os.path.join('subdir', 'foo')))
> +        for snap_dir in [f for f in common.SNAPCRAFT_FILES if '.' not in f]:
> +            cwd = os.getcwd()
> +            os.chdir('subdir')
> +            source_dir = os.path.join(snap_dir, 'foo')
> +            local = sources.Local('..', source_dir)
> +            local.pull()
> +            os.chdir(cwd)
>
> This is inside the loop. But addCleanup would run after the test 
> finishes. You can use test scenarios instead of a for inside the test, 
> and then IMO things are clearer.
>

Yes, this would be welcome indeed!

> And I'm not sure if I'm missing something, but tests.TestCase is 
> already using TempCWD, so you never have to worry about leaving the 
> cwd in a different dir, the parent test will clean that up.
>

He wants to run from a subdir in that tempdir though ;-)

3v1n0:
>> You can use test scenarios instead of a for inside the test, and then IMO things are clearer.
> Yes, this would be welcome indeed!

Ok, @sergiusens, @elopio. I've done that.

snapcraft/internal/sources.py
```diff
@@ -526,16 +526,13 @@ def pull(self):
         elif os.path.isdir(self.source_dir):
             shutil.rmtree(self.source_dir)
 
+        current_dir = os.getcwd()
         source_abspath = os.path.abspath(self.source)
 
         def ignore(directory, files):
-            if directory is source_abspath:
+            if directory is source_abspath or \
```
sergiusens: Can you change this to `directory == source_abspath`?

3v1n0:
@sergiusens I was sure I already commented this, but... I pushed the change you requested.

sergiusens:
No worries @3v1n0 I just wanted to test run this a bit today

sergiusens:
and ftr I did see your push, I get emails for those too, so as long as the commit message is descriptive I know what it is about :-)

3v1n0:
@sergiusens ah, ok... So I'll stop bothering when this happens... You know, I'm used to launchpad silence in this case ;-)

PR #928, opened 2016-11-24
Add missing dependencies to install_requires

javacruft:
A number of runtime dependencies are missing in install_requires;
add missing deps to ensure that snapcraft can be installed directly
to a virtualenv from its github.com repository.


sergiusens:
The reason we don't add these is because it does funky things on dpkg-buildpackage, if you know of a way to resolve this that would be great

javacruft:
Refreshed to what I think will work well with both travis and for use with dpkg-buildpackage.

javacruft:
I'll get this right in a minute - it appears that progressbar in Ubuntu has been fixed to work with python 3, whereas the pure upstream version in pypi does not.

sergiusens:
Thanks.

Strange that progressbar doesn't work as my setup for snapcraft is

```
pip3 install --user pip wheel setuptools
pip install -r requirements.txt requirements-devel.txt
alias snapcraft=<source>/bin/snapcraft
```

javacruft:
requirements.txt has 'progressbar33' which I suspect is a horrible fork for progressbar for py3 and better...

javacruft:
as Ubuntu has a patched up version we could use a pydist override to switch progressbar33 -> progressbar for the packaging.

sergiusens:
Ah, that settles it, thanks

javacruft:
Something looks a little broken with python3-magic on 16.04

```
Searching for file-magic

Note: Bypassing https://pypi.python.org/simple/file-magic/ (disallowed host; see http://bit.ly/1dg9ijs for details).

Couldn't find index page for 'file-magic' (maybe misspelled?)
Scanning index of all packages (this may take a while)

Note: Bypassing https://pypi.python.org/simple/ (disallowed host; see http://bit.ly/1dg9ijs for details).

No local packages or download links found for file-magic
error: Could not find suitable distribution for Requirement.parse('file-magic')
E: pybuild pybuild:274: test: plugin distutils failed with: exit code=1: python3.5 setup.py test 
dh_auto_test: pybuild --test -i python{version} -p 3.5 --dir . returned exit code 13
```

sergiusens:
taking this offline until further reasearch is done.

PR #929, opened 2016-11-25
Snap snapcraft (based on pylxd branch)

kalikiana:


kyrofa:
We really appreciate the work here, but can we please focus on getting #904 passing its tests and land it before expending effort on building on top of it? I'm going to close this for now, please do reopen once #904 lands and you can rebase.

PR #930, opened 2016-11-28
parser: support remote dependencies

3v1n0:
Ensure that we re-parse the parts that have missing dependencies after we've added all the valid ones to the `master_parts_list`

I would have fixed this in a cleaner way (allowing to include parts of a snapcraft.yaml that are good, and ignoring the bad ones only) but it would have caused a major refactor of the parser which probably is better to avoid.

Fixes [LP: #1645350](https://bugs.launchpad.net/snapcraft/+bug/1645350)

@josepht, please have a look to this.

### DIFF
--- a/snapcraft/internal/parser.py
+++ b/snapcraft/internal/parser.py
@@ -216,7 +216,9 @@ def _process_entry(data):
     return parts_list, after_parts
 
 
-def _process_wiki_entry(entry, master_parts_list):
+def _process_wiki_entry(
+        entry, master_parts_list, master_missing_parts,
+        pending_validation_entries):
     """Add valid wiki entries to the master parts list"""
     # return the number of errors encountered
     try:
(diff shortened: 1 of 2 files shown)


josepht:
The approach looks good to me.  Thanks for working on this.  Please fix the complexity issue and check that the coverage for the files you've modified hasn't decreased and that all new code is covered.  Coverage can be a bit wonky at times. :)

3v1n0:
@josepht 

> Please fix the complexity issue

I'm not sure how to deal with that, should I just split the function? :o

> Check that the coverage for the files you've modified hasn't decreased

This is due to the exception that I protect from in the for-cycle over pending_validation_entries.
The fact is that I don't think is possible that an exception might be triggered there, but wasn't wrong to protect, isn't it?
Otherwise I'd just drop the `try`-`except`.

josepht:
I think you can split out the post processing for the pending entries into its own method.

You can probably trigger that exception by referencing a pending part that has a missing 'parts' in the yaml.  I think we want to keep the `try - except` since _process_wiki_entry() raises a couple of exceptions.

josepht:
Nice job killing two birds with one stone. :-) +1

3v1n0:
@sergiusens completely agree... And in fact so it was "almost" stating my initial comment on PR first post, but things were already so much procedural, so I didn't want to rewrite everything.
But I guess is something that has to be done, sooner or later (but it wasn't something for me this time :-P).

PR #931, opened 2016-11-28
parser: add support for origin-{branch,commit,tag}

3v1n0:
Add support for more sources options, so that you can use multiple branches on remote part repositories.

@josepht

### DIFF
--- a/docs/snapcraft-parts.md
+++ b/docs/snapcraft-parts.md
@@ -186,7 +186,10 @@ required fields:
 
 Optional fields are:
 
- * origin-type - A hint about the type of project that contains the part. (e.g. bzr, git, tar)
+ * origin-type - the type of project that contains the part. (e.g. bzr, git, tar)
+ * origin-branch - the source branch that contains the part.
+ * origin-commit - the source commit that contains the part.
+ * origin-tag - the source tag that contains the part.
 
 ### Note:
 
--- a/integration_tests/__init__.py
+++ b/integration_tests/__init__.py
@@ -80,6 +80,16 @@ class TestCase(testtools.TestCase):
 
         return snapcraft_output
 
+    def run_snapcraft_parser(self, arguments):
+        try:
+            snapcraft_output = subprocess.check_output(
+                [self.snapcraft_parser_command, '-d'] + arguments,
+                stderr=subprocess.STDOUT, universal_newlines=True)
+        except subprocess.CalledProcessError as e:
+            self.addDetail('output', content.text_content(e.output))
+            raise
+        return snapcraft_output
+
     def run_apt_autoremove(self):
         deb_env = os.environ.copy()
         deb_env.update({
--- a/integration_tests/test_parser.py
+++ b/integration_tests/test_parser.py
@@ -15,6 +15,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import os
+import fixtures
 import subprocess
 
 import integration_tests
@@ -24,6 +25,13 @@ from snapcraft.tests import fixture_setup
 
 class ParserTestCase(integration_tests.TestCase):
     """Test bin/snapcraft-parser"""
+
+    def setUp(self):
+        super().setUp()
+
+        tempdir = fixtures.TempDir()
+        self.useFixture(tempdir)
+        self.useFixture(fixtures.EnvironmentVariable('TMPDIR', tempdir.path))
 
     def call_parser(self, wiki_path, expect_valid):
         args = [self.snapcraft_parser_command, '--index', wiki_path,
--- a/integration_tests/test_parser.py
+++ b/integration_tests/test_parser.py
@@ -28,15 +28,13 @@ class ParserTestCase(integration_tests.TestCase):
 
     def setUp(self):
         super().setUp()
+        self.useFixture(fixtures.EnvironmentVariable('TMPDIR', self.path))
 
-        tempdir = fixtures.TempDir()
-        self.useFixture(tempdir)
-        self.useFixture(fixtures.EnvironmentVariable('TMPDIR', tempdir.path))
-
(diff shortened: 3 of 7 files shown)


josepht:
Thank you for adding this support.  +1 from me.

3v1n0:
Updated after rebasing on new master

3v1n0:
@sergiusens well, I thought it was too much, since there are already tests covering this for `sources` in general, but I can look at it.

sergiusens:
El 01/12/16 a las 09:00, Marco Trevisan escribió:
>
> @sergiusens <https://github.com/sergiusens> well, I thought it was too 
> much, since there are already tests covering this for |sources| in 
> general, but I can look at it.
>

Yeah, parser changes like this have bit us in the past which is why I 
would feel more comfortable with an integration test.

3v1n0:
> Yeah, parser changes like this have bit us in the past which is why I would feel more comfortable with an integration test.

@sergiusens Ok, i've updated the branch. Also with some `integration_tests.test_parser` refactor that @elopio would probably like to check. Basically now tests are ran using scenarios and using a new fixture tempdir path, so that if you re-run them they will always use a fresh environment.

integration_tests/test_parser.py
```diff
@@ -15,64 +15,76 @@
...
+        self.useFixture(tempdir)
+        self.useFixture(fixtures.EnvironmentVariable('TMPDIR', tempdir.path))
+        self.tempdir = tempdir.path
 
-        self.assertTrue(os.path.exists('parts.yaml'))
+    def call_parser(self, wiki_path, expect_valid, expect_output=True):
+        part_file = os.path.join(self.tempdir, 'parts.yaml')
+        args = [self.snapcraft_parser_command, '--index', wiki_path,
+                '--debug',
+                '--output', part_file]
```
elopio: I would move this to the parent class, because it should be almost the same as run_snapcraft which is defined there. And I would call it run_snapcraft_parser. In fact, I'm doing it here:
https://github.com/snapcore/snapcraft/pull/917/files#diff-e0600b57bbf57ae4d401975ecad5aca2R83

Your branch will land earlier than mine probably.

integration_tests/test_parser.py
```diff
@@ -15,64 +15,76 @@
...
 
-        args = [self.snapcraft_parser_command, '--index',
-                fixture.fake_parts_wiki_fixture.url,
-                '--debug', '--output', 'parts.yaml']
-        subprocess.check_call(args, stderr=subprocess.DEVNULL,
-                              stdout=subprocess.DEVNULL)
+        tempdir = fixtures.TempDir()
+        self.useFixture(tempdir)
+        self.useFixture(fixtures.EnvironmentVariable('TMPDIR', tempdir.path))
+        self.tempdir = tempdir.path
```
elopio: The cwd set by the parent test class is already a temp dir, and you can access it with self.path. Do we need another dir?

integration_tests/test_parser.py
```diff
@@ -15,64 +15,76 @@
...
+                subprocess.check_call, args, stderr=subprocess.DEVNULL,
+                stdout=subprocess.DEVNULL)
 
-        args = [self.snapcraft_parser_command, '--index',
-                os.path.join(os.path.dirname(__file__), 'hidden_parts_wiki'),
-                '--debug',
-                '--output', 'parts.yaml']
-        subprocess.check_call(args, stderr=subprocess.DEVNULL,
-                              stdout=subprocess.DEVNULL)
+        self.assertEqual(os.path.exists(part_file), expect_output)
```
elopio: I think these tests should check the output content, not just if there exists any output. Of course, that's not something you have to fix here.
But, I wouldn't put the assertion inside the call. What do you think of tests like this:
```
parts_file_path = self.run_snapcraft_parser(...)
self.assertFileContents(parts_file_path, expectedContents)
```

integration_tests/test_parser.py
sergiusens: this is actually fine and for `TMPDIR`

integration_tests/test_parser.py
3v1n0: Thanks, merged (and rebased) that commit into mine.

integration_tests/test_parser.py
3v1n0: Right... Done, thanks. I didn't noticed that.

docs/snapcraft-parts.md
```diff
@@ -187,6 +187,9 @@ required fields:
 Optional fields are:
 
  * origin-type - A hint about the type of project that contains the part. (e.g. bzr, git, tar)
+ * origin-branch - A hint about the source branch that contains the part.
+ * origin-commit - A hint about the source commit that contains the part.
+ * origin-tag - A hint about the source tag that contains the part.
```
elopio: I don't think hint is the right word here. It seems wrong also in orgin-type. I would just remove "A hint about" In all the entries.

integration_tests/__init__.py
```diff
@@ -80,6 +80,16 @@ def run_snapcraft(self, command, project_dir=None, yaml_dir=None):
...
 
+    def run_snapcraft_parser(self, arguments):
+        try:
+            snapcraft_output = subprocess.check_output(
+                [self.snapcraft_parser_command, '-d'] + arguments,
+                stderr=subprocess.STDOUT, universal_newlines=True)
+        except subprocess.CalledProcessError as e:
+            self.addDetail('output', content.text_content(e.output))
+            raise
+        return snapcraft_output
```
elopio: such a nice piece of code! ;)

3v1n0: Ahah, incredible!
I didn't forget to credit you in the commit tho :-D

docs/snapcraft-parts.md
3v1n0: Removed that too.

PR #932, opened 2016-11-28
cli: implement `enable-ci travis --refresh` command

cprov:
This will allow users to refresh credentials bound to expire after 1 year
(clock-ticking bomb) or more likely when there is a SSO password change
(rotation).

Travis 'deploy' phase will fail on expired credentials, so users will know
when it's case to refresh.

We could issue macaroons that never expire (age or passwd-changes), but I guess it ends up being a lot more difficult to handle in case of compromise.

LP: #1646660

### DIFF
--- a/snapcraft/integrations/__init__.py
+++ b/snapcraft/integrations/__init__.py
@@ -47,9 +47,7 @@ def enable_ci(ci_system, refresh_only):
 
     if refresh_only:
         module.refresh()
-        return
-
-    print(module.__doc__)
-
-    if input('Continue (y/N): ') == 'y':
-        module.enable()
+    else:
+        print(module.__doc__)
+        if input('Continue (y/N): ') == 'y':
+            module.enable()
--- a/snapcraft/integrations/__init__.py
+++ b/snapcraft/integrations/__init__.py
@@ -27,7 +27,7 @@ SUPPORTED_CI_SYSTEMS = (
 )
 
 
-def enable_ci(ci_system):
+def enable_ci(ci_system, refresh_only):
     if not ci_system:
         # XXX cprov 20161116: we could possibly auto-detect currently
         # integration systems in master ?
@@ -44,6 +44,10 @@ def enable_ci(ci_system):
 
     module = importlib.import_module(
         'snapcraft.integrations.{}'.format(ci_system))
+
+    if refresh_only:
+        module.refresh()
+        return
 
     print(module.__doc__)
 
--- a/snapcraft/integrations/tests/test_travis.py
+++ b/snapcraft/integrations/tests/test_travis.py
@@ -79,7 +79,8 @@ class TravisPreconditionsTestCase(tests.TestCase):
             self.run_command(self.entry_point)
 
         self.assertEqual([
-            'Travis CLI (`travis version`) is not functional.',
+            'Travis CLI (`travis settings`) is not functional or you are '
+            'not allowed to access this repository settings.',
             'Make sure it works correctly in your system before trying this '
             'command again.',
         ], str(raised.exception).splitlines())
--- a/snapcraft/integrations/tests/test_travis.py
+++ b/snapcraft/integrations/tests/test_travis.py
@@ -52,15 +52,15 @@ class TravisPreconditionsTestCase(tests.TestCase):
         ('refresh', {'entry_point': 'refresh'}),
     ]
 
-    def run_command(self):
-        getattr(travis, self.entry_point)()
+    def run_command(self, cmd):
+        getattr(travis, cmd)()
 
     @mock.patch('subprocess.check_call')
     def test_missing_travis_cli(self, mock_check_call):
         mock_check_call.side_effect = [FileNotFoundError()]
 
         with self.assertRaises(RequiredCommandNotFound) as raised:
-            self.run_command()
+            self.run_command(self.entry_point)
 
         self.assertEqual([
             'Travis CLI (`travis`) is not available.',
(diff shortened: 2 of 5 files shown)


snapcraft/integrations/__init__.py
```diff
@@ -45,6 +45,10 @@ def enable_ci(ci_system):
     module = importlib.import_module(
         'snapcraft.integrations.{}'.format(ci_system))
 
+    if refresh_only:
+        module.refresh()
+        return
```
sergiusens: This is a pet peeve so you can ignore me if you want, but maybe split into some methods or wrap what follows in an `else` clause? I just don't like empty `return` statements

snapcraft/integrations/travis.py
```diff
@@ -53,7 +53,17 @@
         snapcraft && snapcraft push *.snap --release edge"
       on:
         branch: master
+
+The dedicated credentials will be functional for exactly 1 year or until
+a password change on the related Ubuntu One SSO account. In order to
+refresh the project credentials, please run the following command::
+
+    $ snapcraft enable-ci travis --refresh
```
sergiusens: Won't macaroon be refreshed on every `push` and/or `release` as well? Maybe the 1 year since last use?

snapcraft/integrations/travis.py
```diff
@@ -146,12 +100,98 @@ def enable():
...
         'Encrypting authorization for Travis and adjusting project to '
         'automatically decrypt and use it during "after_success".')
-
-    os.makedirs(os.path.dirname(LOCAL_CONFIG_FILENAME), exist_ok=True)
     with tempfile.NamedTemporaryFile(mode='w') as fd:
         store.conf.parser.write(fd)
         fd.flush()
-        _encrypt_config(fd.name)
+        os.makedirs(
+            os.path.dirname(LOCAL_CONFIG_FILENAME), exist_ok=True)
```
sergiusens: why is this in the `NamedTemporaryFile` context, I fail to see the reason for a move.

snapcraft/integrations/travis.py
```diff
@@ -146,12 +100,98 @@ def enable():
...
+
+
+@requires_travis_preconditions()
+def refresh():
+    series = storeapi.constants.DEFAULT_SERIES
+    project_config = load_config()
+    snap_name = project_config.data['name']
+    logger.info(
+        'Refreshing credentials to push and release "{}" snaps '
+        'to edge channel in series {}'.format(snap_name, series))
```
sergiusens: Can you change the `"{}"` and `{}` to `{!r}`. Same for other messages bellow.

snapcraft/integrations/tests/test_travis.py
```diff
@@ -117,18 +114,32 @@ def test_enable_broken_git_repo(self, mock_check_call):
...
+    def setUp(self):
+        super().setUp()
+        self.fake_logger = fixtures.FakeLogger(level=logging.DEBUG)
+        self.useFixture(self.fake_logger)
+        self.fake_terminal = tests.fixture_setup.FakeTerminal()
+        self.useFixture(self.fake_terminal)
+
+    def make_travis_yml(self, content):
+        with open('.travis.yml', 'w') as fd:
+            fd.write(content)
```
elopio: One liner: open('.travis.yml', 'w').close()

elopio:
+1 on the contextmanager change. It looks a lot nicer, thanks.

elopio:
I tried this in somebody else's branch, and it got stuck:
```
git clone https://github.com/magne4000/quassel-webserver
cd quassel-webserver
snapcraft enable-ci travis --refresh
[...]
Login successful.
Encrypting authorization for Travis and adjusting project to automatically decrypt and use it during "after_success".
`travis encrypt-file` failed: 1
detected repository as magne4000/quassel-webserver
resource not found ({"error":"Couldn't find repository"})
```

And I had to kill the process. This doesn't seem to be introduced in this branch. I tried again using my fork of the same project, and it worked as expected. I approve this one here, but can you please check into the issue with other people's branches?

snapcraft/integrations/travis.py
cprov: Because I didn't want to leave the plain credentials behind, where it would be picked up by snapcraft.

snapcraft/integrations/__init__.py
cprov: Sure, fixed in the next rev.

snapcraft/integrations/tests/test_travis.py
```diff
@@ -45,25 +45,22 @@
...
+        getattr(travis, self.entry_point)()
 
     @mock.patch('subprocess.check_call')
-    def test_enable_missing_travis_cli(self, mock_check_call):
+    def test_missing_travis_cli(self, mock_check_call):
         mock_check_call.side_effect = [FileNotFoundError()]
 
         with self.assertRaises(RequiredCommandNotFound) as raised:
-            travis.enable()
+            self.run_command()
```
cprov: fixed in a subsequent commit (sorry, I've missed your review)

cprov:
@elopio I have the same problem with repositories that are not mine, effectively we should not run on the because only the owner (and collaborators) have permission to encrypt data in that context (travis private key for this repo). I will figure out a way to prevent run and instruct users to get perms.

cprov:
@elopio both problems you pointed (python hanging and not identifying lack of permission) solved in https://github.com/snapcore/snapcraft/pull/932/commits/52c53dd9721d328363ca518063b3f5069a16fb83

PR #933, opened 2016-11-29
_filedir takes an extension, not a glob

chipaca:
`_filedir "*.snap"` means only files that match `*.*.snap` get tab-completed.

### DIFF
--- a/debian/snapcraft-completion
+++ b/debian/snapcraft-completion
@@ -19,7 +19,7 @@ _snapcraft()
         return 0
         ;;
     upload | push | sign-build)
-        _filedir '*.snap'
+        _filedir 'snap'
         return 0
         ;;
     enable-ci)


kyrofa:
Indeed, from the docs:

```bash
# This function performs file and directory completion. It's better than
# simply using 'compgen -f', because it honours spaces in filenames.
# @param $1  If `-d', complete only on directories.  Otherwise filter/pick only
#            completions with `.$1' and the uppercase version of it as file
#            extension.
```

Thanks @chipaca!

sergiusens:
I don't see a difference with this change as with the current form I see snaps on tab complete, this is still correct

chipaca:
Heh. Ok, adapting from another similar bug report: given
```console
$ ls
xyzzx.two.snap  xyzzy.snap  xyzzz/
```
with the old completion,
```console
$ snapcraft push xyzz
xyzzx.two.snap  xyzzz/          
```
with this fix,
```console
$ snapcraft push xyzz
xyzzx.two.snap  xyzzy.snap      xyzzz/          
```
ok?

sergiusens:
El 30/11/16 a las 18:39, John Lenton escribió:
>
> Heh. Ok, adapting from another similar bug report: given
>
> $ls
> xyzzx.two.snap xyzzy.snap xyzzz/
>
> with the old completion,
>
> $snapcraft push xyzz
> xyzzx.two.snap xyzzz/
>
> with this fix,
>
> $snapcraft push xyzz
> xyzzx.two.snap xyzzy.snap xyzzz/
>
> ok?
>

Yeah, I go it after sending the request. It was really hidden as all the 
snaps snapcraft generates have a `.` lately due to the version string in 
the filename ;-)

Already merged so all good!

PR #935, opened 2016-11-29
cmake plugin: utilize MakePlugin build logic within CMakePlugin

larryprice:
The CMakePlugin inherits from MakePlugin but ignores all of the interesting customization available in the MakePlugin. This commit breaks some of the build logic for make into a separate method such that it can be called during a cmake build.

Also See: https://bugs.launchpad.net/snapcraft/+bug/1645853


snapcraft/plugins/make.py
```diff
@@ -94,9 +94,7 @@ def __init__(self, name, options, project):
         super().__init__(name, options, project)
         self.build_packages.append('make')
 
-    def build(self):
-        super().build()
-
+    def _build(self, env):
```
kyrofa: If this is meant to be called by subclasses, I suggest you make it part of the plugin's public API.

larryprice: Thanks for the quick response. Do you mean changing the name to something like `make`? I've pushed up a very small commit to update the name and usage. Let me know what else might need changing.

larryprice:
I believe the reason that "coverage decreased" is that I removed a couple lines of code from cmake.py that are now obviated by make.py

snapcraft/plugins/make.py
```diff
@@ -94,9 +94,7 @@ def __init__(self, name, options, project):
         super().__init__(name, options, project)
         self.build_packages.append('make')
 
-    def build(self):
-        super().build()
-
+    def make(self, env):
```
sergiusens: if anything I'd make env optional

```python
def make(self, env=None):
    if not env:
        env = os.environ.copy()
```

snapcraft/plugins/make.py
```diff
@@ -119,4 +117,8 @@ def build(self):
         else:
             install_param = self.options.make_install_var + '=' + \
                 self.installdir
-            self.run(command + ['install', install_param])
+            self.run(command + ['install', install_param], env=env)
+
+    def build(self):
+        super().build()
+        self.make(env=os.environ.copy())
```
sergiusens: and just make this `self.make()`

snapcraft/plugins/make.py
kyrofa: Yeah that'll work, I just meant not beginning with an underscore.

larryprice:
Great idea - I've made env an optional param defaulting to `None` and added a test which passes a non-`None` param to `make`

PR #937, opened 2016-12-01
Incorporate all part properties into state tracking.

kyrofa:
This PR fixes LP: [#1646289](https://bugs.launchpad.net/snapcraft/+bug/1646289) by adding all part properties into state tracking. The state->property map is as follows:

**Pull State**:
  - plugin
  - stage-packages
  - source
  - source-commit
  - source-depth
  - source-tag
  - source-type
  - source-branch
  - source-subdir

**Build State**:
  - after
  - build-packages
  - disable-parallel
  - organize

**Stage State**:
  - stage
  - filesets

**Prime State**:
  - snap

Note that this change will mark old trees dirty due to the new options being included in the state tracking.


sergiusens:
This test is going to be complicated `test_build_old_tree_still_filters`, either we discard and not check for values we haven't found previously or assume it needs rebuilds, it would happen in any case if plugins are extended. With the future work to rebuild automatically by tracking changes, it almost feels like this test needs different expectations.

snapcraft/internal/states/_prime_state.py
```diff
@@ -41,18 +41,18 @@ def __init__(self, files, directories, dependency_paths=None,
             self.dependency_paths = dependency_paths
 
     def properties_of_interest(self, part_properties):
-        """Extract the properties concerning this step from part_properties.
+        '''Extract the properties concerning this step from part_properties.
 
         The only property of interest to the prime step is the `snap` keyword
         used to filter out files with a white or blacklist.
-        """
+        '''
```
sergiusens: Didn't you ask @elopio about this on irc? It was ok as it was `"""` :wink:

snapcraft/tests/test_commands_clean.py
```diff
@@ -25,11 +25,32 @@
 from snapcraft.main import main
 from snapcraft.internal import (
     pluginhandler,
+    project_loader,
     states,
 )
 from snapcraft import tests
 
 
+class _CompareExpectedStates():
```
sergiusens: I think I would rather have this be a method ala `assertStatesEqual(state1, state2)` b extending from the `TestCase` class (call it `StateTestClass` if out of ideas).

sergiusens:
@elopio fwiw, a test run of this would be nice.

snapcraft/internal/states/_prime_state.py
elopio: He never listens ;)

snapcraft/tests/test_pluginhandler.py
```diff
@@ -987,17 +956,16 @@ def setUp(self):
...
-                                                 plugin_name='nil')
+        self.handler = _load_plugin('test_part')
+
         self.handler.makedirs()
 
     def test_mark_done_clears_later_steps(self):
         for index, step in enumerate(common.COMMAND_ORDER):
-            shutil.rmtree(self.parts_dir)
+            with contextlib.suppress(FileNotFoundError):
+                shutil.rmtree(self.parts_dir)
```
elopio: rmtree has an ignore_errors argument.

schema/snapcraft.yaml
```diff
@@ -137,26 +137,28 @@ properties:
...
+          source-tag:
+            type: string
+            default: ''
+          source-type:
+            type: string
+            default: ''
           disable-parallel:
               type: boolean
               default: false
-          stage-packages:
```
elopio: Could you please explain why is this not needed here anymore?

elopio:
Test in progress: https://travis-ci.org/elopio/snapcraft-de-noche/builds/180299760

snapcraft/internal/states/_prime_state.py
kyrofa: Yes I did! And he answered "three double quotes." Thus logic dictates that I replace all triple double quotes with single quotes, apparently.

snapcraft/tests/test_pluginhandler.py
```diff
@@ -1036,7 +1003,16 @@ def test_pull_state(self, ubuntu_mock):
...
+        self.assertEqual(9, len(state.properties))
+        self.assertTrue('source' in state.properties)
+        self.assertTrue('source-branch' in state.properties)
+        self.assertTrue('source-commit' in state.properties)
+        self.assertTrue('source-depth' in state.properties)
+        self.assertTrue('source-subdir' in state.properties)
+        self.assertTrue('source-tag' in state.properties)
+        self.assertTrue('source-type' in state.properties)
+        self.assertTrue('plugin' in state.properties)
+        self.assertTrue('stage-packages' in state.properties)
```
kyrofa: Hahaha, coming from the guy who wanted to outlaw `for`s in tests :grin: .

schema/snapcraft.yaml
kyrofa: It was duplicated, actually-- expand the file below, you'll see it again next to `build-packages`.

kyrofa:
> What is more interesting here though is that this will make every project dirty, with proof of the integration test that currently fails.

Yeah, I don't think we want that. This needs a little more work.

schema/snapcraft.yaml
```diff
@@ -180,8 +182,10 @@ properties:
             default: []
           organize:
             type: object
+            default: {}
```
sergiusens: this makes me think we need to copy the default instead of just assigning it

snapcraft/tests/test_pluginhandler.py
kyrofa: Yeah but that would ignore permissions errors etc. as well, I assume? I'd want to see those errors.

schema/snapcraft.yaml
kyrofa: I think so. I added a `deepcopy` in there.

snapcraft/internal/states/_prime_state.py
kyrofa: Fixed.

snapcraft/tests/test_pluginhandler.py
kyrofa: Done.

kyrofa:
>> What is more interesting here though is that this will make every project dirty, with proof of the integration test that currently fails.
>
> Yeah, I don't think we want that. This needs a little more work.

Alright, that should be solved with the latest commit.

kyrofa:
> Ideally it would be nice to see a test where we explicitly see that pull-properties and build-properties from plugins are respected (either when defined in schema or in the new methods).

Done. I think this is ready for another look, now.

integration_tests/snaps/local-plugin-legacy-build-properties/snapcraft.yaml
```diff
@@ -0,0 +1,12 @@
+name: test-package
+version: 0.1
+summary: one line summary
```
elopio: please explain the purpose of this snap here.

integration_tests/test_state_migration_2_22_1.py
```diff
@@ -0,0 +1,28 @@
...
+
+class StateMigrationTestCase(integration_tests.TestCase):
+
+    def test_all_steps_not_dirty(self):
+        project_dir = '2.22.1-prime-state'
+        output = self.run_snapcraft('pull', project_dir)
+        self.assertRegex(output, "Skipping pull part-name")
+        self.assertRegex(output, "Skipping build part-name")
+        self.assertRegex(output, "Skipping stage part-name")
+        self.assertRegex(output, "Skipping prime part-name")
```
elopio: single quotes :p

snapcraft/internal/lifecycle.py
```diff
@@ -203,12 +204,36 @@ def _create_meta(self, step, part_names):
...
+
+            if dirty_report.dirty_project_options:
+                humanized_options = formatting_utils.humanize_list(
+                    dirty_report.dirty_project_options, 'and')
+                pluralized_connection = formatting_utils.pluralize(
+                    dirty_report.dirty_project_options, 'option appears',
+                    'options appear')
+                message_components.append(
+                    'The {} project {} to have changed.\n'.format(
+                        humanized_options, pluralized_connection))
```
elopio: just a note that this is not nice for translations. It's ok for now because we are not doing translations, but hopefully the tool we choose for our messages will make this nicer. https://docs.python.org/3/library/gettext.html#gettext.ngettext

elopio:
Other than a few small comments and:
integration_tests/snaps/local-plugin-legacy-pull-properties/parts/plugins/x_local_plugin.py:32:1: W391 blank line at end of file

I'm +1.

integration_tests/test_state_migration_2_22_1.py
kyrofa: Argh, done.

snapcraft/internal/lifecycle.py
kyrofa: Ah indeed, once we start using that stuff this can be simplified a bit.

sergiusens: Translations is not in the nearby future, we discussed with the store team as they'd need to internationalize their responses as well.

sergiusens:
That's a lot of new code ;-)

The only problem I have with this approach is if we add a new property, say we did `source-depth: 0` as a default right after this branch; it wouldn't be marked dirty. `source-depth: 0` is not really a relevant option to showcase the problem, but say we add negated logic to a plugin for a new property, it means the built part would never be marked dirty.

kyrofa:
> That's a lot of new code ;-)

Yeah, not marking old trees dirty ended up being a bit of an undertaking.

> The only problem I have with this approach is if we add a new property, say we did source-depth: 0 as a default right after this branch; it wouldn't be marked dirty. source-depth: 0 is not really a relevant option to showcase the problem, but say we add negated logic to a plugin for a new property, it means the built part would never be marked dirty.

Hmm, good point. It really should be using the default values then, huh? I'll give that some thought while I resolve the conflicts here.

PR #938, opened 2016-12-01
store: return specific error when already owned

sergiusens:
On `snapcraft register` of a snap we already own we can just say
it is already owned by using the specific error code returned by
the store for this.

LP: #1602091


sergiusens:
Darn coveralls :-/

sergiusens:
We have #942 which is the same as well

sergiusens:
Merged the on on the branch

PR #939, opened 2016-12-01
Replace coveralls with codecov

elopio:
LP: #1646572


elopio:
This requires to enable the codecov integration, and disable the coveralls integration.
I'm not sure if everybody can see this report, but it looks a lot better than coveralls:
https://codecov.io/gh/elopio/snapcraft/branch/codecov

sergiusens:
thank you!

PR #940, opened 2016-12-01
store: implement delta uploads in push.

squidsoup:
LP: #1646303

A few additional notes:

This branch reverts the hard linking of cache files, as the cached file needs an immutable hash and should not change when a new snap is built.

A new error code, `processing_upload_delta_error` has been introduced, which the SCA notification endpoint will return in the event of a delta reconstitution failure. In the event of this error, or the client fails to generate a delta, snapcraft will fall back to the normal upload path.

The branch also introduces a new dependency, pysha3, which is required for calculating sha3_384 hashes. While this can be slow across large files, a later optimisation would be skipping the source snap hash calculation once snaps are cached by their sha3_384 hash rather than revision. 

As a small optimisation, deltas will not be uploaded and fallback if they are 90% or over the size of the target snap.

Additionally, this branch re-implements snap caching to use sha3-384 hashes exclusively, instead of parsing  snap filenames for revisions which was potentially brittle.


squidsoup:
So, looks like we're in a holding pattern until we have pysha3 packaged for xenial, yackkety and zesty. @thomir and I will look at this today.

snapcraft/_store.py
```diff
@@ -375,39 +379,104 @@ def sign_build(snap_filename, key_name=None, local=False):
...
 
-    result = tracker.track()
+    snap_cache = cache.SnapCache(project_name=snap_name)
+    source_snap = snap_cache.get_latest(snap_name)
+
+    if os.environ.get('DELTA_UPLOADS_EXPERIMENTAL') and source_snap:
+        try:
+            result = _upload_delta(snap_name, snap_filename, source_snap)
+        except StoreDeltaApplicationError:
+            logger.warning('Unable to upload delta, '
```
sergiusens: let's say "Unable to push delta,"

snapcraft/_store.py
```diff
@@ -375,39 +379,104 @@ def sign_build(snap_filename, key_name=None, local=False):
...
+                           'falling back to uploading full snap.')
+            result = _upload_snap(snap_name, snap_filename)
+    else:
+        result = _upload_snap(snap_name, snap_filename)
```
sergiusens: _push_snap please

snapcraft/_store.py
```diff
@@ -375,39 +379,104 @@ def sign_build(snap_filename, key_name=None, local=False):
...
-        snap_cache.cache(snap_filename, result['revision'])
-        snap_cache.prune(keep_revision=result['revision'])
 
     if release_channels:
         release(snap_name, result['revision'], release_channels)
 
 
+def _upload_snap(snap_name, snap_filename):
+    store = storeapi.StoreClient()
+    logger.info('Uploading {}'.format(snap_filename))
```
sergiusens: Pushing

snapcraft/_store.py
```diff
@@ -375,39 +379,104 @@ def sign_build(snap_filename, key_name=None, local=False):
...
+        delta_filename = xdelta_generator.make_delta()
+    except DeltaGenerationError:
+        raise StoreDeltaApplicationError
+
+    snap_hashes = {'source_hash': calculate_sha3_384(source_snap),
+                   'target_hash': calculate_sha3_384(target_snap),
+                   'delta_hash': calculate_sha3_384(delta_filename)}
+
+    try:
+        logger.info('Uploading delta {}'.format(delta_filename))
```
sergiusens: Pushing delta

snapcraft/_store.py
```diff
@@ -375,39 +379,104 @@ def sign_build(snap_filename, key_name=None, local=False):
...
+            delta_tracker = store.upload(
+                snap_name,
+                delta_filename,
+                delta_format=delta_format,
+                source_hash=snap_hashes['source_hash'],
+                target_hash=snap_hashes['target_hash'],
+                delta_hash=snap_hashes['delta_hash'])
+        result = delta_tracker.track()
+        delta_tracker.raise_for_code()
+    except storeapi.errors.StoreReviewError as e:
```
sergiusens: we've been putting this logic in `storeapi.errors`, what are the gains of going this path?

snapcraft/internal/cache/_snap.py
```diff
@@ -36,7 +36,7 @@ def _setup_snap_cache(self):
         return snap_cache_path
 
     def cache(self, snap_filename, revision):
-        """Cache snap revision in XDG cache.
+        """Cache snap revision in XDG cache, unless exists.
```
sergiusens: I may have a problem parsing this, "unless *it* exists *already*". ?

snapcraft/_store.py
squidsoup: I wanted to make it clear that a specific failure in the delta infrastructure will result in a fallback to the normal path. I'd be concerned that moving that into `errors` would potentially make that less obvious, but certainly open to refactoring.

squidsoup:
All green now, python3-pysha3 has landed in xenial universe!

snapcraft/_store.py
```diff
@@ -413,37 +427,100 @@ def push(snap_filename, release_channels=None):
     snap_name = snap_yaml['name']
     store = storeapi.StoreClient()
 
-    logger.info('Pushing {!r} to the store.'.format(snap_filename))
+    logger.info('Pushing {!r} to the store'.format(snap_filename))
```
sergiusens: what happened to the `.`?

snapcraft/internal/cache/_snap.py
```diff
@@ -45,7 +45,8 @@ def cache(self, snap_filename, revision):
             revision)
         cached_snap_path = os.path.join(self.snap_cache_dir, cached_snap)
         try:
-            link_or_copy(snap_filename, cached_snap_path)
+            if not os.path.isfile(cached_snap_path):
+                shutil.copyfile(snap_filename, cached_snap_path)
```
sergiusens: Please add a code comment on why copyfile so the next person coming in doesn't "optimize" this :wink:

snapcraft/_store.py
```diff
@@ -550,6 +619,13 @@ def history(snap_name, series, arch):
     print(tabulated_revisions)
 
 
+def get_latest_revision(snap_name, series=None, arch=None):
+    store = storeapi.StoreClient()
+    with _requires_login():
+        history = store.get_snap_history(snap_name, series, arch)
+    return history[-1]
```
sergiusens: this seems like unused code.

sergiusens: it will be useful once we download something to cache before doing a delta or I might take this for the "download and unpack the core snap" work to make classic buildable. But it should leave this PR as it will be forgotten :-)

snapcraft/_store.py
```diff
@@ -413,37 +427,100 @@ def push(snap_filename, release_channels=None):
...
+            except OSError:
+                logger.warning(
+                    'Unable to remove delta {}'.format(delta_filename))
+    return result
+
+
 def _get_text_for_opened_channels(opened_channels):
     if len(opened_channels) == 1:
-        return 'The {!r} channel is now open.'.format(opened_channels[0])
+        return 'The {!r} channel is now open'.format(opened_channels[0])
```
sergiusens: why did we lose the `.`?

snapcraft/_store.py
```diff
@@ -413,37 +427,100 @@ def push(snap_filename, release_channels=None):
...
     else:
         channels = ('{!r}'.format(channel) for channel in opened_channels[:-1])
-        return 'The {} and {!r} channels are now open.'.format(
+        return 'The {} and {!r} channels are now open'.format(
```
sergiusens: why did we lose the .?

snapcraft/_store.py
squidsoup: Yep, fair enough, thanks.

snapcraft/_store.py
squidsoup: While working on this branch iirc, I noticed some inconsistencies in user facing log messages (some have fullstops, others do not). Probably best to discuss and consider for another branch, I'll just revert these.

snapcraft/_store.py
```diff
@@ -417,27 +431,90 @@ def push(snap_filename, release_channels=None):
...
 
-    with _requires_login():
-        tracker = store.upload(snap_name, snap_filename)
+    snap_cache = cache.SnapCache(project_name=snap_name)
+    source_snap = snap_cache.get_latest(snap_name)
+
+    sha3_384_available = hasattr(hashlib, 'sha3_384')
+
+    if os.environ.get('DELTA_UPLOADS_EXPERIMENTAL') and \
+       sha3_384_available and source_snap:
```
kyrofa: Personally I prefer parentheses to escaped newlines, and that matches the rest of the codebase.

snapcraft/_store.py
```diff
@@ -417,27 +431,90 @@ def push(snap_filename, release_channels=None):
...
 
-    result = tracker.track()
     # This is workaround until LP: #1599875 is solved
     if 'revision' in result:
         logger.info('Revision {!r} of {!r} created.'.format(
-            result['revision'], snap_name))
-    else:
-        logger.info('Uploaded {!r}'.format(snap_name))
-    tracker.raise_for_code()
+            result['revision'], snap_name))\
```
kyrofa: What is this backslash here?

snapcraft/_store.py
```diff
@@ -417,27 +431,90 @@ def push(snap_filename, release_channels=None):
...
+
+    sha3_384_available = hasattr(hashlib, 'sha3_384')
+
+    if os.environ.get('DELTA_UPLOADS_EXPERIMENTAL') and \
+       sha3_384_available and source_snap:
+        try:
+            result = _push_delta(snap_name, snap_filename, source_snap)
+        except (StoreDeltaApplicationError, storeapi.errors.StorePushError):
+            logger.warning('Unable to push delta, '
+                           'falling back to pushing full snap.')
```
kyrofa: Would it be useful to state why, here?

snapcraft/file_utils.py
```diff
@@ -15,17 +15,23 @@
...
+
 
 from snapcraft.internal.errors import (
     RequiredCommandFailure,
     RequiredCommandNotFound,
     RequiredPathDoesNotExist,
 )
 
+if sys.version_info < (3, 6):
+    import sha3  # noqa
```
kyrofa: This doesn't seem to be used here-- `hashlib` is used. Is this necessary?

squidsoup: Yes, this is required to patch hashlib for sha3_384.

snapcraft/_store.py
squidsoup: Typo, thanks for catching that!

snapcraft/_store.py
squidsoup: @kyrofa Yes, afaict catching StorePushError handles that case (anything other than the 202).  We could differentiate between a store error and a delta generation error and tell the user, but I'm not sure how useful to them that would be?

snapcraft/internal/deltas/_deltas.py
```diff
@@ -84,6 +87,15 @@ def _check_delta_gen_tool(self):
         if not file_utils.executable_exists(self.delta_tool_path):
             raise DeltaToolError(delta_tool=self.delta_tool_path)
 
+    def _check_delta_size_constraint(self, delta_path):
+        """Ensure delta is sufficiently smaller than target snap."""
```
kyrofa: Can you explain this a bit? This will cause the full snap to be pushed instead, when it seems the delta would still result in bandwidth savings. 90% seems a bit arbitrary, I'd like to see some comments addressing that.

snapcraft/_store.py
kyrofa: I was just imagining someone coming to us with "Hey, I'm getting 'Unable to push delta.' Why?" We really have nothing to tell them, right?

snapcraft/internal/deltas/_deltas.py
squidsoup: 90% is a bit arbitrary, but it is the threshold that was established for delta downloads. The idea here being that although there is still some bandwidth saved, there is still a cost for delta application in the delta service, and pushing the full snap is _probably_ faster in that case. We could certainly spend more time benchmarking to have more concrete data to support this.

snapcraft/internal/cache/_snap.py
```diff
@@ -45,7 +45,10 @@ def cache(self, snap_filename, revision):
             revision)
         cached_snap_path = os.path.join(self.snap_cache_dir, cached_snap)
         try:
-            link_or_copy(snap_filename, cached_snap_path)
+            if not os.path.isfile(cached_snap_path):
+                # this must not be hard-linked,
+                # hence avoids using fileutils.link_or_copy
+                shutil.copyfile(snap_filename, cached_snap_path)
```
squidsoup: Hardlinking means any rebuilt snap invalidates the snap cached with the same name.

snapcraft/internal/deltas/_deltas.py
kyrofa: That's fair. Mind adding a comment?

snapcraft/internal/cache/_snap.py
```diff
@@ -76,6 +79,16 @@ def prune(self, *, keep_revision):
                         'Unable to purge snap {}.'.format(cached_snap))
         return pruned_files_list
 
+    def get_latest(self, snap_name):
+        """Get most recently cached snap."""
+
+        cached_snaps = [
+            os.path.join(self.snap_cache_dir, f) for f in os.listdir(
+                self.snap_cache_dir) if f.startswith(snap_name)]
```
squidsoup: Sure thing, I can make that a loop.

snapcraft/internal/cache/_snap.py
kyrofa: (Note that I understand why given the PR description, but someone looking at this won't)

sergiusens: your reply to @kyrofa should be the code comment :-)

sergiusens:
This is what I am seeing when playing with this feature:

```
$ DELTA_UPLOADS_EXPERIMENTAL=1 snapcraft push vim_8.0_amd64.snap 
Pushing 'vim_8.0_amd64.snap' to the store.
Found cached source snap /home/sergiusens/.cache/snapcraft/vim/revisions/vim_8.0_amd64_2.snap.
Generating xdelta3 delta for vim_8.0_amd64_2.snap --> vim_8.0_amd64.snap.
Pushing delta /home/sergiusens/vim/vim_8.0_amd64.snap.xdelta3.
Uploading /home/sergiusens/vim/vim_8.0_amd64.snap.xdelta3 [                         ]   0%
Uploading /home/sergiusens/vim/vim_8.0_amd64.snap.xdelta3 [=========================] 100%
Unable to push delta, falling back to pushing full snap.
```

squidsoup:
Hi Sergio, this is not enabled on production yet (waiting for the branch to
land), you'll need to use staging for testing.

squidsoup:
That being said it's just a matter of flipping a feature flag on SCA - easy
enough.

squidsoup:
Have improved the user facing error in the case of a failed delta generation or a failed push to the store, as well as improving comments as per @kyrofa's suggestions. @sergiusens  I'll ask Michael (noodles) to flip the deltas flag in production SCA on Tuesday, but in the meantime please test against staging. Thanks for your help!

snapcraft/internal/cache/_snap.py
```diff
@@ -76,6 +80,20 @@ def prune(self, *, keep_revision):
                         'Unable to purge snap {}.'.format(cached_snap))
         return pruned_files_list
 
+    def get_latest(self, snap_name):
```
sergiusens: can this be `get`, also defined in `_cache` and isn't `project_name` the `snap_name` already?

```yaml
def get(hash=None):
    """Get the revision determined in hash or the latest cached item."""
```

snapcraft/internal/cache/_snap.py
```diff
@@ -76,6 +80,20 @@ def prune(self, *, keep_revision):
...
+        """Get most recently cached snap."""
+        cached_snaps = []
+        for f in os.listdir(self.snap_cache_dir):
+            if f.startswith(snap_name):
+                cached_snaps.append(os.path.join(self.snap_cache_dir, f))
+        # cached_snaps = [
```
sergiusens: I see dead code :-)

snapcraft/internal/cache/_snap.py
sergiusens: if not `def get(snap_name, *, hash=None)` works

sergiusens: s/hash/revision/ for parity.

squidsoup:
@sergiusens Hey sergio, I have re-implemented the snap caching to use sha3 384 hashes, and tested against staging/production. I have a few issues with using an arch specific directory for snap caches however (and this is not implemented currently):

1. Snaps built against different arches with the same revision should have different hashes, and should co-exist happily in the same cache dir without collision.
2. Caching a snap to the appropriate arch specific dir would be easy enough, as we can introspect the snap to determine arch. _Getting_ a cached snap by hash would require iterating through every arch dir, but more problematically, I'm not sure how `cache.get` with no arguments would work without passing arch as an argument?

sergiusens:
Hey, just to reply here:
1. this is true
2. we do need to pass the arch

The reason to do this is to prune per arch.

sergiusens:
This is the `zesty` test that failed:

```
FAIL: test_parse_wiki (test_parser.TestParserWikis)
test_parser.TestParserWikis.test_parse_wiki (Missing .snapcraft.yaml file)
testtools.testresult.real._StringException: output: {{{
Processing origin 'lp:~joetalbott/+junk/testpart-missing'
bzr: ERROR: Invalid http response for https://xmlrpc.launchpad.net/bazaar/: Unable to handle http code 502: Bad Gateway
  File "/usr/bin/snapcraft-parser", line 29, in <module>
    wiki_errors = snapcraft.internal.parser.main()
  File "/usr/lib/python3/dist-packages/snapcraft/internal/parser.py", line 94, in main
    return run(args)
  File "/usr/lib/python3/dist-packages/snapcraft/internal/parser.py", line 352, in run
    data = _process_index(output)
  File "/usr/lib/python3/dist-packages/snapcraft/internal/parser.py", line 320, in _process_index
    pending_validation_entries)
  File "/usr/lib/python3/dist-packages/snapcraft/internal/parser.py", line 284, in _try_process_entry
    pending_validation_entries)
  File "/usr/lib/python3/dist-packages/snapcraft/internal/parser.py", line 261, in _process_wiki_entry
    parts_list, after_parts = _process_entry(data)
  File "/usr/lib/python3/dist-packages/snapcraft/internal/parser.py", line 214, in _process_entry
    handler.pull()
  File "/usr/lib/python3/dist-packages/snapcraft/internal/sources/_bazaar.py", line 55, in pull
    subprocess.check_call(cmd)
  File "/usr/lib/python3.5/subprocess.py", line 271, in check_call
subprocess.CalledProcessError: Command '['bzr', 'branch', 'lp:~joetalbott/+junk/testpart-missing', '/home/ubuntu/.cache/snapcraft-parser/lpjoetalbottjunktestpart-missing']' returned non-zero exit status 3
}}}

  File "/tmp/autopkgtest.Q47zNq/build.8M9/snapcraft/integration_tests/test_parser.py", line 82, in test_parse_wiki
    self.call_parser(wiki_file, self.expect_valid, self.expect_output)
  File "/tmp/autopkgtest.Q47zNq/build.8M9/snapcraft/integration_tests/test_parser.py", line 44, in call_parser
    self.assertEqual(os.path.exists(part_file), expect_output)
  File "/usr/lib/python3/dist-packages/testtools/testcase.py", line 350, in assertEqual
    self.assertThat(observed, matcher, message)
  File "/usr/lib/python3/dist-packages/testtools/testcase.py", line 435, in assertThat
    raise mismatch_error
testtools.matchers._impl.MismatchError: False != True
```

so merging anyways.

PR #941, opened 2016-12-02
sources: support symlinks in deb sources

kyrofa:
This PR fixes LP: [#1634813](https://bugs.launchpad.net/snapcraft/+bug/1634813) by using python3-debian to extract debs instead of python3-apt, whose `extractall` method creates directories when it encounters symlinks.

Note that this does add another dependency-- did I put it everywhere?


.travis.yml
```diff
@@ -14,8 +14,8 @@ env:
...
   matrix:
   - TEST_SUITE=static DEPENDENCIES="apt install -y python3-pip && python3 -m pip install -r requirements-devel.txt"
-  - TEST_SUITE=unit DEPENDENCIES="apt install -y libnacl-dev libsodium-dev libffi-dev libapt-pkg-dev libarchive-dev python3-pip squashfs-tools xdelta3 && python3 -m pip install -r requirements-devel.txt -r requirements.txt && python3 -m pip uninstall -y coverage && apt install -y python3-coverage"
-  - TEST_SUITE=integration TEST_STORE=fake DEPENDENCIES="apt install -y bzr curl git libnacl-dev libsodium-dev libffi-dev libapt-pkg-dev libarchive-dev mercurial python3-pip subversion squashfs-tools sudo snapd xdelta3 && python3 -m pip install -r requirements-devel.txt -r requirements.txt"
+  - TEST_SUITE=unit DEPENDENCIES="apt install -y libnacl-dev libsodium-dev libffi-dev libapt-pkg-dev libarchive-dev python3-pip python3-debian squashfs-tools xdelta3 && python3 -m pip install -r requirements-devel.txt -r requirements.txt && python3 -m pip uninstall -y coverage && apt install -y python3-coverage"
```
sergiusens: is this really needed? Can't it be solved by adding it to `requirements.txt`?

Also missing an entry in `HACKING.md`

kyrofa: Yeah good call, done.

kyrofa:
python-debian is not installing from pip. This needs a little more investigation.

flexiondotorg:
@kyrofa After looking at https://travis-ci.org/snapcore/snapcraft/jobs/180859624 I noticed `chardet` is failing to import. If `python-debian` is pip installed into a clean virtualenv it does not require or install `chardet`, even though it clearly uses it. I suggest adding `chardet` to one of the requirements files.

In contrast `python3-debian` Depends: `python3-chardet`.

kyrofa:
> If python-debian is pip installed into a clean virtualenv it does not require or install chardet, even though it clearly uses it. I suggest adding chardet to one of the requirements files.

Indeed, it turns out it's a [known bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=838695). Barry's on it. I've added chardet to `requirements.txt` until that fix is available.

PR #947, opened 2016-12-05
Add 'aliases' support to 'apps'

josepht:
LP: #1646905


sergiusens:
Thanks for the code but mind fixing the tests :smile:

josepht:
@sergiusens I'm working on the tests.  I just pushed the PR to get @kyrofa's help with an issue I was having.

snapcraft/internal/errors.py
```diff
@@ -39,6 +39,13 @@ class MissingState(Exception):
     pass
 
 
+class DuplicateAliasError(SnapcraftError):
+
+    fmt = (
+        'Multiple parts have the same alias defined: {alias!r}'
```
sergiusens: this could be on a single line, right?

snapcraft/internal/project_loader.py
```diff
@@ -131,6 +132,15 @@ def __init__(self, project_options=None):
         snapcraft_yaml = self._expand_filesets(snapcraft_yaml)
         self.data = self._expand_env(snapcraft_yaml)
 
+        # Prevent multiple parts from having duplicate alias names
+        aliases = []
+        for app in self.data.get('apps', []):
+            app = self.data['apps'][app]
+            for alias in app.get('aliases', []):
```
sergiusens: maybe something like this might be easier
```ipython
In [1]: l = ['1', '1', '2', '3']

In [2]: m = set(l)

In [3]: len(m)
Out[3]: 3

In [4]: len(l)
Out[4]: 4

In [5]: len(m) != len(l)
Out[5]: True
```

runtests.sh
```diff
@@ -81,7 +87,7 @@ if [ ! -z "$RUN_STATIC" ] ; then
 fi
 
 if [ ! -z "$RUN_UNIT" ]; then
-    run_unit_tests
+    run_unit_tests "$@"
```
kyrofa: Ah, handy, thanks!

schema/snapcraft.yaml
```diff
@@ -123,6 +123,11 @@ properties:
             uniqueItems: true
             items:
               type: string
+          aliases:
+            type: array
+            items:
+              type: string
+              pattern: "^[a-zA-Z0-9][-_.a-zA-Z0-9]*$"
```
kyrofa: You're using the same regex as that used in snapd, but why are aliases validated differently than apps (e.g. why can aliases end in a hyphen but apps can't)? I've reached out to snapd to see why they're different.

snapcraft/_schema.py
```diff
@@ -73,7 +73,12 @@ def validate(self):
             messages = [e.message]
             path = []
             while e.absolute_path:
-                path.append(e.absolute_path.popleft())
+                element = e.absolute_path.popleft()
+                # assume numbers are indices and use 'xxx[123]' notation.
+                if int == type(element):
```
kyrofa: Would `isinstance(element, int)` be more pythonic here?

snapcraft/_schema.py
```diff
@@ -73,7 +73,12 @@ def validate(self):
             messages = [e.message]
             path = []
             while e.absolute_path:
-                path.append(e.absolute_path.popleft())
+                element = e.absolute_path.popleft()
+                # assume numbers are indices and use 'xxx[123]' notation.
```
kyrofa: Thanks for the comment :smile: .

kyrofa:
Other than sergio's comments, this looks mostly good. The differing regex bothers me, so I'd like to get a solid answer on that before landing this.

snapcraft/internal/project_loader.py
josepht: Shouldn't we show the user the alias that is duplicated?  I initially thought of doing something along these lines but opted for this approach to be able to show the duplicated alias.  I'm happy to change this if you think that not showing the duplicate alias is okay.

schema/snapcraft.yaml
kyrofa: Alright, Gustavo got back to me about why these are different (wanting to support lower- and upper- case, numbers, underscores, periods, etc. in aliases) but he did say that they shouldn't end in a hyphen, underscore, or period.

pedronis: @kyrofa that's ok, whatever we do we need to port it back to snapd, I think we want to allow one letter aliases though (not very common but there is stuff like /usr/bin/X that would be a precedent), here in lot of cases we are trying to accommodate a not too crazy set of upstream already chosen command names,  that's why I didn't try to restrict a lot in the first place

kyrofa: Alright, I say this is fine here since it's currently what snapd supports-- no need to hold up this PR. The validation tools are probably already written, too. If we change the regex we'll have to do it in a coordinated manner.

snapcraft/internal/project_loader.py
kyrofa: +1 to printing the duplicated aliases.

schema/snapcraft.yaml
sergiusens: you should add `uniqueItems: true`

snapcraft/internal/errors.py
```diff
@@ -39,6 +39,11 @@ class MissingState(Exception):
     pass
 
 
+class DuplicateAliasError(SnapcraftError):
```
sergiusens: This code shouldn't be needed thanks to `uniqueItems: true`

snapcraft/internal/project_loader.py
sergiusens: I don't think we need this if we do `uniqueItems: true` now

josepht:
This is doing duplicate checking across all parts not just a single part
which I think your suggestion addresses.

snapcraft/internal/project_loader.py
```diff
@@ -133,13 +134,16 @@ def __init__(self, project_options=None):
...
-                    raise errors.DuplicateAliasError(alias=alias)
-                aliases.append(alias)
+        aliases = [
+            app.get('aliases', []) for app_name, app in self.data.get(
+                'apps', {}).items()]
+        aliases = list(itertools.chain.from_iterable(aliases))
+        unique_aliases = set(aliases)
+        if len(aliases) != len(unique_aliases):
+            for alias in unique_aliases:
+                aliases.remove(alias)
```
kyrofa: Another annoying comment: This seems to be O(n^2), as you're traversing `unique_aliases` as well as `aliases` (remove() is O(n)).

```python
seen = set()
duplicated = set()
for alias in aliases:
    if alias in seen:
        duplicated.add(alias)
    else:
        seen.add(alias)
```

Seems like O(n), as `set().add()` is O(1) and `x in set()` is also O(1).

josepht: Good catch!  I've updated to use your suggestion.

snapcraft/internal/errors.py
kyrofa: It still is in order to catch duplicate aliases _between_ apps.

snapcraft/internal/project_loader.py
```diff
@@ -136,12 +136,15 @@ def __init__(self, project_options=None):
...
-
-            raise errors.DuplicateAliasError(aliases=aliases)
+        seen = set()
+        duplicates = set()
+        for alias in aliases:
+            if alias in seen:
+                duplicates.add(alias)
+            else:
+                seen.add(alias)
+        if len(duplicates) > 0:
```
kyrofa: According to this book I'm reading, this should just be `if duplicates`.

snapcraft/tests/test_project_loader.py
```diff
@@ -50,6 +51,99 @@ def setUp(self):
...
+parts:
+  part1:
+    plugin: go
+    stage-packages: [fswebcam]
+""")
+        raised = self.assertRaises(
+            DuplicateAliasError,
+            project_loader.Config)
+
+        print("JOE: raised: {}".format(raised))
```
sergiusens: seems wrong :wink:

integration_tests/snaps/alias/snapcraft.yaml
```diff
@@ -0,0 +1,19 @@
...
+confinement: devmode
+
+apps:
+  hello:
+    command: hello.sh
+    aliases: [hi.sh, howdy.sh]
+
+parts:
+  my-part:
+    # See 'snapcraft plugins'
```
sergiusens: Remove this comment please

integration_tests/snaps/alias/snapcraft.yaml
```diff
@@ -0,0 +1,19 @@
...
+grade: devel
+confinement: devmode
+
+apps:
+  hello:
+    command: hello.sh
+    aliases: [hi.sh, howdy.sh]
+
+parts:
+  my-part:
```
sergiusens: Call this aliases

snapcraft/internal/project_loader.py
```diff
@@ -131,6 +132,20 @@ def __init__(self, project_options=None):
         snapcraft_yaml = self._expand_filesets(snapcraft_yaml)
         self.data = self._expand_env(snapcraft_yaml)
 
+        # Prevent multiple parts from having duplicate alias names
```
sergiusens: parts?

snapcraft/internal/project_loader.py
```diff
@@ -131,6 +132,20 @@ def __init__(self, project_options=None):
...
+        seen = set()
+        duplicates = set()
+        for alias in aliases:
+            if alias in seen:
+                duplicates.add(alias)
+            else:
+                seen.add(alias)
+        if duplicates:
+            raise errors.DuplicateAliasError(aliases=duplicates)
+
```
sergiusens: Put this in a method named `_ensure_ no _duplicate_app_aliases`

PR #949, opened 2016-12-06
project: support building classic snaps

sergiusens:
This requires setting up some paths to the core snap.

<a href='https://bugs.launchpad.net/bugs/1647699'>LP: #1647699</a>

### DIFF
--- a/demos/git/snapcraft.yaml
+++ b/demos/git/snapcraft.yaml
@@ -2,17 +2,25 @@ name: git
 version: 2.8.0
 summary: Git is a free and open source distributed version control system.
 description: This example is not really production quality
-confinement: strict
+confinement: classic
 
 apps:
-  server:
-    command: bin/git
+  git:
+    command: env PATH=$PATH:$SNAP/libexec/git-core GIT_TEMPLATE_DIR=$SNAP/share/git-core/templates git
 
 parts:
   git:
-    plugin: make
     source: https://github.com/git/git
     source-type: git
-    make-parameters:
-      - prefix=
-    build-packages: [gettext, libssl-dev, libcurl4-openssl-dev, libexpat1-dev]
+    source-depth: 1
+    plugin: autotools
+    configflags:
+      - --with-curl
+      - --with-expat
+    build-packages:
+        - gettext
+        - libssl-dev
+        - libcurl4-openssl-dev
+        - libexpat1-dev
+    stage-packages:
+        - libcurl3
--- a/snapcraft/_options.py
+++ b/snapcraft/_options.py
@@ -30,7 +30,7 @@ _ARCH_TRANSLATIONS = {
         'cross-compiler-prefix': 'arm-linux-gnueabihf-',
         'cross-build-packages': ['gcc-arm-linux-gnueabihf'],
         'triplet': 'arm-linux-gnueabihf',
-        'core_dynamic_linker': 'lib/ld-linux-armhf.so.3',
+        'core-dynamic-linker': 'lib/ld-linux-armhf.so.3',
     },
     'aarch64': {
         'kernel': 'arm64',
@@ -38,7 +38,7 @@ _ARCH_TRANSLATIONS = {
         'cross-compiler-prefix': 'aarch64-linux-gnu-',
         'cross-build-packages': ['gcc-aarch64-linux-gnu'],
         'triplet': 'aarch64-linux-gnu',
-        'core_dynamic_linker': 'lib/ld-linux-aarch64.so.1',
+        'core-dynamic-linker': 'lib/ld-linux-aarch64.so.1',
     },
     'i686': {
         'kernel': 'x86',
@@ -51,6 +51,7 @@ _ARCH_TRANSLATIONS = {
         'cross-compiler-prefix': 'powerpc64le-linux-gnu-',
         'cross-build-packages': ['gcc-powerpc64le-linux-gnu'],
         'triplet': 'powerpc64le-linux-gnu',
+        'core-dynamic-linker': '/lib64/ld64.so.2',
     },
     'ppc': {
         'kernel': 'powerpc',
@@ -63,7 +64,7 @@ _ARCH_TRANSLATIONS = {
         'kernel': 'x86',
         'deb': 'amd64',
         'triplet': 'x86_64-linux-gnu',
-        'core_dynamic_linker': 'lib64/ld-linux-x86-64.so.2',
+        'core-dynamic-linker': 'lib64/ld-linux-x86-64.so.2',
     },
     's390x': {
         'kernel': 's390x',
@@ -71,6 +72,7 @@ _ARCH_TRANSLATIONS = {
         'cross-compiler-prefix': 's390x-linux-gnu-',
         'cross-build-packages': ['gcc-s390x-linux-gnu'],
         'triplet': 's390x-linux-gnu',
+        'core-dynamic-linker': '/lib/ld64.so.1',
     }
 }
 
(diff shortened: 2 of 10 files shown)


sergiusens:
@zyga done

demos/git/snapcraft.yaml
```diff
@@ -2,17 +2,25 @@ name: git
...
 apps:
-  server:
-    command: bin/git
+  git:
+    command: env PATH=$PATH:$SNAP/libexec/git-core GIT_TEMPLATE_DIR=$SNAP/share/git-core/templates git
 
 parts:
   git:
-    plugin: make
     source: https://github.com/git/git
```
kyrofa: If we used `https://github.com/git/git.git` here instead the `source-type: git` could be removed.

snapcraft/internal/meta.py
```diff
@@ -155,7 +155,9 @@ def _write_wrap_exe(self, wrapexec, wrappath,
         args = ' '.join(args) + ' "$@"' if args else '"$@"'
         cwd = 'cd {}'.format(cwd) if cwd else ''
 
-        assembled_env = common.assemble_env().replace(self._snap_dir, '$SNAP')
+        include_library_paths = self._config_data['confinement'] != 'classic'
```
kyrofa: A comment for why we don't want LD_LIBRARY_PATHS in this case would be helpful.

snapcraft/internal/project_loader.py
```diff
@@ -299,10 +303,34 @@ def _build_env(root, arch_triplet):
         for envvar in ['CPPFLAGS', 'CFLAGS', 'CXXFLAGS']:
             env.append(formatting_utils.format_path_variable(
                 envvar, paths, prepend='-I', separator=' '))
+
+    if confinement == 'classic':
+        if not core_linker:
+            raise EnvironmentError('classic confinement requires the '
+                                   'core_linker to be set')
+
+        core_path = os.path.join('/snap', 'core', 'current')
```
kyrofa: Note that this is still `ubuntu-core` in the most recent snapd available in xenial (2.16ubuntu3). We probably want to support both unless we want a hard dependency on a newer snapd version.

In a similar vein, have we considered how we might we support other cores? I assume they'd be named differently (thus mounted in a different location).

snapcraft/internal/project_loader.py
```diff
@@ -299,10 +303,34 @@ def _build_env(root, arch_triplet):
...
+        snap_rpaths = common.get_library_paths(snap_path, arch_triplet,
+                                               existing_only=False)
+
+        rpaths = formatting_utils.combine_paths(
+            core_rpaths + snap_rpaths, prepend='', separator=':')
+        env.append('LDFLAGS="$LDFLAGS '
+                   # Building tools to continue the build becomes problematic
+                   # with nodefaultlib.
+                   '-Wl,-z,nodefaultlib '
+                   '-Wl,--enable-new-dtags '
```
kyrofa: I'm curious why this is used if we also remove LD_LIBRARY_PATHs from our generated environment?

sergiusens:
El 6 dic. 2016 7:13 PM, "Kyle Fazzari" <notifications@github.com> escribió:

*@kyrofa* requested changes on this pull request.

Looks mostly good, though I have a few comments/questions.
------------------------------

In snapcraft/_options.py
<https://github.com/snapcore/snapcraft/pull/949#pullrequestreview-11688906>:

> @@ -61,6 +61,7 @@
         'kernel': 'x86',
         'deb': 'amd64',
         'triplet': 'x86_64-linux-gnu',
+        'core_linker': '/lib/{triplet}/ld-linux-x86-64.so.2'

I don't quite understand this. The code below shows that the fallback is
/lib/ld-linux.so.2, but for example libc6 on arm64 doesn't provide that
either (it's /lib/ld-linux-aarch64.so.1). Do we need to make sure we
specify this for all supported architectures?


Yes we do. I just needed someone to hunt those down for me 😁

------------------------------

In snapcraft/internal/meta.py
<https://github.com/snapcore/snapcraft/pull/949#pullrequestreview-11688906>:

> @@ -155,7 +155,9 @@ def _write_wrap_exe(self, wrapexec, wrappath,
         args = ' '.join(args) + ' "$@"' if args else '"$@"'
         cwd = 'cd {}'.format(cwd) if cwd else ''

-        assembled_env = common.assemble_env().replace(self._snap_dir, '$SNAP')
+        include_library_paths = self._config_data['confinement'] != 'classic'

A comment for why we don't want LD_LIBRARY_PATHS in this case would be
helpful.


It wouldn't be used and we don't want them by mistake either. But I'll add
a comment.

------------------------------

In snapcraft/internal/project_loader.py
<https://github.com/snapcore/snapcraft/pull/949#pullrequestreview-11688906>:

> @@ -299,10 +303,34 @@ def _build_env(root, arch_triplet):
         for envvar in ['CPPFLAGS', 'CFLAGS', 'CXXFLAGS']:
             env.append(formatting_utils.format_path_variable(
                 envvar, paths, prepend='-I', separator=' '))
+
+    if confinement == 'classic':
+        if not core_linker:
+            raise EnvironmentError('classic confinement requires the '
+                                   'core_linker to be set')
+
+        core_path = os.path.join('/snap', 'core', 'current')

Note that this is still ubuntu-core in the most recent snapd available in
xenial (2.16ubuntu3). We probably want to support both unless we want a
hard dependency on a newer snapd version.

In a similar vein, have we considered how we might we support other cores?
I assume they'd be named differently (thus mounted in a different location).


We only want to support core for now. The snappy core team will deal with
the transition.

------------------------------

In snapcraft/internal/project_loader.py
<https://github.com/snapcore/snapcraft/pull/949#pullrequestreview-11688906>:

> +                                   'core_linker to be set')
+
+        core_path = os.path.join('/snap', 'core', 'current')
+        core_rpaths = common.get_library_paths(core_path, arch_triplet,
+                                               existing_only=False)
+        snap_path = os.path.join('/snap', snap_name, 'current')
+        snap_rpaths = common.get_library_paths(snap_path, arch_triplet,
+                                               existing_only=False)
+
+        rpaths = formatting_utils.combine_paths(
+            core_rpaths + snap_rpaths, prepend='', separator=':')
+        env.append('LDFLAGS="$LDFLAGS '
+                   # Building tools to continue the build becomes problematic
+                   # with nodefaultlib.
+                   '-Wl,-z,nodefaultlib '
+                   '-Wl,--enable-new-dtags '

I'm curious why this is used if we also remove LD_LIBRARY_PATHs from our
generated environment?


With nodefaultlib we basically disable LD_LIBRARY_PATH so DT_RUNTIMEPATH is
the only way to find shared libraries. We also force the linker to be the
one in core. The whole idea of classic is to resolve the dependencies with
core and the snap but also be able to run and see stuff on the classic
system.

sergiusens:
hm, replying by email isn't as nice :-/

demos/git/snapcraft.yaml
sergiusens: I say this particular on is fine as is, unless you really really want the change

snapcraft/_options.py
```diff
@@ -61,6 +61,7 @@
         'kernel': 'x86',
         'deb': 'amd64',
         'triplet': 'x86_64-linux-gnu',
+        'core_linker': '/lib/{triplet}/ld-linux-x86-64.so.2'
```
sergiusens: If it breaks on some architectures we can always fix it

snapcraft/internal/meta.py
sergiusens: done

snapcraft/internal/project_loader.py
sergiusens: Just `core` for now

snapcraft/internal/project_loader.py
sergiusens: I explained in the email reply ;-)

demos/git/snapcraft.yaml
kyrofa: Nah, no strong feelings.

kyrofa:
> hm, replying by email isn't as nice :-/

Yeah I noticed that, too :stuck_out_tongue: .

PR #951, opened 2016-12-07
snapcraft plugins -> snapcraft list-plugins

tsdgeos:
snapcraft plugins does not exist so i guess snapcraft list-plugins is what we really want there?

### DIFF
--- a/snapcraft/internal/lifecycle.py
+++ b/snapcraft/internal/lifecycle.py
@@ -56,7 +56,7 @@ confinement: devmode # use 'strict' once you have the right plugs and slots
 
 parts:
   my-part:
-    # See 'snapcraft plugins'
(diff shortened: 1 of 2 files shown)


sergiusens:
Thanks for looking into this, we should actually support the plugins alias instead (this was written by Product Strategy so we should follow that)

If still interested, all this involves is something like:
https://github.com/snapcore/snapcraft/pull/890/files#diff-07ab19acf782b5f8bd05c71b8d04f963R99

then adding to:
https://github.com/snapcore/snapcraft/pull/890/files#diff-07ab19acf782b5f8bd05c71b8d04f963R256

and a command test in snapcraft/tests/test_commands_list_plugins.py maybe just using `testscenarios` as in the `list-registered` one.

This bug exists for this PR -> LP: #1644678

PR #953, opened 2016-12-10
sources: refactor base sources into new package.

kyrofa:
This PR makes progress on LP: [#1648906](https://bugs.launchpad.net/snapcraft/+bug/1648906) by moving `snapcraft/internal/sources.py` into the `sources` package, and extracting `Base` and `FileBase` into their own module. This is the first in a series of PRs; subsequent onces will refactor the other sources into their own modules.


kyrofa:
@elopio can you please re-trigger xenial autopkg tests here?

sergiusens:
done @kyrofa

elopio:
retried.

snapcraft/internal/new_sources/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from snapcraft.internal.new_sources._base import (  # noqa
```
sergiusens: this should be relative,

```python
from ._base import (
```

setup.py
```diff
@@ -42,6 +42,7 @@
               'snapcraft.internal',
               'snapcraft.internal.cache',
               'snapcraft.internal.deltas',
+              'snapcraft.internal.new_sources',
```
sergiusens: still call this `sources`, I don't see a reason for `new_sources`. The migration could be transparent and if it isn't it doesn't matter *as much* as it lives in `internal`

snapcraft/internal/new_sources/__init__.py
kyrofa: Done.

setup.py
kyrofa: :+1: , I didn't realize GitHub would follow renames without dumping a bunch of red and green. This is a much better path than the intermediate package.

PR #954, opened 2016-12-12
pluginhandler: convert to package

sergiusens:
LP: #1649301


setup.py
```diff
@@ -42,6 +42,7 @@
               'snapcraft.internal',
               'snapcraft.internal.cache',
               'snapcraft.internal.deltas',
+              'snapcraft.internal.pluginhandler',
```
kyrofa: "plugin handler" was a good class name, but are we sure we want to continue using it for the package?

sergiusens:
@kyrofa a package is just a better way to organize a module so it seems correct. This module changed names a couple of times already and we always come back to this one.

PR #955, opened 2016-12-12
sources: convert to package.

kyrofa:
This PR makes progress on LP: [#1648906](https://bugs.launchpad.net/snapcraft/+bug/1648906) by moving `snapcraft/internal/sources.py` into its own package. Subsequent PRs will further refactor this by extracting each source into its own module.


kyrofa:
> You can also move the tests to a snapcraft.tests.sources as you originally did[...]

I'll just do it piecemeal in the subsequent PRs, if that's alright.

PR #956, opened 2016-12-12
tests: idempotent store installs.

sergiusens:


debian/tests/control
```diff
@@ -13,6 +13,7 @@ Depends: @,
          python-flake8,
          python3-fixtures,
          python3-pexpect,
+         python-requests-unixsocket,
```
kyrofa: I'm assuming this actually needs to be `python3-requests-unixsocket`? From the failures, it also seems to be needed on `snapstests`, below.

PR #957, opened 2016-12-12
sources: refactor base sources into module.

kyrofa:
This PR makes progress on LP: [#1648906](https://bugs.launchpad.net/snapcraft/+bug/1648906) by extracting `Base` and `FileBase` sources into `sources/_base.py`.


snapcraft/internal/sources/_base.py
```diff
@@ -0,0 +1,70 @@
...
+import shutil
+
+import snapcraft.internal.common
+from snapcraft.internal.indicators import (
+    download_requests_stream,
+    download_urllib_source
+)
+
+
+class IncompatibleOptionsError(Exception):
```
sergiusens: can you put this in `snapcraft.sources.errors` and make it inherit from the base error class we have in `snapcraft.errors` (notice no `_`)

sergiusens: This would require defining a `fmt` class attribute as well.

snapcraft/tests/sources/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import unittest.mock
```
sergiusens: take the opportunity to make this, `from unittest.mock import patch`

snapcraft/tests/sources/test_base.py
```diff
@@ -0,0 +1,100 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+import unittest.mock
```
sergiusens: ditto

snapcraft/internal/sources/__init__.py
```diff
@@ -87,9 +86,11 @@
 
 from snapcraft.internal import common
 from snapcraft import file_utils
-from snapcraft.internal.indicators import (
-    download_requests_stream,
-    download_urllib_source
+
+from ._base import (  # noqa
+    Base,
```
sergiusens: does this need to be exported at all?

snapcraft/internal/sources/__init__.py
```diff
@@ -87,9 +86,11 @@
...
+    FileBase,
```
sergiusens: does this need to be exported at all?

snapcraft/internal/sources/__init__.py
```diff
@@ -87,9 +86,11 @@
...
+    IncompatibleOptionsError
```
sergiusens: instead `from . import errors`

snapcraft/internal/sources/__init__.py
kyrofa: It was previously so I continued doing so, but no-- technically not. I've replaced this with an import of _base.

snapcraft/internal/sources/__init__.py
kyrofa: It was previously so I continued doing so, but no-- technically not. I've replaced this with an import of _base.

snapcraft/internal/sources/_base.py
kyrofa: Done.

snapcraft/tests/sources/test_base.py
kyrofa: Done.

snapcraft/internal/sources/__init__.py
kyrofa: Done.

snapcraft/internal/sources/__init__.py
```diff
@@ -87,10 +86,8 @@
 
 from snapcraft.internal import common
 from snapcraft import file_utils
-from snapcraft.internal.indicators import (
-    download_requests_stream,
-    download_urllib_source
-)
+from . import errors
+from . import _base
```
sergiusens: I say don't even import base, it is not needed. Your internal package modules should import using relative paths

snapcraft/internal/sources/__init__.py
```diff
@@ -169,20 +121,20 @@ def download(self):
         os.chmod(self.file, st.st_mode | stat.S_IEXEC)
 
 
-class Bazaar(Base):
+class Bazaar(_base.Base):
```
sergiusens: ok, I see now, add a TODO to remove that `from . import _base` :wink:

kyrofa: Indeed, though a TODO isn't necessary-- once it's not needed, it'll be a static failure.

snapcraft/internal/sources/_base.py
```diff
@@ -0,0 +1,64 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2016 Canonical Ltd
```
elopio: This is older code, so 2015, 2016.

snapcraft/internal/sources/errors.py
```diff
@@ -0,0 +1,25 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from snapcraft.internal import errors
```
elopio: I prefer to keep all our errors in the same module. I would put the ones from sources also in snapcraft.internal.errors.

kyrofa: Indeed, if I want to update an error message specific to sources, I'd like to update that message in sources, not have to go find it elsewhere. Not to mention having a large file of only somewhat-related things is what led us to the mess that was sources.py in the first place :stuck_out_tongue: .

snapcraft/internal/sources/_base.py
kyrofa: Ah indeed, I pretended this was all new. Updated, thanks :smiley: .

PR #958, opened 2016-12-13
Add source name to error message

tsdgeos:
"no handler to manage source (gdsr://src)"
is a better error message than
"no handler to manage source"
since it makes easier to find out where the mistake is


sergiusens:
Please also remember to affect the bug this fixes

sergiusens:
Thank you for your contribution!

PR #959, opened 2016-12-13
cli: make plugins be an alias of list-plugins

tsdgeos:
LP: #1644678


tsdgeos:
> One detail, please update debian/snapcraft-completion

Done

PR #960, opened 2016-12-13
pluginhandler: install scriptlet support

sergiusens:
LP: #1649683

### DIFF
--- a/snapcraft/internal/pluginhandler/__init__.py
+++ b/snapcraft/internal/pluginhandler/__init__.py
@@ -42,6 +42,7 @@ from snapcraft.internal import (
     sources,
     states,
 )
+from ._scriptlets import ScriptRunner
 
 logger = logging.getLogger(__name__)
 
@@ -386,7 +387,10 @@ class PluginHandler:
         shutil.copytree(self.code.sourcedir, self.code.build_basedir,
                         symlinks=True, ignore=ignore)
 
+        script_runner = ScriptRunner(builddir=self.code.build_basedir)
+
         self.code.build()
+        script_runner.install(scriptlet=self._part_properties.get('install'))
 
         self.mark_build_done()
 
--- a/snapcraft/internal/pluginhandler/_scriptlets.py
+++ b/snapcraft/internal/pluginhandler/_scriptlets.py
@@ -1,0 +1,53 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2016 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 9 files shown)


snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -386,7 +387,10 @@ def ignore(directory, files):
         shutil.copytree(self.code.sourcedir, self.code.build_basedir,
                         symlinks=True, ignore=ignore)
 
+        script_runner = ScriptRunner(builddir=self.code.build_basedir)
```
kyrofa: This means the script runner ignores `source-subdir`. Is that on purpose?

kyrofa: This means the script runner ignores `source-subdir`. Is that on purpose?

snapcraft/internal/pluginhandler/_scriptlets.py
```diff
@@ -0,0 +1,52 @@
...
+                f.write('#!/bin/sh\n')
+                f.write(scriptlet)
+                f.flush()
+                scriptlet_path = f.name
+
+            os.chmod(scriptlet_path, 0o755)
+            run([scriptlet_path], cwd=self._builddir)
+        finally:
+            if os.path.exists(scriptlet_path):
+                os.unlink(scriptlet_path)
```
kyrofa: This should probably just attempt the removal while suppressing a `FileNotFoundError`. Though I'm curious why you don't just run from within the context manager and let `delete=True`?

sergiusens: I cannot `chmod` a descriptor in use, so it needs to be closed.

snapcraft/internal/pluginhandler/__init__.py
sergiusens: @kyrofa yes it is, it feels safer to have people type `$whatever-subdir/something` than `../someother-thing`

snapcraft/internal/pluginhandler/_scriptlets.py
kyrofa: Ah, makes sense. Okay, just suppress the exception and I'm happy.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -386,7 +387,10 @@ def ignore(directory, files):
...
+
         self.code.build()
+        script_runner.install(scriptlet=self._part_properties.get('install'))
```
elopio: it's not clear why you initialize script_runner, then do something else, and then use it.
If there's a reason for that, can you explain it in a comment?

elopio:
I think I can find this useful. But it needs docs.

sergiusens:
@elopio as good as it gets doc added

snapcraft/internal/pluginhandler/__init__.py
sergiusens: You will see in the next PR, it is to avoid a code change once `prepare` is in place.
I will propose it as soon as this lands.

eyelash:
I cannot find the documentation for this on http://snapcraft.io/docs. Where is it?

caldav:
@eyelash the deployment there will happen tomorrow, sorry for the delay

PR #971, opened 2016-12-14
tests: fix integration tests in armhf

elopio:
LP: #1649899


elopio:
WIP

sergiusens:
ok, next week task for @kyrofa

elopio:
damn it!

elopio:
Temporary failure in name resolution. I'll retry later.

sergiusens:
El 16/12/16 a las 11:29, Leo Arias escribió:
>
> Temporary failure in name resolution. I'll retry later.
>
The only reason I say later is because this introduces a big slow down 
in our landings and we've had enough of those this week ;-)

elopio:
No, I agree. I would like you to dput as early as possible, to start my holidays today :) I yelled at the elders of the internet, not at you.

sergiusens:
I would avoid usage of `platform.machine` as much as possible and use project_options in the future as this detection is bound to change in the future given that we care about the userspace and not the cpu as proved by the foundations team

elopio:
:dancer:

elopio:
@sergiusens we don't have access to the project options from these tests. Should snapcraft offer a way to retrieve that?

PR #973, opened 2016-12-14
ci: check the license agreement on Travis.

elopio:


elopio:
Some of this checks might not be needed.
For example, if the PR is from a Canonical employee and he's not using his canonical email address, then it's ok to fail. He should probably change his git configuration to send code identifying himself as a canonicaler.

Then this could be reduced to two cases:
- a canonical address: good
- member of the launchpad cla team: good

All the rest could be considered as failures. Let me know what do you think, and I you prefer I'll remove the other two cases.

tools/cla_check_in_travis.sh
```diff
@@ -0,0 +1,28 @@
...
+    # We assume that the members of the snapcore team have signed the agreement.
+    echo 'Branch from a snapcore team member. CLA OK.'
+    exit 0
+elif [ "$(echo $email | cut -d @ -f 2)" = 'canonical.com' ]; then
+    echo 'Branch from a Canonical employee. CLA OK.'
+    exit 0
+elif ./tools/cla_check_in_launchpad.py $email; then
+    echo 'Branch from a signed contributor. CLA OK.'
+    exit 0
+elif ./tools/past_contributor_check.py $owner; then
```
sergiusens: I'd rather not do this last else

elopio: ack. Then I will leave only the two cases mentioned above.

elopio:
@sergiusens do you want me to keep on working on the script to check if it's a member of the canonical team in launchpad, or do you think it's enough to check the email address?

sergiusens:
@elopio if we check for anything we should check that if the submitter is a member of canonical they should submit the PR/commits with the canonical address

tools/cla_check_in_travis.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+set -ev
+
+email="$(git log -1 $TRAVIS_COMMIT --pretty="%aE")"
+echo "Checking the licence agreement for $email..."
```
kyrofa: s/licence/license/

elopio: @kyrofa I copied it from the website. But now I see it's used inconsistently, so I reported a bug:
https://github.com/ubuntudesign/www.ubuntu.com/issues/1200

I think we should use licence, even though it looks ugly.

kyrofa: Ah, british spelling. I thought it was a typo :stuck_out_tongue: .

sergiusens:
Mind fixing the conflicts?

PR #974, opened 2016-12-14
Updated maven plugin to use get_build_properties()

ZenHarbinger:
Updated the maven.py plugin to use the get_build_properties() method instead of modifying the schema.

https://bugs.launchpad.net/snapcraft/+bug/1649986

### DIFF
--- a/snapcraft/tests/test_plugin_maven.py
+++ b/snapcraft/tests/test_plugin_maven.py
@@ -20,6 +20,7 @@ from unittest import mock
 from xml.etree import ElementTree
 
 import fixtures
+from testtools.matchers import HasLength
 
 import snapcraft
 from snapcraft import tests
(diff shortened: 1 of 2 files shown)


sergiusens:
I suspect you will need to update the unit tests too. Thanks for working on this!

ZenHarbinger:
Is the get_build_properties() not called during testing?
I'll work up some stuff in a bit.

sergiusens:
it isn't and it would be good to add a check, but the real problem is that `.schema` is being called and checked against with those two *deprecated* properties that were removed :smile:

ZenHarbinger:
Is there a similar mechanism in place to be able to verify a custom schema w/o calling .schema()?

snapcraft/tests/test_plugin_maven.py
```diff
@@ -64,64 +65,19 @@ def assertSettingsEqual(self, expected, observed):
...
-                         'it was "{}"'.format(maven_targets['minitems']))
-
-        self.assertTrue(
-            'uniqueItems' in maven_targets,
-            'Expected "uniqueItems" to be included in "maven-targets"')
-        self.assertTrue(
-            maven_targets['uniqueItems'],
-            'Expected "maven-targets" "uniqueItems" to be "True"')
-
-        build_properties = schema['build-properties']
```
sergiusens: only this part needed to be removed from the test, the rest should stay.
Thanks.

ZenHarbinger:
Is there something extra to get the xenial- and yakkety- amd64 builds to pass?

sergiusens:
this is fine, it is breaking on some network issues related to the download of node-v4.4.4-linux-x64.tar.gz which we have been seeing lately

snapcraft/tests/test_plugin_maven.py
```diff
@@ -57,6 +57,15 @@ def _canonicalize_settings(settings):
                 default_namespace='http://maven.apache.org/SETTINGS/1.0.0')
             return f.getvalue() + '\n'
 
+    def test_get_build_properties(self):
+        plugin = maven.MavenPlugin('test-part', self.options,
```
sergiusens: a side comment, since this is a `classmethod` you can do
```python
expected_build_properties = ['maven-options', 'maven-targets']
resulting_build_properties = maven.MavenPlugin.get_build_properties()
self.assertThat(resulting_build_properties, HasLength(len(expected_build_properties))
for property in expected_build_properties:
    self.assertIn(property, resulting_build_properties)
```

Mind doing it like this? If yes you will need to add this import right bellow `import fixtures` so it looks like:

```python
import fixtures
from testtools.matchers import HasLength

import snapcraft
```

sergiusens:
By the way, feel free to join https://rocket.ubuntu.com/channel/snapcraft for faster iteration cycles if you want :-)

PR #976, opened 2016-12-14
demos: do not force arch for the tomcat demo

sergiusens:
The `architectures` keyword just forces an architecture to be set on a built snap regardless the contents of the snap. This is effectively useless in this case.

LP: #1591414


elopio:
lgtm

PR #980, opened 2016-12-15
sources: add optional "source-checksum" property

pachulo:
# Features
- Solution for [LP: #1585913](https://bugs.launchpad.net/snapcraft/+bug/1585913) bug based on work done by @tsimonq2 in PR #619 but with support only for checksums specified in the snapcraft.yml file.
- The code adds the optional property **"source-checksum"** to check the integrity of the source files (tar, zip, deb or rpm).
- The format of the property is **\<algorithm\>/\<digest\>**. For example:
  - sha256/de2fb61252548af3c87c4aab17e82601691d19e37fd3d29ea6288e56
- Currently, the supported algorithms are: **md5**, **sha1**, **sha224**, **sha256**, **sha384**,  **sha512**, **sha3_256**, **sha3_384** & **sha3_512**
- The sha3 support comes from [pysha3](https://pypi.python.org/pypi/pysha3) in systems where python version < 3.6.
- ~~The size of **sha2** digests is automatically detected, supporting the following bit sizes:~~
  - ~~224~~
  - ~~256~~
  - ~~384~~
  - ~~512~~

# Failsafes

- It has the following failsafes to ensure the code always works as intended:
  - Special exception for a checksum that doesn't match
  - Raises an exception for a checksum that is invalid, or incompatible with the checksums supported

# Testing coverage

- I have the following testing in place:
  - Integration tests:
    - Testing of each supported algorithm in raw form with a created file (tar) that has a static checksum in the snapcraft.yaml file.
    - A single test in simple-zip.
    - Tests deb-with-checksum and rpm-with-checksum created.
  - Unit tests:
    - Testing for sources that don't allow digest.
    - Testing for sources that allow digest and where the digest:
        - is correct
        - is wrong
        - uses an invalid algorithm


sergiusens:
Thanks for your contribution. The timing is a bit unfortunate in that we just finished refactoring that source package. It is unfortunate to not have sha3 in our current version of python which we might just be sticking with for a while. For this reason one of our colleagues is SRUing pysha3 into xenial. Maybe @squidsoup can give us an update on that.

squidsoup:
pysha3 is currently in the proposed queue for zesty (https://launchpad.net/ubuntu/zesty/+queue?queue_state=0&queue_text=pysha3), after which it will be SRU'd to xenial. Unfortunately I can't provide any timeframes.

tsimonq2:
I would be interested to see what this looks like once the conflicts are resolved.

Thanks for finishing that up, and sorry for going MIA!

pachulo:
So what should I do @sergiusens ? Rebase the PR so it can be merged?
Can we go on without the SHA3 support for now and add it when the library is available in the repos?
I think that this feature is a good one to have in snapcraft as soon as possible.

Thanks a lot to you @tsimonq2 !

kyrofa:
> So what should I do @sergiusens ? Rebase the PR so it can be merged?

I'd say yes.

> Can we go on without the SHA3 support for now and add it when the library is available in the repos?

I'd also say yes to this.

pachulo:
@sergiusens @kyrofa rebase done!

Additionally I:
- Tried to put everything in it's place: the new error handle, the function that verifies the checksum, etc.
- Deleted the code to support sha3 digests (instead of leaving it commented): once the library is added to the repos, it should be a breeze to add it.

Waiting for your comments!

snapcraft/internal/sources/_base.py
```diff
@@ -62,3 +65,35 @@ def download(self):
             request.raise_for_status()
 
             download_requests_stream(request, self.file)
+
+    def verify_checksum(source_checksum, checkfile):
```
kyrofa: Also, is this intended to be a method on `FileBase`? It shouldn't be, and, well, it's not (no `self`), but then it's used as if it is. It should probably be in `sources/__init__.py`.

snapcraft/internal/sources/errors.py
```diff
@@ -23,3 +23,11 @@ class IncompatibleOptionsError(errors.SnapcraftError):
...
     def __init__(self, message):
         super().__init__(message=message)
+
+
+class ChecksumDoesNotMatchError(errors.SnapcraftError):
+
+    fmt = '{message}'
+
+    def __init__(self, message):
+        super().__init__(message=message)
```
kyrofa: This would do more for you if it accepted an expected and actual digest and printed a decent formatted message on its own.

snapcraft/internal/sources/__init__.py
```diff
@@ -33,6 +33,12 @@
     control system or compression algorithm. The source-type key can tell
     snapcraft exactly how to treat that content.
 
+  - source-checksum: <algorithm>/<digest>
+
+    Snapcraft will use the digest specified to verify the integrity of the
+    source. The source-type needs to be a file (tar, zip, deb or rpm) and
+    the algorithm either md5, sha1, sha2 or sha3.
```
pachulo: Yes indeed!

snapcraft/internal/sources/_base.py
```diff
@@ -17,21 +17,24 @@
...
 
 
 class Base:
 
-    def __init__(self, source, source_dir, source_tag=None, source_commit=None,
-                 source_branch=None, source_depth=None,
-                 command=None):
+    def __init__(self, source, source_dir, source_checksum=None,
+                 source_tag=None, source_commit=None, source_branch=None,
+                 source_depth=None, command=None):
```
pachulo: not really, modified!

snapcraft/internal/sources/_base.py
pachulo: No, I didn't really know where to put it, so I put it in `FileBase`, but it makes much more sense in `sources/__init__.py`
Moved and modified the implementation to one similar to your example.

snapcraft/internal/sources/errors.py
pachulo: Yeah, it makes a lot of sense to do in the the error. Modified.

snapcraft/internal/sources/__init__.py
```diff
@@ -170,3 +180,20 @@ def _get_source_type_from_uri(source, ignore_errors=False):
...
 
     return source_type
+
+
+def verify_checksum(source_checksum, checkfile):
+    try:
+        algorithm, digest = source_checksum.split('/', 1)
+
+    except ValueError:
+            raise ValueError('invalid checksum format: {!r}'
```
kyrofa: Indentation is a little weird through here.

snapcraft/internal/sources/errors.py
```diff
@@ -23,3 +23,13 @@ class IncompatibleOptionsError(errors.SnapcraftError):
 
     def __init__(self, message):
         super().__init__(message=message)
+
+
+class DigestDoesNotMatchError(errors.SnapcraftError):
```
kyrofa: The `SnapcraftError` class is a little smarter than that:

```python
class DigestDoesNotMatchError(errors.SnapcraftError):
    fmt = 'Expected the digest for source to be {expected}, but it was {calculated}'

    def __init__(self, expected, calculated):
        super().__init__(expected=expected, calculated=calculated)
```

pachulo: Didn't really know how to do it...thanks for the tip!

snapcraft/internal/sources/__init__.py
pachulo: Modified!

integration_tests/snaps/simple-zip/snapcraft.yaml
```diff
@@ -10,3 +10,39 @@ parts:
     source: simple.zip
     files:
         "*": .
+  checksum-md5:
+    plugin: copy
```
elopio: please use dump instead of copy

integration_tests/snaps/simple-tar/snapcraft.yaml
```diff
@@ -40,3 +40,27 @@ parts:
     plugin: tar-content
     source: simple.tar.bz2
     destination: destdir1/destdir2
+  checksum-md5:
+    plugin: tar-content
```
elopio: please use dump instead of tar-content

snapcraft/internal/sources/_bazaar.py
```diff
@@ -37,6 +37,9 @@ def __init__(self, source, source_dir, source_tag=None, source_commit=None,
             raise errors.IncompatibleOptionsError(
                 'can\'t specify both source-tag and source-commit for '
                 'a bzr source')
+        if source_checksum:
+            raise errors.IncompatibleOptionsError(
+                'can\'t specify a source-checksum for a bzr source')
```
elopio: tiny detail, it's more readable like:
"can't specify a source ..."
It's the only exception where we like double quotes. I know the lines above are not following it, we can fix them later.

snapcraft/internal/sources/errors.py
```diff
@@ -23,3 +23,12 @@ class IncompatibleOptionsError(errors.SnapcraftError):
...
+
+    fmt = 'Expected the digest for source to be {expected}, '\
```
elopio: no need for the \ in there. You can just continue strings in the following line.

integration_tests/snaps/simple-zip/snapcraft.yaml
pachulo: Done!

integration_tests/snaps/simple-tar/snapcraft.yaml
pachulo: Done!

snapcraft/internal/sources/_bazaar.py
pachulo: OK!

snapcraft/internal/sources/errors.py
pachulo: OK!

pachulo:
Modified with the commented stuff, but I will definitely need guidance with those unit tests @elopio !

pachulo:
@elopio I added the unit tests for the sources that don't allow digests, but I would need help to create the ones for the sources that do admit them: Can you give me an example?

snapcraft/internal/sources/errors.py
tsimonq2: I remember writing special exceptions, didn't know we got a new error interface since I initially wrote this, cool!

pachulo:
@elopio I added most of the unit tests for the sources that allow digests in snapcraft/tests/sources/test_checksum.py: I think that this is what you asked for, but I'm not really sure...

Besides that, I would need help to create the unit test for a deb with a correct checksum: if I understood well, I need a real deb file for this one, but I don't really know how to create a dummy one from python.
I've tried to use small_0.1-1.deb (from integration_tests/snaps/simple-deb) but couldn't get the path right...

snapcraft/tests/sources/test_checksum.py
```diff
@@ -0,0 +1,154 @@
...
+import unittest
+import hashlib
+import zipfile
+import libarchive
+
+from snapcraft.internal import sources
+
+from snapcraft import tests
+
+#from snapcraft.tests.sources import SourceTestCase
```
elopio: please remove this commented line.

elopio:
nice work @pachulo! Your tests are correct and nicely coded, but now lets take into account the integration tests so we don't have the same tests twice at different levels.

Please, let me know what do you think about this proposal.

Make a new snap in integration_tests/snaps/checksum-algorithms, and move there all the checksums you have added to simple-tar, and add a test integration_tests/test_checksum_algorithms.py and run snapcraft pull in there. Then simple tar can keep simple, and it will be easier to find the checksum tests.

This integration test will be checking the pull step with all the combinations of algorithms + valid/invalid.
So we could add a valid checksum to simple-deb, simple-zip, simple-rpm, to check the other sources.

This way the integration tests cover the pull step. In the unit tests, we can have a few that directly call verify_checksum. One for correct, one for incorrect, one for invalid algorithm. 

After that, the codecov report will tell us which lines we are missing. I think we will only be missing that when a source is instantiated with source_checksum, the provision step will call verify_checksum. But as that is already covered by the integration tests, we can do use mock.patch to inject a fake instead of the real verification.

As I said above, this is just a proposal. Feel free to disagree or to deviated.

sergiusens:
pysha3 is now in the repos, can you re add the support?

sergiusens:
Can you remove `integration_tests/snaps/checksum-algorithms/icon.png` and its icon entry in `snapcraft.yaml` ?

pachulo:
Hi @elopio !

I think that what you say makes a lot of sense!

I've made the first refactor and I have some comments/doubts:

> This integration test will be checking the pull step with all the combinations of algorithms + valid/invalid.

I think that right now the integration tests only check for algorithm+**valid**, am I wrong?

> In the unit tests, we can have a few that directly call verify_checksum. One for correct, one for incorrect, one for invalid algorithm.

As for now only the test for **invalid algorithm** is working, I'm still working on it...

I'm getting a weird error for the one with the correct checksum:

```
  File "/home/ubuntu/snapcraft/pachulo-snapcraft/snapcraft/tests/sources/test_checksum.py", line 64, in test_correct_checksum
    sources.verify_checksum(source_checksum, 'src/test.zip')
  File "/home/ubuntu/snapcraft/pachulo-snapcraft/snapcraft/internal/sources/__init__.py", line 187, in verify_checksum
    algorithm, digest = source_checksum.split('/', 1)
AttributeError: 'NoneType' object has no attribute 'split'
```

pachulo:
> pysha3 is now in the repos, can you re add the support?

I think that after the changes proposed by @kyrofa a month ago, I will only need to change the documentation after this PR is accepted: https://github.com/snapcore/snapcraft/pull/940

pachulo:
@elopio I think that all tests are working and that they (mostly) adhere to what you proposed!
Can you check my previous comment?

> I think that right now the integration tests only check for algorithm+valid, am I wrong?

If so, how should I create an integration test to check algorithm+invalid?

sergiusens:
@pachulo I am not sure how this has not come up earlier, but we require a CLA to be signed. Have you done so? I ask as the only check that fails is the one to check the CLA

pachulo:
@sergiusens I thought that I did, but I did it again just in case!

sergiusens:
thanks @pachulo I triggered the cla test to rerun. There are however, some conflicts to be taken care of.

snapcraft/tests/sources/test_checksum.py
```diff
@@ -0,0 +1,82 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2016 Canonical Ltd
```
elopio: please update the copyright year.

pachulo:
Good night! I think that now everything is in it's place! I'm waiting for your comments (or even approval)!

integration_tests/test_checksum_algorithms.py
```diff
@@ -0,0 +1,65 @@
...
+                          ['pull', 'checksum-sha3-256'],
+                          project_dir)
+        self.assertRaises(subprocess.CalledProcessError,
+                          self.run_snapcraft,
+                          ['pull', 'checksum-sha3-384'],
+                          project_dir)
+        self.assertRaises(subprocess.CalledProcessError,
+                          self.run_snapcraft,
+                          ['pull', 'checksum-sha3-512'],
+                          project_dir)
```
elopio: This would look better as scenarios, don't you think?

integration_tests/test_checksum_algorithms.py
```diff
@@ -0,0 +1,65 @@
...
+
+import subprocess
+
+import integration_tests
+
+
+class ChecksumAlgorithmsTestCase(integration_tests.TestCase):
+
+    def test_checksum_algorithms(self):
+        project_dir = 'checksum-algorithms'
```
elopio: you could add deb-with-checksum and rpm-with-checksum as scenarios here.

snapcraft/internal/sources/__init__.py
```diff
@@ -84,13 +93,19 @@
 from ._subversion import Subversion  # noqa
 from ._tar import Tar                # noqa
 from ._zip import Zip                # noqa
+from . import errors
+
+if sys.version_info < (3, 6):
```
elopio: why is this here? What happens on 3.6?

snapcraft/tests/sources/test_checksum.py
```diff
@@ -0,0 +1,85 @@
...
+        super().setUp()
+
+    def test_invalid_checksum(self):
+        # Create zip file for testing
+        os.makedirs(os.path.join('src'))
+        file_to_zip = os.path.join('src', 'test.txt')
+        open(file_to_zip, 'w').close()
+        zip_file = zipfile.ZipFile(os.path.join('src', 'test.zip'), 'w')
+        zip_file.write(file_to_zip)
+        zip_file.close()
```
elopio: on this one, there's no need to make a zip file, right?
The error should happen before verify_checksum tries to open the file, so you can just pass a string like 'dummy.zip', and you test the same code path faster.

snapcraft/internal/sources/__init__.py
pachulo: I took the example from [PR#940](https://github.com/snapcore/snapcraft/pull/940) [snapcraft/file_utils.py](https://github.com/snapcore/snapcraft/pull/940/files#diff-8672afd5aee283a8c90cf4487fb51689): In python 3.6 sha3 support is upstreamed in hashlib.

snapcraft/tests/sources/test_checksum.py
pachulo: Yes, we need to make a file. That was my first attempt, but I found out that the error happens **after** verify_checksum opens the file.

integration_tests/test_checksum_algorithms.py
pachulo: I don't really know how scenarios look like, but I guess that you are right. Any hint example to follow?

snapcraft/internal/sources/__init__.py
elopio: got it. Please leave that as a comment in the code.

snapcraft/tests/sources/test_checksum.py
elopio: And does it need to be a valid zip file, or any file will do?

I see that this is not your code, this is the library following the python rule of ask forgiveness, not permission, so it's ok.

integration_tests/test_checksum_algorithms.py
elopio: grep for `scenarios` in the tests directories. You'll find plenty. You need to move this test to it's own test class, and put the parts names in the scenarios list.

Something like:

`scenarios = [(part, {'part': part}) for part in ['checksum-md5', 'checksum-sha1', ...]]`

Then, during the test will be copied for every part, and self.part will have the value.

snapcraft/tests/sources/test_checksum.py
pachulo: Now we only create an empty file.

tsimonq2:
\o/

kyrofa:
Note that the CLA check is timing out (there are too many commits, haha). I think we can ignore it at this point. I'm going to update the branch and trigger the autopkgtests.

PR #981, opened 2016-12-15
Support download and validate on branded stores.

cprov:
Allow users to set `SNAPCRAFT_UBUNTU_STORE` and perform operations
on branded stores.

So far it only affects interactions with the Snap Index (CPI), which
already support the 'X-Ubuntu-Store' header and further permission
checks.

It will be probably used as the foundation for supporting push and
release for branded stores, instead of discrete '--store'.

LP: #1650210

### DIFF
--- a/snapcraft/storeapi/__init__.py
+++ b/snapcraft/storeapi/__init__.py
@@ -354,11 +354,31 @@ class SnapIndexClient(Client):
             'UBUNTU_STORE_SEARCH_ROOT_URL',
             constants.UBUNTU_STORE_SEARCH_ROOT_URL))
 
+    def get_default_headers(self):
+        """Return default headers for CPI requests.
+
+        Tries to build an 'Authorization' header with local credentials
+        if they are available.
+        Also pin specific branded store if `SNAPCRAFT_UBUNTU_STORE`
+        environment is set.
+        """
+        headers = {}
+
+        with contextlib.suppress(errors.InvalidCredentialsError):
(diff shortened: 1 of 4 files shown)


snapcraft/storeapi/__init__.py
```diff
@@ -354,11 +354,31 @@ def __init__(self, conf):
...
+        if they are available.
+        Also pin specific branded store if `SNAPCRAFT_UBUNTU_STORE`
+        environment is set.
+        """
+        headers = {}
+
+        with contextlib.suppress(errors.InvalidCredentialsError):
+            headers['Authorization'] = _macaroon_auth(self.conf)
+
+        branded_store = os.getenv("SNAPCRAFT_UBUNTU_STORE")
```
elopio: single quotes :zipper_mouth_face:

elopio:
@cprov now that we are running unit tests in arm, your assertions need to be more permissive to take into account that the arch might be different.

Also, we don't have integration tests against staging for validate. Roberto said he would add the required ids into staging, but he has not done it. In a similar way, I would like a test branded store that we could use to test this validate change. Is that possible?

cprov:
@elopio fixed the tests to consider the testbed arch, thanks for pointing it.

The lack of staging tests for validation (and others depending on specific snaps or keys to be available) is worrying, we should have come up with a solution for this, either will precious sample-data or some sort of high-level fixture to ensure what we expect is in place in staging.

I will talk to Roberto and Colin and see if we can move this forward somehow, anything is better that not having real integration tests at all.

PR #984, opened 2016-12-15
parser: improve output

josepht:
* Add debug to VCS classes .pull() method.
    * Only show pull output when --debug is given
    * Note the index being processed and the output being written.
    
    LP:#1591015


elopio:
Hum, I'm happy about the output, but now convinced about passing the debug parameter.
In the snapcraft command, we store the debug value in options. Ideally, we should use the same mechanism in both, which I think should be a config of some kind.
If I understood correctly, snapcraft without -d would still do a pull(debug=True), which is confusing.

I reported this, because everything that prints messages is a mess right now: https://bugs.launchpad.net/snapcraft/+bug/1649687
I would prefer to first design that presentation layer that can be used by all our commands. What do you think?

sergiusens:
I agree with @elopio 
Maybe remove the debug from pull for now.

snapcraft/internal/sources/_git.py
```diff
@@ -40,33 +40,39 @@ def __init__(self, source, source_dir, source_tag=None, source_commit=None,
...
+        subprocess.check_call([self.command, '-C', self.source_dir,
+                               'pull', '--recurse-submodules=yes',
+                               self.source, refspec])
+
+        # Merge any updates for the submodules (if any).
+        subprocess.check_call([self.command, '-C', self.source_dir,
+                              'submodule', 'update'])
+
+    def _clone_new(self):
+        command = [self.command, 'clone', '--recursive']
```
sergiusens: what happened to submodules here?

josepht: Isn't that what '--recursive' is?  From the 'git log' history I didn't change that line.

sergiusens:
can you resolve the conflicts here?

snapcraft/internal/sources/_git.py
sergiusens: yeah, I see that now, thanks

PR #986, opened 2016-12-15
parser: clean up help

josepht:
* Add defaults for 'index' and 'output'
* Explain what the parser does

LP:#1591011
LP:#1591012


snapcraft/internal/parser.py
```diff
@@ -18,6 +18,8 @@
 """
 snapcraft-parser
 
+Parse a wiki index of parts and output formated YAML for use by 'snapcraft'
```
elopio: Could you explain how snapcraft uses the formatted YAML in a few words? Or would that be too much info?

snapcraft/internal/parser.py
```diff
@@ -26,8 +28,10 @@
   -v --version                          show program version and exit
   -d --debug                            print debug information while executing
                                         (including backtraces)
-  -i --index=<filename>                 a file containing a part index.
+  -i --index=<filename|URL>             a file containing a part index.
```
elopio: this would be a filepath, right?

snapcraft/internal/parser.py
```diff
@@ -59,7 +63,11 @@ class MissingSnapcraftYAMLError(Exception):
 
 # TODO: make this a temporary directory that get's removed when finished
 BASE_DIR = os.getenv('TMPDIR', '/tmp')
-PARTS_FILE = "snap-parts.yaml"
+PARTS_FILE = 'snap-parts.yaml'
+DEFAULT_INDEX = 'http://wiki.ubuntu.com/snapcraft/parts?action=raw'
+
+__doc__ = __doc__.format(default_index=DEFAULT_INDEX,
+                         default_parts_file=PARTS_FILE)
```
elopio: wow, cool trick. I'm not sure if it's breaking any python rule, but I'm +1! :)

kyrofa: I agree that it's a neat trick, but there's a reason docopt uses the docstring for this-- so the documentation can be available to someone reading the code as _well_ as someone running `-h`. I appreciate the deduplication, but this makes it hard to read for the person in the code.

PR #987, opened 2016-12-15
pluginhandler: prepare scriptlet support

sergiusens:
LP: #1642371


snapcraft/__init__.py
```diff
@@ -225,6 +225,18 @@
     `prime/` directory reflects the file structure of the snap with no
     extraneous content).
 
+  - prepare: shell script
+
+    If present, the shell script defined here is run before the `build` step
+    of the plugin starts. The working directory is the base build
+    directory for the given part. The defined script is run with `/bin/sh`.
```
kyrofa: I just realized: `/bin/sh` will be different for different people, which means the same `snapcraft.yaml` will succeed for some people, and fail for others. Should we instead be using a specific shell?

sergiusens: I don't think we need to worry, all of snapcraft is tied to `dash` now and so are wrapper scripts and everything else

snapcraft/internal/pluginhandler/_scriptlets.py
```diff
@@ -44,6 +44,14 @@ def _run(self, scriptlet):
             with suppress(FileNotFoundError):
                 os.unlink(scriptlet_path)
 
+    def prepare(self, *, scriptlet):
```
sergiusens: I thought about it, maybe...

snapcraft/__init__.py
kyrofa: Alright.

snapcraft/internal/pluginhandler/_scriptlets.py
kyrofa: Or hand it a few different scriptlets in `__init__` and made the `install()` and `prepare()` run them.

snapcraft/tests/pluginhandler/test_scriptlets.py
```diff
@@ -24,6 +24,16 @@
 
 class ScriptletTestCase(tests.TestCase):
 
+    def test_run_prepare_scriptlet(self):
+        handler = mocks.loadplugin(
+            'test-part', part_properties={'prepare': 'touch before-build'})
```
kyrofa: Heh, cool test.

snapcraft/tests/pluginhandler/test_scriptlets.py
```diff
@@ -24,6 +24,16 @@
...
+
+        handler.build()
+
+        before_build_file_path = os.path.join(handler.code.build_basedir,
+                                              'before-build')
+        self.assertThat(before_build_file_path, FileExists())
```
elopio: But this is not testing the order, right?
It would be cool to have a test, maybe integration, that makes sure that prepare runs first, then build, and then install.

snapcraft/tests/pluginhandler/test_scriptlets.py
```diff
@@ -24,6 +24,16 @@
 
 class ScriptletTestCase(tests.TestCase):
 
+    def test_run_prepare_scriptlet(self):
+        handler = mocks.loadplugin(
```
elopio: I know I reviewed the mocks PR, but technically this is a fake: http://xunitpatterns.com/Fake%20Object.html
(testxplaning)

kyrofa: It's not either, really-- it's a real plugin :stuck_out_tongue: .

sergiusens: I put it in `mock` as I had no elegant name to encompass all in there and used `unittest.mock` as precedent where you can create spies which aren't mocks either.

PR #988, opened 2016-12-15
pluginhandler: build scriptlet support

sergiusens:
LP: #1649684


snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -390,7 +390,10 @@ def ignore(directory, files):
         script_runner = ScriptRunner(builddir=self.code.build_basedir)
 
         script_runner.run(scriptlet=self._part_properties.get('prepare'))
-        self.code.build()
+        if self._part_properties.get('build'):
+            script_runner.run(scriptlet=self._part_properties.get('build'))
```
kyrofa: Small nit: this is used twice in immediate succession. Putting it in a variable would be less error prone and perform better.

PR #989, opened 2016-12-15
pluginhandler: add support for disabling system library migration

kyrofa:
Currently snapcraft, during the prime step for a part, checks to see if that part needs any libraries from the system. This is not always desired, as in some cases the libraries will be provided by another snap via the content interface.

This PR fixes LP: [#1639831](http://bugs.launchpad.net/bugs/1639831) by introducing a new property to the base part YAML: `build-attributes`. It's simply a list of build attributes, only one of which exists right now: `no-system-libraries`. For example:
    
    parts:
      # ...
      example:
        plugin: cmake
        build-attributes: [no-system-libraries]
      # ...
    
Using the `no-system-libraries` attribute has the effect of not pulling any libraries from the system during the prime step for that part.


sergiusens:
More than `build-options` I'd call these `build-attributes`

kyrofa:
> More than `build-options` I'd call these `build-attributes`

Works for me-- done.

kyrofa:
> Looks good, the in-code and md documentation needs to be updated to add this though.

Ah of course! Done.

PR #990, opened 2016-12-16
tests: fix snaps tests in armhf

elopio:
LP: #1650458

### DIFF
--- a/integration_tests/test_maven_plugin.py
+++ b/integration_tests/test_maven_plugin.py
@@ -22,7 +22,7 @@ import integration_tests
 class MavenPluginTestCase(integration_tests.TestCase):
 
     def test_build_maven_plugin(self):
-        if snapcraft.project_options().deb_arch == 'armv7l':
+        if snapcraft.ProjectOptions().deb_arch == 'armv7l':
             # https://bugs.launchpad.net/snapcraft/+bug/1647405
             self.skipTest('The maven plugin does not support armhf')
         project_dir = 'simple-maven'
--- a/integration_tests/test_maven_plugin.py
+++ b/integration_tests/test_maven_plugin.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright (C) 2015, 2016 Canonical Ltd
+# Copyright (C) 2015, 2016, 2017 Canonical Ltd
 #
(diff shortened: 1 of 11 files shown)


kyrofa:
(arm tests are failing here).

snaps_tests/demos_tests/test_git.py
```diff
@@ -0,0 +1,36 @@
...
+
+import snaps_tests
+
+
+class GitTestCase(snaps_tests.SnapsTestCase):
+
+    snap_content_dir = 'git'
+
+    def test_gopaste(self):
+        if platform.machine() == 'armv7l':
```
sergiusens: please use an instance of `project_options` for these so we don't have to make a global change everywhere later.

snaps_tests/__main__.py
```diff
@@ -47,7 +48,11 @@ def main():
 
     arguments = docopt.docopt(__doc__)
 
-    snaps_tests.config['skip-install'] = arguments['--skip-install']
+    if platform.machine() == 'armv7l':
```
sergiusens: project_options

elopio: @sergiusens but project options is available only when snapcraft is running. These tests just call snapcraft. I'm not sure how you want me to access project_options.

Here I'm just checking for the machine identified as armv7l in our autopkgtests. I'm not yet preoccupied about the actual arch, because that should happen in the plugin. If it doesn't support a particular architecture, it should display a nice error to the user.

sergiusens:
```python
from snapcraft import project_options
...
...
if project_options().deb_arch == 'armhf':
```

sergiusens: Or in the main test class
```python
self.deb_arch = project_options().deb_arch
```

sergiusens: Or don't use project_options if you are really against it, but set it up only once in `setUp`

integration_tests/test_maven_plugin.py
```diff
@@ -14,15 +14,15 @@
...
+import snapcraft
 
 import integration_tests
 
 
 class MavenPluginTestCase(integration_tests.TestCase):
 
     def test_build_maven_plugin(self):
-        if platform.machine() == 'armv7l':
+        if snapcraft.project_options().deb_arch == 'armv7l':
```
sergiusens: this change is certainly incomplete, right?

PR #992, opened 2016-12-16
plugins: update make plugin to use get_build_properties() 

ZenHarbinger:
Updated make plugin to use get_build_properties().
Updated test_plugin_make as well to test this method.

https://bugs.launchpad.net/snapcraft/+bug/1650548

### DIFF
--- a/snapcraft/plugins/cmake.py
+++ b/snapcraft/plugins/cmake.py
@@ -57,7 +57,7 @@ class CMakePlugin(snapcraft.plugins.make.MakePlugin):
     def get_build_properties(cls):
         # Inform Snapcraft of the properties associated with building. If these
         # change in the YAML Snapcraft will consider the build step dirty.
-        return ['configflags']
+        return super().get_build_properties() + ['configflags']
 
     def __init__(self, name, options, project):
         super().__init__(name, options, project)
--- a/snapcraft/plugins/make.py
+++ b/snapcraft/plugins/make.py
@@ -89,7 +89,7 @@ class MakePlugin(snapcraft.BasePlugin):
     def get_build_properties(cls):
         # Inform Snapcraft of the properties associated with building. If these
         # change in the YAML Snapcraft will consider the build step dirty.
-        return ['makefile', 'make-parameters', 'make-install-var']
+        return ['makefile', 'make-parameters', 'make-install-var', 'artifacts']
 
     def __init__(self, name, options, project):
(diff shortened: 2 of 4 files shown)


snapcraft/plugins/make.py
```diff
@@ -83,12 +83,13 @@ def schema(cls):
...
+
+    @classmethod
+    def get_build_properties(cls):
         # Inform Snapcraft of the properties associated with building. If these
         # change in the YAML Snapcraft will consider the build step dirty.
-        schema['build-properties'].extend(
-            ['makefile', 'make-parameters', 'make-install-var'])
-
-        return schema
+        return ['makefile', 'make-parameters', 'make-install-var']
```
kyrofa: `artifacts` probably need to be here as well.

ZenHarbinger:
Added, should be testing now!

snapcraft/plugins/make.py
```diff
@@ -83,12 +83,13 @@ def schema(cls):
...
+
+    @classmethod
+    def get_build_properties(cls):
         # Inform Snapcraft of the properties associated with building. If these
         # change in the YAML Snapcraft will consider the build step dirty.
-        schema['build-properties'].extend(
-            ['makefile', 'make-parameters', 'make-install-var'])
-
-        return schema
+        return ['makefile', 'make-parameters', 'make-install-var', 'artifacts']
```
kyrofa: Note that, after #978, this will overwrite the `cmake` build properties. This PR is likely the place to fix that (`cmake` should be building on `make`'s version).

ZenHarbinger: Yes, I am aware of having to fix cmake.  It will hit a PR once possible to fix w/ this code base.
cmake already uses get_build_properites(), so I'll have to call MakePlugin.get_buid_properties() to fix, but can't until this PR, I think....

ZenHarbinger: Understood.  I have the update in place.

snapcraft/plugins/cmake.py
```diff
@@ -57,7 +57,7 @@ def schema(cls):
     def get_build_properties(cls):
         # Inform Snapcraft of the properties associated with building. If these
         # change in the YAML Snapcraft will consider the build step dirty.
-        return ['configflags']
+        return super().get_build_properties() + ['configflags']
```
kyrofa: Thank you.

PR #995, opened 2016-12-16
plugins: update kernel plugin to use get_build_properties()

ZenHarbinger:
Updated kernel to use get_build_properties()

https://bugs.launchpad.net/snapcraft/+bug/1650597

### DIFF
--- a/snapcraft/plugins/kernel.py
+++ b/snapcraft/plugins/kernel.py
@@ -130,14 +130,16 @@ class KernelPlugin(kbuild.KBuildPlugin):
             'enum': ['gz'],
         }
 
+        return schema
+
+    @classmethod
+    def get_build_properties(cls):
         # Inform Snapcraft of the properties associated with building. If these
         # change in the YAML Snapcraft will consider the build step dirty.
-        schema['build-properties'].extend([
+        return [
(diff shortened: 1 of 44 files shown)


snapcraft/plugins/kernel.py
```diff
@@ -130,14 +130,16 @@ def schema(cls):
...
         }
 
+        return schema
+
+    @classmethod
+    def get_build_properties(cls):
         # Inform Snapcraft of the properties associated with building. If these
         # change in the YAML Snapcraft will consider the build step dirty.
-        schema['build-properties'].extend([
+        return kbuild.KBuildPlugin.get_build_properties() + [
```
kyrofa: `super().get_build_properties()` might be a bit nicer, here.

ZenHarbinger: If I switch to super.get_build_properties() I get the following error:
'super' has no attribute 'get_build_properties'
Because it's a classmethod.

kyrofa: `super` is a method: `super()`.

ZenHarbinger: Hey! I learned something. OK, I'll get it.

PR #996, opened 2016-12-16
Updated nodejs plugin to use get_build_properties()

ZenHarbinger:
Updated nodjs to use get_build_properties() and get_pull_properties()
Updated the schema testing in the unit tests.
Added get_build_properties() and get_pull_properties() to the test file.

https://bugs.launchpad.net/snapcraft/+bug/1650600

### DIFF
--- a/snapcraft/plugins/nodejs.py
+++ b/snapcraft/plugins/nodejs.py
@@ -90,14 +90,19 @@ class NodePlugin(snapcraft.BasePlugin):
         if 'required' in schema:
             del schema['required']
 
-        # Inform Snapcraft of the properties associated with building. If these
-        # change in the YAML Snapcraft will consider the build step dirty.
-        schema['build-properties'].extend(['node-packages', 'npm-run'])
+        return schema
+
+    @classmethod
+    def get_build_properties(cls):
         # Inform Snapcraft of the properties associated with pulling. If these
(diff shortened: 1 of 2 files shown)


snapcraft/plugins/nodejs.py
```diff
@@ -90,14 +90,19 @@ def schema(cls):
...
             del schema['required']
 
-        # Inform Snapcraft of the properties associated with building. If these
-        # change in the YAML Snapcraft will consider the build step dirty.
-        schema['build-properties'].extend(['node-packages', 'npm-run'])
+        return schema
+
+    @classmethod
+    def get_build_properties(cls):
         # Inform Snapcraft of the properties associated with pulling. If these
```
kyrofa: There are the properties associated with _building_, not pulling.

snapcraft/tests/test_plugin_nodejs.py
```diff
@@ -158,27 +159,89 @@ class Options:
                          'architecture not supported (fantasy-arch)')
 
     def test_schema(self):
```
kyrofa: This was a hefty rewrite, thanks for that.

snapcraft/plugins/nodejs.py
ZenHarbinger: Fixed.

PR #1002, opened 2016-12-16
plugins: update python plugin to support get_pull_properties()

ZenHarbinger:
Also updated schema test in unit tests.
Added tests for get_pull_properties() to test.

https://bugs.launchpad.net/snapcraft/+bug/1650626

### DIFF
--- a/snapcraft/tests/test_plugin_python.py
+++ b/snapcraft/tests/test_plugin_python.py
@@ -269,7 +269,7 @@ class PythonPluginTestCase(tests.TestCase):
         self.assertListEqual(expected_fileset, fileset)
 
     @mock.patch.object(python.PythonPlugin, 'run')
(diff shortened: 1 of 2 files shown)


snapcraft/tests/test_plugin_python.py
```diff
@@ -257,7 +269,7 @@ def test_fileset_ignores(self):
         self.assertListEqual(expected_fileset, fileset)
 
     @mock.patch.object(python.PythonPlugin, 'run')
-    def test_build_fixes_python_shebangs(self, run_mock):
+    def test_pull_fixes_python_shebangs(self, run_mock):
```
kyrofa: Why this change? The shebang rewrite is happening in build, and this test tests the build.

ZenHarbinger:
Yes, that was an unexpected rename of the test method.
Renamed and re-committed.

PR #1006, opened 2016-12-17
schema: replace `snap` filter with `prime` filter.

kyrofa:
This PR fixes LP: [#1627781](https://bugs.launchpad.net/snapcraft/+bug/1627781) by deprecating the `snap` keyword, and replacing it with `prime`. It also makes sure that they can't both be specified.


integration_tests/test_prime_filter.py
```diff
@@ -0,0 +1,53 @@
...
+        self.assertThat(os.path.join(primedir, 'prime2'), Not(FileExists()))
+
+    def test_snap_filter_is_deprecated(self):
+        project_dir = 'simple-prime-filter'
+        output = self.run_snapcraft(['prime', 'snap-keyword'], project_dir)
+
+        # Verify that the `snap` keyword is deprecated.
+        self.assertThat(
+            output, Contains(
+                "DEPRECATED: The 'snap' keyword has been replaced by 'prime'"))
```
sergiusens: assert it contains a link to the KB https://github.com/snapcore/snapcraft/wiki/Knowledge-Base#kb001

change kb001 to something else if appropriate.

kyrofa: I took your suggestion and ran with it a little, adding a central mechanism for dealing with deprecation notices and linking them back to the wiki. Let me know what you think.

kyrofa:
> I do wonder though, why do you not pop the `snap` keyword at the project loader level, is it just because of status tracking?

Indeed, it's to make sure we know when that keyword changes, thus knowing when `prime` is dirty.

> Given it is automatic anyways, does it make sense at all?

I'm assuming you're referring to the fact that the prime step is automatically cleaned/re-run if we discover that it's dirty. However, we still need to be able to know if it's dirty in order to clean/re-run it, which is why we need to track this keyword. If it changes, the `prime` step is considered dirty, it is automatically cleaned, and re-run.

kyrofa:
> While there is an integration test, maybe a unit test for the deprecation warning would be good.

Sure thing-- done.

PR #1007, opened 2016-12-19
lifecycle: clean without parsing if possible.

kyrofa:
Currently snapcraft cleans each step of each part individually, even when `clean` is run without any parts or steps. This wastes time parsing YAML, and will even fail to clean if the YAML is invalid.

This PR fixes LP: [#1596596](https://bugs.launchpad.net/snapcraft/+bug/1596596) by cleaning smarter; it parses YAML as a last resort. If no parts are supplied, it cleans entire steps.


integration_tests/snaps/invalid-snap/snapcraft.yaml
```diff
@@ -0,0 +1,16 @@
+# This snap is missing a name
```
sergiusens: Fix the entries here to use proper descriptions and/or entries

kyrofa:
> this does have the side-effect of leaving orphaned files in prime and stage, right? Not sure that is something we want to silently happen.

Indeed that would be bad-- no, that's not the case. The only time files are left in prime/stage in the current snapcraft is if individual parts are cleaned (while others remained staged or primed). That's still the case here. This PR [only affects cleaning when no parts are specified][1], in which case we don't need to know anything about parts (don't need to parse the YAML and don't need to remove files piecemeal).

[1]: https://github.com/snapcore/snapcraft/pull/1007/files#diff-2c5b7b62eb9a54c9bea1bbb72b29345bR508

integration_tests/snaps/invalid-snap/snapcraft.yaml
kyrofa: Done.

snapcraft/internal/lifecycle.py
```diff
@@ -442,32 +438,81 @@ def _cleanup_common_directories(config, project_options):
...
+                except NotADirectoryError:
+                    os.remove(path)
+    for part in parts:
+        part.mark_cleaned('build')
+        part.mark_cleaned('pull')
 
 
 def clean(project_options, parts, step=None):
+    if not step:
+        step = 'pull'
```
sergiusens: why don't you make the default `step=None` `step='pull'` instead?

snapcraft/internal/lifecycle.py
```diff
@@ -442,32 +438,81 @@ def _cleanup_common_directories(config, project_options):
...
+
+    if not parts and step == 'pull':
```
sergiusens: it feels like what you want here is
`if not parts and not step` and keep `step` as None? If we ever add a new pre-pull step this will be a point of failure, I get that there may be others.

snapcraft/tests/commands/test_clean.py
```diff
@@ -178,31 +178,19 @@ def test_per_step_cleaning(self, mock_clean):
...
-            'clean2': states.PrimeState({'clean2'}, set()),
-        }
-
-        self.assertEqual('DEPRECATED: Use `prime` instead of `strip` as '
-                         'the step to clean\n', fake_logger.output)
-        mock_clean.assert_called_with(
-            expected_staged_state, expected_primed_state, 'prime')
+        self.assertThat(
+            fake_logger.output, MatchesRegex(
+                'DEPRECATED: Use `prime` instead of `strip` as the step to '
```
sergiusens: will this point to the NB as well?

snapcraft/internal/lifecycle.py
kyrofa: Because the user can run `snapcraft clean -s pull` as well, which would set `step='pull'` here, not `None`.

snapcraft/internal/lifecycle.py
kyrofa: Well, I... I... have no idea. Fixed.

sergiusens:
```
ERROR: test_clean_invalid_yaml (test_clean.CleanTestCase)
test_clean.CleanTestCase.test_clean_invalid_yaml
testtools.testresult.real._StringException: output: {{{Issues while validating snapcraft.yaml: 'name' is a required property}}}
  File "/home/travis/build/snapcore/snapcraft/integration_tests/test_clean.py", line 54, in test_clean_invalid_yaml
    self.run_snapcraft('clean', project_dir)
  File "/home/travis/build/snapcore/snapcraft/integration_tests/__init__.py", line 77, in run_snapcraft
    stderr=subprocess.STDOUT, universal_newlines=True)
    **kwargs).stdout
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['/home/travis/build/snapcore/snapcraft/bin/snapcraft', 'clean', '-d']' returned non-zero exit status 1
```

kyrofa:
Ah ha! That's why `step=None` and is then set to `'pull'`: `args['--step']` is None when it's not supplied. Reverted and working again.

snapcraft/internal/lifecycle.py
```diff
@@ -442,32 +438,83 @@ def _cleanup_common_directories(config, project_options):
...
+                except NotADirectoryError:
+                    os.remove(path)
+    for part in parts:
+        part.mark_cleaned('build')
+        part.mark_cleaned('pull')
 
 
 def clean(project_options, parts, step=None):
+    # step defaults to None because that's how it comes from docopt when it's
+    # not set.
```
sergiusens: In retrospect, it would of been better to do this on the docopt side; just leave it as is not though ;-)

kyrofa: Yeah, that's a little embarrassing. That's definitely the way to do it.

PR #1009, opened 2016-12-20
store: implement push pre-check.

squidsoup:
LP: #1610776

This branch adds a precheck request to the store to `snapcraft push`, which in the case of an unregistered snap will report the error to the user before attempting a full upload.

### DIFF
--- a/snapcraft/_store.py
+++ b/snapcraft/_store.py
@@ -409,11 +409,14 @@ def push(snap_filename, release_channels=None):
         raise FileNotFoundError(
             'The file {!r} does not exist.'.format(snap_filename))
 
-    logger.info('Uploading {}.'.format(snap_filename))
-
     snap_yaml = _get_data_from_snap_file(snap_filename)
     snap_name = snap_yaml['name']
     store = storeapi.StoreClient()
+
+    with _requires_login():
+        store.push_precheck(snap_name)
+
+    logger.info('Uploading {}.'.format(snap_filename))
     with _requires_login():
         tracker = store.upload(snap_name, snap_filename)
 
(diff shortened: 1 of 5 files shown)


snapcraft/_store.py
```diff
@@ -409,11 +409,14 @@ def push(snap_filename, release_channels=None):
...
             'The file {!r} does not exist.'.format(snap_filename))
 
-    logger.info('Uploading {}.'.format(snap_filename))
-
     snap_yaml = _get_data_from_snap_file(snap_filename)
     snap_name = snap_yaml['name']
     store = storeapi.StoreClient()
+
+    with _requires_login():
+        store.push_precheck(snap_name)
```
kyrofa: Does this step take any significant amount of time? Should a message be printed about what it's doing in case it hangs at all?

squidsoup:
It does require an http request, which could potentially timeout if there's
an issue with the store. Logging something might be useful, but I'm not
sure what the best way to convey what the 'precheck' is for to the user.
Any suggestions? Maybe "Checking snap status in store..."?

kyrofa:
> Maybe "Checking snap status in store..."?

Yeah that's fine, or even as opaque as "Checking with the store..." would be fine. The reason here is I think twofold:

- Even though this isn't a step of which the user needs to be explicitly aware, this would provide a better experience if issues do happen.
- If issues do happen, this makes the log more useful to us.

snapcraft/_store.py
```diff
@@ -409,11 +409,14 @@ def push(snap_filename, release_channels=None):
...
+
+    logger.info('Uploading {}.'.format(snap_filename))
```
sergiusens: can we take the opportunity to call change this to `'Pushing {!r}'`?

snapcraft/_store.py
sergiusens: or do this after logging the message for which I want to change anyways ;-)

PR #1010, opened 2016-12-20
meta: add 'desktop' entry for apps

oSoMoN:
LP: #1588359


snapcraft/internal/meta.py
```diff
@@ -122,10 +123,15 @@ def _setup_from_setup(self):
...
         gui_src = os.path.join(setup_dir, 'gui')
         gui_dst = os.path.join(self.meta_dir, 'gui')
-        if os.path.exists(gui_dst) and os.path.exists(gui_src):
-            shutil.rmtree(gui_dst)
         if os.path.exists(gui_src):
-            shutil.copytree(gui_src, gui_dst)
+            for f in os.listdir(gui_src):
+                if os.path.splitext(f)[1] == '.desktop':
+                    logger.warning(
+                        "DEPRECATED: use the 'desktop' key for apps instead "
```
sergiusens: I wouldn't necessarily deprecate this, new projects might not care about existing conventions.

schema/snapcraft.yaml
```diff
@@ -88,6 +88,12 @@ properties:
           stop-command: ["daemon"]
         additionalProperties: false
         properties:
+          desktop:
+            type: string
+            description: path to a desktop file representing the app, relative to the prime directory
+          icon:
```
sergiusens: is this here to support replacing icon names in existing desktop files instead of paths?

snapcraft/internal/meta.py
oSoMoN: I don’t have a strong opinion on this, so I can undeprecate it if you wish.

schema/snapcraft.yaml
oSoMoN: This is here to allow overriding the icon in the upstream desktop file with a custom one for the snap.

For example, for webbrowser-app the Icon line for the input desktop file (before transformation) is:

    Icon=@CMAKE_INSTALL_FULL_DATADIR@/webbrowser-app/webbrowser-app.png

Typically on a deb system that would end up being:

    Icon=/usr/share/webbrowser-app/webbrowser-app.png

But when building a snap from the source tree, the line in the resulting desktop file in the prime directory will look like this:

    Icon=/share/webbrowser-app/webbrowser-app.png

Allowing to override the icon in snapcraft.yaml means upstream doesn’t need to care to support snaps.

sergiusens: That last case is where I'd like to be smart, detect it is a path and apply the correct snapcraft magic foo of prefixing $SNAP in that path.

snapcraft/internal/meta.py
```diff
@@ -215,19 +217,62 @@ def _wrap_exe(self, command, basename=None):
...
+            try:
+                app[k] = self._wrap_exe(app[k], '{}-{}'.format(k, name))
+            except CommandError as e:
+                raise EnvironmentError(
+                    'The specified command {!r} defined in the app {!r} '
+                    'does not exist or is not executable'.format(str(e), name))
+        if 'desktop' in app:
+            self._wrap_desktop_and_icon(name, app)
+
+    def _wrap_desktop_and_icon(self, name, app):
```
sergiusens: we actually reformat the desktop file instead of wrapping it. If you want to change the name to reflect that it would be great

snapcraft/internal/meta.py
```diff
@@ -215,19 +217,62 @@ def _wrap_exe(self, command, basename=None):
...
+                'desktop file'.format(desktop_file))
+        if 'Exec' not in desktop_contents[section]:
+            raise EnvironmentError(
+                'The specified desktop file {!r} is missing the '
+                '"Exec" key'.format(desktop_file))
+        # XXX: do we want to allow more parameters for Exec?
+        exec_value = '{}.{} %U'.format(self._config_data['name'], name)
+        desktop_contents[section]['Exec'] = exec_value
+        if 'Icon' in desktop_contents[section]:
+            icon = desktop_contents[section]['Icon']
```
sergiusens: would be good to fail or warn if it does not exist

sergiusens: maybe fallback to the snap icon in meta/gui/icon if it exists, but this would be a different PR ;-)

oSoMoN: Done, a warning will now be issued if the icon does not exist.

oSoMoN: And I can confirm that I tested this branch on the webbrowser-app snap, and it works as expected. I haven’t tested on other snaps though.

PR #1011, opened 2016-12-20
ci: use a named docker instance with proper working dir and env

3v1n0:
And export there the host's TERM variable.

LP: #1659372

### DIFF
--- a/snapcraft/integrations/travis.py
+++ b/snapcraft/integrations/travis.py
@@ -48,8 +48,8 @@ See the example below::
     deploy:
       skip_cleanup: true
       provider: script
-      script: docker run -v $(pwd):$(pwd) -t ubuntu:xenial sh -c
-        "apt update -qq && apt install snapcraft -y && cd $(pwd) &&
+      script: docker run -e TERM -v $PWD:$PWD -w $PWD -t ubuntu:xenial sh -c
+        "apt update -qq && apt install snapcraft -yq &&
         snapcraft && snapcraft push *.snap --release edge"
       on:
         branch: master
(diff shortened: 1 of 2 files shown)


elopio:
@3v1n0 Thanks. What were we missing without the TERM variable? Can you please report a bug for the problem that you are fixing?

3v1n0:
@elopio bug added.

> What were we missing without the TERM variable?

This will ensure that the docker instance will inherit the TERM variable from the parent terminal, thus it will optionally hide all the unneeded content from logs (if set to TERM=dumb) as per the same reasons that it was introduced in PR: #906.

snapcraft/integrations/travis.py
```diff
@@ -218,9 +218,8 @@ def enable():
             'skip_cleanup': True,
             'provider': 'script',
             'script': (
-                'docker run -v $(pwd):$(pwd) -t ubuntu:xenial sh -c '
-                '"apt update -qq && apt install snapcraft -y && cd $(pwd) && '
-                'snapcraft && snapcraft push *.snap --release edge"'),
+                'docker run -e TERM -v $PWD:$PWD -w $PWD -t snapcore/snapcraft'
+                ' sh -c "snapcraft && snapcraft push *.snap --release edge"'),
```
elopio: @3v1n0 I think this will also need apt update and apt upgrade, because the snapcore/snapcraft image is not being updated: https://bugs.launchpad.net/snapcraft/+bug/1639112

elopio:
I like -w instead of cd, and I didn't know about it, thanks.

However, being a little pita, there are three bugs here:
1. we are not exporting TERM
2. we are using cd instead of -w
3. we are not using a named docker instance.

I'm ok with fixing the three in the same PR, but my problem is that the description of the PR only mentions 1. The bug mentions 2 and 3. And the PR fixes 1 and 2. :)

Also, this is not your fault, but you are changing two things in the code, and only one in the tests. That means that we are missing a test.

sergiusens:
you might want to look at #1125 wrt usage of `-e`

kyrofa:
Where does this PR stand?

sergiusens:
Closing due to lack of activity, feel free to reopen once updated.

PR #1016, opened 2016-12-22
cmake: extend parte build properties as it extends make

3v1n0:
I guess we want to properly update the builds properties extending partent's one too.

In case we don't care about those in cmake plugin, we should instead probably drop them from the schema too.


ZenHarbinger:
This is in #992 , that as my bad and should be in the master branch by now.

3v1n0:
Nice, thanks.

PR #1017, opened 2016-12-22
make: add support for cwd path

3v1n0:
This can be useful for plugins that extends Make so that they can define a subfolder where to run make


kyrofa:
It's not clear to me why this is necessary. This is what `source-subdir` is for. Perhaps you could log a bug about it (as requested in the [contribution guide](https://github.com/snapcore/snapcraft/blob/master/CONTRIBUTING.md) so we can discuss?

kyrofa:
Closing due to inactivity. Please reopen once you're ready to continue. Thank you!

PR #1018, opened 2016-12-22
autotools: extend Make plugin instead of repeating code

3v1n0:
We can use the self.make function from make instead of redefining the build process again here.

This has to be rebased once PR: #1017 lands.

### DIFF
--- a/snapcraft/plugins/autotools.py
+++ b/snapcraft/plugins/autotools.py
@@ -42,10 +42,10 @@ In addition, this plugin uses the following plugin-specific keywords:
 import os
 import stat
 
-import snapcraft
+from snapcraft.plugins import make
 
 
-class AutotoolsPlugin(snapcraft.BasePlugin):
+class AutotoolsPlugin(make.MakePlugin):
 
     @classmethod
     def schema(cls):
@@ -71,7 +71,7 @@ class AutotoolsPlugin(snapcraft.BasePlugin):
     def get_build_properties(cls):
         # Inform Snapcraft of the properties associated with building. If these
         # change in the YAML Snapcraft will consider the build step dirty.
-        return ['configflags', 'install-via']
+        return super().get_build_properties() + ['configflags', 'install-via']
 
     def __init__(self, name, options, project):
         super().__init__(name, options, project)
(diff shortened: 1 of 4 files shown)


snapcraft/tests/test_plugin_autotools.py
```diff
@@ -105,7 +110,8 @@ def test_schema(self):
                          'but it was "{}"'.format(installvia_default))
 
     def test_get_build_properties(self):
-        expected_build_properties = ['configflags', 'install-via']
+        expected_build_properties = make.MakePlugin.get_build_properties() + \
+            ['configflags', 'install-via']
```
kyrofa: Personally I like parentheses a bit more than escaped newlines.

snapcraft/plugins/make.py
```diff
@@ -116,9 +116,12 @@ def make(self, env=None):
                     snapcraft.file_utils.link_or_copy(
                         source_path, destination_path)
         else:
-            install_param = self.options.make_install_var + '=' + \
-                self.installdir
-            self.run(command + ['install', install_param], env=env)
+            command.append('install')
+            if self.options.make_install_var:
+                command.append("{}={}".format(
```
kyrofa: Single quotes here, please.

snapcraft/plugins/autotools.py
```diff
@@ -113,18 +111,15 @@ def build(self):
...
             # Use an empty prefix since we'll install via DESTDIR
             configure_command.append('--prefix=')
-            make_install_command.append('DESTDIR=' + self.installdir)
         else:
             configure_command.append('--prefix=' + self.installdir)
 
         self.run(configure_command + self.options.configflags)
-        self.run(['make', '-j{}'.format(self.parallel_build_count)])
-        self.run(make_install_command)
+        self.make()
```
kyrofa: Forgive my density (I only just started the coffee IV), but how does this use #1017?

kyrofa:
Please also log a bug regarding this issue (as outlined in the [contribution guide](https://github.com/snapcore/snapcraft/blob/master/CONTRIBUTING.md)). Thank you!

kyrofa:
Closing due to inactivity. Please reopen once you're ready to continue. Thank you!

snapcraft/plugins/autotools.py
3v1n0: No, it doesn't.But next hunk will conflict with that...

snapcraft/tests/test_plugin_autotools.py
3v1n0: Ack

3v1n0:
> Overall the approach looks good. Although since it essentially saves two lines of code by adding several more I'm not sure I'd use the "... instead of repeating code" line :stuck_out_tongue: . That said the Make plugin has a few features not contained within autotools, and I could see them being useful.

Yeah, in fact that's the actual reason... Inheriting all the configuration options. :-)

Added also the bug, and please reopen this PR, as I don't have the rights for.

kyrofa:
> Added also the bug, and please @kyrofa reopen this PR, as I don't have the rights for.

Ah, it seems you force-pushed-- GitHub won't let me re-open it now, either. Please go ahead and re-propose it.

PR #1020, opened 2016-12-23
Change hello.sh to be executable

liu-xiao-guo:
The hello.sh in the alias test code is not runnable.


kyrofa:
Huh, that feature seemed to have landed without an integration test that uses that snap, so this is a slightly larger fix I'm afraid. I've done it in #1021, so I'll close this in favor of that.

Thanks!

PR #1024, opened 2016-12-30
gradle plugin: update gradle plugin to support both gradle and gradlew

ZenHarbinger:
Updated gradle plugin to check if ./gradlew exists in os.cwd().  
If it does not, then 'gradle' is added to the get_build_dependencies() list.  
If it does, then ./gradlew is called, if it does not, then gradle is called.  

Updated tests to use both gradle and gradlew (so a complete duplication of all tests).

Added an option to specify the gradle-output-dir which will by copied to the installdir.
The default value here is 'build/libs' which is what the old default was hard-coded to be.

https://bugs.launchpad.net/snapcraft/+bug/1653182


ZenHarbinger:
### Info on why I have modified this plugin

Gradle is a build system for Java applications.  Currently, snapcraft supports the gradle wrapper (graldlew) which is what is used so that developers can build the source without the full gradle suite installed.

However, the plugin to snapcraft is limited to supporting only this wrapper.  There are other applications that support/require using the full gradle suite.

### Changes
Within snapcraft, the invocation of gradle v. ./gradlew is determined by the existence of the wrapper script.  Same goes for if there is a build dependency which snapcraft must install.

The arguments passed to wrapper and to gradle are the same.

On top of these changes, I have added in the ability to specify what the target output directory is from gradle.  Gradle has default output directories, but it is very flexible and is possible to change the output directory to be anything.  Thus, snapping the data from the gradle plugin is always easily done.  The default is build/libs which works if nothing is altered.

### Example
On the snapcraft mailing list, there was  a request for help building a gradle application.  The output of this application was 'BUILD/'.

This modification would bring in the full support of gradle or gradlew into being where before it was only 'gradlew'.

snapcraft/plugins/gradle.py
```diff
@@ -56,28 +56,39 @@ def schema(cls):
             },
             'default': [],
         }
+        schema['properties']['gradle-output-dir'] = {
```
sergiusens: this would require documentation updates

snapcraft/plugins/gradle.py
```diff
@@ -110,3 +119,27 @@ def _get_proxy_options(self):
                     proxy_options.append(
                         '-D{}.proxyPort={}'.format(var, parsed_url.port))
         return proxy_options
+
+
+def _link_or_copy(source, destination, boundary):
```
sergiusens: it would be nice to extend `fileutils.link_or_copy` with this logic. Maybe a new parameter `copy_out_of_boundary` (set to `False` by default to not break compatibility.

ZenHarbinger: Can I recommend this do a separate PR?
I copied this from the `dump` plugin.  Where I should really just be using file_utils.link_or_copy, which is what I'm going to check in.
https://github.com/snapcore/snapcraft/blob/master/snapcraft/plugins/dump.py

snaps_tests/demos_tests/test_gradle_with_outputdir.py
```diff
@@ -0,0 +1,29 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import snaps_tests
+
+
+class GradleTestCase(snaps_tests.SnapsTestCase):
+
+    snap_content_dir = 'gradle-with-outputdir'
+
+    def test_java_hello_world(self):
```
sergiusens: this would need a similar construct to the one in https://github.com/snapcore/snapcraft/blob/master/snaps_tests/demos_tests/test_gradle.py#L27

Or fix the armhf issue we have which would also be a welcome addition!

sergiusens: disregard, the test ran fine for gradle

sergiusens:
Ignoring the travis error since:
```
$  cla-check matt.aguirre@gmail.com
Logging into Launchpad...
Contributor: https://api.launchpad.net/1.0/~matt-aguirre
The contributor has signed the CLA.
```

PR #1026, opened 2017-01-05
Add support for hooks.

kyrofa:
Hooks are now supported in snapd and the click reviewer tools. Time to finish off LP: [#1586465](https://bugs.launchpad.net/ubuntu/+source/snapd/+bug/1586465) by supporting them in snapcraft as well. This is done by convention, and two methods are supported:

1. Place hook executables in the `snap/hooks/` directory, and they will be copied into the snap automatically.
2. Install hook executables from parts into the `snap/hooks/` directory.

### DIFF
--- a/snapcraft/internal/meta.py
+++ b/snapcraft/internal/meta.py
@@ -14,6 +14,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import contextlib
 import os
 import configparser
 import logging
@@ -49,6 +50,7 @@ _OPTIONAL_PACKAGE_KEYS = [
     'confinement',
     'epoch',
     'grade',
+    'hooks',
 ]
 
 
@@ -58,7 +60,7 @@ class CommandError(Exception):
 
 def create_snap_packaging(config_data, snap_dir, parts_dir):
     """Create snap.yaml and related assets in meta.
-    Create  the meta directory and provision it with snap.yaml
+    Create the meta directory and provision it with snap.yaml
     in the snap dir using information from config_data.
 
     :param dict config_data: project values defined in snapcraft.yaml.
@@ -67,6 +69,7 @@ def create_snap_packaging(config_data, snap_dir, parts_dir):
     packaging = _SnapPackaging(config_data, snap_dir, parts_dir)
     packaging.write_snap_yaml()
     packaging.setup_assets()
+    packaging.write_snap_directory()
 
     return packaging.meta_dir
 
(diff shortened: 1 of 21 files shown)


snapcraft/internal/meta.py
```diff
@@ -58,7 +60,7 @@ class CommandError(Exception):
 
 def create_snap_packaging(config_data, snap_dir, parts_dir):
     """Create snap.yaml and related assets in meta.
-    Create  the meta directory and provision it with snap.yaml
+    Create the meta directory and provision it with snap.yaml
     in the snap dir using information from config_data.
```
sergiusens: since you edited this text you can add the part about checking for `snap`

snapcraft/internal/meta.py
```diff
@@ -116,6 +119,38 @@ def setup_assets(self):
...
+
+            for file_path in files:
+                source = os.path.join(root, file_path)
+                destination = os.path.join(self._snap_dir, source)
+                with contextlib.suppress(FileNotFoundError):
+                    os.remove(destination)
+                file_utils.link_or_copy(source, destination)
+
+        # Now write hook wrappers for any hooks included in that directory.
+        self._generate_hook_wrappers()
```
sergiusens: I would only generate hook wrappers for things coming from parts just to not break any assumption a manually crafted hook would have.
To do so, I guess it is just enough to rename `_general_hook_wrappers` to a public one and call it before `packaging.write_snap_directory()` above.

It makes sense to move it out as well as `write_snap_directory` will eventually do more.

kyrofa: > I would only generate hook wrappers for things coming from parts just to not break any assumption a manually crafted hook would have.

If we don't generate wrappers for the manually crafted ones, there's nothing connecting the dots between our magic location (snap/hooks) and snapd's (meta/hooks). Are you thinking we'd just generate a dumb wrapper?

I'm trying to think of assumptions a manually crafted hook might have that would break with our setting environment variables... I'm not coming up with any, do you know? However, I can quickly come up with a breakage if we don't set them: if the manually crafted hook uses any python executables within it, it'll immediately break. And there's no way around it except to move them out of the `snap` directory and install them via a part. The inconsistency there might make for a poor experience.

sergiusens: An example, a hook wanting to use the system python where the wrapper gets in the way.

kyrofa: How would that be done in an app?

sergiusens: If I have this tree:

```
./
./snap/hooks/configure
```

After snapping I expect this `configure` to land as is on
```
./prime/
    meta/hooks/configure
    snap/hooks/configure
```

If in the other case this came from a part, those two `configure`'s would be different, the former the wrapped one while the latter the built one.

sergiusens: "How would that be done in an app?" How are apps related?

sergiusens: We don't have `apps` by convention

kyrofa: Apps are related in that they're runnable things exposed by snapcraft, as are hooks. One is run by the user, one is run by snapd, but they're both exposed by snapcraft. What I mean to say is, if we don't expect developers to want to do such a thing in apps, why do we expect them to want to do such a thing in a hook?

sergiusens: we don't have apps byt convention now, do we? We were also given the guidelines to not compare these to apps ;-)

kyrofa: This is sitting right next to the code that is doing the same job for apps. It's in the same file, a few lines away! It's difficult not to compare how the two are handled.

kyrofa: Alright, done.

snapcraft/internal/meta.py
kyrofa: Done.

PR #1028, opened 2017-01-06
[Highly experimental] Run the integration suite in parallel

elopio:


sergiusens:
interestingly I thought this would choke on build packages

elopio:
It does. 11 tests failed. We need to either put all the tests with build packages in the same queue, or implement a retry strategy for apt. I'm going to close this for now. I'll keep trying it later. I want to see the time savings, but first need all the tests to run.

PR #1030, opened 2017-01-06
tests: fix broken delta upload unit test.

shawn111:
Clear DELTA_UPLOADS_EXPERIMENTAL environment variable.

LP: #1654507


shawn111:
It is an invalid PR, removed it.

PR #1032, opened 2017-01-06
Use more secure temporary directory for parser runs.

josepht:
LP:#1650427


3v1n0:
Wasn't `XDG_RUNTIME_DIR` better for such thing?

PR #1034, opened 2017-01-07
rust plugin: respect fetch parameters.

sergiusens:
LP: #1654718

### DIFF
--- a/integration_tests/snaps/rust-with-revision/Cargo.toml
+++ b/integration_tests/snaps/rust-with-revision/Cargo.toml
@@ -1,7 +1,9 @@
 [package]
-name = "simple-rust"
+name = "rust-with-revision"
 version = "0.1.0"
 authors = ["Marius Gripsgard <mariogrip@ubuntu.com>"]
 
 [dependencies]
 log = "*"
+rustc_version = "= 0.1.2"
+rustc_version_runtime = "0.1.0"
--- a/integration_tests/snaps/simple-rust-with-revision/Cargo.toml
+++ b/integration_tests/snaps/simple-rust-with-revision/Cargo.toml
@@ -1,0 +1,7 @@
+[package]
+name = "simple-rust"
+version = "0.1.0"
+authors = ["Marius Gripsgard <mariogrip@ubuntu.com>"]
+
+[dependencies]
+log = "*"
--- a/integration_tests/snaps/simple-rust-with-revision/snapcraft.yaml
+++ b/integration_tests/snaps/simple-rust-with-revision/snapcraft.yaml
@@ -1,0 +1,13 @@
+name: test-rust-on-revision
+version: 0.1
+summary: A simple rust project using a specific rust revision.
+description: |
+  Build a simple project tied to a rust revision defined in the
+  plugin's schema.
+confinement: strict
+
+parts:
+  simple-rust:
+    plugin: rust
+    rust-revision: 1.13.0
+    source: .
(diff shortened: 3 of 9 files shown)


integration_tests/test_rust_plugin.py
```diff
@@ -29,6 +29,14 @@ def test_stage_rust_plugin(self):
             os.path.join('stage', 'bin', 'simple-rust'), cwd=project_dir)
         self.assertEqual('There is rust on snaps!\n', binary_output)
 
+    def test_stage_rust_with_revision(self):
+        project_dir = 'simple-rust-with-revision'
+        self.run_snapcraft('stage', project_dir)
+
+        binary_output = self.get_output_ignoring_non_zero_exit(
+            os.path.join('stage', 'bin', 'simple-rust'), cwd=project_dir)
+        self.assertEqual('There is rust on snaps!\n', binary_output)
```
elopio: This is not checking that we compiled with the right rust.
I think this would be better: http://paste.ubuntu.com/23758746/

integration_tests/snaps/simple-rust-with-revision/snapcraft.yaml
```diff
@@ -0,0 +1,13 @@
+name: test-rust-on-revision
```
elopio: I found this week that I dislike that we are calling everything simple-*
Instead, it would be easier to find things if we called them plugin-*
Can you rename this dir to something like integration_tests/snaps/rust-with-revision ?

integration_tests/snaps/simple-rust-with-revision/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "simple-rust"
```
elopio: s/simple-rust/rust-with-revision

integration_tests/snaps/simple-rust-with-revision/snapcraft.yaml
sergiusens: sure, I just dislike inconsistencies more :-P

integration_tests/snaps/simple-rust-with-revision/Cargo.toml
sergiusens: sure

integration_tests/snaps/simple-rust-with-revision/snapcraft.yaml
elopio: I can fix them all next week. Feel free to leave it, one more, one less, no big difference :)

integration_tests/snaps/rust-with-revision/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "rust-with-revision"
+version = "0.1.0"
+authors = ["Marius Gripsgard <mariogrip@ubuntu.com>"]
+
+[dependencies]
+log = "*"
+rustc_version = "= 0.1.2"
+rustc_version_runtime = "0.1.0"
```
elopio: ouch, they fixed the bug I reported a lot faster than what I expected:
https://github.com/seppo0010/rustc-version-runtime-rs/issues/1

From the few bits I learned from cargo this morning, I think that will make it break.
You should change the two lines above for this one:
rustc_version_runtime = "0.1.3"

elopio:
@sergiusens beagle board classic explodes when I try to run this example.
Merge it as is, and I'll make a PR trying rust 1.12.

PR #1038, opened 2017-01-07
misc: delete bzr ignore

elopio:
LP: [#1654619](https://bugs.launchpad.net/snapcraft/+bug/1654619)


elopio:
oh, I didn't notice that. Thanks @sergiusens.
I used the git commands, but the diff doesn't seem to understand that it was a rename.

PR #1039, opened 2017-01-08
Check the older ubuntu-core snap name for core dynamic linker.

thomir:
Fixes LP: [#1650946](https://bugs.launchpad.net/snapcraft/+bug/1650946) for those of us who have the older 'ubuntu-core' snap installed.

### DIFF
--- a/snapcraft/_options.py
+++ b/snapcraft/_options.py
@@ -168,18 +168,20 @@ class ProjectOptions:
         However if core is not installed None will be returned.
         """
         core_path = os.path.join('/snap', 'core', 'current')
+        fallback_core_path = os.path.join('/snap', 'ubuntu-core', 'current')
         core_dynamic_linker = self.__machine_info.get('core-dynamic-linker',
                                                       'lib/ld-linux.so.2')
 
-        try:
-            dynamic_linker_resolved_path = os.readlink(
-                os.path.join(core_path, core_dynamic_linker))
-            dynamic_linker_path = os.path.join(
-                core_path, dynamic_linker_resolved_path.lstrip('/'))
-        except FileNotFoundError:
-            dynamic_linker_path = None
+        for lookup_path in (core_path, fallback_core_path):
+            try:
+                dynamic_linker_resolved_path = os.readlink(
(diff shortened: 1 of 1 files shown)


elopio:
@sergiusens can you just quickly confirm that the new ubuntu-core will work the same for classic snaps?

kyrofa:
Yeah make sure @sergiusens takes a look, [he seemed to only want core](https://github.com/snapcore/snapcraft/pull/949#discussion_r91182679).

sergiusens:
I am sorry to say this won't work moving forward, the requirement is to have `core` only, `ubuntu-core` is deprecated and going away and once it does, all these snaps will just fail to work.

Thanks for the PR

thomir:
I'm confused - which snaps will fail to work? Surely this is something that has an effect at the build-time of snaps, not at the run-time?

By all means close the PR, especially if there's a migration path coming for people who have 'ubuntu-core' installed. If there's no migration path however, snapcraft becomes unusable to everyone in this situation.

Can you please elaborate on what your plan is to fix this?

sergiusens:
@thomir runtime is also affected, the linker is forced to be the one in the snap at runtime as well

```
$ readelf -a /snap/vim/current/bin/vim |grep INTERP -A3
  INTERP         0x0000000000000238 0x0000000000400238 0x0000000000400238
                 0x0000000000000033 0x0000000000000033  R      1
      [Requesting program interpreter: /snap/core/current/lib/x86_64-linux-gnu/ld-2.23.so]
  LOAD           0x0000000000000000 0x0000000000400000 0x0000000000400000
```

niemeyer:
The proper fix for this will be worked on over the next few days. We'll migrate ubuntu-core to core.

kyrofa:
> By all means close the PR, especially if there's a migration path coming for people who have 'ubuntu-core' installed. If there's no migration path however, snapcraft becomes unusable to everyone in this situation.

Last I heard the snapd folks were working on a migration plan.

thomir:
Great, thanks @niemeyer !

PR #1052, opened 2017-01-17
Make git pulls less error prone due to history changes.

josepht:
Bugs/1614520


snapcraft/internal/sources/_git.py
```diff
@@ -49,6 +49,11 @@ def _pull_existing(self):
...
             refspec = self.source_commit
 
+        subprocess.check_call([self.command, '-C', self.source_dir,
+                               'fetch', '--recurse-submodules=yes'])
+        subprocess.check_call([self.command, '-C', self.source_dir,
+                               'reset', '--hard', refspec])
+
         # Pull changes to this repository and any submodules.
         subprocess.check_call([self.command, '-C', self.source_dir,
                                'pull', '--recurse-submodules=yes',
```
kyrofa: Wait wait, why are you pulling after the fetch/reset?

josepht: Does the reset recurse over the submodules?  I think I'll need to do something along the lines of what is suggested here: http://stackoverflow.com/a/17871677

kyrofa: Hmm, that's probably a good call, but some good testing is in order.

josepht: Updated and test added.

sergiusens:
leaving the call to you as you are going to do repull/rebuild

snapcraft/internal/sources/_git.py
```diff
@@ -49,14 +53,18 @@ def _pull_existing(self):
...
         subprocess.check_call([self.command, '-C', self.source_dir,
-                               'pull', '--recurse-submodules=yes',
-                               self.source, refspec])
+                               'reset', '--hard', reset_spec], **self.kwargs)
 
         # Merge any updates for the submodules (if any).
         subprocess.check_call([self.command, '-C', self.source_dir,
-                              'submodule', 'update'])
+                              'submodule', 'update', '--remote'],
+                              **self.kwargs)
```
kyrofa: Should this include `--recursive` to recursively make sure submodules within submodules are up-to-date?

josepht: Updated

snapcraft/tests/sources/test_git.py
```diff
@@ -161,3 +174,126 @@ def test_init_with_source_tag_and_commit_raises_exception(self):
...
+            shutil.rmtree(dir)
+
+    def clean_dir(self, dir):
+        self.rm_dir(dir)
+        os.mkdir(dir)
+        self.addCleanup(self.rm_dir, dir)
+
+    def clone_repo(self, repo, tree):
+        self.clean_dir(tree)
+        self.call(['git', 'clone', repo, tree])
```
sergiusens: this missed a `Build-Depends` on `git` in `debian/control` :cry:

PR #1053, opened 2017-01-17
meta: ensure snap.yaml is desktop free

sergiusens:
snap.yaml has no concept of desktop files, this
is a by-convention mechanism in snap.

LP: #1656487

### DIFF
--- a/snapcraft/internal/meta.py
+++ b/snapcraft/internal/meta.py
@@ -90,7 +90,7 @@ class _SnapPackaging:
         self._arch_triplet = project_options.arch_triplet
 
         self._meta_dir = os.path.join(self._snap_dir, 'meta')
-        self._config_data = config_data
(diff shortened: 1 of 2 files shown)


snapcraft/internal/meta.py
```diff
@@ -336,8 +337,9 @@ def _reformat_desktop(self, name, app):
                 else:
                     logger.warning(
                         'Icon {} specified in desktop file {} not found '
-                        'in prime directory'.format(icon, app['desktop']))
+                        'in prime directory'.format(icon, desktop_file))
         target = os.path.join(gui_dir, os.path.basename(desktop_file))
+        print('target', target)
```
kyrofa: Should this be here?

sergiusens: oh noes!

PR #1057, opened 2017-01-19
godeps plugin: support for go-packages.

sergiusens:
This is the on-par support for what is in the go plugin.
LP: #1657212


integration_tests/snaps/godeps-with-go-packages/snapcraft.yaml
```diff
@@ -0,0 +1,14 @@
+name: test-package
```
kyrofa: This doesn't seem to be using the dependencies.tsv and main.go copied in from `simple-godeps` (they're used in the new `godeps` snap instead). Should they be removed from here? Note also that the built .snap version of this was committed.

integration_tests/test_godeps_plugin.py
```diff
@@ -31,15 +32,44 @@ class GodepsPluginTestCase(testscenarios.WithScenarios,
...
+        hash_command = [binary, 'hash', '10', 'password']
+
+        output = subprocess.check_output(
+            hash_command, cwd=project_dir)
+
+        check_hash_command = [binary, 'check', output, 'password']
+        output = subprocess.check_output(
+            check_hash_command, cwd=project_dir)
+
+        self.assertEqual('Equal', output.decode('UTF-8').strip(' \n'))
```
kyrofa: These functions have the actual test in common, they just set it up a little differently, right? Perhaps the assertions should be moved into a helper function so you don't need to make two changes.

snapcraft/plugins/godeps.py
```diff
@@ -61,13 +70,28 @@ def schema(cls):
...
         # The import path must be specified.
         schema['required'].append('go-importpath')
 
         return schema
 
     @classmethod
+    def get_build_properties(cls):
+        # Inform Snapcraft of the properties associated with building. If these
+        # change in the YAML Snapcraft will consider the build step dirty.
+        return ['go-packages']
```
kyrofa: The `go` plugin grabs these in the `pull` step. Is there a reason we're adding it to the `build` step here? I suspect that'll fail on LP since internet isn't available during `build`.

integration_tests/snaps/godeps-with-go-packages/snapcraft.yaml
sergiusens: That is the exact purpose of the test, being there and not being built

sergiusens: that is in the description a few lines below btw ;-)

integration_tests/test_godeps_plugin.py
sergiusens: ok

integration_tests/snaps/godeps-with-go-packages/snapcraft.yaml
kyrofa: Ah ha! Fair enough. Though note in the description, s/unsing/using/.

snapcraft/plugins/godeps.py
kyrofa: Ah, I see now, okay-- that makes sense, thanks for clarifying :smiley: .

snapcraft/plugins/godeps.py
```diff
@@ -28,12 +28,11 @@
 
     - go-packages:
       (list of strings)
-      Go packages to fetch, these must be a "main" package. Dependencies
-      are pulled in automatically by `go get`.
+      Go packages to build/install, these must be a "main" package.
+      Dependencies should have already been retrieved by the `godeps-file`
+      used for this part.
```
kyrofa: Yeah that makes it a bit more clear.

PR #1059, opened 2017-01-20
pluginhandler: support more complex stage-packages.

kyrofa:
Currently snapcraft only supports a flat list of stage packages to be staged regardless of target architecture. This PR resolves LP: [#1637282](https://bugs.launchpad.net/snapcraft/+bug/1637282) by introducing a more complex grammar that allows one to filter stage packages depending on various selectors (target arch for right now), as well as specify optional packages.

The grammar is made up of two statements: `on` and `try`.

```yaml
- on <selector>[,<selector>...]:
  - ...
- else[ fail]:
  - ...
```

The body of the `on` clause is taken into account if every (AND, not OR) selector is true for the target build environment. Currently the only selectors supported are target architectures (e.g. amd64).

If the `on` clause doesn't match and it's immediately followed by an `else` clause, the `else` clause must be satisfied. An `on` clause without an `else` clause is considered satisfied even if no selector
matched. The `else fail` form allows erroring out if an `on` clause was not matched.

```yaml
- try:
  - ...
- else:
  - ...
```

The body of the `try` clause is taken into account only when all packages contained within it are valid. If not, if it's immediately followed by `else` clauses they are tried in order, and one of them must be satisfied. A `try` clause with no `else` clause is considered satisfied even if it contains invalid packages.

This PR is larger than I'd like it to be, I'm sorry about that-- I didn't see a logical place to split. Lots of the line count is tests.

Note also that I tried to test a number of combinations, but the grammar is recursive, allowing for unlimited combinations. This could use some good poking.


sergiusens:
interesteningly this is missing documentation for `snapcraft help plugins` and general ones in docs

sergiusens:
... with examples

sergiusens:
@niemeyer mind giving this a semantic/sytax review

sergiusens:
btw, you seem to have change the style of your docsrtrings, not sure why https://github.com/snapcore/snapcraft/blob/master/snapcraft/file_utils.py#L33

kyrofa:
> btw, you seem to have change the style of your docsrtrings, not sure why [...]

Yeah, I'm still learning. Item 49 of Effective Python gave this as the recommended style... I'm not actually sure where I got the previous style. I'm totally open to suggestions though, we don't use a tool to process these so it's really up to us. Opinions?

kyrofa:
(I'll wait for the semantics review before updating docs and making a PR to snappy-docs).

niemeyer:
The logic described above in the PR description sounds like what we agreed upon.

Haven't looked at the code in detail, but one issue to watch out for might be the spacing.. can we please ensure the clause enforces the lack of spaces between the comma-separated entries?  ("ubuntu,amd64", not "ubuntu,<space>amd64"). This allows us to extend the format later in a few different ways, and also make it more readable. Not bad in the short term either.

kyrofa:
> can we please ensure the clause enforces the lack of spaces between the comma-separated entries?

Indeed, that is not enforced here. I'll fix that.

kyrofa:
>> can we please ensure the clause enforces the lack of spaces between the comma-separated entries?

> Indeed, that is not enforced here. I'll fix it.

Done. I also updated docs here.

snapcraft/internal/pluginhandler/stage_package_grammar/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from ._errors import StagePackageSyntaxError  # noqa
```
sergiusens: the pattern we've been following for all packages is to call the module `errors` (instead of `_errors`) and not automatically import all its symbols.

snapcraft/internal/pluginhandler/stage_package_grammar/_on.py
```diff
@@ -0,0 +1,128 @@
...
+    >>> with tempfile.TemporaryDirectory() as cache_dir:
+    ...     ubuntu = repo.Ubuntu(cache_dir)
+    ...     options = ProjectOptions(target_deb_arch='i386')
+    ...     clause = OnStatement('on amd64', ['foo'], options, ubuntu)
+    ...     clause.add_else(['bar'])
+    ...     clause.process()
+    {'bar'}
+    """
+
+    def __init__(self, on, body, project_options, ubuntu):
```
sergiusens: can you please to

```python
def __init__(self, *, on, body, project_options, ubuntu):
```

or at least

```python
def __init__(self, on, *, body, project_options, ubuntu):
```

personally, the former feels better given the class name

snapcraft/internal/pluginhandler/stage_package_grammar/_on.py
```diff
@@ -0,0 +1,128 @@
...
+        # selectors are matched with an AND, not OR, there should only be one
+        # selector.
+        if (len(self.selectors) == 1) and (target_arch in self.selectors):
+            packages = process_grammar(
+                self._body, self._project_options, self._ubuntu)
+        else:
+            for else_body in self._else_bodies:
+                if not else_body:
+                    # Handle the 'else fail' case.
+                    raise EnvironmentError(
```
sergiusens: can we get a specific exception?

snapcraft/internal/pluginhandler/stage_package_grammar/_on.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def __init__(self, on, body, project_options, ubuntu):
+        """Create an _OnStatement instance.
+
+        Arguments:
+            on: String containing the 'on <selectors>' part of the clause.
+            body: List containing the body of the 'on' clause.
+            project_options: Instance of ProjectOptions to use to process
+                             clause.
+            ubuntu: repo.Ubuntu instance used for checking package validity.
```
sergiusens: for future proofing, can this be named `repo_instance` ?

snapcraft/internal/pluginhandler/stage_package_grammar/_processor.py
```diff
@@ -0,0 +1,138 @@
...
+    for section in grammar:
+        try:
+            if else_fail_pattern.match(section):
+                _handle_else(statement, None)
+            else:
+                # If it's just a string, add it to the package set.
+                packages.add(section)
+        except TypeError:
+            # Alright, it wasn't an unconditional stage package, so it must be
+            # a dict.
```
sergiusens: why not make use of `isisnstace`? Feels like there might be less nesting.

snapcraft/internal/pluginhandler/stage_package_grammar/_processor.py
```diff
@@ -0,0 +1,138 @@
...
+            for key, value in section.items():
+                if on_clause_pattern.match(key):
+                    # We've come across the begining of an 'on' statement.
+                    # That means any previous statement we found is complete.
+                    statements.add(statement)
```
sergiusens: can you walk me throught this? The first match to occur will add a `satement = None` and then redefine it below?

snapcraft/internal/pluginhandler/stage_package_grammar/_on.py
```diff
@@ -0,0 +1,128 @@
...
+                        'unable to satisfy {!r}, failure forced'.format(self))
+
+                packages = process_grammar(
+                    else_body, self._project_options, self._ubuntu)
+                if packages:
+                    break
+
+        return packages
+
+    def __eq__(self, other):
```
sergiusens: is this used anywhere?

snapcraft/internal/parts.py
```diff
@@ -270,7 +270,8 @@ def load_plugin(self, part_name, plugin_name, part_properties):
             plugin_name=plugin_name,
             part_properties=part_properties,
             project_options=self._project_options,
-            part_schema=self._validator.part_schema)
+            part_schema=self._validator.part_schema,
+            definitions_schema=self._validator.definitions_schema)
```
kyrofa: Yeah definitely. I didn't actually realize those replacements could be recursive (I see that now). Want me to try that? Or would we rather move to all json in the future, in which case this would need to stay?

snapcraft/internal/pluginhandler/_stage_package_handler.py
```diff
@@ -0,0 +1,100 @@
...
+    ...    handler = StagePackageHandler(['foo'], cache_dir)
+    ...    handler.fetch()
+    ...    handler.unpack(unpack_dir)
+    """
+
+    def __init__(self, stage_packages_grammar, cache_dir, *, sources=None,
+                 project_options=None):
+        """Create new StagePackageHandler.
+
+        Arguments:
```
kyrofa: I asked @elopio about it, he said to leave it until he had done his research since other docstrings would need to change as well.

@elopio is now a good time to change these?

snapcraft/internal/pluginhandler/stage_package_grammar/_on.py
kyrofa: Yeah, right [here](https://github.com/snapcore/snapcraft/pull/1059/files#diff-68c6d965d4ff21ecd61544afbc4bb051R121). It's what causes an error if the grammar contains two `on` statements with the same selectors at the same level.

snapcraft/internal/pluginhandler/stage_package_grammar/_processor.py
kyrofa: Honestly I'd prefer to use `isinstance`, but I thought duck typing was more pythonic. Happy to change.

snapcraft/internal/pluginhandler/stage_package_grammar/_processor.py
kyrofa: Let's use this as an example:

```
- on amd64:
  - foo
- try:
  - bar
- else:
  - baz
```
`process_grammar()` is called with this entire thing, and it walks through the top level list (`on`, `try`, `else`).

1. **Parsing `on amd64`**
    Since `statement` is `None` initially, yes, when it hits the `on` it adds `None` to the `statements` collection (which will [ignore it](https://github.com/snapcore/snapcraft/pull/1059/files#diff-68c6d965d4ff21ecd61544afbc4bb051R119)). Then it creates a new `OnStatement` and holds onto it.
2. **Parsing `try`**
    Since we [hit a new statement](https://github.com/snapcore/snapcraft/pull/1059/files#diff-68c6d965d4ff21ecd61544afbc4bb051R68) we know the `OnStatement` we were holding onto is complete, so it's added to the collection. It then creates a new `TryStatement` and holds onto it.
3. **Parsing `else`**
    `else` clauses mean nothing by themselves; they [must be attached to an `OnStatement` or `TryStatement`](https://github.com/snapcore/snapcraft/pull/1059/files#diff-68c6d965d4ff21ecd61544afbc4bb051R95), otherwise their presence is an error. Since we already have a `TryStatement`, it gets attached to that.
4. **Parsing is complete**
    But we're still holding onto that `TryStatement`. [Add it to the collection](https://github.com/snapcore/snapcraft/pull/1059/files#diff-68c6d965d4ff21ecd61544afbc4bb051R77).
5. **[Process all statements in collection](https://github.com/snapcore/snapcraft/pull/1059/files#diff-68c6d965d4ff21ecd61544afbc4bb051R78)**

sergiusens: can we then add a comment for each `statement -> None` potential action as a code comment and I guess this is to solve the initial value in a loop problem.

snapcraft/internal/pluginhandler/stage_package_grammar/_on.py
sergiusens: thanks

snapcraft/internal/parts.py
sergiusens: let's use yaml, I don't see us changing this any time soon. If you want to do the work, yes please.

docs/snapcraft-syntax.md
```diff
@@ -82,8 +82,43 @@ contain.
...
+          - try:
+            - ...
+          - else:
+            - ...
+
+      The body of the 'try' clause is taken into account only when all packages
+      contained within it are valid. If not, if it's immediately followed by
+      'else' clauses they are tried in order, and one of them must be satisfied.
+      A 'try' clause with no 'else' clause is considered satisfied even if it
+      contains invalid packages.
```
elopio: wow, this is hard. I think I got it, but I will try to avoid it for sure.

elopio: no, scratch that. I didn't get what are the selectors.

integration_tests/snaps/stage-package-grammar/snapcraft.yaml
```diff
@@ -0,0 +1,53 @@
...
+        - foo
+      - else:
+        - hello
+
+  on-other-arch-else-fail:
+    plugin: nil
+    stage-packages:
+      - on other-arch:
+        - foo
+      - else fail
```
elopio: you are missing a scenario with multiple selectors.

snapcraft/internal/pluginhandler/_stage_package_handler.py
elopio: I'm +1 on sphinx docstrings and found no nicer alternative. So yes, please.

snapcraft/internal/pluginhandler/stage_package_grammar/_processor.py
elopio: +1 isinstance.

elopio:
@caldav, can you please review the documentation?

snapcraft/internal/parts.py
kyrofa: Unfortunately jsonschema explodes with a `RecursionError` when the YAML contains recursive anchors, so it seems that using jsonschema refs appears to be our only option.

snapcraft/internal/pluginhandler/stage_package_grammar/_on.py
kyrofa: Sure, though may I ask why? Requiring optional arguments to be keyword-only makes sense, but in this case they aren't optional. Are you saying you like keyword-only arguments always?

sergiusens: yes, I like keyword arguments always so when I read code that uses this class or method I don't need to read the signature/definition. I think it is also a recommendation in effective python, where it makes it easier to add keywords later without having to have a massive refactor.

kyrofa: Alright, I'm making this change now. By the way, yes, Item 21 does indeed recommend using keyword-only arguments to enforce clarity, but note that he doesn't necessarily recommend doing it for all arguments, just ones that aren't clear (e.g. his example only made two out of four arguments keyword-only as they were both booleans and "it's easy to confuse the position of the two Boolean arguments").

sergiusens: I am taking it to the extreme, the thing with positional is that the first positional argument is the one with more weight, but what if we come up with something that should change the order to make it feel appropriate, or would you way that any order here matters?

To be fair, I sort of also brought this with me from `golint` and initializing `sruct` s

kyrofa: True, keyword-only removes any ordering from the function call, which is nice, but of course has the downside of requiring more typing.  What you're suggesting, though, has broader implications. Should we start using keyword-only in every function, then? Not even golang is that verbose :wink: . Given that you're bringing this from `golint` and `struct`s, do you really feel this way only about `__init__()`?

integration_tests/snaps/stage-package-grammar/snapcraft.yaml
kyrofa: Such a scenario is currently impossible, since:

1. Selectors are matched with an AND, not an OR
2. The only selector currently supported is the architecture.

docs/snapcraft-syntax.md
kyrofa: Done, though I did it in `snapcraft/__init__.py`

snapcraft/__init__.py
```diff
@@ -144,13 +144,45 @@
...
+          - ...
+        - else:
+          - ...
+
+    The body of the 'try' clause is taken into account only when all packages
+    contained within it are valid. If not, if it's immediately followed by
+    'else' clauses they are tried in order, and one of them must be satisfied.
+    A 'try' clause with no 'else' clause is considered satisfied even if it
+    contains invalid packages.
+
```
kyrofa: Done.

snapcraft/internal/pluginhandler/_stage_package_handler.py
kyrofa: Done.

snapcraft/internal/pluginhandler/stage_package_grammar/__init__.py
kyrofa: No problem-- done.

snapcraft/internal/pluginhandler/stage_package_grammar/_on.py
kyrofa: Done.

snapcraft/internal/pluginhandler/stage_package_grammar/_on.py
kyrofa: Good call, done.

snapcraft/internal/pluginhandler/stage_package_grammar/_on.py
kyrofa: Definitely, done.

snapcraft/internal/pluginhandler/stage_package_grammar/_processor.py
kyrofa: Done.

snapcraft/internal/pluginhandler/stage_package_grammar/_processor.py
kyrofa: Done.

elopio:
When you separate the on clause with a comma and a space, the error doesn't give a clue of how to fix it:

Invalid syntax for stage packages: 'on other-arch, foo' is not a valid 'on' clause

kyrofa:
> When you separate the on clause with a comma and a space, the error doesn't give a clue of how to fix it

Yeah I took the easy way out: it's regex. In order for me to give helpful errors I'll need to split the regex into a number of manual checks. Would you prefer that?

elopio:
Yes, I think this is hard and users will need as much help as possible.
But maybe you can just check if there's ',\s' and suggest to remove the space.

elopio:
Is this a valid construct?

  try-else:
    plugin: nil
    stage-packages:
      - try:
        - on other-arch:
          - foo
        - else
      - else:
        - foo
      - else:
        - bar
      - else fail

I don't understand what it should do, but it doesn't fail.

kyrofa:
> I don't understand what it should do, but it doesn't fail.

Indeed: `bar` is a valid package.

elopio:
sorry for the wrong formatting. So what does it mean when one puts multiple else clauses?

elopio:
This is not likely introduced by this PR, so feel free to report a bug instead of fixing it. But I think we should validate the package names. They shouldn't have spaces:

```
  on-other-arch-else-fail:
    plugin: nil
    stage-packages:
      - on other-arch:
        - foo
      - else faill
```

Error downloading stage packages for part 'on-other-arch-else-fail': The Ubuntu package 'else faill' was not found.

kyrofa:
> Error downloading stage packages for part 'on-other-arch-else-fail': The Ubuntu package 'else faill' was not found.

| Name | Score |
|----------|-------------|
| Kyle's regex skills | 1 |
| Leo | 0 |

Agreed, but this is a separate issue.

kyrofa:
> So what does it mean when one puts multiple else clauses?

They're tried in order until one of them provides usable packages.

kyrofa:
> Yes, I think this is hard and users will need as much help as possible.

Done. I added a more helpful message if there are no selectors, as well. Note that regexy things (multiple commas, starting/ending with commas, etc.) I left to be handled by the regex. I really don't want to write regex parsing by hand :stuck_out_tongue: .

elopio:
Thanks :)
I'm ok with the generic message for some errors. It will be easier to notice that you are starting with a comma, that's usually wrong, than to notice that you have an extra space, which is usually ok.

kalikiana:
How is other-arch going to be validated? Literal arch names may not always be unique. A part or package name could be other-arch. In the future even host versus target arch could be desirable. Same with other selectors, it will be difficult to tell a defined selector from an arch that is accepted as a selector implicitly.

kyrofa:
> A part or package name could be other-arch.

That's won't hurt anything, since "on other-arch" won't be matched up to either.

> Same with other selectors, it will be difficult to tell a defined selector from an arch that is accepted as a selector implicitly. 

I'm not sure I understand, could you give me an example?

sergiusens:
I think this means to say something like

what if a distro name collides with an arch? In the sense that we need namespacing maybe (future, implicit `arch:`)... i.e.; `on arch:amd64 os:armhf` (if an `os` called `armhf` existed).

This is not a problem right now and we can namespace as soon as we introduce a new selector type.

paroj:
wouldnt it make sense to have this for `build-packages` as well?

PR #1060, opened 2017-01-20
options: use actual userspace architecture for building in 32bit

3v1n0:
It might happen in 32bit containers running on 64bit hosts, that
they are detected as 64bit systems since platform's machine()
returns the kernel's architecture, not the userspace one.

Using `gcc -v` `Target:`'s value instead

LP: [#1641123](https://bugs.launchpad.net/snapcraft/+bug/1641123)

### DIFF
--- a/snapcraft/plugins/gulp.py
+++ b/snapcraft/plugins/gulp.py
@@ -81,7 +81,7 @@ class GulpPlugin(snapcraft.BasePlugin):
         super().__init__(name, options, project)
         self._npm_dir = os.path.join(self.partdir, 'npm')
         self._nodejs_tar = sources.Tar(nodejs.get_nodejs_release(
-            self.options.node_engine), self._npm_dir)
+            self.options.node_engine, self.project), self._npm_dir)
 
     def pull(self):
         super().pull()
--- a/snapcraft/plugins/gulp.py
+++ b/snapcraft/plugins/gulp.py
@@ -81,7 +81,7 @@ class GulpPlugin(snapcraft.BasePlugin):
         super().__init__(name, options, project)
         self._npm_dir = os.path.join(self.partdir, 'npm')
         self._nodejs_tar = sources.Tar(nodejs.get_nodejs_release(
-            self.options.node_engine, self.project), self._npm_dir)
+            self.options.node_engine, self.project.deb_arch), self._npm_dir)
 
     def pull(self):
         super().pull()
--- a/snapcraft/plugins/nodejs.py
+++ b/snapcraft/plugins/nodejs.py
@@ -107,7 +107,7 @@ class NodePlugin(snapcraft.BasePlugin):
         super().__init__(name, options, project)
         self._npm_dir = os.path.join(self.partdir, 'npm')
         self._nodejs_tar = sources.Tar(get_nodejs_release(
-            self.options.node_engine, self.project), self._npm_dir)
+            self.options.node_engine, self.project.deb_arch), self._npm_dir)
 
     def pull(self):
         super().pull()
@@ -140,8 +140,7 @@ class NodePlugin(snapcraft.BasePlugin):
             self.run(['npm', 'run', target], cwd=rootdir)
 
 
-def _get_nodejs_base(node_engine, project_options):
-    machine = project_options.deb_arch
+def _get_nodejs_base(node_engine, machine):
     if machine not in _NODEJS_ARCHES:
         raise EnvironmentError('architecture not supported ({})'.format(
             machine))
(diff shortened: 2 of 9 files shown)


kyrofa:
Looks like some static test failures here.

adconrad:
You can (and should) skip the ^Target regex entirely here by using 'gcc -dumpmachine' instead of parsing the output of 'gcc -v'

It's worth noting, however, that triplets will differ between distros.  Not everyone uses the upstream defaults.  Some may say i386-linux-gnu instead of i686, some may include the vendor tag (ie: i686-redhat-linux-gnu), some alter the machine instead of the ABI (armhl-linux-gnu instead of arm-linux-gnueabihf).

That's not to say this approach is wrong, but there are lots of small corner cases you'll have to cater to in a switch to get it right for all your target platforms.  Thankfully, logging in to a bunch of different distros and testing this out probably isn't super hard, but it's certainly annoying.

3v1n0:
@adconrad mh, I've actually followed what it was written in the [related bug](https://bugs.launchpad.net/snapcraft/+bug/1641123) as it's what @sergiusens also suggested; my first implementation was actually using [a different approach (see commit)](https://github.com/snapcore/snapcraft/pull/1060/commits/df83f4364fd05de71a9a61e46f3fec734218d8a5) which was using python's platform architecture (which checks for the arch of a binary file - `python` by default), and that's probably more universal than this then.

I can go back to that way instead... To me checking the python binary type that has been launched seems pretty sane, but.... Indeed there could be still many corner cases.

snapcraft/plugins/nodejs.py
```diff
@@ -142,7 +141,7 @@ def _npm_install(self, rootdir):
 
 
 def _get_nodejs_base(node_engine):
-    machine = platform.machine()
+    machine = _options._get_platform_architecture()
```
sergiusens: hmm, can't you pass `project_options.deb_arch` and change the map to use `deb_arch` instead?

snapcraft/plugins/nodejs.py
```diff
@@ -140,15 +140,15 @@ def _npm_install(self, rootdir):
             self.run(['npm', 'run', target], cwd=rootdir)
 
 
-def _get_nodejs_base(node_engine):
-    machine = _options._get_platform_architecture()
+def _get_nodejs_base(node_engine, project_options):
```
sergiusens: this should recieve `machine`

snapcraft/plugins/nodejs.py
```diff
@@ -107,7 +107,7 @@ def __init__(self, name, options, project):
         super().__init__(name, options, project)
         self._npm_dir = os.path.join(self.partdir, 'npm')
         self._nodejs_tar = sources.Tar(get_nodejs_release(
-            self.options.node_engine), self._npm_dir)
+            self.options.node_engine, self.project), self._npm_dir)
```
sergiusens: you should pass `self.project.debarch`

snapcraft/plugins/gulp.py
```diff
@@ -81,7 +81,7 @@ def __init__(self, name, options, project):
         super().__init__(name, options, project)
         self._npm_dir = os.path.join(self.partdir, 'npm')
         self._nodejs_tar = sources.Tar(nodejs.get_nodejs_release(
-            self.options.node_engine), self._npm_dir)
+            self.options.node_engine, self.project), self._npm_dir)
```
sergiusens: you should pass `self.project.debarch`

snapcraft/plugins/nodejs.py
3v1n0: Ah, ok... I thought you meant to pass the whole project in the previous req...

adconrad:
I think the "just ask python" method seems cleaner.  Yes, it won't work in cross-build cases, but those are explicitly not catered for here and, really, the only sane way to cross is to explicitly request an arch via a flag, not hope you can guess implicitly.

In the dpkg world, dpkg-dev will fall back to 'gcc -dumpmachine' (which might be where people got the idea from) if there's no dpkg binary around to talk to, but that's a worst case bootstrapping scenario.  Usually, we have an actual dpkg binary to ask about itself, and we trust the answer.  The analog here of trusting python seems like the correct one, and perhaps making sure you have the hooks in place for a future explicit passing of arch for crossing.

adconrad:
(And sorry if any of my comments along the way ended up adding to the confusion instead of helping resolve it)

3v1n0:
@adconrad no worries, I'm happy we ended up with this decision too...

> making sure you have the hooks in place for a future explicit passing of arch for crossing.

Well the `--target-arch` is already there and it's exactly for this reason I believe.

PR #1061, opened 2017-01-20
Fix Exec key in desktop files for apps named like their parent package (LP: #1658123).

oSoMoN:
https://bugs.launchpad.net/snapcraft/+bug/1658123

### DIFF
--- a/snapcraft/internal/meta.py
+++ b/snapcraft/internal/meta.py
@@ -305,20 +305,18 @@ class _SnapPackaging:
         if desktop_file_name:
             desktop_file = _DesktopFile(
                 name=name, filename=desktop_file_name,
-                snap_name=self._config_data['name'], prime_dir=self._snap_dir,
-                gui_dir=os.path.join(self.meta_dir, 'gui'))
+                snap_name=self._config_data['name'], prime_dir=self._snap_dir)
             desktop_file.parse_and_reformat()
-            desktop_file.write()
+            desktop_file.write(gui_dir=os.path.join(self.meta_dir, 'gui'))
 
 
 class _DesktopFile:
 
-    def __init__(self, *, name, filename, snap_name, prime_dir, gui_dir):
+    def __init__(self, *, name, filename, snap_name, prime_dir):
         self._name = name
         self._filename = filename
         self._snap_name = snap_name
         self._prime_dir = prime_dir
-        self._gui_dir = gui_dir
         self._path = os.path.join(prime_dir, filename)
         if not os.path.exists(self._path):
             raise EnvironmentError(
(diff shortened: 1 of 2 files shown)


snapcraft/internal/meta.py
```diff
@@ -325,7 +325,9 @@ def _reformat_desktop(self, name, desktop_file):
                 'The specified desktop file {!r} is missing the '
                 '"Exec" key'.format(desktop_file))
         # XXX: do we want to allow more parameters for Exec?
-        exec_value = '{}.{} %U'.format(self._config_data['name'], name)
+        snap_name = self._config_data['name']
+        exec_value = '{} %U'.format(
+            name if name == snap_name else '{}.{}'.format(snap_name, name))
```
kyrofa: The nested if/else and format cause a double-take. I good old-fashioned if/else would be easier to read.

sergiusens: interesting, this was what mostly bugged me about this PR and queueing it for later :-P

oSoMoN: I initially wrote it as a normal if else clause, but when running the tests flake8 complained that the function was too complex.

sergiusens: maybe split the function in two, one to retrieve a valid desktop file, another to reformat and write or 3 and make a reformat and a separate write.

This could also be a small `class DesktopParser` or similar and moved out of Packaging (this would be the best)

kyrofa: +1 to `DesktopParser`, that will simplify things.

snapcraft/internal/meta.py
```diff
@@ -301,50 +301,68 @@ def _wrap_app(self, name, app):
...
+            desktop_file = _DesktopFile(name, desktop_file_name,
+                                        self._config_data['name'],
+                                        self._snap_dir, self.meta_dir)
+            desktop_file.parse_and_reformat()
+            desktop_file.write()
+
+
+class _DesktopFile:
+
+    def __init__(self, name, filename, snap_name, snap_dir, meta_dir):
```
sergiusens: instead of `meta_dir`, why not pass `gui_dir`?

sergiusens: Can you call `snap_dir` `prime_dir` instead? It is the new name after all :-) (no need to change the old

sergiusens: Last but not least, Can you do
`def __init__(self, *, <rest of args>)`

This will force parametrized calls which makes it easier in the future for future refactoring and reading.

snapcraft/internal/meta.py
```diff
@@ -301,50 +301,68 @@ def _wrap_app(self, name, app):
...
+                name=name, filename=desktop_file_name,
+                snap_name=self._config_data['name'], prime_dir=self._snap_dir,
+                gui_dir=os.path.join(self.meta_dir, 'gui'))
+            desktop_file.parse_and_reformat()
+            desktop_file.write()
+
+
+class _DesktopFile:
+
+    def __init__(self, *, name, filename, snap_name, prime_dir, gui_dir):
```
sergiusens: darn, no I think `filename` and `gui_dir` should just be a parameter in `write` as in
```python
def write(self, *, desktop_path)
```

and
```python
def parse_and_reformat(self, *, desktop_file_path):
```

and removed from here

oSoMoN: Agreed for gui_dir, not so much for filename as it’s being used in all three functions (__init__, parse_and_reformat, write), so I see little point in passing it as a parameter to all of them. Plus, I think it makes more sense semantically speaking to have filename as a parameter of __init__, given that a _DesktopFile instance operates on one given desktop file.

PR #1063, opened 2017-01-20
file_utils: copy symlinks to directories as symlinks.

kyrofa:
This PR fixes LP: [#1658155](https://bugs.launchpad.net/snapcraft/+bug/1658155) by ensuring that `file_utils.link_or_copy_tree()` handles symlinks to directories the same way it handles symlinks to files.

### DIFF
--- a/snapcraft/tests/__init__.py
+++ b/snapcraft/tests/__init__.py
@@ -65,6 +65,36 @@ class IsExecutable:
         return None
 
 
+class LinkExists:
+    """Match if a file path is a symlink."""
+
+    def __init__(self, expected_target=None):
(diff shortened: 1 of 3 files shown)


snapcraft/tests/__init__.py
```diff
@@ -65,6 +65,36 @@ def match(self, file_path):
         return None
 
 
+class LinkExists:
```
elopio: What do you think of naming it LinksTo ?

sergiusens: I thought of changing the name but it follows the `FileExists` matcher

PR #1064, opened 2017-01-20
pluginhandler: collide with directories and symlinks

kyrofa:
Snapcraft's collision detection currently only checks files. This PR fixes LP: [#1658225](https://bugs.launchpad.net/snapcraft/+bug/1658225) by adding directories and symlinks to the list, making sure to catch files that collide with directories, and symlinks that point to different places.

### DIFF
--- a/snapcraft/internal/pluginhandler/__init__.py
+++ b/snapcraft/internal/pluginhandler/__init__.py
@@ -1063,22 +1063,21 @@ def check_for_collisions(parts):
     parts_files = {}
     for part in parts:
         # Gather our own files up
-        part_files, _ = part.migratable_fileset_for('stage')
+        part_files, part_directories = part.migratable_fileset_for('stage')
+        part_contents = part_files | part_directories
 
         # Scan previous parts for collisions
         for other_part_name in parts_files:
-            common = part_files & parts_files[other_part_name]['files']
+            common = part_contents & parts_files[other_part_name]['files']
             conflict_files = []
             for f in common:
                 this = os.path.join(part.installdir, f)
                 other = os.path.join(
                     parts_files[other_part_name]['installdir'],
                     f)
-                if os.path.exists(this) and os.path.exists(other):
-                    if os.path.islink(this) and os.path.islink(other):
-                        continue
-                    if _file_collides(this, other):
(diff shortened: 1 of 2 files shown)


kyrofa:
Note that the test failures here are simply because this fix better exposes why #897 is so necessary. That PR should probably be merged first.

sergiusens:
This depends on a closed PR, please update and reopen when ready

kyrofa:
This is still an issue, but should be solvable now that organize is handled differently. Brought this PR up to date and re-opened.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -1087,10 +1086,37 @@ def check_for_collisions(parts):
...
 
+def _paths_collide(path1: str, path2: str) -> bool:
+    if not (os.path.lexists(path1) and os.path.lexists(path2)):
+        return False
+
+    path1_is_dir = os.path.isdir(path1)
+    path2_is_dir = os.path.isdir(path2)
+    path1_is_link = os.path.islink(path1)
+    path2_is_link = os.path.islink(path2)
+    collide = False
```
sergiusens: instead of this, why not add an 
```python
else:
    collide = False
```
to the end of your conditionals (it would feel more complete).

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -1087,10 +1086,37 @@ def check_for_collisions(parts):
...
+
+    if path1_is_link and path2_is_link:
+        # They're both symlinks; verify they're pointing to the
+        # same place.
+        if os.readlink(path1) != os.readlink(path2):
+            collide = True
+    elif path1_is_link or path2_is_link:
+        collide = True
+    elif path1_is_dir and path2_is_dir:
+        collide = False
```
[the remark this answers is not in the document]
kyrofa: Otherwise we'd end up calling `_file_collides` with directories, which it doesn't expect.

snapcraft/internal/pluginhandler/__init__.py
kyrofa: Done. I actually got rid of `collide` all together.

PR #1066, opened 2017-01-23
store: don't recommend 2FA as it's not available.

kyrofa:
This PR fixes LP: [#1635567](https://bugs.launchpad.net/snapcraft/+bug/1635567) by removing the strong recommendation to use two-factor authentication. Ideally that would simply be made available for everyone to use, but until then, we shouldn't recommend it.

### DIFF
--- a/integration_tests/__init__.py
+++ b/integration_tests/__init__.py
@@ -155,9 +155,6 @@ class StoreTestCase(TestCase):
         process.sendline(email)
         process.expect_exact('Password: ')
         process.sendline(password)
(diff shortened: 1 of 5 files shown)


sergiusens:
This is actually something that needs to be fixed store side, sorry but this would require a discussion on the mailing list to go forward

sergiusens:
closing this until that discussion happens

elopio:
Here's the server side bug: https://bugs.launchpad.net/canonical-identity-provider/+bug/1636645

PR #1068, opened 2017-01-24
tests: use a temporary directory for snaps tests

elopio:
LP: [#1643658](https://bugs.launchpad.net/snapcraft/+bug/1643658)

### DIFF
--- a/snaps_tests/__init__.py
+++ b/snaps_tests/__init__.py
@@ -197,7 +197,7 @@ class SnapsTestCase(testtools.TestCase):
                 'mktemp -d').strip()
             self.addCleanup(
                 self.snappy_testbed.run_command,
-                ['rm', '-r', tmp_in_testbed])
+                ['rm', '-rf', tmp_in_testbed])
             self.snappy_testbed.copy_file(snap_local_path, tmp_in_testbed)
             snap_file_name = os.path.basename(snap_local_path)
             snap_path_in_testbed = os.path.join(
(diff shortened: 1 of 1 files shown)


kyrofa:
Yay! \o/

snaps_tests/__init__.py
```diff
@@ -193,14 +193,15 @@ def install_store_snap(self, snap_name):
     def install_snap(self, snap_local_path, snap_name, version,
                      devmode=False):
         if not config.get('skip-install', False):
-            self.snappy_testbed.copy_file(snap_local_path, '/home/ubuntu')
-            snap_file_name = os.path.basename(snap_local_path)
-            snap_path_in_testbed = os.path.join(
-                '/home/ubuntu/', snap_file_name)
-            # Remove the snap file from the testbed.
+            tmp_in_testbed = self.snappy_testbed.run_command(
+                'mktemp -d').strip()
```
kyrofa: Why not use python utilities for doing this (tempfile)?

kyrofa: Oh right, this is on another machine, huh? Scratch this comment.

snaps_tests/__init__.py
```diff
@@ -193,14 +193,15 @@ def install_store_snap(self, snap_name):
...
             self.addCleanup(
                 self.snappy_testbed.run_command,
-                ['rm', snap_path_in_testbed])
+                ['rm', '-r', tmp_in_testbed])
```
kyrofa: Would it be worth adding a `-f` here just in case?

elopio: +1. I pushed the change.

snaps_tests/__init__.py
elopio: well, potentially another machine. We are not running in a remote testbed anymore, so it would be worth thinking if it's better to simplify these scripts.
However, if I understand correctly, the snapcraft snap will only work on classic. Then we still need to test the snaps in remote core some times.

PR #1071, opened 2017-01-25
Download all Python packages in the python plugin using one command.

bjornt:
Before, there was one command for downloading the packages in
requirements file, one for the ones in setup.py, and one for downloading
the manually specified packages.

That broke the case where you specify high-level dependencies in
setup.py without version-specific contstraints and use requirements.txt
to pin the packages to specific versions. The end result was that even
though packages were pinned in requirements.txt, snapcraft would
download the latest versions of packages.

LP: [#1659234](https://bugs.launchpad.net/snapcraft/+bug/1659234)

### DIFF
--- a/snapcraft/plugins/python.py
+++ b/snapcraft/plugins/python.py
@@ -191,23 +191,26 @@ class PythonPlugin(snapcraft.BasePlugin):
         return env
 
     def _get_commands(self, setup):
-        commands = []
+        args = []
+        cwd = None
         if self.options.requirements:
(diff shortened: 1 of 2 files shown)


sergiusens:
Looks sane, there are some static test/linting failures and we now check that either the commit is done with a @canonical.com address or the email in question is in the "has signed the CLA" camp. Mind updating?

bjornt:
I pushed a new version that addresses all the issues I could find.

PR #1075, opened 2017-01-25
travis: allow to pass --channels parameter to generate proper config

3v1n0:
This allows to generate credentials for uploading to other channels than 'edge'.
Although this is still experimental, having the ability to override the default edge channel is something to have IMHO.

LP: [#1659352](https://bugs.launchpad.net/snapcraft/+bug/1659352)

### DIFF
--- a/snapcraft/integrations/__init__.py
+++ b/snapcraft/integrations/__init__.py
@@ -27,7 +27,7 @@ SUPPORTED_CI_SYSTEMS = (
 )
 
 
-def enable_ci(ci_system, refresh_only):
+def enable_ci(ci_system, refresh_only, channels):
     if not ci_system:
         # XXX cprov 20161116: we could possibly auto-detect currently
         # integration systems in master ?
@@ -46,8 +46,8 @@ def enable_ci(ci_system, refresh_only):
         'snapcraft.integrations.{}'.format(ci_system))
 
     if refresh_only:
-        module.refresh()
+        module.refresh(channels)
     else:
         print(module.__doc__)
         if input('Continue (y/N): ') == 'y':
-            module.enable()
+            module.enable(channels)
--- a/snapcraft/integrations/tests/test_travis.py
+++ b/snapcraft/integrations/tests/test_travis.py
@@ -139,7 +139,7 @@ class TravisSuccessfulTestCase(tests.TestCase):
 
     scenarios = [
         ('default', dict(
-            channels=None)),
+            channels=[])),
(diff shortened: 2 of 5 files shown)


sergiusens:
This is something for @evandandrea :-)

I believe we need to have humans do the part of releasing to other channels just like we do for any other software work on in Ubuntu. While you can I don't think we should promote this. That said, final word goes to Evan.

kyrofa:
> I believe we need to have humans do the part of releasing to other channels just like we do for any other software work on in Ubuntu. While you can I don't think we should promote this. That said, final word goes to Evan.

Promoting and supporting some else's workflow are two different things, IMHO.

3v1n0:
> I believe we need to have humans do the part of releasing to other channels just like we do for any other software work on in Ubuntu. While you can I don't think we should promote this.

I agree, in fact 'edge' is still the default one, but if you want to force this, or still allowing auto-uploads to channels such as 'beta', you should be able to do that without having to hack snapcraft.

There are some maintainers that were interested in having auto-uploads in stable too, it's their responsibility at the end. We can say what the good practices are, we can't really force anyone to follow them.

cprov:
I think it gives people the wrong incentives. We have being messaging emphatically that CI snaps go to edge, where it's clearer the risk is *high*,  and then promoted to more stable channels as they pass *human* reviews.

IMHO, the fact that it's technically possible (in other words, not easily preventable like users leaking their SSO password) is not a excuse to point developers to this direction and make it a 'supported' way. Once they get invested, it will be much harder to change.

3v1n0:
Again, it's up to upstream to do that... They could even do this in local scripts, and in any case they have to force this. Some upstreams really want to do automatic uploads on git tags (i.e. releases; as [remmina, for example](https://github.com/FreeRDP/Remmina/blob/next/scripts/travis-build.sh#L64)) and having to hack snapcraft to do it, causes some annoyances.

kyrofa:
> Some upstreams really want to do automatic uploads on [releases][...]

Indeed, Nextcloud does this as well. In fact, they have daily master builds going to edge (so literally their edge), daily stable builds going to beta, then since they're using GitFlow to develop the snap itself, the development branch is published on candidate and finally when they release to master it's published to stable. It seems unwise to force a workflow on such projects.

sergiusens:
Sorry, people can workflow as they want, but even the macaroon that this integration will provide will eventually only allow you to push to edge. We should lead on good practices.

PR #1079, opened 2017-01-26
Add snapcraft plugin for Qt Build Suite (qbs)

dpniel:
This plugin supports building any applications that use gcc or clang
toolchains, as well as Qt(4&5) based applications.

LP: [#1659537](https://bugs.launchpad.net/snapcraft/+bug/1659537)

### DIFF
--- a/integration_tests/snaps/qbs-nil/snapcraft.yaml
+++ b/integration_tests/snaps/qbs-nil/snapcraft.yaml
@@ -1,0 +1,13 @@
+name: test-qbs-plugin
+version: 0.1
+summary: qbs snapcraft plugin
+description: Tests if parts can be built using qbs and gcc
+confinement: strict
+
+parts:
+  qbs-project:
+    plugin: qbs
+    source: .
+    qbs-build-variant: debug
+    qbs-options: ['project.printString:false']
+    qbs-profile: gcc
--- a/integration_tests/snaps/qbs-qt/snapcraft.yaml
+++ b/integration_tests/snaps/qbs-qt/snapcraft.yaml
@@ -1,0 +1,14 @@
+name: test-qbs-plugin
+version: 0.1
+summary: qbs snapcraft plugin
+description: Tests if Qt based parts can be built using qbs
+confinement: strict
+
+parts:
+  qbs-project:
+    plugin: qbs
+    source: .
+    qbs-options: ['project.printString:true']
+    qbs-profile: clang
+    qbs-build-variant: release
+    qt-version: qt5
--- a/integration_tests/test_qbs_plugin.py
+++ b/integration_tests/test_qbs_plugin.py
@@ -1,0 +1,41 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2017 Dan Chapman <dpniel@ubuntu.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+import os
+
+import testscenarios
+import integration_tests
+
+
+class QbsPluginTestCase(testscenarios.WithScenarios,
+                        integration_tests.TestCase):
+
+    scenarios = [
+        # we expect qbs-nil to output the wrong string here as
+        # 'project.printString' is set to false in the yaml file.
+        ('qbs-nil', dict(project_dir='qbs-nil',
+                         expected_value='Wrong snapcraft string')),
+        ('qbs-qt', dict(project_dir='qbs-qt',
+                        expected_value='Hello Snapcraft World')),
+    ]
+
+    def test_stage_qbs_plugin(self):
+
+        self.run_snapcraft('stage', self.project_dir)
+
+        # The test binaries have the same name as the project dir
+        binary_output = self.get_output_ignoring_non_zero_exit(
+            os.path.join(self.stage_dir, 'bin', self.project_dir))
+        self.assertEqual(self.expected_value, binary_output)
--- a/snapcraft/plugins/qbs.py
+++ b/snapcraft/plugins/qbs.py
@@ -1,0 +1,153 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2017 Dan Chapman <dpniel@ubuntu.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""The QBS plugin is useful for building parts that use the Qt Build Suite.
+
+These are projects that are built using .qbs files and are using gcc or
+clang based toolchains.
+
+This plugin uses the common plugin keywords as well as those for "sources".
+For more information check the 'plugins' topic for the former and the
+'sources' topic for the latter.
+
+Additionally, this plugin uses the following plugin-specific keywords:
+
+    - qbs-build-variant:
+      (enum, 'debug' or 'release')
+      Set either debug or release build variant. Default is release.
+    - qbs-options:
+      (list of strings)
+      Additional options to pass to the qbs invocation.
+    - qbs-profile:
+      (enum, 'gcc' or 'clang')
+      Set either gcc or clang as the base build profile. Default is gcc.
+    - qt-version:
+      (enum, 'qt4' or 'qt5')
+      Optionally you can set up profiles for qt4 and qt5 applications.
+      If no qt version is set then the base gcc/clang profile will be used.
+"""
+import os
+import shutil
+import snapcraft
+
+
+class QbsPlugin(snapcraft.BasePlugin):
+
+    @classmethod
+    def schema(cls):
+        schema = super().schema()
+
+        schema['properties']['qbs-build-variant'] = {
+            'enum': ['debug', 'release'],
+            'default': 'release',
+        }
+
+        schema['properties']['qbs-options'] = {
+            'type': 'array',
+            'minitems': 1,
+            'uniqueItems': True,
+            'items': {
+                'type': 'string',
+            },
+            'default': [],
+        }
+
+        schema['properties']['qbs-profile'] = {
+            'enum': ['gcc', 'clang'],
+            'default': 'gcc',
+        }
+
+        schema['properties']['qt-version'] = {
+            'enum': ['qt4', 'qt5'],
+            'default': None
+        }
+
+        return schema
+
+    @classmethod
+    def get_pull_properties(cls):
+        # Inform Snapcraft of the properties associated with pulling. If these
+        # change in the YAML Snapcraft will consider the pull step dirty.
+        return ['qt-version', 'qbs-profile']
+
+    @classmethod
+    def get_build_properties(cls):
+        # Inform Snapcraft of the properties associated with building. If these
+        # change in the YAML Snapcraft will consider the build step dirty.
+        return ['qbs-options', 'qbs-build-variant']
+
+    def __init__(self, name, options, project):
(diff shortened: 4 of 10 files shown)


snapcraft/plugins/qbs.py
```diff
@@ -0,0 +1,146 @@
...
+
+    @classmethod
+    def get_build_properties(cls):
+        # Inform Snapcraft of the properties associated with building. If these
+        # change in the YAML Snapcraft will consider the build step dirty.
+        return ['options', 'build-variant']
+
+    def __init__(self, name, options, project):
+        super().__init__(name, options, project)
+        self.build_packages.extend(['qbs', 'gcc', 'clang'])
```
3v1n0: Add Make too

dpniel: Why is Make required here? qbs doesn't use make and handles the execution on the build graph itself.

3v1n0: Sorry... I misread `/usr/bin/qmake` with `/usr/bin/make` :-)

dpniel: :-) few!

snapcraft/plugins/qbs.py
```diff
@@ -0,0 +1,146 @@
...
+Additionally, this plugin uses the following plugin-specific keywords:
+
+    - options:
+      (list of strings)
+      additional options to pass to the qbs invocation.
+    - qt-version:
+      (enum, 'nil' or 'qt4' or 'qt5')
+      Version of Qt to use for qbs profile. If nil the base gcc/clang profile
+      will be used.
+    - profile:
```
sergiusens: does it make sense to call this `qbs-profile`?

I am trying to get to a place where we can somehow namespace things that come from plugins and things that come from core.

snapcraft/plugins/qbs.py
```diff
@@ -0,0 +1,146 @@
...
+            },
+            'default': [],
+        }
+
+        schema['properties']['profile'] = {
+            'enum': ['gcc', 'clang'],
+            'default': 'gcc',
+        }
+
+        schema['properties']['build-variant'] = {
```
sergiusens: this is not documented, also, can this be renamed to `qbs-build-variant`?

snapcraft/plugins/qbs.py
```diff
@@ -0,0 +1,146 @@
...
+    - options:
+      (list of strings)
+      additional options to pass to the qbs invocation.
+    - qt-version:
+      (enum, 'nil' or 'qt4' or 'qt5')
+      Version of Qt to use for qbs profile. If nil the base gcc/clang profile
+      will be used.
+    - profile:
+      (enum, 'gcc' or 'clang')
+      Set either gcc or clang as the base build profile
```
sergiusens: mention that gcc is the default please

snapcraft/plugins/qbs.py
```diff
@@ -0,0 +1,146 @@
...
+
+        build_profile = ""
+
+        if self.options.qt_version == 'nil':
+            build_profile = self.options.profile
+        else:
+            # a unique'ish name for snap builds. Hopefully this shouldn't clash
+            # with any other local profiles. Each profile should look something
+            # like: profiles.snapcraftqt5clang or profiles.snapcraftnilgcc
+            build_profile = 'snapcraft' + \
```
sergiusens: how about
```python
build_profile = 'snapcraft-qbs-{}-{}'.format(
    self.options.qt_version,
    self.options.profile)
```

and update the comment as well if you take this

snapcraft/plugins/qbs.py
```diff
@@ -0,0 +1,146 @@
...
+                            self.options.qt_version + \
+                            self.options.profile
+
+            # Setup the Qt profile.
+            self.run(['qbs', 'setup-qt', '/usr/bin/qmake', build_profile],
```
sergiusens: since calling qmake on a specific path, shouldn't you before do this:

```
if not os.path.exists('/usr/bin/qmake'):
    raise RuntimeError('The qbs plugin requires qmake on /usr/bin/qmake')
```

Do you also want to hardcode it to that path?

dpniel:
@sergiusens Thanks for the review :-) 

I think I have addressed all your comments now.

I also noticed that the integration tests didn't cover the qbs-options property so I have added that in to switch the binary output and make sure it's being passed through correctly. 

Also I'm not sure if having to create a file to fake qmake being on PATH, in order to make the unit test happy is the right thing to do. Is there some better way to fake it's existence on PATH?

sergiusens:
@dpniel I'll defer to the QA guy who goes after the call sign @elopio :smile:

integration_tests/snaps/qbs-nil/snapcraft.yaml
```diff
@@ -0,0 +1,16 @@
...
+description: Tests if parts can be built using qbs and gcc
+confinement: strict
+
+parts:
+  qbs-project:
+    plugin: qbs
+    source: .
+    qbs-options: [
+      'project.printString:true'
+    ]
```
elopio: nit comment, so far I haven't seen this style in yaml.
I've seen:

qbs-options: ['project.printString:true']

and 

qbs-options:
  - 'project.printString:true'

Have you seen that usage in other places? If not, please change it.

integration_tests/snaps/qbs-qt/snapcraft.yaml
```diff
@@ -0,0 +1,16 @@
...
+description: Tests if Qt based parts can be built using qbs
+confinement: strict
+
+parts:
+  qbs-project:
+    plugin: qbs
+    source: .
+    qbs-options: [
+      'project.printString:true'
+    ]
```
elopio: ditto: style on the yaml list.

integration_tests/test_qbs_plugin.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def test_stage_qbs_plugin(self):
+
+        self.run_snapcraft('stage', self.project_dir)
+
+        # The test binaries have the same name as the project dir
+        binary_output = self.get_output_ignoring_non_zero_exit(
+            os.path.join('stage', 'bin', self.project_dir),
+            cwd=self.project_dir)
+        self.assertEqual("Hello Snapcraft World", binary_output)
```
elopio: Please use single quotes, for consistency.

elopio: You are not testing the Wrong snapcraft string message. That would be a nice thing to add here.

snapcraft/plugins/qbs.py
```diff
@@ -0,0 +1,157 @@
...
+            # like: profiles.snapcraft-qbs-qt5-clang
+            build_profile = 'snapcraft-qbs-{}-{}'.format(
+                self.options.qt_version,
+                self.options.qbs_profile)
+
+            # Find full qmake path as qbs setup-qt requires it
+            qmake = shutil.which('qmake')
+            if qmake is None:
+                # Let's just check it's not in a common location
+                qmake = os.path.join('usr', 'bin', 'qmake')
```
elopio: I don't see a case where qmake is in /usr/bin, but which can't find it.
Do you think this conditional is necessary?

snapcraft/plugins/qbs.py
```diff
@@ -0,0 +1,157 @@
...
+                if not os.path.exists(qmake):
+                    raise RuntimeError(
+                        'The qbs plugin requires qmake be on PATH')
```
elopio: What happens when your machine has both qt5-qmake and qt4-qmake installed? Which one will be /usr/bin/qmake?

snapcraft/plugins/qbs.py
```diff
@@ -0,0 +1,157 @@
...
+                  '-f', self.sourcedir,
+                  self.options.qbs_build_variant,
+                  'qbs.installRoot:' + self.installdir,
+                  'profile:' + build_profile] + self.options.qbs_options,
+                 env=env)
+
+    def _build_environment(self):
+        env = os.environ.copy()
+        env['QT_SELECT'] = self.options.qt_version
+        return env
```
elopio: I like your comments as a guide inside the code blocks. Thanks!

snapcraft/plugins/qbs.py
```diff
@@ -0,0 +1,157 @@
...
+For more information check the 'plugins' topic for the former and the
+'sources' topic for the latter.
+
+Additionally, this plugin uses the following plugin-specific keywords:
+
+    - qbs-options:
+      (list of strings)
+      Additional options to pass to the qbs invocation.
+    - qt-version:
+      (enum, 'nil' or 'qt4' or 'qt5')
```
dpniel: For this I was trying to be explicit in the intentions of the plugin. as qbs can be used to build any project and not just Qt apps. I'm happy to remove nil and make it optional if you think it would be clear enough. I'll update the doc string as well to indicate the `qt_version` is optional.

snapcraft/plugins/qbs.py
```diff
@@ -0,0 +1,157 @@
...
+            'items': {
+                'type': 'string',
+            },
+            'default': [],
+        }
+
+        schema['properties']['qbs-profile'] = {
+            'enum': ['gcc', 'clang'],
+            'default': 'gcc',
+        }
```
dpniel: Heh, that's cool! I'm happy to get this right and learn how you guys like it done :-)

snapcraft/plugins/qbs.py
```diff
@@ -0,0 +1,157 @@
...
+            raise RuntimeError('Unsupported Qt version: {!r}'.format(
+                self.options.qt_version))
+
+    def build(self):
+        super().build()
+
+        env = self._build_environment()
+
+        # Setup the toolchains, there will only be gcc or clang by default
+        self.run(['qbs', 'setup-toolchains', '--detect'], env=env)
```
dpniel: At this point qbs is only interested in finding toolchains it knows about and creating the base profiles for them.

The place where trouble would be caused is in the `qbs config profile.*.baseProfile = 'clang'`. Which would cause the build to fail. But really you would want that to error out at that point, as clang is expected to be in the detected profiles if clang is set in the yaml.

snapcraft/plugins/qbs.py
```diff
@@ -0,0 +1,157 @@
...
+
+        build_profile = ""
+
+        if self.options.qt_version == 'nil':
+            build_profile = self.options.qbs_profile
+        else:
+            # a unique'ish name for snap builds. Hopefully this shouldn't clash
+            # with any other local profiles. Each profile should look something
+            # like: profiles.snapcraft-qbs-qt5-clang
```
dpniel: I did think about uuids, but I think it would make debugging profile issues somewhat difficult. You would only really want one profile for each possible configuration in the `~/.config/QtProject/qbs/profiles` directory and a uuid would make it difficult to figure out which profile you need to be looking at. Plus you would end up with multiple duplicates of the same configuration if running snapcraft locally. Which would get messy.

snapcraft/plugins/qbs.py
dpniel: Good point. will remove

snapcraft/plugins/qbs.py
dpniel: Hopefully the correct one :-D 

This is handled by setting the QT_SELECT env var which is done in `QbsPlugin._build_environment()`. The qt-chooser magic takes care of the rest at run time, so you should always get the correct qmake.

dpniel:
@elopio Regarding you question:

"And a big question: if we find a bug with this qbs plugin, will you be around to give us a hand?"

Yep no problem, just give me a ping. I'm lurking in the background on rocket.u.c/snapcraft or i'm easily found on irc with nick "DanChapman" :-)

snapcraft/plugins/qbs.py
elopio: thanks for the info.

snapcraft/plugins/qbs.py
elopio: makes sense.

integration_tests/test_qbs_plugin.py
```diff
@@ -0,0 +1,38 @@
...
+    ]
+
+    def test_stage_qbs_plugin(self):
+
+        self.run_snapcraft('stage', self.project_dir)
+
+        # The test binaries have the same name as the project dir
+        binary_output = self.get_output_ignoring_non_zero_exit(
+            os.path.join(self.stage_dir, 'bin', self.project_dir))
+        self.assertNotEqual('Wrong snapcraft string', binary_output)
```
elopio: sorry, I meant to build them again with project.printString, to make sure that those arguments work.

dpniel: I've changed this so qbs-nil builds with project.printString:false and qbs-qt with project.printString:true.

From what I can see it's not possible to change `qbs-options` from within the test (am i missing something?), and would require adding more test directories with different snapcraft.yaml files. Which seems like it's needlessly duplicating code.

bjorn:
@dpniel Would there be any chance of getting this reverted plugin added again? Maybe for some visibility LP: [#1659537](https://bugs.launchpad.net/snapcraft/+bug/1659537) should also be reopened.

bjorn:
No idea why it claims I've unassigned @elopio. :man_shrugging:

cjp256:
@bjorn  I would be glad to assist if someone in the community wanted to revive the plugin and iron out any major issues.

bjorn:
@cjp256 Alright, I've revived this PR as #2892 with the goal of trying to make those tests pass. Any help would be really appreciated!

PR #1080, opened 2017-01-26
python plugin: avoid the use of PYTHON* env vars.

sergiusens:
Avoid the use of environment variables as they leak into whatever else
calls them. This helps classic confinement but also other things like
snaps with python2 and python3 parts.

`PYTHONHOME` can be determined by `argv0` and `PYTHONUSEBASE` is setup in `site.py`

LP: #1657504


sergiusens:
just want to state here that @matiasb gave me a _loose_ :+1: :smile:

sergiusens:
Might be replaced with #1093

PR #1081, opened 2017-01-26
local source: preserve symlinks to directories.

kyrofa:
This PR fixes LP: [#1659653](https://bugs.launchpad.net/snapcraft/+bug/1659653) by preserving any symlinks to directories within local sources instead of creating a new directory.

### DIFF
--- a/snapcraft/internal/sources/_local.py
+++ b/snapcraft/internal/sources/_local.py
@@ -47,5 +47,5 @@ class Local(Base):
             else:
                 return []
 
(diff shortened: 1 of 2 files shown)


sergiusens:
let's merge this after #1083 just so we get three similar ducks in a row :-)

kyrofa:
> let's merge this after #1083 just so we get three similar ducks in a row :-)

kyrofa:
> let's merge this after #1083 just so we get three similar ducks in a row :-)

Sounds good.

PR #1082, opened 2017-01-26
project: new plugin directory location

sergiusens:
LP: #1659085


integration_tests/snaps/local-plugin/snap/plugins/x_local_plugin.py
```diff
@@ -0,0 +1,22 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2015 Canonical Ltd
```
kyrofa: Just to double-check: this isn't a new file (i.e. no need to update this)?

integration_tests/test_local_plugin.py
```diff
@@ -30,13 +31,31 @@ def test_stage_local_plugin(self):
...
+
+class LocalPluginCleanTestCase(testscenarios.WithScenarios,
+                               integration_tests.TestCase):
+
+    scenarios = [
+        ('local-plugin', dict(project_dir='local-plugin',
+                              base_dir='snap')),
+        ('local-plugin-in-parts', dict(project_dir='local-plugin-in-parts',
+                                       base_dir='parts')),
+    ]
```
kyrofa: Nice, thanks for testing this.

snapcraft/internal/deprecations.py
```diff
@@ -25,7 +25,8 @@
 import logging
 
 _DEPRECATION_MESSAGES = {
-    'dn1': "The 'snap' keyword has been replaced by 'prime'."
+    'dn1': "The 'snap' keyword has been replaced by 'prime'.",
+    'dn2': "Custom plugins should now be placed in `snap/plugins`.",
```
kyrofa: We sort of have a mix of markdown-style code blocks (the backticks) and single quotes, like above. Can we stick with either quotes or backticks in the deprecation notices? I have a slight preference to single quotes since these are surfaced directly to the user, but I have no strong feelings.

integration_tests/snaps/local-plugin/snap/plugins/x_local_plugin.py
sergiusens: it is a copy

snapcraft/internal/deprecations.py
sergiusens: single quotes is fine and acceptable

PR #1083, opened 2017-01-27
project: support for gui in snap

sergiusens:
Depecrate the setup directory in favor of
the snap directory.

LP: #1659085


snapcraft/internal/deprecations.py
```diff
@@ -27,6 +27,7 @@
 _DEPRECATION_MESSAGES = {
     'dn1': "The 'snap' keyword has been replaced by 'prime'.",
     'dn2': "Custom plugins should now be placed in 'snap/plugins'.",
+    'dn3': "Assets in 'setup/gui' should now be placed in to 'snap/gui'.",
```
kyrofa: Small nit: I suggest scratching the "to": "Assets in 'setup/gui' should now be placed in 'snap/gui'."

snapcraft/internal/meta.py
```diff
@@ -163,25 +164,28 @@ def write_snap_directory(self):
...
-                    raise CommandError('hook {!r} is not executable'.format(
-                        hook_name))
-
-                with contextlib.suppress(FileNotFoundError):
-                    os.remove(destination)
-
-                file_utils.link_or_copy(source, destination)
+        # Now copy the assets contained within the snap directory directly into
+        # meta (they don't get wrappers for hooks like the ones that come from
+        # parts).
```
kyrofa: This comment doesn't quite make sense changing 'hooks' to 'assets,' as assets has nothing to do with wrappers.

snapcraft/internal/meta.py
```diff
@@ -163,25 +164,28 @@ def write_snap_directory(self):
...
+                os.makedirs(dst_dir, exist_ok=True)
+                for asset in os.listdir(src_dir):
+                    source = os.path.join(src_dir, asset)
+                    destination = os.path.join(dst_dir, asset)
+
+                    # First, verify that the hook is actually executable
+                    if (origin == 'hooks' and
+                            not os.stat(source).st_mode & stat.S_IEXEC):
+                        raise CommandError(
+                            'hook {!r} is not executable'.format(asset))
```
kyrofa: Would it be better/more clear to refactor this into a standalone function that accepts a path and some sort of `checker` function?

snapcraft/internal/deprecations.py
sergiusens: oh darn, that is re-edit at its best

snapcraft/internal/meta.py
sergiusens: yes, it was my thought as well, but felt lazy

snapcraft/internal/meta.py
```diff
@@ -177,10 +176,8 @@ def write_snap_directory(self):
                     destination = os.path.join(dst_dir, asset)
 
                     # First, verify that the hook is actually executable
-                    if (origin == 'hooks' and
-                            not os.stat(source).st_mode & stat.S_IEXEC):
-                        raise CommandError(
-                            'hook {!r} is not executable'.format(asset))
+                    if origin == 'hooks':
+                        _validate_hook(source)
```
kyrofa: I was really talking about the overall migration function, something like `_migrate_into_snapdir(path, checker=None)` where the checker could validate the hook when the `path='hooks'`. But yeah, this is fine.

sergiusens: we don't have time for this today, add a technical debt card please

PR #1084, opened 2017-01-27
make: add support for cwd path to make() function

3v1n0:
This can be useful for plugins that extends Make so that they can define a subfolder where to run make.

LP: [#1659810](https://bugs.launchpad.net/snapcraft/+bug/1659810)

### DIFF
--- a/snapcraft/plugins/make.py
+++ b/snapcraft/plugins/make.py
@@ -95,7 +95,7 @@ class MakePlugin(snapcraft.BasePlugin):
         super().__init__(name, options, project)
         self.build_packages.append('make')
 
-    def make(self, env=None):
+    def make(self, cwd=None, env=None):
         command = ['make']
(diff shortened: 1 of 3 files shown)


sergiusens:
Hey, thanks for his. Just to clear my ignorance, is there a difference between this and doing `make -f` which we alreday support?

Can we add an integration test with a custom plugin that makes use of this feature if `make -f` is not suitable?

sergiusens:
Meh, we can certainly support this even if `make -f` is suitable.

3v1n0:
`make -f` would require a Makefile path, and it's fine, but different... Instead I could replace the chdir thing by using `make -C` instead, but there's really not much difference.

kyrofa:
Closing due to inactivity. Please re-propose when you've made the changes requested.

PR #1085, opened 2017-01-27
snapcraft: fix or ignore PEP8 static errors

3v1n0:
I was annoyed by these warnings... 

LP: [#1659816](https://bugs.launchpad.net/snapcraft/+bug/1659816)

### DIFF
--- a/snapcraft/internal/deltas/__init__.py
+++ b/snapcraft/internal/deltas/__init__.py
@@ -15,7 +15,6 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 
-
-from . import errors # noqa
-from ._deltas import BaseDeltasGenerator # noqa
-from ._xdelta3 import XDelta3Generator # noqa
+from . import errors  # noqa
+from ._deltas import BaseDeltasGenerator  # noqa
+from ._xdelta3 import XDelta3Generator  # noqa
--- a/snapcraft/internal/sources/__init__.py
+++ b/snapcraft/internal/sources/__init__.py
@@ -147,7 +147,7 @@ def get_source_handler(source, *, source_type=''):
 _tar_type_regex = re.compile(r'.*\.((tar(\.(xz|gz|bz2))?)|tgz)$')
 
 
-def _get_source_type_from_uri(source, ignore_errors=False):
+def _get_source_type_from_uri(source, ignore_errors=False):  # noqa: C901
     source_type = ''
     if source.startswith('bzr:') or source.startswith('lp:'):
         source_type = 'bzr'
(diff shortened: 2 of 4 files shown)


sergiusens:
aren't you using the pinned versions in requirements-devel.txt?

3v1n0:
Oh, maybe not?

Anyway, are these changes wrong with those versions too?

kyrofa:
> Oh, maybe not... I preferred only use archive stuff

Me too, but I don't think the fix is to ignore stuff, because then even more recent versions of the tools won't check them.

elopio:
what's C901?

3v1n0:
> what's C901?

```
snapcraft/internal/sources/__init__.py:150:1: C901 '_get_source_type_from_uri' is too complex (11)
snapcraft/tests/fake_servers.py:336:1: C901 'FakeStoreAPIRequestHandler.do_POST' is too complex (11)```

> Me too, but I don't think the fix is to ignore stuff, because then even more recent versions of the tools won't check them.

I'm only ignoring the too-complex error as above, not anything else... I fixed the syntax of other lines ignoring stuff (that was not using two leading spaces)

elopio:
I think that a function with complexity 11 should be split in two functions. Ideally, more than that. I wonder why that is not caught by travis...

3v1n0:
@elopio if you check those functions, splitting them would cause their complexity to grow IMHO (unless we don't use maps or something).

That rule is fine, but it doesn't apply to all the cases.

elopio:
@3v1n0 I think I would like to see a map for the source types. But not a big deal, +1 from me.

sergiusens:
@3v1n0 we don't do static tests as part of the archive integration. We only rely on pypi and the reason is, we want to support the same codebase for xenial, yakkety and zesty.

sergiusens:
oh, and we locked it to xenial's versions for people developing using the archive packages instead of pypi

kyrofa:
> I'm only ignoring the too-complex error as above

Right, that's what I'm talking about. Ignoring it means that the function can further grow in complexity, and even the newer review tools will never catch it. I'd rather live with the static error, or split it.

3v1n0:
Ok, fine... You want me to remove the `# noqa: C901` and leave the rest or just ignoring all this?

sergiusens:
I'll leave that call to @elopio, but if this goes in please update `requirements-devel.txt` with the appropriate static checker versions

elopio:
I said I'm ok with this. The branch makes the code a little better, and then if somebody comes with a clever way to simplify the function while keeping it clear and simple, we can remove the #noqa. Also I don't think we are at risk of filling our repo with noqa, this are two special cases.

The one that's getting pretty ugly is the fake server. I can take an action to make it better after this sprint.

PR #1086, opened 2017-01-27
print snapcraft's version on startup when running with --debug

chipaca:
LP: [1659817](https://bugs.launchpad.net/snapcraft/+bug/1659817)


kyrofa:
Thanks for the PR @chipaca! It looks like you need to update some of the expected output from the integration tests, though (they all run in debug).

PR #1089, opened 2017-01-28
options: add armv8l to list of translations.

sergiusens:
platform.machine() is returning armv8l instead of aarch64 on
some architectures when on a 32bit userspace. This addresses that
problem.

LP: #1641123

### DIFF
--- a/snapcraft/_options.py
+++ b/snapcraft/_options.py
@@ -79,8 +79,9 @@ _ARCH_TRANSLATIONS = {
 }
 
 
-_USERSPACE_ARCHITECTURE = {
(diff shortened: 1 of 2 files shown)


elopio:
This will fix the execution, so +1 there. But I don't know enough about linux architectures to confirm that this is a correct fix.
If this is needed to finish the release, we can land it and ask for confirmation later, as it will not represent any regression and we can easily change it later.

sergiusens:
we'll see how it goes

sergiusens:
I can already see `Priming ros-project` so must be good.

PR #1090, opened 2017-01-28
core: classic with no exported variables.

sergiusens:
To avoid unexpected errors we wipe clean the envrionment
variables inside the wrappers, if a developer needs it
they could export it themselves.

LP: #1657504


sergiusens:
Might be replace with #1093

PR #1091, opened 2017-01-29
tests:  add ubuntu user to sudoers on every adt platform 

elopio:
LP: #1660459

### DIFF
--- a/debian/tests/integrationtests
+++ b/debian/tests/integrationtests
@@ -1,7 +1,5 @@
 #!/bin/sh
 
-if [ "$(uname -i)" = 'armv7l' ]; then
-    echo 'ubuntu ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers
-fi
+echo 'ubuntu ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers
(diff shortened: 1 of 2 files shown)


debian/tests/integrationtests
```diff
@@ -1,7 +1,5 @@
 #!/bin/sh
 
-if [ "$(uname -i)" = 'armv7l' ]; then
-    echo 'ubuntu ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers
-fi
+echo 'ubuntu ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers
```
sergiusens: might be better to use `/etc/sudoers.d`

elopio: That sounds like a good idea. I'm still not sure if this will fix it though. The only thing that makes sense with this is error is that something changed in zesty.

elopio:
@sergiusens I think this worked. The remaining issues seem to be failures to connect to the rust server.
I will make an RT.

elopio:
oh, no, it's rustup: unknown CPU type: armv8l

PR #1093, opened 2017-01-30
python plugin: do the right thing with classic.

sergiusens:
To fully enable classic confinment for python based snaps, there are
two big things that need to get done; no env var leaking and
use a python interpreter with the classic compilation flags.

LP: #1657504


battlemidget:
Will this allow a successful installation of snaps on trusty as well?

sergiusens:
@battlemidget yes; there is only one last thing for trusty support being tracked on https://bugs.launchpad.net/bugs/1657504

There is however one last issue involving core on comment https://bugs.launchpad.net/snapd/+bug/1657504/comments/15

For classic, you *will* need to provide an interpreter, a quick way is like

```yaml
name: asciinema
version: '1.3.0'
summary: Record and share your terminal sessions, the right way.
description: |
  Forget screen recording apps and blurry video. Enjoy a lightweight,
  purely text-based approach to terminal recording.

grade: devel
confinement: classic

apps:
  asciinema:
    command: bin/asciinema

parts:
  asciinema:
    plugin: python
    python-packages: [$SNAPCRAFT_PROJECT_NAME==$SNAPCRAFT_PROJECT_VERSION]
    after: [python3]
  python3:
    source: https://www.python.org/ftp/python/3.6.0/Python-3.6.0.tar.xz
    plugin: autotools
    configflags: [--prefix=/usr]
    build-packages: [libssl-dev]
```

snapcraft/internal/meta.py
```diff
@@ -253,33 +254,41 @@ def _write_wrap_exe(self, wrapexec, wrappath,
...
-        classic_library_paths = self._config_data['confinement'] == 'classic'
-        assembled_env = common.assemble_env(classic_library_paths,
-                                            self._arch_triplet)
-        assembled_env = assembled_env.replace(self._snap_dir, '$SNAP')
+        # binaries are linked with `nodefaultlib` but we still do
+        # not want to leak PATH or other environment variables
+        # that would affect the applications view of the classic
+        # environment it is dropped into.
         replace_path = r'{}/[a-z0-9][a-z0-9+-]*/install'.format(
             self._parts_dir)
```
chipaca: has `_parts_dir` been sanitized to not have regexpish things in it? otherwise you want `re.escape(self._parts_dir)` here

snapcraft/plugins/python.py
```diff
@@ -163,24 +185,51 @@ def clean_pull(self):
...
         pip = _Pip(exec_func=subprocess.check_call,
-                   runnable=self.system_pip_command,
+                   runnable=pip_command,
                    package_dir=self._python_package_dir, env=env,
                    extra_install_args=['--ignore-installed'])
 
         if download:
             pip.download(args)
-        pip.wheel(args)
+        # pip.wheel(args)
```
chipaca: `# leftover? Or ...?`

snapcraft/internal/meta.py
```diff
@@ -253,33 +254,41 @@ def _write_wrap_exe(self, wrapexec, wrappath,
...
         with open(wrappath, 'w+') as f:
-            f.write(script)
+            f.write('#!/bin/sh\n')
+            if assembled_env:
+                f.write('{}\n'.format(assembled_env))
+                f.write('export LD_LIBRARY_PATH=$SNAP_LIBRARY_PATH:'
+                        '$LD_LIBRARY_PATH\n')
+            if cwd:
+                f.write('{}\n'.format(cwd))
+            f.write('exec {} {}\n'.format(executable, args))
```
chipaca: your use of `f.write` is entertaining :-)

snapcraft/plugins/python.py
sergiusens: yeah, already removed locally, didn't want to stress the weak infra ;-)

snapcraft/internal/meta.py
sergiusens: I can change to `print` :wink:

snapcraft/internal/meta.py
sergiusens: it is sanitized but good idea

snapcraft/shutil.py
```diff
@@ -0,0 +1,31 @@
...
+
+"""Utilities to run inside a correct snapcrafting environment."""
+
+
+import tempfile
+
+from snapcraft.internal import common
+
+
+def which(command, **kwargs):
```
kyrofa: Why not use the real `shutil.which()` with `assemble_env()`'s PATH?

kyrofa: If we want to keep this, can we name the file something else? I'm afraid `shutil` will cause confusion later.

snapcraft/__init__.py
```diff
@@ -304,6 +304,7 @@
 from snapcraft import plugins                       # noqa
 from snapcraft import sources                       # noqa
 from snapcraft import file_utils                    # noqa
+from snapcraft import shutil                        # noqa
```
kyrofa: Is this necessary? I don't see it used.

snapcraft/_baseplugin.py
```diff
@@ -73,13 +73,21 @@ def PLUGIN_STAGE_SOURCES(self):
         """Define alternative sources.list."""
         return getattr(self, '_PLUGIN_STAGE_SOURCES', [])
 
+    @property
+    def stage_packages(self):
+        return self._stage_packages
+
+    @stage_packages.setter
+    def stage_packages(self, value):
+        self._stage_packages = value
```
kyrofa: Using this property and setter combo, I don't see how this changes from using a simple property as we were doing. Am I missing something?

snapcraft/__init__.py
kyrofa: Ah, there it is. Nevermind.

snapcraft/internal/meta.py
```diff
@@ -253,33 +254,41 @@ def _write_wrap_exe(self, wrapexec, wrappath,
...
-                                            self._arch_triplet)
-        assembled_env = assembled_env.replace(self._snap_dir, '$SNAP')
+        # binaries are linked with `nodefaultlib` but we still do
+        # not want to leak PATH or other environment variables
+        # that would affect the applications view of the classic
+        # environment it is dropped into.
         replace_path = r'{}/[a-z0-9][a-z0-9+-]*/install'.format(
-            self._parts_dir)
-        assembled_env = re.sub(replace_path, '$SNAP', assembled_env)
+            re.escape(self._parts_dir))
```
kyrofa: This is used a few times. `re.compile()` might be a good idea.

snapcraft/internal/meta.py
```diff
@@ -253,33 +254,41 @@ def _write_wrap_exe(self, wrapexec, wrappath,
...
         with open(wrappath, 'w+') as f:
-            f.write(script)
+            print('#!/bin/sh', file=f)
+            if assembled_env:
+                print('{}'.format(assembled_env), file=f)
+                print('export LD_LIBRARY_PATH=$SNAP_LIBRARY_PATH:'
+                      '$LD_LIBRARY_PATH', file=f)
+            if cwd:
+                print('{}'.format(cwd), file=f)
+            print('exec {} {}'.format(executable, args), file=f)
```
kyrofa: Why not just collect all the lines in a list and finally `f.write('\n'.join(lines))` at the end?

snapcraft/plugins/python.py
```diff
@@ -163,24 +179,56 @@ def clean_pull(self):
...
+        # since we are using an independent env we need to export this too
+        # TODO: figure out if we can move back to common.run
+        env['SNAPCRAFT_STAGE'] = self.project.stage_dir
+        env['SNAPCRAFT_PART_INSTALL'] = self.installdir
 
         args = ['pip', 'setuptools', 'wheel']
 
+        pip_command = [self._get_python_command(), '-m', 'pip']
+
+        # if python_command it is not from stage we don't have pip
```
kyrofa: s/if python_command it is/if python_command is/. Although from the comment I'm not sure how that relates to the setting of `PYTHONHOME`.

snapcraft/plugins/python.py
```diff
@@ -163,24 +179,56 @@ def clean_pull(self):
...
         env = os.environ.copy()
+        env['PYTHONUSERBASE'] = self.installdir
+        if self._get_python_command().startswith(self.project.stage_dir):
+            env['PYTHONHOME'] = os.path.join(self.project.stage_dir, 'usr')
+        else:
+            env['PYTHONHOME'] = os.path.join(self.installdir, 'usr')
+
+        env['PATH'] = '{}:{}'.format(
+            os.path.join(self.installdir, 'usr', 'bin'),
+            os.path.expandvars('$PATH'))
```
kyrofa: Huh, I didn't know about `os.path.expandvars()`, nice.

snapcraft/shutil.py
kyrofa: Why not use the real shutil.which() with assemble_env()'s PATH? If we want to keep this, can we name the file something else? I'm afraid shutil will cause confusion later.

snapcraft/internal/meta.py
sergiusens: I guess this boils down to personal preference, doesn't it?

snapcraft/shutil.py
sergiusens: this is a common practice though.

sergiusens: also, we want the specific env to take precedence (PATH)

snapcraft/plugins/python.py
sergiusens: this can be clarified

snapcraft/shutil.py
sergiusens: `assembled_env`'s `PATH` is a list with multiple entries of `PATH`

snapcraft/internal/meta.py
kyrofa: Indeed it does. Feel free to ignore.

battlemidget:
I attempted to get my conjure-up snap working with this branch on trusty. This is the packaged snap contents of that:

http://paste.ubuntu.com/23907436/

I was able to install without an issue this time, however, here is the error i received:

`/snap/conjure-up/x1/wrappers/conjure-up: 14: exec: /snap/conjure-up/x1/bin/conjure-up: not found`

which goes inline with the contents of the snap. 

On a plus though, my original issue with the systemd services not being installed have been resolved \o/

For reference here is my snapcraft im using https://github.com/conjure-up/conjure-up/blob/master/snap/snapcraft.yaml

battlemidget:
I've successfully tested this branch and was able to build https://github.com/conjure-up/conjure-up on Trusty with huge success.

snapcraft/internal/meta.py
```diff
@@ -253,33 +254,41 @@ def _write_wrap_exe(self, wrapexec, wrappath,
...
+        # that would affect the applications view of the classic
+        # environment it is dropped into.
+        replace_path = re.compile(r'{}/[a-z0-9][a-z0-9+-]*/install'.format(
+            re.escape(self._parts_dir)))
+        if self._config_data['confinement'] == 'classic':
+            assembled_env = None
+        else:
+            assembled_env = common.assemble_env()
+            assembled_env = assembled_env.replace(self._snap_dir, '$SNAP')
+            assembled_env = re.sub(replace_path, '$SNAP', assembled_env)
```
kyrofa: You probably want `replace_path.sub('$SNAP', assembled_env)` here, no?

sergiusens: ah, you should of been more specific; the current mechanism works though :-P

snapcraft/internal/meta.py
```diff
@@ -253,33 +254,41 @@ def _write_wrap_exe(self, wrapexec, wrappath,
...
+
         executable = '"{}"'.format(wrapexec)
-        if shebang is not None:
+
+        if shebang:
+            if shebang.startswith('/usr/bin/env '):
+                shebang = shell_utils.which(shebang.split()[1])
             new_shebang = re.sub(replace_path, '$SNAP', shebang)
```
kyrofa: Right here too.

snapcraft/plugins/python.py
```diff
@@ -205,7 +205,10 @@ def _install_pip(self, download):
 
         pip_command = [self._get_python_command(), '-m', 'pip']
 
-        # if python_command it is not from stage we don't have pip
+        # If python_command it is not from stage we don't have pip, which means
+        # we are going to need to resort to the pip installed on the system
+        # that came from build-packages. This shouldn't be a problem as
+        # stage-packages and build-packages should match.
```
kyrofa: :+1: thank you. Though the typo is still there: s/python_command it is not/python_command is not/

PR #1095, opened 2017-01-31
Plainbox providers run validate

jocave:
Modify the plainbox-provider plugin to run the "validate" command during the build step to ensure buggy providers don't make it in to checkbox snap packages.

As providers can reference jobs from other providers (i.e. have a dependency on that provider) the environment has a variable PROVIDERPATH populated. Coupled with appropriate use of the "after" keyword this allows "validate" steps to identify that jobs are defined correctly. An integration test to confirm this functionality is included.

LP: [#1660692](https://bugs.launchpad.net/snapcraft/+bug/1660692)


integration_tests/test_plainbox_provider_plugin.py
```diff
@@ -32,3 +32,7 @@ def test_snap_simple_provider(self):
                 project_dir, 'stage', 'providers', 'simple-plainbox-provider',
                 'plainbox-provider-simple.provider'),
             FileExists())
+
+    def test_snap_provider_with_deps(self):
+        project_dir = 'plainbox-provider-with-deps'
+        self.run_snapcraft('prime', project_dir)
```
elopio: Please make a comment here to justify that the test doesn't have an assertion.
Something like:
# If the dependency can't be found, the prime step will fail.

What about a test for a failed validation?

snapcraft/plugins/plainbox_provider.py
```diff
@@ -43,6 +44,13 @@ def __init__(self, name, options, project):
 
     def build(self):
         super().build()
+        env = os.environ.copy()
+        provider_stage_dir = os.path.join(self.project.stage_dir, "providers")
```
elopio: please use single quotes

elopio: oh, I see there is a combination of single and double quotes in this function. Now that you are touching this code, I would appreciate if you update all of them to single quotes :)

snapcraft/plugins/plainbox_provider.py
```diff
@@ -43,6 +44,13 @@ def __init__(self, name, options, project):
...
+        if os.path.exists(provider_stage_dir):
+            providerdirs = [os.path.join(provider_stage_dir, provider)
+                            for provider in os.listdir(provider_stage_dir)]
+            env['PROVIDERPATH'] = ':'.join(providerdirs)
+        self.run(["python3", "manage.py", "validate"], env=env)
```
elopio: There is a unit test that checks these calls. You can add validate in there.
And you can also add a couple of unit tests for the env, one with provider_stage_dir, and one without it.

snapcraft/plugins/plainbox_provider.py
```diff
@@ -43,6 +44,13 @@ def __init__(self, name, options, project):
 
     def build(self):
         super().build()
+        env = os.environ.copy()
+        provider_stage_dir = os.path.join(self.project.stage_dir, "providers")
+        if os.path.exists(provider_stage_dir):
+            providerdirs = [os.path.join(provider_stage_dir, provider)
```
elopio: nit: provider_dirs.

sergiusens:
Mind upating the integration test name to @elopio's new world order of `<plugin-name>-test` seen in #1105 ?

sergiusens:
Just a reminder that if you want this in 2.27 you will need to update by tomorrow

PR #1096, opened 2017-01-31
schema,copy plugin: better errors when item has no value.

kyrofa:
Snapcraft supports a few different objects in its schema that don't enforce the actual presence of a vaue in its items. Of interest in this PR is the `organize` keyword in the core, as well as the `files` keyword in the copy plugin.

This PR fixes LP: [#1587455](https://bugs.launchpad.net/snapcraft/+bug/1587455) by modifying the schema of both the core and the copy plugin to enforce both the type and the presence of values.

### DIFF
--- a/snapcraft/_schema.py
+++ b/snapcraft/_schema.py
@@ -15,30 +15,15 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import contextlib
 import os
 
 import jsonschema
 import yaml
 
-from snapcraft.internal import common
-
-# dict of validator -> cause pairs. Wish jsonschema just gave us better
-# messages.
-_VALIDATION_ERROR_CAUSES = {
-    'maxLength': 'maximum length is {validator_value}',
-    'minLength': 'minimum length is {validator_value}',
-}
-
-
-class SnapcraftSchemaError(Exception):
-
-    @property
-    def message(self):
-        return self._message
-
-    def __init__(self, message):
-        self._message = message
+from snapcraft.internal import (
+    common,
+    errors
+)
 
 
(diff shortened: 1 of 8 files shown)


snapcraft/_schema.py
```diff
@@ -78,48 +63,4 @@ def validate(self):
...
-    """Attempt to determine a cause from validation error.
-
-    Returns:
-        A string representing the cause of the error (it may be empty if no
-        cause can be determined).
-    """
-
-    return _VALIDATION_ERROR_CAUSES.get(error.validator, '').format(
-        validator_value=error.validator_value)
+            raise errors.SnapcraftSchemaError.from_validation_error(e)
```
elopio: nice! please update the copyright year.

kyrofa: Ah, thanks. I'll wait to see if the integration tests pass (I changed some errors) first, then I'll push.

snapcraft/internal/project_loader.py
```diff
@@ -33,22 +33,13 @@
...
-from snapcraft._schema import Validator, SnapcraftSchemaError
+from snapcraft._schema import Validator
 from snapcraft.internal.parts import SnapcraftLogicError, get_remote_parts
 
 
 logger = logging.getLogger(__name__)
 
 
-@jsonschema.FormatChecker.cls_checks('file-path')
-def _validate_file_exists(instance):
```
elopio: good catch!

kyrofa:
> You can use a dash... for now ;)

Haha, you still haven't heard back?!

elopio:
nop. And I forgot to ping during my morning.

snapcraft/_schema.py
kyrofa: Done.

PR #1097, opened 2017-02-01
lifecycle: print the command needed to clean the dirty part

elopio:
LP: [#1590811](https://bugs.launchpad.net/snapcraft/+bug/1590811)

### DIFF
--- a/snapcraft/internal/lifecycle.py
+++ b/snapcraft/internal/lifecycle.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright (C) 2015, 2016 Canonical Ltd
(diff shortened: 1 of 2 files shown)


snapcraft/internal/lifecycle.py
```diff
@@ -234,8 +234,9 @@ def _handle_dirty(self, part, step, dirty_report):
                         humanized_options, pluralized_connection))
 
             message_components.append(
-                "\nPlease clean that part's {!r} step in order to "
-                'continue'.format(step))
+                "\nIn order to continue, please clean that part's {0!r} step "
+                "running:\n"
```
sergiusens: I think this should be `"by running:\n"` also, it may be better to put it on the same line (no `\n`)

elopio:
@sergiusens done. Thanks for reviewing.

PR #1099, opened 2017-02-01
catkin plugin: don't pass args to setup.sh.

kyrofa:
Currently, apps created in snaps that use the Catkin plugin source ROS's setup.sh. However, this means that the setup.sh attempts to parse any command-line parameters passed to the app in question, which is never desired and will break things most of the time parameters are actually used.

This PR fixes LP: [#1660852](https://bugs.launchpad.net/snapcraft/+bug/1660852) by by saving off any parameters, zeroing them right before sourcing setup.sh, and restoring them before execing the binary in question.

Thanks to @didrocks for the initial approach, though I doubt you thought it'd be this yucky, eh?

### DIFF
--- a/snapcraft/plugins/catkin.py
+++ b/snapcraft/plugins/catkin.py
@@ -41,9 +41,10 @@ import glob
 import os
 import tempfile
 import logging
-import shutil
 import re
+import shutil
 import subprocess
+import textwrap
 
 import snapcraft
 from snapcraft import (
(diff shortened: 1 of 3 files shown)


kyrofa:
See the output in the bug. With this PR, you can see that args make it to the actual app in question instead of the setup.sh:

```
$ ros-example.launch-project --help
Usage: roslaunch [options] [package] <filename> [arg_name:=value...]
       roslaunch [options] <filename> [<filename>...] [arg_name:=value...]

If <filename> is a single dash ('-'), launch XML is read from standard input.

Options:
  -h, --help            show this help message and exit
  --files               Print list files loaded by launch file, including
                        launch file itself
  --args=NODE_NAME      Print command-line arguments for node
  --nodes               Print list of node names in launch file
  --find-node=NODE_NAME
                        Find launch file that node is defined in
  -c NAME, --child=NAME
                        Run as child service 'NAME'. Required with -u
  --local               Do not launch remote nodes
  --screen              Force output of all local nodes to screen
  -u URI, --server_uri=URI
                        URI of server. Required with -c
  --run_id=RUN_ID       run_id of session. Required with -c
  --wait                wait for master to start before launching
  -p PORT, --port=PORT  master port. Only valid if master is launched
  --core                Launch core services only
  --pid=PID_FN          write the roslaunch pid to filename
  -v                    verbose printing
  --dump-params         Dump parameters of all roslaunch files to stdout
  --skip-log-check      skip check size of log folder
  --ros-args            Display command-line arguments for this launch file
  --disable-title       Disable setting of terminal title
  -w NUM_WORKERS, --numworkers=NUM_WORKERS
                        override number of worker threads. Only valid for core
                        services.
  -t TIMEOUT, --timeout=TIMEOUT
                        override the socket connection timeout (in seconds).
                        Only valid for core services.
```

snapcraft/plugins/catkin.py
```diff
@@ -195,13 +196,37 @@ def env(self, root):
...
+        # restore all args for the wrapper's `exec` line.
+        script = textwrap.dedent('''
+        if [ -e {0} ]; then
+            # Shell quote arbitrary string by replacing every occurrence of '
+            # with '\\'', then put ' at the beginning and end of the string.
+            # Prepare yourself, fun regex ahead.
+            quote()
+            {{
+                for i; do
+                    printf %s\\\\n "$i" | sed "s/\'/\'\\\\\\\\\'\'/g;1s/^/\'/;\$s/\$/\' \\\\\\\\/"
```
elopio: :sob:

kyrofa: Yeah, beautiful eh?

kyrofa:
> What about a regression integration test?

Really the only thing I could test is the presence of the script in the wrapper, which I'm doing (in some sense) in the unit test. Ideally I'd actually run the binary in the snaps test, but it's not run everywhere. Is that okay?

snapcraft/plugins/catkin.py
sergiusens: is this some mogragic? :wink:

elopio:
well, as building a ros snap is way too slow, I was thinking of uploading somewhere a snapcraft.yaml, and documenting in the manual test files how to build and run it to confirm that the bug is fixed.

kyrofa:
> I was thinking of uploading somewhere a snapcraft.yaml, and documenting in the manual test files how to build and run it to confirm that the bug is fixed.

Super easy-- done.

PR #1100, opened 2017-02-02
repo: remove symlinks to libc.

kyrofa:
libc is never staged, as it's always contained in the core snap. However, currently snapcraft leaves symlinks to its libraries alone and intact, which causes issues when it comes to content sharing and building against tarred staging areas. It also doesn't add any functionality, as those libraries are in the standard library paths anyway.

Instead of allowing those symlinks to dangle, this PR fixes LP: [#1658774](https://bugs.launchpad.net/snapcraft/+bug/1658774) by removing them.


integration_tests/test_staging_links_to_libc.py
```diff
@@ -0,0 +1,41 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import subprocess
+
+import integration_tests
+
+
+class StageLibcLinksTestCase(integration_tests.TestCase):
```
sergiusens: Can this class be put into one of the existing modules?

kyrofa: Certainly-- test_stage.py it is!

kyrofa: Done.

PR #1103, opened 2017-02-02
meta: support for the environment keyword.

sergiusens:
LP: #1583259

### DIFF
--- a/schema/snapcraft.yaml
+++ b/schema/snapcraft.yaml
@@ -80,6 +80,11 @@ properties:
     type: object
     description: environment entries for the snap as a whole
     minItems: 1
+    additionalProperties:
+      anyOf:
+        - type: string
+          minLength: 1
+        - type: number
   apps:
     type: object
     additionalProperties: false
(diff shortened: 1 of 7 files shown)


kyrofa:
\o/

schema/snapcraft.yaml
```diff
@@ -76,6 +76,10 @@ properties:
   epoch:
     description: the snap epoch, used to specify upgrade paths
     format: epoch
+  environment:
+    type: object
+    description: environment entries for the snap as a whole
+    minItems: 1
```
kyrofa: We should probably catch empty values here. Perhaps by being more explicit:

```
environment:
    type: object
    description: environment entries for the snap as a whole
    minItems: 1
    additionalProperties:
        anyOf:
            - type: string
              minLength: 1
            - type: number
```

schema/snapcraft.yaml
```diff
@@ -134,6 +138,10 @@ properties:
             items:
               type: string
               pattern: "^[a-zA-Z0-9][-_.a-zA-Z0-9]*$"
+          environment:
+            type: object
+            description: environment entries for the specific app.
+            minItems: 1
```
kyrofa: We should probably catch empty values here as well. You might consider putting this into definitions so as to not duplicate.

kyrofa: Err... sorry, the definitions aren't working without #1059 .

PR #1105, opened 2017-02-04
tests: rename the integration test snaps

elopio:
LP: [#1661893](https://bugs.launchpad.net/snapcraft/+bug/1661893)


elopio:
I said a few weeks ago to Sergio that we must change this. And here it is.
Next thing I think should be to make sure that each plugin has a hello snap.

PR #1109, opened 2017-02-06
meta: allow `post-stop-command` for an app entry in `apps`.

elespike:
LP: #1662240


schema/snapcraft.yaml
```diff
@@ -106,6 +106,9 @@ properties:
           stop-command:
             type: string
             description: command executed to stop a service
+          post-stop-command:
+            type: string
+            description: command executed after stopping a service
```
elopio: oh, wait, I think you are missing the dependencies. Take a look a few lines up:

```
dependencies:
  stop-command: ["daemon"]
```

Just as stop-command, post-stop-command doesn't make sense if this is not a daemon. So there's a missing test here too. Can you please write a test that checks that you can't write a snapcraft.yaml with `post-stop-command` but without `daemon`?

elespike: Good catch! I've made the change and added tests (for stop-command as well).

snapcraft/tests/test_project_loader.py
```diff
@@ -1838,10 +1838,44 @@ def test_valid_app_daemons(self):
...
+        self.assertEqual(
+            "The 'apps/service1' property does not match the required schema: "
+            "'daemon' is a dependency of 'stop-command'",
+            str(raised))
+
+        self.data['apps'] = {
+            'service1': {
+                'command': 'binary1',
+                'post-stop-command': 'binary1 --post-stop',
+            },
```
elopio: @elespike thank you! It would be better if you use test scenarios here.

basically, you make a separate test case class for these two tests and you define a class variable called scenarios, like:

```
scenarios = [
    ('stop-command', {apps: {
        'service1': {...}
    )
    ('post-stop-command', {apps: {
        'service1': {...}
    )
]
```

Search for the word `scenarios` in other test files to get an idea of how we use them, and I'm available in https://rocket.ubuntu.com/channel/snapcraft if you need a hand.

elespike: Thanks for the help!

PR #1112, opened 2017-02-06
core: setup core to support classic confinement.

sergiusens:
When doing ci, most systems won't have /snap/core/current as they
are not snappy capable. The work here allows for one to export
SNAPCRAFT_SETUP_CORE to get the core snap and unsquash it
onto the expected locations.

LP: #1662351


sergiusens:
This branch still depends on whatever comes out of #940

integration_tests/test_stage.py
```diff
@@ -64,3 +66,26 @@ def test_staging_libc_links(self):
...
+                self.run_snapcraft(['stage'], project_dir)
+            except subprocess.CalledProcessError as e:
+                pass
+            else:
+                self.fail(
+                    'This should fail as SNAPCRAFT_SETUP_CORE is not set')
+
+        # Now we set the required environment variable
+        self.useFixture(fixtures.EnvironmentVariable(
+                'SNAPCRAFT_SETUP_CORE', '1'))
```
elopio: so, how do we set this variable in launchpad?

snapcraft/__init__.py
```diff
@@ -282,12 +282,14 @@
 from snapcraft._baseplugin import BasePlugin        # noqa
 from snapcraft._options import ProjectOptions       # noqa
 from snapcraft._help import topic_help              # noqa
+# TODO migrate to snapcraft.store
```
elopio: This is the type of things that we read a month from now and we no longer remember why they made sense before and not now.
Can you please explain why, and add the date and your name?

snapcraft/internal/cache/_snap.py
```diff
@@ -76,8 +76,24 @@ def prune(self, *, keep_revision):
                         'Unable to purge snap {}.'.format(cached_snap))
         return pruned_files_list
 
+    def get(self, *, snap_name, revision=None):
+        cached_snaps = []
+        for f in os.listdir(self.snap_cache_dir):
+            if f.startswith(snap_name):
```
elopio: isn't this risky with snaps whose name is a subset of other snaps? Like hello and hello-world?

snapcraft/internal/lifecycle.py
```diff
@@ -112,6 +118,39 @@ def execute(step, project_options, part_names=None):
             'type': config.data.get('type', '')}
 
 
+def _setup_core(deb_arch):
+    core_path = os.path.join(os.path.sep, 'snap', 'core', 'current')
+    if os.path.exists(core_path):
+        logger.debug('{!r} already exists, skipping core setup'.format(
+            core_path))
+        return
```
elopio: This if condition deserves a unit test, I think.

snapcraft/internal/lifecycle.py
```diff
@@ -112,6 +118,39 @@ def execute(step, project_options, part_names=None):
...
+        with tempfile.TemporaryDirectory() as d:
+            snap_name = os.path.join(d, 'core_stable_{}.snap'.format(deb_arch))
+            snapcraft.download('core', 'stable', snap_name, deb_arch)
+            # FIXME there is a chance of incorrecly caching the wrong
+            # revision here.
+            snap_cache.cache(snap_name, revision=core_revision)
+    # This should now be a hit.
+    core_snap = snap_cache.get(snap_name='core', revision=core_revision)
+    if not core_snap:
+        raise RuntimeError('Error while retrieving the core snap')
```
elopio: missing unit test.

snapcraft/internal/lifecycle.py
```diff
@@ -112,6 +118,39 @@ def execute(step, project_options, part_names=None):
...
+    core_snap = snap_cache.get(snap_name='core', revision=core_revision)
+    if not core_snap:
+        with tempfile.TemporaryDirectory() as d:
+            snap_name = os.path.join(d, 'core_stable_{}.snap'.format(deb_arch))
+            snapcraft.download('core', 'stable', snap_name, deb_arch)
+            # FIXME there is a chance of incorrecly caching the wrong
+            # revision here.
+            snap_cache.cache(snap_name, revision=core_revision)
+    # This should now be a hit.
+    core_snap = snap_cache.get(snap_name='core', revision=core_revision)
```
elopio: maybe, you could simplify this with a try catch. Make get raise an exception instead of return None.

```
try:
    core_snap = cache.get...
except NotFoundInCache:
    core_snap = download  # Raise if fails to download.
    save_to_cache(core_snap)
```

I don't see why you need to call core_snap twice, and check if not core_snap twice.

snapcraft/internal/lifecycle.py
```diff
@@ -112,6 +118,39 @@ def execute(step, project_options, part_names=None):
...
+    snap_cache.prune(keep_revision=core_revision)
+
+    # And try to get this revision.
+    core_snap = snap_cache.get(snap_name='core', revision=core_revision)
+    if not core_snap:
+        with tempfile.TemporaryDirectory() as d:
+            snap_name = os.path.join(d, 'core_stable_{}.snap'.format(deb_arch))
+            snapcraft.download('core', 'stable', snap_name, deb_arch)
+            # FIXME there is a chance of incorrecly caching the wrong
+            # revision here.
```
elopio: you are not really explaining here why. And if it's a FIXME, it should have a bug reported to fix it later.

integration_tests/test_stage.py
kyrofa: It sounds like cjwatson will be mounting the core snap into place, so perhaps we don't need to worry about setting this in LP.

sergiusens: I already discussed this with him. If they need it, they will set it as they have done with all other env var enabled features (as it is their preference)

snapcraft/__init__.py
sergiusens: I'll add a bug but name and date are the reason we use a VCS ;-)

snapcraft/internal/cache/_snap.py
sergiusens: this isn't a problem once I merge the new work on delta branches.

snapcraft/internal/lxd.py
```diff
@@ -50,7 +51,8 @@ def _pull_file(self, src, dst):
                     '{}/{}'.format(self._container_name, src), dst])
 
     def _container_run(self, cmd):
-        check_call(['lxc', 'exec', self._container_name, '--'] + cmd)
+        check_call(['lxc', 'exec', self._container_name, '--'] + cmd,
+                   env=self._env)
```
kyrofa: This sets the environment on the _host_ for this command, doesn't it? Does that get propagated into the `lxc exec`, or should this be set within the command itself?

snapcraft/internal/lifecycle.py
```diff
@@ -112,6 +118,39 @@ def execute(step, project_options, part_names=None):
...
+            snap_cache.cache(snap_name, revision=core_revision)
+    # This should now be a hit.
+    core_snap = snap_cache.get(snap_name='core', revision=core_revision)
+    if not core_snap:
+        raise RuntimeError('Error while retrieving the core snap')
+
+    # Now unpack
+    logger.info('Setting up {!r} in {!r}'.format(core_snap, core_path))
+    check_call(['sudo', 'mkdir', '-p', os.path.dirname(core_path)])
+    output = check_output(['sudo', 'unsquashfs', '-d', core_path, core_snap])
```
kyrofa: I'm curious: why unsquash versus mounting? Is that not possible from within lxc confinement?

sergiusens:
```
sergiusens@snapcraft:~/vim$ sudo mount /home/sergiusens/.cache/snapcraft/projects/snapcraft-core/snap_hashes/amd64/af028f2f34d9a0fbc399a36d1809e509abba1258ec9fe15d998fc87bcb779c90bc1fae7d857a3bab9eb710ddc7f4bfaa /snap/core/current
mount: /snap/core/current: mount failed: Unknown error -1
sergiusens@snapcraft:~/vim$ file /home/sergiusens/.cache/snapcraft/projects/snapcraft-core/snap_hashes/amd64/af028f2f34d9a0fbc399a36d1809e509abba1258ec9fe15d998fc87bcb779c90bc1fae7d857a3bab9eb710ddc7f4bfaa
/home/sergiusens/.cache/snapcraft/projects/snapcraft-core/snap_hashes/amd64/af028f2f34d9a0fbc399a36d1809e509abba1258ec9fe15d998fc87bcb779c90bc1fae7d857a3bab9eb710ddc7f4bfaa: Squashfs filesystem, little endian, version 4.0, 79831621 bytes, 12340 inodes, blocksize: 131072 bytes, created: Fri Jan 13 20:42:12 2017
```

sergiusens: `mount` is reserved in containers and only allowed if there is a fuse interface for it

kyrofa: Yeah that sounds familiar, thanks for the sanity check.

snapcraft/_store.py
```diff
@@ -623,11 +623,23 @@ def close(snap_name, channel_names):
...
-def download(snap_name, channel, download_path, arch):
-    """Download snap from the store to download_path"""
+def download(snap_name, channel, download_path, arch, but_hash=''):
+    """Download snap from the store to download_path.
+    :param str snap_name: The snap name to download.
+    :param str channel: the channel to get the snap from.
+    :param str download_path: the path to write the downloaded snap to.
+    :param str arch: the architecture of the download as a deb arch.
+    :param str but_hash: a sha3_384 hash to indicate that the download should
+                         not take place.
```
kyrofa: This variable name and description don't make sense to me. I don't understand what it does.

snapcraft/internal/lifecycle.py
```diff
@@ -112,6 +118,42 @@ def execute(step, project_options, part_names=None):
...
+        # The current hash matches the filename
+        current_hash = os.path.splitext(os.path.basename(core_snap))[0]
+    else:
+        current_hash = ''
+
+    with tempfile.TemporaryDirectory() as d:
+        download_path = os.path.join(d, 'core.snap')
+        download_hash = snapcraft.download('core', 'stable', download_path,
+                                           deb_arch, but_hash=current_hash)
+        print('download', download_hash, 'current', current_hash)
```
kyrofa: This should probably be removed.

snapcraft/internal/lxd.py
```diff
@@ -59,6 +59,9 @@ def _create_container(self):
                 'lxc', 'launch', '-e',
                 'ubuntu:xenial/{}'.format(self._project_options.deb_arch),
                 self._container_name])
+            check_call([
+                'lxc', 'config', 'set', self._container_name,
+                'environment.SNAPCRAFT_SETUP_CORE', '1'])
```
kyrofa: Much better.

snapcraft/internal/lifecycle.py
```diff
@@ -112,6 +118,42 @@ def execute(step, project_options, part_names=None):
...
+        if download_hash != current_hash:
+            snap_cache.cache(snap_filename=download_path)
+            snap_cache.prune(deb_arch=deb_arch, keep_hash=download_hash)
+
+    core_snap = snap_cache.get(deb_arch=deb_arch)
+
+    # Now unpack
+    logger.info('Setting up {!r} in {!r}'.format(core_snap, core_path))
```
kyrofa: Can we add "may prompt for sudo" here?

sergiusens: I will let you log a bug so we do the same thing for every place we already call sudo ;-)

sergiusens:
Addressing your points:

*Make the core mount point configurable*

This was my only standing thought except it is not something I like... :-/

* Change the rmdir/mkdir/unsquashfs to not use sudo unless it gets a permissions error*

It needs to have the right permissions, so this is not a good idea.

*Create a unit test that creates a "core" snap that contains a single file, place it in the cache, and then you should be able to "setup core" and verify that it happened as expected.*

It seems too much side stepping just to run unsquashfs (and the integration test already does this).

sergiusens:
@elopio @elopio @elopio!!!!!

elopio:
It's failing everywhere! 
snapcraft.tests.test_lifecycle.CoreSetupTestCase.test_core_setup_skipped_if_core_exists
OSError: classic confinement requires the core snap to be installed. Install it by running `snap install core`.

snapcraft/tests/test_lifecycle.py
```diff
@@ -607,3 +617,109 @@ def test_pull_is_dirty_if_target_arch_changes(
...
+        core_path_mock = patcher.start()
+        core_path_mock.return_value = self.core_path
+        self.addCleanup(patcher.stop)
+
+        self.tempdir = os.path.join(self.path, 'tmpdir')
+        patcher = mock.patch('snapcraft.internal.lifecycle.TemporaryDirectory')
+        self.tempdir_mock = patcher.start()
+        self.addCleanup(patcher.stop)
+
+        # Create a fake sudo that just echos
```
elopio: this looks like a good fixture for fixture_setup.py

elopio:
I'm +1, once it's all green. Thanks Sergio.

sergiusens:
@elopio then add an approve so i don't need to ping you again, or is that what you want?

elopio:
This +1 is a high level of trust <3 :D

PR #1113, opened 2017-02-07
repo: cache stage packages for faster future use.

kyrofa:
Snapcraft currently caches stage packages, but doesn't actually use them. This PR fixes LP: [#1661383](https://bugs.launchpad.net/snapcraft/+bug/1661383) by using them, and making sure cleanbuild can use the cache as well.

The way this ended up being done isn't ideal. Since `apt.Cache().fetch_archives()` seems to mangle (some) package names for no apparent reason, it's impossible to fetch stage packages with it and later determine which packages were pulled as the filenames on disk don't match up to the filenames in the archive. The settled-upon solution was to pull stage-packages using `apt.package.Version().fetch_binary()`. This works well and actually simplifies the caching mechanism, but has a few drawbacks:

- It prints directly to stdout when a package is already cached.
- Fetching stage packages one-by-one means one loses a sense of how long the overall process will actually take.

Note also that this PR **copies** the cache into the cleanbuild container as opposed to bind-mounting read-only. This was done so the caching code was more robust. Otherwise it would need to account for being unable to do the following things:

- Being unable to clean the cache when sources changes
- Being unable to update the index
- Being unable to fetch packages

And have backup plans. Such code would be complex and difficult to test, and it seemed it wasn't worth it for a relatively small efficiency gain, particularly considering that ideally the cache would be editable from the container anyway, that just doesn't seem possible today without modifications to /`etc/subgid`. Of course this is up for debate!


kyrofa:
Actually... would it be insane if we pulled the cached packages back from the container before stopping it? Then they can essentially modify it, albeit inefficiently.

snapcraft/internal/repo.py
```diff
@@ -198,106 +189,85 @@ def _setup_apt(self, download_dir):
...
-            apt_cache = apt.Cache(rootdir=rootdir, memonly=True)
+            apt_cache = self._setup_apt(cache_dir)
             apt_cache.open()
-            self._restore_cached_packages(apt_cache.get_changes(),
-                                          package_cache_dir, download_dir)
+
             yield apt_cache
-            self._store_cached_packages(package_cache_dir, download_dir)
+
+            apt_cache.close()
```
sergiusens: this should be in a `finally` clause with a check if it is open.

integration_tests/__init__.py
```diff
@@ -51,9 +53,39 @@ def setUp(self):
...
             'XDG_CONFIG_HOME', os.path.join(self.path, '.config')))
         self.useFixture(fixtures.EnvironmentVariable(
+            'XDG_CACHE_HOME', os.path.join(self.path, '.cache')))
+        self.useFixture(fixtures.EnvironmentVariable(
             'XDG_DATA_HOME', os.path.join(self.path, 'data')))
         self.useFixture(fixtures.EnvironmentVariable('TERM', 'dumb'))
 
+        patcher = mock.patch(
+            'xdg.BaseDirectory.xdg_config_home',
+            new=os.path.join(self.path, '.config'))
```
elopio: so, just setting HOME to something didn't work?

kyrofa: No, `xdg.BaseDirectory` has those as _variables_! They're evaluated at import time, which is way before `setUp()` runs.

snapcraft/internal/repo.py
kyrofa: You mean in case it was closed in the `yield`?

sergiusens: what if an exception occurs during the `yield`, you should jump to `finally` and `.close` there which will happen no matter what

kyrofa: Yeah good call. Done.

kyrofa:
As a real-world test, note that Nextcloud builds/runs fine with this.

snapcraft/_options.py
```diff
@@ -177,14 +178,18 @@ def debug(self):
...
-                 target_deb_arch=None, debug=False):
+                 target_deb_arch=None, debug=False, cache_dir=None):
         # TODO: allow setting a different project dir and check for
         #       snapcraft.yaml
         self.__project_dir = os.getcwd()
         self.__use_geoip = use_geoip
         self.__parallel_builds = parallel_builds
         self._set_machine(target_deb_arch)
         self.__debug = debug
+        if cache_dir:
```
sergiusens: you should be using or at least integrating with `snapcraft.internal.cache`

kyrofa: Alright I removed this from `ProjectOptions` and simply made them optional keyword arguments in the relevant areas so the tests could hook into them.

kyrofa:
Closed in favor of #1122.

PR #1115, opened 2017-02-07
kernel plugin: fix kernel snap layout

lool:
Hi,

This fixes the kernel snap layout for firmware/ and modules/; I've build tested this against the ONL example linked in the bug and prime/ now properly contains a modules/ and firmware/ toplevel.

(Also small fix to list of required packages; these packages were missing on top of a clean Ubuntu 16.04 EC2 instance)

Cheers,
-- Loïc

### DIFF
--- a/snapcraft/plugins/kbuild.py
+++ b/snapcraft/plugins/kbuild.py
@@ -102,7 +102,7 @@ class KBuildPlugin(BasePlugin):
 
     def __init__(self, name, options, project):
         super().__init__(name, options, project)
-        self.build_packages.append('make')
+        self.build_packages.extend(['bc', 'gcc', 'make'])
 
         self.make_targets = []
         self.make_install_targets = ['install']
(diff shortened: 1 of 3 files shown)


lool:
Yup, sorry; I have just fixed the extraneous "lib" in the current path to match the code changes.

BTW, I've noticed that the kernel plugin is a bit fragile, it installs in the build() method and doesn't like when steps are rerun (e.g. if build doesn't complete, the next one will fail to overwrite the previous bits); it looks like it needs a larger refactoring

elopio:
@lool can you please also add a test to manual-tests.md so we don't regress?

sergiusens:
@lool you will need to reauthor your commits with your @canonical.com address or sign the CLA ;-)

lool:
I've done so now (with my @ubuntu.com address)

elopio:
@sergiusens an email in ubuntu.com doesn't mean that the agreement is signed, right?
We know lool so do we really need him to commit with the canonical email?

lool:
(I've also signed the CLA with @ubuntu.com now)

PR #1122, opened 2017-02-08
 repo: cache stage packages for faster future use.

kyrofa:
Snapcraft currently caches stage packages, but doesn't actually use them. This PR fixes LP: [#1661383](https://bugs.launchpad.net/snapcraft/+bug/1661383) by using them, and making sure cleanbuild can use the cache as well.

The way this ended up being done isn't ideal. Since `apt.Cache().fetch_archives()` seems to mangle (some) package names for no apparent reason, it's impossible to fetch stage packages with it and later determine which packages were pulled as the filenames on disk don't match up to the filenames in the archive. The settled-upon solution was to pull stage-packages using `apt.package.Version().fetch_binary()`. This works well and actually simplifies the caching mechanism, but has a few drawbacks:

- It prints directly to stdout when a package is already cached.
- Fetching stage packages one-by-one means one loses a sense of how long the overall process will actually take.

Note also that this PR **copies** the cache into the cleanbuild container as opposed to bind-mounting read-only. This was done so the caching code was more robust. Otherwise it would need to account for being unable to do the following things:

- Being unable to clean the cache when sources changes
- Being unable to update the index
- Being unable to fetch packages

And have backup plans. Such code would be complex and difficult to test, and it seemed it wasn't worth it for a relatively small efficiency gain, particularly considering that ideally the cache would be editable from the container anyway, that just doesn't seem possible today without modifications to /`etc/subgid`. Of course this is up for debate!


integration_tests/test_stage_package_cache.py
```diff
@@ -0,0 +1,58 @@
...
+        # linked) by comparing inodes.
+        cached_deb_inode = os.stat(cached_deb).st_ino
+        self.assertThat(cached_deb_inode, Equals(os.stat(staged_deb).st_ino))
+
+        # Now clean the part and pull again.
+        self.run_snapcraft('clean', 'dump')
+        self.run_snapcraft(['pull', 'oneflatwithstagepackages'], 'dump')
+
+        # Verify that the staged deb is _still_ the same one from the cache.
+        self.assertThat(cached_deb_inode, Equals(os.stat(staged_deb).st_ino))
```
elopio: :+1:

snapcraft/internal/cache/_cache.py
```diff
@@ -44,3 +44,11 @@ def __init__(self, *, project_name):
         super().__init__()
         self.project_cache_root = os.path.join(
             self.cache_root, 'projects', project_name)
+
+
+class SnapcraftStagePackageCache(SnapcraftCache):
```
sergiusens: At first I thought, is this going to be reusable? if not just implement this in `_apt`. Now I think you are forward thinking :smile:

snapcraft/internal/cache/_apt.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def __init__(self, *, sources_digest):
+        """Create a new AptStagePackageCache.
+
+        :param str sources_digest: Unique digest of the current apt sources.
+        """
+
+        super().__init__()
+        cache_base_dir = os.path.join(self.stage_package_cache_root, 'apt')
+        _clear_old_caches(cache_base_dir, sources_digest)
```
sergiusens: snap up a ros project and then one that uses the system and you roll without a cache, right?

Which is why `prune` could be implemented separately and later.

kyrofa: Great catch. I've removed the pruning code and logged LP: [#1663051](https://bugs.launchpad.net/snapcraft/+bug/1663051) so we can be smart about it.

snapcraft/internal/cache/_cache.py
kyrofa: Ha! I want that refactor to be as easy as possible :stuck_out_tongue: .

PR #1128, opened 2017-02-09
 tests: support bzr branches for external tests

elopio:
LP: [#1663294](https://bugs.launchpad.net/snapcraft/+bug/1663294)


external_snaps_tests/__main__.py
```diff
@@ -68,14 +74,27 @@ def _is_git(repo):
 
 def _git_clone(url, repo_branch=None):
     temp_dir = tempfile.mkdtemp(prefix='snapcraft-')
-    command = ['git', 'clone', url, temp_dir]
+    command = ['git', 'clone', '--progress', url, temp_dir]
```
kyrofa: Why `--progress` here? Is the git progress not otherwise reported in a subprocess exception?

elopio: I was hitting a problem where travis stayed 10 minutes doing the clone, and timed out. So I added verbosity to bzr and git to avoid it. It probably was a network problem, because it doesn't make sense that a clone of these repos would take that long, but still more output here is better.

I could move it to a separate branch if you prefer.

PR #1130, opened 2017-02-09
catkin plugin: produce build-ready staging area.

kyrofa:
Currently the Catkin plugin massages some ROS-specific files in order to get it to build from within the part's install directory. This works well enough, but produces a staging area that, if used to build more parts in the future, will attempt to use the old install directory.

This PR fixes LP: [#1663385](https://bugs.launchpad.net/snapcraft/+bug/1663385) by updating the Catkin plugin such that, once the build has completed, these files are massaged to be usable from the staging area.


snapcraft/plugins/catkin.py
```diff
@@ -348,36 +348,65 @@ def _prepare_build(self):
         # dependencies contain .cmake files pointing to system paths (e.g.
         # /usr/lib, /usr/include, etc.). They need to be rewritten to point to
         # the install directory.
-        def rewrite_paths(match):
+        def _new_path(path):
+            if not path.startswith(self.installdir):
+                return self.installdir + path
```
elopio: os.path.join?

kyrofa: This is actually because the `path` here is absolute. If I were to use `os.path.join` I'd have to do `path[1:]`. Is that preferable?

kyrofa: Note that it's [guaranteed to be absolute](https://github.com/snapcore/snapcraft/pull/1130/files#diff-3cc26b49a70ee86523598a96dd385301R363).

elopio: well, a comment would be nice.

elopio:
lgtm. It would be nicer to have another catkin expert to do these reviews, but with the tests passing I'm happy.

snapcraft/plugins/catkin.py
kyrofa: Sure! Done :smiley: .

kyrofa:
(I'm ignoring the tomcat failure in xenial-amd64).

PR #1133, opened 2017-02-10
tests: fix the test that was modifying the environment variables

elopio:
LP: [#1663396](https://bugs.launchpad.net/snapcraft/+bug/1663396)


kyrofa:
Nice find! Some handy grep work? I'll test in the morning to verify this fix

elopio:
I failed with grep because I was looking for PATH, not os.environ. I added an assertion on the setup test to fail when PATH was empty, so the test that was causing the problem was the one before the failure.

PR #1138, opened 2017-02-10
python plugin: exclude the RECORD files

elopio:
LP: [#1663739](https://bugs.launchpad.net/snapcraft/+bug/1663739)

### DIFF
--- a/integration_tests/snaps/python-with-two-parts/snap/snapcraft.yaml
+++ b/integration_tests/snaps/python-with-two-parts/snap/snapcraft.yaml
@@ -1,0 +1,16 @@
+name: python-with-two-parts
+version: 0
+summary: test the python plugin
+description: |
+  This is a snap with two python parts. It's a regression test for
+  https://bugs.launchpad.net/snapcraft/+bug/1663739
+confinement: strict
+grade: devel
+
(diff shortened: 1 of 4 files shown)


integration_tests/snaps/python-with-two-parts/snap/snapcraft.yaml
```diff
@@ -0,0 +1,16 @@
...
+confinement: strict
+grade: devel
+
+parts:
+  part1:
+    source: .
+    plugin: python
+  part2:
+    source: .
+    plugin: python
```
kyrofa: This will test the python2 bits. Should we also test python3?

elopio: The opposite, this tests python3. The issue is on pip, and on both py2 and py3 we are using the same pip, so I don't think running the test again gives us a lot of value.

It would be simple to add though, and just a couple of minutes more on the integration suite. If you want it I'll add it.

kyrofa: Err, right I knew that.

I'm good with this if you are.

PR #1140, opened 2017-02-15
[WIP] catkin plugin: support building with an underlay.

kyrofa:
Currently the Catkin plugin simply pulls down all dependencies of the workspace being built using rosdep. This doesn't work for the cases of trying to create a ROS system that utilizes content sharing, as it forces each snap to bundle their dependencies even when it's undesired.

This PR resolves LP: [#1664776](https://bugs.launchpad.net/snapcraft/+bug/1664776) by adding support for an `underlay` property to the Catkin plugin, which accepts a build-time underlay path for when the consumer snap is being built (which is searched for dependencies before fetching them), as well as a run-time underlay path for when the consumer snap is being run (so the workspace can be properly overlaid).

**Note:** This is the first of two PRs. The second adds a demo (and a snaps_test) that utilizes this. It might be helpful to see how this is used, though, so feel free to [check out the demo](https://github.com/kyrofa/snapcraft/tree/feature/1664776/catkin_underlay_snaps_tests/demos/shared-ros).


snapcraft/plugins/catkin.py
```diff
@@ -34,6 +34,19 @@
...
+      overlaying a workspace from another snap via content sharing. Made up of
+      two properties:
+      - build-path:
+        (string)
+        Build-time path to existing workspace to underlay the one being built,
+        for example '$SNAPCRAFT_STAGE/opt/ros/kinetic'.
+      - run-path:
+        (string)
+        Run-time path of the underlay workspace (e.g. a subdirectory of the
+        content interface's 'target' attribute.)
```
elopio: What about a link to the example here?

kyrofa: Sure, I can add that in the PR that adds the example.

snapcraft/plugins/catkin.py
```diff
@@ -196,37 +223,17 @@ def env(self, root):
...
+            script = self._source_setup_sh(root, None)
 
         # Each of these lines is prepended with an `export` when the
         # environment is actually generated. In order to inject real shell code
-        # we have to hack it in by appending it on the end of an item in the
-        # environment. FIXME: There should be a better way to do this.
-        env[-1] = env[-1] + script
+        # we have to hack it in by appending it on the end of an item already
+        # in the environment. FIXME: There should be a better way to do this.
+        env[-1] = env[-1] + '\n\n' + script
```
elopio: can we make this nicer now that we have environment vars in the yaml?

snapcraft/plugins/catkin.py
```diff
@@ -243,11 +250,37 @@ def pull(self):
...
                 'Unable to find package path: "{}"'.format(
                     self._ros_package_path))
 
+        # Validate the underlay. Note that this validation can't happen in
+        # __init__ as the underlay will probably only be valid once a
+        # dependency has been staged.
+        catkin = None
+        underlay = None
+        if self.options.underlay:
+            underlay = self.options.underlay['build-path']
```
elopio: I'm not sure I like calling this underlay. What about underlay_build_path?

snapcraft/plugins/catkin.py
kyrofa: That doesn't help us when building, though. It doesn't help us after the fact either as that functionality isn't exposed to plugins.

snapcraft/plugins/catkin.py
```diff
@@ -260,9 +293,12 @@ def pull(self):
...
         rosdep.setup()
 
+        self._setup_dependencies(rosdep, catkin)
+
+    def _setup_dependencies(self, rosdep, catkin):
         # Parse the Catkin packages to pull out their system dependencies
-        system_dependencies = _find_system_dependencies(self.catkin_packages,
-                                                        rosdep)
+        system_dependencies = _find_system_dependencies(
+            self.catkin_packages, rosdep, catkin)
```
elopio: :1st_place_medal:

snapcraft/plugins/catkin.py
kyrofa: Sure, done. I did the same for the run path, below.

kyrofa:
Note that #1146 broke this. Don't merge for now.

snapcraft/plugins/catkin.py
```diff
@@ -121,10 +148,16 @@ def __init__(self, name, options, project):
         super().__init__(name, options, project)
         self.build_packages.extend(['libc6-dev', 'make'])
 
+        # roslib is the base requiremet to actually create a workspace with
```
sergiusens: requirement, but it is a TODO for next time :-)

PR #1141, opened 2017-02-15
parser: use the project loader code to find the origin's snapcraft.yaml

josepht:
Rather than roll our own snapcraft.yaml logic use the existing code in project_loader.

LP: #1665007


snapcraft/internal/parser.py
```diff
@@ -51,9 +51,16 @@
...
-from snapcraft.internal import log, repo, sources
-from snapcraft.internal.errors import SnapcraftError, InvalidWikiEntryError
-from snapcraft.internal.project_loader import replace_attr
+from snapcraft.internal import log, project_loader, repo, sources
+from snapcraft.internal.errors import (
+    SnapcraftError,
+    InvalidWikiEntryError,
+    SnapcraftEnvironmentError,
+)
+from snapcraft.internal.project_loader import (
```
sergiusens: why do you import project_loader above and then do this finer grained import instead of using `project_loader`?

josepht:
I'm not sure what you mean by 'formatted correctly'.

kyrofa:
> I'm not sure what you mean by 'formatted correctly'.

- For the bugref, gbp-dch will be looking for LP:<space>#<bugnum>
- For the PR/commit title, we've starting following the convention of structuring it as: <effected subsystem>: <what changed>. So this might be "parser: use project loader to find origin's snapcraft.yaml" or something similar. This allows for one to extract meaning from `git log --oneline`, for instance.

josepht:
Kyle: Thanks for explaining :)

FIxed.

sergiusens:
Ok, better, a test would be nice though

integration_tests/snap.snapcraft.yaml
```diff
@@ -0,0 +1,5 @@
+---
+description: A test part with snap/snapcraft.yaml
+maintainer: John Doe <john.doe@example.com>
+parts: [snap-testpart]
+origin: lp:~joetalbott/+junk/snap-testpart
```
kyrofa: Does this make anyone else a tad nervous? I know this is the pattern followed elsewhere in the parser, where everything seems to be in @josepht's junk... but at least in my own experience, that's only for throw-away stuff. If this is accidentally deleted our tests start failing.

Is there a better place to put this stuff?

josepht: I agree that it's suboptimal.  I'm working on an update to move all the parser tests to use git repos generated on-the-fly on the local disk.  Should speed up integration tests a bit too.

josepht: I've pushed an update now.

sergiusens:
nice on the locals, but how about you put all the wiki stubs in a `wiki` directory, kind of like what is done with `keys`

PR #1142, opened 2017-02-15
state: asset tracking - store versions of stage-packages.

josepht:
Keep track of the versions of the stage-packages and their dependencies during the pull step and store the list in the PullState object.

LP: #1547158


josepht:
Depends on https://github.com/snapcore/snapcraft/pull/1157

josepht:
The test failure looks unrelated to my changes.

kyrofa:
This PR seems to be missing a bug.

josepht:
https://bugs.launchpad.net/snapcraft/+bug/1547158 seems appropriate to me.

kyrofa:
> https://bugs.launchpad.net/snapcraft/+bug/1547158 seems appropriate to me.

Agreed! Thanks for adding the link.

snapcraft/internal/states/_pull_state.py
```diff
@@ -43,11 +43,13 @@ def _schema_properties():
...
     yaml_tag = u'!PullState'
 
-    def __init__(self, property_names, part_properties=None, project=None):
+    def __init__(self, property_names, part_properties=None, project=None,
+                 pkg_list=None):
         # Save this off before calling super() since we'll need it
         # FIXME: for 3.x the name `schema_properties` is leaking
         #        implementation details from a higher layer.
         self.schema_properties = property_names
+        self.pkg_list = pkg_list
```
kyrofa: Meh... I hate adding more properties to the state. Remember we're also planning on tracking VCS info here. Would it be better to nest all this asset tracking under a dict called `assets`? Then it would take a form like:

```python
assets = {
    'stage-packages': [<the package list you already have>],
    'source': <VCS stuff we have yet to do>,
}
```

That gives us room to expand to tracking more assets without further cluttering up the root of the state (since essentially `vars(state)` is written out as YAML). Thoughts?

snapcraft/internal/repo.py
```diff
@@ -288,8 +288,11 @@ def is_valid(self, package_name):
             return package_name in apt_cache
 
     def get(self, package_names):
+        pkg_list = []
         with self._apt.archive(self._cache.base_dir) as apt_cache:
-            self._get(apt_cache, package_names)
+            pkg_list = self._get(apt_cache, package_names)
```
kyrofa: Why not just `return self._get([...])` from within the context manager?

snapcraft/internal/pluginhandler/_stage_package_handler.py
```diff
@@ -81,10 +81,13 @@ def fetch(self):
         the cache.
         """
 
+        self.pkg_list = []
```
kyrofa: Why save this into the instance?

sergiusens:
@josepht can you fix the conflicts on this?

sergiusens:
@kyrofa just waiting for you :-)

integration_tests/test_pull_properties.py
```diff
@@ -45,6 +45,24 @@ def assert_expected_pull_state(self, project_dir):
 
         # Verify that the contents of the dependencies made it in as well.
         self.assertTrue('foo' in state.properties)
+        self.assertTrue(len(state.assets['stage-packages']) > 0)
```
kyrofa: The rest of this test actually verifies that the right thing is in here, this just verifies that there's _something_ there. That happens more explicitly in another test, but then why check anything here?

PR #1143, opened 2017-02-16
Switch Track and Arch in channel maps

josepht:
LP:#1665457


josepht:
Hold off on this until I can get 'Series' removed as well please.

PR #1145, opened 2017-02-17
tests: expect failures for the tests that can't be run in arm64

elopio:
LP: [#1662917](https://bugs.launchpad.net/snapcraft/+bug/1662917)

### DIFF
--- a/integration_tests/test_rust_plugin.py
+++ b/integration_tests/test_rust_plugin.py
@@ -24,11 +24,14 @@ import integration_tests
 
 class RustPluginTestCase(integration_tests.TestCase):
 
-    def setUp(self):
-        super().setUp()
+    def run_snapcraft(self, command, project_dir=None, debug=True):
         if snapcraft.ProjectOptions().deb_arch == 'armh64':
             # https://github.com/rust-lang/rustup.sh/issues/82
-            self.skipTest('The rustup does not support arm64')
+            self.expectFailure(
+                'The rustup script does not support arm64.',
+                super().run_snapcraft, command, project_dir, debug)
+
+        super().run_snapcraft(command, project_dir, debug)
 
     def test_stage_rust_plugin(self):
         self.run_snapcraft('stage', 'rust-hello')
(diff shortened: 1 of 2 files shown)


integration_tests/test_rust_plugin.py
```diff
@@ -18,11 +18,21 @@
...
 
 
 class RustPluginTestCase(integration_tests.TestCase):
 
+    def run_snapcraft(self, command, project_dir=None, debug=True):
+        if snapcraft.ProjectOptions().deb_arch == 'armh64':
+            # https://github.com/rust-lang/rustup.sh/issues/82
+            self.expectFailure(
+                'The rustup script does not support arm64.',
+                super().run_snapcraft, command, project_dir, debug)
```
elopio: This can be removed once we update the plugin to the new rustup.

snaps_tests/demos_tests/test_ros.py
```diff
@@ -29,6 +30,12 @@ class ROSTestCase(snaps_tests.SnapsTestCase):
     @skipUnless(linux_distribution()[2] == 'xenial',
                 'This test fails on yakkety LP: #1614476')
     def test_ros(self):
+        if snapcraft.ProjectOptions().deb_arch == 'arm64':
+            # https://bugs.launchpad.net/snapcraft/+bug/1662915
+            self.expectFailure(
+                'There are no ros-indigo packages in the ros archive',
```
kyrofa: To be explicit, there are no arm64 ROS Indigo packages in the ROS archive (not just no packages).

elopio:
There's something happening here that I don't yet understand :pensive:

elopio:
The only remaining issue here is https://bugs.launchpad.net/snapcraft/+bug/1665165, which is scheduled for the 2.28 milestone, so I think there's no need to skip it.

kyrofa:
> We can ask @kyrofa to work on his plan to switch to kinetic by default since we live on a xenial base.

Yeah, now that we have arm64 autopkgtests actually hooked up, I'll get on that.

PR #1148, opened 2017-02-17
kernel plugin: if dtb target == NULL and arch == (arm||arm64), build and install all dtbs

piso77:
If dtb target == NULL and arch == (arm||arm64), build and install all dtbs by default.

LP: #1665656


sergiusens:
mind updating the unit tests here?

snapcraft/tests/plugins/test_kernel.py
```diff
@@ -863,15 +863,15 @@ def test_kernel_image_target_as_map(self):
         plugin = kernel.KernelPlugin('test-part', self.options,
                                      project_options)
 
-        self.assertEqual(plugin.make_targets, ['Image', 'modules'])
+        self.assertEqual(plugin.make_targets, ['Image', 'modules', 'dtbs'])
```
kyrofa: Do we have any tests that cover when _not_ building on these architectures, and thus makes sure dtbs aren't built for them?

sergiusens: Yes we do although not directly visible on https://github.com/snapcore/snapcraft/pull/1148/files#diff-6f5410d7a6a3500fab904417fc36123dR779

We should probably refactor the kernel tests into many smaller classes but Paolo shouldn't be the one to do that.

snapcraft/plugins/kernel.py
```diff
@@ -173,12 +173,18 @@ def _set_kernel_targets(self):
...
         if self.dtbs:
             self.make_targets.extend(self.dtbs)
-        self.make_install_targets = [
-            'modules_install', 'INSTALL_MOD_PATH={}'.format(self.installdir)]
+        elif (self.project.kernel_arch == "arm" or
+                self.project.kernel_arch == "arm64"):
+            self.make_targets.append('dtbs')
+            self.make_install_targets.extend([
+                'dtbs_install',
+                'INSTALL_DTBS_PATH={}/dtbs'.format(self.installdir)])
```
elopio: We also run this tests in armhf and arm64. The tests were not updated to take into account that they could be running on these architectures, so now we have new failures:

https://objectstorage.prodstack4-5.canonical.com/v1/AUTH_77e2ada1e7a84929a74ba3b87153c0ac/autopkgtest-xenial-snappy-dev-snapcraft-daily/xenial/armhf/s/snapcraft/20170307_030329_f94e4@/log.gz

This is pretty important, because we can't land in the archives if the tests start failing. @piso77, can you please take a look?

PR #1149, opened 2017-02-17
kernel plugin: if kernel's target == NULL, use per-arch default target

piso77:
Linux's per-arch default targets are:

i386/amd64: bzImage
arm: zImage
arm64: Image.gz

LP: #1665659


sergiusens:
mind updating the unit tests here?

sergiusens:
#1209 replaces this, fixes some issues and adds tests

PR #1150, opened 2017-02-17
kernel plugin: rework MAKEFLAGS from the environment

piso77:
Linux's kernel Makefile gets confused if it is invoked with the
environment already setup by another Linux's Makefile:

linux/package/Makefile -> snapcraft -> linux/Makefile

fix the problem removing the offending make variable (MAKEFLAGS)

LP: #1665661

### DIFF
--- a/snapcraft/plugins/kbuild.py
+++ b/snapcraft/plugins/kbuild.py
@@ -59,6 +59,7 @@ import logging
 import os
 import shutil
 import subprocess
+import re
 
 from snapcraft import BasePlugin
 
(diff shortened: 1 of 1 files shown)


snapcraft/plugins/kbuild.py
```diff
@@ -149,8 +149,14 @@ def do_remake_config(self):
...
     def do_build(self):
+        # Linux's kernel Makefile gets confused if it is invoked with the
+        # environment setup by another Linux's Makefile:
+        # linux/package/Makefile -> snapcraft -> linux/Makefile
+        # fix the problem removing the offending make variable
+        env = os.environ.copy()
+        del env['MAKEFLAGS']
         # build the software
-        self.run(self.make_cmd + self.make_targets)
+        self.run(self.make_cmd + self.make_targets, env=env)
```
sergiusens: can you restore the variable after this call?

PR #1154, opened 2017-02-20
Expose parallel_build_count to scriptlets (LP: #1666271).

oSoMoN:


snapcraft/internal/parts.py
```diff
@@ -317,6 +317,9 @@ def build_env_for_part(self, part, root_part=True):
             env += dep_part.env(stagedir)
             env += self.build_env_for_part(dep_part, root_part=False)
 
+        env.append('PARALLEL_BUILD_COUNT={}'.format(
```
sergiusens: since this method is called more than once (look at `for dep_part in part.deps`) can you put this inside the `if root_part` block?

sergiusens:
hmm, I am having second thoughts about the variable name; can this be renamed to `SNAPCRAFT_PARALLEL_BUILD_COUNT`?

oSoMoN:
Sure thing!

PR #1156, opened 2017-02-21
python plugin: use stage headers if applicable.

sergiusens:
When using staged python, the plugin should also use staged
python headers instead of those on the system.

LP: #1666488

### DIFF
--- a/snapcraft/plugins/python.py
+++ b/snapcraft/plugins/python.py
@@ -193,24 +193,35 @@ class PythonPlugin(snapcraft.BasePlugin):
         else:
             return unstaged_python
 
+    def _get_python_headers(self):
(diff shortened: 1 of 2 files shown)


snapcraft/plugins/python.py
```diff
@@ -193,25 +193,34 @@ def _get_python_command(self):
         else:
             return unstaged_python
 
+    def _get_python_headers(self):
+        base_match = os.path.join('usr', 'include', '{}*'.format(
+            self.options.python_version))
+        unstaged_python = glob(os.path.join(os.path.sep, base_match))
```
sergiusens: but then `os.path.join(self.project.stage_dir, base_match)` won't _join_ below (thanks python :wink:)

PR #1157, opened 2017-02-21
repo: support versioned stage-packages

josepht:
LP: #1666656


snapcraft/internal/repo.py
```diff
@@ -527,3 +530,26 @@ def _try_copy_local(path, target):
...
+    """Break package name into base parts"""
+
+    name = pkg_name
+    version = arch = None
+    if '=' in pkg_name:
+        name, version = pkg_name.split('=')
+        if ':' in name:
+            name, arch = name.split(':')
+
+    if arch:
```
sergiusens: sorry, I don't understand; why do you split to rejoin again?

josepht: Thanks for catching that.  It was a holdover from my mistaken thinking that the format was <pkg>=<version>:<arch>  I'm testing an update now.

josepht:
The test failure looks unrelated to my changes.

snapcraft/internal/repo.py
```diff
@@ -299,6 +299,9 @@ def _get(self, apt_cache, package_names):
                 logger.debug(
                     'Marking {!r} (and its dependencies) to be fetched'.format(
                         name))
+                version = None
+                name_arch, version = _get_pkg_name_parts(name)
+                _set_candidate(apt_cache[name_arch], version)
```
kyrofa: Assuming I understand this correctly, this will silently ignore the case when an invalid or non-existing version is used, and use the most recent version instead. We should probably raise an exception instead so we don't do something unexpected. What do you think? If you agree, please add a test for it as well.

snapcraft/internal/repo.py
```diff
@@ -546,3 +549,21 @@ def _try_copy_local(path, target):
...
         raise MissingCommandError([command])
+
+
+def _get_pkg_name_parts(pkg_name):
+    """Break package name into base parts"""
+
+    name = pkg_name
+    version = None
+    if '=' in pkg_name:
+        name, version = pkg_name.split('=')
```
kyrofa: Small nit: this isn't very pythonic, and checks for '=' twice. Something like this might be better on both fronts:

```python
def _get_pkg_name_parts(pkg_name):
    name = pkg_name
    version = None
    with contextlib.suppress(ValueError):
        name, version = pkg_name.split('=')
    return name, version
```

integration_tests/snaps/asset-tracking/snapcraft.yaml
```diff
@@ -0,0 +1,11 @@
+name: asset-tracking
```
kyrofa: This isn't actually used within any integration tests.

josepht: No, it's supposed to be added in the asset-tracking PR.

snapcraft/internal/repo.py
josepht: Good catch.  Fixed and added an integration check.

snapcraft/internal/repo.py
josepht: updated with your suggestion.

integration_tests/snaps/stage-package-version/snap/snapcraft.yaml
```diff
@@ -0,0 +1,13 @@
+name: stage-package-version
+version: '0.1'
+summary: install a specific version of a stage-package
+description: |
+  Install a specific version of  a stage-package
```
kyrofa: This summary/description implies to me that this is expected to work.

josepht: Yep, fixed.

integration_tests/snaps/stage-package-version/snap/snapcraft.yaml
```diff
@@ -0,0 +1,13 @@
+name: stage-package-version
+version: '0.1'
+summary: install a specific, but non-existent version of a stage-package
+description: |
+  Install a specifi, but non-existent  version of  a stage-package
```
kyrofa: Haha, poor guy, now you have a typo.

kyrofa: Wouldn't it be easier to test this at the unit level?

josepht: I thought that as well but the apt-cache bits are mocked in the unit tests.  Unless I'm missing something.  It seemed like I'd be testing something so faked that it couldn't fail unless I mis-wrote the test.  I'm open to ideas though.

josepht: I fixed the typo btw.

kyrofa: Hmm.... yeah you're right, I can't find a good spot either. Integration it is!

snapcraft/tests/test_repo.py
```diff
@@ -60,6 +60,21 @@ def _fetch_binary(download_dir, **kwargs):
         self.mock_cache.return_value.get_changes.return_value = [
             self.mock_package]
 
+    def test__get_pkg_name_parts_name_only(self):
```
sergiusens: why the double underscore here?

snapcraft/tests/test_repo.py
```diff
@@ -60,6 +60,21 @@ def _fetch_binary(download_dir, **kwargs):
...
+        name, version = repo._get_pkg_name_parts('hello')
+        self.assertEqual('hello', name)
+        self.assertEqual(None, version)
+
+    def test__get_pkg_name_parts_all(self):
```
sergiusens: why the double underscore here?

snapcraft/tests/test_repo.py
```diff
@@ -60,6 +60,21 @@ def _fetch_binary(download_dir, **kwargs):
...
+        name, version = repo._get_pkg_name_parts('hello:i386=2.10-1')
+        self.assertEqual('hello:i386', name)
+        self.assertEqual('2.10-1', version)
+
+    def test__get_pkg_name_parts_no_arch(self):
```
sergiusens: why the double underscore here?

snapcraft/internal/repo.py
```diff
@@ -299,7 +299,10 @@ def _get(self, apt_cache, package_names):
                 logger.debug(
                     'Marking {!r} (and its dependencies) to be fetched'.format(
                         name))
-                apt_cache[name].mark_install()
+                version = None
```
sergiusens: this version assignment here is unnecessary

snapcraft/internal/repo.py
```diff
@@ -299,7 +299,10 @@ def _get(self, apt_cache, package_names):
...
+                name_arch, version = _get_pkg_name_parts(name)
+                _set_candidate(apt_cache[name_arch], version)
```
sergiusens: this would be better read as
```python
if version:
    _set_pkg_version(apt_cache[name_arch], version)
```

And adapt `_set_candidate` accordingly.

snapcraft/tests/test_repo.py
josepht: I've always done that then the method I'm testing starts with an underscore.  More as a mental note than anything.  If you want me to make it a single quote I'm happy to do so.

snapcraft/internal/repo.py
josepht: Since `_set_candidate` also checks version I can remove the check here.  Which helps keep me under the complexity limit :-)

josepht:
@sergiusens I think I've addressed all of your comments.

snapcraft/internal/repo.py
sergiusens: is this due to complexity? doing nothing on set inside set if no version is set seems counterintuitive though

kyrofa: No way... this is a tiny function!

josepht: The complexity is for `_get()` not `_set_candidate()`.  Sergio brings up a valid point I'll see if there's another way to reduce the complexity.

kyrofa: Heh, yeah I thought it was the other `get` for a minute, my apologies. To reverse my statement: this function is actually quite large. Splitting it up probably makes sense.

josepht: Let me know what you think of my update.  It's just a minor refactor to get complexity down not a major overhaul.

snapcraft/internal/repo.py
```diff
@@ -291,18 +291,24 @@ def get(self, package_names):
...
+                if version:
+                    _set_pkg_version(apt_cache[name_arch], version)
+                apt_cache[name_arch].mark_install()
             except KeyError:
                 raise PackageNotFoundError(name)
 
+    def _get(self, apt_cache, package_names):
+        manifest_dep_names = self._manifest_dep_names(apt_cache)
+
+        self._mark_install(apt_cache, package_names)
```
kyrofa: I think this is going a good direction, but we might as well complete the refactor here. This function is made up of three steps:

1. Mark requested packages for installation
2. Unmark base packages for installation
3. Actually fetch archives.

I say we simply put each in its own function. No real code changes.

josepht:
Don't merge this yet.  I've got a bit more refactoring to push.  Waiting on local tests.

kyrofa:
CLA issue for you here, @sergiusens.

sergiusens:
@kyrofa fixed locally and going to ignore it this time... sad but wth

PR #1158, opened 2017-02-22
packaging: snapcraft as a snap

sergiusens:
Integration work to build snapcraft as a snap.

LP: #1666908


sergiusens:
thank @Conan-Kudo I will add a task for that though, it is much more work to get gnupg2 in place.

PR #1160, opened 2017-02-22
cli: rename `history` to 'list-revisions' with `revisions` alias.

josepht:
LP: #1667034


josepht:
Should 'history' be kept as a command, perhaps with a deprecation warning?

snapcraft/main.py
```diff
@@ -118,7 +119,8 @@
                supersedes this command.
   release      Release a revision of a snap to a specific channel.
   status       Show the current status of a snap per channel and architecture.
-  history      List all revisions of a snap.
+  list-revisions List all revisions of a snap.
+  revisions    Alias for list-revisions.
```
kyrofa: Can you please space out the rest of these so it stays pretty?

snapcraft/main.py
```diff
@@ -358,8 +361,8 @@ def _run_store_command(args):  # noqa: C901
     elif args['status']:
         snapcraft.status(
             args['<snap-name>'], args['--series'], args['--arch'])
-    elif args['history']:
-        snapcraft.history(
+    elif args['revisions']:
+        snapcraft.revisions(
```
kyrofa: Does `list-revisions` need to be handled here as well?

integration_tests/test_store_history.py
```diff
@@ -25,53 +25,53 @@
 import integration_tests
 
 
-class HistoryTestCase(integration_tests.StoreTestCase):
+class RevisionsTestCase(integration_tests.StoreTestCase):
```
kyrofa: We should probably include a test case for `list-revisions`.

snapcraft/main.py
josepht: yes

snapcraft/main.py
josepht: yes

integration_tests/test_store_history.py
josepht: I added a single test method since both are handled by the same logic in the source tree.  Is that sufficient or would you prefer an entire copy of the RevisionsTestCase class?

snapcraft/internal/deprecations.py
```diff
@@ -28,6 +28,7 @@
     'dn1': "The 'snap' keyword has been replaced by 'prime'.",
     'dn2': "Custom plugins should now be placed in 'snap/plugins'.",
     'dn3': "Assets in 'setup/gui' should now be placed in 'snap/gui'.",
+    'history': "The 'history' keyword has been replaced by 'revisions'.",
```
kyrofa: This needs to use 'dn4' for the ID (that's mostly historical at this point, but it conforms with the rest). Also, s/keyword/command/ . Finally, was it replaced by `revisions` or `list-revisions` (since one is an alias of the other...)?

snapcraft/main.py
```diff
@@ -121,6 +122,7 @@
   status       Show the current status of a snap per channel and architecture.
   list-revisions List all revisions of a snap.
   revisions    Alias for list-revisions.
+  history      List all revisions of a snap.
```
kyrofa: These descriptions need to be spaced out as well. Also, instead of a description for history, I'd say "Deprecated alias for list-revisions" or something similar.

snapcraft/main.py
```diff
@@ -121,6 +122,7 @@
   status       Show the current status of a snap per channel and architecture.
   list-revisions List all revisions of a snap.
   revisions    Alias for list-revisions.
```
kyrofa: Just to throw this out as a rhetorical comment: I dislike having two commands that do the same thing (`revisions` and `list-revisions`), it seems confusing and I'm not clear on the benefit.

snapcraft/main.py
```diff
@@ -361,7 +368,11 @@ def _run_store_command(args):  # noqa: C901
     elif args['status']:
         snapcraft.status(
             args['<snap-name>'], args['--series'], args['--arch'])
-    elif args['revisions']:
+    elif args['history']:
+        deprecations.handle_deprecation_notice('history')
+        snapcraft.revisions(
+            args['<snap-name>'], args['--series'], args['--arch'])
```
kyrofa: Rather than calling `snapcraft.revisions` twice, I suggest something a little less performant but less error-prone:

```python
elif args['revisions'] or args['list-revisions'] or args['history']:
    if args['history']:
        deprecations.handle_deprecation_notice('dn4')
    snapcraft.revisions([...])
```

integration_tests/test_store_history.py
kyrofa: That seems fine, although I don't see that here. Have you not pushed it? Make sure you add a test case for `history` as well, verifying that it's deprecated etc.

kyrofa: That seems fine, although I don't see that here. Have you not pushed it? Make sure you add a test case for history as well, verifying that it's deprecated etc.

josepht:
This should be ready for a second look.  I've addressed all the comments.

josepht:
the PR for snappy-docs is here: https://github.com/CanonicalLtd/snappy-docs/pull/44

snapcraft/internal/deprecations.py
```diff
@@ -28,6 +28,7 @@
     'dn1': "The 'snap' keyword has been replaced by 'prime'.",
     'dn2': "Custom plugins should now be placed in 'snap/plugins'.",
     'dn3': "Assets in 'setup/gui' should now be placed in 'snap/gui'.",
+    'dn4': "The 'history' command has been replaced by 'revisions'.",
```
sergiusens: shouldn't this be `list-revisions`?

elopio:
snapcraft/tests/commands/test_list_revisions.py:56:1: E302 expected 2 blank lines, found 1

josepht:
The integration test failure is a connection timeout.

PR #1163, opened 2017-02-27
docs: update the directory where the API pages are generated

elopio:
LP: [#1668413](https://bugs.launchpad.net/snapcraft/+bug/1668413)


tools/push_api_gh_pages.sh
```diff
@@ -6,7 +6,7 @@ GH_REPO="@github.com/snapcore/snapcraft.git"
 
 FULL_REPO="https://$GH_TOKEN$GH_REPO"
 
-cd docs/api/_build/html
+cd tools/api/_build/html
```
kyrofa: I'm not quite sure what this is, but the `_build/` dir doesn't exist in my branch. Are you creating it somewhere else?

elopio: Yes, it's created by the Makefile in tools/api.

PR #1166, opened 2017-03-01
tests: Fix name registration window limit test to latest changes

fgallina:
Now it's possible to register N names (currently 10) during the course
of a time window (currently 10 minutes). The integration test checking
for the limit now reflects that.

### DIFF
--- a/integration_tests/test_store_register.py
+++ b/integration_tests/test_store_register.py
@@ -92,14 +92,17 @@ class RegisterTestCase(integration_tests.StoreTestCase):
         # This test has a potential to fail if working off a slow
         # network.
         self.login()
-        snap_name_1 = 'good-snap{}'.format(uuid.uuid4().int)
-        snap_name_2 = 'test-too-fast{}'.format(uuid.uuid4().int)
 
-        self.register(snap_name_1, wait=False)
+        error = None
+        for idx in range(self.test_store.register_count_limit + 1):
+            snap_name = 'test-too-fast{}-{}'.format(uuid.uuid4().int, idx)
+            try:
+                self.register(snap_name, wait=False)
+            except subprocess.CalledProcessError as exc:
+                error = exc
+                break
 
-        error = self.assertRaises(
-            subprocess.CalledProcessError,
-            self.register, snap_name_2, wait=False)
+        self.assertIsNotNone(error, 'An error must be raised.')
         expected = (
             '.*You must wait \d+ seconds before trying to register your '
             'next snap.*')
--- a/snapcraft/tests/fixture_setup.py
+++ b/snapcraft/tests/fixture_setup.py
@@ -328,16 +328,19 @@ class TestStore(fixtures.Fixture):
         test_store = os.getenv('TEST_STORE', 'fake')
         if test_store == 'fake':
             self.useFixture(FakeStore())
+            self.register_count_limit = 10
             self.register_delay = 0
             self.reserved_snap_name = 'test-reserved-snap-name'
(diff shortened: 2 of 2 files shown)


fgallina:
This does fix the test as intended but now there's one major drawback which is that now every name registration that's successful makes the suite to wait for 10 minutes. To fix this, we could:

- Make `StoreTestCase.register` always attempt registration, if failed and `wait=True` determine whether the call failed because of time limit and if so retry, otherwise just raise `subprocess.CalledProcessError` as it currently does. This will allow to take advantage of all allowed registrations during the time window efficiently.

- Additionally, we also could set a secondary acceptance test user and use that one to test the name registration limit. With that, we'd ensure such test won't cause blocking on other ones exercising name registration.

elopio:
Thanks @fgallina! Also instead of attempt the registration, we could keep a counter of the number of registration attempts, and only wait when it's 10.

Your code will work nicely on staging, but the fake servers need to be updated for the new behaviour.

integration_tests/test_store_register.py
```diff
@@ -92,14 +92,17 @@ def test_registrations_in_a_row_fail_if_too_fast(self):
...
-        snap_name_1 = 'good-snap{}'.format(uuid.uuid4().int)
-        snap_name_2 = 'test-too-fast{}'.format(uuid.uuid4().int)
 
-        self.register(snap_name_1, wait=False)
+        error = None
+        for idx in range(self.test_store.register_count_limit + 1):
+            snap_name = 'test-too-fast{}-{}'.format(uuid.uuid4().int, idx)
+            try:
+                self.register(snap_name, wait=False)
+            except subprocess.CalledProcessError as exc:
```
elopio: This doesn't really catch that the 11th registration causes a failure.
Maybe, put self.register in a loop 10 times. Then use assertRaises to catch the 11th failure?

sergiusens:
any movement here?

integration_tests/test_store_register.py
fgallina: The intent here is to trigger the registration limit at any of the attempts. Whether this is going to be on the 1st or the 11th really depends on how tests got ordered and how many names the test user registered until that point.

As an extra related note, fake server tests against this trigger on first attempt as the 'test-too-fast' prefix is actually the trigger for the window error.

fgallina:
I think @elopio suggestion may be a nice idea but then external things could be triggering name registration (even aborted runs) for the test user and so the counter may be an unreliable source, possibly causing failures until the time window is clean again.

For reference, on our [staging acceptance tests](http://bazaar.launchpad.net/~ubuntuone-pqm-team/ols-store-tests/store-acceptance-tests/view/head:/tests/api/snap/test_register_name.py#L18) we take a different approach on the name registration tests: we attempt to register whenever needed and if a `429` HTTP code is returned then we wait (if kwarg `wait=True`) for the amount of time received in the `Retry-After` header and retry. This approach has been reliable regardless of the current state of the test user registrations in the store and may be a fine approach to follow here.

Unfortunately at this time I have some priority work to focus on before I can come back to to this. So it'd be great if you can pick it up from here, otherwise I'll come back as soon as I'm able to.

elopio:
@fgallina I'm finishing this branch. Your suggestion with retry on wait seems pretty simple.
But, it sounds like you forgot to push your changes on the fake servers, do you still have them around?

elopio:
@fgallina forget I said anything. Just now I understood that no changes are needed on the fake servers. Sorry for being slow, twice.

elopio:
Closing in favor of #1224.

PR #1167, opened 2017-03-02
tests:install wget in the container that triggers the beta tests

elopio:
LP: [#1669247](https://bugs.launchpad.net/snapcraft/+bug/1669247)


elopio:
This makes no sense. It was working before and we didn't change anything related to the container packages. But oh well... I tried the container with wget and it successfully triggered the autopkgtests.

PR #1168, opened 2017-03-02
sources: add support for 7-zip files

tim-sueberkrueb:
This allows snapcraft to work with sources in form of `.7z` files.
Unpacking `.7z` is handled using `libarchive` which is already a
dependency of the project.
Tests: As [discussed on rocket chat](https://rocket.ubuntu.com/channel/snapcraft), since there is no support for [writing `7z` files with libarchive yet](https://github.com/libarchive/libarchive/wiki/LibarchiveFormats), I included a simple `test.7z` file only including a file `test.txt` with the text `Text` in it.


sergiusens:
Can you please check the contribution guidelines

```
Contributor: https://api.launchpad.net/1.0/~tim-sueberkrueb
The contributor has a Launchpad account, but has not signed the CLA.
```

tim-sueberkrueb:
Thanks for responding to my pull request.
Squashed my previous commits. Formatted commit message as described. Signed the CLA for this contribution.

snapcraft/internal/sources/_7z.py
```diff
@@ -0,0 +1,62 @@
...
+import os
+import libarchive
+import shutil
+import tempfile
+
+from . import errors
+from ._base import FileBase
+
+
+class _7z(FileBase):
```
elopio: no need for this underscore here.

snapcraft/internal/sources/_7z.py
```diff
@@ -0,0 +1,62 @@
...
+            shutil.rmtree(dst)
+            os.makedirs(dst)
+            shutil.move(tmp_7z, _7z_file)
+
+        # Ensure dst does not have trailing slash
+        dst = dst.rstrip('/')
+        # Open the 7z file and extract it to destination
+        with libarchive.file_reader(_7z_file) as archive:
+            for file_entry in archive:
+                entrypath = file_entry.pathname.lstrip('./')
```
elopio: you don't have a test for this lstrip, do you?

snapcraft/internal/sources/_7z.py
tim-sueberkrueb: @elopio thanks for you response! 
I don't think it's possible to start class names in Python with numbers, is it?

elopio: you are right.

kyrofa: (you could spell it out if you prefer)

snapcraft/internal/sources/_7z.py
tim-sueberkrueb: @elopio thanks! Actually this lstrip isn't required for 7z files as far as I know (I based my work on the rpm module and this is a rpm specific feature). I removed it.

tim-sueberkrueb:
@elopio added the integration test.

snapcraft/internal/sources/_7z.py
tim-sueberkrueb: @kyrofa @elopio what do you prefer?

kyrofa: No preference on my part, leave as-is unless @elopio prefers otherwise.

elopio: I would prefer python to support numbers. This is fine as a workaround, it will just be a little confusing with our understanding of _ as internal.

kyrofa: > it will just be a little confusing with our understanding of _ as internal.

Yeah... thinking on that a little more, I change my vote from neutral to "please spell it out."

snapcraft/internal/sources/_7z.py
```diff
@@ -0,0 +1,61 @@
...
+
+        if clean_target:
+            tmp_7z = tempfile.NamedTemporaryFile().name
+            shutil.move(_7z_file, tmp_7z)
+            shutil.rmtree(dst)
+            os.makedirs(dst)
+            shutil.move(tmp_7z, _7z_file)
+
+        # Ensure dst does not have trailing slash
+        dst = dst.rstrip('/')
```
elopio: what happens when dst ends with a trailing slash? There is no test for this condition.

snapcraft/tests/sources/test_7z.py
```diff
@@ -0,0 +1,62 @@
...
+            '7z', 'test.7z'
+        )
+
+        dest_dir = 'src'
+        os.makedirs(dest_dir)
+
+        _7z_source = sources._7z(test_7z_file_path, dest_dir)
+        _7z_source.pull()
+
+        self.assertEqual(os.listdir(dest_dir), ['test.txt'])
```
elopio: If the 7zip archive has only one file, this test is not really exercising the for loop. I would add two more files.

snapcraft/tests/sources/test_7z.py
```diff
@@ -0,0 +1,62 @@
...
+class Test7z(tests.TestCase):
+
+    @staticmethod
+    def relative_path():
+        return os.path.dirname(os.path.realpath(__file__))
+
+    def test_pull_7z_file_must_extract(self):
+        test_7z_file_path = os.path.join(
+            Test7z.relative_path(),
+            '7z', 'test.7z'
```
elopio: Ideally, the unit test will build the test.7z archive instead of using a binary. 
What are our options here?

snapcraft/internal/sources/_7z.py
tim-sueberkrueb: @elopio ups, since we're using `os.path.join`, this actually shouldn't be required since `os.path.join` handles stuff like this. `dst` is not used anywhere else, so this line can be removed.

snapcraft/tests/sources/test_7z.py
tim-sueberkrueb: @elopio since `libarchive` doesn't support writing 7z files right now, to my understanding we could ...
* find another package which provides 7z writing (couldn't find one after a quick search)
* invoke a shell command like done in this wrapper: https://github.com/Argomirr/py7z/blob/master/py7z.py#L111

snapcraft/internal/sources/_7z.py
tim-sueberkrueb: @kyrofa okay, will do. Btw, congrats for 300 GitHub-🌟 :)

snapcraft/internal/sources/_7z.py
tim-sueberkrueb: Done.

snapcraft/internal/sources/_7z.py
tim-sueberkrueb: Done

elopio:
@tim-sueberkrueb I'm sorry we left your branch here in the queue for so long. With the available options, I think that adding the 7zip file to the repo is ok. But now this has a conflict, can you please resolve it?

+1 on everything else from me. Just waiting for the tests to be green.

snapcraft/tests/sources/test_7z.py
tim-sueberkrueb: Will do.

tim-sueberkrueb: Done.

tim-sueberkrueb:
@elopio thanks for the review :) Looks like the tests are finally [green](https://www.youtube.com/watch?v=sRZh42fWLbg) :D

PR #1169, opened 2017-03-02
tests: update the ftp source for integration test

elopio:
kernel.org has just shut down the ftp server, so we need to use a different one.
To avoid this happening again, the test starts a local ftp server.

This adds python3-pyftpdlib to the autopkgtest dependencies.

LP: [#1669541](https://bugs.launchpad.net/snapcraft/+bug/1669541)

### DIFF
--- a/integration_tests/test_dump_plugin.py
+++ b/integration_tests/test_dump_plugin.py
@@ -16,13 +16,48 @@
 
 import os
 import fixtures
+import tarfile
+import threading
+
+from pyftpdlib import (
+    authorizers,
+    handlers,
+    servers
+)
 
 from testtools.matchers import (
     DirExists,
+    FileContains,
     FileExists
 )
 
 import integration_tests
+
+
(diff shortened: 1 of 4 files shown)


integration_tests/test_dump_plugin.py
```diff
@@ -16,15 +16,50 @@
...
 from testtools.matchers import (
     DirExists,
+    FileContains,
     FileExists
 )
 
 import integration_tests
 
 
+class FtpServerRunning(fixtures.Fixture):
```
kyrofa: Why not just call this `FtpServer`?

kyrofa:
This is lovely @elopio, thanks! I only have one comment on the FTP server name.

integration_tests/test_dump_plugin.py
kyrofa: After a quick chat on Rocket, this is fine.

PR #1170, opened 2017-03-02
core: fix symlink resolution in get_core_dynamic_linker

mwhudson:
The existing code did not handle relative symlinks correctly, they were
expanded relative to /snap/core/current, not the directory they resided in.
This is why classic snap builds only worked on amd64 and ppc64el on Launchpad.

LP: [#1665165](https://bugs.launchpad.net/snapcraft/+bug/1665165)

No test because I don't see how to add one (afaict, get_core_dynamic_linker is currently untested).

### DIFF
--- a/snapcraft/_options.py
+++ b/snapcraft/_options.py
@@ -21,6 +21,7 @@ import platform
 
 from snapcraft.internal import common
 from snapcraft.internal.deprecations import handle_deprecation_notice
+from snapcraft.internal.errors import SnapcraftEnvironmentError
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


mwhudson:
(I have tested this in an i386 lxd and it worked there)

kyrofa:
I'm going to go ahead and hit the "update" button here which will hopefully fix that red X from Travis.

snapcraft/_options.py
```diff
@@ -193,18 +193,31 @@ def get_core_dynamic_linker(self):
...
+        dynamic_linker_path = os.path.join(
+            core_path,
+            self.__machine_info.get('core-dynamic-linker',
+                                    'lib/ld-linux.so.2'))
+
+        # We can't use os.path.realpath because any absolute symlinks
+        # have to be interpreted relative to core_path, not the real
+        # root.
+        seen_paths = set()
+        while True:
```
kyrofa: Would using os.path.realpath simplify this a bit?

kyrofa: Eh... I suppose not. Since those symlinks may be absolute, we need to make sure we look inside the core snap for that, not on the host.

snapcraft/_options.py
```diff
@@ -193,18 +193,31 @@ def get_core_dynamic_linker(self):
...
+            if dynamic_linker_path in seen_paths:
+                raise EnvironmentError(
```
kyrofa: This won't do anything about the errno in `EnvironmentError`. Can we please use `snapcraft.internal.errors.SnapcraftEnvironmentError` instead?

snapcraft/_options.py
```diff
@@ -193,18 +193,31 @@ def get_core_dynamic_linker(self):
...
+                    "found symlink loop resolving dynamic linker path")
+            seen_paths.add(dynamic_linker_path)
+            if not os.path.exists(dynamic_linker_path):
+                return None
+            if not os.path.islink(dynamic_linker_path):
+                return dynamic_linker_path
+            link_contents = os.readlink(dynamic_linker_path)
+            if link_contents.startswith('/'):
```
kyrofa: Would `os.path.isabs()` be better?

snapcraft/_options.py
mwhudson: Sure.

snapcraft/_options.py
mwhudson: /me tosses a coin, let's say yes.

sergiusens:
Thanks for this @mwhudson! You are correct that there is no unit test for this, we do however have integration tests and other architectures are run through adt, unfortunately we are not allowed to run subsets of tests of things we know work on other architectures and are restricted to always fail until everything works (it is the ubuntu-dev way) which hides the test results from us. We can however setup some test fixtures on the side to at least run these subsets during development and detach ourselves from the adt results from what would be a deb's tests. Do you mind looking into that @elopio?

mwhudson:
Given that your autopkgtests currently succeed on armhf, which has this problem, I suspect they are not sufficient either (none of the tests of classic snaps actually install and run the built snap, afaics?)

sergiusens:
the `git` demo is also a classic snap and has the potential to run (it it isn't setup already).

kyrofa:
@sergiusens does @elopio need to look into that before this lands?

sergiusens:
ok this does solve the problem too https://code.launchpad.net/~sergiusens/+snap/test-snap (at least for i386)

PR #1174, opened 2017-03-07
copy, dump plugin: ignore symlinks to libc.

kyrofa:
libc is not typically staged, as it's always contained in the core snap. However, Snapcraft leaves symlinks to its libraries from stage-packages alone and intact. This causes issues when it comes to content sharing and building against tarred staging areas, as various plugins work really hard to ensure the snap doesn't include dangling symlinks, and they end up following the libc symlinks in the tarred staging area and
pulling in libc. The naive solution to this is to simply remove such symlinks from the stage-packages, and this was indeed the solution introduced in commit 3052b0f459ded440dad2e2362946ee37e09c8606. However, didn't work quite as well as desired and led to odd bugs such as ld linking to static libraries in the part instead of shared libraries on the host, etc.

This PR is a second take on fixing LP: [#1658774](https://bugs.launchpad.net/snapcraft/+bug/1658774): instead of removing symlinks to libc, modify the copy and dump plugins to ignore (i.e. not follow) symlinks to libc.


kyrofa:
I wanted to propose this for discussion. I have a few questions:

1. Currently the libc library list is implemented in `snapcraft.internal.libraries`, but it's used within plugins. I'd rather not use an internal API within a public plugin, so we need a better place to put this. Should I make a `library_utils` module?
2. What happens if one is attempting to cross-build a snap? This will use the libc library list from the host, not the target. That's a problem today as well, and that's not really something we support anyway... but I wanted to mention it.

snapcraft/internal/libraries.py
```diff
@@ -86,7 +87,7 @@ def get_dependencies(elf):
 
     This may include libraries contained within the project.
     """
-    logger.debug('Getting dependencies for {!r}'.format(elf))
+    logger.debug('Getting dependencies for {!r}'.format(str(elf)))
```
sergiusens: there's a specific bug for this iirc

snapcraft/internal/libraries.py
```diff
@@ -102,3 +103,18 @@ def get_dependencies(elf):
     libs = [l for l in ldd_out if not os.path.basename(l) in system_libs]
 
     return libs
+
+
+_libc_lib_list = None
```
sergiusens: thanks for moving this out of `repo`

sergiusens: on second thought, this should be in repo since it call `dpkg` and will break the crossdistro story (or cause cyclic imports which is also an unhappy story).

elopio:
This is the execution of the external suite with this branch:
https://travis-ci.org/elopio/snapcraft-de-noche/builds/209576692

Only two failures: docker, which has a pr open to fix it. And quassel-webserver, that I still have to investigate but it's not related to this branch. So all good from this side.

snapcraft/internal/libraries.py
kyrofa: Hmm, I'm not finding one. I can make one and extract this fix if you like.

snapcraft/internal/libraries.py
kyrofa: Well that's an easy answer to question (1)!

kyrofa:
> So all good from this side.

kyrofa:
> So all good from this side.

Thanks for that, @elopio. That's a good safety net.

kyrofa:
Alright, this is ready for another look.

snapcraft/internal/repo.py
```diff
@@ -576,3 +568,18 @@ def _set_pkg_version(pkg, version):
         pkg.candidate = version
     else:
         raise PackageNotFoundError('{}={}'.format(pkg.name, version))
+
+
+_libc_lib_list = None
+
+
+def libc_library_list():
```
sergiusens: Maybe name this 
```python
def get_pkg_libs(pkg):
```

kyrofa: Yeah, if you see this being more generally useful, sure. Still want me to save the result, just make it a dict instead?

sergiusens: hmm, global dicts; make sure it is non racy and works with copies at all times (more so for whatever it returns.

kyrofa: Alright, done. I also added docs to that now-public function.

kyrofa: What race are you concerned about? I didn't add a mutex in there, for example.

PR #1179, opened 2017-03-09
godeps plugin: add git to build-packages

cjwatson:
The godeps plugin runs "go get github.com/rogpeppe/godeps", which
requires git to be installed.

LP: #1671546


sergiusens:
oh, thanks for this!

PR #1180, opened 2017-03-09
tests: support snap directory in external tests

elopio:
LP: [#1671620](https://bugs.launchpad.net/snapcraft/+bug/1671620)

### DIFF
--- a/external_snaps_tests/__main__.py
+++ b/external_snaps_tests/__main__.py
@@ -102,9 +102,8 @@ def _build_snaps(path, cleanbuild=False, keep_dir=False):
                     or 'snapcraft.yaml' in filenames
                     or '.snapcraft.yaml' in filenames):
                 _build_snap(dirpath, cleanbuild, keep_dir)
-                if 'snap' in dirnames:
-                    # Do not recurse in any directory.
-                    dirnames = []
+                # Do not recurse in any directory.
+                dirnames = []
     except subprocess.CalledProcessError as e:
         sys.exit(e.returncode)
     finally:
--- a/external_snaps_tests/__main__.py
+++ b/external_snaps_tests/__main__.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright (C) 2016 Canonical Ltd
+# Copyright (C) 2016-2017 Canonical Ltd
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License version 3 as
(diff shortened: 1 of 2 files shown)


external_snaps_tests/__main__.py
```diff
@@ -97,9 +97,13 @@ def _bzr_branch(url):
...
     try:
-        for dirpath, _, filenames in os.walk(path):
-            if 'snapcraft.yaml' in filenames or '.snapcraft.yaml' in filenames:
+        for dirpath, dirnames, filenames in os.walk(path, topdown=True):
+            if ('snap' in dirnames
+                    or 'snapcraft.yaml' in filenames
+                    or '.snapcraft.yaml' in filenames):
                 _build_snap(dirpath, cleanbuild, keep_dir)
+                # Do not recurse in any directory.
+                dirnames = []
```
kyrofa: I've never seen this before. It prevents the `walk` from going into those directories?

elopio: Just if you use the top down argument.

elopio:
I'm testing this with a run of the nightlies, so wait a moment before landing.

external_snaps_tests/__main__.py
kyrofa: @elopio but you're not actually modifying the list in-place, right?

> When topdown is True, the caller can modify the dirnames list in-place (perhaps using del or slice assignment), and walk() will only recurse into the subdirectories whose names remain in dirnames[...]

elopio: I am. I think. I don't really understand your question.
The docs say that recursion will happen only on the directories that remain in dirnames. If dirnames is empty, no recursion will happen.

elopio:
Hold this. There's a directory called snap in etcd, that's not related to us. This breaks the docker test.

external_snaps_tests/__main__.py
kyrofa: > I am. I think. I don't really understand your question.

Let me explain with this:

```python
>>> real = [1, 2, 3]
>>> ref = real
>>> real
[1, 2, 3]
>>> ref
[1, 2, 3]
>>> ref = []  # you're doing this
>>> real
[1, 2, 3]
>>> ref
[]
>>> ref = real
>>> del ref[:]
>>> real
[]
>>> ref
[]
```

elopio: You are right, thanks Kyle.
And with this, the change for etcd wouldn't have been necessary. But it doesn't hurt either.

elopio:
Ready!

PR #1182, opened 2017-03-10
meta: do not quote the command wrapper args.

sergiusens:
The arguments for the command wrapper should not be quoted, that is
`$@` should not be surrounded by quotes.

LP: #1671674


mwhudson:
I really doubt this is correct, btw.

sergiusens:
@mwhudson yeah, I am thinking this is more of something to be done in https://github.com/lxc/lxd-pkg-ubuntu/pull/27 but my scripting in this area is horrible

PR #1184, opened 2017-03-10
store: enable retries for store calls.

sergiusens:
Calling anything related to the store, more so when dealing with CDNs
has become extremely flaky. This registers an adapter to enabled a
maximum retry count of 5 before giving up on connection resets.

LP: #1671817

### DIFF
--- a/snapcraft/storeapi/__init__.py
+++ b/snapcraft/storeapi/__init__.py
@@ -1,7 +1,7 @@
 
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright (C) 2016 Canonical Ltd
+# Copyright (C) 2016-2017 Canonical Ltd
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License version 3 as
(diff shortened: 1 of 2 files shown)


snapcraft/storeapi/__init__.py
```diff
@@ -91,6 +92,10 @@ def __init__(self, conf, root_url):
         self.conf = conf
         self.root_url = root_url
         self.session = requests.Session()
+        # Setup max retries for all store URLs and the CDN
+        self.session.mount('http://', HTTPAdapter(max_retries=5))
+        self.session.mount('https://', HTTPAdapter(max_retries=5))
```
kyrofa: This looks magical! Although I want to point out: I often see multiple PRs fail at the same time, which seems like the CDN goes down completely for a period of time. Any idea how quickly these retries occur?

sergiusens: from the error in https://travis-ci.org/snapcore/snapcraft/jobs/209838628 and given --debug I know it works somehow.... we can make it wait a few seconds, that would be a tad more complex, want me to do that?

kyrofa: This is simple enough, we might as well see if it fixes the issue. If it doesn't we can make it wait a little longer.

PR #1185, opened 2017-03-10
repo: add version support for build-packages

josepht:
This is really just adjusting the use of apt_cache to not include
the version in the package name.

LP: #1671939


snapcraft/internal/repo.py
```diff
@@ -83,7 +83,8 @@ def install_build_packages(packages):
     with apt.Cache() as apt_cache:
         for pkg in unique_packages:
             try:
-                if not apt_cache[pkg].installed:
+                pkg_name, _ = _get_pkg_name_parts(pkg)
+                if not apt_cache[pkg_name].installed:
```
kyrofa: What happens if there are multiple versions available, and one is installed, but the YAML specifies another?

sergiusens: with this logic it will do nothing, but if we get the version here we can tell apt to install our preferred version

josepht: I'm working on an update to check if the installed version matches the requested version and install the requested version if they don't match.  I'm assuming apt will do the right thing here.

integration_tests/test_build_package_version.py
```diff
@@ -0,0 +1,34 @@
...
+
+import subprocess
+
+import integration_tests
+
+
+class BuildPackageVersionTestCase(integration_tests.TestCase):
+
+    def test_build_package_gets_version(self):
+        self.run_snapcraft('pull', 'build-package-version')
```
elopio: This is good. It would be better if we could check that the version pulled is the right one. Do we have a way to do it?

josepht: I can check the state file contains the proper versioned package name.  I guess you'd also want to verify that the actual package installed is the expected version, right?

snapcraft/internal/repo.py
josepht: I've pushed the update to handle installing a versioned package when there is another version installed.

integration_tests/test_build_package_version.py
elopio: Ideally, yes. dpkg something something to get the version, but I don't really know how to do that. If it's easy and readable, please add it.

josepht: I assume using apt_cache to check the installed version is sufficient.

josepht:
The integration test failure seems unrelated but I've seen it several times now.

josepht:
The pkg version check changes got lost during the merge of the repo rebase.  Re-added in 8eb2eb6

integration_tests/test_build_package_version.py
kyrofa: Yeah that looks good.

snapcraft/tests/repo/test_deb.py
```diff
@@ -241,12 +244,29 @@ def install_test_packages(self, test_pkgs, mock_apt, mock_env):
...
+
+    @patch('snapcraft.repo._deb.is_dumb_terminal')
+    @patch('subprocess.check_call')
+    def test_install_versioned_build_package(
             self, mock_check_call, mock_is_dumb_terminal):
         mock_is_dumb_terminal.return_value = False
         self.install_test_packages(self.test_packages)
 
         installable = self.get_installable_packages(self.test_packages)
+        installable.append('versioned-package=0.2')
```
kyrofa: Since this package isn't installed, shouldn't it already be in the installable list?

josepht: Yes, updated.

snapcraft/tests/repo/test_deb.py
```diff
@@ -241,7 +244,23 @@ def install_test_packages(self, test_pkgs, mock_apt, mock_env):
 
     @patch('snapcraft.repo._deb.is_dumb_terminal')
     @patch('subprocess.check_call')
-    def test_install_buid_package(
+    def test_install_build_package(
```
kyrofa: Err, on second look, isn't this now the exact same test as the one below?

integration_tests/snaps/build-package-version/snap/snapcraft.yaml
```diff
@@ -0,0 +1,14 @@
+name: build-package-version
```
kyrofa: Mind adding a global build package with a version as well?

integration_tests/snaps/build-package-version/snap/snapcraft.yaml
```diff
@@ -7,6 +7,7 @@ description: |
 
 grade: stable
 confinement: strict
+build-packages: ['make=4.1-6']
```
kyrofa: (and make sure it's installed in the integration test)

integration_tests/snaps/build-package-version/snap/snapcraft.yaml
josepht: done

integration_tests/snaps/build-package-version/snap/snapcraft.yaml
josepht: duh, done. :)

kyrofa:
Cannot seem to get the core snap downloaded on Travis, but the autopkgtests work.

PR #1187, opened 2017-03-13
tests: take into account the new current link.

elopio:
snapd in zesty now creates a symlink from `~/snap/<name>/current`
to `~/snap/<name>/<revision>`. We need to take this into account
when searching for files in the snaps directories.

LP: #1672463


elopio:
This is good to go. Only failures were caused by https://bugs.launchpad.net/snapd/+bug/1672495

PR #1188, opened 2017-03-13
store: set User-Agent header in store requests.

josepht:
LP: #1663383


snapcraft/storeapi/__init__.py
```diff
@@ -97,7 +97,8 @@ def __init__(self, conf, root_url):
         self.session.mount('https://', HTTPAdapter(max_retries=5))
 
         self._snapcraft_headers = {
-            'X-SNAPCRAFT-VERSION': snapcraft.__version__
+            'X-SNAPCRAFT-VERSION': snapcraft.__version__,
+            'User-Agent': 'snapcraft/{}'.format(snapcraft.__version__),
```
josepht: Will do

cprov: Snapd does `snapd/123 (series 16; classic; testing) ubuntu/16.10 (amd64) linux/4.8.0-3-generic`
IFAICS, snapcraft does not depend on kernels (am I wrong ?), so we could start with snapcraft and host entities: `snapcraft/123 (classic; testing) ubuntu/16.10 (amd64)`

josepht: what is 'testing'?  Do I need to check if snapcraft is installed via a deb package, a snap?

josepht:
Here's an example output:

```
snapcraft/2.27.1 Ubuntu/16.04 (x86_64)
```

snapcraft/tests/store/test_store_client.py
```diff
@@ -1183,3 +1185,16 @@ def test_sign_dev_agreement_exception_store_down(self):
...
+
+    def test_user_agent(self):
+        arch = ProjectOptions().deb_arch
+        expected = 'snapcraft/{} {} ({})'.format(
+            snapcraft_version,
+            '/'.join(platform.dist()[0:2]),  # i.e. Ubuntu/16.04
+            arch,
+        )
+        actual = storeapi._get_user_agent()
+        self.assertEqual(expected, actual)
```
elopio: ok, this is not very useful as it will be really difficult to make it fail. Using fakes instead of the real calls to deb_arch, version and dist would make it more readable, but using fakes is the downside.
Oh well, at least it will catch typos or when somebody tries to change the function.

cprov:
@josepht  thanks. It will be sufficient to start collecting data on the store side. It think we will also want snapcraft qualifiers (terms between parenthesis) like `(<deb|snap>; [testing])` but we can add them when we have a concrete case.

sergiusens:
@cprov are you using the `X-SNAPCRAFT-VERSION` entry at all, if not, let's remove it in this PR as well please.

cprov:
@sergiusens good point,  `X-SNAPCRAFT-VERSION` is not used in any metrics.

sergiusens:
@cprov @josepht this might be a stretch but maybe it would be nice to capture if a CI environment is being used (travis or something from gitlab).

snapcraft/storeapi/__init__.py
```diff
@@ -81,10 +81,23 @@ def _deserialize_macaroon(value):
         raise errors.InvalidCredentialsError('Failed to deserialize macaroon')
 
 
+def _is_ci_env():
```
josepht: Done

snapcraft/storeapi/_agent.py
```diff
@@ -0,0 +1,42 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+import platform
+
+import snapcraft
+
+
+def _is_ci_env():
+    env_prefixes = ['TRAVIS']
```
elopio: When autopkgtests run they set $AUTOPKGTEST_TMP. You can use that one too.

snapcraft/tests/store/test_agent.py
```diff
@@ -0,0 +1,49 @@
...
+        expected_post = ' {} ({})'.format(
+            '/'.join(platform.dist()[0:2]),  # i.e. Ubuntu/16.04
+            arch,
+        )
+        actual = storeapi._agent._get_user_agent()
+        self.assertTrue(actual.startswith(expected_pre))
+        self.assertTrue(actual.endswith(expected_post))
+
+    def test_in_ci_env(self):
+        os.environ['TRAVIS_STUFF'] = 'stuff'
```
elopio: You should use the EnvironmentVariable fixture:
https://github.com/testing-cabal/fixtures/blob/master/fixtures/_fixtures/environ.py

snapcraft/tests/store/test_agent.py
```diff
@@ -0,0 +1,49 @@
...
+
+        self.assertTrue(storeapi._agent._is_ci_env())
+
+    def test_not_in_ci_env(self):
+        os.environ = {}
```
elopio: instead of reseting the whole environment, use EnvironmentVariable to set the vars you care about to None.

snapcraft/storeapi/_agent.py
sergiusens: good call (even though we never run against production).

snapcraft/tests/store/test_agent.py
```diff
@@ -0,0 +1,66 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2016 Canonical Ltd
```
elopio: 2017, please

snapcraft/storeapi/_agent.py
```diff
@@ -0,0 +1,42 @@
...
+    matches = []
+
+    for prefix in env_prefixes:
+        matches += [
+            var for var in os.environ.keys() if var.startswith(prefix)]
+
+    return len(matches) > 0
+
+
+def _get_user_agent():
```
elopio: I would remove the _ from this function, to kind of indicate that it's used by a different python file.
Not a big deal though, as we haven't defined clear rules of _.

josepht:
Can someone re-kick the Travis tests please?

kyrofa:
> Can someone re-kick the Travis tests please?

Done.

PR #1189, opened 2017-03-13
core: resolve ld link first.

sergiusens:
When building with classic confinement the logic was to check
if what was defined as ld existed, but it wouldn't if it was
a symbolic link so check for that in the first check as well.

LP: #1672496

### DIFF
--- a/snapcraft/_options.py
+++ b/snapcraft/_options.py
@@ -209,8 +209,7 @@ class ProjectOptions:
                     "found symlink loop resolving dynamic linker path")
 
             seen_paths.add(dynamic_linker_path)
-            if not (os.path.exists(dynamic_linker_path) or
-                    os.path.islink(dynamic_linker_path)):
+            if not os.path.lexists(dynamic_linker_path):
                 return None
             if not os.path.islink(dynamic_linker_path):
                 return dynamic_linker_path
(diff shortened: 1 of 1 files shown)


snapcraft/_options.py
```diff
@@ -207,11 +207,14 @@ def get_core_dynamic_linker(self):
             if dynamic_linker_path in seen_paths:
                 raise SnapcraftEnvironmentError(
                     "found symlink loop resolving dynamic linker path")
+
             seen_paths.add(dynamic_linker_path)
-            if not os.path.exists(dynamic_linker_path):
+            if not (os.path.exists(dynamic_linker_path) or
```
mwhudson: I don't think this is right, the correct fix is to replace os.path.exists with os.path.lexists. Sorry about this!

sergiusens: ah, that is what I was looking for, thanks @mwhudson

sergiusens:
The errors in here are all related to non related issues like:

```
requests.packages.urllib3.exceptions.NewConnectionError: <requests.packages.urllib3.connection.VerifiedHTTPSConnection object at 0xf5deb210>: Failed to establish a new connection: [Errno -3] Temporary failure in name resolution
```

And zesty failed with what PR #1187 will fix

PR #1190, opened 2017-03-14
.travis.yml: Don't use sudo to install dependencies as it's not even installed yet

Roadmaster:
LP: [#162788](https://bugs.launchpad.net/snapcraft/+bug/1672788)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ env:
 
 matrix:
   include:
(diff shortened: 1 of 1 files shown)


elopio:
This is wrong. The CLA check doesn't run in docker, it runs directly in the travis machine, and that one needs sudo. The problem in the bug is different, probably caused when I added the CLA check, but not related to sudo.

elopio:
I proposed #1191 instead.

PR #1192, opened 2017-03-14
repo: refactor into a package.

sergiusens:
Refactoring repo into a package makes it extremely easy to add
new functionality outside of apt/deb by adding such enablement
into `__init__` and by extending `BaseRepo`.

LP: #1602258


sergiusens:
@Conan-Kudo here's take 1

snapcraft/__init__.py
```diff
@@ -138,15 +138,15 @@
 
   - build-packages: [deb, deb, deb...]
 
-    A list of Ubuntu packages to install on the build host before building
+    A list of packages to install on the build host before building
```
elopio: It would be worth explaining where are the packages from. For now, maybe add: currently, only packages from the ubuntu archive of the currently installed distribution are supported.

snapcraft/internal/repo/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+    raise NotImplementedError()
+
+
+if _is_deb_based():
+    from . import _deb
+    Repo = _deb.Ubuntu                                                # noqa
+    get_packages_for_source_type = _deb.get_packages_for_source_type  # noqa
+    get_pkg_libs = _deb.get_pkg_libs                                  # noqa
+    install_build_packages = _deb.install_build_packages              # noqa
+    is_package_installed = _deb.is_package_installed                  # noqa
```
elopio: I don't like these statements that run during import.
Why don't you put them in a method get_repo? or a class method of Repo? This would return a class that implements the right functions depending on the current platform.

snapcraft/internal/repo/_base.py
```diff
@@ -0,0 +1,196 @@
...
+        is called all along the core of snapcraft to obtain the
+        package_names and store them in some repo defined location to
+        later unpack.
+
+        :param str package_names: list of packages to obtain.
+        """
+        raise NotImplemented()
+
+    def unpack(self, rootdir):
+        """Unpack obtained packages into rootdir.
```
elopio: destination instead of rootdir?

snapcraft/internal/repo/_base.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def get(self, package_names):
+        """Get the packages.
+
+        This method needs to be implemented by the inheriting class. It
+        is called all along the core of snapcraft to obtain the
+        package_names and store them in some repo defined location to
+        later unpack.
+
+        :param str package_names: list of packages to obtain.
```
elopio: it would be useful to document the return value.

snapcraft/__init__.py
```diff
@@ -138,15 +138,15 @@
 
   - build-packages: [deb, deb, deb...]
```
Conan-Kudo: I would change this to `pkg` instead of `deb`.

snapcraft/__init__.py
Conan-Kudo: I would not do this. Leaving this part generic is perfectly fine.

snapcraft/tests/repo/test_base.py
```diff
@@ -0,0 +1,270 @@
...
+
+        for test_file in self.files:
+            self.assertThat(
+                test_file['path'], FileContains(test_file['expected']))
+
+
+class FixShebangTestCase(RepoBaseTestCase):
+
+    scenarios = [
+        ('python bin dir', {
```
Conan-Kudo: I would add tests here for `python2` as well. I've seen this show up more often.

snapcraft/__init__.py
Conan-Kudo: I would not suggest doing that. Leaving this part generic is perfectly fine.

snapcraft/internal/repo/__init__.py
sergiusens: because of backwards compatibility and cyclic dependencies... try it locally and see the mess it is to keep backwards compat ;-)

snapcraft/internal/repo/_base.py
sergiusens: yeah

snapcraft/internal/repo/_base.py
sergiusens: gotcha

snapcraft/__init__.py
sergiusens: indeed

snapcraft/tests/repo/test_base.py
sergiusens: python2 and python3... I noticed this while refactoring and was going to fix in a separate PR

snapcraft/internal/repo/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+def is_package_installed():
+    """Stub method for the platform specific implementation."""
+    raise NotImplementedError()
+
+
+if _is_deb_based():
+    from . import _deb
+    Repo = _deb.Ubuntu                                                # noqa
+    get_packages_for_source_type = _deb.get_packages_for_source_type  # noqa
+    get_pkg_libs = _deb.get_pkg_libs                                  # noqa
```
sergiusens: doing this work brought up this inconsistency, which I wish to change... given that this is only in master and not in a tagged revision I would want to change this to `get_package_libraries`

@elopio just in case you miss it ^

snapcraft/internal/repo/_base.py
sergiusens: hmm, rootdir is approriate though (given the language of repos)

snapcraft/internal/repo/_base.py
sergiusens: next time you ask me to do this, fetch out the syntax or get us going with markdown for docs :-)

snapcraft/__init__.py
sergiusens: fixed

snapcraft/internal/repo/_base.py
sergiusens: fixed

snapcraft/__init__.py
sergiusens: Right, if we want to be generic, this messaging has to be generic. Which is why I also updated the errors mentioning Ubuntu (given that we already work on debian too :wink:)

snapcraft/internal/repo/__init__.py
sergiusens: OK, I was being dumb, the next series of commits being pushed in a bit solves all this. I verified and we should be good with this change.

sergiusens:
I am ready for another review

integration_tests/test_multiarch_stage_packages.py
```diff
@@ -31,6 +31,6 @@ def test_bad_multiarch_stage_packages(self):
         self.assertThat(
             exception.output, Contains(
                 "Error downloading stage packages for part 'my-part': The "
-                "Ubuntu package 'hello:fake-arch' was not found.\n"
+                "package 'hello:fake-arch' was not found.\n"
```
Conan-Kudo: Is this an enforced convention? At least in DNF, pushing an arch works by using `pkg.arch` rather than `pkg:arch`.

integration_tests/test_stage_package_version.py
```diff
@@ -27,6 +27,6 @@ def test_stage_package_gets_version(self):
             self.run_snapcraft, 'pull', 'stage-package-version')
         self.assertIn(
             "Error downloading stage packages for part 'hello': "
-            "The Ubuntu package 'hello=x.y-z' was not found.",
+            "The package 'hello=x.y-z' was not found.",
```
Conan-Kudo: Same as before... Is this enforced convention, or just apt-ism in the assert?

snapcraft/_store.py
```diff
@@ -259,7 +259,7 @@ def list_keys():
 
 
 def create_key(name):
-    if not repo.is_package_installed('snapd'):
+    if not repo.Repo.is_package_installed('snapd'):
         raise EnvironmentError(
             'The snapd package is not installed. In order to use '
             '`create-key`, you must run `apt install snapd`.')
```
Conan-Kudo: This should be more generic somehow.

snapcraft/_store.py
```diff
@@ -303,7 +303,7 @@ def _maybe_prompt_for_key(name):
 
 
 def register_key(name):
-    if not repo.is_package_installed('snapd'):
+    if not repo.Repo.is_package_installed('snapd'):
         raise EnvironmentError(
             'The snapd package is not installed. In order to use '
             '`register-key`, you must run `apt install snapd`.')
```
Conan-Kudo: Likewise, should be more generic.

snapcraft/_store.py
```diff
@@ -349,7 +349,7 @@ def _generate_snap_build(authority_id, snap_id, grade, key_name,
 
 
 def sign_build(snap_filename, key_name=None, local=False):
-    if not repo.is_package_installed('snapd'):
+    if not repo.Repo.is_package_installed('snapd'):
         raise EnvironmentError(
             'The snapd package is not installed. In order to use '
             '`sign-build`, you must run `apt install snapd`.')
```
Conan-Kudo: Once more, with feeling! Should be more generic...

snapcraft/internal/lifecycle.py
```diff
@@ -325,7 +325,7 @@ def _tar_filter(tarinfo):
 
 
 def cleanbuild(project_options, remote=''):
-    if not repo.is_package_installed('lxd'):
+    if not repo.Repo.is_package_installed('lxd'):
         raise EnvironmentError(
             'The lxd package is not installed, in order to use `cleanbuild` '
             'you must install lxd onto your system. Refer to the '
```
Conan-Kudo: Genericized, please?

snapcraft/internal/repo/_platform.py
```diff
@@ -0,0 +1,36 @@
...
+from ._deb import Ubuntu
+
+_DEB_BASED_PLATFORM = [
+    'Ubuntu',
+    'Debian',
+]
+
+
+def _is_deb_based():
+    return platform.linux_distribution()[0] in _DEB_BASED_PLATFORM
```
Conan-Kudo: Forewarning here, `platform.linux_distribution()` is going away in Python 3.7. Use [`distro`](https://pypi.python.org/pypi/distro). You don't need to address this now if you don't want to, but it's a good time to warn about this.

snapcraft/plugins/copy.py
```diff
@@ -115,7 +115,7 @@ def _link_or_copy(source, destination, boundary):
         normalized = os.path.normpath(os.path.join(destination_dirname, link))
         if os.path.isabs(link) or not normalized.startswith(boundary):
             # Only follow symlinks that are NOT pointing at libc (LP: #1658774)
-            if link not in snapcraft.repo.get_pkg_libs('libc6'):
+            if link not in snapcraft.repo.Repo.get_package_libraries('libc6'):
```
Conan-Kudo: `libc6` goes by different names in different distros. For example, RPM based distributions call this `glibc`.

snapcraft/plugins/dump.py
```diff
@@ -57,7 +57,7 @@ def _link_or_copy(source, destination, boundary):
         normalized = os.path.normpath(os.path.join(destination_dirname, link))
         if os.path.isabs(link) or not normalized.startswith(boundary):
             # Only follow symlinks that are NOT pointing at libc (LP: #1658774)
-            if link not in snapcraft.repo.get_pkg_libs('libc6'):
+            if link not in snapcraft.repo.Repo.get_package_libraries('libc6'):
```
Conan-Kudo: `libc6` goes by different names in different distros. For example, RPM based distributions call this `glibc`.

snapcraft/tests/plugins/test_copy.py
```diff
@@ -335,7 +335,7 @@ def test_copy_symlinks_to_libc(self):
 
         # Even though this symlink is absolute, since it's to libc the copy
         # plugin shouldn't try to follow it or modify it.
-        libc_libs = snapcraft.repo.get_pkg_libs('libc6')
+        libc_libs = snapcraft.repo.Repo.get_package_libraries('libc6')
```
Conan-Kudo: `libc6` goes by different names in different distros. For example, RPM based distributions call this `glibc`.

snapcraft/tests/plugins/test_dump.py
```diff
@@ -170,7 +170,7 @@ def test_dump_symlinks_to_libc(self):
 
         # Even though this symlink is absolute, since it's to libc the copy
         # plugin shouldn't try to follow it or modify it.
-        libc_libs = snapcraft.repo.get_pkg_libs('libc6')
+        libc_libs = snapcraft.repo.Repo.get_package_libraries('libc6')
```
Conan-Kudo: `libc6` goes by different names in different distros. For example, RPM based distributions call this `glibc`.

snapcraft/internal/repo/_platform.py
sergiusens: yeah, I know about this warning, wasn't aware of the replacement, will do this in a different branch/PR though

Conan-Kudo: @msuchy had recently replaced our usage of `platform.linux_distribution()` with `python-distro` in [Mock](https://github.com/rpm-software-management/mock). That's how I knew about this.

Conan-Kudo: @xsuchy had recently replaced our usage of `platform.linux_distribution()` with `python-distro` in [Mock](https://github.com/rpm-software-management/mock). That's how I knew about this. It's fine that you don't address this now. But I wanted to let you know of the replacement.

integration_tests/test_multiarch_stage_packages.py
sergiusens: This is apt-ism, when dnf is implemented this test should just be expanded to behave differently depending on the platform it runs on, the error message is generated generically though

integration_tests/test_stage_package_version.py
sergiusens: Nope, this is not enforced, implemented purely in the `Ubuntu` class and the `.` syntax can be implemented in whatever the rpm one is called.

integration_tests/test_multiarch_stage_packages.py
Conan-Kudo: Okay, then.

integration_tests/test_stage_package_version.py
Conan-Kudo: Cool!

snapcraft/_store.py
sergiusens: I could implement a class method to get equivalents

Conan-Kudo: That would be a good idea. 👍

sergiusens:
To make this PR slightly smaller I create #1194

snapcraft/internal/repo/_platform.py
xsuchy: http://miroslav.suchy.cz/blog/archives/2016/10/31/how_to_replace_deprecated_platform_distro/index.html

snapcraft/_store.py
sergiusens: and given some thought, in a followup PR to not make this too big of a review

snapcraft/internal/repo/_platform.py
sergiusens: thanks, this will be a problem wrt wanting to have single packaging across xenial, yakkety and zesty but I'll figure it out

```
$ rmadison python3-distro
 python3-distro | 1.0.1-2 | zesty/universe | all
```

sergiusens:
@Conan-Kudo If only access to package name changes remain, I'll address that in a follow-up PR to not make this grow more than it needs to.

snapcraft/internal/repo/_base.py
elopio: I liked the google syntax you gave me yesterday. But I don't mind about this sphinx one either. Your choice.

snapcraft/internal/repo/_base.py
elopio: with the comment in the docstring, this is ok for me.

sergiusens:
oh great `Could not resolve proxy: squid.internal` ...

sergiusens:
meh `fatal: unable to access 'https://go.googlesource.com/crypto/': Proxy CONNECT aborted`

sergiusens:
shared-ros, what a pain you were!

PR #1193, opened 2017-03-14
repo: track per-part build-packages.

josepht:
LP: #1547158


josepht:
Depends on https://github.com/snapcore/snapcraft/pull/1185

integration_tests/test_pull_properties.py
```diff
@@ -58,11 +59,26 @@ def test_pull(self):
...
+        self.assertTrue(len(state.assets['build-packages']) > 0)
         self.assertIn('hello=2.10-1', state.assets['stage-packages'])
+        self.assertIn('make=4.1-6', state.assets['build-packages'])
+
+    def test_pull_global_build_packages(self):
+        project_dir = 'asset-tracking-global'
+        self.run_snapcraft('pull', project_dir)
+
+        state_file = os.path.join(
+            self.parts_dir, project_dir, 'state', 'pull')
```
kyrofa: Wait, no, I thought we talked about this. You're now tracking duplicate global state in every part. I thought the global state tracking was going to be used for this? Has that plan changed?

integration_tests/test_pull_properties.py
```diff
@@ -58,11 +59,26 @@ def test_pull(self):
...
         with open(state_file) as f:
             state = yaml.load(f)
 
         # Verify that the correct version of 'hello' is installed
         self.assertTrue(len(state.assets['stage-packages']) > 0)
+        self.assertTrue(len(state.assets['build-packages']) > 0)
         self.assertIn('hello=2.10-1', state.assets['stage-packages'])
+        self.assertIn('make=4.1-6', state.assets['build-packages'])
+
+    def test_pull_global_build_packages(self):
```
kyrofa: Haha, I got confused. This test is really for testing specifically that global build packages is _not_ tracked per part. The name doesn't really imply that.

josepht:
This needs to be re-kicked.

kyrofa:
> This needs to be re-kicked.

Done.

snapcraft/internal/lifecycle.py
```diff
@@ -243,6 +244,18 @@ def _run_step(self, step, part, part_names):
         common.env.extend(self.config.project_env())
 
         part = _replace_in_part(part)
+
+        # Add the annotated list of build packages
+        part_build_packages = part._part_properties['build-packages']
+        part.build_packages = []
+        for pkg in self.build_packages:
```
josepht: self.build_packages is the list of *all* the build packages.  The loop is to get only the versioned package that are used by this part.

kyrofa: Right. Let me explain my confusion with an example:

```yaml
name: foo
# ...
build-packages: [bar]
parts:
  part1:
    plugin: nil
    build-packages: [baz]
```

At this point in the code, the build packages will have been installed, and self.build_packages will contain, say, `['bar=1.2', 'baz=3,4']`. Right? Then you loop over these, and compare each one to the build-packages for the part, which is just 'baz', and doesn't match anything. So then you split the package name and version apart, and place 'bar' and 'baz' into `part.build_packages`. Do I understand correctly?

josepht: it's adding `pkg` to the parts build-package list, which would match for `baz=3.4` but not for `bar=1.2` since it is not in the parts build-package list from the originial snapcraft.yaml.  Neither `bar` nor `bar=1.2` will match within the loop and thus won't (correctly) get added to the part's build-package list in the assets dict.

kyrofa: Ah, thanks for bearing with me there. I'm good with this.

josepht: @kyrofa are you satisfied on this point after our HO discussion?

integration_tests/snaps/asset-tracking-global/snapcraft.yaml
```diff
@@ -0,0 +1,11 @@
+name: asset-tracking-global
+version: 0.1
+summary: A snap with a global build-package
+description: A snap with a global build-package
+architectures: [all]
+confinement: strict
+build-packages: ['make=4.1-6']
```
kyrofa: Is this the build-package that isn't available on later ubuntu releases?

integration_tests/snaps/build-package-version-global/snap/snapcraft.yaml
```diff
@@ -0,0 +1,14 @@
+name: build-package-version
+version: '0.1'
+summary: install a specific version of a build-package
+description: |
+  Install a specific version of  a build-package.  One part with
+  a good version and one with a bad version
```
kyrofa: This description doesn't look accurate.

snapcraft/internal/lifecycle.py
kyrofa: Indeed.

integration_tests/snaps/asset-tracking-global/snapcraft.yaml
josepht: merged

integration_tests/snaps/build-package-version-global/snap/snapcraft.yaml
josepht: updated

integration_tests/snaps/asset-tracking/snapcraft.yaml
```diff
@@ -9,3 +9,4 @@ parts:
   asset-tracking:
     plugin: nil
     stage-packages: ['hello=2.10-1']
+    build-packages: ['make=4.1-6']
```
kyrofa: I'm going to run the zesty autopkgtests before this merges. Will this pass? :wink: .

integration_tests/snaps/asset-tracking-global/snapcraft.yaml
```diff
@@ -0,0 +1,11 @@
+name: asset-tracking-global
```
kyrofa: How is this snap different from build-package-version-global?

integration_tests/test_pull_properties.py
```diff
@@ -58,11 +59,30 @@ def test_pull(self):
...
         # Verify that the correct version of 'hello' is installed
         self.assertTrue(len(state.assets['stage-packages']) > 0)
+        self.assertTrue(len(state.assets['build-packages']) > 0)
         self.assertIn('hello=2.10-1', state.assets['stage-packages'])
+        self.assertIn('make=4.1-6', state.assets['build-packages'])
+
+    def test_pull_global_build_packages_are_excluded(self):
+        """
+        Ensure global build-packages are not included in each part's
+        build-packages data.
```
kyrofa: Very clear, thank you!

snapcraft/internal/lifecycle.py
```diff
@@ -109,12 +109,13 @@ def execute(step, project_options, part_names=None):
...
     config = snapcraft.internal.load_config(project_options)
     repo.Repo.install_build_packages(config.build_tools)
+    build_packages = repo.Repo.get_installed_package_list(config.build_tools)
 
     if (os.environ.get('SNAPCRAFT_SETUP_CORE') and
             config.data['confinement'] == 'classic'):
         _setup_core(project_options.deb_arch)
 
-    _Executor(config, project_options).run(step, part_names)
+    _Executor(config, project_options, build_packages).run(step, part_names)
```
kyrofa: Now that we actually have a function for extracting build package versions using the repo and config (which is already passed to the executor), we can probably stop passing them around and calculate them when needed. When is the latest we could do it? Does it even need to be in the lifecycle?

snapcraft/internal/lifecycle.py
```diff
@@ -243,6 +245,18 @@ def _run_step(self, step, part, part_names):
...
         part = _replace_in_part(part)
+
+        # Add the annotated list of build packages
+        part_build_packages = part._part_properties['build-packages']
+        part.build_packages = []
+        for pkg in self.build_packages:
+            if pkg in part_build_packages:
+                part.build_packages.append(pkg)
+            else:
+                pkg_name, version = repo._deb._get_pkg_name_parts(pkg)
```
kyrofa: This is calling a private method on a private module. Note also that the build-packages syntax is package-format-agnostic, so there's no reason to keep this function specific to debs. Can we promote this method within repo to somewhere more public? `repo.<something>` would be nice.

snapcraft/internal/repo/_deb.py
```diff
@@ -227,6 +228,20 @@ def install_build_packages(cls, package_names):
                     .format(e))
 
     @classmethod
+    def get_installed_package_list(cls, package_names):
```
kyrofa: `get_installed_build_packages()` might be more intuitive.

snapcraft/tests/repo/test_deb.py
```diff
@@ -250,6 +250,7 @@ def test_install_build_package(
         self.install_test_packages(self.test_packages)
 
         installable = self.get_installable_packages(self.test_packages)
+        installable.append('versioned-package=0.2')
```
kyrofa: Wait... didn't I comment on this already? It's not needed, right?

josepht: yes, merge issues

integration_tests/snaps/asset-tracking/snapcraft.yaml
josepht: Ooops! Fixing

integration_tests/snaps/asset-tracking-global/snapcraft.yaml
josepht: It's not, fixing.

snapcraft/internal/lifecycle.py
josepht: Updated.  I didn't get it out of lifecycle but moved it a bit deeper.

snapcraft/internal/lifecycle.py
josepht: exposed is as repo.get_pkg_name_parts()

snapcraft/internal/repo/_deb.py
josepht: Sounds good. Changed.

snapcraft/tests/repo/test_deb.py
josepht: fixed

integration_tests/snaps/asset-tracking/snapcraft.yaml
josepht: fixed

integration_tests/snaps/asset-tracking-global/snapcraft.yaml
josepht: removed.

josepht:
Updated for @kyrofa 's comments.

josepht:
The integration tests timed out.  Can someone please re-kick the travis tests?

snapcraft/internal/lifecycle.py
```diff
@@ -243,6 +243,20 @@ def _run_step(self, step, part, part_names):
         common.env.extend(self.config.project_env())
 
         part = _replace_in_part(part)
+
+        # Add the annotated list of build packages
+        build_packages = repo.Repo.get_installed_build_packages(
```
kyrofa: My only concern with having this here is that it'll get re-calculated for every step of the same part, which will slow things down. Is there a place where we can calculate it only once? Perhaps where the part is created, or something similar? What do you think?

josepht: Updated.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -333,9 +334,24 @@ def pull(self, force=False):
     def mark_pull_done(self):
         pull_properties = self.code.get_pull_properties()
 
+        # Add the annotated list of build packages
```
kyrofa: Ah ha, that's what I'm talking about! However, at this point, is there any advantage to actually storing it in the class instance at all? Or should we just store it in a temporary variable here just to shove into the state?

josepht: Updated to use a local variable

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -338,19 +338,19 @@ def mark_pull_done(self):
         part_build_packages = self._part_properties.get('build-packages', [])
         build_packages = repo.Repo.get_installed_build_packages(
             part_build_packages)
-        self.build_packages = []
+        ann_build_packages = []
```
kyrofa: What does `ann` mean? Annotated? Perhaps `versioned_build_packages` would be better?

kyrofa:
Running autopkgtests for xenial and zesty.

PR #1196, opened 2017-03-16
store: enable delta uploads by default when pushing.

sergiusens:
LP: #1673455


sergiusens:
here you go @squidsoup

squidsoup:
Looks great! :shipit:

PR #1197, opened 2017-03-16
tests: increase the timeout in the shared ros test

elopio:
LP: [#1673621](https://bugs.launchpad.net/snapcraft/+bug/1673621)


sergiusens:
And interestingly the timeout still showed up here. We will need something smarter to solve this. Let's leave this bug open for that cause.

PR #1199, opened 2017-03-17
cleanbuild: packaging independent detection.

sergiusens:
Don't use the repo to detect if lxd is available as this will prevent use when lxd is provided as a snap or manually installed.

It also makes the logic packaging , agnosting, meaning that a translation layer to figure out what provides the `lxc` command is not necessary.

The work here also migrates the detection out of the lifecycle and into the lxd module.

LP: #1666735

### DIFF
--- a/snapcraft/internal/lxd.py
+++ b/snapcraft/internal/lxd.py
@@ -151,9 +151,10 @@ def _get_default_remote():
         default_remote = check_output(['lxc', 'remote', 'get-default'])
     except CalledProcessError:
         raise SnapcraftEnvironmentError(
-            'LXD is either not installed or configured properly, in order '
-            'to use `cleanbuild` you must have a proper LXD setup '
-            'on your system.\nRefer to the documentation at '
+            'You must have LXD installed in order to use cleanbuild. '
+            'However, it is either not installed or not configured '
+            'properly.\n'
+            'Refer to the documentation at '
             'https://linuxcontainers.org/lxd/getting-started-cli.')
     return default_remote.decode(sys.getfilesystemencoding()).strip()
 
--- a/snapcraft/internal/lxd.py
+++ b/snapcraft/internal/lxd.py
@@ -17,8 +17,9 @@
 
 import logging
 import os
+import sys
 from contextlib import contextmanager
-from subprocess import check_call, CalledProcessError
+from subprocess import check_call, check_output, CalledProcessError
 from time import sleep
 
 import petname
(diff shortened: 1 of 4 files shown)


snapcraft/internal/lxd.py
```diff
@@ -147,3 +134,47 @@ def _wait_for_network(self):
...
+    Use the lxc command to query for the default lxd remote. In most
+    cases this will return the local remote.
+
+    :returns: default lxd remote.
+    :rtype: string.
+    :raises snapcraft.internal.errors.SnapcraftEnvironmentError:
+        raised if the lxc call fails.
+    """
+    try:
+        default_remote = check_output(['lxc', 'remote', 'get-default'])
```
elopio: I like this because it's python's ask for forgiveness, not for permission.
But there might be other reasons here. Like lxd is installed, but you are not in the lxd group. You are swallowing the error that lxc returns, so there is no way to know it.
Maybe put a more generic message, and print what lxc returns as an error?

sergiusens: Since I am not redirecting stderr it should be printed out; but if this command fails it probably means you need to go through the setup anyways.

sergiusens: Here's the output after I remove myself from the group:
```
sergiusens@mirkwood:~/source/snapcraft/demos/hooks$ ~/source/snapcraft/bin/snapcraft cleanbuild
Permission denied, are you in the lxd group?
There are either no permissions or the remote 'local' does not exist.
Verify the existing remotes by running `lxc remote list`
To setup a new remote, follow the instructions on
https://linuxcontainers.org/lxd/getting-started-cli/#multiple-hosts
```

elopio: oh, so if you are not in the group you can still get the default remote. Wfm.

snapcraft/internal/lxd.py
```diff
@@ -147,3 +134,47 @@ def _wait_for_network(self):
...
+    except CalledProcessError:
+        raise SnapcraftEnvironmentError(
+             'The lxd package is not installed, in order to use '
+             '`cleanbuild` you must install lxd onto your system. '
+             'Refer to the "Ubuntu Desktop and Ubuntu Server" section on '
+             'https://linuxcontainers.org/lxd/getting-started-cli/'
+             '#ubuntu-desktop-and-ubuntu-server to enable a proper setup.')
```
evandandrea: And what if they're using something other than Ubuntu? I think we should instead suggest they install the lxd snap.

sergiusens: I am fine with that, but I'd prefer to refer to upstream documentation. Their first step in that section is to install the snap but the anchor/title would need a new name.

evandandrea: That works for me. Can we just point them at the root of https://linuxcontainers.org/lxd/getting-started-cli/ and we'll work to get snaps as the first option?

evandandrea: https://github.com/lxc/linuxcontainers.org/pull/245

PR #1202, opened 2017-03-21
python plugin: support pbr/setup.cfg projects.

sergiusens:
Create the entry_points/console_scripts defined in setup.cfg
when using pbr in setup.py

LP: #1670852

### DIFF
--- a/integration_tests/test_python_plugin.py
+++ b/integration_tests/test_python_plugin.py
@@ -128,7 +128,7 @@ class PythonPluginTestCase(integration_tests.TestCase):
             self.stage_dir, 'bin', 'python3_test')
         self.assertThat(python3_entry_point, FileExists())
         console_script = console_script_template.format(
-            version='3', console_scripts='console_scripts')
+            version='3', console_scripts="'console_scripts'")
         self.assertThat(python3_entry_point, FileContains(console_script))
 
     def test_build_does_not_keep_pyc_or_pth_files_in_install(self):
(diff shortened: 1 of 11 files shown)


integration_tests/test_python_plugin.py
```diff
@@ -84,6 +87,48 @@ def test_build_rewrites_shebangs(self):
...
+            for python_dir in ['python2', 'python3']:
+                d = os.path.join(self.path, python_dir)
+                check_output(['git', 'init'], cwd=d)
+                check_output(
+                    ['git', 'config', 'user.name', 'Test User'], cwd=d)
+                check_output(
+                    ['git', 'config', 'user.email', '<test.user@example.com'],
+                    cwd=d)
+                check_output(['git', 'add', '.'], cwd=d)
+                check_output(['git', 'commit', '-m', 'initial'], cwd=d)
```
elopio: why not just running these steps, instead of putting them in a callable?

sergiusens: I didn't have a split API to get this done. Let's create a bug and get it done in the future. I sort of prefer this mechanism though.

sergiusens: iow, this is really a corner case scenario due to some limitations/features in git. If it is used often enough then maybe we can split, but it should be unlikely.

elopio:
The integration tests fail because of this:
# PBR Generated from u'console_scripts'
# PBR Generated from 'console_scripts'

PR #1203, opened 2017-03-21
Add some initial checks to intelligently build the initial snapcraft.yaml

mhall119:
Creating a PR for comments, this is not ready to merge yet.

This extends the 'init' command to take an optional source argument, for example:
    snapcraft init https://github.com/snapcore/snapcraft.git

It will then check the source's name (URL or local directory) for the snap's name, and introspect it's top-level files for hints as to what kind of plugin it might need.

This is a very limited proof of concept, it would need to be expanded to support a wider variety of sources and plugins before landing.

### DIFF
--- a/snapcraft/internal/lifecycle.py
+++ b/snapcraft/internal/lifecycle.py
@@ -39,33 +39,84 @@ from snapcraft.internal import (
 from snapcraft.internal.cache import SnapCache
 from snapcraft.internal.indicators import is_dumb_terminal
 from snapcraft.internal.project_loader import replace_attr
+from snapcraft.internal.sources import _get_source_type_from_uri
 
 
 logger = logging.getLogger(__name__)
 
 
-_TEMPLATE_YAML = """name: my-snap-name # you probably want to 'snapcraft register <name>'
-version: '0.1' # just for humans, typically '1.2+git' or '1.3.2'
-summary: Single-line elevator pitch for your amazing snap # 79 char long summary
+_TEMPLATE_YAML = """name: {name} # you probably want to 'snapcraft register <name>'
+version: '{version}' # just for humans, typically '1.2+git' or '1.3.2'
+summary: {summary} # 79 char long summary
 description: |
-  This is my-snap's description. You have a paragraph or two to tell the
-  most important story about your snap. Keep it under 100 words though,
-  we live in tweetspace and your description wants to look good in the snap
-  store.
+  {description}
 
 grade: devel # must be 'stable' to release into candidate/stable channels
 confinement: devmode # use 'strict' once you have the right plugs and slots
 
+apps:
+  {name}:
+    command: {command} # Command to run this app
+    daemon: {daemon}
+    plugs: {plugs}
+    
 parts:
-  my-part:
+  {name}:
     # See 'snapcraft plugins'
-    plugin: nil
+    plugin: {plugin}
+    source: {source}
+    source-type: {source-type}
 """  # noqa, lines too long.
 
 _STEPS_TO_AUTOMATICALLY_CLEAN_IF_DIRTY = {'stage', 'prime'}
 
+_PLUGIN_HINT_FILES = {
+    'package.json': ['nodejs'],
+    'CMakeLists.txt': ['cmake'],
+    'Makefile': ['make', 'autotools'],
+    'configure': ['autotools'],
+    'setup.py': ['python'],
+    'requirements.txt': ['python'],
+}
 
-def init():
+def _list_root_files(source, source_type):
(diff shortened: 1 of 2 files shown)


popey:
@mhall119 asked me to pop this link in here. @stuartlangridge also looked at doing this kinda thing but differently. https://github.com/stuartlangridge/syg

snapcraft/internal/lifecycle.py
```diff
@@ -76,7 +127,45 @@ def init():
         raise EnvironmentError('snapcraft.yaml already exists!')
     elif os.path.exists('.snapcraft.yaml'):
         raise EnvironmentError('.snapcraft.yaml already exists!')
-    yaml = _TEMPLATE_YAML.strip()
+        
+    # Attempt to pre-populate some package information
+    if source is None:
```
sergiusens: we've been using the `if not source` construct for this.

snapcraft/internal/lifecycle.py
```diff
@@ -76,7 +127,45 @@ def init():
...
+  most important story about your snap. Keep it under 100 words though,
+  we live in tweetspace and your description wants to look good in the snap
+  store.""",
+        'plugin': 'dump',
+        'source': source,
+        'source-type': 'local',
+        'command': 'echo "Hello World"',
+        'daemon': 'none',
+        'plugs': []
+    }# noqa, lines too long.
```
sergiusens: just add a parens in `description` to get rid of the `#noqa`

snapcraft/internal/lifecycle.py
```diff
@@ -39,33 +39,84 @@
...
+    plugin: {plugin}
+    source: {source}
+    source-type: {source-type}
 """  # noqa, lines too long.
 
 _STEPS_TO_AUTOMATICALLY_CLEAN_IF_DIRTY = {'stage', 'prime'}
 
-
-def init():
+_PLUGIN_HINT_FILES = {
```
sergiusens: this is not complete, right?

snapcraft/internal/lifecycle.py
```diff
@@ -39,33 +39,84 @@
...
+    return []
+
+def _guess_plugin(source, source_type):
+    """Step through pre-defined hint files to see if they exist in the project
+    root directory. Assign each plugin type a score based on the presence of
+    files it would use, and return the plugin type with the highest score.
+    
+    Defaults to 'dump' if no hint file is found"""
+    detect_score = {'dump': 0}
+    for line in _list_root_files(source, source_type):
```
sergiusens: this could probably be made more functional with some list comprehensions.

snapcraft/internal/lifecycle.py
```diff
@@ -76,7 +127,45 @@ def init():
         raise EnvironmentError('snapcraft.yaml already exists!')
     elif os.path.exists('.snapcraft.yaml'):
         raise EnvironmentError('.snapcraft.yaml already exists!')
-    yaml = _TEMPLATE_YAML.strip()
+        
+    # Attempt to pre-populate some package information
+    if source is None:
+        source = '.'
+
+    snap_meta = {
```
sergiusens: this should probably be an OrderedDict so that writes to the file are in order.

snapcraft/internal/lifecycle.py
```diff
@@ -76,7 +127,45 @@ def init():
...
+        'daemon': 'none',
+        'plugs': []
+    }# noqa, lines too long.
+    
+    # Use the argument passed to the commandline to determine the source type
+    try:
+        detect_type = _get_source_type_from_uri(source)
+        if detect_type:
+            snap_meta['source-type'] = detect_type
+    except ValueError:
```
sergiusens: instead: 
```python
with contextlib.suppress(ValueError):
    detect_type = ...
    snap_meta[...] = ...
```

sergiusens:
@popey and @mhall119 yeah, @stuartlangridge 's code looks good too with the benefit of looking more moduralized out. I'll let you decide on the approach but having this logic not tightly coupled into the lifecycle is a nice touch.

sergiusens:
OK, I am closing this due to lack of response. Please reopen if you have updates but let me entice you to propose this on the forums instead ;-)

snapcraft/internal/lifecycle.py
mhall119: No, it's just a start, we should add more items to this list to cover more plugins

snapcraft/internal/lifecycle.py
mhall119: Possibly, but it works and it's not code that's frequently run or time sensitive, so I didn't put effort into optimizing it.

PR #1204, opened 2017-03-21
target-arch: decouple target arch from deb, and use a separate field …

piso77:
target field for better matching - aka introduce aliases for some archs

LP: #1674752

### DIFF
--- a/snapcraft/_options.py
+++ b/snapcraft/_options.py
@@ -31,6 +31,7 @@ _ARCH_TRANSLATIONS = {
     'armv7l': {
         'kernel': 'arm',
         'deb': 'armhf',
+        'uts_machine': 'arm',
         'cross-compiler-prefix': 'arm-linux-gnueabihf-',
         'cross-build-packages': ['gcc-arm-linux-gnueabihf'],
         'triplet': 'arm-linux-gnueabihf',
@@ -39,6 +40,7 @@ _ARCH_TRANSLATIONS = {
     'aarch64': {
         'kernel': 'arm64',
         'deb': 'arm64',
+        'uts_machine': 'aarch64',
         'cross-compiler-prefix': 'aarch64-linux-gnu-',
         'cross-build-packages': ['gcc-aarch64-linux-gnu'],
         'triplet': 'aarch64-linux-gnu',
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
What would use the `target` field?

piso77:
The kernel plugin, so i don't have to:

-set_snaparch() {
- # Attempt to find the correct architecture
- case "$UTS_MACHINE" in
- i386|ia64|alpha)
- snaparch="$UTS_MACHINE" ;;
- x86_64)
- snaparch=amd64 ;;
- sparc*)
- snaparch=sparc ;;
- s390*)
- snaparch=s390$(grep -q CONFIG_64BIT=y $KCONFIG_CONFIG && echo x || true) ;;
- ppc*)
- snaparch=$(grep -q CPU_LITTLE_ENDIAN=y $KCONFIG_CONFIG && echo ppc64el || echo powerpc) ;;
- parisc*)
- snaparch=hppa ;;
- mips*)
- snaparch=mips$(grep -q CPU_LITTLE_ENDIAN=y $KCONFIG_CONFIG && echo el || true) ;;
- aarch64)
- snaparch=arm64 ;;
- arm*)
- if grep -q CONFIG_AEABI=y $KCONFIG_CONFIG; then
- if grep -q CONFIG_VFP=y $KCONFIG_CONFIG; then
- snaparch=armhf
- else
- snaparch=armel
- fi
- else
- snaparch=arm
- fi
- ;;
- *)
- snaparch=$(dpkg --print-architecture)
- echo "" >&2
- echo "** ** ** WARNING ** ** **" >&2
- echo "" >&2
- echo "Your architecture doesn't have it's equivalent" >&2
- echo "Debian userspace architecture defined!" >&2
- echo "Falling back to using your current userspace instead!" >&2
- echo "Please add support for $UTS_MACHINE to ${0} ..." >&2
- echo "" >&2
- esac
- forcearch="--target-arch=$snaparch"
-}
-

just to convert from the $UTS_MACHINE arch (that is defined by the linux kernel kbuild system) to an equivalent deb ach

kyrofa:
This seems reasonable to me. Let's run the autopkgtests.

sergiusens:
@kalikiana this is supposed to live here, it is what we already supported once upon a time before moving to user space, it would however provide a good fallback.

PR #1206, opened 2017-03-22
Arm64 witness

elopio:
Just to check if all the tests pass after merging sergio's python console scripts branch.


sergiusens:
too bad the witness wasn't run with arm64 ;-)

elopio:
what do you mean?
xenial-arm64 — autopkgtest finished (success) 
Apparently, we have connection problems on armhf machines though :/

PR #1208, opened 2017-03-22
kernel plugin: fix modprobe output parsing.

sergiusens:
The module path is always the second item.

LP: #1675029


sergiusens:
I am ignoring this classic error:
```ERROR: test_classic_confinement (test_stage.StageTestCase)

test_stage.StageTestCase.test_classic_confinement
```

PR #1211, opened 2017-03-22
sources: add git source tracking

josepht:
LP: #1547158

### DIFF
--- a/integration_tests/__init__.py
+++ b/integration_tests/__init__.py
@@ -198,7 +198,7 @@ class StoreTestCase(TestCase):
 
     def logout(self):
         output = self.run_snapcraft('logout')
-        expected = (r'.*Clearing credentials for Ubuntu One SSO.\n.*\n'
+        expected = (r'.*Clearing credentials for Ubuntu One SSO.\n'
                     r'Credentials cleared.\n.*')
         self.assertThat(output, MatchesRegex(expected, flags=re.DOTALL))
 
--- a/integration_tests/__init__.py
+++ b/integration_tests/__init__.py
@@ -26,7 +26,6 @@ from distutils import dir_util
 import fixtures
 import pexpect
 from unittest import mock
-import testscenarios
 import testtools
 from testtools import content
 from testtools.matchers import MatchesRegex
@@ -34,7 +33,7 @@ from testtools.matchers import MatchesRegex
 from snapcraft.tests import fixture_setup
 
 
-class TestCase(testscenarios.WithScenarios, testtools.TestCase):
+class TestCase(testtools.TestCase):
 
     def setUp(self):
         super().setUp()
--- a/integration_tests/__init__.py
+++ b/integration_tests/__init__.py
@@ -26,6 +26,7 @@ from distutils import dir_util
 import fixtures
 import pexpect
 from unittest import mock
+import testscenarios
 import testtools
 from testtools import content
 from testtools.matchers import MatchesRegex
@@ -33,7 +34,7 @@ from testtools.matchers import MatchesRegex
 from snapcraft.tests import fixture_setup
 
 
-class TestCase(testtools.TestCase):
+class TestCase(testscenarios.WithScenarios, testtools.TestCase):
 
     def setUp(self):
         super().setUp()
--- a/integration_tests/test_pull_properties.py
+++ b/integration_tests/test_pull_properties.py
@@ -18,6 +18,7 @@ import os
 import subprocess
 import yaml
 
+import testscenarios
 from testtools.matchers import FileExists
 
 import integration_tests
@@ -89,7 +90,8 @@ class AssetTrackingTestCase(integration_tests.TestCase):
         self.assertNotIn('hello=2.10-1', state.assets['build-packages'])
 
 
-class GitAssetTrackingTestCase(integration_tests.TestCase):
+class GitAssetTrackingTestCase(testscenarios.WithScenarios,
+                               integration_tests.TestCase):
 
     def _create_git_repo(self, name):
         def _call(cmd):
@@ -160,8 +162,6 @@ class GitAssetTrackingTestCase(integration_tests.TestCase):
             state = yaml.load(f)
 
         self.assertIn('source-details', state.assets)
-
-        print("JOE: {}".format(state.assets['source-details']))
 
         if 'branch' in self.expected:
             self.assertEqual(self.expected['branch'],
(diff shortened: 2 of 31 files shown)


josepht:
This one needs a re-kick please.

integration_tests/test_pull_properties.py
```diff
@@ -53,9 +54,32 @@ def assert_expected_pull_state(self, project_dir):
...
+
+        def _call_with_output(cmd):
+            return subprocess.check_output(cmd).decode('utf-8').strip()
+
+        os.makedirs(name)
+        _call(['git', '-C', name, 'init'])
+        _call(['git', '-C', name, 'config',
+               '--local', 'user.name', 'Test User'])
+        _call(['git', '-C', name, 'config',
+               '--local', 'user.email', 'testuser@example.com'])
```
kyrofa: `--local` is the default, you probably don't need it here.

integration_tests/test_pull_properties.py
```diff
@@ -53,9 +54,32 @@ def assert_expected_pull_state(self, project_dir):
...
+        with open(os.path.join(name, 'testing'), 'w') as fp:
+            fp.write('testing')
+
+        _call(['git', '-C', name, 'add', 'testing'])
+        _call(['git', '-C', name, 'commit', '-am', 'testing'])
+        _call(['git', '-C', name, 'branch', 'feature'])
+        _call(['git', '-C', name, 'tag', 'feature-tag'])
+        return _call_with_output(['git', '-C', name, 'rev-parse', 'HEAD'])
```
kyrofa: Having all of these pointing to the same commit seems like it's possible for the test to pass even if the hash-saving code saves the wrong hash. Do you think adding separate commits for master, feature, and feature-tag would make for more robust tests?

snapcraft/internal/sources/_git.py
```diff
@@ -90,3 +90,25 @@ def pull(self):
             self._pull_existing()
         else:
             self._clone_new()
+        self.assets = self._get_source_details()
+
+    def _get_source_details(self):
```
kyrofa: So let's fast-forward here: how do you envision this information being used when we actually save it into the `$SNAP/snap/snapcraft.yaml` and need to re-consume it? For example, just a few lines above there's the logic for erroring out if both `source-commit` and `source-branch` are provided, and yet you're saving both here (among other things). How do you plan on that being disambiguated?

integration_tests/test_pull_properties.py
josepht: removed

integration_tests/test_pull_properties.py
josepht: updated as well as the unit test

snapcraft/internal/sources/_git.py
josepht: updated to only get the commit if nothing else has been supplied.  I plan for each part to get only what is supplied and to only look up whichever bit is needed to correctly identify the exact source.  Thus I can just directly add 'source-branch', 'source-tag', 'source-commit', etc.  I might have to add some logic to each source class if there are identifiers that are only needed for some source types.

josepht:
The conflicts have been resolved.

integration_tests/test_pull_properties.py
```diff
@@ -68,7 +106,52 @@ def test_pull(self):
         self.assertTrue(len(state.assets['stage-packages']) > 0)
         self.assertTrue(len(state.assets['build-packages']) > 0)
         self.assertIn('hello=2.10-1', state.assets['stage-packages'])
-        self.assertIn('hello=2.10-1', state.assets['build-packages'])
+        self.assertIn('source-details', state.assets)
+
+    def test_pull_git(self):
```
sergiusens: These tests could probably be better suited for test scenarios in a `GitAssetTrackingTestCase` class inheriting for thing current one. This would also make it more maintainable as `_create_git_repo` could be moved to this new class.

snapcraft/internal/sources/_base.py
```diff
@@ -37,6 +37,7 @@ def __init__(self, source, source_dir, source_tag=None, source_commit=None,
         self.source_branch = source_branch
         self.source_depth = source_depth
         self.source_checksum = source_checksum
+        self.assets = None
```
sergiusens: `assets` is a bad name, it is a source asset for snapcraft itself, but for the source class in question it is not an asset.

snapcraft/internal/sources/_git.py
sergiusens: This is missing `source-checksum`. I also don't understand how this will work if done in `__init__` where the source hasn't been pulled to `source_dir` yet.

sergiusens: It seems like this needs to be a public method the source instance for the part (pluginhandler) can call into at the appropriate time instead of at init.

josepht:
@sergiusens Git._get_source_details() is called at the end of Git.pull() not in __init__.

snapcraft/internal/sources/_base.py
josepht: agreed, updated to 'source_details'

integration_tests/test_pull_properties.py
josepht: updated.  I'm thinking test_pull_properties.py isn't the best place for these tests.  Any objections to having all these in test_source_tracking.py?

integration_tests/__init__.py
```diff
@@ -26,14 +26,15 @@
...
+import testscenarios
 import testtools
 from testtools import content
 from testtools.matchers import MatchesRegex
 
 from snapcraft.tests import fixture_setup
 
 
-class TestCase(testtools.TestCase):
+class TestCase(testscenarios.WithScenarios, testtools.TestCase):
```
sergiusens: don't do this.

integration_tests/test_pull_properties.py
```diff
@@ -92,80 +128,47 @@ def _add_and_commit_file(path, filename, contents=None, message=None):
...
-
-        state_file = os.path.join(
-            self.parts_dir, project_dir, 'state', 'pull')
-        self.assertThat(state_file, FileExists())
-        with open(state_file) as f:
-            state = yaml.load(f)
-
-        self.assertTrue(len(state.assets['build-packages']) == 0)
-        self.assertNotIn('hello=2.10-1', state.assets['build-packages'])
+        print("JOE: {}".format(state.assets['source-details']))
```
sergiusens: loose cannon

integration_tests/__init__.py
sergiusens: look at the other integration tests using scenarios, this really really isn't needed.

integration_tests/test_pull_properties.py
josepht: I inadvertently broke my pre-commit hook that checks for that.  Fixed now. :-)

integration_tests/__init__.py
josepht: Fixed

PR #1212, opened 2017-03-22
store: better retry strategy for GETs.

sergiusens:
Use the retry class which allows for more fine grained settings such as forcing a retry when a status code is returned.

LP: #1671817

### DIFF
--- a/snapcraft/storeapi/__init__.py
+++ b/snapcraft/storeapi/__init__.py
@@ -286,9 +286,21 @@ class StoreClient():
                 name, download_path))
             return
         logger.info('Downloading {}'.format(name, download_path))
-        request = self.cpi.get(download_url, stream=True)
-        request.raise_for_status()
-        download_requests_stream(request, download_path)
+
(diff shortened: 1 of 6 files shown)


snapcraft/storeapi/__init__.py
```diff
@@ -278,9 +286,22 @@ def _download_snap(self, name, channel, arch, download_path,
                 name, download_path))
             return
         logger.info('Downloading {}'.format(name, download_path))
-        request = self.cpi.get(download_url, stream=True)
-        request.raise_for_status()
-        download_requests_stream(request, download_path)
+
+        # HttpAdapter cannot help here as this is a stream.
+        # LP: #1617765
+        not_downloaded = True
```
elopio: I find it easier to have positive booleans. Use `downloaded`, and `not downloaded`.

elopio:
This is less pretty than before, but still looks good. I just left a nit about the boolean name.

PR #1216, opened 2017-03-23
cleanbuild: don't copy cache into container.

sergiusens:
The cache copy logic needs some more design. Currently it copies in the cache regardless of the  target. We need to be smart and only copy the matching hashes. It also needs to weigh in non local remotes and provide user visible feedback in the case where the cache is copied to a non local remote.

LP: #1669730

### DIFF
--- a/snapcraft/internal/lxd.py
+++ b/snapcraft/internal/lxd.py
@@ -1,7 +1,7 @@
 #!/usr/bin/python3
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright (C) 2016 Canonical Ltd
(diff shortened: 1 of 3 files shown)


popey:
Maybe I'm doing something wrong, but this still seems to be copying the cache.

```
alan@gort:~/Development/Snappy/triggerhappy$ /home/alan/Development/snapcraft/snapcraft/bin/snapcraft cleanbuild --remote colossus
Creating snapcraft-wildly-safe-eft
Starting snapcraft-wildly-safe-eft
Setting up container with project assets
./
./snapcraft.yaml
./scripts/
./scripts/launch-th.sh
./parts/
Copying snapcraft cache into container
```

I am seeing large data transfers to the remote host while that step happens.

PR #1218, opened 2017-03-23
asset-tracking: use a more likely to be found global build-package

josepht:
LP: #1675447

### DIFF
--- a/integration_tests/snaps/build-package-version-global/snap/snapcraft.yaml
+++ b/integration_tests/snaps/build-package-version-global/snap/snapcraft.yaml
@@ -2,8 +2,7 @@ name: build-package-version
 version: '0.1'
 summary: install a specific version of a build-package
 description: |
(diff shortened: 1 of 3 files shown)


integration_tests/snaps/build-package-version-global/snap/snapcraft.yaml
```diff
@@ -0,0 +1,14 @@
+name: build-package-version
+version: '0.1'
+summary: install a specific version of a build-package
+description: |
+  Install a specific version of  a build-package.  One part with
+  a good version and one with a bad version
```
kyrofa: This description is no longer correct.

josepht: Updated

elopio:
integration tests passed in zesty and yakkety. This is good to go. And I filed a bug to get a better fix.

PR #1219, opened 2017-03-23
kernel plugin: learn how to assemble the ubuntu config using kconfigflavour

piso77:
Instead of gluing together shell scripts or anything, simply teach snapcraft how to extract the entire ubuntu config from debian.$branch/config dir.

LP: #1675454


snapcraft/plugins/kbuild.py
```diff
@@ -114,14 +124,44 @@ def __init__(self, name, options, project):
 
     def do_base_config(self, config_path):
         # if kconfigfile is provided use that
+        # elif kconfigflavour is provided, assemble the ubuntu.flavour config
         # otherwise use defconfig to seed the base config
-        if self.options.kconfigfile is None:
+        if self.options.kconfigfile:
+            shutil.copy(self.options.kconfigfile, config_path)
+        elif self.options.kconfigflavour:
```
kyrofa: Are `kconfigfile` and `kconfigflavour` mutually exclusive, then? If so, perhaps the schema should indicate that as the case.

kyrofa: Judging from the descriptions for the other options, perhaps that's a more general comment on the plugin :stuck_out_tongue: .

elopio: This big block inside the elif statement should be extracted to its own function.

PR #1221, opened 2017-03-24
pluginhandler: factor out state bits into the state package.

josepht:
LP: #1675881

### DIFF
--- a/snapcraft/internal/states/_state.py
+++ b/snapcraft/internal/states/_state.py
@@ -14,6 +14,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import os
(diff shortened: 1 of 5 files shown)


snapcraft/internal/states/_state.py
```diff
@@ -78,3 +79,17 @@ def _get_differing_keys(dict1, dict2):
...
+    state = None
+    state_file = step_state_file(state_dir, step)
+    if os.path.isfile(state_file):
+        with open(state_file, 'r') as f:
+            state = yaml.load(f.read())
+
+    return state
+
+
+def step_state_file(state_dir, step):
```
sergiusens: this needs a get prefix as well

josepht: Done

PR #1223, opened 2017-03-28
kernel plugin: check_config() and check_initrd() support

piso77:
Make snapcraft smarter by parsing .config and initrd modules list and printing out if there are mandatory options missing, or if a boot required piece of code is not builtin or part packaged into initrd

LP: #1676806

### DIFF
--- a/snapcraft/plugins/kbuild.py
+++ b/snapcraft/plugins/kbuild.py
@@ -163,6 +163,8 @@ class KBuildPlugin(BasePlugin):
         except OSError as e:
             raise RuntimeError('Unable to access {!r}: '
                                '{}'.format(e.filename, e.strerror))
+    def get_config_path(self):
+        return os.path.join(self.builddir, '.config')
 
     def do_base_config(self, config_path):
         # if kconfigfile is provided use that
@@ -222,13 +224,19 @@ class KBuildPlugin(BasePlugin):
                  ['CONFIG_PREFIX={}'.format(self.installdir)] +
                  self.make_install_targets)
 
-    def build(self):
+    def prep_build(self):
         super().build()
 
-        config_path = os.path.join(self.builddir, '.config')
+        config_path = self.get_config_path()
 
         self.do_base_config(config_path)
         self.do_patch_config(config_path)
         self.do_remake_config()
+
+    def finish_build(self):
         self.do_build()
         self.do_install()
+
+    def build(self):
+        self.prep_build()
+        self.finish_build()
--- a/snapcraft/plugins/kbuild.py
+++ b/snapcraft/plugins/kbuild.py
@@ -224,7 +224,7 @@ class KBuildPlugin(BasePlugin):
                  ['CONFIG_PREFIX={}'.format(self.installdir)] +
                  self.make_install_targets)
 
-    def prep_build(self):
+    def do_configure(self):
         super().build()
 
         config_path = self.get_config_path()
@@ -238,5 +238,5 @@ class KBuildPlugin(BasePlugin):
         self.do_install()
 
     def build(self):
-        self.prep_build()
+        self.do_configure()
         self.finish_build()
--- a/snapcraft/plugins/kbuild.py
+++ b/snapcraft/plugins/kbuild.py
@@ -233,10 +233,7 @@ class KBuildPlugin(BasePlugin):
         self.do_patch_config(config_path)
         self.do_remake_config()
 
-    def finish_build(self):
-        self.do_build()
-        self.do_install()
-
     def build(self):
         self.do_configure()
-        self.finish_build()
+        self.do_build()
+        self.do_install()
--- a/snapcraft/plugins/kbuild.py
+++ b/snapcraft/plugins/kbuild.py
@@ -163,6 +163,7 @@ class KBuildPlugin(BasePlugin):
         except OSError as e:
             raise RuntimeError('Unable to access {!r}: '
                                '{}'.format(e.filename, e.strerror))
+
     def get_config_path(self):
         return os.path.join(self.builddir, '.config')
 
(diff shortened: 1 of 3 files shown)


sergiusens:
Thanks! Can I start by asking why this is in `kbuild.py` and not in `kernel.py`?

piso77:
Because since i wanted to check .config stuff, i thought kbuild was the correct place - after all .config / Kconfig mangling happens there.

sergiusens:
yeah, but you would never do this when building busybox and I suspect if I run the demo builds that are pending this will fail :wink:

snapcraft/plugins/kernel.py
```diff
@@ -355,11 +379,78 @@ def _copy_dtbs(self):
...
+                    if val == 'Y':
+                        builtin.append(opt)
+                    elif val == 'M':
+                        modules.append(opt)
+
+    def _do_check_config(self):
+        # check the resulting .config has all the necessary options
+        msg = 'Your kernel config is missing some feature that ubuntu core \
+recommend / requires,\nand while we won\'t prevent you from building this\
+kernel snap,\n we suggest you take a look at these:'
```
elopio: This would look a little prettier like this:
```
msg = ("Your kernel config is missing some features that ubuntu core "
            "recommends or requires,\n"
            "and while we won't prevent you from building this "
            "kernel snap,\n "
            "we suggest you take a look at these:")
```

Note that you also have to correct featureS and recommendS.

snapcraft/plugins/kernel.py
```diff
@@ -355,11 +379,78 @@ def _copy_dtbs(self):
...
+        required_opts = required_generic + required_security + \
+            required_snappy + required_systemd
```
elopio: Please use parentheses instead of the `\`.

snapcraft/plugins/kernel.py
```diff
@@ -355,11 +379,78 @@ def _copy_dtbs(self):
...
+            elif opt in modules:
+                continue
+            else:
+                missing.append(opt)
+
+        if missing:
+            logger.warn('\n{}\n'.format(msg))
+            for opt in missing:
+                logger.warn('  - {}'.format(opt))
+            logger.warn('')
```
elopio: It's better to call logger.warn only once. First build the message, and then pass it full to warn.
The log could have some format added to the message, like timestamps.

snapcraft/plugins/kernel.py
```diff
@@ -355,11 +379,78 @@ def _copy_dtbs(self):
...
+
+    def _do_check_initrd(self):
+        # check all required_boot[] items are either builtin or part of initrd
+        msg = 'The following feature is boot essential, consider making\n\
+it static or adding the relevant module to initrd:'
```
elopio: parentheses instead of `\` for line break.s

snapcraft/plugins/kbuild.py
```diff
@@ -166,13 +169,19 @@ def do_install(self):
                  ['CONFIG_PREFIX={}'.format(self.installdir)] +
                  self.make_install_targets)
 
-    def build(self):
+    def prep_build(self):
```
elopio: we have a new keyword called prepare. This name would make it confusing. What about something like `prepare_config` or `configure`.

snapcraft/plugins/kbuild.py
```diff
@@ -166,13 +169,19 @@ def do_install(self):
...
         super().build()
```
elopio: This super here is wrong, it should be in the build method.

snapcraft/plugins/kbuild.py
```diff
@@ -166,13 +169,19 @@ def do_install(self):
...
         self.do_patch_config(config_path)
         self.do_remake_config()
+
+    def finish_build(self):
         self.do_build()
         self.do_install()
+
+    def build(self):
+        self.prep_build()
+        self.finish_build()
```
elopio: prepare and finish sounds like a poor separation for this function. Maybe we can find better names. I would like something like:
```

def build(self):
    self.configure()
    self.do_build()
    self.do_install()
```

It's still poor naming because inside build you have another call to build, but that's because snapcraft doesn't yet separate the build and install phases. I hope we can improve that soon.

snapcraft/plugins/kernel.py
```diff
@@ -355,11 +379,78 @@ def _copy_dtbs(self):
...
+            for line in f:
+                tok = line.strip().split('=')
+                items = len(tok)
+                if items == 2:
+                    opt = tok[0].upper()
+                    val = tok[1].upper()
+                    if val == 'Y':
+                        builtin.append(opt)
+                    elif val == 'M':
+                        modules.append(opt)
```
elopio: Can we make this a function that returns builtin and modules, instead of storing them in a global variable?

snapcraft/plugins/kernel.py
```diff
@@ -355,11 +379,78 @@ def _copy_dtbs(self):
...
         super().pull()
         snapcraft.download(
             'ubuntu-core', 'edge', self.os_snap, self.project.deb_arch)
 
+    def build(self):
+        super().prep_build()
+        self._do_parse_config(self.get_config_path())
+        self._do_check_config()
+        self._do_check_initrd()
+        super().finish_build()
```
elopio: Following my suggestions above, this could look like:
```

def build(self):
   self.prepare_config()
   builtin, modules = self.parse_config()
   self.check_config(builtin, modules)
   self.check_initrd(builtin, modules)
   self.do_build()
   self.do_install()
```

snapcraft/plugins/kernel.py
```diff
@@ -355,11 +379,78 @@ def _copy_dtbs(self):
...
+
     def do_install(self):
         super().do_install()
```
elopio: Now that we are here, this method is not necessary if it just calls the parent. You can just remove it.

snapcraft/plugins/kbuild.py
sergiusens: done

PR #1224, opened 2017-03-28
tests: update name registration window limit test.

elopio:
LP: [#1676725](https://bugs.launchpad.net/snapcraft/+bug/1676725)


elopio:
This requires #1225

elopio:
@fgallina @cprov ping. I had to split the store suite because it was timing out in travis. This will require a change on your script to call `TEST_SUITE=store`, instead of `integration`.

sergiusens:
Can you update the branch please?

PR #1225, opened 2017-03-29
channels: Fix staging store test for Tracks.

josepht:
LP: #1677365


elopio:
All good here, I added this commit in my branch that runs on staging because I made it in the snapcore organization, and it's green.

Thanks @josepht !

PR #1229, opened 2017-04-03
sources: add VCS asset tracking

josepht:
supercedes https://github.com/snapcore/snapcraft/pull/1211, https://github.com/snapcore/snapcraft/issues/1213, https://github.com/snapcore/snapcraft/pull/1214, and https://github.com/snapcore/snapcraft/pull/1215

* git support
* bzr support
* svn support
* hg support

LP: #1547158


josepht:
Can someone re-kick the travis test please?

josepht:
The test failures appear to be network related.  Can someone please re-run the tests?

integration_tests/_source_helpers.py
```diff
@@ -0,0 +1,122 @@
...
+def call(cmd):
+    subprocess.check_call(cmd, stdout=subprocess.DEVNULL,
+                          stderr=subprocess.DEVNULL)
+
+
+def call_with_output(cmd):
+    return subprocess.check_output(cmd).decode('utf-8').strip()
+
+
+def create_bzr_repo(name):
```
sergiusens: these could all be fixtures.

integration_tests/test_pull_properties.py
```diff
@@ -86,3 +88,140 @@ def test_pull_global_build_packages_are_excluded(self):
...
+        self.assertIn('source-details', state.assets)
+
+        if 'branch' in self.expected:
+            self.assertEqual(self.expected['branch'],
+                             state.assets['source-details']['branch'])
+        elif 'tag' in self.expected:
+            self.assertEqual(self.expected['tag'],
+                             state.assets['source-details']['tag'])
+        else:
+            self.assertEqual(expected_commit,
```
sergiusens: I thought we were going to record the commit regardless, if that is the case, this needs to live outside of the `else`

integration_tests/test_pull_properties.py
```diff
@@ -86,3 +88,140 @@ def test_pull_global_build_packages_are_excluded(self):
...
+
+        state_file = os.path.join(
+            self.parts_dir, self.part_name, 'state', 'pull')
+        self.assertThat(state_file, FileExists())
+        with open(state_file) as f:
+            state = yaml.load(f)
+
+        self.assertIn('source-details', state.assets)
+
+        if 'branch' in self.expected:
```
sergiusens: if you use smart serialization values in scenarios, this conditional would not be necessary.

integration_tests/test_pull_properties.py
josepht: The issue is that we don't accept a snapcraft.yaml with both a `source-tag` and `source-commit`

.gitmodules
```diff
@@ -1,3 +0,0 @@
-[submodule "docs"]
-	path = docs
-	url = https://github.com/CanonicalLtd/snappy-docs.git
```
elopio: This shouldn't be here, right?

integration_tests/test_pull_properties.py
josepht: Hrm, I thought I was using smart serialization values.  What do you suggest?

.travis.yml
```diff
@@ -14,7 +14,7 @@ env:
   matrix:
   - DOCKER=1 CHECK_CLA=1 DEPENDENCIES="apt install -y git python-launchpadlib"
   - DOCKER=1 TEST_SUITE=static DEPENDENCIES="apt install -y python3-pip && python3 -m pip install -r requirements-devel.txt"
-  - DOCKER=1 TEST_SUITE=unit DEPENDENCIES="apt install -y git libnacl-dev libsodium-dev libffi-dev libapt-pkg-dev libarchive-dev python3-pip squashfs-tools xdelta3 && python3 -m pip install -r requirements-devel.txt -r requirements.txt && python3 -m pip uninstall -y coverage && apt install -y python3-coverage"
+  - DOCKER=1 TEST_SUITE=unit DEPENDENCIES="apt install -y git bzr subversion mercurial libnacl-dev libsodium-dev libffi-dev libapt-pkg-dev libarchive-dev python3-pip squashfs-tools xdelta3 && python3 -m pip install -r requirements-devel.txt -r requirements.txt && python3 -m pip uninstall -y coverage && apt install -y python3-coverage"
```
elopio: Remember that the unit tests are also executed when buildling the deb, so these should be added to debian/control.
Also, this PR shouldn't be landed before the autopkgtests for xenial:amd64 pass.

integration_tests/_source_helpers.py
```diff
@@ -0,0 +1,122 @@
...
+        os.chdir(cwd)
+
+
+def call(cmd):
+    subprocess.check_call(cmd, stdout=subprocess.DEVNULL,
+                          stderr=subprocess.DEVNULL)
+
+
+def call_with_output(cmd):
+    return subprocess.check_output(cmd).decode('utf-8').strip()
```
elopio: call and call with output are not specific to sources. Maybe put them in the base test case?

elopio: nevermind, I see you need them out side of the test case. Still, they should be probably called subprocess utils or something like that.

integration_tests/snaps/asset-tracking/snapcraft.yaml
```diff
@@ -10,3 +10,39 @@ parts:
...
+    source-type: subversion
+  hg-part:
+    plugin: nil
+    source: hg-source
+    source-type: mercurial
+  hg-part-tag:
+    plugin: nil
+    source: hg-source
+    source-type: mercurial
+    source-tag: feature-tag
```
elopio: This is a big snap. Which is not so bad, but what would you think about making one for each source? asset-tracking-with-git, asset-tracking-with-hg, asset-tracking-with-bzr...

integration_tests/test_pull_properties.py
sergiusens: Something where you can do without an `if/else`, your expected values are of one entry point only, so if your data structure were to be something like

```python
'expected': ('key', 'value'),

...

self.assertThat(state.assets['source-details'][self.expected[0], Equals(self.expected[1]))
```

NamedTuples might be better here. If you are going to keep that `plain` with an empy expected the most you need to do is add an `if self.expected` (even rename to `self.expected_details`).

I bet you can add `_source_helpers.create_git_repo('git-source')` into the `expected of `plain` unless we agree it needs checking for all cases.

integration_tests/test_pull_properties.py
sergiusens: ok then

sergiusens: a comment in the test would be good. With the previous comment then we'd do something like

```python
if self.expected_details:
   self.assertThat(state.assets['source-details'][self.expected_details.key, Equals(self.expected_details.value))
else:
    # check the source-commit
    ...

```

`key` and `value` come from using named tuples.

.gitmodules
sergiusens: seems like accidental removal.

integration_tests/test_pull_properties.py
```diff
@@ -68,7 +70,7 @@ def test_pull(self):
         self.assertTrue(len(state.assets['stage-packages']) > 0)
         self.assertTrue(len(state.assets['build-packages']) > 0)
         self.assertIn('hello=2.10-1', state.assets['stage-packages'])
-        self.assertIn('hello=2.10-1', state.assets['build-packages'])
```
elopio: so, we don't save the hello build-pacakge any more?

.gitmodules
josepht: yeah, I'll add it back.

integration_tests/snaps/asset-tracking/snapcraft.yaml
```diff
@@ -10,3 +10,39 @@ parts:
...
     build-packages: ['hello=2.10-1']
+  git-part:
+    plugin: nil
+    source: git-source
+    source-type: git
+  git-part-branch:
+    plugin: nil
+    source: git-source
+    source-type: git
+    source-branch: feature
```
elopio: I personally would find this more clear if the branch was called test-branch, or something like that.
I get confused making the link feature<->branch.

integration_tests/snaps/asset-tracking/snapcraft.yaml
josepht: I'm fine with that.

integration_tests/test_pull_properties.py
josepht: another merge mistake, I'll re-add it.

snapcraft/internal/sources/_bazaar.py
```diff
@@ -55,4 +61,25 @@ def pull(self):
...
+
+        branch = None
+        source = self.source
+
+        return {
+            'commit': commit,
+            'branch': branch,
+            'source': source,
+            'tag': tag,
+        }
```
elopio: I would prefer an object called SourceDetails.

snapcraft/internal/states/_pull_state.py
```diff
@@ -44,14 +44,16 @@ class PullState(State):
...
+                 stage_packages=None, build_packages=None,
+                 source_details=None):
         # Save this off before calling super() since we'll need it
         # FIXME: for 3.x the name `schema_properties` is leaking
         #        implementation details from a higher layer.
         self.schema_properties = property_names
         self.assets = {
             'stage-packages': stage_packages,
             'build-packages': build_packages,
+            'source-details': source_details,
```
elopio: ahhh, I see why a dict is is handy instead of an object. I would still prefer an object, and also one for assets. But no need to do it here if you prefer dicts. We can discuss later.

snapcraft/tests/sources/test_bazaar.py
```diff
@@ -14,12 +14,47 @@
...
+        """Return command output converted to a string."""
+        return subprocess.check_output(cmd).decode('utf-8').strip()
 
-from snapcraft.tests.sources import SourceTestCase
+    def rm_dir(self, dir):
+        if os.path.exists(dir):
+            shutil.rmtree(dir)
 
+    def clean_dir(self, dir):
+        self.rm_dir(dir)
```
elopio: why would this dir exists? aren't we running all tests in a clean dir?
mkdir and not taking care about any clean up should work. If not, we need to add the temp dir fixture to the base test case.

snapcraft/tests/sources/test_git.py
```diff
@@ -187,13 +195,17 @@ def test_source_checksum_raises_exception(self):
...
+class GitBaseTestCase(tests.TestCase):
 
     def call(self, cmd):
+        """Call a command ignoring output."""
         subprocess.check_call(
             cmd, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
 
+    def call_with_output(self, cmd):
+        """Return command output converted to a string."""
+        return subprocess.check_output(cmd).decode('utf-8').strip()
```
elopio: yeah, this is everywhere. It should be in a common file.

snapcraft/internal/sources/_bazaar.py
sergiusens: I don't mind this :wink:

snapcraft/internal/sources/_bazaar.py
```diff
@@ -41,6 +42,11 @@ def __init__(self, source, source_dir, source_tag=None, source_commit=None,
             raise errors.IncompatibleOptionsError(
                 "can't specify a source-checksum for a bzr source")
 
+        self.kwargs = {}
```
sergiusens: `self._call_kwargs` even

sergiusens:
Updating the branch which has the working node version (previous one was pulled out)

snapcraft/tests/sources/test_bazaar.py
josepht: I was working around the mocks in tests.sources.SourceTestCase.  Fix coming.

snapcraft/tests/sources/test_git.py
josepht: Fix coming to use the same one everywhere.

snapcraft/internal/states/_pull_state.py
josepht: I'm going to leave this as is for now.  The upcoming PR to add the source details to the snap's snapcraft.yaml takes advantage of dict.update().  We can discuss later.

snapcraft/internal/sources/_bazaar.py
josepht: this is staying a dict for now.  See following comment.

snapcraft/internal/sources/_bazaar.py
josepht: updated to _call_kwargs in upcoming push

integration_tests/snaps/asset-tracking/snapcraft.yaml
josepht: I'll update this to `test-branch`

integration_tests/_source_helpers.py
josepht: I'm adding snapcraft.tests.subprocess_utils.py in my upcoming push

.travis.yml
josepht: added in upcoming push

integration_tests/_source_helpers.py
josepht: upcoming push will have them as fixtures.  Thanks for the suggestion.

integration_tests/snaps/asset-tracking/snapcraft.yaml
josepht: update pushed

josepht:
I've pushed an update that should address the issues brought up in the last round of reviews.  Thanks for the reviews.

snapcraft/internal/sources/_subversion.py
```diff
@@ -53,14 +59,37 @@ def pull(self):
...
+        tag = None
+        source = self.source
+        commit = self.source_commit
+
+        lines = subprocess.check_output(['svn', 'info', self.source_dir]
+                                        ).decode('utf-8').split('\n')
+        prefix = 'Last Changed Rev: '
+        for line in lines:
+            if line.startswith(prefix):
+                commit = line.replace(prefix, '')
```
kyrofa: This variable may have already been set, in which case you're throwing it away here. In the other sources you check to make sure that isn't the case. Is there a reason that's not happening here?

josepht: Thanks for catching that.  Updated.

josepht:
Please retest this.

snapcraft/internal/states/_pull_state.py
sergiusens: For now is not an option, it is forever as we cannot break state afterwards.

josepht: That's a good point.  I'm for keeping it a dict forever then.

josepht:
xenial-amd64 test failure seems unrelated to this PR.

PR #1230, opened 2017-04-04
lxd: refactor Cleanbuilder into Containerbuild and add Project.

kalikiana:
I refactored the Cleanbuilder class into a new base class, Containerbuild and introduced a new class, Project, which creates and re-uses a persistent container named after the project (snap). The "snap" as well as the argumentless command now prompts the user for a containerized build (which in the future will mount folders from the host into the container instead of building off a tarball).


snapcraft/internal/lxd.py
```diff
@@ -34,19 +36,36 @@
...
+            return None
+        return tarinfo
+    return _tar_filter
 
-    def __init__(self, snap_output, tar_filename, project_options,
-                 remote=None):
+
+class Containerbuild:
+
+    def __init__(self, snap_output, source_folder, project_options,
```
sergiusens: Can you add a `*` after `self`? So it looks like
```python
def __init__(self, *, snap_output, source_folder, project_options,
```

snapcraft/internal/lxd.py
```diff
@@ -34,19 +36,36 @@
...
+                 config, container_name, remote=None):
```
sergiusens: Can we pass in a smaller massaged data set instead of the full of the config data structure? A smaller thing maybe called `snap_meta` with just the meta data, could be a method of `config`, so callers would pass `config.get_metadata()`

snapcraft/internal/lxd.py
```diff
@@ -120,6 +142,36 @@ def _wait_for_network(self):
...
+
+
+class Project(Containerbuild):
+
+    def __init__(self, snap_output, source_folder, project_options,
+                 config, remote=None):
+        super().__init__(snap_output, source_folder, project_options,
+                         config, config.data['name'], remote)
+
+    def _create_container(self):
```
sergiusens: it would be better here to do maybe a `_ensure_container(container_name)` (akin to `_verify_remote`) and then move to `_ensure_started(container_name)`

snapcraft/tests/commands/test_snap.py
```diff
@@ -29,6 +31,100 @@
...
+type: {}
+confinement: strict
+grade: stable
+
+parts:
+    part1:
+      plugin: nil
+"""
+
+    def setUp(self):
```
sergiusens: You should create a common `ContainerTestClass` with this common `setUp` (and `make_snapcraft_yaml`) that both Cleanbuild and this class can inherit from. If the only thing that really makes these different is the arguments to `main` then please consider using `testscenarios`.

snapcraft/internal/lxd.py
kalikiana: That would result in "__init__() takes 1 positional argument but 7 were given" because all arguments don't have names. What's your intention here?

sergiusens:
On Apr 5, 2017 5:56 AM, "Christian Dywan" <notifications@github.com> wrote:

*@kalikiana* commented on this pull request.
------------------------------

In snapcraft/internal/lxd.py
<https://github.com/snapcore/snapcraft/pull/1230#discussion_r109862372>:

>
-    def __init__(self, snap_output, tar_filename, project_options,
-                 remote=None):
+
+class Containerbuild:
+
+    def __init__(self, snap_output, source_folder, project_options,

That would result in "*init*() takes 1 positional argument but 7 were
given" because all arguments don't have names. What's your intention here?


To force keyword arguments

sergiusens:
@kalikiana christian@twotoasts.de has not signed the CLA, can either sign it with that address or `git --amend --author ...` using your @canonical.com address?

snapcraft/internal/lifecycle.py
```diff
@@ -312,30 +311,11 @@ def _handle_dirty(self, part, step, dirty_report):
...
-            return None
-        elif fn.endswith('.snap'):
-            return None
-        return tarinfo
-    return _tar_filter
-
-
 def cleanbuild(project_options, remote=''):
     config = snapcraft.internal.load_config(project_options)
-    tar_filename = '{}_{}_source.tar.bz2'.format(
```
sergiusens: moving this logic had me at awe at first and now I know why, we only want to tar up for a cleanbuild, for a project build we want the container to have the mount as we discussed, so the code path for project would do something like

    lxc config device add $container_name homedir disk source=$SNAPCRAFT_PROJECT_DIRECTORY path=/$project_name
    lxc start unmapped

snapcraft/internal/lxd.py
```diff
@@ -120,6 +146,39 @@ def _wait_for_network(self):
         logger.info('Network connection established')
 
 
+class Cleanbuilder(Containerbuild):
+
+    def __init__(self, *, output, source, project_options,
+                 metadata=None, remote=None):
+        container_name = petname.Generate(3, '-')
```
kyrofa: This gets `snapcraft-` prepended to it, which makes for quite a long container name. Would it be worth shortening this to 2?

elopio:
I'm copying here my comment about the unit test failure:

/home/ubuntu/workspace/canonical/snapcraft/snapcraft/internal/project_loader.py(412)load_config()
-> return Config(project_options)
(Pdb) n
snapcraft.internal.errors.SnapcraftSchemaError: The 'type' property does not match the required schema: OrderedDict() is not of type 'string'
you are hitting a sys.exit(1) because of that, which kills the test. We shouldn't exit there, we should just exit on main.
@kalikiana and I found this with pdb. You have to call it with stdout, because we are redirecting it
pdb.Pdb(stdout=sys.__stdout__).set_trace()
I have lots of things to document about running tests, sorry.

snapcraft/internal/lxd.py
kalikiana: No strong preference personally, but that's identical to behavior on master.

snapcraft/internal/lifecycle.py
kalikiana: Right, that's the long-term plan. I didn't think we wanted to get everything done in the same PR, but happy to go forward with that here.

sergiusens: I just don't want the tar logic to move into `lxd`. Either duplicate code (which should be fine as it is temprorary) or implement the mount stuff here (it is a few lines after all.

snapcraft/main.py
```diff
@@ -308,7 +308,13 @@ def run(args, project_options):  # noqa
         parts.define(args['<part-name>'])
     elif args['search']:
         parts.search(' '.join(args['<query>']))
+    elif os.environ.get('SNAPCRAFT_CONTAINER_BUILDS'):
```
sergiusens: Just rediscovered `os.getenv` too, not sure which one is preferred. But I write here due to another reason, is the next PR going to consume this env globally and make all the commands pass into the ProjectContainer class so that I can do

`SNAPCRAFT_CONTAINER_BUILDS=1 snapcraft pull`
`SNAPCRAFT_CONTAINER_BUILDS=1 snapcraft build part-n`
`SNAPCRAFT_CONTAINER_BUILDS=1 snapcraft clean`

We probably need to guard (or ignore the variable) when using `cleanbuild`

snapcraft/tests/commands/test_snap.py
```diff
@@ -31,6 +32,75 @@
...
+            'Mounting {} into container\n'
+            'Waiting for a network connection...\n'
+            'Network connection established\n'.format(source),
+            fake_logger.output)
+
+        container_name = '{}:snapcraft-snap-test'.format(self.remote)
+        project_folder = 'build_snap-test'
+        fake_lxd.check_call_mock.assert_has_calls([
+            call(['lxc', 'start', container_name]),
+            call(['lxc', 'config', 'device', 'add', container_name,
```
sergiusens: We might want to check before `set`, so maybe extract this into an `_ensure_disk` (or `_ensure_$better_name`)

sergiusens:
For the second run (in contrast to the first), I get:
`error: device already exists`

I am not sure if we need to run `apt-get update` on every run either (but that we can figure out in a different PR) as we probably want to introduce some new command to handle this.

snapcraft/main.py
kalikiana: Right now only "snapcraft"/ "snapcraft snap" honors the flag. There's a separate task "pass through snapcraft commands into the container" in the backlog which is about that (unless I misunderstood). So I was assuming once this branch is done, we would discuss and integrate the other commands.

snapcraft/internal/lxd.py
```diff
@@ -152,10 +152,9 @@ def __init__(self, *, output, source, project_options,
                          remote=remote)
 
     def _ensure_container(self):
-        try:
-            check_call([
-                'lxc', 'start', self._container_name])
-        except CalledProcessError:
+        containers = check_output([
+            'lxc', 'list', self._container_name])
```
sergiusens: smart, I did not think this would act as a filter, we just need to be careful in the `in` check below
```
$ lxc list snapcraft
+----------------+---------+----------------------+-----------------------------------------------+------------+-----------+
|      NAME      |  STATE  |         IPV4         |                     IPV6                      |    TYPE    | SNAPSHOTS |
+----------------+---------+----------------------+-----------------------------------------------+------------+-----------+
| snapcraft      | RUNNING | 10.120.128.94 (eth0) | fd42:40b3:c2e2:92a2:216:3eff:fe46:a89f (eth0) | PERSISTENT | 0         |
+----------------+---------+----------------------+-----------------------------------------------+------------+-----------+
| snapcraft-godd | STOPPED |                      |                                               | PERSISTENT | 0         |
+----------------+---------+----------------------+-----------------------------------------------+------------+-----------+
```

snapcraft/tests/test_lxd.py
```diff
@@ -32,6 +32,11 @@
 
 class LXDTestCase(tests.TestCase):
 
+    scenarios = [
+        ('local', dict(remote='local')),
+        ('remote', dict(remote='my-remote')),
+    ]
```
sergiusens: \o/

snapcraft/internal/lxd.py
sergiusens:
```
lxc list --format=json snapcraft
```

snapcraft/main.py
sergiusens: wfm

PR #1231, opened 2017-04-04
pluginhandler: exclude `/snap/` from libraries.

sergiusens:
When collecting libraries when working with classic confinement
we currently collect snaps of our own doing if installed. This
small fix avoids that.

Ideally it would have been good to just add the `no-system-libs`
`build-attributes` but the side effects there are a bit too big
as some libraries are still collected when dealing with snaps
that depend on other snaps.

LP: #1670100


sergiusens:
done @elopio

sergiusens:
Apparently v6.10.1 doesn't exist anymore and we need v6.10.2 :cry:

PR #1233, opened 2017-04-05
docs: update contribution guide.

sergiusens:
Reflect the fact that we no longer require bugs.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,10 +9,10 @@ here.
 give us permission to use your contributions.
 
 1. If there is a [Snapcraft bug][2] you are trying to fix, please refer to
-   it here. If it is a feature that has not been discussed before, please
-   raise awareness on https://forum.snapcraft.io under the *snapcraft* topic
-   so we are on the same track and you don't work in vain or step over what
-   someone else is already working on, doing so actually saves time!
+   it here. If it is a feature that has not been discussed, please raise
+   awareness on https://forum.snapcraft.io under the *snapcraft* topic. This
+   will ensure that we're all on the same page, and your work is not in vain
+   or duplicating what someone else is already doing. This actually saves time!
 
 2. We use a forking, feature-based workflow.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -8,12 +8,11 @@ here.
...
-   feature you're about to add. This is not just paperwork required in order to
-   land something: this is where you state what's happening, and why it's a
-   shortcoming. Doing this before you start working on a fix also gives the
-   Snapcraft team a chance to give you feedback and advice, which saves time
-   for everyone!
+1. If there is a [Snapcraft bug][2] you are trying to fix, please refer to
+   it here. If it is a feature that has not been discussed before, please
+   raise awareness on https://forum.snapcraft.io under the *snapcraft* topic
+   so we are on the same track and you don't work in vain or step over what
+   someone else is already working on, doing so actually saves time!
```
kyrofa: This is a pretty long sentence. Consider breaking it up, perhaps something like:

"If it is a feature that has not been discussed, please raise awareness on https://forum.snapcraft.io under the *snapcraft* topic. This will ensure that we're all on the same page, and your work is not in vain or duplicating what someone else is already doing. This actually saves time!"

CONTRIBUTING.md
```diff
@@ -38,10 +37,7 @@ give us permission to use your contributions.
 
        More detailed explanatory text, if necessary. Wrap it to 72 characters.
        Think of this like an email, where you have a subject line and a body.
-       Make sure you mention the bug being fixed on a line all by itself at the
-       end, like so:
 
-       LP: #<bug number>
```
kyrofa: Do we not want still want the buglink in the commit if there is one?

sergiusens: only if there is one

sergiusens: but I don't think of it as mandatory either anymore as I don't want to not trust people about not including them and triple checking

elopio:
+1 with @kyrofa's reworded paragraph.

The bugs were really useful for me because they usually had the instructions to run a regression test, and that's what I do during releases. Please, be careful when reviewing branches to make sure that there is a way to verify manually that the change works as expected.

PR #1234, opened 2017-04-05
core: find the correct libraries as a snap.

sergiusens:
The current implementation as of 3.6 in for find_library in ctypes.util makes use of `ldconfig -p` which depends on the current ld cache which in effect has no knowledge of any library in $SNAP. What makes matters worse is that it will provide results for on host libraries we do NOT want.

LP: #1678289

### DIFF
--- a/snapcraft/__init__.py
+++ b/snapcraft/__init__.py
@@ -356,6 +356,28 @@ from collections import OrderedDict                 # noqa
 import pkg_resources                                # noqa
 import yaml                                         # noqa
 
+import os as _os
+if _os.environ.get('SNAP_NAME') == 'snapcraft':
+    # The current implementation as of 3.6 in for find_library in
+    # ctypes.util makes use of `ldconfig -p` which depends on the
+    # current ld cache which in effect has no knowledge of any library
+    # in $SNAP. What makes matters worse is that it will provide
+    # results for on host libraries we do NOT want.
+    import re as _re
+
+    def find_library(name):
+        regex = r'lib{name}\.[^\s]+'.format(name=_re.escape(name))
+        snap_root = _os.environ.get('SNAP')
+        for root, directories, files in _os.walk(snap_root):
+            for filename in files:
+                res = _re.search(regex, filename)
+                if res:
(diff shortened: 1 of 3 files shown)


elopio:
With your explanation, it seems that the code makes sense. But I don't really know about this, please write a test I can run to verify that it breaks without your fix, and passes with it.

snapcraft/__init__.py
```diff
@@ -356,6 +356,28 @@
...
+        snap_root = _os.environ.get('SNAP')
+        for root, directories, files in _os.walk(snap_root):
+            for filename in files:
+                res = _re.search(regex, filename)
+                if res:
+                    return res.group(0)
+
+    from unittest.mock import patch as _patch
+    _p = _patch('ctypes.util.find_library').start()
+    _p.side_effect = find_library
```
elopio: why mock? The idea of mock is that you get it reverted when the test ends. In this case, it would be "better" just to overwrite find_library.

sergiusens: how would I do that?

sergiusens:
@elopio it is not a build problem, it is a runtime problem and the test beds are super contaminated to use regular ctypes. I can give you manual, build the snapcraft snap as is in the repo, create a clean container and run it.

snapcraft/__init__.py
elopio: ctypes.util.find_library = find_library

elopio:
Did you tried it @sergiusens?

I built the snap using your branch and tried to install it in a xenial lxc container, but got this error: https://paste.ubuntu.com/24318092/

sergiusens:
On Apr 5, 2017 12:59 AM, "Leo Arias" <notifications@github.com> wrote:

Did you tried it @sergiusens <https://github.com/sergiusens>?

I built the snap using your branch and tried to install it in a xenial lxc
container, but got this error: https://paste.ubuntu.com/24318092/


Of course I tried. The snapcraft in edge has this.

sergiusens:
@elopio please try again

PR #1236, opened 2017-04-05
snap: use the gpg tarball instead of git://.

sergiusens:
Since launchpad doesn't support git:// references in code and
the https handler for gnupg is currently broken it seems
appropriate to use the release tarball.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -108,9 +108,7 @@ parts:
         - zlib1g-dev
     after: [python]
   gpg:
-      source: git://git.gnupg.org/gnupg.git
-      source-tag: gnupg-1.4.20
-      source-depth: 1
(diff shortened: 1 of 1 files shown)


snap/snapcraft.yaml
```diff
@@ -108,9 +108,7 @@ parts:
         - zlib1g-dev
     after: [python]
   gpg:
-      source: git://git.gnupg.org/gnupg.git
-      source-tag: gnupg-1.4.20
-      source-depth: 1
+      source: https://gnupg.org/ftp/gcrypt/gnupg/gnupg-1.4.21.tar.bz2
```
elopio: what about adding a checksum?

sergiusens: I think we need signature verification for this one and checksum is not that important (upstream doesn't publish).

PR #1237, opened 2017-04-06
repo: enable elementary.

sergiusens:
Added `elementary` and `"elementary"`. The former for when the
latter is fixed if at all.

LP: #1678286


popey:
Tested on Elementary, appears to work fine. No more crashes, and I can build snaps now.

PR #1239, opened 2017-04-06
catkin plugin: create completely valid environment

kyrofa:
The environment created by the Catkin plugin is really only completely valid at runtime (i.e. when packaged as a snap). It's partially valid at build-time. However, it turns out that as a result of sourcing `setup.sh` (which is required at build-time), various dependencies actually need the valid run-time environment (e.g. `LD_LIBRARY_PATH`, valid `ROS_HOME`, etc.).

This PR fixes LP: [#1674487](https://bugs.launchpad.net/snapcraft/+bug/1674487) by making sure `LD_LIBRARY_PATH` is defined before we source `setup.sh`. It also makes sure `ROS_HOME` is always pointing to a valid place.

### DIFF
--- a/snapcraft/plugins/catkin.py
+++ b/snapcraft/plugins/catkin.py
@@ -191,6 +191,10 @@ deb http://${{security}}.ubuntu.com/${{suffix}} {0}-security main universe
     def env(self, root):
         """Runtime environment for ROS binaries and services."""
 
+        paths = common.get_library_paths(root, self.project.arch_triplet)
+        ld_library_path = formatting_utils.combine_paths(
+            paths, prepend='', separator=':')
+
         env = [
             # This environment variable tells ROS nodes where to find ROS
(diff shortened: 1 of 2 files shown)


kyrofa:
> Is the turtlebot needed to verify the fix in real life?

No, it can be verified by creating a snap of some Turtlebot-focused software. I thought about making it a snaps test, but if you think the talker/listener ROS snap takes a while, you ain't seen nothin.

elopio:
@kyrofa I'd still be interested in some steps I can run here, even if they take a lot. Maybe you can put them in a gist for testing robotics, and we keep adding things to check once in a while?

kyrofa:
@elopio yeah that's a good idea. [Here you go](https://gist.github.com/kyrofa/c01366d9ef68872efd0d01ee5aa047fe).

sergiusens:
@kyrofa good candidate for the nightlies

PR #1240, opened 2017-04-06
cli: improve push output.

sergiusens:
Use basename of file to leave space for the widget (useful when pushing a delta).
Don't print the 'Upload' message twice and don't leave a trailing spinner after the release is complete.
Upload renamed to push to be more consistent.

LP: #1625784

### DIFF
--- a/snapcraft/storeapi/__init__.py
+++ b/snapcraft/storeapi/__init__.py
@@ -762,16 +762,19 @@ class StatusTracker:
 
         content = {}
         for indicator_count in itertools.count():
+            progress_indicator.update(indicator_count)
             if not queue.empty():
                 content = queue.get()
                 if isinstance(content, Exception):
                     raise content
-                widgets[0] = self._get_message(content)
-            progress_indicator.update(indicator_count)
             if content.get('processed'):
                 break
+            else:
+                widgets[0] = self._get_message(content)
             sleep(0.1)
         progress_indicator.finish()
+        # Print at the end to avoid a left over spinner artifact
+        print(self._get_message(content))
 
         self.__content = content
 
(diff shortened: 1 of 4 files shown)


snapcraft/storeapi/__init__.py
```diff
@@ -762,16 +762,19 @@ def track(self):
...
-                widgets[0] = self._get_message(content)
-            progress_indicator.update(indicator_count)
             if content.get('processed'):
                 break
+            else:
+                widgets[0] = self._get_message(content)
             sleep(0.1)
         progress_indicator.finish()
+        # Print at the end to avoid a left over spinner artifact
+        print(self._get_message(content))
```
elopio: I cry every time I see a print :'(
Let's make a sprint in Costa Rica to fix the UI and the logging!

snapcraft/tests/commands/test_push.py
```diff
@@ -89,7 +89,7 @@ def test_push_a_snap(self):
 
         self.assertRegexpMatches(
             self.fake_logger.output,
-            ".*Pushing 'my-snap-name_0\.1_\w*.snap' to the store\.\n"
+            ".*push 'my-snap-name_0\.1_\w*.snap' to the store\.\n"
```
elopio: why you are not checking the "Preparing to " part of thte message?

snapcraft/storeapi/__init__.py
sergiusens: ok, but this should be a print, not a logger message.

snapcraft/tests/commands/test_push.py
sergiusens: I don't think these are as relevant.

elopio: I don't think either, but now that you are changing it, it seemed low effort.

PR #1241, opened 2017-04-06
help: replace dashes with underscores when printing plugins help

EduardoVega:
Plugins named with dashes must be modified before importing them to get the docs, dashes must be replaced with underscores, so we can find the right plugin name.

LP: [#1604312](https://bugs.launchpad.net/snapcraft/+bug/1604312)


snapcraft/_help.py
```diff
@@ -46,7 +46,7 @@ def _topic_help(module_name, devel):
 def _module_help(module_name, devel):
     try:
         module = importlib.import_module(
-            'snapcraft.plugins.{}'.format(module_name))
+            'snapcraft.plugins.{}'.format(module_name.replace("-", "_")))
```
elopio: please use single quotes instead of double quotes:

    '-', '_'

It's a small detail, but consistency in the details is what makes python pretty :)

snapcraft/tests/commands/test_help.py
```diff
@@ -69,6 +69,15 @@ def test_print_module_help_for_valid_plugin(self):
...
                          '{!r} instead'.format(expected, output))
 
+    def test_print_module_with_dashes_help_for_valid_plugin(self):
+        main(['help', 'plainbox-provider'])
+
+        expected = ' Create parts containing'
+        output = self.fake_terminal.getvalue()[:len(expected)]
+        self.assertEqual(output, expected,
+                         'The help message does not start with {!r} but with '
+                         '{!r} instead'.format(expected, output))
```
elopio: There's a testtools matcher that you can use here to make it simpler:

https://testtools.readthedocs.io/en/latest/for-test-authors.html#startswith-endswith

EduardoVega: Thanks for the feedback, working on it. I have to suggestions that could be worked in the future:

I think testtools startswith-endswith does make the test easier to understand, so it would be nice to also add it to the rest of tests doing similar asserts in test_help.py.

Also I realized that plugins help message is standard (e.g The 'plugin name' plugin is...), but the one that I'm using for this test 'plainbox-provider' isn't, so we should modify it so it prints same message format.

elopio: yes, good catches. Please report bugs for them, and obviously, feel free to fix them if you have more time to contribute with us..

snapcraft/tests/commands/test_help.py
```diff
@@ -69,6 +69,15 @@ def test_print_module_help_for_valid_plugin(self):
...
                          '{!r} instead'.format(expected, output))
 
+    def test_print_module_named_with_dashes_help_for_valid_plugin(self):
+        main(['help', 'plainbox-provider'])
+
+        expected = ' Create parts'
+        output = self.fake_terminal.getvalue()
+        self.assertThat(output, tests.testtools.matchers.StartsWith(expected),
+                         'The help message does not start with {!r} but with '
+                         '{!r} instead'.format(expected, output[:len(expected)]))
```
elopio: Great! 
Just one more minor detail. testtools takes care of the formatting of the error message, so you can remove there the third argument. Give it a try forcing the test to fail with the wrong expected text.

EduardoVega: Cool, didn't know that... Thanks

PR #1242, opened 2017-04-07
kernel_plugin: use CROSS_COMPILE to override the default toolchain

piso77:

LP: #1680863


sergiusens:
Do you mind adding an entry in manual-tests.md on how to test this?

piso77:
A bootable kernel is a product of the kernel src itself and the toolchain used to build it, and if we let the user choose any random toolchian, we have no idea if the resulting kernel will still boot.

Since this feature is about letting the developer choose a different toolchain than the one we hard code in options.py, we should only check for this precise case - that the developer was able to override the default toolchain and pick a new one using the CROSS_COMPILE env variable - considering this, i added a unit tests for precisely this case, and removed the manual test that i previously pushed in manual-tests.md since the unit test already cover this, and it's automatic/more efficient.

elopio:
I'm happy with your solution for the tests. Thanks @piso77

sergiusens:
I made a fix to the way the environment was being set. After this passes we are good.

PR #1243, opened 2017-04-07
tests: update the location of the upstream retry autopkgtests script

elopio:
LP: [#1680960](https://bugs.launchpad.net/snapcraft/+bug/1680960)


sergiusens:
habit is hard to change, no need for a bug :wink:

PR #1246, opened 2017-04-10
store: plumbing for collaboration support.

psivaa:
Carry out the plumbing work for developer collaboration feature (snap_developer) for the UI to be wired in later.

GET/PUT snap-developer assertions
Refactor 'get|put_validate' helpers to generalise the use.
Sign snap-developer assertion using snap sign
LP: #1680123


sergiusens:
Tried your branch as a passthrough and got a store error it seems:
```
sergiusens@mirkwood:~/source/snapcraft$  snapcraft staging servers> python3 -m snapcraft.main collaborate awesome-databse --key-name staging
{'snap_developer': []}
Signing developers assertion for dt96PQCUvRUcu9GYh90xcXGC4kCldlVn
Error signing developers assertion for dt96PQCUvRUcu9GYh90xcXGC4kCldlVn: error: cannot parse the assertion input as JSON: json: cannot unmarshal array into Go value of type map[string]interface {}
```

snapcraft/_store.py
```diff
@@ -793,19 +793,76 @@ def validate(snap_name, validations, revoke=False, key=None):
...
+    try:
+        return store.get_assertion(snap_id, 'developers')
+    except storeapi.errors.StoreValidationError as e:
+        if e.error_list[0]['code'] == 'snap-developer-not-found':
+            return {'snap_developer': {
+                'type': 'snap-developer',
+                'authority-id': publisher_id,
+                'publisher-id': publisher_id,
+                'snap-id': snap_id,
+                'developers': []}}
```
emgee: There may be no `developers` item in the assertion returned from the store so I would leave it out here too. It will force the calling code to handle it properly.

snapcraft/_store.py
```diff
@@ -793,19 +793,76 @@ def validate(snap_name, validations, revoke=False, key=None):
...
+        raise
+
+
+def _sign_developers(snap_id, assertion, key):
```
emgee: Is this func worth it? The calling code already has a `StoreClient` instance anyway so this is really adding extra lines of (real) code and conflating signing and storing.

snapcraft/_store.py
```diff
@@ -793,19 +793,76 @@ def validate(snap_name, validations, revoke=False, key=None):
...
+    #          'since': '2017-02-10T08:35:00.390258Z',
+    #          'until': '2018-02-10T08:35:00.390258Z'
+    #         }],
+    #      }
+    # }
+
+    _sign_developers(snap_id, developers['snap_developer'], key)
+
+
+def _get_developers(snap_id, publisher_id):
```
emgee: `developers = _get_developers(snap_id, publisher_id)` ... `developers['snap_developer']` above is a bit weird. Maybe this func should return the parsed assertion? I'm not sure how that fits with other snapcraft code though.

snapcraft/_store.py
```diff
@@ -793,19 +793,76 @@ def validate(snap_name, validations, revoke=False, key=None):
...
+    store = storeapi.StoreClient()
+    try:
+        return store.get_assertion(snap_id, 'developers')
+    except storeapi.errors.StoreValidationError as e:
+        if e.error_list[0]['code'] == 'snap-developer-not-found':
+            return {'snap_developer': {
+                'type': 'snap-developer',
+                'authority-id': publisher_id,
```
emgee: `authority-id` must be the publisher's id even if the assertion came from the store (see also above).

snapcraft/_store.py
```diff
@@ -793,19 +793,78 @@ def validate(snap_name, validations, revoke=False, key=None):
...
+
+    with _requires_login():
+        account_info = store.get_account_information()
+    publisher_id = account_info['account_id']
+
+    release = storeapi.constants.DEFAULT_SERIES
+    try:
+        snap_id = account_info['snaps'][release][snap_name]['snap-id']
+    except KeyError:
+        raise storeapi.errors.SnapNotFoundError(snap_name)
```
emgee: Is `snap_name` definitely the only lookup that could cause a KeyError? I suspect so, but thought I'd ask or the exception could be very confusing.

snapcraft/_store.py
```diff
@@ -793,19 +793,78 @@ def validate(snap_name, validations, revoke=False, key=None):
...
+    developers = _get_developers(snap_id, publisher_id)
+
+    # XXX: Do the amendments via UI here.
+    #
+    # The data will look like:
+    # {'snap_developer': {
+    #      'type: 'snap-developer',
+    #      'authority-id': <account_id of the publisher Or `canonical`>,
```
emgee: Sort of. By the time the UI gets involved the `authority-id` must be changed to the publisher's id as that's whos key is signing the new assertion.

snapcraft/_store.py
```diff
@@ -793,19 +793,78 @@ def validate(snap_name, validations, revoke=False, key=None):
...
+        if e.error_list[0]['code'] == 'snap-developer-not-found':
+            return {'snap_developer': {
+                'type': 'snap-developer',
+                'authority-id': publisher_id,
+                'publisher-id': publisher_id,
+                'snap-id': snap_id}}
+        raise
+
+
+def _sign_developers(snap_id, assertion, key):
```
emgee: Is this func worth it? The calling code already has a StoreClient instance anyway so this is really adding extra lines of (real) code. I also feel it conflates signing (local) and storing (remote).

snapcraft/_store.py
psivaa: Yes. Unless we mess up here on the release values. Even in that case, the error of no `snap_name` available is correct, but not specific enough. (Also the same segment is used in a few other places, which needs changing and that raises a point that this part of `find_snap_id` can be a helper. Not sure if it falls within this context of this PR. Could do a follow up if all agree)

snapcraft/_store.py
psivaa: Let me only have `publisher_id` here.

snapcraft/_store.py
psivaa: Good suggestion. Fixed.

snapcraft/_store.py
psivaa: Good catch again. Removed and merged them in `collaborate`

snapcraft/_store.py
emgee: Sorry, terrible place to leave this comment. I'll add an improved comment where it makes more sense and delete this.

snapcraft/_store.py
```diff
@@ -793,19 +793,74 @@ def validate(snap_name, validations, revoke=False, key=None):
...
+
+    assertion = _sign_assertion(snap_id, assertion, key, 'developers')
+
+    store.push_assertion(snap_id, assertion, 'developers')
+
+
+def _get_developers(snap_id, publisher_id):
+    store = storeapi.StoreClient()
+    try:
+        return store.get_assertion(snap_id, 'developers')['snap_developer']
```
emgee: When the assertion comes from the store, we don't know if the `authority-id` is "canonical" or the publisher's ID so it must be set to the publisher's ID. Otherwise the signing key does not belong to the signing authority and the assertion will be rejected when it's saved.

Personally, I would do this in the same place as `revision` is incremented but I'll leave that to you.

psivaa: Ok, got it. Fixed this too. Thanks for the comment.

snapcraft/_store.py
sergiusens: the calling code is this code base, righ? care to clarify?

emgee: @sergiusens yes, specifically the `collaborate` func atm.

snapcraft/_store.py
```diff
@@ -815,17 +873,20 @@ def _check_validations(validations):
...
+def _sign_assertion(target, assertion, key, endpoint):
     cmdline = ['snap', 'sign']
     if key:
         cmdline += ['-k', key]
     snap_sign = Popen(
         cmdline, stdin=subprocess.PIPE, stdout=subprocess.PIPE,
         stderr=subprocess.PIPE)
     data = json.dumps(assertion).encode('utf8')
-    logger.info('Signing validation {}'.format(validation))
+    logger.info('Signing {} assertion for {}'.format(endpoint, target))
```
sergiusens: It would be better here if `target` were the snap name instead (maybe with `target` in parenthesis indicating it is the `snap-id`). So far, no elements of the UI of snap nor snapcraft have exposed these IDs so my suggestion is to just hide them.

This is my last comment fwiw

psivaa: Good catch. That was different in validation vs snap developers. Fixed now. Thanks for the spot.

PR #1247, opened 2017-04-10
Fixing Store integration tests.

cprov:
Coping with recently added snap-name size limit (40-chars) in Store.

Pending more specific error response from the store. See LP: #1681545

### DIFF
--- a/integration_tests/__init__.py
+++ b/integration_tests/__init__.py
@@ -273,14 +273,30 @@ class StoreTestCase(TestCase):
         process.close()
         return process.exitstatus
 
+    def get_unique_name(self, prefix='snapcrafttest'):
+        """Return a unique snap name.
+
+        It uses a UUIDv4 to create unique names and limits its full size
+        to 40 chars (as defined in the snap specification).
+        """
+        unique_id = uuid.uuid4().int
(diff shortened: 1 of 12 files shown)


integration_tests/__init__.py
```diff
@@ -273,14 +273,30 @@ def list_registered(self, expected_snaps):
         process.close()
         return process.exitstatus
 
+    def get_unique_name(self, prefix='u1test'):
+        """Return an unique snap name.
```
sergiusens: a unique

cprov: Fixed in the next revision.

cprov:
@sergiusens great opportunity to drop 'u1test' prefix, thanks to pointing it. Using 'snapcrafttest' in the next rev.

snapcraft/tests/fake_servers.py
```diff
@@ -603,7 +607,7 @@ def _handle_register_409(self, error_code):
         self.wfile.write(json.dumps(response).encode())
 
     def _handle_register_429(self, error_code):
-        self.send_response(409)
+        self.send_response(429)
```
sergiusens: hah

PR #1249, opened 2017-04-11
Add Linux Mint support

nefelim:

### DIFF
--- a/snapcraft/internal/repo/_platform.py
+++ b/snapcraft/internal/repo/_platform.py
@@ -26,6 +26,7 @@ _DEB_BASED_PLATFORM = [
     '"elementary"',
     'debian',
     'neon',
+    'LinuxMint',
 ]
 
 
--- a/snapcraft/internal/repo/_platform.py
+++ b/snapcraft/internal/repo/_platform.py
@@ -26,7 +26,7 @@ _DEB_BASED_PLATFORM = [
     '"elementary"',
     'debian',
     'neon',
-    'LinuxMint',
+    'linuxmint',
 ]
 
 
@@ -35,7 +35,7 @@ def _is_deb_based(distro):
 
 
 def _get_repo_for_platform():
-    distro = _linux_distribution()[0]
+    distro = _linux_distribution()[0].lower()
     if _is_deb_based(distro):
         return Ubuntu
     else:


snapcraft/internal/repo/_platform.py
```diff
@@ -26,6 +26,7 @@
     '"elementary"',
     'debian',
     'neon',
+    'LinuxMint',
```
Conan-Kudo: This is wrong. It should be `linuxmint`. The value being checked is the `ID` field in `/etc/os-release`.

nefelim: Very strange...
`python3
Python 3.5.2 (default, Nov 17 2016, 17:05:23) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import platform
>>> platform.linux_distribution()[0]
'LinuxMint'
>>> 
`

nefelim: cat /etc/os-release
NAME="Linux Mint"
VERSION="18 (Sarah)"
ID=linuxmint
ID_LIKE=ubuntu
PRETTY_NAME="Linux Mint 18"
VERSION_ID="18"
HOME_URL="http://www.linuxmint.com/"
SUPPORT_URL="http://forums.linuxmint.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/linuxmint/"
UBUNTU_CODENAME=xenial

zyga: LinuxMint is spelled in different ways across releases AFAIR. Please look at https://github.com/zyga/os-release-zoo/tree/master/linuxmint

At one time it was indeed `LinuxMint` but I think they corrected that now.

ogra1: seems it only has that info in 18, seemingly 17 has the ubuntu version of the file
https://bugs.launchpad.net/linuxmint/+bug/1641491

nefelim: For my case code has been work. 
Test python script has return "LinuxMint".
File /etc/os-release has string "ID=linuxmint".
Who should I believe?

nefelim: 1. I has run script:
'(echo "import platform" ; echo "print (platform.linux_distribution()[0])") | strace python3'
2. View logs:
'
......
open("/etc/lsb-release", O_RDONLY|O_CLOEXEC) = 3
fstat(3, {st_mode=S_IFREG|0644, st_size=105, ...}) = 0
ioctl(3, TCGETS, 0x7fffcf805070)        = -1 ENOTTY (Inappropriate ioctl for device)
lseek(3, 0, SEEK_CUR)                   = 0
ioctl(3, TCGETS, 0x7fffcf805010)        = -1 ENOTTY (Inappropriate ioctl for device)
lseek(3, 0, SEEK_CUR)                   = 0
read(3, "DISTRIB_ID=LinuxMint\nDISTRIB_REL"..., 8192) = 105
read(3, "", 8192)                       = 0
close(3)                                = 0
write(1, "LinuxMint\n", 10LinuxMint
......
'
3. And dump file:
cat /etc/lsb-release
DISTRIB_ID=LinuxMint
DISTRIB_RELEASE=18
DISTRIB_CODENAME=sarah
DISTRIB_DESCRIPTION="Linux Mint 18 Sarah"

ogra1: smells like platform.linux_distribution() actually uses info from lsb-release (which seems wrong)

nefelim: Alternatively, you can add one more line 'linuxmint'.

niemeyer: Alternatively, lowercase the string before attempting to match.

nefelim: +1

ogra1: long term it would be good if snapcraft could switch to a function that uses /etc/os-release so we are all on the same page (i dont think snapd uses lsb stuff anywhere, since there is no guarantee it is even installed)

nefelim:
1. Change constant platform name to  lower case
2. Add code for transform platform name to lower case.

snapcraft/internal/repo/_platform.py
Conan-Kudo: It's not hard to write a replacement function to just process the `os-release(5)` file. I did it some time ago. I can submit a PR to replace the function call with direct parsing of the file.

snapcraft/internal/repo/_platform.py
```diff
@@ -34,7 +35,7 @@ def _is_deb_based(distro):
 
 
 def _get_repo_for_platform():
-    distro = _linux_distribution()[0]
+    distro = _linux_distribution()[0].lower()
```
sergiusens: Doing a `lower` here will fail all the other matches it needs to do. I would suggesst against doing a lower here unless everything is lowered in `_DEB_BASED_PLATFORM` but that makes the latter less strict so I am more inclined on adding all the matches you need.

sergiusens:
Closing this due to lack of activity, feel free to reopen once the comments have been addressed.

PR #1253, opened 2017-04-13
go plugin: cross-compilation support

kalikiana:
To be used with the --target-arch= argument.

Fixes [bug 1664857](https://bugs.launchpad.net/snapcraft/+bug/1664857).


sergiusens:
I've added someone as a reviewer who will give you some magic lines :wink:

chipaca:
At a glance, this is missing `CGO_ENABLED=1` (otherwise Go runs as if `CGO_ENABLED=0` were set, when cross-compiling, at least with ). Without cgo enabled simple things like current user lookup, and full nssswitch support, won't work. You also probably need a few more flags to get the right ld and etc; I recommend having a simple go program that uses a shared C library for testing.

If you want to suffer, make that a simple go program that uses a shared C++ library. Mix both of these if you want to lose your mind.

The best quick read for this that I know of is still [xnox's blog post](http://blog.surgut.co.uk/2014/06/cross-compile-go-code-including-cgo.html) from a few years ago.

Also note the intricacies of this differ between go versions, so test it on all of them.

chipaca:
As a start, try this:
```go
package main

/*
#cgo pkg-config: glib-2.0
#include <math.h>
#include <glib.h>

double mysin(double x) {
    return sinf(x);
}

const char* username() {
    return g_get_user_name();
}
*/
import "C"

import (
	"fmt"
	"os/user"
)

func sin(x float64) float64 {
	return float64(C.mysin(C.double(x)))
}

func main() {
	u, err := user.Current()
	if err != nil {
		panic(err)
	}
	fmt.Println("user according to go:", u.Username)
	fmt.Println("user according to glib:", C.GoString(C.username()))
	fmt.Println("more or less 1:", sin(3.14/2))
}
```

sergiusens:
you might need to commit --reauthor your commit with christian@twotoasts.de or sign the CLA ;-)

sergiusens:
@mwhudson mind taking a look at this?

mwhudson:
It all looks OK to me.

snapcraft/plugins/go.py
```diff
@@ -191,4 +197,20 @@ def _build_environment(self):
...
+            env['CGO_ENABLED'] = '1'
+            # See https://golang.org/doc/install/source#environment
+            go_archs = {
+                'armhf': 'arm',
+                'i386': '386',
+                'ppc64el': 'ppc64le',
+            }
+            env['GOARCH'] = go_archs.get(self.project.deb_arch,
+                                         self.project.deb_arch)
+            env['GOARM'] = '7'
```
sergiusens: do you want to set `GOARM` even if `GOARCH` is not `arm`?

mwhudson: It's harmless, if a little strange.

integration_tests/snaps/go-cgo/snapcraft.yaml
```diff
@@ -0,0 +1,18 @@
...
+version: 0.1
+summary: A go project using libraries written in C.
+description: |
+  This is a basic go snap that uses cgo to access C APIs.
+  The name of the binary will be that of the directory containing it,
+  just like it is seen when using `go build` or `go install`.
+  If you want to add other functionalities to this snap, please don't.
+  Make a new one.
+
+icon: icon.png
```
sergiusens: let's remove this (and the actual icon file)

integration_tests/__init__.py
```diff
@@ -31,6 +33,7 @@
 from testtools.matchers import MatchesRegex
 
 from snapcraft.tests import fixture_setup
+from snapcraft.internal.repo import _deb
```
sergiusens: this gives me a bad vibe, not sure how to solve it yet. Maybe reimplement here, @elopio ?

integration_tests/__init__.py
```diff
@@ -129,6 +132,32 @@ def run_snapcraft_parser(self, arguments):
             raise
         return snapcraft_output
 
+    @contextmanager
+    def setup_multi_arch_sources(self, arch):
```
elopio: This is nice, but in tests we usually have fixtures instead of context managers. I personally would prefer to follow that practice, but I don't have a very strong opinion against this context manager.

integration_tests/plugins/test_go_plugin.py
```diff
@@ -33,3 +34,11 @@ def test_stage_go_plugin(self):
...
     def test_building_multiple_main_packages(self):
         self.run_snapcraft('stage', 'go-with-multiple-main-packages')
+
+    def test_cross_compiling(self):
+        if snapcraft.ProjectOptions().deb_arch != 'amd64':
+            self.skipTest('The test only handles amd64 to armhf')
+
+        arch = 'armhf'
+        with self.setup_multi_arch_sources(arch):
+            self.run_snapcraft('stage', 'go-cgo', '--target={}'.format(arch))
```
elopio: looks good! Thanks :)

snapcraft/tests/plugins/test_go.py
```diff
@@ -24,6 +24,55 @@
...
+                'GOARCH' in env, 'Expected environment to include GOARCH')
+            self.assertEqual(env['GOARCH'], self.go_arch)
+
+    scenarios = [
+        ('armv7l', dict(deb_arch='armhf', go_arch='arm')),
+        ('aarch64', dict(deb_arch='arm64', go_arch='arm64')),
+        ('i386', dict(deb_arch='i386', go_arch='386')),
+        ('x86_64', dict(deb_arch='amd64', go_arch='amd64')),
+        ('ppc64le', dict(deb_arch='ppc64el', go_arch='ppc64le')),
+    ]
```
elopio: Please move the scenarios before the test.

snapcraft/internal/parts.py
```diff
@@ -273,7 +273,13 @@ def load_plugin(self, part_name, plugin_name, part_properties):
             part_schema=self._validator.part_schema,
             definitions_schema=self._validator.definitions_schema)
 
-        self.build_tools += part.code.build_packages
+        # Add multi-arch suffix for cross-compilation to dev packages
+        for package in part.code.build_packages:
+            if self._project_options.is_cross_compiling:
+                if package.endswith('-dev'):
```
sergiusens: to be future proof and this will happen, we need to check for `:` to detect a hardcoding of an arch, an `all` entry or an `any`; in the case of no `:` or `any` I think is when we can append.

sergiusens: @cjwatson can we pretty please get your insight here?

cjwatson: I'm pretty sure that the whole approach of appending architecture suffixes is wrong here, not least because apt has a whole load of non-trivial metadata about which packages need which kind of multi-arch treatment that's being ignored.  Just checking for `-dev` suffixes is going to go wrong in both directions.

Here's the [relevant spec](https://wiki.ubuntu.com/MultiarchCross) for the system-level behaviour.  To use this correctly, you need to somehow arrange to call `apt-get build-dep -a<target-arch>` (the behaviour of `apt-get build-dep` behaves differently in subtle but important ways from `apt-get install` - for instance the `:native` suffix isn't defined for ordinary dependencies, only build-dependencies).

My recommendation would be a somewhat different strategy, which is a bit more work but should be a lot more robust: generate a temporary `.dsc` file with the contents of `build-packages` in its `Build-Depends`, and pass that to `apt-get build-dep` or `apt-get build-dep -a<target-arch>` depending on whether you're cross-compiling.  On xenial you can just pass the `.dsc` directly to `apt-get build-dep`, but if you need to support trusty you need to faff about with turning it into a `Sources` file the same kind of way that `sbuild` does.

If it were me then I'd do that in a separate PR and rebase this one on top of it later ...

sergiusens: thanks @cjwatson

snapcraft/tests/fixture_setup.py
```diff
@@ -405,6 +406,89 @@ def call_effect(*args, **kwargs):
...
+        elif not_available:
+            note = 'Depends: {}{} but it is not installable'
+        else:
+            note = '{}{}'
+        errors = []
+        for package in packages:
+            errors.append(note.format(package, arch))
+        if not_cached or not_available:
+            self.exception = True
+            details = '''Some packages could not be installed. This may mean that you have
```
sergiusens: mind using `dedent` here?

snapcraft/tests/fixture_setup.py
```diff
@@ -405,6 +406,89 @@ def call_effect(*args, **kwargs):
...
+class FakeAptGetBuildDep(fixtures.Fixture):
+    '''Mock apt-get build-dep output'''
+
+    def __init__(self, packages, arch='', update_error=False,
+                 not_cached=False, not_available=False):
+        self.filename = '{}/{}abcdef.dsc'.format(tempfile.gettempdir(),
+                                                 tempfile.gettempprefix())
+        if arch:
+            arch = ':{}'.format(arch)
+        prolog = '''NOTE: This is only a simulation!
```
sergiusens: mind using `dedent` here? `from textwrap import dedent` (look at the new-cli branch occurrences for quick inspiration)

snapcraft/plugins/go.py
sergiusens: I would add an `if self.project.debarch == 'armhf'` here, `GOARM` is probably 8 for arm64 anyways

snapcraft/internal/repo/_deb.py
```diff
@@ -186,20 +188,91 @@ def get_packages_for_source_type(cls, source_type):
...
+            subprocess.check_call(
+                ['sudo', 'dpkg', '--add-architecture', arch])
+
+        sources = _get_local_sources_list()
+        if '[arch={}]'.format(arch) not in sources:
+            logger.info('Adding sources for {}'.format(arch))
+            release = platform.linux_distribution()[2]
+            sources = _format_sources_list(None, deb_arch=arch,
+                                           release=release, foreign=True)
+            with tempfile.NamedTemporaryFile() as sources_arch:
```
sergiusens: can you rename `source_arch` to source_arch_file`? makes the code below a faster read

snapcraft/internal/repo/_deb.py
```diff
@@ -186,20 +188,91 @@ def get_packages_for_source_type(cls, source_type):
         return packages
 
     @classmethod
-    def install_build_packages(cls, package_names):
-        unique_packages = set(package_names)
+    def _setup_multi_arch(cls, apt_cache, package_name):
```
sergiusens: Given the complexity (not the code, but the task, it is invasive after all), can we add a docstring?

snapcraft/internal/repo/_deb.py
```diff
@@ -186,20 +188,91 @@ def get_packages_for_source_type(cls, source_type):
...
+            if 'Err:' in update_output:
+                raise subprocess.CalledProcessError(
+                    100, ['sudo', 'apt-get', 'update'], update_output)
+        except subprocess.CalledProcessError as e:
+            logger.error(e.output)
+            raise e
+        return package_name in apt_cache
+
+    @classmethod
+    def _get_build_deps(cls, package_names, arch):
```
sergiusens: Can we get a docstring here too?

sergiusens:
Seeing some issues

First try, go is not installed
```
sergiusens@mirkwood:~/source/snapcraft/integration_tests/snaps/go-cgo$ ../../../bin/snapcraft snap --target-arch=armhf
Setting target machine to 'armhf'
Preparing to pull go-cgo 
Pulling go-cgo 
Please consider setting `go-importpath` for the 'go-cgo' part
go get -t -d ./go-cgo/...
/tmp/tmpirt49ylk: 10: exec: go: not found
....
subprocess.CalledProcessError: Command '['/bin/sh', '/tmp/tmpirt49ylk', 'go', 'get', '-t', '-d', './go-cgo/...']' returned non-zero exit status 127
```

Second try, first without x-compile
```
sergiusens@mirkwood:~/source/snapcraft/integration_tests/snaps/go-cgo$ ../../../bin/snapcraft pull
Installing build dependencies: golang-1.7-go golang-1.7-src golang-go golang-src libglib2.0-dev libglib2.0-dev-bin libpcre3-dev libpcre32-3 libpcrecpp0v5
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Suggested packages:
  libglib2.0-doc
Recommended packages:
  golang-1.7-race-detector-runtime golang-race-detector-runtime
The following NEW packages will be installed:
  golang-1.7-go golang-1.7-src golang-go golang-src libglib2.0-dev
  libglib2.0-dev-bin libpcre3-dev libpcre32-3 libpcrecpp0v5
0 upgraded, 9 newly installed, 0 to remove and 1 not upgraded.
Need to get 0 B/28,0 MB of archives.
After this operation, 154 MB of additional disk space will be used.
...
Preparing to pull go-cgo 
Pulling go-cgo 
Please consider setting `go-importpath` for the 'go-cgo' part
go get -t -d ./go-cgo/...
```

Now I can x-compile, but wait...
```
sergiusens@mirkwood:~/source/snapcraft/integration_tests/snaps/go-cgo$ ../../../bin/snapcraft snap --target-arch=armhf
Setting target machine to 'armhf'
Preparing to pull go-cgo 
Pulling go-cgo 
Please consider setting `go-importpath` for the 'go-cgo' part
go get -t -d ./go-cgo/...
Preparing to build go-cgo 
Building go-cgo 
Please consider setting `go-importpath` for the 'go-cgo' part
go build -o /home/sergiusens/source/snapcraft/integration_tests/snaps/go-cgo/parts/go-cgo/go/bin/go-cgo ./go-cgo/...
# runtime/cgo
exec: "arm-linux-gnueabihf-gcc": executable file not found in $PATH
...
```

snapcraft/internal/repo/_deb.py
```diff
@@ -186,8 +188,100 @@ def get_packages_for_source_type(cls, source_type):
...
+                        name += ':{}'.format(arch)
+                    else:
+                        version = version_with_arch
+                    name += '={}'.format(version)
+                pkgs.append(name)
+            return pkgs
+
+    @classmethod
+    def install_build_packages(cls, package_names, arch):
+        unique_packages = set(cls._get_build_deps(package_names, arch))
```
elopio: This is crashing for me:

 > /home/ubuntu/workspace/canonical/snapcraft/snapcraft/internal/repo/_deb.py(286)install_build_packages()
-> new_packages = []
(Pdb) unique_packages
{'for', 'Keep', 'privileges', 'so', 'package', 'get', 'build', 'in', 'relevance', "don't", "'/tmp/tmpvnu0mvxc.dsc'", 'tree...', 'file', 'the', 'is', 'dependencies', 'depend', 'real', 'on', 'also', 'deactivated,', 'locking', 'execution.', 'Reading', 'information...', 'current', 'needs', 'Building', 'mind', 'using', 'root', 'state', 'to', 'situation!', 'lists...', 'that', 'Note,', 'dependency'}

$ apt-get build-dep -q -s -aarm64 golang-go:native        
NOTE: This is only a simulation!
      apt-get needs root privileges for real execution.
      Keep also in mind that locking is deactivated,
      so don't depend on the relevance to the real current situation!
E: No architecture information available for arm64. See apt.conf(5) APT::Architectures for setup

snapcraft/plugins/go.py
```diff
@@ -191,4 +195,23 @@ def _build_environment(self):
         env['CGO_LDFLAGS'] = '{} {} {}'.format(
             env.get('CGO_LDFLAGS', ''), flags, env.get('LDFLAGS', ''))
 
+        if self.project.is_cross_compiling:
+            env['CC'] = '{}-gcc'.format(self.project.arch_triplet)
+            env['CXX'] = '{}-g++'.format(self.project.arch_triplet)
+            env['CGO_ENABLED'] = '1'
+            env['PKG_CONFIG_PATH'] = '/usr/lib/{}/pkgconfig'.format(
```
kyrofa: This should not be needed. Check project loader for more details.

snapcraft/plugins/go.py
```diff
@@ -144,22 +144,26 @@ def _get_local_go_package(self):
...
-        for go_package in self.options.go_packages:
-            self._run(['go', 'install'] + tags + [go_package])
-        if not self.options.go_packages:
-            self._run(['go', 'install'] + tags +
-                      ['./{}/...'.format(self._get_local_go_package())])
+        packages = self.options.go_packages
+        if not packages:
+            packages = ['./{}/...'.format(self._get_local_go_package())]
+        for package in packages:
+            binary = self._gopath_bin + '/' + self._binary_name(package)
```
kyrofa: os.path.join

kalikiana:
This branch was split into smaller following pieces:

- [go plugin: Add support for cross-compilation](https://github.com/snapcore/snapcraft/pull/1338)
- [go plugin: Cross compile with CGo](https://github.com/snapcore/snapcraft/pull/1343)
- Figure out build dependencies for the target architecture
- Setup a multi-arch environment

PR #1255, opened 2017-04-13
Update rust plugin to set RUSTFLAGS

ChrisMacNaughton:
Update the rust plugin to link shared libraries in correctly

### DIFF
--- a/snapcraft/plugins/rust.py
+++ b/snapcraft/plugins/rust.py
@@ -38,6 +38,7 @@ import shutil
 
 import snapcraft
 from snapcraft import sources
+from snapcraft.internal import common
 
 _RUSTUP = 'https://static.rust-lang.org/rustup.sh'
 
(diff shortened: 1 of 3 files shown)


snapcraft/plugins/rust.py
```diff
@@ -105,9 +106,18 @@ def _build_env(self):
...
                     "RUSTDOC": self._rustdoc,
-                    "RUST_PATH": self._rustlib})
+                    "RUST_PATH": self._rustlib,
+                    'RUSTFLAGS': self._rustflags()})
         return env
 
+    def _rustflags(self):
+        ldflags = shell_utils.getenv('LDFLAGS')
+        rustldflags = ''
+        flags = {flag for flag in ldflags.split(' ') if flag}
```
sergiusens: this could probably just be
```python
set(ldflags.split(' '))
```

ChrisMacNaughton: I ran into some weirdness where the split had extra whitespace, which ended up generating a command that had an empty argument :-/

snapcraft/plugins/rust.py
```diff
@@ -105,9 +106,19 @@ def _build_env(self):
...
-                    "RUST_PATH": self._rustlib})
+                    "RUST_PATH": self._rustlib,
+                    'RUSTFLAGS': self._rustflags()})
         return env
 
+    def _rustflags(self):
+        ldflags = shell_utils.getenv('LDFLAGS')
+        rustldflags = ''
+        flags = set(ldflags.split(' '))
+        flags.remove('')
```
sergiusens: what does this accomplish?

ChrisMacNaughton: it removes empty strings ex:
```python
>>> parts = set('  test '.split(' '))
>>> parts
{'', 'test'}
>>> parts.remove('')
>>> parts
{'test'}
```

sergiusens: oh, in that case, the list (set) comprehension was better.

sergiusens: I'll roll that back and setup the core setup flag for the test

sergiusens: I will reset instead of revert as you seemed to use your gmail email and that email has not been used to sign the CLA

snapcraft/plugins/rust.py
sergiusens: the `if flag` in the original fixes that, we can add `flag.strip() for flag in ...`  but I don't think we need that.

ChrisMacNaughton: exactly :)

PR #1257, opened 2017-04-13
core: support running from other operating systems.

sergiusens:
As long as the operating system provides the right toolset, allow
cleanbuilds to go through.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,7 @@ progressbar33==2.4
 PyYAML==3.11
 pyxdg==0.25
 requests==2.9.1
-https://launchpad.net/ubuntu/+archive/primary/+files/python-apt_1.1.0~beta1build1.tar.xz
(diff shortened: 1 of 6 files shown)


requirements.txt
```diff
@@ -7,7 +7,7 @@ progressbar33==2.4
 PyYAML==3.11
 pyxdg==0.25
 requests==2.9.1
-https://launchpad.net/ubuntu/+archive/primary/+files/python-apt_1.1.0~beta1build1.tar.xz
+https://launchpad.net/ubuntu/+archive/primary/+files/python-apt_1.1.0~beta1build1.tar.xz; sys_platform == 'linux'
```
kyrofa: This isn't necessary for `python-debian`?

sergiusens:
I don't want to QA what will happen if we don't add that message

requirements.txt
sergiusens: no it is not

PR #1258, opened 2017-04-17
python-plugin: support pip list columns mode

sdeancos:
When pip list use format=columns mode the regex not work, this patch add new regex.

LP: [#1683395](https://bugs.launchpad.net/snapcraft/+bug/1683395)


sergiusens:
Thanks for the contribution! Some travis errors have however showed up:

```
snapcraft/plugins/python.py:459:5: E303 too many blank lines
Checking the licence agreement for sdeancos@gmail.com...
The contributor has a Launchpad account, but has not signed the CLA.
```

Mind sorting that out?

sdeancos:
Hi! Thanks so much, sorry, i fix for the E303 bug.

I'm trying signed the CLA but I do not know what I have to fill in the field "Please add the Canonical Project Manager or contact" maybe can you help me?

elopio:
@sdeancos you can fill Jamie Bennett as the manager.

snapcraft/plugins/python.py
```diff
@@ -443,9 +443,14 @@ def list(self, exec_func=None):
         output = exec_func(cmd, env=self._env)
         package_listing = {}
         version_regex = re.compile('\((.+)\)')
+        version_regex_alt = re.compile('((\d+)\.)?((\d+)\.)?(\*|\d+)')
         for line in output.splitlines():
             line = line.split()
             m = version_regex.search(line[1])
+            if not m:
```
elopio: version_regex_alt is only used inside this if block, so it would be better if you move the variable assignment here.

snapcraft/plugins/python.py
```diff
@@ -443,9 +443,14 @@ def list(self, exec_func=None):
         output = exec_func(cmd, env=self._env)
         package_listing = {}
         version_regex = re.compile('\((.+)\)')
+        version_regex_alt = re.compile('((\d+)\.)?((\d+)\.)?(\*|\d+)')
```
elopio: `alt` is not a very good name. Maybe `version_regex_with_columns` ?
Also, I had no idea this columns thing existed. What about adding a few lines with a comment explaining why is this here?

sdeancos:
Hi, I'm trying write an integration test but its a problem of environment, in that case (in my machine):

$ cat ~/.pip/pip.conf 
[list]
format=columns

kalikiana:
You can set PIP_CONFIG_FILE to avoid using the actual configuration in home

elopio:
So, from @kalikiana's suggestion, I'm now thinking that it would be better to disable PIP_CONFIG_FILE for all the commands in the plugin. There are some things that if overwritten in the user config file would make the build/install fail. For example, take a look at https://github.com/ContinuumIO/anaconda-issues/issues/1709

@sdeancos it would be nice to have this in the next version of snapcraft. Will you have some time to continue with this pulll request? Thanks for your work so far, and for bringing this issue to our attention.

sdeancos:
Hi all!

I'm going to try to close this topic, I'm sorry, lately I have not had time.

kyrofa:
I've gone ahead and closed it. Please let us know if you'd like to continue with it.

elopio:
Hey @sdeancos, I'm going to take over this. My fix would be to call pip list --format=json instead, so we won't depend on your local configuration. Your review will be nice, it should be ready tomorrow.

sdeancos:
Hi @elopio ! It's a very good idea. I think it's a good solution. :D

PR #1260, opened 2017-04-17
meta: version from git support.

sergiusens:
If `version: git` is set, snapcraft will determine
the version using git and its tags.

### DIFF
--- a/integration_tests/test_git_source.py
+++ b/integration_tests/test_git_source.py
@@ -219,4 +219,3 @@ class GitGenerateVersionTestCase(GitSourceBaseTestCase):
             exception.output,
             Contains('fatal: Not a git repository (or any of the parent '
                      'directories): .git'))
-
--- a/integration_tests/test_git_source.py
+++ b/integration_tests/test_git_source.py
@@ -18,7 +18,7 @@ import subprocess
 import shutil
 from textwrap import dedent
 
-from testtools.matchers import FileExists
+from testtools.matchers import Contains, FileExists
 import integration_tests
 
 
@@ -209,3 +209,14 @@ class GitGenerateVersionTestCase(GitSourceBaseTestCase):
         revno = self.get_revno()[:7]
         expected_file = 'git-test_0+git.{}_amd64.snap'.format(revno)
         self.assertThat(expected_file, FileExists())
+
+    def test_no_git(self):
+        shutil.rmtree('.git')
+
+        exception = self.assertRaises(
+            subprocess.CalledProcessError, self.run_snapcraft, ['snap'])
+        self.assertThat(
+            exception.output,
+            Contains('fatal: Not a git repository (or any of the parent '
+                     'directories): .git'))
+
--- a/integration_tests/test_git_source.py
+++ b/integration_tests/test_git_source.py
@@ -13,27 +13,50 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
-
+import os
 import subprocess
 import shutil
+from textwrap import dedent
 
+from testtools.matchers import FileExists
 import integration_tests
 
 
-class GitSourceTestCase(integration_tests.TestCase):
+class GitSourceBaseTestCase(integration_tests.TestCase):
 
     def setUp(self):
         super().setUp()
         if shutil.which('git') is None:
             self.skipTest('git is not installed')
 
-    def _init_and_config_git(self):
+    def init_and_config_git(self):
         subprocess.check_call(
             ['git', 'init', '.'], stdout=subprocess.DEVNULL)
         subprocess.check_call(
             ['git', 'config', '--local', 'user.name', '"Example Dev"'])
         subprocess.check_call(
             ['git', 'config', '--local', 'user.email', 'dev@example.com'])
+
+    def add_file(self, file_path):
+        subprocess.check_call(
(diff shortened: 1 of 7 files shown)


snapcraft/internal/sources/_git.py
```diff
@@ -15,14 +15,54 @@
...
+            output = subprocess.check_output([
+                'git', '-C', source_dir, 'describe',
+                '--dirty']).decode(sys.getfilesystemencoding()).strip()
+        except subprocess.CalledProcessError:
+            output = subprocess.check_output([
+                'git', '-C', source_dir, 'describe', '--tags', '--dirty',
+                '--always']).decode(sys.getfilesystemencoding()).strip()
+            return '0+git.{}'.format(output)
+
+        m = re.search(r'(?P<tag>.*)-(?P<revs_ahead>.*)-g(?P<commit>.*)',
```
chipaca: to avoid surprises I'd suggest anchoring this at both ends (i.e. start it with `^`, end it with `$`). Also if `revs_ahead` is expected to be a number, you want `\d+`, not `.*`. Also nothing here should be `.*`; you want at least one char, and they should all be non-whitespace probably (so `\S+`), and I suspect `commit` is expected to be hexadecimal so `[0-9a-fA-F]+`.

sergiusens: thanks, this is exactly why I wanted you to review!

chipaca: hah! you know me too well 😁

sergiusens: For the tag, it can be anything but whitespace, so I stole our regex in place from `schema/snapcraft.yaml` which for version is `^[a-zA-Z0-9.+~-]*$`

integration_tests/test_git_source.py
```diff
@@ -142,3 +165,47 @@ def test_pull_git_with_depth(self):
...
+            print(dedent("""\
+                name: git-test
+                version: git
+                summary: test git generated version
+                description: test git generated version with git hint
+                architectures: [amd64]
+                parts:
+                    nil:
+                        plugin: nil
+                """), file=f)
```
elopio: Almost all the other tests use an existing yaml instead of writing one. Is there a reason here to do it differently?

snapcraft/internal/sources/_git.py
```diff
@@ -15,14 +15,57 @@
...
 from ._base import Base
 
 
 class Git(Base):
 
+    @classmethod
+    def generate_version(cls, *, source_dir=None):
+        """Return the latest git tag from PWD or defined source_dir.
+
+        The output depends on the use of annotated tags, an will return
```
elopio: typo: an -> and

snapcraft/internal/sources/_git.py
```diff
@@ -15,14 +15,57 @@
...
+        hash of the latest commit.
+        """
+        if not source_dir:
+            source_dir = os.getcwd()
+
+        try:
+            output = subprocess.check_output([
+                'git', '-C', source_dir, 'describe',
+                '--dirty']).decode(sys.getfilesystemencoding()).strip()
+        except subprocess.CalledProcessError:
```
elopio: It would be nice to add a comment here about what does this exception mean.

snapcraft/tests/sources/test_git.py
```diff
@@ -383,3 +385,36 @@ def test_git_details_tag(self):
...
+        super().setUp()
+
+        patcher = mock.patch('subprocess.check_output')
+        self.output_mock = patcher.start()
+        self.addCleanup(patcher.stop)
+
+    def test_version(self):
+        if self.tag:
+            self.output_mock.return_value = self.return_value.encode('utf-8')
+        else:
```
elopio: I would prefer to move the no-tag case to a different test, to avoid the `if` block here.

integration_tests/test_git_source.py
sergiusens: Because there are no files aside from snapcraft.yaml added. I've been finding it a pain to easily troubleshoot these split ones when they break.

sergiusens:
> And finally, PR should be accompanied by a pr in snappy-docs explaining the new feature

This is what the forum post is for :wink:

snapcraft/internal/sources/_git.py
sergiusens: done

snapcraft/tests/sources/test_git.py
sergiusens: done

snapcraft/internal/sources/_git.py
sergiusens: fixed

sergiusens:
```
fatal: Not a git repository (or any of the parent directories): .git
fatal: Not a git repository (or any of the parent directories): .git
Command '['git', '-C', '/tmp/local-plugin', 'describe', '--tags', '--dirty', '--always']' returned non-zero exit status 128
```
is fixed

sergiusens:
> When I don't have a tag, this message is printed:
```
fatal: No names found, cannot describe anything.
```
this is no longer shown

elopio:
@sergiusens: I don't think that this is enough:
fatal: Not a git repository (or any of the parent directories): .git

It's not mentioning that the error comes from the version field. My first reaction would be to check the parts, not the version. And as the parts are likely to come from git, I would find myself confused for a while.

I'm pretty happy with the rest, and tested it enough. So I'll leave my +1. But please consider extending the message.

PR #1261, opened 2017-04-18
storeapi:  improve the error message for the case the Store answers an upload needs manual review.

facundobatista:
LP: [#1670471](https://bugs.launchpad.net/snapcraft/+bug/1670471)


snapcraft/storeapi/errors.py
```diff
@@ -266,10 +266,14 @@ def __init__(self, snap_name, response):
 class StoreReviewError(StoreError):
 
     __FMT_NEED_MANUAL_REVIEW = (
-        'Publishing checks failed.\n'
-        'To release this to stable channel please request a review on '
-        'the snapcraft list.\n'
-        'Use devmode in the edge or beta channels to disable confinement.')
+        "The Store automatic review failed.\n"
+        "A human will soon review your snap, but if you can't wait please "
+        "write to the snapcraft mailing list asking for the manual review "
```
elopio: now the forum should be used, instead of the mailing list.

facundobatista:
Thanks @elopio ! 

There I changed the text to mention the forum, not the mailing list.

Regarding the test case, I don' t think there is value in providing a test for just a simple text (with "simple" meaning that the text is not built or constructed in any way) as we have in this case.

Imagine how the test for this constant would be, just the text repeated in the code and the test, and if a change needed, the change is reproduced in both places, no really asserting nothing.

IMO constants are not to be tested, is like testing specific values of settings or configs.

Thanks again!

elopio:
@facundobatista I agree more or less about the text. At some point I would love to have tests specifying the UX, instead of google docs, but we are not close to that yet. As we are usually at least checking parts of the message, I thought that we didn't even have a test for the exception. But I checked and we have test_upload_requires_review. All good, +1

PR #1263, opened 2017-04-19
lxd: pass through commands into the container

kalikiana:
All commands around building snaps and dealing with parts will be passed through into the lxd container.

### DIFF
--- a/manual-tests.md
+++ b/manual-tests.md
@@ -53,8 +53,8 @@
 
 # Test containerized building
 
-1. Setup a remote as described on
-   https://linuxcontainers.org/lxd/getting-started-cli/#multiple-hosts
+1. Setup LXD as described on
+   https://linuxcontainers.org/lxd/getting-started-cli/
 2. Select a project <project> to build.
 3. Run `SNAPCRAFT_CONTAINER_BUILDS=1 snapcraft`.
 4. Run `SNAPCRAFT_CONTAINER_BUILDS=1 snapcraft clean` and observe that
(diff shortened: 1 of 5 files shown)


snapcraft/main.py
```diff
@@ -281,16 +281,34 @@ def _get_command_from_arg(args):
     return functions[function[0]]
 
 
+def _containerbuild(args):
```
sergiusens: the conventions in python is to prefix with `is`, so this would be `_is_...`

sergiusens:
This is the problem I see with remotes:
```
sergiusens@mirkwood:~/source/snapcraft/demos/godd$ SNAPCRAFT_CONTAINER_BUILDS=1 ~/source/snapcraft/bin/snapcraft --remote shire
Creating snapcraft-godd
Mounting /home/sergiusens/source/snapcraft/demos/godd into container
error: Failed to setup device: Source path /home/sergiusens/source/snapcraft/demos/godd doesn't exist for device /build_godd
Stopping shire:snapcraft-godd
Command '['lxc', 'config', 'device', 'add', 'shire:snapcraft-godd', 'build_godd', 'disk', 'source=/home/sergiusens/source/snapcraft/demos/godd', 'path=/build_godd']' returned non-zero exit status 1
```

manual-tests.md
```diff
@@ -50,6 +50,17 @@
 3. Run `snapcraft cleanbuild --remote <remote>` where `<remote>` is
    the name you gave the remote on step 1.
 
+
+# Test containerized building
+
+1. Setup a remote as described on
```
sergiusens: The remote step here should now be gone and replaced with something about having a proper lxd setup

PR #1264, opened 2017-04-19
tests: do not print coverage check message.

sergiusens:
Filter the output when running ./runtests.sh where we check if coverage is
available by running `python3 -m coverage`.


elopio:
I would prefer something like #1266

PR #1265, opened 2017-04-19
shell_utils: code cleanup.

sergiusens:
Clean up code for better reuse and make more of the logic
make use of these helpers.


sergiusens:
@elopio `common.run_output` already decodes and has enough fall back logic to not break everywhere  :smile:

PR #1266, opened 2017-04-19
tests: report coverage only in unit tests

elopio:


sergiusens:
this doesn't fix the spurious error when I run it locally though.

sergiusens:
I still think you need to implement what is in my branch

elopio:
@sergiusens ah, I see what you were getting at there. I copied your change here. We could have just landed both, sorry :)

elopio:
Please don't land this branch yet. Coverage is not working because of:
`can use starred expression only as assignment target`

I think that's because codecov we have in travis is py2, and that's py3 syntax. Maybe, we can do everything in docker, I'll try.

elopio:
Score! This is ready to go.

PR #1270, opened 2017-04-20
release changelog for 2.29

sergiusens:

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,72 @@
+snapcraft (2.29) xenial; urgency=medium
+
+  [ Sergio Schvezov ]
+  * store: better retry strategy for GETs. (#1212)
+  * docs: remove docs and link to snappy-docs (#1227)
+  * nodejs plugin: switch to the newer LTS. (#1228)
+  * pluginhandler: exclude `/snap/` from libraries. (#1231)
+  * nodejs plugin: switch to the newer LTS. (#1228) (#1232)
+  * core: find the correct libraries as a snap. (#1234)
+  * snap: use the gpg tarball instead of git://. (#1236)
+  * docs: update contribution guide. (#1233)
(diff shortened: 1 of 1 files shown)


debian/changelog
```diff
@@ -1,3 +1,72 @@
+snapcraft (2.29) xenial; urgency=medium
+
+  [ Sergio Schvezov ]
+  * store: better retry strategy for GETs. (#1212)
+  * docs: remove docs and link to snappy-docs (#1227)
```
elopio: nit. It's missing the period at the end.

debian/changelog
```diff
@@ -1,3 +1,72 @@
...
+
+  [ Colin Watson ]
+  * ant plugin: honour proxy configuration (#1256)
+
+  [ Chris MacNaughton ]
+  * rust plugin: update plugin to set RUSTFLAGS (#1255)
+
+  [ Andy Li ]
+  * docs: fixed links to doc (#1250)
+  * tests: report coverage only in unit tests. (#1266)
```
elopio: this is not from Andy Li, it's weird that it ended here.

debian/changelog
```diff
@@ -1,3 +1,72 @@
+snapcraft (2.29) xenial; urgency=medium
+
+  [ Sergio Schvezov ]
+  * store: better retry strategy for GETs. (#1212)
+  * docs: remove docs and link to snappy-docs. (#1227)
+  * nodejs plugin: switch to the newer LTS. (#1228)
+  * pluginhandler: exclude `/snap/` from libraries. (#1231)
+  * nodejs plugin: switch to the newer LTS. (#1228) (#1232)
```
elopio: That (#1228) should probably be removed. Maybe you should also remove the previous entry about switching to the newer LTS.

PR #1271, opened 2017-04-20
tests: increase the staging registration limit to 100

elopio:
LP: [#1680066](https://bugs.launchpad.net/snapcraft/+bug/1680066)


elopio:
Despite the failure, this is good to go because it fixes the registration limit tests.
The remaining errors are an already reported bug because the push times out, and a new one because staging now allows 50 characters in snap names. I'm talking to the store people to see what's the right fix in that case.

PR #1274, opened 2017-04-21
Simple explanation of what the test groups mean

facundobatista:


HACKING.md
```diff
@@ -77,3 +77,10 @@ We assume you have run through the installation instructions, to run all the tes
 You can selectively run a selective group of tests like:
 
     ./runtests.sh [static|unit|integration|snaps]
+
+The test groups mean:
+
+- static: static code analysis done with flake8
+- unit: run the standard unit tests
+- integration: shells to snapcraft, builds
```
elopio: sadly, we had to split integration because it was timing out. So now we have integration|plugins|store.
I couldn't find a better name for the ones that remain in the `integration` dir :(
And I hope this is temporary until we can do proper parallelization with cleanbuild.

So, I don't now how to explain all this. Any ideas are welcome. For now, +1. I'll come back and expand on integration.

PR #1275, opened 2017-04-21
init: add a newline at the end of the file

roxyd:
LP: [#1660880](https://bugs.launchpad.net/snapcraft/+bug/1660880)

### DIFF
--- a/snapcraft/internal/lifecycle.py
+++ b/snapcraft/internal/lifecycle.py
@@ -60,7 +60,7 @@ parts:
   my-part:
     # See 'snapcraft plugins'
     plugin: nil
-"""  # noqa, lines too long.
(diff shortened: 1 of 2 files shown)


snapcraft/internal/lifecycle.py
```diff
@@ -60,7 +60,7 @@
   my-part:
     # See 'snapcraft plugins'
     plugin: nil
-"""  # noqa, lines too long.
+""" # noqa, lines too long.
```
elopio: It's usually a good idea to check the `git diff` before making the commit, to make sure that only the lines you intended changed. In this case it's just a whitespace so no big deal.

roxyd: Thank you!

PR #1276, opened 2017-04-24
sources: validate unknown source-type in yaml

EduardoVega:
Invalid source-type defaults to Local source (handler) and it doesn't print an error message

LP: [#1650943](https://bugs.launchpad.net/snapcraft/+bug/1650943)


snapcraft/internal/sources/__init__.py
```diff
@@ -164,7 +165,14 @@ def get_source_handler(source, *, source_type=''):
     if not source_type:
         source_type = _get_source_type_from_uri(source)
 
-    return _source_handler.get(source_type, Local)
+    handler = _source_handler.get(source_type, None)
+
+    if not handler:
+        raise ValueError((
+            'invalid snapcraft.yaml: '
+            'unknown source type: {}').format(source_type))
```
elopio: @sergiusens can you please check if this error is consistent with our other messages?
I think it could be something like:
ValueError('no handler to manage source type ({})'.format(source_type))

EduardoVega:
@sergiusens and @elopio , thanks for the feedback. I wasn't sure about the change either, I did it because, I wanted to avoid changing the schema every time a new source handler is added to the _source_handler dictionary. 

So basically creating an enum in the snapcraft.yaml with the current souce handlers in  _source_handler dictionary should do the work?

sergiusens:
Thanks, just one more minor thing to be complete now. For users of the API, so they get a proper `KeyError` we should change

```python
return _source_handler.get(source_type, Local)
```

to 
```python
return _source_handler.get[source_type]
```

Make sense? I hope so :smile:

EduardoVega:
@sergiusens , Yes, it does. But I would also add:
```
_source_handler = {
    'bzr': Bazaar,
    'git': Git,
    'hg': Mercurial,
    'mercurial': Mercurial,
    'subversion': Subversion,
    'svn': Subversion,
    'tar': Tar,
    'zip': Zip,
    '': Local,
}
```

Because the default value in `def get_source_handler(source, *, source_type=''):` and in the `snapcraft.yaml` is `''`, so whitout it, it would send KeyError every time default is used.

Let me know what do you think?

elopio:
@EduardoVega the CLA automated check is failing there in travis because you made some commits or the PR with the email eduardo.vega.valerio@intel.com.
You will have to ammend your commits, or sign the CLA also with that address to get it green.

PR #1277, opened 2017-04-24
Handle revoked-uploads case on snap-developer revokes. 

psivaa:
Revoking developers via snap-developer means that the uploads they performed in the past will need to be invalidated too.
This PR is to warn the user about the existence of such uploads and to enable a force option.


snapcraft/_store.py
```diff
@@ -847,7 +848,20 @@ def collaborate(snap_name, key):
...
+    except storeapi.errors.StoreValidationError as e:
+        if e.error_list[0]['code'] == 'revoked-uploads':
+            print("This will revoke the following uploads: {}".format(
+                e.error_list[0]['extra']))
+            if input("Are you sure you want to continue (y/N): ") == "y":
+                store.push_assertion(
+                    snap_id, assertion, 'developers', force=True)
+            else:
+                print("The collaborators for this snap haven't been altered")
+                sys.exit(1)
```
elopio: Is it common to exit with error if the user cancels the action? It doesn't sound ok to me, but I don't know.
Anyway, the right way to exit is raising an exception with the message, that then will be printed by main and exit.

snapcraft/tests/commands/test_collaborate.py
```diff
@@ -90,3 +91,17 @@ def test_collaborate_bad_request(self):
...
+        mock_input = patcher.start()
+        mock_input.return_value = 'y'
+        self.client.login('dummy', 'test correct password')
+        err = self.assertRaises(
+            storeapi.errors.StoreValidationError,
+            collaborate,
+            'revoked', 'keyname')
+
+        self.assertEqual(
+            'Received error 409: "The assertion\'s `developers` would revoke '
```
elopio: The exception to the rule of preferring ' for strings is when ' is inside the string. In that case, we use " to avoid the \.

elopio:
No integration test here @psivaa ? I get nervous when the store commands get only unit tests.

sergiusens:
@elopio only now that snapd 2.24 has reach the masses we can add integration tests.

snapcraft/tests/commands/test_collaborate.py
psivaa: This error message was a nested one, with both `"` and `'`  inside the string. Used str.format() instead now to avoid `\`.

snapcraft/_store.py
psivaa: Thanks. This is now fixed to print a message and re-raise StoreValidationError

psivaa:
@elopio @sergiusens I don't think I can add integration tests for this particular branch changes since the store side implentation of 'revoked-uploads' has not landed yet. 
But I should be able to add integ tests for the happy paths, which I'll do as a follow up PR.

snapcraft/tests/fake_servers.py
```diff
@@ -841,6 +841,10 @@ def do_GET(self):  # noqa: C901
             self._DEV_API_PATH, 'snaps/no-dev/developers')
         bad_developers_path = urllib.parse.urljoin(
             self._DEV_API_PATH, 'snaps/badrequest/developers')
+        revoked_developers_path = urllib.parse.urljoin(
+            self._DEV_API_PATH, 'snaps/revoked/developers')
+        not_revoked_developers_path = urllib.parse.urljoin(
+            self._DEV_API_PATH, 'snaps/no-revoked/developers')
```
elopio: Note the refactors here to make this shorter:
https://github.com/snapcore/snapcraft/pull/1288/files#diff-2a6de018c4c141826eaa1f76b9b4c31aR839

They will cause conflicts with your branch.

kyrofa:
This hasn't moved for a few months, I'm going to go ahead and close it. We can reopen when we're ready to make more progress.

PR #1278, opened 2017-04-25
recording: save the snapcraft.yaml in the resulting snap

elopio:
Record the snapcraft.yaml in prime/snap/snapcraft.yaml.

This is missing the annotation, it's just a simple version to get started and discuss the details.


snapcraft/tests/test_lifecycle.py
```diff
@@ -618,6 +618,31 @@ def test_pull_is_dirty_if_target_arch_changes(
...
+version: 0
+summary: test
+description: test
+confinement: strict
+grade: stable
+parts:
+  test-part:
+    plugin: nil
+    prime: []
+    snap: []
```
elopio: @sergiusens this item shouldn't be here, right? I think we should remove it when we read the yaml, so we don't carry over the deprecated name until this late step.

snapcraft/tests/test_lifecycle.py
```diff
@@ -618,6 +618,31 @@ def test_pull_is_dirty_if_target_arch_changes(
...
+        expected = ("""name: test
+version: 0
+summary: test
+description: test
+confinement: strict
+grade: stable
+parts:
+  test-part:
+    plugin: nil
+    prime: []
```
elopio: The default values for prime and snap are not [], they are ['*']. We should take them too from the state files, right?

sergiusens: hmm, this needs updating in our schema then

snapcraft/tests/test_lifecycle.py
sergiusens: not sure how this makes it here as we pop it when loading the project. Would be nice to track down.

snapcraft/internal/lifecycle.py
```diff
@@ -252,6 +253,14 @@ def _create_meta(self, step, part_names):
             meta.create_snap_packaging(self.config.data,
                                        self.project_options)
 
+    def _record_snapcraft(self, step, part_names):
```
sergiusens: Take into account that this logic is already implemented in `meta`, somewhat.

snapcraft/tests/test_lifecycle.py
elopio: the schema is correct, but it somehow doesn't make it to self.config.data. I can dig back.

snapcraft/internal/lifecycle.py
elopio: I see it now. I missed it because of the flag. I would just remove it. Or do you have a good reason for it being in meta ?

sergiusens: well at least I don't think it is correct to have it in lifecycle

sergiusens: we would also be splitting logic and ownership if it is not in meta; the `snap` directory logic would be split up otherwise

elopio: I moved it to meta.py. But lets discuss it tomorrow in the standup, because I think we can do better.
I would like to split the logic that puts things in prime/meta from the one that puts things in prime/snap. 

Hooks, .desktop and icon are taken from $PROJECT_ROOT/snap and put into prime/meta. For now, I would leave those functions in meta.py. But we are calling them assets, maybe assets.py could work too.

elopio:
@sergiusens I'm digging into the prime and stage value, and how to remove the snap key. But I prefer to do that in a separate branch.

elopio:
@sergiusens everything is loaded in ordered dicts by https://github.com/snapcore/snapcraft/blob/master/snapcraft/__init__.py#L447

Which kind of sucks, because it's a side-effect of the import. As such, it affects the unit tests too, but we don't really control it. It works, good for now.

PR #1279, opened 2017-04-25
misc: rename the snap dir variable to prime dir

elopio:


elopio:
@sergiusens if you are happy with this one, let's land it first. It will make the rest more readable.

PR #1280, opened 2017-04-25
Added support for 'branches' in Store responses (release, close, and status).

facundobatista:
While doing this, I also changed the behaviour for the case of not understanding some branch info that could come from the server: now, instead of exploding, it just logs in error and present that channel empty (this will ease future evolutions that may be needed; for example, currently these commands would explode when working on a package with branches!).

Also, I removed the "_tabulated_status" function, as wasn't used.

Finally, removed several FakeTerminal instantiaton in a lot of tests that the only thing they do was to shadow the same thing from their base class.

LP: [#1686162](https://bugs.launchpad.net/snapcraft/+bug/1686162)


sergiusens:
Why are branches only for stable?
"Branch must target stable risk"

sergiusens:
Also, we call it grade everywhere else... isn't there mixed terminology

PR #1282, opened 2017-04-26
parts: remove the deprecated snap keyword from the internal parts representation

elopio:


elopio:
This requires #1278

snapcraft/internal/parts.py
```diff
@@ -188,10 +188,11 @@ def _process_parts(self):
             if 'filesets' in properties:
                 del properties['filesets']
 
-            # FIXME: snap is deprecated, rewrite it to prime instead.
-            if properties.get('snap'):
+            # Handle the deprecated snap keyword.
+            snap = properties.pop('snap')
```
elopio: This could be done earlier, but that would mean another sweep on the entire loaded yaml. I think it's fine to keep it here.

PR #1286, opened 2017-04-27
lxd: Setup image and target arch for cross-compilation

kalikiana:
To support cross-compilation in LXD containers the image for the host architecture needs to be used, rather than for the target architecture, which would require slow qemu emulation. It's worth noting that host here means the host running LXD which may not be the same as the host running snapcraft eg. snapcraft executed on an armhf machine using an amd64 remote building for arm64.

Note: this is orthogonal to actual support for cross-compilation in individual plugins and only deals with the container setup.

### DIFF
--- a/snapcraft/internal/lxd.py
+++ b/snapcraft/internal/lxd.py
@@ -40,11 +40,6 @@ _PROXY_KEYS = ['http_proxy', 'https_proxy', 'no_proxy', 'ftp_proxy']
 
 class Containerbuild:
 
-    def _get_server_info(self):
-        remote = self._container_name.split(':')[0]
-        return yaml.load(check_output([
-            'lxc', 'info', '{}:'.format(remote)]).decode())
-
     def __init__(self, *, output, source, project_options,
                  metadata, container_name, remote=None):
         if not output:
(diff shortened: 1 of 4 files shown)


snapcraft/internal/lxd.py
```diff
@@ -38,6 +40,11 @@
 
 class Containerbuild:
 
+    def _get_server_info(self):
```
elopio: Another nit, can you move this method after __init__?
I love some of the suggestions from the [clean code](https://www.safaribooksonline.com/library/view/clean-code/9780136083238/) book. One of those is vertical organization, you put first the caller method, and after the called method. This leaves the functions close to were they were called first, but the nicest thing for me is that you can read a source code file from top to bottom. In here if you find _get_server_info you have no context about what it does. If you instead find __init__ first, you will see the context where _get_server_info is called and then you can decide if the internals of that method are relevant to what you are looking for, or you can just skip it.

elopio: shouldn't it be called _get_remote_info instad?

snapcraft/tests/test_lxd.py
```diff
@@ -46,20 +47,27 @@ def test_cleanbuild(self, mock_pet):
 
         mock_pet.return_value = 'my-pet'
 
-        project_options = ProjectOptions()
+        with patch('platform.machine') as machine_mock, \
```
elopio: nit: prefer parentheses instead of splitting lines with `\`

kalikiana: I don't see how that's possible here. But I figured I can use decorators anyway so there's no need for the with at all.

PR #1287, opened 2017-04-27
tests: minor cleanups on the spread tests

elopio:


elopio:
Thanks to @fgimenez for the suggestions.

PR #1288, opened 2017-04-28
store: collaboration UI for snaps

sergiusens:


snapcraft/_store.py
```diff
@@ -828,16 +825,87 @@ def collaborate(snap_name, key):
...
+
+def _edit_collaborators(developers):
+    """Spawn an editor to modify the snap-developer assertion for a snap."""
+    editor_cmd = os.getenv('EDITOR', 'vi')
+
+    developer_wrapper = {'developers': _reformat_time_from_assertion(
+        developers)}
+
+    with tempfile.NamedTemporaryFile() as ft:
+        ft.close()
```
elopio: why do you close it?

sergiusens: because the `with open` below opens it again. I didn't want any need to flush.

snapcraft/tests/commands/test_collaborate.py
```diff
@@ -84,9 +106,23 @@ def test_collaborate_bad_request(self):
...
         self.assertEqual(
                 'Received error 400: "The given `snap-id` does not match '
                 'the assertion\'s."', str(err))
+
+    def test_collaborate_unchanged_collaborators(self):
+        self.edit_collaborators_mock.return_value = [{
+            'developer-id': 'test-dev-id',
+            'since': '2017-02-10T08:35:00.390258Z',
+            'until': '2018-02-10T08:35:00.390258Z'
+        }]
```
elopio: There's a bug here. _edit_collaborators will always return the times without milliseconds. So we will never hit this case in real life.
The comparison needs to ignore milliseconds.

snapcraft/tests/store/test_store_client.py
```diff
@@ -373,7 +387,21 @@ def test_get_account_information_refreshes_macaroon(self):
...
+                        'snap-id': 'test-snap-id-with-dev',
+                        'status': 'Approved'
+                    },
+                    'test-snap-with-no-validations': {
+                        'price': None,
+                        'private': False,
+                        'since': '2016-12-12T01:01:01Z',
+                        'snap-id': 'test-snap-id-with-no-validations',
+                        'status': 'Approved'
+                    },
```
elopio: These tests are very bad, as they are just checking the fake server. But I won't fix them here.

sergiusens: we cannot really test this with the staging servers

snapcraft/main.py
```diff
@@ -336,6 +336,11 @@ def run(args, project_options):  # noqa
         if _is_containerbuild():
             lifecycle.containerbuild('snap', project_options, args['--output'],
                                      args['<directory>'])
+        elif os.getenv('SNAPCRAFT_COLLABORATE'):
+            # this is only for testing
+            snapcraft.collaborate(
+                os.getenv('SNAPRAFT_COLLABORATORS_SNAP_NAME'),
```
psivaa: I was caught by this SNAPRAFT when I was expecting SNAPCRAFT

snapcraft/tests/store/test_store_client.py
elopio: I know. But the thing is that in unit tests we shouldn't just check the reply from a fake server. We can maybe check that the store client returns a json, and a couple of fields, not all the details of what we hardcoded in the fake.

snapcraft/_store.py
```diff
@@ -828,16 +824,98 @@ def collaborate(snap_name, key):
...
+    reformatted_developers = []
+    for developer in developers:
+        developer_it = {'developer-id': developer['developer-id']}
+        for range_ in ['since', 'until']:
+            if range_ in developer:
+                if developer[range_] == 'now':
+                    date = datetime.now()
+                else:
+                    date = datetime.strptime(developer[range_],
+                                             '%Y-%m-%d %H:%M:%S')
```
psivaa: Was wondering if we could default 00:00:00 when there is no corresponding HH:MM:SS input, to provide easy UE, but that not a huge issue.

snapcraft/_store.py
```diff
@@ -828,16 +824,98 @@ def collaborate(snap_name, key):
     #         }],
     #      }
     # }
+    developers = _edit_collaborators(assertion.get('developers', []))
+    if _are_developers_unchanged(developers, assertion.get('developers', [])):
+        logger.warning('Aborting due to unchanged collaborators list.')
```
sergiusens: a new line is the default

snapcraft/_store.py
sergiusens: the datetime objects will always have that information.

snapcraft/_store.py
psivaa: I meant an additional one. But upto you

snapcraft/_store.py
psivaa: But `developer[range_]` is  a string and you're free to append a ' 00:00:00' to that before creating the datetime object.

snapcraft/_store.py
```diff
@@ -828,16 +824,98 @@ def collaborate(snap_name, key):
...
     assertion['authority-id'] = publisher_id
 
-    assertion = _sign_assertion(snap_name, assertion, key, 'developers')
+    signed_assertion = _sign_assertion(snap_name, assertion, key, 'developers')
+    store.push_assertion(snap_id, signed_assertion, 'developers')
+
 
-    store.push_assertion(snap_id, assertion, 'developers')
+_COLLABORATION_HEADER = dedent("""\
+    # Change which developers may build or upload snaps on the publisher's behalf.
```
elopio: I changed it to push or release.

PR #1292, opened 2017-04-29
tests: fix the recording tests to work in multiple architectures

elopio:


elopio:
@kalikiana we can't run different architectures in travis, only amd64. That's why we have the autopkg runs, but as they are slower we do them only nightly. I caught this because #1291 started failing.

Thanks for the review. On the second run, integration tests passed, which is what we are fixing here. The suite failed building the ros snap, which is sad because it probably means that the arm machines are even slower than before. I'll keep an eye on the beta to see if it happens again.

PR #1293, opened 2017-04-30
recording: record stage packages installed in the snap

elopio:


elopio:
Requires #1292

snapcraft/tests/fixture_setup.py
```diff
@@ -548,3 +548,50 @@ def setUp(self):
             revno = call_with_output(['hg', 'id']).split()[0]
 
             self.commit = revno
+
+
+class FakeAptCache(fixtures.Fixture):
```
elopio: This is the most simple fake to get this working.
With some more time, and patience to read the apt source code we could use a fake /etc/apt directory instead. I think this is good for now.

snapcraft/tests/test_lifecycle.py
```diff
@@ -656,8 +657,42 @@ def test_prime_with_build_info_records_snapcraft_yaml(self):
...
+summary: test
+description: test
+confinement: strict
+grade: stable
+parts:
+  test-part:
+    plugin: nil
+    prime: []
+    stage: []
+    stage-packages: [test-package1=test-version1, test-package2=test-version2]
```
kalikiana: The order in the test results is reversed. By the looks of it, FakeAptCache will have an undefined order since it's using a dict, as is the YAML that is put in a dict by snapcraft internally. I think you should be going through the packages in the test and checking that they're in there without relying on the order.

integration_tests/test_snapcraft_recording.py
```diff
@@ -18,30 +18,190 @@
...
         self.assertEqual(recorded_yaml, source_yaml)
+
+    def test_prime_without_arch_records_current_arch(self):
+        """Test the recorded snapcraft.yaml for a basic snap
+
+        This snap doesn't have stage or build packages and it is not declared
+        that it works on all architectures, which makes it specific to the
+        current architecture.
+        """
+        self.run_snapcraft('prime', project_dir='basic-without-arch')
```
kalikiana: I think you forgot to stage the folder.

snapcraft/tests/test_lifecycle.py
elopio: I changed FakeAptCache to use an OrderedDict. The yaml package is configured in snapcraft/__init__.yaml to also use OrderedDict, so I think this is enough to guarantee the same order always.

integration_tests/test_snapcraft_recording.py
elopio: right, sorry. Added :)

integration_tests/test_snapcraft_recording.py
```diff
@@ -18,30 +18,190 @@
...
+        with open(recorded_yaml_path) as recorded_yaml_file:
+            recorded_yaml = yaml.load(recorded_yaml_file)
+
+        self.assertEqual(recorded_yaml, source_yaml)
+
+    def test_prime_records_stage_packages_version(self):
+        """Test the recorded snapcraft.yaml for a snap with stage packages
+
+        This snap declares all the stage packages that it requires, there are
+        no additional depenencies. The stage packages specify their version.
```
sergiusens: dependencies.

integration_tests/test_snapcraft_recording.py
```diff
@@ -18,30 +18,190 @@
...
+        with open(recorded_yaml_path) as recorded_yaml_file:
+            recorded_yaml = yaml.load(recorded_yaml_file)
+
+        self.assertEqual(recorded_yaml, source_yaml)
+
+    def test_prime_without_stage_packages_version(self):
+        """Test the recorded snapcraft.yaml for a snap with stage packages
+
+        This snap declares all the stage packages that it requires, there are
+        no additional depenencies. The stage packages dont' specify their
```
sergiusens: dependencies

sergiusens: don't or do not

snapcraft/internal/meta.py
```diff
@@ -137,7 +139,17 @@ def _record_snapcraft(self):
...
             os.makedirs(record_dir, exist_ok=True)
             with open(record_file_path, 'w') as record_file:
-                yaml.dump(self._config_data, record_file)
+                annotated_snapcraft = self._annotate_snapcraft(
+                    copy.deepcopy(self._config_data))
+                yaml.dump(annotated_snapcraft, record_file)
+
+    def _annotate_snapcraft(self, data):
+        for part in data['parts']:
+            pull_state = get_state(
```
sergiusens: wouldn't this be better handled in the `state` package? A `get_state(part_name, state)` doing the right thing so that

```python
pull_stage = get_state(part_name, state)
```

IOW, `get_state` shouldn't need exposing paths for the case where these change.

elopio: makes a lot of sense, but that IMO is a different branch because it will touch many other files.

@sergiusens would you like it as a prerequisite of this one, or can I add it to my pile and do it after all the existing ones are merged?

sergiusens: sure, after is fine

elopio: all right, I fixed the typos and merged with master. Thanks for the review.

PR #1294, opened 2017-05-02
meson plugin: Add plugin for meson build system

JulianLiu:
LP: [#1653128](https://bugs.launchpad.net/snapcraft/+bug/1653128)


integration_tests/snaps/meson-hello/snapcraft.yaml
```diff
@@ -0,0 +1,18 @@
+name: meson-hello
+version: 0.1
+summary: test the meson plugin
+description: |
+  This is a basic meson snap. It just prints a hello world.
+  If you want to add other functionalities to this snap, please don't.
+  Make a new one.
+icon: icon.png
```
elopio: Please remove this entry and the .png file.
This optional, and it's good to keep the integration tests as small as possible.

snapcraft/plugins/meson.py
```diff
@@ -0,0 +1,84 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""The meson plugin is useful for building meson based parts.
+
+Meson based projects are projects that have a meson.build that drives the
+build.
+
+This plugin always runs 'meson snapbuild' followed by 'ninja' and
```
elopio: what's snapbuild? is that a thing internal to meson?

snapcraft/plugins/meson.py
```diff
@@ -0,0 +1,84 @@
...
+        return ['meson-parameters']
+
+    def __init__(self, name, options, project):
+        super().__init__(name, options, project)
+        self.snapbuildname = 'snapbuild'
+        self.mesonbuilddir = os.path.join(self.builddir, self.snapbuildname)
+        self.build_packages.append('meson')
+        self.build_packages.append('ninja-build')
+
+    def _meson(self):
```
elopio: I would prefer this to be called _run_meson

snapcraft/plugins/meson.py
```diff
@@ -0,0 +1,84 @@
...
+        os.makedirs(self.mesonbuilddir, exist_ok=True)
+        meson_command = ['meson']
+        if self.options.meson_parameters:
+            meson_command.extend(self.options.meson_parameters)
+        meson_command.append(self.snapbuildname)
+        self.run(meson_command)
+
+    def _ninja(self):
```
elopio: what do you think about splitting this in two? _run_ninja and _run_ninja_install.
It's a nit, but I think the build method would be nice and clear like this:

def build(self):
    super().build()
    self._run_meson()
    self._run_ninja_build_default()
    self._run_ninja_install()

snapcraft/tests/plugins/test_meson.py
```diff
@@ -0,0 +1,118 @@
...
+import snapcraft
+from snapcraft import tests
+from snapcraft.plugins import meson
+
+
+class MesonPluginTestCase(tests.TestCase):
+    """Plugin to provide snapcraft support for the meson build system"""
+
+    def setUp(self):
+        super(MesonPluginTestCase, self).setUp()
```
elopio: No need to define the setUp if it just calls the parent setUp. That's the default action.

integration_tests/snaps/meson-hello/snapcraft.yaml
JulianLiu: Sure.

snapcraft/plugins/meson.py
JulianLiu: 'snapbuild' is just a arbitrary string, any better name will fit :)

snapcraft/plugins/meson.py
JulianLiu: Sure

snapcraft/plugins/meson.py
JulianLiu: Sure, I will refactor here later

snapcraft/tests/plugins/test_meson.py
JulianLiu: I thought this setUp did more than calling parent's setUp here. 😃

snapcraft/plugins/meson.py
```diff
@@ -0,0 +1,87 @@
...
+        env = os.environ.copy()
+        env['DESTDIR'] = self.installdir
+        ninja_install_command = ['ninja', 'install']
+        self.run(ninja_install_command, env=env, cwd=self.mesonbuilddir)
+
+    def build(self):
+        super().build()
+        self._run_meson()
+        self._run_ninja_build_default()
+        self._run_ninja_install()
```
elopio: nice! Thanks. And my last pita comment, copied from #1286 

I love some of the suggestions from the clean code book. One of those is vertical organization, you put first the caller method, and after the called method. This leaves the functions close to were they were called first, but the nicest thing for me is that you can read a source code file from top to bottom. In here if you find _run_meson first you have no context about what it does. If you instead find build first, you will see the context where _run_meson is called and then you can decide if the internals of that method are relevant to what you are looking for, or you can just skip it.

So, can you please reorder the defs like this:
```
def build(self):
    [...]

def _run_meson(self):
    [...]

def _run_ninja_build_default(self):
    [...]

def _run_ninja_install(self):
    [...]
```

JulianLiu: Sure, nice suggestions.

eyelash:
Does this enable optimizations (`--buildtype release`)? I couldn't find anything in the code.

JulianLiu:
You can add it in meson-parameters

PR #1296, opened 2017-05-03
rust snaps can now use source-subdir without failing on pull

roxyd:
LP: [1685563](https://bugs.launchpad.net/snapcraft/+bug/1685563)

### DIFF
--- a/integration_tests/plugins/test_rust_plugin.py
+++ b/integration_tests/plugins/test_rust_plugin.py
@@ -69,7 +69,7 @@ class RustPluginTestCase(RustPluginBaseTestCase):
 
     def test_stage_rust_with_source_subdir(self):
         self.run_snapcraft('stage', 'rust-subdir')
-        
+
         binary_output = self.get_output_ignoring_non_zero_exit(
             os.path.join(self.stage_dir, 'bin', 'rust-subdir'))
         self.assertEqual('Rust in a subdirectory works\n', binary_output)
--- a/integration_tests/plugins/test_rust_plugin.py
+++ b/integration_tests/plugins/test_rust_plugin.py
@@ -89,7 +89,7 @@ class RustPluginTestCase(RustPluginBaseTestCase):
 
         self.assertThat(
             os.path.join('parts', 'rust-subdir', 'src', 'subdir',
-            'Cargo.lock'), FileExists())
+                         'Cargo.lock'), FileExists())
 
 
 class RustPluginConfinementTestCase(testscenarios.WithScenarios,
--- a/integration_tests/plugins/test_rust_plugin.py
+++ b/integration_tests/plugins/test_rust_plugin.py
@@ -69,13 +69,27 @@ class RustPluginTestCase(RustPluginBaseTestCase):
 
     def test_stage_rust_with_source_subdir(self):
         self.run_snapcraft('stage', 'rust-subdir')
-
+        
         binary_output = self.get_output_ignoring_non_zero_exit(
             os.path.join(self.stage_dir, 'bin', 'rust-subdir'))
         self.assertEqual('Rust in a subdirectory works\n', binary_output)
         # Test for bug https://bugs.launchpad.net/snapcraft/+bug/1654764
         self.assertThat('Cargo.lock', Not(FileExists()))
 
+    def test_stage_rust_with_source_and_source_subdir(self):
+        self.copy_project_to_cwd('rust-subdir')
+        with open('snapcraft.yaml') as snapcraft_yaml_file:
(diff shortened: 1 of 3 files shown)


integration_tests/plugins/test_rust_plugin.py
```diff
@@ -69,13 +69,27 @@ def test_stage_rust_plugin_with_conditional_feature(self):
 
     def test_stage_rust_with_source_subdir(self):
         self.run_snapcraft('stage', 'rust-subdir')
-
+
```
sergiusens: can you remove this new empty line please

snapcraft/plugins/rust.py
```diff
@@ -152,6 +152,13 @@ def _fetch_rust(self):
                   '--disable-sudo', '--save'] + options)
 
     def _fetch_deps(self):
+
```
sergiusens: can we please remove this new empty line?

integration_tests/plugins/test_rust_plugin.py
elopio: @roxyd note in here that you replaced an empty line with a line that has spaces.
Our automated static checks will fail here in travis. Please remove the spaces, and leave just the empty line.

snapcraft/plugins/rust.py
elopio: An empty line after a function definition is not very useful, IMO. I would remove it.

snapcraft/plugins/rust.py
```diff
@@ -152,6 +152,13 @@ def _fetch_rust(self):
                   '--disable-sudo', '--save'] + options)
 
     def _fetch_deps(self):
+        
+        source_subdir = getattr(self.options, 'source_subdir', None)
```
sergiusens: This could be 

```python
   sourcedir = os.path.join(self.sourcedir, source_subdir)
else:
    sourcedir = self.sourcedir
 ```

roxyd:
thank you @sergiusens, I've enjoyed contributing!

snapcraft/plugins/rust.py
roxyd: I changed it to:

```     def _fetch_deps(self):
            sourcedir = os.path.join(self.sourcedir, source_subdir)
        else:
            sourcedir = self.sourcedir

        self.run([self._cargo, 'fetch',
                  '--manifest-path',
                  os.path.join(sourcedir, 'Cargo.toml')])

When I did this, it failed a test. Let me see if I can troubleshoot to make this work, as it looks like your way would be more effective

integration_tests/plugins/test_rust_plugin.py
roxyd: removed and pushed, thanks!

snapcraft/plugins/rust.py
roxyd: removed and pushed, thanks!

elopio:
The static tests failed in travis because:
snapcraft/tests/plugins/test_rust.py:179:9: F841 local variable 'rustdir' is assigned to but never used

snapcraft/tests/plugins/test_rust.py:184:80: E501 line too long (94 > 79 characters)

integration_tests/plugins/test_rust_plugin.py:87:1: W293 blank line contains whitespace

integration_tests/plugins/test_rust_plugin.py:89:1: W293 blank line contains whitespace

integration_tests/plugins/test_rust_plugin.py:91:80: E501 line too long (80 > 79 characters)

integration_tests/plugins/test_rust_plugin.py:91:81: W291 trailing whitespace

integration_tests/plugins/test_rust_plugin.py:94:1: E302 expected 2 blank lines, found 1

I love python, but it's high maintenance :D

roxyd:
indeed, it is very specific on whitespace. I will make changes and push later today!

snapcraft/plugins/rust.py
roxyd: I can't really figure out why this isn't working; but perhaps this can be a later code change
it's saying that source_subdir is not defined.
and AttributeError: 'Options' object has no attribute 'source_subdir'

elopio: @roxyd Sergio made a mistake there. On the if block, it should be
`sourcedir = os.path.join(self.sourcedir, self.options.source_subdir)`
I'm not sure if that would solve the error you are getting though.

roxyd: it still says that source_subdir is not defined. I think we have to have the statement above it that defines it

roxyd: @sergiusens , @elopio showed me how to make it work so now source_subdir is part of the options class and this suggestion works! I think it makes the code even better, because self.options.source_subdir can be used in the future if there's any other use for it. thank you!

PR #924, opened 2016-11-22
New plugin: jhbuild

attente:
LP: #1643924

PR #1298, opened 2017-05-04
jhbuild plugin: new plugin

diddledan:
LP: #1643924
Previous PR: #924 

The previous PR was closed due to the attempt relying on a package which isn't available in Xenial. I have taken that PR (props to @attente for the original work, which I have based this PR on) and removed the dependency on Bubblewrap. I then massaged the code to simplify and unify duplicated code. The tests pass on my Xenial LXC container and also pass Travis-CI's suite.


sergiusens:
Nice work, I am not fluent in dhbuild but will find someone who is to do a first pass on this and see if it is all good.

Thanks!

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,773 @@
...
+                                 os.path.join(self.installdir, 'usr'))
+            jhbuildrc_file.write('use_local_modulesets = True\n')
+            jhbuildrc_file.write('skip = [%s]\n' % ', '.join([
+                '\'%s\'' % module
+                for module in self.skip_modules
+            ]))
+            jhbuildrc_file.write(
+                'module_autogenargs[\'gdk-pixbuf\'] = '
+                '\'--disable-gio-sniffing\'\n'
+            )
```
jhenstridge: This might be a bit easier if you specify the whole configuration in a single go, and use the `str.format()` method to substitute in named parameters.  Something like:

    config = '''
    moduleset = {module_set!r}
    ...
    '''
    jhbuildrc_file.write(config.format(module_set=self.options.module_set, ...))

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,773 @@
...
+control. See the JHBuild manual for more details:
+https://developer.gnome.org/jhbuild/stable/.
+
+The plugin can be customized with the following keys:
+
+  - modules: (required) a list of modules to include in the part, modules that
+             should be skipped should be prefixed with '-', e.g. '-WebKit'
+  - module-set: (default: gnome-world) the module set JHBuild should use
+  - module-set-dir: the directory where the module set is located
+  - snap-revision: the revision of the snap this part is included in
```
attente: You mean the store can potentially overwrite the revision of the uploaded snap to something other than what the developer specified in the snapcraft.yaml? That's really a big problem for this commit then because it's making a pretty big assumption that the location of /snap/<snap-name>/<snap-revision> is fixed.

attente:
The dependency magic and exception list could probably be removed in favour of what's already in JHBuild, but that would mean staging -dev packages, and at the time, I wanted to minimize what went into the final snap. But yeah, really the added complexity probably wasn't worth the effort.

jhenstridge:
With how Snapcraft works right now, you can add `libfoo-dev` to `build-packages` for your snap, which might also install a `libfoo42` package on the build system.  If the executable the snap primes depends on `libfoo.so.42` but that library hasn't been staged, it will copy the library from the system into the snap.

For many library dependencies, this is sufficient: it's often only necessary to manually stage a library package if the library depends on additional data files found in that package.

That's why I was asking how much of the dependency resolution magic in your plugin was necessary.

snapcraft/plugins/jhbuild.py
jhenstridge: The currently active revision of the snap will be available at `/snap/$snap_name/current` if you need a fixed path.  Although using the `$SNAP` environment variable at runtime is better if possible.

attente:
I'm not sure if we're on the same page here, or if I'm fundamentally misunderstanding some magic snapcraft might be doing. What I meant was that JHBuild would end up putting `libfoo-dev` into the snap when really the snap only needs `libfoo42` at runtime. But anyways, I think you are right that all the magic is really not necessary since the gains are fairly small compared to the complexity introduced.

jhenstridge:
What I'm saying is that in many cases it isn't necessary to stage either `libfoo42` or `libfoo-dev`.  If I put the following in my snapcraft project:

    build-packages:
     - libfoo-dev

... then the shared library from `libfoo42` will be copied into my snap if any executables or libraries depend on it.  There is no need to list either package via `stage-packages`.

attente:
That kind of magic worries me though. Copying a shared library into the snap doesn't seem as complete as fulfilling a runtime dependency. What happens if `libfoo42` depends on `libfoo-data` which provides something like a gsettings schema or some other required non-shared-library file? Is that case covered?

diddledan:
I've got an updated version of this plugin which addresses some of the raised points. I'll be uploading soon (hopefully) to get further feedback

elopio:
Thanks for the update @diddledan. @attente, @jhenstridge: could you please review it again?

integration_tests/plugins/test_jhbuild_plugin.py
```diff
@@ -0,0 +1,97 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2016 Canonical Ltd
```
elopio: The year should be 2017.

integration_tests/plugins/test_jhbuild_plugin.py
```diff
@@ -0,0 +1,97 @@
...
+            self.handle_request()
+
+
+class JHBuildPluginTestCase(integration_tests.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        cls.port = 8000
+        handler = StoppableHTTPRequestHandler
+        httpd = StoppableHTTPServer(("127.0.0.1", cls.port), handler)
```
elopio: another nit, please use single quotes for strings, for consistency with the rest of the code.

integration_tests/plugins/test_jhbuild_plugin.py
```diff
@@ -0,0 +1,97 @@
...
+        cls.server = threading.Thread(target=httpd.serve_forever)
+        cls.server.setDaemon(True)
+        cls.server.start()
+
+    @classmethod
+    def tearDownClass(cls):
+        conn = http.client.HTTPConnection("127.0.0.1:%d" % cls.port)
```
elopio: Single quotes, and please use `.format` instead of `%`.

integration_tests/plugins/test_jhbuild_plugin.py
```diff
@@ -0,0 +1,97 @@
...
+        conn.request("QUIT", "/")
```
elopio: single quotes :)

integration_tests/plugins/test_jhbuild_plugin.py
```diff
@@ -0,0 +1,97 @@
...
+
+import http.client
+import http.server
+import glob
+import os
+import threading
+import integration_tests
+
+
+class StoppableHTTPRequestHandler(http.server.SimpleHTTPRequestHandler):
```
elopio: Take a look at how we deal with fake servers in fixtures:
https://github.com/snapcore/snapcraft/blob/master/snapcraft/tests/fixture_setup.py#L253

You can reuse some things there, instead of implementing the quit in tear down yourself.

integration_tests/plugins/test_jhbuild_plugin.py
```diff
@@ -0,0 +1,97 @@
...
+
+    @classmethod
+    def tearDownClass(cls):
+        conn = http.client.HTTPConnection("127.0.0.1:%d" % cls.port)
+        conn.request("QUIT", "/")
+        conn.getresponse()
+        cls.server.join()
+
+    def test_snap(self):
+
```
elopio: Please remove this empty line.

integration_tests/plugins/test_jhbuild_plugin.py
```diff
@@ -0,0 +1,97 @@
...
+            ],
+            'snap': [],
+        }
+
+        for stage in ['pull', 'build', 'stage', 'prime', 'snap']:
+            self.run_snapcraft(stage, 'jhbuild')
+
+            for path in files[stage]:
+                self.assertTrue(os.path.exists(path),
+                                "path '%s' does not exist" % path)
```
elopio: In here it's ok to use double quotes, because you have a single quote inside. But please use .format.
Also, you can use the testtools assertion: http://testtools.readthedocs.io/en/latest/for-test-authors.html?highlight=fileexists#fileexists 

I usually don't like fors in tests, but I like this one! :D

integration_tests/snaps/jhbuild/simple-jhbuild.modules
```diff
@@ -0,0 +1,23 @@
...
+  <repository name="system" type="system"/>
+  <repository name="simple-jhbuild" type="tarball" href="http://localhost:8000/" default="yes"/>
+  <systemmodule id="libtool">
+    <branch repo="system"/>
+    <systemdependencies>
+      <dep type="path" name="libtoolize"/>
+    </systemdependencies>
+  </systemmodule>
+  <autotools id="dep-module" autogen-sh="autoreconf">
+    <branch version="1.0.0" module="dep-module.tar.bz2" checkoutdir="dep-module"/>
```
elopio: It would be nice to explain what are these binary files and how you generated them.

integration_tests/snaps/jhbuild/snapcraft.yaml
```diff
@@ -0,0 +1,24 @@
+name: test-jhbuild
+version: 1.0.0
+grade: stable
+confinement: strict
+summary: Simple JHBuild integration test
+description: >
```
elopio: Please replace this `>` with `|`.

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,383 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2016 Canonical Ltd
```
elopio: 2017.

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,383 @@
...
+https://developer.gnome.org/jhbuild/stable/.
+
+The plugin can be customized with the following keys:
+
+  - modules: (required) a list of modules to include in the part, modules that
+             should be skipped should be prefixed with '-', e.g. '-WebKit'
+  - module-set: (default: gnome-world) the module set JHBuild should use
+  - module-set-dir: the directory where the module set is located
+               Either leave unset to use the default moduleset definitions, or
+             set to `.` to reference your project folder.
```
elopio: There is a weird alignment here.

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,383 @@
...
+  - jhbuild-archive: the source tarball directory on the build host
+  - jhbuild-mirror: the DVCS repository directory on the build host
+
+Advice:
```
elopio: This is a cool thing to add, I think!
What do you think @kyrofa @kalikiana ?

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,383 @@
...
+"""
+
+import glob
+import logging
+import os
+import subprocess
+import snapcraft
+from snapcraft.internal import common
+
+HOME = os.path.expanduser('~') if os.geteuid() else os.path.join('/', 'root')
```
elopio: I'm not sure about why this is needed. Wouldn't os.environ.get('HOME') work for both root and non-root?

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,383 @@
...
+    'python-dev',
+    'python3-dev',
+    'subversion',
+    'symlinks',
+    'yelp-tools',
+    'zlib1g-dev',
+}
+
+
+LOG = logging.getLogger(__name__)
```
elopio: we call this variable `logger` everywhere else in the code.

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,383 @@
...
+
+
+def _get_jhbuild_user():
```
elopio: a docstring in this function would be useful, because it's not clear for me why it's needed.

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,383 @@
...
+    myuid = os.getuid()
+    if myuid != 0:
+        return myuid
```
elopio: we don't prefix variables with `my`. `uid` seems like a good name for me here.

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,383 @@
...
+
+    try:
+        uid = common.run_output(['id', '-u', 'jhbuild'])
+    except subprocess.CalledProcessError:
+        uid = None
+    return uid
+
+
+def _get_jhbuild_group():
```
elopio: as on the other, a docstring would be appreciated. and `gid` instead of `mygid`.

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,383 @@
...
+            },
+            'jhbuild-archive': {
+                'type': 'string',
+                'default': '',
+            },
+            'jhbuild-mirror': {
+                'type': 'string',
+                'default': '',
+            },
+            'cflags': {
```
elopio: cflags is not explained in the plugin help.

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,383 @@
...
+            module[1:] for module in self.options.modules if module[0] == '-']
+
+        self.build_packages += list(BUILD_PACKAGES)
+
+        self.jhbuild_src = os.path.join(self.partdir, 'jhbuild', 'src')
+        self.jhbuild_program = os.path.join(
+            self.partdir, 'jhbuild', 'usr', 'bin', 'jhbuild')
+        self.jhbuildrc_path = os.path.join(self.partdir, 'jhbuildrc')
+
+    def enable_cross_compilation(self):
```
elopio: This is the default from the base class, so no need to define it in this plugin.

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,383 @@
...
+        gid = common.run_output(['id', '-g', 'jhbuild'])
+    except subprocess.CalledProcessError:
+        gid = 65534
+    return gid
+
+
+class JHBuildPlugin(snapcraft.BasePlugin):
+    """
+    JHBuildPlugin provides jhbuild functionality to snapcraft
+    """
```
elopio: Another nit, we use a single line when the docstrings are a single line. No need for the `"""` in their own lines.

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,383 @@
...
+
+    def enable_cross_compilation(self):
+        """
+        We don't support cross compilation (yet?)
+        """
+        raise NotImplementedError(
+            'The {!s} plugin does not support cross compilation'.format(
+                self.name))
+
+    def run(self, cmd, cwd=None, **kwargs):
```
elopio: Why are you defining run and run_output only to call the definitions from the supper class? It would be the same to just remove them from this class.

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,383 @@
...
+        stdout.
+
+        :param list cmd: command arguments, first is the executable
+        :param str cwd: working directory
+        :return: the output of the command
+        :rtype: str
+        """
+        return super().run_output(cmd, cwd=cwd, **kwargs)
+
+    def jhbuild(self, args, output=True, **kwargs):
```
elopio: I think that a better name here would be run_jhbuild.

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,383 @@
...
+        """Run JHBuild in the build stage.
```
elopio: instead of stage, we call them steps.

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,383 @@
...
+
+        :param list args: command arguments without executable
+        :return: the output of the command if captured
+        :rtype: str
+        """
+
```
elopio: No need for this empty line.

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,383 @@
...
+    def _pull_jhbuild(self):
+        LOG.info('Pulling JHBuild')
+
+        repository = 'https://git.gnome.org/browse/jhbuild'
+
+        if self.options.jhbuild_mirror:
+            repository = self.options.jhbuild_mirror
+
+        if os.path.isdir(self.jhbuild_src):
```
elopio: is the jhbuild_src dir always a git repository? And why do we need it in our own directory instead of calling it from the directory defined in the property?

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,383 @@
...
+                                   self.skip_modules]),
+                              extra_prefixes=', '.join(
+                                  ['\'%s\'' % prefix for prefix in
+                                   extra_prefixes]),
+                              cflags=self.options.cflags,
+                              ))
+
+    def pull(self):
+        self._pull_jhbuild()
+        self._setup_jhbuild()
```
elopio: It's nicer when you are able to read the source code from top to bottom. For that, please move the definition of _pull_jhbuild, _setup_jhbuild and jhbuild after this pull method.
Then somebody new to the code will find first the pull method, and if he wants to dig further, he will also read the definitions of the other 3 methods. If he only wants a high level overview, he will just read pull. With the current implementation he will find first the definition of setup_jhbuild with no context to know if it's an implementation detail, or something critical to understand what the plugin does.

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,383 @@
...
+
+        self.jhbuild(['sanitycheck'], output=False)
+
+        modules = self.jhbuild(['list'] + self.modules,
+                               output=True).splitlines()
+
+        if 'gtk+' in modules or 'gtk+-3' in modules:
+            self.modules.append('adwaita-icon-theme')
+            self.stage_packages.append('ttf-ubuntu-font-family')
```
elopio: Could you please explain this with a comment in the source code?

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,383 @@
...
+
+    def build(self):
+        self._setup_jhbuild()
+
+        LOG.info('Building modules')
+        self.jhbuild(['build'] + self.modules,
+                     cwd=self.builddir, output=False)
+
+        LOG.info('Fixing symbolic links')
+        self.run(['symlinks', '-c', '-d', '-r', '-s', '-v', self.installdir])
```
elopio: could you please explain with a comment in the source code why this is needed?

snapcraft/tests/plugins/test_jhbuild.py
```diff
@@ -0,0 +1,53 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2016 Canonical Ltd
```
elopio: 2017

kyrofa:
> That kind of magic worries me though. Copying a shared library into the snap doesn't seem as complete as fulfilling a runtime dependency. What happens if libfoo42 depends on libfoo-data which provides something like a gsettings schema or some other required non-shared-library file? Is that case covered?

Indeed, this is why Snapcraft 3.x will stop the magic: it's not clear at what point the magic stops! Currently snapcraft literally runs `ldd` on every ELF file in the snap, pulling libraries from the system as necessary. That's it. It will _not_ work for anything other than a shared library (e.g. your gsettings example would not work).

snapcraft/plugins/jhbuild.py
kyrofa: Indeed, "stage" is actually a step :stuck_out_tongue: . Terminology minefield here.

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,383 @@
...
+            cmd = ['sudo', '-H', '-u', 'jhbuild', '--']
+
+        cmd = cmd + [self.jhbuild_program,
+                     '--no-interact', '-f', self.jhbuildrc_path]
+
+        run = self.run_output if output else self.run
+        return run(cmd + args, cwd=cwd, **kwargs)
+
+    def _create_jhbuild_user(self):
+        common.run(['adduser', '--shell', '/bin/bash', '--disabled-password',
```
kyrofa: Does this potentially require sudo, then? Might be worth mentioning in the docs at the top. In fact, it seems like a run may or may not fail depending on whether or not it was run with sudo and whether or not that user already exists. Should we instead _use_ sudo, here, thus prompting for a password if necessary?

I don't like creating users under the covers like this, by the way, but I assume it's required for jhbuild.

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,383 @@
...
+        for path in paths:
+            for filename in glob.iglob('%s/**' % path, recursive=True):
+                if not os.path.exists(filename):
+                    continue
+                os.chown(filename, jhbuild_user, jhbuild_group)
+
+    def _pull_jhbuild(self):
+        LOG.info('Pulling JHBuild')
+
+        repository = 'https://git.gnome.org/browse/jhbuild'
```
kyrofa: Can we define this as a constant at the top, please?

snapcraft/plugins/jhbuild.py
kyrofa: Whoa, cool tool, we have a lot of symlink logic elsewhere, just in Python-- I wonder if we should start using this instead. @elopio, thoughts?

.vscode/launch.json
```diff
@@ -1,33 +0,0 @@
...
-            "program": "${workspaceRoot}/units.py",
-            "cwd": "${workspaceRoot}",
-            "debugOptions": [
-                "WaitOnAbnormalExit",
-                "WaitOnNormalExit",
-                "RedirectOutput"
-            ]
-        }
-    ]
-}
```
diddledan: I removed that because I figured I had accidentally created it and thus it only had stuff relevant for me. I'll restore it.

snapcraft/plugins/jhbuild.py
diddledan: I'll remove the usage of symlinks on the premise that snapcraft will "do the right thing" - that might mean including the symlinks command elsewhere in snapcraft.

snapcraft/plugins/jhbuild.py
diddledan: jhbuild refuses to run under the root account, so we need to `sudo` to an unprivileged user when cleanbuilding. I think we can limit the `sudo` and `adduser` calls to only running when the current euid is 0.

snapcraft/plugins/jhbuild.py
```diff
@@ -0,0 +1,383 @@
...
+  - jhbuild-mirror: the DVCS repository directory on the build host
+
+Advice:
+
+  - You only need to specify a single module in most cases. JHBuild knows the
+    module dependency graph.
+
+  - The desktop-gtk3 helper is incompatible with GTK built by jhbuild. A custom
+    helper is currently required.
+    See github.com/diddledan/corebird for example.
```
diddledan: The problem is that the `desktop-launch` script which gets built by `desktop-gtk3` requires multiple changes to work around the different installation locations of ubuntu's packages vs jhbuild-built packages.

snapcraft/plugins/jhbuild.py
kyrofa: That's not what I meant by my comment-- you may very well need to fix these up. @elopio was just wanting a comment in the code as to why.

snapcraft/plugins/jhbuild.py
diddledan: there is now documentation for the cflags parameter.

elopio:
Thanks for all the changes @diddledan. I changed my review to +1, and now we only need to wait for @jhenstridge to approve this.

jhenstridge:
Sorry for the delays.  This looks like it corrects the main problems I had with the initial versions of the plugin (shipping a diff against jhbuild itself, and creating a new system for discovering dependencies specific to the plugin).  So I've got no objections now.

PR #1295, opened 2017-05-03
recording: record build packages installed in the snap

elopio:

PR #1299, opened 2017-05-04
asset-tracking: save the dependencies of build packages

elopio:
LP: [#1688151](https://bugs.launchpad.net/snapcraft/+bug/1688151)

### DIFF
--- a/integration_tests/test_snapcraft_recording.py
+++ b/integration_tests/test_snapcraft_recording.py
@@ -144,4 +144,6 @@ class SnapcraftRecordingPackagesTestCase(
         with open(recorded_yaml_path) as recorded_yaml_file:
             recorded_yaml = yaml.load(recorded_yaml_file)
 
-        self.assertEqual(recorded_yaml, source_yaml)
+        self.assertEqual(
+            recorded_yaml['parts'][part_name][self.packages_type],
+            expected_stage_packages)
--- a/integration_tests/test_snapcraft_recording.py
+++ b/integration_tests/test_snapcraft_recording.py
@@ -132,7 +132,6 @@ class SnapcraftRecordingPackagesTestCase(
         with open(os.path.join('snap', 'snapcraft.yaml')) as source_yaml_file:
             source_yaml = yaml.load(source_yaml_file)
         part_name = 'part-with-{}'.format(self.packages_type)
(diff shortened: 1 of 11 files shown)


elopio:
Requires #1295

sergiusens:
this now has conflicts

elopio:
@sergiusens conflicts resolved.

integration_tests/test_snapcraft_recording.py
```diff
@@ -45,6 +45,7 @@ def test_prime_with_architectures(self):
         This snap doesn't have stage or build packages and is declared that it
         works on all architectures.
         """
+
```
sergiusens: you used to tell me to get rid of these!

snapcraft/internal/repo/_deb.py
```diff
@@ -229,23 +229,31 @@ def install_build_packages(cls, package_names):
...
-        return pkg_list
+                if str(installed_package) not in installed_packages:
+                    installed_packages.append(str(installed_package))
+                    for depends in installed_package.get_dependencies(
+                            'Depends'):
+                        # deps is a list of or dependencies. We are taking
+                        # the first one that satisfies the dependency, which
+                        # might or might not be problematic.
+                        # --elopio - 20170504
+                        build_packages.append(depends[0].name)
```
sergiusens: popping and appending, and losing history might lead to an eternal loop, will it not?

integration_tests/test_snapcraft_recording.py
elopio: When the name of the test is clear, they are not needed. But here I couldn't fit good names in 80 chars.
The names are confusing, because with architecture could mean that the source has arch, or that the record has arch. Without dependencies could mean that the packages have no dependencies, or that all the dependencies are declared. I welcome suggestions for better test names.

elopio: ok, empty line removed. Sorry I misunderstood.

snapcraft/internal/repo/_deb.py
elopio: I think not, because installed_packages is a list of already visited nodes. When we find a node that's already in installed, we ignore it and don't add its dependencies to the list of nodes to visit.

PR #1300, opened 2017-05-06
tools: add a script to install autopkgtest dependencies

elopio:


tools/install_autopkgtest_dependencies.py
```diff
@@ -0,0 +1,49 @@
...
+from debian import deb822
+
+
+def main():
+    if len(sys.argv) != 2:
+        sys.exit('Usage: {} test_name'.format(sys.argv[0]))
+    test_name = sys.argv[1]
+    dependencies = get_dependencies(
+        os.path.join('debian', 'tests', 'control'), test_name)
+    subprocess.check_call(['sudo', 'apt', 'install'] + dependencies)
```
sergiusens: Can you first check which ones are missing from an installation in order to then run

`sudo apt-mark auto $packages`?

tools/install_autopkgtest_dependencies.py
```diff
@@ -0,0 +1,49 @@
...
+
+from debian import deb822
+
+
+def main():
+    if len(sys.argv) != 2:
+        sys.exit('Usage: {} test_name'.format(sys.argv[0]))
+    test_name = sys.argv[1]
+    dependencies = get_dependencies(
+        os.path.join('debian', 'tests', 'control'), test_name)
```
sergiusens: this path should contemplate the path of this script using `__file__`

tools/install_autopkgtest_dependencies.py
elopio: I added a check with dpkg -l. If you know a better way, let me know.

tools/install_autopkgtest_dependencies.py
elopio: done.

elopio:
Thanks for the review @sergiusens!

sergiusens:
so one final question, why would this be useful? Maybe document it on HACKING.md

elopio:
@sergiusens that's where I want to get. I will make a branch that adds steps on HACKING.md to run the tests using the travis docker (after #1318), using an autopkgtest runner and using debs locally (where this script is useful). On #1311 are also instructions to install from pip, but I don't like that option. Should we include it?

sergiusens:
On Tue, May 16, 2017 at 12:17 PM, Leo Arias <notifications@github.com>
wrote:

> @sergiusens <https://github.com/sergiusens> that's where I want to get. I
> will make a branch that adds steps on HACKING.md to run the tests using the
> travis docker (after #1318
> <https://github.com/snapcore/snapcraft/pull/1318>), using an autopkgtest
> runner and using debs locally (where this script is useful). On #1311
> <https://github.com/snapcore/snapcraft/pull/1311> are also instructions
> to install from pip, but I don't like that option. Should we include it?
>
No, for development we keep pip, we moved to it for a reason!

PR #1301, opened 2017-05-06
recording: record global build-packages installed on the host

elopio:


elopio:
Requires #1299

elopio:
These conflicts are too much for me.

PR #1302, opened 2017-05-09
lxd: mount project folder via sshfs in case of a remote

### LINKED ISSUE #1457: remote per-project container
#1302 [design](https://forum.snapcraft.io/t/remote-per-project-containers/535/1)

kalikiana:
Serve project folder to remote LXD containers via sftp.

Fixes  #1457


snapcraft/internal/lifecycle.py
```diff
@@ -326,8 +326,12 @@ def _tar_filter(tarinfo):
 
 def containerbuild(step, project_options, output=None, args=[]):
     config = snapcraft.internal.load_config(project_options)
+    remote = os.environ.get('SNAPCRAFT_CONTAINER_BUILDS')
+    if remote in ['1', '0', None, '']:
+        remote = None
```
elopio: I find it weird that when the env var is set to 1, then it's the same as not setting it.
Could you raise an error if an invalid value is set?

kalikiana: You're misunderstanding the code. This is so you can do SNAPCRAFT_CONTAINER_BUILDS=myremote to specify a remote or SNAPCRAFT_CONTAINER_BUILDS=1 to use the default (local) remote. There's no such thing is an invalid value here.

elopio: so, 1, 0 and None mean local?

kalikiana: Yes. snapcraft/main.py has the separate check for whether to use containers or not.

kalikiana: Note that I'm specifically sticking to the environment variable here to avoid exposing it in the CLI - we might consider to at some point use the --remote argument for all commands that can be run in a container.

elopio: ok, I get it now. Thanks for the explanation.
I think a comment in the code explaining this wouldn't hurt.

snapcraft/internal/lxd.py
```diff
@@ -183,18 +197,69 @@ def _ensure_container(self):
...
 
+    def _container_login(self):
+        network = self._get_container_status()['state']['network']['eth0']
+        for address in network['addresses']:
+            if address['family'] == 'inet':
+                return address['address'], '22'
+        raise RuntimeError('No IP found for {}'.format(self._container_name))
+
+    def _background(self, cmd, stdin=None, stdout=None):
+        self._processes += [Popen(cmd, stdin=stdin, stdout=stdout)]
```
elopio: remember to put these after the place where they are called.

snapcraft/internal/lxd.py
```diff
@@ -183,18 +197,69 @@ def _ensure_container(self):
...
     def _ensure_mount(self, destination, source):
         logger.info('Mounting {} into container'.format(source))
-        devices = self._get_container_status()['devices']
-        if destination not in devices:
-            check_call([
-                'lxc', 'config', 'device', 'add', self._container_name,
-                destination, 'disk', 'source={}'.format(source),
-                'path=/{}'.format(destination)])
+        remote, container_name = self._container_name.split(':')
+        if remote != 'local':
```
elopio: This is a big if block. Can you put it inside a method?

snapcraft/internal/lxd.py
```diff
@@ -183,18 +197,69 @@ def _ensure_container(self):
...
+            if destination in devices:
+                check_call([
+                    'lxc', 'config', 'device', 'remove', self._container_name,
+                    destination, 'disk', 'source={}'.format(source),
+                    'path=/{}'.format(destination)])
+
+            self._install_packages(['sshfs'])
+            keyfile = 'id_{}'.format(self._container_name)
+            if not os.path.exists(keyfile):
+                check_output(['ssh-keygen', '-o', '-N', '', '-f', keyfile])
```
elopio: if you are not using the output, you can use check_call instead.

snapcraft/internal/lxd.py
```diff
@@ -183,18 +197,69 @@ def _ensure_container(self):
...
+            self._container_run(['chmod', '600',
+                                 '/root/.ssh/authorized_keys'])
+            self._container_run(['mkdir', '-p', '/{}'.format(destination)])
+            self._container_run(['mkdir', '-p', source])
+            r1, w1 = os.pipe()
+            r2, w2 = os.pipe()
+            ssh_hostname, ssh_port = self._container_login()
+            logger.info('Connecting via SSH to {}:{}'.format(
+                ssh_hostname, ssh_port))
+            self._background(['/usr/lib/sftp-server'], stdin=r1, stdout=w2)
```
elopio: what do you think about calling this _container_run_in_background, or something like that?

snapcraft/internal/lxd.py
```diff
@@ -183,18 +197,69 @@ def _ensure_container(self):
...
+            self._container_run(['chmod', '700', '/root/.ssh'])
+            self._container_run(['tee', '-a',
+                                 '/root/.ssh/authorized_keys'],
+                                stdin=open('{}.pub'.format(keyfile), 'r'))
+            self._container_run(['chmod', '600',
+                                 '/root/.ssh/authorized_keys'])
+            self._container_run(['mkdir', '-p', '/{}'.format(destination)])
+            self._container_run(['mkdir', '-p', source])
+            r1, w1 = os.pipe()
+            r2, w2 = os.pipe()
```
elopio: instead of r1 and w2, why not calling this stdin and stdout?
And when you don't use variables, you use _ instead of the name.
`stdin, _ = os.pipe()`

snapcraft/internal/lxd.py
```diff
@@ -183,18 +197,69 @@ def _ensure_container(self):
     def _setup_project(self):
         self._ensure_mount(self._project_folder, self._source)
 
+    def _container_login(self):
```
elopio: I read this thinking that it performed a login. It would be clearer named like _get_container_login.
I would personally call it _get_container_address, but I can see how it could be confusing because of the network address.

snapcraft/internal/lxd.py
```diff
@@ -183,18 +197,69 @@ def _ensure_container(self):
...
+                    'lxc', 'config', 'device', 'remove', self._container_name,
+                    destination, 'disk', 'source={}'.format(source),
+                    'path=/{}'.format(destination)])
+
+            self._install_packages(['sshfs'])
+            keyfile = 'id_{}'.format(self._container_name)
+            if not os.path.exists(keyfile):
+                check_output(['ssh-keygen', '-o', '-N', '', '-f', keyfile])
+            self._container_run(['mkdir', '-p', '/root/.ssh'])
+            self._container_run(['chmod', '700', '/root/.ssh'])
```
elopio: It would be nice to have comments for why these are needed.

sergiusens:
On Mon, May 15, 2017 at 5:26 PM, Leo Arias <notifications@github.com> wrote:

> *@elopio* commented on this pull request.
> ------------------------------
>
> In snapcraft/internal/lxd.py
> <https://github.com/snapcore/snapcraft/pull/1302#discussion_r116592157>:
>
> > -                destination, 'disk', 'source={}'.format(source),
> -                'path=/{}'.format(destination)])
> +        remote, container_name = self._container_name.split(':')
> +        if remote != 'local':
> +            # Remove project folder in case it was used "locally" before
> +            devices = self._get_container_status()['devices']
> +            if destination in devices:
> +                check_call([
> +                    'lxc', 'config', 'device', 'remove', self._container_name,
> +                    destination, 'disk', 'source={}'.format(source),
> +                    'path=/{}'.format(destination)])
> +
> +            self._install_packages(['sshfs'])
> +            keyfile = 'id_{}'.format(self._container_name)
> +            if not os.path.exists(keyfile):
> +                check_output(['ssh-keygen', '-o', '-N', '', '-f', keyfile])
>
> if you are not using the output, you can use check_call instead.
>
Unless you want to output easily hidden.

debian/control
```diff
@@ -62,6 +62,7 @@ Depends: python3-apt,
          python3-xdg,
          python3-yaml,
          squashfs-tools,
+         openssh-sftp-server,
```
sergiusens: only thing I can add here is that I would dynamically install this instead

kalikiana: That would effectively make project/cleanbuild deb-specific at runtime. Any particular reason to special-case sftp? Is there precedent for lazily installing packages in other cases?

snapcraft/internal/lxd.py
```diff
@@ -80,10 +81,14 @@ def _pull_file(self, src, dst):
         check_call(['lxc', 'file', 'pull',
                     '{}/{}'.format(self._container_name, src), dst])
 
-    def _container_run(self, cmd):
+    def _container_run(self, cmd, **kwargs):
         # Set HOME here as 'lxc config set ... environment.HOME' doesn't work
-        check_call(['lxc', 'exec', self._container_name, '--env',
-                   'HOME=/{}'.format(self._project_folder), '--'] + cmd)
+        # Use 'cd' because --env has no effect with sshfs mounts
```
kyrofa: This doesn't actually set `$HOME` at all though, right? Does it need to be set? Or does the comment above need to be removed?

snapcraft/internal/lxd.py
```diff
@@ -106,12 +111,19 @@ def _ensure_started(self):
...
             check_call(['lxc', 'stop', '-f', self._container_name])
 
+    def _install_packages(self, packages):
+        try:
+            self._wait_for_network()
+            self._container_run(['apt-get', 'update'])
+            self._container_run(['apt-get', 'install', '-y', *packages])
+        finally:
+            # Always remove apt lock in case we stop during install
+            self._container_run(['rm', '-f', '/var/lib/apt/lists/lock'])
```
kyrofa: Apt doesn't do this?

snapcraft/internal/lxd.py
```diff
@@ -204,18 +218,83 @@ def _ensure_container(self):
     def _setup_project(self):
         self._ensure_mount(self._project_folder, self._source)
 
+    def _get_container_address(self):
+        network = self._get_container_status()['state']['network']['eth0']
```
kyrofa: Not an enormous deal, but this makes me a little nervous. We're using the default profile, right? What if the user changes that profile to configure their network devices differently? This device may not even exist, or may be an interface with no external network access.

snapcraft/internal/lxd.py
```diff
@@ -204,18 +218,83 @@ def _ensure_container(self):
...
         devices = self._get_container_status()['devices']
-        if destination not in devices:
+        if destination in devices:
             check_call([
-                'lxc', 'config', 'device', 'add', self._container_name,
+                'lxc', 'config', 'device', 'remove', self._container_name,
                 destination, 'disk', 'source={}'.format(source),
                 'path=/{}'.format(destination)])
 
+        # Generate an SSH key and add it to the container's known keys
```
kyrofa: The diff isn't being helpful here: where would this keyfile end up? Would it be useful to default to `~/.ssh/id_rsa.pub` if it already exists rather than dumping random keys on disk?

snapcraft/internal/lxd.py
```diff
@@ -204,18 +218,83 @@ def _ensure_container(self):
...
+        except CalledProcessError as e:
+            # XXX: This needs to be extended once we support other distros
+            raise SnapcraftEnvironmentError(
+                'sftp-server could not be run.\n'
+                'On Debian, Ubuntu and derivatives, the package '
+                'openssh-sftp-server needs to be installed.'
+                ) from e
+        self._host_run(['ssh', '-C', '-F', '/dev/null',
+                        '-o', 'IdentityFile={}'.format(keyfile),
+                        '-o', 'StrictHostKeyChecking=no',
```
kyrofa: We have a handle to the container. Would it be a good idea to obtain the host key so we don't have to disable checking?

snapcraft/internal/lxd.py
```diff
@@ -204,18 +218,83 @@ def _ensure_container(self):
...
+                             '/root/.ssh/authorized_keys'])
+
+        # Use sshfs in slave mode inside SSH to reverse mount destination
+        self._install_packages(['sshfs'])
+        self._container_run(['mkdir', '-p', '/{}'.format(destination)])
+        self._container_run(['mkdir', '-p', source])
+        ssh_address = self._get_container_address()
+        logger.info('Connecting via SSH to {}'.format(ssh_address))
+        stdin1, stdout1 = os.pipe()
+        stdin2, stdout2 = os.pipe()
```
kyrofa: These appear to never be used, no? Why are we using Popen at all?

snapcraft/internal/lxd.py
kalikiana: That's why this line is there, yes. Not sure how to make this more clear.

snapcraft/internal/lxd.py
kalikiana: There's no need for the user to manage the key. We don't even want them to SSH into the container. So I think that's rather the opposite of what we want.

snapcraft/internal/lxd.py
kalikiana: Note the following two invokations of _host_run to see where and how they are used.

snapcraft/internal/lxd.py
elopio: It was clear for me. I think @kyrofa needs to ctrl+c in the middle of an update, to see how things go crazy and non-user-friendly.

snapcraft/internal/lxd.py
elopio: That was my same thinking when I first reviewed it. Maybe this means that a comment in the code could be useful.

elopio:
@kalikiana I've been running the manual test here, and I get this error:
http://paste.ubuntu.com/24905555/

snapcraft/internal/lxd.py
kyrofa: Huh. This is fine of course, just surprised that apt doesn't release the lock.

snapcraft/internal/lxd.py
kyrofa: Then where are these keys going?

snapcraft/internal/lxd.py
kalikiana: It occurred to me this makes more sense in a separate PR, since LXD builds in general are affected by it. And to make it more interesting, I managed to get stuck with dpkg telling to run "dpkg --configure -a". So I removed it again from this one.

snapcraft/internal/lxd.py
kalikiana: I didn't want to make this a running target, what with IP changes and such - since we're in charge I figured it's safe to ignore that check, we're always SSHing into the container and using the key we generated.

snapcraft/internal/lxd.py
kalikiana: The keys are pushed to the LXD container. We never touch the configuration on the host running snapcraft.
Note I moved it to the runtime, as I did in the other branch.

snapcraft/internal/lifecycle.py
```diff
@@ -336,8 +337,16 @@ def _tar_filter(tarinfo):
 
 def containerbuild(step, project_options, output=None, args=[]):
     config = snapcraft.internal.load_config(project_options)
+    remote = os.environ.get('SNAPCRAFT_CONTAINER_BUILDS')
+    # Default remote if it's a truthy value - otherwise it's a remote name
+    try:
+        util.strtobool(remote)
+        remote = None
+    except ValueError:
+        pass
```
kyrofa: Can you explain the logic here? If `SNAPCRAFT_CONTAINER_BUILDS` is "0", `remote` is `None`. If `SNAPCRAFT_CONTAINER_BUILDS` is "1", `remote` is `None`. The only time `remote` is not None is if `SNAPCRAFT_CONTAINER_BUILDS` is "foo" or something.

snapcraft/internal/lxd.py
kyrofa: Yes, much better now, thank you.

snapcraft/internal/lifecycle.py
kalikiana: Exactly. Quoting myself from the exact same question asked by @elopio in this same PR: 

> This is so you can do SNAPCRAFT_CONTAINER_BUILDS=myremote to specify a remote or SNAPCRAFT_CONTAINER_BUILDS=1 to use the default (local) remote.

It's either a name or a truthy value. But this isn't where container builds are enabled or disabled.

sergiusens:
This is part of #1457

snapcraft/internal/lxd.py
```diff
@@ -304,24 +304,97 @@ def _ensure_container(self):
...
                 'path={}'.format(destination)])
 
+    def _remote_mount(self, destination, source):
+        # Generate an SSH key and add it to the container's known keys
+        keyfile = 'id_{}'.format(self._container_name)
+        keyfile_path = os.path.join(self._tmp, keyfile)
+        if not os.path.exists(keyfile_path):
+            # Swallow output and allow exceptions to save stdout
+            check_output(['ssh-keygen', '-o', '-N', '', '-f', keyfile_path])
+        # os.sep needs to be `/` and on Windows it will be set to `\`
```
kyrofa: I don't understand how this comment applies here. You're not using `os.sep`, right?

snapcraft/internal/lxd.py
```diff
@@ -304,24 +304,97 @@ def _ensure_container(self):
...
 
     def _setup_project(self):
         self._ensure_mount(self._project_folder, self._source)
 
+    def _get_container_address(self):
+        network = self._get_container_status()['state']['network']['eth0']
+        for address in network['addresses']:
+            if address['family'] == 'inet':
+                return address['address']
+        raise RuntimeError('No IP found for {}'.format(self._container_name))
```
kyrofa: Please make this an exception from SnapcraftError. Might be nice to create a base LxdError class and inherit from that even.

snapcraft/internal/lxd.py
kalikiana: For the record, you'll never run into the scenario where there's no external network access because we probe connectivity via a Python snippet after creating the container. Though the device may have a different name, so I added support for it as well as a test case which I just noticed I didn't have.

snapcraft/internal/lxd.py
kalikiana: I changed to SnapcraftEnvironmentError like the other existing ones and [proposed a separate PR for a new error class](#1505).

snapcraft/internal/lxd.py
kalikiana: Indeed not. I re-phrased it.

snapcraft/internal/lxd.py
```diff
@@ -317,11 +317,13 @@ def _setup_project(self):
...
-        network = self._get_container_status()['state']['network']['eth0']
-        for address in network['addresses']:
-            if address['family'] == 'inet':
-                return address['address']
-        raise RuntimeError('No IP found for {}'.format(self._container_name))
+        network = self._get_container_status()['state']['network']
+        for interface in network:
+            for address in network[interface]['addresses']:
+                if address['family'] == 'inet':
+                    return address['address']
```
kyrofa: Yeah I like this.

snapcraft/internal/lxd.py
```diff
@@ -343,12 +345,13 @@ def _remote_mount(self, destination, source):
         if not os.path.exists(keyfile_path):
             # Swallow output and allow exceptions to save stdout
             check_output(['ssh-keygen', '-o', '-N', '', '-f', keyfile_path])
-        # os.sep needs to be `/` and on Windows it will be set to `\`
+
+        # Don't use os.path.join because a Windows client would use \
```
kyrofa: Ah ha, that makes more sense indeed.

sergiusens:
This new implementation requires me to do something like this on the host running snapcraft:
```
sudo ip route add 10.113.236.0/24 via 192.168.1.121
```
Where `10.113.236.0/24` is the network the remote `lxdbr0` is on and `192.168.1.121` is the ip of the host with the lxd server.

snapcraft/internal/lifecycle.py
sergiusens: This logic could be improved, you also need to warn if the `remote` is valid and if a remote is actually going to be used, bonus for telling people why this remote is being set, something like;

```
Using a remote lxd instance as SNAPCRAFT_CONTAINER_BUILDS is set to use 'shire' as a remote
```

sergiusens: splitting out into its own thing might be good

snapcraft/internal/lxd.py
```diff
@@ -316,18 +316,88 @@ def _ensure_container(self):
...
+        # Use sshfs in slave mode inside SSH to reverse mount destination
+        self._container_run(['apt-get', 'install', '-y', 'sshfs'])
+        self._container_run(['mkdir', '-p', destination])
+        # Pipes for sshfs and sftp-server to communicate
+        stdin1, stdout1 = os.pipe()
+        stdin2, stdout2 = os.pipe()
+        try:
+            self._host_run(['/usr/lib/sftp-server'],
+                           stdin=stdin1, stdout=stdout2)
+        except CalledProcessError as e:
```
sergiusens: it seems I didn't get caught by this (same issue as that lxd bug you fixed, `FileNotFoundError`)
```
snapcraft
Waiting for a network connection...
  File "<string>", line 1, in <module>
  File "/usr/lib/python3.5/urllib/request.py", line 163, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/lib/python3.5/urllib/request.py", line 466, in open
    response = self._open(req, data)
  File "/usr/lib/python3.5/urllib/request.py", line 484, in _open
    '_open', req)
  File "/usr/lib/python3.5/urllib/request.py", line 444, in _call_chain
    result = func(*args)
  File "/usr/lib/python3.5/urllib/request.py", line 1282, in http_open
    return self.do_open(http.client.HTTPConnection, req)
  File "/usr/lib/python3.5/urllib/request.py", line 1257, in do_open
    r = h.getresponse()
  File "/usr/lib/python3.5/http/client.py", line 1197, in getresponse
    response.begin()
  File "/usr/lib/python3.5/http/client.py", line 297, in begin
    version, status, reason = self._read_status()
  File "/usr/lib/python3.5/http/client.py", line 258, in _read_status
    line = str(self.fp.readline(_MAXLINE + 1), "iso-8859-1")
  File "/usr/lib/python3.5/socket.py", line 575, in readinto
    return self._sock.recv_into(b)
socket.timeout: timed out
Network connection established
Get:1 http://security.ubuntu.com/ubuntu xenial-security InRelease [102 kB]
Hit:2 http://archive.ubuntu.com/ubuntu xenial InRelease                   
Get:3 http://archive.ubuntu.com/ubuntu xenial-updates InRelease [102 kB]      
Get:4 http://archive.ubuntu.com/ubuntu xenial-backports InRelease [102 kB]    
Fetched 306 kB in 3s (85.0 kB/s)                                  
Reading package lists... Done
Mounting /home/sergiusens/source/snapcraft/demos/godd into container
ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDquEdseWFnTTJkYdRHgpK8C6k+kSXVAgAll7dR5HAd0CWNg9Yfb0rqHCcj+YKMpPXd7hMTvMQ2BKdqimx52s055PEZab/GnFRALoXYLa/6i0P3CsYUA7FNqZAMBF5FVSjUIM+tqiwJNbGA7J7yAR/zmhOGMXYgKWxjcOnu8YGp72qOtS9tSZV5ZVovWjaAvOCBR519Khwnd123gH5L8XtXs2zP4ZDcVIgp0NXwSPOqWfDkykcFXgN3PpwI2ZtIdGTxSG/drTlnVjdTOnCINZ8gwSz5u+R+jerPQaT4dHb0iiJSV4cQ/gJYyLHsuX5nGODxIn3TwniqK0nhmjA1hp7x sergiusens@mirkwood
Connecting to 10.113.236.26 via SSH
Warning: Permanently added '10.113.236.26' (ECDSA) to the list of known hosts.
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following packages were automatically installed and are no longer required:
  gir1.2-gudev-1.0 golang-1.6-go golang-1.6-race-detector-runtime
  golang-1.6-src golang-go golang-race-detector-runtime golang-src
  libglib2.0-bin libglib2.0-dev libgudev-1.0-0 libgudev-1.0-dev libpcre16-3
  libpcre3-dev libpcre32-3 libpcrecpp0v5 pkg-config zlib1g-dev
Use 'apt autoremove' to remove them.
The following NEW packages will be installed:
  sshfs
0 upgraded, 1 newly installed, 0 to remove and 76 not upgraded.
Need to get 41.7 kB of archives.
After this operation, 138 kB of additional disk space will be used.
Get:1 http://archive.ubuntu.com/ubuntu xenial/universe amd64 sshfs amd64 2.5-1ubuntu1 [41.7 kB]
Fetched 41.7 kB in 2s (19.1 kB/s)
Selecting previously unselected package sshfs.
(Reading database ... 38315 files and directories currently installed.)
Preparing to unpack .../sshfs_2.5-1ubuntu1_amd64.deb ...
Unpacking sshfs (2.5-1ubuntu1) ...
Processing triggers for man-db (2.7.5-1) ...
Setting up sshfs (2.5-1ubuntu1) ...
Stopping shire:snapcraft-godd
  File "/home/sergiusens/venv/snapcraft/bin/snapcraft", line 11, in <module>
    load_entry_point('snapcraft', 'console_scripts', 'snapcraft')()
  File "/home/sergiusens/venv/snapcraft/lib/python3.6/site-packages/pkg_resources/__init__.py", line 561, in load_entry_point
  File "/home/sergiusens/venv/snapcraft/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2631, in load_entry_point
    return ep.load()
  File "/home/sergiusens/venv/snapcraft/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2291, in load
    return self.resolve()
  File "/home/sergiusens/venv/snapcraft/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2297, in resolve
  File "/home/sergiusens/source/snapcraft/snapcraft/cli/__main__.py", line 19, in <module>
    run(prog_name='snapcraft')
  File "/home/sergiusens/venv/snapcraft/lib/python3.6/site-packages/click/core.py", line 722, in __call__
  File "/home/sergiusens/venv/snapcraft/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/home/sergiusens/venv/snapcraft/lib/python3.6/site-packages/click/core.py", line 1043, in invoke
    return Command.invoke(self, ctx)
  File "/home/sergiusens/venv/snapcraft/lib/python3.6/site-packages/click/core.py", line 895, in invoke
  File "/home/sergiusens/venv/snapcraft/lib/python3.6/site-packages/click/core.py", line 535, in invoke
  File "/home/sergiusens/venv/snapcraft/lib/python3.6/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/sergiusens/source/snapcraft/snapcraft/cli/__init__.py", line 110, in run
    ctx.forward(lifecyclecli.commands['snap'])
  File "/home/sergiusens/venv/snapcraft/lib/python3.6/site-packages/click/core.py", line 553, in forward
    return self.invoke(cmd, **kwargs)
  File "/home/sergiusens/venv/snapcraft/lib/python3.6/site-packages/click/core.py", line 535, in invoke
  File "/home/sergiusens/source/snapcraft/snapcraft/cli/lifecycle.py", line 129, in snap
    lifecycle.containerbuild('snap', project_options, output, directory)
  File "/home/sergiusens/source/snapcraft/snapcraft/internal/lifecycle.py", line 354, in containerbuild
    metadata=config.get_metadata()).execute(step, args)
  File "/home/sergiusens/source/snapcraft/snapcraft/internal/lxd.py", line 409, in execute
    super().execute(step, args)
  File "/home/sergiusens/source/snapcraft/snapcraft/internal/lxd.py", line 144, in execute
    self._setup_project()
  File "/home/sergiusens/source/snapcraft/snapcraft/internal/lxd.py", line 317, in _setup_project
    self._ensure_mount(self._project_folder, self._source)
  File "/home/sergiusens/source/snapcraft/snapcraft/internal/lxd.py", line 332, in _ensure_mount
    return self._remote_mount(destination, source)
  File "/home/sergiusens/source/snapcraft/snapcraft/internal/lxd.py", line 382, in _remote_mount
    stdin=stdin1, stdout=stdout2)
  File "/home/sergiusens/source/snapcraft/snapcraft/internal/lxd.py", line 395, in _host_run
    self._processes += [Popen(cmd, **kwargs)]
  File "/usr/lib/python3.6/subprocess.py", line 707, in __init__
    restore_signals, start_new_session)
  File "/usr/lib/python3.6/subprocess.py", line 1333, in _execute_child
    raise child_exception_type(errno_num, err_msg)
FileNotFoundError: [Errno 2] No such file or directory: '/usr/lib/sftp-server'
```

snapcraft/internal/lxd.py
```diff
@@ -316,18 +316,88 @@ def _ensure_container(self):
...
+            # XXX: This needs to be extended once we support other distros
+            raise SnapcraftEnvironmentError(
+                'sftp-server could not be run.\n'
```
sergiusens: This message seems wrong
```
sudo apt install sftp-server
[sudo] password for sergiusens: 
Reading package lists... Done
Building dependency tree       
Reading state information... Done
E: Unable to locate package sftp-server
```

sergiusens:
I thought that to avoid filesystem issues (e.g.; ntfs) we were going to create entries on the container and mount from the host into the container.

This has the benefit of not needing to install sftp on the host opening some holes along the way.

snapcraft/internal/lxd.py
```diff
@@ -316,18 +316,88 @@ def _ensure_container(self):
...
+                'On Debian, Ubuntu and derivatives, the package '
+                'openssh-sftp-server needs to be installed.'
+                ) from e
+        self._host_run(ssh_cmd + ['sshfs -o slave -o nonempty :{} {}'.format(
```
sergiusens: this command could be run with `lxc exec`, right?
If so, and this is the direction to take, `ssh_cmd` could be removed and that check as well, avoiding the need to add routes.

sergiusens:
After installing `openssh-sftp-server` things started to move. I am still concerned with the approach given the need to add a route and open a port on the host.

These checks should also be done very early in the process.

@Saviq mind adding your thoughts to this PR?

snapcraft/internal/lxd.py
kalikiana: You'll want to re-read that error message.

Saviq:
> @Saviq <https://github.com/saviq> mind adding your thoughts to this PR?

Not sure what changed here, but originally the flow was:

sftp-server ↔STDIO↔ ssh client@localhost ↔SSH↔ ssh server@remote \
↔STDIO↔ sshfs

That meant no port needed opening on localhost, because all
communication was going via ssh, with a localhost client connecting to
the server. Were there issues with this approach?

As for routing, the above approach has the advantage of only requiring
the remote side to listen, which has to happen for the LXD remote to
work in the first place.

snapcraft/internal/lxd.py
kalikiana: No it can't. I guess I should rename the method.

snapcraft/cli/env.py
```diff
@@ -14,7 +14,40 @@
...
+from snapcraft.internal import errors
 
 
-def is_containerbuild():
-    return os.environ.get('SNAPCRAFT_CONTAINER_BUILDS')
+class ContainerConfig:
+
+    def __init__(self):
+        # The environment variable can be used in 3 ways
+        # 1. SNAPCRAFT_CONTAINER_BUILDS=1 enables local containers
```
sergiusens: This is really nice, you can even make this comment block into a docstring for `__init__` with some text massaging...
"""
Defines if a container should be used and if a remote other than local should be used.
Checks for the existence of `SNAPCRAFT_CONTAINER_BUILDS` and reads its value which can be interpreted in three different ways
"""

Just an illustration.

sergiusens:
There are conflicts here (due to the other lxd PR that landed). Mind rebasing and squashing as we discussed on Monday?

sergiusens:
there is a lot of legacy in this, a merge might be fine

This fixes #1457 so if you rebase, make sure to add that in the commit message

kalikiana:
I would rather not rebase this one.

I added the issue to the description.

snapcraft/internal/lxd.py
```diff
@@ -304,34 +307,101 @@ def _ensure_container(self):
...
+                    'lxc', 'config', 'device', 'add', self._container_name,
+                    'fuse', 'unix-char', 'path=/dev/fuse'
+                    ])
             check_call([
                 'lxc', 'start', self._container_name])
 
     def _setup_project(self):
         self._ensure_mount(self._project_folder, self._source)
 
+    def _get_container_address(self):
```
sergiusens: given your latest change this is dead code. Please remove

sergiusens:
And another follow up PR, clean is only destroying the container and not the data:
``` 
sergiusens  ⓔ  snapcraft  ⋯  snapcraft  demos  godd  snapcraft clean
Using LXD remote 'shire' from SNAPCRAFT_CONTAINER_BUILDS
Deleting shire:snapcraft-godd
 sergiusens  ⓔ  snapcraft  ⋯  snapcraft  demos  godd  ls
godd_1.0_amd64.snap  parts  prime  snap  stage
```

PR #1303, opened 2017-05-09
kernel plugin: slightly improve the messaging of check_config()

piso77:
Slightly improve the messaging of check_config(), by specifying if a kernel option applies only to a certain kernel version:

...
CONFIG_CC_STACKPROTECTOR_STRONG (4.1.x and later versions only)
...
CONFIG_DEVPTS_MULTIPLE_INSTANCES (4.8.x and earlier versions only)
...

No functional changes were introduced in this diff, just the messaging text.

LP: #1689564

### DIFF
--- a/snapcraft/plugins/kernel.py
+++ b/snapcraft/plugins/kernel.py
@@ -420,15 +420,12 @@ class KernelPlugin(kbuild.KBuildPlugin):
         if missing:
             warn = '\n{}\n'.format(msg)
             for opt in missing:
-                ver = ''
+                note = ''
                 if opt == 'CONFIG_CC_STACKPROTECTOR_STRONG':
-                    ver = '(4.1.x and later versions only)'
+                    note = '(4.1.x and later versions only)'
                 elif opt == 'CONFIG_DEVPTS_MULTIPLE_INSTANCES':
-                    ver = '(4.8.x and earlier versions only)'
-                warn += '{}'.format(opt)
-                if ver:
-                    warn += ' {}'.format(ver)
-                warn += '\n'
+                    note = '(4.8.x and earlier versions only)'
+                warn += '{} {}'.format(opt, note)
             logger.warn(warn)
 
     def _do_check_initrd(self, builtin, modules):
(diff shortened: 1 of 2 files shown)


snapcraft/plugins/kernel.py
```diff
@@ -81,25 +81,30 @@
...
+                   'SQUASHFS': '', 'SQUASHFS_XATTR': '', 'SQUASHFS_XZ': '',
+                   'DEVPTS_MULTIPLE_INSTANCES':
+                   '(4.8.x and earlier versions only)'}
+
+required_systemd = {'DEVTMPFS': '', 'CGROUPS': '', 'INOTIFY_USER': '',
+                    'SIGNALFD': '', 'TIMERFD': '', 'EPOLL': '',
+                    'NET': '', 'SYSFS': '', 'PROC_FS': '', 'FHANDLE': '',
+                    'DMIID': '', 'BLK_DEV_BSG': '', 'NET_NS': '',
+                    'IPV6': '', 'AUTOFS4_FS': '',
+                    'TMPFS_POSIX_ACL': '', 'TMPFS_XATTR': '', 'SECCOMP': ''}
```
elopio: Hey @piso77, I'm not sure that a dictionary is the right structure for this, because it's mostly empty.
The nicer message seems good to me, but if it only applies to two cases, adding two `if`s in the code is better.

snapcraft/plugins/kernel.py
```diff
@@ -420,7 +420,15 @@ def _do_check_config(self, builtin, modules):
...
-                warn += '{}\n'.format(opt)
+                ver = ''
+                if opt == 'CONFIG_CC_STACKPROTECTOR_STRONG':
+                    ver = '(4.1.x and later versions only)'
+                elif opt == 'CONFIG_DEVPTS_MULTIPLE_INSTANCES':
+                    ver = '(4.8.x and earlier versions only)'
+                warn += '{}'.format(opt)
+                if ver:
+                    warn += ' {}'.format(ver)
+                warn += '\n'
```
elopio: You could do it like this:

```
version_restriction = ''
if opt == 'CONFIG_CC_STACKPROTECTOR_STRONG':
    version_restriction = ' (4.1.x and later versions only)'
elif opt == 'CONFIG_DEVPTS_MULTIPLE_INSTANCES':
    version_restriction = ' (4.8.x and earlier versions only)'
warn += '{}{}\n'.format(opt, version_restriction)
```

A little shorter, but I'm just being a PITA, I'm ok with how it is now.

PR #1304, opened 2017-05-10
cleanbuild: set the container language to C.UTF-8

elopio:
LP: [#1662456](https://bugs.launchpad.net/snapcraft/+bug/1662456)

### DIFF
--- a/manual-tests.md
+++ b/manual-tests.md
@@ -42,6 +42,12 @@
 6. Ensure you are dropped back into your original shell session.
 
 
+# Test cleanbuild with non-ascii characters in the desktop file.
+
+1. Run `snapcraft cleanbuild` for the snap in `integration_tests/snaps/desktop-with-non-ascii`
+   * Check that the build succeeds and you get the `.snap` file.
+
+
 # Test cleanbuild with a remote.
 
 1. Setup a remote as described on
--- a/snapcraft/internal/lxd.py
+++ b/snapcraft/internal/lxd.py
@@ -55,7 +55,7 @@ class Containerbuild:
         _verify_remote(remote)
         self._container_name = '{}:snapcraft-{}'.format(remote, container_name)
         # Use the server architecture to avoid emulation overhead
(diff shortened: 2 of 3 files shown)


ogra1:
shouldn't this be C.UTF-8 (which is always available, contrary to en_*.UTF-8 which needs to be generated (admittedly often by default, but there in no guarantee))

sergiusens:
On Wed, May 10, 2017 at 7:10 AM, Oliver Grawert <notifications@github.com>
wrote:

> shouldn't this be C.UTF-8 (which is always available, contrary to
> en_*.UTF-8 which needs to be generated (admittedly often by default, but
> there in no guarantee))
>
yes it should. Our travis thing should also move to this (diff PR) to not
localegen all the time.

elopio:
That works. Thanks for the reviews.

elopio:
@kalikiana copying here what I mentioned during the stand up. Two problems to add a test here:

Currently, the integration tests for cleanbuild is disabled. We need to reenable it first.
But even with that test working, it takes a long time to start the container. I think it's not worth it to add a slow test to check this condition. I tested it manually, added a unit test to pin the fact that we are setting the environment, and in #1309 (thanks to your suggestion) I added an integration test for a non-ascii desktop file using the C locale. I think that's good enough.

sergiusens:
@kalikiana mind readressing your review after the comments?

@elopio and @kalikiana a test in manual-tests.md would be enough IMO for this particular case so it is at least documented and we have a path to automation.

manual-tests.md
```diff
@@ -42,6 +42,12 @@
 6. Ensure you are dropped back into your original shell session.
 
 
+# Test cleanbuild with non-ascii characters in the desktop file.
+
+1. Run `snapcraft cleanbuild` for the snap in `integration_tests/snaps/desktop-with-non-ascii`
+   * Check that the build succeeds and you get the `.snap` file.
```
elopio: This snap is added by #1309

elopio:
@kalikiana please re-review this.

PR #1305, opened 2017-05-10
Use architectures field which existed in older LXD

kalikiana:
LXD recently renamed the kernelarchitecture field to kernel-architecture so snapcraft is unwittingly not backwards-compatible.

This was filed as [bug 1689712](https://bugs.launchpad.net/snapcraft/+bug/1689712).

### DIFF
--- a/snapcraft/internal/lxd.py
+++ b/snapcraft/internal/lxd.py
@@ -55,7 +55,7 @@ class Containerbuild:
         _verify_remote(remote)
         self._container_name = '{}:snapcraft-{}'.format(remote, container_name)
         # Use the server architecture to avoid emulation overhead
-        kernel = self._get_remote_info()['environment']['kernel_architecture']
+        kernel = self._get_remote_info()['environment']['architectures'][0]
         deb_arch = _get_deb_arch(kernel)
         if not deb_arch:
             raise SnapcraftEnvironmentError(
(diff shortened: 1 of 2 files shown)


snapcraft/internal/lxd.py
```diff
@@ -55,7 +55,7 @@ def __init__(self, *, output, source, project_options,
         _verify_remote(remote)
         self._container_name = '{}:snapcraft-{}'.format(remote, container_name)
         # Use the server architecture to avoid emulation overhead
-        kernel = self._get_remote_info()['environment']['kernel_architecture']
+        kernel = self._get_remote_info()['environment']['architectures'][0]
```
elopio: This is riskier I think.
How do you know that the first architecture is always the kernel architecture? And I'm not sure if yaml.load guarantees the order in lists.
I was thinking of just trying with both values, like:

```
   try:
      kernel = self._get_remote_info()['environment']['kernel_architecture']
   except KeyError:
      kernel = self._get_remote_info()['environment']['kernelarchitecture']
```

kalikiana: I know because that's what LXD does in the source code. But I have no strong opinion one way or the other, so I'm happy to go for the fallback instead.

elopio: If you are sure this is not flaky, I'm ok with the approach that you prefer.

PR #1306, opened 2017-05-10
recording: record global build-packages installed on the host

elopio:


snapcraft/internal/meta.py
```diff
@@ -29,7 +29,10 @@
 
 from snapcraft import file_utils
 from snapcraft import shell_utils
-from snapcraft.internal import common
+from snapcraft.internal import (
+    common,
+    repo
+)
```
sergiusens: not asking for a change, but refactors had taught me that these are easier to handle if they are each one liners :wink:

elopio: but that's ugly!

sergiusens: that really depends on who is looking :wink:

I personally find the use of parens ugly :confounded:

PR #1307, opened 2017-05-10
cli: new UI and internal refactor

sergiusens:
This introduces a completely rewritten CLI for snapcraft based on
click. This is the first step in decoupling library and cli for
snapcraft.


snapcraft/cli/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+from snapcraft.internal import deprecations
+from snapcraft.internal import log
+from . import echo
+from .assertions import assertionscli
+from .discovery import discoverycli
+from .lifecycle import lifecyclecli
+from .store import storecli
+from .parts import partscli
+from .help import helpcli
+from .ci import cicli
```
elopio: I don't like this relative imports. Why do you like them?

snapcraft/cli/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+_cmd_aliases = {
+    'registered': 'list-registered',
+    'keys': 'list-keys',
+    'revisions': 'list-revisions',
+    'plugins': 'list-plugins',
+}
+
+_cmd_deprecation_notices = {
+    'history': 'dn4',
+}
```
elopio: These are all constants. They should be upper case.

snapcraft/cli/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+    def parse_args(self, ctx, args):
+        if not args:
+            args.insert(0, 'snap')
+        return super().parse_args(ctx, args)
+
+
+@click.group(cls=SnapcraftGroup)
+@click.pass_context
+@click.option('--debug', is_flag=True)
+def run(ctx, debug, catch_exceptions=False):
```
elopio: what is that catch_exceptions argument that you are not using?

snapcraft/cli/_options.py
```diff
@@ -0,0 +1,47 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+import click
+
+from snapcraft import ProjectOptions
+
+
+_build_options = [
```
elopio: constant, so upper-case

snapcraft/cli/_options.py
```diff
@@ -0,0 +1,47 @@
...
+    click.option('--no-parallel-builds', is_flag=True,
+                 help='Force a squential build.'),
+    click.option('--target-arch', metavar='<arch>',
+                 help='Target architecture to cross compile to'),
+]
+
+
+def add_build_options():
+    def _add_build_options(func):
+        for option in reversed(_build_options):
```
elopio: Am I missing something obvious here? Why do you reverse it instead of defining the list in the order you want?

snapcraft/cli/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+import logging
+
+import click
+
+import snapcraft.internal.dirs
+from snapcraft.internal import deprecations
+from snapcraft.internal import log
+from . import echo
+from .assertions import assertionscli
```
elopio: These are already in the cli module. Do you think it's necessary to repeat `cli` in the function name?

snapcraft/cli/assertions.py
```diff
@@ -0,0 +1,112 @@
...
+def list_keys():
+    """List the keys available to sign assertions.
+
+    This command has an alias of `keys`.
+    """
+    try:
+        snapcraft.list_keys()
+    except Exception as e:
+        echo.error(e)
+        sys.exit(1)
```
elopio: Having multiple exit points makes things harder to test. Is it possible to keep this like before, with a single point?

snapcraft/cli/__init__.py
elopio: maybe naming them like assertions_command_group, if it needs to be clearer.

snapcraft/internal/common.py
```diff
@@ -183,7 +183,7 @@ def reset_env():
 
 
 def get_terminal_width(max_width=MAX_CHARACTERS_WRAP):
-    if os.isatty(sys.stdout.fileno()):
+    if os.isatty(1):
```
elopio: could you explain why stdout is 1 here? A comment in the code would be nice.

elopio:
I only have style comments. All of which can be ignored and discussed when we are working on the nice API.

elopio:
@sergiusens make sure to pull. I made a tiny change to get the integration tests going.

snapcraft/cli/__init__.py
sergiusens: Because @kyrofa and I like them more :-P

Really means we can move this package around (or others) with less effort later.

snapcraft/cli/__init__.py
sergiusens: If you mean dn4, feel free to change that in a different PR, they are defined in the deprecation_notices package

snapcraft/cli/_options.py
sergiusens: it is so the decoration applies in the order I defined them

snapcraft/cli/_options.py
sergiusens: I would consider this more of a private global... but ok :wink:

snapcraft/cli/__init__.py
sergiusens: they are cli's in themselves (can be independently run as well) so it would be more confusing.

snapcraft/cli/assertions.py
sergiusens: Given that each is an independent command that can be executed on its own, I'd say no... it really isn't harder to test. This also allows for fine grained exception catching to improve the product

snapcraft/internal/common.py
sergiusens: I asked you if you could fix it two days ago ;-)

snapcraft/cli/__init__.py
sergiusens: By default click will catch exceptions and exit with '-1' but the message is lost with the way we wrote our exceptions.

snapcraft/cli/env.py
```diff
@@ -0,0 +1,20 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+import os
+
+
+def is_containerbuild():
+    return os.environ.get('SNAPCRAFT_CONTAINER_BUILDS')
```
kalikiana: Thinking out loud... are you planning on adding other variables here? I'm wondering about namespacing since this one is about LXD - on the other hand it's a general flag that's handled before the actual LXD code comes into play.

snapcraft/cli/lifecycle.py
```diff
@@ -0,0 +1,209 @@
...
+
+    \b
+    Examples:
+        snapcraft clean
+        snapcraft clean my-part --step build
+    """
+    project_options = get_project_options(**kwargs)
+    if env.is_containerbuild():
+        step = step or 'pull'
+        lifecycle.containerbuild('clean', project_options,
```
kalikiana: Might be nice to unify this more so that containerbuild and other lifecycle commands mirror each other. But it probably makes sense to attempt that in a follow-up as it's complex enough as it is.

snapcraft/tests/commands/__init__.py
```diff
@@ -60,3 +67,59 @@ def mock_snap_output(command, *args, **kwargs):
...
+
+
+class CommandBaseTestCase(tests.TestCase):
+
+    def setUp(self):
+        super().setUp()
+        self.runner = CliRunner()
+
+    def run_command(self, args, **kwargs):
+        return self.runner.invoke(run, args, env=os.environ,
```
kalikiana: Why the explicit env=os.environ here? That seems like it should be the default outside of test cases already.

snapcraft/cli/env.py
sergiusens: I really want to reach the point of killing this flag. Namespacing by putting it in the lxd module? That works for me.

snapcraft/cli/lifecycle.py
sergiusens: I was aiming for that, but the code changes just thinned out too much, we can have many follow up PRs for this while we also split UI from logic some more.

snapcraft/tests/commands/__init__.py
sergiusens: I was just going crazy on getting -1 as exit codes and thought it was due t lack of environment. I'll get rid of this 3AM change, it is not needed. Nice catch

snapcraft/cli/__init__.py
kalikiana: Is there any sort of documentation on what is preferred? The style of imports isn't overly consistent between different files right now. Including things like "from foo import bar" versus "import foo" using foo.bar in code. Might be nice to look into that after this PR is done.

snapcraft/cli/env.py
kalikiana: Yeah. Also because eventually it'll be reading LXD-specific settings like the container name.

sergiusens:
I cannot believe this is all finally green :-)

snapcraft/cli/store.py
```diff
@@ -0,0 +1,275 @@
...
+from snapcraft.storeapi.constants import DEFAULT_SERIES
+from . import echo
+
+
+_MESSAGE_REGISTER_PRIVATE = dedent("""\
+    Even though this is private snap, you should think carefully about
+    the choice of name and make sure you are confident nobody else will
+    have a stronger claim to that particular name. If you are unsure
+    then we suggest you prefix the name with your developer identity,
+    As ‘nessita-yoyodyne-www-site-content’.""")
```
elopio: I think this is a pretty bad example.
Maybe better: `username-my-snap`

snapcraft/cli/store.py
```diff
@@ -0,0 +1,275 @@
...
+_MESSAGE_REGISTER_CONFIRM = dedent("""
+    We always want to ensure that users get the software they expect
+    for a particular name.
+
+    If needed, we will rename snaps to ensure that a particular name
+    reflects the software most widely expected by our community.
+
+    For example, most people would expect ‘thunderbird’ to be published by
```
elopio: firefox is a better example than thunderbird, don't you think?

snapcraft/internal/common.py
elopio: I'm confused. Is this related to the `detach`? or something else I completely forgot?

PR #1308, opened 2017-05-10
allow capital Y to accept the dev agreement

felicianotech:
LP: [#1689953](https://bugs.launchpad.net/snapcraft/+bug/1689953)

### DIFF
--- a/snapcraft/_store.py
+++ b/snapcraft/_store.py
@@ -91,7 +91,7 @@ def _check_dev_agreement_and_namespace_statuses(store):
                    storeapi.constants.UBUNTU_STORE_TOS_URL)
             choice = input(
                 storeapi.constants.AGREEMENT_INPUT_MSG.format(url))
-            if choice == 'y':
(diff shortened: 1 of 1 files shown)


elopio:
Thanks @felicianotech! I think it's ok to accept y and Y. It would be nicer if we made the change everywhere to maintain consistency.

But note that there is a huge UI change coming in #1307. So we will leave this PR here for a while, and get back to it once the other PR is merged.

felicianotech:
I didn't realize there are more of those prompts. I can surely find them all and change them however I guess I'll wait for that other PR first.

sergiusens:
On Wed, May 10, 2017 at 11:54 PM, Ricardo N Feliciano <
notifications@github.com> wrote:

> I didn't realize there are more of those prompts. I can surely find them
> all and change them however I guess I'll wait for that other PR first.
>
> FWIW, we are migrating to click which has a click.confirm method that
takes care of all this.

PR #1309, opened 2017-05-11
meta: read and write the desktop file with utf-8 encoding

elopio:
LP: [#1662456](https://bugs.launchpad.net/snapcraft/+bug/1662456)


elopio:
I couldn't add an unit tests, probably because python is already running with utf-8 so I can't patch LC_ALL. So I added an integration test.

PR #1311, opened 2017-05-11
add missing VCS dependencies to HACKING.md

felicianotech:
Noticed that these packages were missing when I ran `./runtests.sh`.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -58,7 +58,7 @@ If you don't have python3 installed, you can use the one from the archives:
 
 Install the needed dependencies.
 
-    sudo apt install build-essential python3-dev python3-debian libapt-pkg-dev libsodium-dev gcc libffi-dev libarchive13 squashfs-tools xdelta3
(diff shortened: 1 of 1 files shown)


HACKING.md
```diff
@@ -58,7 +58,7 @@ If you don't have python3 installed, you can use the one from the archives:
 
 Install the needed dependencies.
 
-    sudo apt install build-essential python3-dev python3-debian libapt-pkg-dev libsodium-dev gcc libffi-dev libarchive13 squashfs-tools xdelta3
+    sudo apt install build-essential python3-dev python3-debian libapt-pkg-dev libsodium-dev gcc libffi-dev libarchive13 squashfs-tools xdelta3 bzr git mercurial subversion
```
elopio: @sergiusens I'm not sure why do we recommend installing with pip.
Wouldn't it be better to recommend debs?

    $ sudo apt install build-dep snapcraft
    $ ./tools/install_autopkgtest_dependencies.py integrationtests
    $ ./tools/install_autopkgtest_dependencies.py snapstests

To me, that seems easier to the users and easier to keep in sync.

sergiusens: Because then you would get different versions for python packages depending on where you are working from

PR #1312, opened 2017-05-12
state: fix the name of the source details

elopio:


snapcraft/internal/sources/_bazaar.py
```diff
@@ -78,8 +78,8 @@ def _get_source_details(self):
         source = self.source
 
         return {
-            'commit': commit,
-            'branch': branch,
+            'source-commit': commit,
+            'source-branch': branch,
             'source': source,
```
elopio: Joe was naming this source-source. That's wrong, right???

sergiusens: @kyrofa did the original review for that, he might know the whys and whats

elopio: From @kyrofa "yeah that's probably safe"

elopio:
@kalikiana what I would like is that we get a suite for reproducible builds. So build once, then build again using prime/snap/snapcraft and check that the snaps are equal.

Now that this bug is fixed, I don't see a lot of value in a static check of the recorded yaml, because the unit tests are pinning the current behaviour.

PR #1314, opened 2017-05-12
catkin plugin: add support for rosinstall files

kyrofa:
ROS projects are often made up of many sub-projects. Some created by the community, some created by OSRF, others created by the robotics vendor themselves. Projects use rosinstall files for keeping track of all these dependencies and pulling them into the same workspace. However, currently Snapcraft doesn't support rosinstall files.

This PR resolves LP: [#1533735](https://bugs.launchpad.net/snapcraft/+bug/1533735) by introducing support for rosinstall files in the Catkin plugin. It does this by adding a new property to the plugin: `rosinstall-files`. Multiple rosinstall files can be added, and they will be merged into the workspace in the order given.


demos/rosinstall/src/quittable_listener/package.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0"?>
+<package>
+  <name>quittable_listener</name>
+  <version>0.0.0</version>
+  <description>The listener package</description>
+  <maintainer email="me@example.com">me</maintainer>
+  <license>TODO</license>
```
elopio: a testing tip, not for now but to take it into account. Usually, we use the word dummy to identify parts of the test that need to be present but affect the test in no way. Instead of TODO, DUMMY would be a better placeholder here, I think.

snapcraft/plugins/catkin.py
```diff
@@ -34,6 +34,10 @@
     - include-roscore:
       (boolean)
       Whether or not to include roscore with the part. Defaults to true.
+    - rosinstall-files:
+      (list of strings)
+      List of rosinstall files to merge while pulling. Paths are relative to
+      the source.
```
elopio: What do you think about writing a post in the forum or in your blog, to accompany this feature?
A place where we can point people to if they want to use it, with a nicely explained example?

demos/rosinstall/src/quittable_listener/package.xml
kyrofa: That's fair, though note that this scaffolding is generated by `catkin_create_pkg`, not written by hand.

snapcraft/plugins/catkin.py
kyrofa: Yeah I think that's an excellent idea :+1: .

snapcraft/plugins/catkin.py
```diff
@@ -919,3 +961,58 @@ def _get_highest_version_path(path):
...
+        self._ubuntu_sources = ubuntu_sources
+        self._wstool_path = wstool_path
+        self._wstool_install_path = os.path.join(wstool_path, 'install')
+        self._project = project
+
+    def setup(self):
+        os.makedirs(self._wstool_install_path, exist_ok=True)
+
+        # wstool isn't a dependency of the project, so we'll unpack it
+        # somewhere else, and use it from there.
```
elopio: can't you make it a build dependency if the rosinstall property is set in the yaml?
Adding dependencies here would affect the recording of build artifacts. It's doable, but it would be nicer to keep all the build packages in the same place.

kyrofa: Not really, no. This is similar to the compilers and catkin bits: they (in most cases) come from the ROS repos, not Ubuntu's. We can't assume those repos are setup on the host machine, and we certainly don't want to add them. So we don't install them at all: we fetch them off to the side and use them. They don't end up in the snap, and they don't provide any libs, so no need to track them at all, I'd say.

snapcraft/tests/plugins/test_catkin.py
```diff
@@ -249,6 +260,55 @@ def test_schema(self):
...
+                        '"rosinstall-files"')
+        self.assertTrue('minitems' in rosinstall_files,
+                        'Expected "minitems" to be included in '
+                        '"rosinstall-files"')
+        self.assertTrue('uniqueItems' in rosinstall_files,
+                        'Expected "uniqueItems" to be included in '
+                        '"rosinstall-files"')
+        self.assertTrue('items' in rosinstall_files,
+                        'Expected "items" to be included in '
+                        '"rosinstall-files"')
```
elopio: I would drop all those assertion messages, and just use self.assertIn instead.
Not a big deal, or -1, just my preference.

snapcraft/tests/plugins/test_catkin.py
```diff
@@ -91,13 +92,23 @@ class props:
...
+        self.addCleanup(patcher.stop)
+
     def verify_rosdep_setup(self, rosdistro, package_path, rosdep_path,
                             sources):
         self.rosdep_mock.assert_has_calls([
             mock.call(rosdistro, package_path, rosdep_path, sources,
                       self.project_options),
             mock.call().setup()])
 
+    def verify_wstool_setup(self, package_path, wstool_path, sources):
```
elopio: assert instead of verify?

elopio:
This looks good to me. I will trigger the autopkgtests.

snapcraft/plugins/catkin.py
elopio: If we want reproducible builds, we have to track them. But if they are not part of the build packages, then this has to be part of the future work to let plugins save stuff in the state that will then end up in prime/snap/snapcraft.yaml. Nothing to do for now I guess.

demos/rosinstall/src/quittable_listener/package.xml
kyrofa: I updates all ROS demos to use GPLv3 here.

snapcraft/tests/plugins/test_catkin.py
kyrofa: Good idea. I rewrote this entire test to be less verbose and use `assertThat`.

snapcraft/tests/plugins/test_catkin.py
kyrofa: Sure, done.

kyrofa:
This is currently blocked behind (unrelated) broken autopkgtests on arm. Waiting for those to be fixed.

kyrofa:
Are we still waiting for release before this can be merged?

snapcraft/plugins/catkin.py
```diff
@@ -919,3 +961,58 @@ def _get_highest_version_path(path):
...
+            ['update', '-j{}'.format(self._project.parallel_build_count),
+             '-t{}'.format(self._ros_package_path)]).strip()
+
+    def _run(self, arguments):
+        env = os.environ.copy()
+
+        # We want to make sure we use our own rosdep (which is python)
+        env['PATH'] += ':' + os.path.join(self._wstool_install_path, 'usr',
+                                          'bin')
+        env['PYTHONPATH'] = os.path.join(self._wstool_install_path, 'usr',
```
sergiusens: this technically shouldn't be needed as this site packages entry is dynamically detected from the execution path. If that is not the case, mind adding a comment?

snaps_tests/demos_tests/test_rosinstall.py
```diff
@@ -0,0 +1,41 @@
...
+import snaps_tests
+
+
+class RosinstallTestCase(snaps_tests.SnapsTestCase):
+
+    snap_content_dir = 'rosinstall'
+
+    @skipUnless(linux_distribution()[2] == 'xenial',
+                'This test fails on yakkety LP: #1614476')
+    def test_rosinstall(self):
```
sergiusens: only works on `xenial` or does it work on `zesty` and `artful`?

kyrofa: Yeah, later releases barf on the key. Might be sha1-related.

snapcraft/plugins/catkin.py
kyrofa: Removing this causes wstool to not be found, so this does indeed seem to be required. Comment added.

kyrofa:
(tests keep failing due to connection resets. Retrying...)

snaps_tests/demos_tests/test_rosinstall.py
sergiusens: yeah, the comment is still incorrect though :wink:

PR #1315, opened 2017-05-13
lxd: mock platform in the FakeLXD fixture

kalikiana:
Mocking the platform should be done in the FakeLXD fixture rather than individual tests so assumptions on server architecture and mock cross-compiling hold true consistently.


elopio:
lgtm. I'll trigger arm tests here.

PR #1316, opened 2017-05-13
Run unit tests on Travis with mock armhf

kalikiana:
Since **Travis** doesn't run tests on _architectures_ other than **amd64** certain failures such as [bug 1690294](https://bugs.launchpad.net/snapcraft/+bug/1690294) only show up in autopkgtest runs. By mocking **platform.machine** globally they can be caught.


sergiusens:
the goal is fine but the approach feels odd. The test should just mock _get_platform_machine

elopio:
@kalikiana it seems you will have to fix all the failing tests here, in order to add the new faked test.
The extra 10 minutes that the suite will take on all PRs seems well worth it, this is a good idea, thanks.

snapcraft/tests/__init__.py
```diff
@@ -139,6 +139,13 @@ def setUp(self):
         self.parts_dir = os.path.join(os.getcwd(), 'parts')
         self.local_plugins_dir = os.path.join(self.snap_dir, 'plugins')
 
+        machine = os.environ.get('MOCK_MACHINE', None)
```
elopio: Would it make sense to try to namespace the environment variables? Something like SNAPCRAFT_TEST_MOCK_MACHINE.

Here are some env vars that only apply to travis, like $DEPENDENCIES and $TEST_SUITE, I'm not sure if those should also be namespaced. But $TEST_SNAP_WITH_TRACKS I think should be something like $SNAPCRAFT_TEST_SNAP_WITH_TRACKS. If we agree on a style, I can apply it in a later branch.

kyrofa: +1 to namespacing.

PR #1317, opened 2017-05-14
recording: save the details of the source pulled

elopio:


elopio:
Requires #1312

integration_tests/__init__.py
```diff
@@ -212,6 +216,108 @@ def set_package_version(
...
+
+
+class GitSourceBaseTestCase(TestCase):
+
+    def setUp(self):
+        super().setUp()
+        if shutil.which('git') is None:
+            self.skipTest('git is not installed')
+
+    def init_and_config_git(self):
```
kalikiana: The arbitration of init methods strikes me as odd. Why include the name of the VCS here and nowhere else? And can we have "init" and "config" separately for clarity?

integration_tests/test_snapcraft_recording.py
```diff
@@ -197,3 +198,108 @@ def test_prime_with_packages_missing_dependency(self):
...
+
+        subprocess.check_call(
+            ['svn', 'checkout',
+             'file:///{}'.format(os.path.join(self.path, 'repo')),
+             'local'],
+            stdout=subprocess.DEVNULL)
+
+        open(os.path.join('local', 'file'), 'w').close()
+        subprocess.check_call(
+            ['svn', 'add', 'file'], stdout=subprocess.DEVNULL, cwd='local/')
```
kalikiana: This should probably be moved to add/commit/update methods in SubversionSourceBaseTestCase.

integration_tests/__init__.py
elopio: When reading the tests, if the function was only named `init` it wasn't very clear for me. Names like `commit` and `tag` clearly refer to vcs, but init doesn't give me a clue of what it is initializing.
Maybe I can rename it to `init_source_control`, so it's the same for all the classes.

elopio: And I think we can just call it init, remove the config part. But I'd prefer to leave it in a single function, at least until we find a test that requires init but not the part setting the email.

integration_tests/test_snapcraft_recording.py
elopio: Yes, I'll move it. But ugh, svn is so ugly. I will have to add an argument for the cwd.

elopio:
@kalikiana please re-review

integration_tests/test_snapcraft_recording.py
kalikiana: What about the other svn operations?

elopio:
@kalikiana I didn't want to touch svn too much, but oh well, I refactored it all ^_^

integration_tests/test_git_source.py
```diff
@@ -13,6 +13,7 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
```
sergiusens: why?

elopio: consistency. It's like that everywhere else, isn't it?

PR #1318, opened 2017-05-16
tests: refactor the travis jobs using stages

elopio:


.travis.yml
```diff
@@ -3,44 +3,27 @@ dist: trusty
...
+jobs:
+  include:
+    # CLA check, only in pull requests.
+    - stage: test
+      script: if [ "$TRAVIS_PULL_REQUEST" != "false" ] && [ "$TRAVIS_EVENT_TYPE" != 'cron' ]; then ./tools/travis/run_cla_check.sh; fi
+    # Tests, only when not triggered by the daily cron.
+    - stage: test
+      script: if [ "$TRAVIS_EVENT_TYPE" != "cron" ]; then ./tools/travis/run_tests.sh static; fi
+    - stage: test
+      script: if [ "$TRAVIS_EVENT_TYPE" != "cron" ]; then ./tools/travis/run_tests.sh unit && pip install codecov coverage urllib3[security] && python3 -m coverage xml && codecov; fi
```
kalikiana: Also shouldn't it be `stage: coverage` since previously after_success was used to only run it afterwards. ?

.travis.yml
```diff
@@ -3,44 +3,27 @@ dist: trusty
...
-
-after_success:
-  - if [ "$TEST_SUITE" == "unit" ]; then pip install codecov coverage && python3 -m coverage xml && codecov; fi
+jobs:
+  include:
+    # CLA check, only in pull requests.
+    - stage: test
+      script: if [ "$TRAVIS_PULL_REQUEST" != "false" ] && [ "$TRAVIS_EVENT_TYPE" != 'cron' ]; then ./tools/travis/run_cla_check.sh; fi
+    # Tests, only when not triggered by the daily cron.
+    - stage: test
```
elopio: I found that confusing, because I'm not very used to relying in the yaml order. But I will give it a try.

.travis.yml
elopio: I moved coverage to the script.
It shouldn't be a separate stage, because it would run in a new runner, and we need it to run in the same machine that the unit tests ran.

elopio:
@kyrofa this is very ready! Waiting for the second +1.

.travis.yml
```diff
@@ -3,44 +3,22 @@ dist: trusty
...
-  # Generate docs, only in pushes to master.
-  - if [ ! -z $GEN_DOCS ] && [ "$TRAVIS_EVENT_TYPE" = "push" ] && [ "$TRAVIS_BRANCH" = "master" ]; then docker exec -i builder ./tools/gen_api_docs.sh && ./tools/push_api_gh_pages.sh; fi
-
-after_success:
-  - if [ "$TEST_SUITE" == "unit" ]; then pip install codecov coverage && python3 -m coverage xml && codecov; fi
+jobs:
+  include:
+    - stage: test
+      # CLA check, only in pull requests.
+      script: if [ "$TRAVIS_PULL_REQUEST" != "false" ] && [ "$TRAVIS_EVENT_TYPE" != 'cron' ]; then ./tools/travis/run_cla_check.sh; fi
```
kyrofa: I know this isn't new in this PR, but what value is there in using `if` here as opposed to just this?

```bash
[ "$TRAVIS_PULL_REQUEST" != "false" ] && [ "$TRAVIS_EVENT_TYPE" != 'cron' ] && ./tools/travis/run_cla_check.sh
```

.travis.yml
```diff
@@ -3,44 +3,22 @@ dist: trusty
...
+    # Tests, only when not triggered by the daily cron.
+    - script: if [ "$TRAVIS_EVENT_TYPE" != "cron" ]; then ./tools/travis/run_tests.sh static; fi
```
kyrofa: I feel like the static tests should be in their own stage, run before everything else. We won't accept a PR that fails those, why run the heavier tests? It will also catch those stupid quick-fix errors without wasting time. Thoughts? I'd say the same thing for CLA if that wasn't a manual process.

.travis.yml
elopio: For somebody not used to bash, I find the `if` a lot more readable.
You made a good point that for one liners, the brackets might be better because you don't have to use `;` and `then`, which are weird. But I am still slightly more inclined to use `if`.

.travis.yml
elopio: I agree with you now. So I've split into more stages:

static: CLA && python static check
unit
integration: integration, plugins and store

kyrofa: I suggest leaving the CLA check in the last stage, only because it's a manual process. I imagine something like this:

1. A PR is made.
2. Tests immediately fail because CLA isn't signed. We say "hey, sign the CLA."
3. Dev comes back saying "I did already!"
4. We have to wait for the manual process to complete before we can even really review.

elopio: So, in the integration stage, or in a separate stage?
I'm not so convinced, but let's give it a try.

PR #1322, opened 2017-05-17
state: ignore the 'any' architecture in the build packages apt cache

elopio:
LP: [#1691569](https://bugs.launchpad.net/snapcraft/+bug/1691569)

PR #1323, opened 2017-05-17
 state: search for the build package that provides a virtual package

elopio:
LP: [#1691580](https://bugs.launchpad.net/snapcraft/+bug/1691580)

### DIFF
--- a/snapcraft/internal/repo/_deb.py
+++ b/snapcraft/internal/repo/_deb.py
@@ -201,10 +201,9 @@ class Ubuntu(BaseRepo):
                     elif version and installed_version != version:
                         new_packages.append(pkg)
                 except KeyError as e:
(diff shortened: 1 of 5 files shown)


elopio:
Requires #1322

snapcraft/internal/repo/_deb.py
```diff
@@ -380,6 +396,12 @@ def _manifest_dep_names(self, apt_cache):
         return manifest_dep_names
 
 
+def _find_virtual_package_provider(apt_cache, package_name):
```
kalikiana: Why not use apt_cache.get_providing_packages?

elopio: whaaat? I searched the api for something like this, and clearly did a lousy job. Thanks for finding it!

PR #1327, opened 2017-05-23
tests: small updates for manual kernel tests

elopio:


manual-tests.md
```diff
@@ -123,7 +123,7 @@
 
 1. Download https://developer.qualcomm.com/download/db410c/linux-board-support-package-v1.2.zip
 2. Extract it and copy the file `firmware.tar` to the directory `demos/96boards-kernel`.
-3. Run `snapcraft --target-arch arm64` in the `demos/96boards-kernel` directory.
+3. Run `snapcraft snap --target-arch arm64` in the `demos/96boards-kernel` directory.
```
kalikiana: This is necessary because the behavior of options changed with the new CLI. Perhaps this should wait until we've discussed this properly?

sergiusens: we will certainly fix this, this change however is not necessarily wrong.

PR #1328, opened 2017-05-23
qmake plugin: set default qt version

tim-sueberkrueb:
Right now it is necessary to manually specify the Qt version to use for the QMake plugin.
For convenience however, the default value should be qt5, as discussed on Launchpad: 
[LP: #1633278](https://bugs.launchpad.net/snapcraft/+bug/1633278)

### DIFF
--- a/snapcraft/tests/plugins/test_qmake.py
+++ b/snapcraft/tests/plugins/test_qmake.py
@@ -95,12 +95,15 @@ class QMakeTestCase(tests.TestCase):
         qt_version = properties['qt-version']
         self.assertTrue('enum' in qt_version,
                         'Expected "enum" to be included in "qt_version"')
-        self.assertFalse('default' in qt_version,
-                         '"default" was unexpectedly included in "qt_version"')
 
         qt_version_enum = qt_version['enum']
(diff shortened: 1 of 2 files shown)


snapcraft/tests/plugins/test_qmake.py
```diff
@@ -41,8 +41,8 @@ def setUp(self):
 
         class Options:
             options = []
-            qt_version = 'qt5'
             project_files = []
+            qt_version = 'qt5'
```
kalikiana: I'd prefer to avoid this type of noise (moving code around unchanged).

That said, all of this looks good to me.

tim-sueberkrueb: @kalikiana oh, I'm sorry, I didn't noticed that (didn't move that line on purpose ;)). If you want to, I can revert that.

kalikiana: Please do, if it's not too troublesome.

tim-sueberkrueb: @kalikiana absolutely not. Should be fixed now. Thanks for reviewing!

PR #1332, opened 2017-05-23
cli: provide a whoami command

sergiusens:
There are proper fallbacks here as we never
stored the email in our original credentail config location. This provides the proper messaging so this is correct for users wanting to know this.

In some distant future, the account info from the store will also provide this information which is why we loop over the account information .

### DIFF
--- a/snapcraft/storeapi/__init__.py
+++ b/snapcraft/storeapi/__init__.py
@@ -160,6 +160,7 @@ class StoreClient():
         # The macaroon has been discharged, save it in the config
         self.conf.set('macaroon', macaroon)
         self.conf.set('unbound_discharge', unbound_discharge)
+        self.conf.set('email', email)
         if save:
             self.conf.save()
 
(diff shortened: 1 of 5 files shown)


cprov:
Thanks @sergiusens, it looks good and efficient.

elopio:
\o/ All green. With my review and Celso's, that's +2, but kind of cheating.
It would be better if @kalikiana or @kyrofa review it too.

snapcraft/storeapi/__init__.py
```diff
@@ -188,6 +189,25 @@ def _refresh_if_necessary(self, func, *args, **kwargs):
...
             return func(*args, **kwargs)
 
+    def whoami(self):
+        """Return user relevant login information."""
+        account_data = {}
+
+        for k in ('email', 'account_id'):
+            try:
+                value = self.conf.get(k)
+            except KeyError as e:
```
kyrofa: Looking at `snapcraft.config.Config()` ([reference](https://github.com/snapcore/snapcraft/blob/master/snapcraft/config.py#L56)), I don't see how this could ever possibly raise a `KeyError`, so this chunk of code seems to do nothing, no?

elopio: lol, good catch.

PR #1333, opened 2017-05-24
state: search for the dependencies in the archive

elopio:
LP: [#1693058](https://bugs.launchpad.net/snapcraft/+bug/1693058)

### DIFF
--- a/snapcraft/internal/repo/_deb.py
+++ b/snapcraft/internal/repo/_deb.py
@@ -258,7 +258,7 @@ class Ubuntu(BaseRepo):
                     installed_packages.append(str(installed_package))
                     dependencies = cls._get_dependencies(
                         apt_cache, installed_package)
-                    installed_packages.extend(
+                    build_packages.extend(
                         [dependency for dependency in dependencies
                          if dependency not in installed_packages])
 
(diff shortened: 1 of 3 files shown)


snapcraft/internal/repo/_deb.py
```diff
@@ -252,26 +252,52 @@ def get_installed_build_packages(cls, package_names):
...
+            for depend in depends:
+                try:
+                    dependency = cls._get_providing_package(
+                        apt_cache, depend.name)
+                    dependencies.append(dependency.package.name)
+                    break
+                except errors.BuildPackageNotFoundError:
+                    pass
+            else:
+                raise errors.BuildPackageNotFoundError(depends)
```
kalikiana: This looks a little ugly to me for two reasons:

1. You're looping through the list, breaking the loop **unless** there's an exception in which case... the loop continues. This is awkward.
    Can we check if the package is in the cache instead of the _try/ except_?
2. You're raising **BuildPackageNotFoundError** with a list of packages - the exception takes a single package name.
    Perhaps we need a new exception, like BuildDependenciesMissingError, that takes a list. In the future we could also try harder to list all missing packages rather than just the first one.

elopio:
We decided to change the approach, and I proposed #1340.

PR #1334, opened 2017-05-25
tests: use the fake apt cache in deb unit tests

elopio:


snapcraft/tests/fixture_setup.py
```diff
@@ -567,58 +567,104 @@ def setUp(self):
...
-        providing_packages = []
-        for package in self.cache:
-            if package_name in self.cache[package].provides:
-                providing_packages.append(self.cache[package])
-        return providing_packages
+        # Add all the packages in the manifest.
+        with open(os.path.abspath(
+                os.path.join(
+                    __file__, '..', '..',
+                    'internal', 'repo', 'manifest.txt'))) as manifest_file:
```
kalikiana: Any chance we can avoid reading this file and just use _manifest_dep_names? To avoid hard-coding the name in another place.

snapcraft/tests/repo/test_deb.py
```diff
@@ -219,28 +220,30 @@ def test_sources_armhf_trusty(self, mock_cc):
...
+        self.fake_apt_cache = fixture_setup.FakeAptCache()
+        self.useFixture(self.fake_apt_cache)
+        self.test_packages = (
+            'package-not-installed', 'package-installed',
+            'another-uninstalled', 'another-installed', 'repeated-package',
+            'repeated-package', 'versioned-package=0.2', 'versioned-package')
+        for package in self.test_packages:
+            self.fake_apt_cache.cache[package] = (
+                fixture_setup.FakeAptCachePackage(
+                    self.fake_apt_cache.path, package))
```
kalikiana: How about a method add_packages in FakeAptCache that just takes a list?

elopio: done.

snapcraft/tests/fixture_setup.py
elopio: I would prefer to keep this hardcoded here for now, instead of importing the ubuntu backend, or doing a selective import based on the architecture. 

The function to get the packages is in the Ubuntu object, but as you see from the path, this file doesn't mention it's specific to ubuntu. When we start supporting other backends, we will need to move the file and then make the selective import based on the architecture, and then we can use that method here.

PR #1336, opened 2017-05-29
HACKING: make running from virtualenv more obvious

kyrofa:
Snapcraft previously did not run from within a virtualenv, so we only recommended installing its dependencies via pip. This was always confusing, and these days snapcraft actually _does_ run from within a virtualenv, so let's make those directions more prominent in HACKING.md.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -1,8 +1,31 @@
 # Snapcraft
 
+## Installing from pip
+
+First install a few dependencies:
+
+    sudo apt install gcc g++ make python3-venv libffi-dev libsodium-dev libapt-pkg-dev libarchive13
+
+Create and activate a new virtual environment:
+
+    python3 -m venv path/to/venv
+    source path/to/venv/bin/activate
+
+
+Make sure pip is up-to-date:
+
+    (venv) $ pip install --upgrade pip
+
+
+Install snapcraft (and its dependencies):
+
+    (venv) $ pip install -r requirements.txt .
+
+
 ## Running
 
 To see all the commands and options, run `snapcraft --help`.
+
 
 ## Hacking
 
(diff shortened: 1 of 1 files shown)


HACKING.md
```diff
@@ -14,6 +37,35 @@ We'd love the help!
 - We can also be found on the #snappy IRC channel on Freenode
 
 
+### Installing for development
+
+You'll need to install the development dependencies, and you'll also probably want to install it in `editable` mode so your changes actually do something:
```
elopio: I would write something like: `so you can use your changes to the snapcraft source`..

HACKING.md
```diff
@@ -14,6 +37,35 @@ We'd love the help!
...
+
+    sudo apt install squashfs-tools xdelta3 bzr git mercurial subversion
+
+To run all the tests execute:
+
+    ./runtests.sh
+
+You can selectively run a selective group of tests like:
+
+    ./runtests.sh [static|unit|integration|snaps]
```
elopio: [static|unit|integration|plugins|store|snaps]

HACKING.md
```diff
@@ -1,9 +1,32 @@
...
+## Installing from pip
+
+First install a few dependencies:
+
+    sudo apt install gcc g++ make python3-venv libffi-dev libsodium-dev libapt-pkg-dev libarchive13
+
+Create and activate a new virtual environment:
+
+    python3 -m venv path/to/venv
+    source path/to/venv/bin/activate
```
kalikiana: How about just "." as the path? That's easier for someone who doesn't normally use venv, others will probably know.

kalikiana: Correction: . is a bad idea. Something like ../.venv is much better because it won't clutter your source tree.

HACKING.md
kyrofa: Wait a minute... YOU wrote that!

HACKING.md
kyrofa: Take another look-- what do you think of that?

HACKING.md
kyrofa: Done. You owe me.

HACKING.md
kyrofa: I tweaked it a bit, take a look and tell me what you think.

PR #1339, opened 2017-06-01
python plugin: install six before using setuptools

kyrofa:
setuptools v36.0.0 (released today) requires that six is installed before it can be used. This PR fixes LP: [#1694945](https://bugs.launchpad.net/snapcraft/+bug/1694945) by doing exactly that.


kyrofa:
Note that this PR contains no tests as our integration tests actually fail _without_ this patch (at least `plugins/test_python_plugin.py`, probably others).

PR #1340, opened 2017-06-01
state: save all the build packages as global

elopio:
We install all the build packages before pulling the individual parts.
Before this branch, we first installed all the build packages, then pulled each individual part, and then the individual parts saved the versions of packages installed inspecting the dependency chain in the cache.

With this change, we save all the build packages and their versions when they are installed. So this changes a few things:

 * If a build package is already installed, it won't be saved as an asset. So in the end, the recorded annotated snapcraft.yaml will be accurate only in cleanbuild, and requires that we also record the image used during cleanbuild.
 * The recorded annotated snapcraft.yaml will be slightly different than the original, with all the build packages for all the parts as global packages. As we install all the build packages before any part anyways, this is equivalent.

### DIFF
--- a/snapcraft/internal/lifecycle.py
+++ b/snapcraft/internal/lifecycle.py
@@ -111,8 +111,9 @@ def execute(step, project_options, part_names=None):
     config = snapcraft.internal.load_config(project_options)
     installed_packages = repo.Repo.install_build_packages(
         config.build_tools)
-    os.makedirs('snap/.snapcraft', exist_ok=True)
-    with open('snap/.snapcraft/state', 'w') as f:
+    snapcraft_state_dir = os.path.join('snap', '.snapcraft')
+    os.makedirs(snapcraft_state_dir, exist_ok=True)
+    with open(os.path.join(snapcraft_state_dir, 'state'), 'w') as f:
         f.write(yaml.dump(states.GlobalState(installed_packages)))
 
     if (os.environ.get('SNAPCRAFT_SETUP_CORE') and
--- a/snapcraft/internal/lifecycle.py
+++ b/snapcraft/internal/lifecycle.py
@@ -46,6 +46,8 @@ from snapcraft.internal.project_loader import replace_attr
 logger = logging.getLogger(__name__)
 
 
+_SNAPCRAFT_INTERNAL_DIR = os.path.join('snap', '.snapcraft')
+
 _TEMPLATE_YAML = """name: my-snap-name # you probably want to 'snapcraft register <name>'
 version: '0.1' # just for humans, typically '1.2+git' or '1.3.2'
 summary: Single-line elevator pitch for your amazing snap # 79 char long summary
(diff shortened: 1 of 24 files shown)


snapcraft/internal/lifecycle.py
```diff
@@ -108,7 +109,11 @@ def execute(step, project_options, part_names=None):
     :returns: A dict with the snap name, version, type and architectures.
     """
     config = snapcraft.internal.load_config(project_options)
-    repo.Repo.install_build_packages(config.build_tools)
+    installed_packages = repo.Repo.install_build_packages(
+        config.build_tools)
+    os.makedirs('snap/.snapcraft', exist_ok=True)
+    with open('snap/.snapcraft/state', 'w') as f:
```
kyrofa: Please kindly use `os.path.join` here. You're a wonderful person, and your work is appreciated.

snapcraft/internal/lifecycle.py
```diff
@@ -108,7 +109,11 @@ def execute(step, project_options, part_names=None):
     :returns: A dict with the snap name, version, type and architectures.
     """
     config = snapcraft.internal.load_config(project_options)
-    repo.Repo.install_build_packages(config.build_tools)
+    installed_packages = repo.Repo.install_build_packages(
+        config.build_tools)
```
kyrofa: Make sure we raise an exception here if this returns `None`. Thank you.

snapcraft/internal/lifecycle.py
elopio: Done. But you have to practice more to make it sound natural.

snapcraft/internal/lifecycle.py
elopio: done.

elopio:
@sergiusens all green!

snapcraft/internal/lifecycle.py
```diff
@@ -108,7 +109,15 @@ def execute(step, project_options, part_names=None):
...
+    if installed_packages is None:
+        raise ValueError(
+            'The repo backed is not returning the list of installed packages')
```
sergiusens: backend

snapcraft/internal/states/_state.py
```diff
@@ -81,6 +86,11 @@ def _get_differing_keys(dict1, dict2):
     return differing_keys
 
 
+def get_global_state():
+    with open(os.path.join('snap', '.snapcraft', 'state'), 'r') as state_file:
```
sergiusens: Is this cleaned up on `snapcraft clean` ?

sergiusens: does this make itself into the snap?

elopio: No, and yes. Good catches, but agh, because they are not real lifecycle steps I can't find a good place for their unit tests. I will put them in lifecycle.

elopio:
@sergiusens from the discussion in the forum, my understanding is that we will keep the states of each part in the parts directory.
So we will have `snap/.snapcraft/state` and `snap/parts/my-part/state/...`.

snapcraft/internal/states/_state.py
elopio: @sergiusens done: https://github.com/snapcore/snapcraft/pull/1340/commits/3c28849d0d1c974843d1d684f9502af9e6b6bd07
I couldn't find better locations, so please check if they make sense to you.

PR #1343, opened 2017-06-01
go plugin: Cross compile with CGo

kalikiana:
Support cross-compilation when using _CGo_ in a **go** part.

Follow-up on [go plugin: Add support for cross-compilation](https://github.com/snapcore/snapcraft/pull/1338).


integration_tests/plugins/test_go_plugin.py
```diff
@@ -68,3 +68,14 @@ def test_cross_compiling(self):
         binary = os.path.join(self.parts_dir, 'go-hello', 'install', 'bin',
                               os.path.basename(self.path))
         self.assertThat(binary, HasArchitecture('aarch64'))
+
+    def test_cross_compiling_with_cgo(self):
+        if snapcraft.ProjectOptions().deb_arch != 'amd64':
+            self.skipTest('The test only handles amd64 to armhf')
```
elopio: fixed.

PR #1344, opened 2017-06-01
git: don't use --remote for updating submodules

Saviq:
Remote in updating submodules means the tip of the remote branch is
checked out, not the commit recorded in the superproject.

This means the project loses control over what content is actually
checked out.

LP: [#1694765](https://bugs.launchpad.net/snapcraft/+bug/1694765)


sergiusens:
I do want to note that using `--force` will wipe out your local changes on the submodules, perhaps `--remote` did this too and was a hidden thing so this is still good.

PR #1346, opened 2017-06-02
repo: multi-arch build deps

kalikiana:
Figure out dependencies in a multi-arch compatible way using apt-get build-dep.

Follow-up on [go plugin: Cross compile with CGo](https://github.com/snapcore/snapcraft/pull/1343)


elopio:
@kalikiana please resolve the conflics.

snapcraft/tests/fixture_setup.py
```diff
@@ -411,6 +413,119 @@ def call_effect(*args, **kwargs):
...
+        Reading state information...''')
+    _PROBLEMS = dedent('''
+        Some packages could not be installed. This may mean that you have
+        requested an impossible situation or if you are using the unstable
+        distribution that some required packages have not yet been created
+        or been moved out of Incoming.
+        The following information may help to resolve the situation:
+
+        The following packages have unmet dependencies:
+         builddeps:{}{} : {}
```
kyrofa: I typically find that it's clearer to name these when used in a larger template like this. Otherwise changes become quite fragile.

snapcraft/internal/repo/_deb.py
```diff
@@ -209,20 +211,79 @@ def install_build_packages(cls, package_names):
...
+    def _get_build_deps(cls, package_names, arch):
+        """Use apt-get build-dep with a fake source package file to find
+           all dependencies and correct architectures.
+        """
+        build_deps = []
+        if not package_names:
+            return build_deps
+
+        with tempfile.NamedTemporaryFile(suffix='.dsc') as fake_source:
+            depends = 'Build-Depends: {}\n'.format(', '.join(package_names))
```
kyrofa: Can you explain how this has different results than using something like [apt_pkg.parse_depends()](https://apt.alioth.debian.org/python-apt-doc/library/apt_pkg.html#apt_pkg.parse_depends)?

Conan-Kudo: This seems horrifically wrong. Parsing the Debian Source Control file directly should be unnecessary, as APT already has this information...

kalikiana: The code generates the .dsc files based on build-packages and calls apt-get build-dep - there's nothing to parse here.

snapcraft/tests/repo/test_deb.py
```diff
@@ -265,31 +265,24 @@ def setUp(self):
     def get_installable_packages(self, packages, target_arch=''):
         installable = []
         for pkg in packages:
-            if not packages[pkg].installed:
```
elopio: I'm not sure I understand why you removed this check. If a package is already installed, it shouldn't be marked as installable.

elopio: Nevermind, I was looking at the wrong deb.

PR #1347, opened 2017-06-02
cli: do not duplicate errors

sergiusens:
The removal of sys.exit when doing `load_config` had a side effect of printing errors twice. The work here does a bit of cleanup so this mistake does not happen again.

LP: #1686557


snapcraft/internal/errors.py
```diff
@@ -82,6 +78,35 @@ class SnapcraftPartMissingError(SnapcraftError):
...
+
+    fmt = 'Issue detected while analyzing snapcraft.yaml: {message}'
+
+    def __init__(self, message):
+        super().__init__(message=message)
+
+
+class PluginError(SnapcraftError):
+
+    fmt = "Issue while loading part: {message}"
```
elopio: single quotes please

snapcraft/internal/errors.py
```diff
@@ -82,6 +78,35 @@ class SnapcraftPartMissingError(SnapcraftError):
...
+
+    fmt = ("Issues while validating snapcraft.yaml: the 'plugin' keyword is "
+           "missing for the {part_name} part.")
+
+
+class SnapcraftYamlFileError(SnapcraftError):
+
+    fmt = ("Could not find {snapcraft_yaml}. Are you sure you are in the "
+           "right directory?\n"
+           "To start a new project, use `snapcraft init`")
```
elopio: single quotes please.

snapcraft/internal/errors.py
```diff
@@ -82,6 +78,35 @@ class SnapcraftPartMissingError(SnapcraftError):
     )
 
 
+class SnapcraftLogicError(SnapcraftError):
```
elopio: `logic` is a bad name. But this was not introduced in this branch and I can't think now of a better name. So just to keep in mind, let's try to find a better one.

elopio:
Nice refactor, thanks.

```$ snapcraft
Issues while validating snapcraft.yaml: The 'grade' property does not match the required schema: 'invalid' is not one of ['stable', 'devel']
The 'grade' property does not match the required schema: 'invalid' is not one of ['stable', 'devel']
$ git checkout sergiusens/double-double-error
$ /home/ubuntu/workspace/canonical/snapcraft/bin/snapcraft
Issues while validating snapcraft.yaml: The 'grade' property does not match the required schema: 'invalid' is not one of ['stable', 'devel']
```

PR #1350, opened 2017-06-02
rust plugin: Add support for cross-compilation

kalikiana:
Support doing `snapcraft snap --target-arch=` with parts using the **Rust** plugin


integration_tests/plugins/test_rust_plugin.py
```diff
@@ -91,6 +92,17 @@ def test_stage_rust_with_source_and_source_subdir(self):
             os.path.join('parts', 'rust-subdir', 'src', 'subdir',
                          'Cargo.lock'), FileExists())
 
+    def test_cross_compiling(self):
+        if snapcraft.ProjectOptions().deb_arch != 'amd64':
+            self.skipTest('The test only handles amd64 to armhf')
```
elopio: s/armhf/arm64

snapcraft/plugins/rust.py
```diff
@@ -102,6 +103,40 @@ def build(self):
             cmd.append(' '.join(self.options.rust_features))
         self.run(cmd, env=self._build_env())
 
+    def _write_config(self):
+        if not self.project.is_cross_compiling:
+            return
+
+        # Cf. http://doc.crates.io/config.html
+        with suppress(FileExistsError):
+            os.mkdir(os.path.join(self.builddir, '.cargo'))
```
elopio: I prefer os.makedirs(..., exist_ok=True). Was it @kyrofa who prefered suppress? Any reason for that?

snapcraft/plugins/rust.py
```diff
@@ -102,6 +103,40 @@ def build(self):
...
+        with open(os.path.join(self.builddir, '.cargo', 'config'), 'w') as f:
```
elopio: a nit, I would put the cargo_dir in a variable, so there's no need to do the os.path.join again.

snapcraft/plugins/rust.py
```diff
@@ -93,6 +91,9 @@ def __init__(self, name, options, project):
 
     def build(self):
         super().build()
+
+        self._write_config()
```
elopio: This is only for cross-compiling right? What do you think about naming it _write_crosscompile_config?

snapcraft/plugins/rust.py
```diff
@@ -147,9 +188,12 @@ def _fetch_rust(self):
...
             os.makedirs(self._rustpath)
         self._rustup_get.download()
-        self.run([self._rustup,
-                  '--prefix={}'.format(self._rustpath),
-                  '--disable-sudo', '--save'] + options)
+        cmd = [self._rustup,
+               '--prefix={}'.format(self._rustpath),
+               '--disable-sudo', '--save'] + options
+        if self.project.is_cross_compiling:
+            cmd.append("--with-target={}".format(self._target))
```
elopio: single quotes please.

snapcraft/tests/matchers.py
```diff
@@ -30,8 +30,12 @@ def __str__(self):
...
-                    self._expected_arch, arch))
+        try:
+            magic = self._ms.file(file_path)
+            arch = magic.split(',')[1]
+            if self._expected_arch not in arch:
+                return testtools.matchers.Mismatch(
+                    'Expected {!r} to be in {!r}'.format(
+                        self._expected_arch, arch))
+        except Exception as e:
+            raise ValueError('Unexpected magic {!r}'.format(magic)) from e
```
elopio: I'm not understanding this exception. What does it mean to have an unexpected magic?

integration_tests/plugins/test_rust_plugin.py
```diff
@@ -91,6 +92,17 @@ def test_stage_rust_with_source_and_source_subdir(self):
             os.path.join('parts', 'rust-subdir', 'src', 'subdir',
                          'Cargo.lock'), FileExists())
 
+    def test_cross_compiling(self):
+        if snapcraft.ProjectOptions().deb_arch != 'amd64':
+            self.skipTest('The test only handles amd64 to arm64')
+
+        target_arch = 'arm64'
```
kyrofa: Why is this a variable if you only use it once?

snapcraft/plugins/rust.py
```diff
@@ -102,6 +104,39 @@ def build(self):
...
+        # Cf. rustc --print target-list
+        rust_targets = {
+            'armhf': 'armv7-{}-{}eabihf',
+            'arm64': 'aarch64-{}-{}',
+            'i386': 'i686-{}-{}',
+            'amd64': 'x86_64-{}-{}',
+            'ppc64el': 'powerpc64le-{}-{}',
+        }
+        self._target = rust_targets.get(self.project.deb_arch).format(
+            'unknown-linux', 'gnu')
```
kyrofa: What's the point of making all of those templates if this is the only option? Why not just add this into the list?

snapcraft/plugins/rust.py
```diff
@@ -102,6 +104,39 @@ def build(self):
...
+        if not self._target:
```
kyrofa: If there isn't an entry in the above dict for `self.project.deb_arch`, you'll never get here-- the use of `format` will cause `'NoneType' object has no attribute 'format'`. If you get rid of the format this will work, though.

snapcraft/plugins/rust.py
```diff
@@ -130,6 +165,12 @@ def clean_pull(self):
         if os.path.exists(self._rustpath):
             shutil.rmtree(self._rustpath)
 
+    def clean_build(self):
+        super().clean_build()
+
+        with suppress(FileExistsError):
+            shutil.rmtree(self._cargo_dir)
```
kyrofa: Wait... why would `rmtree` raise `FileExistsError`? If a file is in use? That sounds like a situation where we shouldn't suppress. Curious what's happening here.

snapcraft/tests/matchers.py
```diff
@@ -30,7 +30,13 @@ def __str__(self):
...
 
     def match(self, file_path):
-        arch = self._ms.file(file_path).split(',')[1]
+        magic = self._ms.file(file_path)
+        # Catch exceptions on splitting the string to provide context
+        # This includes "cannot open `...' (No such file or directory)"
+        try:
+            arch = magic.split(',')[1]
+        except IndexError as e:
+            raise ValueError('Failed to parse magic {!r}'.format(magic)) from e
```
kyrofa: Thank you for this, very nice.

snapcraft/plugins/rust.py
kalikiana: - Making every single target twice as long adds noise and makes it less easy to see what's relevant.
- If in the future we care about musl we can replace it easily.

ChrisMacNaughton:
Been watching this one, code looks OK but I don't have any ARM hardware to test on ;-)

elopio:
Thanks @ChrisMacNaughton!
@cholcombe973 would you like to give it a look too?

We want to get more expert knowledge into our plugins, so if the two of you could help us with the rust plugin it would be great.

snapcraft/plugins/rust.py
```diff
@@ -102,6 +104,39 @@ def build(self):
...
+                           '{}-gcc'.format(self.project.arch_triplet)))
+
+    def enable_cross_compilation(self):
+        # Cf. rustc --print target-list
+        targets = {
+            'armhf': 'armv7-{}-{}eabihf',
+            'arm64': 'aarch64-{}-{}',
+            'i386': 'i686-{}-{}',
+            'amd64': 'x86_64-{}-{}',
+            'ppc64el': 'powerpc64le-{}-{}',
```
cholcombe973: rustc --print target-list on ubuntu 16.04 x86_64 produces 64 targets for me.  Are these the only ones we want to support?

snapcraft/tests/matchers.py
cholcombe973: :+1:

manual-tests.md
```diff
@@ -76,6 +76,15 @@
 5. Run `go-hello`.
 
 
+# Test cross-compilation with Rust
+
+1. Go to integration_tests/snaps/rust-hello.
+2. Run `snapcraft snap --target-arch=armhf`.
+3. Copy the snap to a Raspberry Pi.
```
cholcombe973: We could also probably help users without a Raspberry Pi test with qemu instructions.

snapcraft/plugins/rust.py
```diff
@@ -76,9 +77,6 @@ def get_build_properties(cls):
     def __init__(self, name, options, project):
         super().__init__(name, options, project)
         self.build_packages.extend([
```
cholcombe973: I think we need the C cross toolchain here and the cross compiled std crates no?

snapcraft/tests/plugins/test_rust.py
```diff
@@ -15,16 +15,113 @@
...
+        patcher = mock.patch.dict(os.environ, {})
+        self.env_mock = patcher.start()
+        self.addCleanup(patcher.stop)
+
+    def test_cross_compile(self):
+        plugin = rust.RustPlugin('test-part', self.options,
+                                 self.project_options)
+        os.makedirs(plugin.sourcedir)
+
+        plugin.enable_cross_compilation()
```
cholcombe973: This needs to catch NotImplementedError

manual-tests.md
cholcombe973: We could also probably help users without a Raspberry Pi test with qemu instructions.

snapcraft/plugins/rust.py
cholcombe973: I think we need the C cross toolchain here and the cross compiled std crates no?
Also I think @ChrisMacNaughton  did you need these libs for the alacrity snap you did awhile back?

kalikiana: I attempted to removed redundant packages here. Unfortunately our integration tests are not 100% trustworthy because **gcc** will implicitly be present on CI I verified this again in a LXD container and gcc is needed, the other packages aren't - at least for the test packages in snapcraft, if there's others we need to add new tests.

manual-tests.md
kalikiana: I was talking about this with @elopio a couple of weeks ago. We could consider an integration test using qemu. But we can't run the test snap on a native armhf machine. So the manual test is specifically for this.

snapcraft/tests/plugins/test_rust.py
kalikiana: Good catch (no pun intended). I added a separate test for that.

manual-tests.md
cholcombe973: I see.  Ok I was just wondering

PR #1352, opened 2017-06-05
Allow source-type to specify local

jocave:
Allows local overriding of a source-type to local

LP: #1695886


sergiusens:
Hey, thanks for this, if no test can we at least get a reason for needing this?

jocave:
Sure, I put a reason on the LP bug, but will reproduce here.

"The recent change to the schema that added an enumeration of valid 'source-types' means that now you cannot specify source-type to be local."

"The particular use case this allows is when you have a remote part with one of the currently valid source types e.g. git, and you want to fast iterate on the part without the requirement to check in to a repo (even one on your local file system). It was previously possible to override the source-type to local by putting any non-valid value in your local snapcraft.yaml source-type field whilst keeping all the other useful stuff like build-packages etc."

sergiusens:
Ah, we certainly need a unit test at least for this so we do not regress again.

elopio:
@EduardoVega this is related to the code you touched. If you have some spare minutes, would you mind reviewing it after @jocave adds the test?

EduardoVega:
Hi @elopio , @sergiusens and @jocave . The change looks good. In fact it makes sense to have the `local` source-type option, so in case it is used, snapcraft will avoid the process to identify the handler for local sources.

snapcraft/tests/sources/test_deb.py
```diff
@@ -60,3 +61,7 @@ def test_extract_and_keep_debfile(self):
 
         with open(deb_download, 'r') as deb_file:
             self.assertEqual('Test fake compressed file', deb_file.read())
+
+    def test_has_source_handler_entry_on_linux(self):
+        if sys.platform == 'linux':
```
elopio: I find it nicer to use skip unless: https://docs.python.org/3/library/unittest.html#unittest.skipUnless

snapcraft/tests/sources/test_rpm.py
```diff
@@ -61,3 +62,9 @@ def test_extract_and_keep_rpmfile(self):
 
         test_output_files = ['test.txt', rpm_file_name]
         self.assertCountEqual(os.listdir(dest_dir), test_output_files)
+
+    def test_has_source_handler_entry_on_linux(self):
+        if sys.platform == 'linux':
+            self.assertTrue(sources._source_handler['rpm'] is sources.Rpm)
+        else:
+            self.assertRaises(KeyError, sources._source_handler['rpm'])
```
elopio: You are doing it differently here than in the deb case. Any reason?

jocave:
@elopio sure, I can add an integration test.

snapcraft/tests/sources/test_rpm.py
jocave: No, I meant for both to be like the rpm case. Opened atom to find not all files had been saved though!

integration_tests/snaps/local-source-type/snapcraft.yaml
```diff
@@ -0,0 +1,12 @@
+name: test-package
+version: 0.1
+summary: one line summary
+description: a longer description
+icon: icon.png
```
kyrofa: Please toast the icon, it's not necessary anymore (yeah I know, some of our older integration tests still have them, not your fault).

jocave: @kyrofa no problem, it's done...

PR #1353, opened 2017-06-05
cli: default options for the implicit snap command

sergiusens:
The `snap` command is an implicit command, but when used the options it carries cannot be used. We add the default options to the command as hidden ones to enable this again.


integration_tests/test_snap.py
```diff
@@ -149,6 +149,14 @@ def test_snap_from_snapcraft_init(self):
 
         self.run_snapcraft('snap')
 
+    def test_implicit_command_with_arch(self):
+        self.assertThat('snapcraft.yaml', Not(FileExists()))
+        self.run_snapcraft('init')
+        self.assertThat(os.path.join('snap', 'snapcraft.yaml'), FileExists())
```
elopio: No need for these two assertions, IMO. We start in a temp home, so the directory will be empty. And if init fails, the following command would fail too.

snapcraft/cli/__init__.py
```diff
@@ -82,8 +83,9 @@ def get_command(self, ctx, cmd_name):
 
 @click.group(cls=SnapcraftGroup, invoke_without_command=True)
 @click.pass_context
+@add_build_options(hidden=True)
 @click.option('--debug', '-d', is_flag=True)
-def run(ctx, debug, catch_exceptions=False):
+def run(ctx, debug, catch_exceptions=False, **kwargs):
```
elopio: I missed this in the previous review, but I'll mention it now. `ctx` is a bad name, `context` is a lot better.

integration_tests/test_snap.py
sergiusens: removed

PR #1354, opened 2017-06-06
catkin plugin: fix pythonpath for catkin_find

mrjogo:
ROS has separate python package location that needs to be on PYTHONPATH before calling catkin_find in the generated bash script. Fixes false script failures, which the plugin interprets as not finding the specified package.

LP: [#1696014](https://bugs.launchpad.net/snapcraft/+bug/1696014)

### DIFF
--- a/snapcraft/plugins/catkin.py
+++ b/snapcraft/plugins/catkin.py
@@ -617,7 +617,6 @@ def _find_system_dependencies(catkin_packages, rosdep, catkin):
                 try:
                     catkin.find(dependency)
                 except CatkinPackageNotFoundError:
-                    logger.info("No underlay package: {}".format(dependency))
                     # No package by that name is available
                     pass
                 else:
--- a/snapcraft/plugins/catkin.py
+++ b/snapcraft/plugins/catkin.py
@@ -889,9 +889,11 @@ class _Catkin:
 
     def _run(self, arguments):
         with tempfile.NamedTemporaryFile(mode='w+') as f:
-            lines = ['export PYTHONPATH={}'.format(os.path.join(
-                self._catkin_install_path, 'usr', 'lib', 'python2.7',
-                'dist-packages'))]
+            lines = ['export PYTHONPATH={}:{}'.format(
+                os.path.join(self._catkin_install_path, 'usr', 'lib',
+                             'python2.7', 'dist-packages'),
+                os.path.join(self._catkin_install_path, 'opt', 'ros',
+                             'kinetic', 'lib', 'python2.7', 'dist-packages'))]
 
             ros_path = os.path.join(
                 self._catkin_install_path, 'opt', 'ros', self._ros_distro)
--- a/snapcraft/plugins/catkin.py
+++ b/snapcraft/plugins/catkin.py
@@ -617,6 +617,7 @@ def _find_system_dependencies(catkin_packages, rosdep, catkin):
                 try:
                     catkin.find(dependency)
                 except CatkinPackageNotFoundError:
+                    logger.info("No underlay package: {}".format(dependency))
                     # No package by that name is available
                     pass
                 else:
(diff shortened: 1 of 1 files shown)


snapcraft/plugins/catkin.py
```diff
@@ -889,9 +889,11 @@ def find(self, package_name):
...
     def _run(self, arguments):
         with tempfile.NamedTemporaryFile(mode='w+') as f:
-            lines = ['export PYTHONPATH={}'.format(os.path.join(
-                self._catkin_install_path, 'usr', 'lib', 'python2.7',
-                'dist-packages'))]
+            lines = ['export PYTHONPATH={}:{}'.format(
+                os.path.join(self._catkin_install_path, 'usr', 'lib',
+                             'python2.7', 'dist-packages'),
+                os.path.join(self._catkin_install_path, 'opt', 'ros',
+                             'kinetic', 'lib', 'python2.7', 'dist-packages'))]
```
kyrofa: I need to do some testing to be sure, but I doesn't sourcing the `setup.sh` as done immediately below this add that to the `PYTHONPATH`? But making this change seems to solve your issue?

kyrofa: This is odd: this change solves your issue? I need to do some testing to be sure, but I swear that sourcing the `setup.sh` as done immediately below this adds that to the PYTHONPATH.

Note also that this hard-codes `kinetic`, it should be using the `ros_path` defined below instead.

mrjogo: I may have jumped the gun a little. While this change does fix the error I was seeing, it's still not finding the underlay packages, which I think is because the _setup_util.py in my underlay is still pointing to the directory it was dumped from. It would make sense that if that were properly sourced it would find the catkin package. I'll report back once I get everything working.

I submitted the CLA just before making the pull request.

mrjogo: Ok, so you are correct that when catkin is part of the underlay, _Catkin.find works as intended. However, I am including the underlay in the snap (to use custom, pre-built tarballs of some ROS packages), and when catkin is part of the underlay, it creates a staging conflict because the catkin deb (a different version) is automatically installed as part of the catkin plugin.

What should be the correct way to build a single snap with an underlay?

kyrofa: @mrjogo I'm sorry, but I'm having trouble understanding what you're doing and what's wrong. Is this https://forum.snapcraft.io/t/python-pkg-resources-only-works-in-certain-paths/978 ? If not, perhaps it would be best to start a forum conversation about this, with some example code?

kyrofa:
I just realized the comment I made is contained in a hidden diff, so I'll make it again here:

@mrjogo I'm sorry, but I'm having trouble understanding what you're doing and what's wrong. Is this https://forum.snapcraft.io/t/python-pkg-resources-only-works-in-certain-paths/978 ? If not, perhaps it would be best to start a forum conversation about this, with some example code?

kyrofa:
Closing this due to inactivity. When you have some more time, let's please discuss this further in the forum so I can understand the problem.

mrjogo:
Sorry for disappearing. This did not have to do with the pkg_resources thread, however, it turns out this pull request fixed a problem very specific to my set up where https://github.com/ros/catkin/blob/b05cd262f4aa992856d4470cfc2b1f5e52fabefc/bin/catkin_find#L9 failed because the catkin python module was installed in a different place. Therefore, this pull request is no longer relevant.

I'm still having conflict issues with underlays, but not because of this, so I'll start a separate forum conversation. Thanks for your patience.

mrjogo:
Unfortunately I spoke too soon and this is still a problem. See discussion https://forum.snapcraft.io/t/problems-building-using-catkin-plugin-and-underlays/1913 and demonstration code in the bug report.

PR #1355, opened 2017-06-06
go plugin: filter the main packages go-packages are defined

sergiusens:
When no go packages are defined, `go build -o` is incompatible with `./...` so we need to provide it the main packages we want to build.

### DIFF
--- a/integration_tests/plugins/test_go_plugin.py
+++ b/integration_tests/plugins/test_go_plugin.py
@@ -39,11 +39,11 @@ class GoPluginTestCase(integration_tests.TestCase):
     def test_building_multiple_main_packages(self):
         self.run_snapcraft('stage', 'go-with-multiple-main-packages')
 
-    def test_building_multiple_main_packages_without_go_packages(self):
-        self.copy_project_to_cwd('go-with-multiple-main-packages')
-
         for bin in ['main1', 'main2', 'main3']:
             self.assertThat(os.path.join('stage', 'bin', bin), FileExists())
+
+    def test_building_multiple_main_packages_without_go_packages(self):
+        self.copy_project_to_cwd('go-with-multiple-main-packages')
 
         snapcraft_yaml_file = 'snapcraft.yaml'
         with open(snapcraft_yaml_file) as f:
--- a/integration_tests/plugins/test_go_plugin.py
+++ b/integration_tests/plugins/test_go_plugin.py
@@ -17,6 +17,9 @@
 import os
 import subprocess
 
+import yaml
+from testtools.matchers import FileContains, FileExists, Not
+
 import integration_tests
 import snapcraft
 from snapcraft.tests.matchers import HasArchitecture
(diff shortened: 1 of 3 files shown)


elopio:
@sergiusens I tried building the godd demo snap with your branch and it still fails.
It would be good to start with an integration test of the most simple go snap that fails, to then confirm the fix.

sergiusens:
oh @elopio, but I was fixing gopaste with this.

sergiusens:
the case where go-packages are not defined nor goimportpath

snapcraft/plugins/go.py
```diff
@@ -144,6 +144,15 @@ def _get_local_go_package(self):
...
 
+    def _get_local_main_packages(self):
+        search_path = './{}/...'.format(self._get_local_go_package())
+        packages = self._run_output(['go', 'list', '-f',
+                                     '{{.ImportPath}} {{.Name}}',
+                                     search_path])
+        packages_split = [p.split() for p in packages.splitlines()]
+        main_packages = [p[0] for p in packages_split if p[1] == 'main']
+        return main_packages
+
```
kalikiana: I like the use of -f here to get sensible output - do we care for a pretty error when the list is empty? This would right now result in "No such file or directory" which isn't necessarily clear.

integration_tests/plugins/test_go_plugin.py
```diff
@@ -36,6 +39,25 @@ def test_stage_go_plugin(self):
     def test_building_multiple_main_packages(self):
         self.run_snapcraft('stage', 'go-with-multiple-main-packages')
 
+    def test_building_multiple_main_packages_without_go_packages(self):
+        self.copy_project_to_cwd('go-with-multiple-main-packages')
+
+        for bin in ['main1', 'main2', 'main3']:
+            self.assertThat(os.path.join('stage', 'bin', bin), FileExists())
```
elopio: You are checking that the stage dir has contents but you haven't called the snapcraft command yet.
`testtools.matchers._impl.MismatchError: stage/bin/main1 does not exist.`

snapcraft/plugins/go.py
```diff
@@ -144,6 +144,15 @@ def _get_local_go_package(self):
             go_package = os.path.basename(os.path.abspath(self.options.source))
         return go_package
 
+    def _get_local_main_packages(self):
```
elopio: could you please move this definition after it was called the first time?

after `def build`

snapcraft/plugins/go.py
```diff
@@ -181,6 +190,10 @@ def _run(self, cmd, **kwargs):
         env = self._build_environment()
         return self.run(cmd, cwd=self._gopath_src, env=env, **kwargs)
 
+    def _run_output(self, cmd, **kwargs):
```
elopio: what about calling this _run_output_with_build_env?

sergiusens: I would rather change this for all plugins in one sweep as this follows the pattern we've been using started by @kyrofa. A bug would come in handy :wink:

snapcraft/plugins/go.py
sergiusens: wait, during the sprint we said before!

integration_tests/plugins/test_go_plugin.py
sergiusens: oh, this was a bad copy paste, I wanted to add the check to the previous test as well :confused:

snapcraft/plugins/go.py
sergiusens: I still prefer to keep it close to the other method defined just above.

elopio: no, the idea is to read the source code from top to bottom, you start with a high level public method, and following it you define all its helpers. So you can choose if you want a high-level view, or dig into the details. If you first find an internal helper, you don't know the context so you have to jump throw the code finding where it is called in order to understand it.

elopio: right, the other method is also not following the top-to-bottom rule.

PR #1356, opened 2017-06-06
common: find data files via sys.prefix

kyrofa:
Currently Snapcraft hard-codes where it looks for its data files (schema, etc.) to support either running from a development tree or running from the Debian package. This actually works for when its installed via pip as well, but only if using an editable installation.

This PR fixes LP: [#1656884](https://bugs.launchpad.net/snapcraft/+bug/1656884) by using sys.prefix to also support being installed via pip (without being editable).

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -4,7 +4,7 @@
 
 First install a few dependencies:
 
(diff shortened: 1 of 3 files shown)


kyrofa:
Sure, manual test added.

PR #1359, opened 2017-06-08
tests: replace grub-doc with haskell-doc in build-packages tests

elopio:
Grub-docs was removed from artful. We need to test using a different package that will work on all the supported versions.


kyrofa:
This test may be better suited to only stable (i.e. released) Ubuntu versions.

sergiusens:
Agree

elopio:
@sergiusens @kyrofa I tested this in xenial and artful. So to not slow things down more, I would prefer to add the skipunless(xenial) later.

kyrofa:
That sounds fine.

PR #1361, opened 2017-06-09
pluginhandler: don't clobber path on local import failure

zyga:
The helper for importing local plugins would clobber sys.path if the
import failed. This is fixed with a simple try/finally statement.

Signed-off-by: Zygmunt Krynicki <zygmunt.krynicki@canonical.com>


elopio:
I added an assertion that fails without @zyga's fix. With this, I'm +1.

PR #1362, opened 2017-06-09
Only install golang-go if it is not found in PATH.

chrisglass:
Do not install golang-go if "go" is already present in the system's
PATH. For example, a user might specifically want a golang program to be
built with the go executable in their PATH and not the golang version
installed by the golang-go package.

This should solve the problem for people like in the linked bug, as well as (like me) people who want to specifically use golang 1.8 (and golang-go installs 1.7).

LP: [#1618720](https://bugs.launchpad.net/snapcraft/+bug/1618720)

### DIFF
--- a/snapcraft/plugins/go.py
+++ b/snapcraft/plugins/go.py
@@ -45,6 +45,7 @@ Additionally, this plugin uses the following plugin-specific keywords:
 import logging
 import os
 import shutil
+import subprocess
 from glob import iglob
 
 import snapcraft
@@ -101,7 +102,13 @@ class GoPlugin(snapcraft.BasePlugin):
 
     def __init__(self, name, options, project):
         super().__init__(name, options, project)
-        self.build_packages.append('golang-go')
+
+        # Do not try to install golang if it is already available in our PATH.
+        try:
+            subprocess.check_output(['go', 'version'])
+        except OSError:
+            self.build_packages.append('golang-go')
+
         self._gopath = os.path.join(self.partdir, 'go')
         self._gopath_src = os.path.join(self._gopath, 'src')
         self._gopath_bin = os.path.join(self._gopath, 'bin')


snapcraft/plugins/go.py
```diff
@@ -101,7 +102,13 @@ def get_pull_properties(cls):
 
     def __init__(self, name, options, project):
         super().__init__(name, options, project)
-        self.build_packages.append('golang-go')
+
+        # Do not try to install golang if it is already available in our PATH.
+        try:
+            subprocess.check_output(['go', 'version'])
+        except OSError:
+            self.build_packages.append('golang-go')
```
kalikiana: This will effectively "hide" the Go being used from the collected artifacts, won't it? Since there's no package with a specific version to track.
@elopio Do we have a story here yet?

chrisglass: My preferred solution would be *not* to install any packages here.

Odds are somebody building a snap from golang sources already has the appropriate golang toolchain set-up.

Preferred solution:
raise OSError("A golang compiler cannot be found. Please make sure an appropriate compiler can be found in your PATH.")

elopio: Thanks for your contribution @chrisglass. However, our plans seem to go in the opposite direction.

We want snapcraft to work on a clean xenial container, where go will not be installed. We are trying to default to build in containers always, and to make reproducible builds. Relying in the go package installed in the developer machine goes against this.

What we want is to support build-snaps, so you can specify on the yaml the snap with your toolchain. If your machine already has this snap installed, then nothing will happen. And this also encourages people to push to the store snaps for compilers and stuff like this.

I don't think we want to support a case for a developer installing his own toolchain from something that's not a snap, but this is open to debate of course. You can open a forum post to discuss with a wider audience.

Please let me know what do you think.

chrisglass: Well... ok, let's step back a second:

I'm trying to package a golang program in a snap. The program in question requires golang 1.8 to build. 

How do you suggest I achieve that, without this PR?

elopio: @chrisglass today, using the remote go part, like in here:
https://github.com/elopio/hashicorp-snaps/blob/master/vault/snap/snapcraft.yaml#L30

That it's a little slower because it compiles go. So in the near future, you would do it with the go snap published by mwhudson. Something like `build-snaps: [go]`.

ChrisMacNaughton: Could the go plugin not do what the Rust plugin does and allow the user to configure a specific version of the compiler to use?

chrisglass: @ChrisMacNaughton that would work for my use case at least, yes.

chrisglass: @elopio "a little slower" is a bit of an understatement :) Rebuilding my toolchain from source every time I want to package my golang application seems a little bit extreme, I think I'll just maintain a local fork of snapcraft with this patch until a more mature solution emerges.

Thanks for your time!

kyrofa:
As discussed, this doesn't quite fit the goals of snapcraft. Keep an eye out for the `build-snaps` feature! I'll go ahead and close this for now, but like @elopio said, please feel free to open a forum post to have a wider discussion about this.

PR #1364, opened 2017-06-13
lxd: Inject snapcraft and core snaps into the container

kalikiana:
When running snapcraft as a snap and building in a LXD container, currently snapcraft from the Xenial archives is used to do the actual build. This means a newer snap may be using an older snapcraft or vice versa. Therefore the same version that's already running on the host should be pushed into the container.

See also [the relevant forum discussion](https://forum.snapcraft.io/t/pushing-the-snapcraft-on-host-into-the-container/387).


snapcraft/internal/lxd.py
```diff
@@ -139,6 +141,56 @@ def _setup_project(self):
         self._push_file(tar_filename, dst)
         self._container_run(['tar', 'xvf', os.path.basename(tar_filename)])
 
+    def _inject_snapcraft(self):
+        if common.is_snap():
+            # Because of https://bugs.launchpad.net/snappy/+bug/1628289
+            self._container_run(['apt-get', 'install', 'squashfuse', '-y'])
+
+            # Push core snap into container
+            self._inject_snap('core', '99T7MUlRhtI3U0QFgl5mXXESAiSwt776')
```
elopio: please put this id in a constant variable.

snapcraft/internal/lxd.py
```diff
@@ -139,6 +141,56 @@ def _setup_project(self):
...
+        else:
+            self._container_run(['apt-get', 'install', 'snapcraft', '-y'])
+
+    def _inject_snap(self, name, id, *, rev=None, classic=False):
+        if not rev:
+            rev = check_output(['readlink',
+                                os.path.join(os.path.sep,
+                                             'var', 'snap',
+                                             name, 'current')]).decode()[:-1]
+        key = \
```
elopio: We prefer parentheses instead of \.
And by *we*, I mean *I*, because I'm not sure if we have discussed this nit before. Do you prefer \ for some reason?

snapcraft/internal/lxd.py
```diff
@@ -139,6 +141,56 @@ def _setup_project(self):
...
+            'BWDEoaqyr25nF5SNCvEv2v7QnM9QsfCc0PBMYD_i2NGSQ32EF2d4D0hqUel3m8ul'
```
elopio: I'm lost with assertions. Please move this to a constant var, and would it be too much to comment the variable with the way you found this value?

snapcraft/internal/lxd.py
kalikiana: We've talked about it in the context of lambdas, where parentheses don't work and I ended up changing the code completely. So there's no generic alternative. Parentheses work here, though, so I'm okay with that.

snapcraft/internal/lxd.py
```diff
@@ -139,6 +144,61 @@ def _setup_project(self):
...
+    def _inject_snap(self, name, *, rev=None):
+        session = requests_unixsocket.Session()
+        snapd_socket = '/run/snapd.socket'.replace('/', '%2F')
+        api = 'http+unix://{}/v2/snaps/{}'.format(snapd_socket, name)
+        try:
+            json = session.request('GET', api).json()
+        except Exception as e:
+            raise RuntimeError('Error connecting to {}'.format(api)) from e
+        if not json['status'] == 'OK':
+            raise RuntimeError('Error querying {} snap: {}'.format(name, json))
```
kyrofa: These are pretty snapcraft-internal exceptions. Should we be using an exception inheriting from `SnapcraftError`?

snapcraft/internal/lxd.py
```diff
@@ -139,6 +144,61 @@ def _setup_project(self):
...
+            # Because of https://bugs.launchpad.net/snappy/+bug/1628289
+            self._container_run(['apt-get', 'install', 'squashfuse', '-y'])
+
+            # Push core snap into container
+            self._inject_snap('core')
+            self._inject_snap('snapcraft', rev=os.environ.get('SNAP_REVISION'))
+        else:
+            self._container_run(['apt-get', 'install', 'snapcraft', '-y'])
+
+    def _inject_snap(self, name, *, rev=None):
```
kyrofa: If we can always get the rev from the snapd API, is there a reason we need to support this? We always want the currently-active one, right?

snapcraft/internal/lxd.py
```diff
@@ -139,6 +144,61 @@ def _setup_project(self):
...
+        except Exception as e:
+            raise RuntimeError('Error connecting to {}'.format(api)) from e
+        if not json['status'] == 'OK':
+            raise RuntimeError('Error querying {} snap: {}'.format(name, json))
+        id = json['result']['id']
+        classic = json['result']['confinement'] == 'classic'
+        if not rev:
+            rev = json['result']['revision']
+
+        self._inject_assertions('{}_{}.assert'.format(name, rev), [
```
kyrofa: As you know from the [forum post](https://forum.snapcraft.io/t/getting-the-revision-of-a-snap-from-snapd/989), there's a concern here where the snaps could very well auto-update away from the revision we want to use. There doesn't seem to be a way to disable that behavior, which is unfortunate. My current (admittedly not ideal) suggestion is to not use the assertions at all so the snaps aren't tied back to the store, and thus cannot update. What do you think about that?

snapcraft/internal/lxd.py
```diff
@@ -139,6 +144,61 @@ def _setup_project(self):
...
+            ['account-key', 'public-key-sha3-384={}'.format(_STORE_KEY)],
+            ['snap-declaration', 'snap-name={}'.format(name)],
+            ['snap-revision', 'snap-revision={}'.format(rev),
+             'snap-id={}'.format(id)],
+        ])
+
+        installed = os.path.join(os.path.sep, 'var', 'lib', 'snapd', 'snaps',
+                                 '{}_{}.snap'.format(name, rev))
```
kyrofa: Have we verified with the snapd team that this path and file-naming convention are stable? We could always ask for the snap path to be contained in the API response.

snapcraft/internal/lxd.py
```diff
@@ -139,6 +144,61 @@ def _setup_project(self):
...
+        filename = os.path.basename(installed)
+        # Copy file to ensure LXD snap can access it
+        shutil.copyfile(installed, filename)
```
kyrofa: Is this copying the snap into cwd?

kalikiana: Yes. I'd prefer to use a temporary folder but currently not sure what I can use that's also readable by the LXD snap.

snapcraft/internal/lxd.py
kalikiana: The folder is apparently guaranteed to exist (unlike /snap)

kyrofa: And the file-naming convention? Not a blocker to land this, but something we should check on.

snapcraft/internal/lxd.py
```diff
@@ -29,7 +29,7 @@
 import yaml
 
 from snapcraft.internal.errors import SnapcraftEnvironmentError
-from snapcraft.internal import common
+from snapcraft.internal.common import format_snap_name, is_snap
```
kyrofa: I prefer leaving this as `common`, personally, unless it's clear from the context where it's coming from (e.g. importing matchers from testtools in a test is clear). Otherwise I feel it decreases readability. @elopio, what are your thoughts?

snapcraft/internal/lxd.py
```diff
@@ -145,60 +145,44 @@ def _setup_project(self):
...
             json = session.request('GET', api).json()
         except Exception as e:
-            raise RuntimeError('Error connecting to {}'.format(api)) from e
+            raise SnapcraftEnvironmentError(
+                'Error connecting to {}'.format(api)) from e
         if not json['status'] == 'OK':
-            raise RuntimeError('Error querying {} snap: {}'.format(name, json))
-        id = json['result']['id']
+            raise SnapcraftEnvironmentError(
+                'Error querying {r!} snap: {}'.format(name, json))
```
kyrofa: Will this print raw JSON, then? Is there a way to format the error nicely?

snapcraft/internal/lxd.py
```diff
@@ -145,60 +145,44 @@ def _setup_project(self):
...
-             'snap-id={}'.format(id)],
-        ])
+        rev = json['result']['revision']
 
         installed = os.path.join(os.path.sep, 'var', 'lib', 'snapd', 'snaps',
                                  '{}_{}.snap'.format(name, rev))
         filename = os.path.basename(installed)
         # Copy file to ensure LXD snap can access it
-        shutil.copyfile(installed, filename)
+        copyfile(installed, filename)
```
kyrofa: Since this copies it to cwd, should we remove it after we've pushed it into the container?

kyrofa: A quick aside: the `lxc` command from the lxd snap only has the `home` and `network` plugs. If the cwd is outside of $HOME this won't work...

kyrofa: Perhaps we could copy it to `$HOME/snap/lxd` if lxc is coming from the snap.

snapcraft/internal/lxd.py
kalikiana: Note the commit message, I explicitly changed it to be able to use mock.patch.

snapcraft/internal/lxd.py
kalikiana: I found another option: /run/user/*/snap.lxd is readable by the 'lxd' snap.

snapcraft/internal/lxd.py
```diff
@@ -165,18 +165,25 @@ def _inject_snap(self, name):
...
-                                 '{}_{}.snap'.format(name, rev))
-        filename = os.path.basename(installed)
-        # Copy file to ensure LXD snap can access it
-        copyfile(installed, filename)
-        self._push_file(filename, os.path.join(self._project_folder, filename))
+                                 filename)
+        # Copy file to a path the 'lxd' snap can access
+        filepath = os.path.join(os.path.sep, 'run', 'user', str(os.getuid()),
+                                'snap.lxd', filename)
+        copyfile(installed, filepath)
```
kyrofa: What if `dirname(filepath)` doesn't exist (e.g. lxd is not running from the snap)? At least from the [docs](https://docs.python.org/3/library/shutil.html#shutil.copyfile) it seems copyfile will not create it.

snapcraft/internal/lxd.py
```diff
@@ -165,18 +165,25 @@ def _inject_snap(self, name):
         except Exception as e:
             raise SnapcraftEnvironmentError(
                 'Error connecting to {}'.format(api)) from e
-        if not json['status'] == 'OK':
+        if json['status'] == 'error':
             raise SnapcraftEnvironmentError(
-                'Error querying {r!} snap: {}'.format(name, json))
+                'Error querying {!r} snap: {}'.format(
+                    name, json['result']['message']))
```
kyrofa: Much better, thank you!

snapcraft/internal/lxd.py
elopio: @kalikiana this is how you can patch the common.is_snap imported in lxd:
http://paste.ubuntu.com/24961616/

elopio: No, scratch that. I see now what you mean. I am more or less ok with your approach.

However, I always avoid importing symbols, so I never patch like this. It is a little confusing that the behaviour of is_snap will be patched in the test only on the call from the lxd module. The fact that you need two different behaviours for that method probably means that you are not testing a unit of code. I am not strongly against this, mainly because the lack of testability is not your fault, and adding this test helps a little to improve things later. But it's a bad smell that we have to change the import to use a test double.

snapcraft/internal/lxd.py
sergiusens: yes

snapcraft/internal/lxd.py
sergiusens: we need to inject the snap so it tracks the correct channel though

snapcraft/tests/commands/test_snap.py
```diff
@@ -139,25 +143,96 @@ def test_snap_containerized(self):
...
+                    'urllib.request.urlopen(' +
+                    '"http://start.ubuntu.com/connectivity-check.html"' +
+                    ', timeout=5)']),
+              call(['apt-get', 'update']),
+              call(['snapcraft', 'snap', '--output',
+                    'snap-test_1.0_amd64.snap']),
+        ])
+
+    @mock.patch('snapcraft.internal.lxd.Containerbuild._container_run')
+    @mock.patch('snapcraft.internal.lxd.is_snap')
```
sergiusens: this could have been `snapcraft.internal.common.is_snap` and would of worked equally fine (wrt to comments above)

sergiusens: All these `is_snap` scenarios could be in their own test class in this module and have the value set through scenarios.

snapcraft/tests/fixture_setup.py
```diff
@@ -443,6 +444,47 @@ def _setUp(self):
...
+            if args[0] == 'GET' and '/v2/snaps/' in args[1]:
+                if self._socket_error:
+                    raise ConnectionError(
+                        'Connection aborted.',
+                        FileNotFoundError(2, 'No such file or directory'))
+
+                class Session:
+                    def __init__(self, name, api_error):
+                        self._name = name
+                        self._confinement = 'core' in name or 'classic'
```
sergiusens: can you enlighten me what this logic implies or what `self._confinement` drives?

snapcraft/internal/lxd.py
```diff
@@ -139,6 +144,54 @@ def _setup_project(self):
...
+        rev = json['result']['revision']
+
+        # https://github.com/snapcore/snapd/blob/master/snap/info.go
+        # MountFile
+        filename = '{}_{}.snap'.format(name, rev)
+        # https://github.com/snapcore/snapd/blob/master/dirs/dirs.go
+        # CoreLibExecDir
+        installed = os.path.join(os.path.sep, 'var', 'lib', 'snapd', 'snaps',
+                                 filename)
+        # Copy file to a path the 'lxd' snap can access
```
sergiusens: what if the deb is being used?

sergiusens: Have you tried straight out copying it to /tmp?

snapcraft/internal/lxd.py
```diff
@@ -139,6 +144,54 @@ def _setup_project(self):
...
+        rundir = os.path.join(os.path.sep, 'run', 'user', str(os.getuid()),
+                              'snap.lxd')
+        os.makedirs(rundir, exist_ok=True)
+        filepath = os.path.join(rundir, filename)
+        copyfile(installed, filepath)
+        self._push_file(filepath, os.path.join(self._project_folder, filename))
+        logger.info('Installing {}'.format(filename))
+        cmd = ['snap', 'install', '--dangerous', filename]
```
sergiusens: why `--dangerous` and not `snap ack` the assertions?

snapcraft/internal/lxd.py
```diff
@@ -139,6 +144,54 @@ def _setup_project(self):
...
+            json = session.request('GET', api).json()
+        except Exception as e:
+            raise SnapcraftEnvironmentError(
+                'Error connecting to {}'.format(api)) from e
+        if json['type'] == 'error':
+            raise SnapcraftEnvironmentError(
+                'Error querying {!r} snap: {}'.format(
+                    name, json['result']['message']))
+        classic = json['result']['confinement'] == 'classic'
+        rev = json['result']['revision']
```
sergiusens: Also make it track the same `channel` as on the system (maybe not as it won't be easier to switch channels from local installs in the same transaction as it is when installing from the store).

snapcraft/internal/lxd.py
sergiusens: Also, only go forward with the install if the revisions differ.

snapcraft/internal/lxd.py
```diff
@@ -111,7 +116,7 @@ def execute(self, step='snap', args=None):
             self._setup_project()
             self._wait_for_network()
             self._container_run(['apt-get', 'update'])
```
sergiusens: side comment, we need a better plan for this as it is really too slow, maybe piggy back it to `snapcraft update` and if it is a containerized instance we do `update` for the parts cache, apt cache and a snap refresh.

snapcraft/internal/lxd.py
kalikiana: - If snapcraft is a dep, this code won't be used at all.
- /tmp is no good because the 'lxd' snap can't read it.

snapcraft/internal/lxd.py
kalikiana: I implemented pushing the assertions initially. Then [we discussed that it's impossible to disable automatic refreshes in snapd](https://forum.snapcraft.io/t/getting-the-revision-of-a-snap-from-snapd/989/9), which could break a build right in the middle. Not using assertions means no updates happen, and snapcraft is responsible for updating.

snapcraft/tests/commands/test_snap.py
kalikiana: - It's not equally fine because the behavior differs - I'm looking into another approach, though, fixing the problem in common that prevented me from global patching.
- Not sure where scenarios would help here? The assertions are completely different.

snapcraft/internal/lxd.py
sergiusens: I would still prefer the other way around or you lose revision check to avoid reinstalling the snap on every run.

snapcraft/internal/lxd.py
sergiusens: not if snapcraft is a deb but if `lxd` is a deb

sergiusens: I guess you will just be creating that path

kalikiana: Yes

snapcraft/__init__.py
```diff
@@ -446,3 +446,8 @@ def str_presenter(dumper, data):
 yaml.add_representer(str, str_presenter)
 yaml.add_representer(OrderedDict, dict_representer)
 yaml.add_constructor(_mapping_tag, dict_constructor)
+
+if common.is_snap():
```
sergiusens: uh oh, this wasn't the best idea...

kalikiana: ?

sergiusens: `common.internal` holds all the good bits and we don't want to expose `is_snap` as a public API.

sergiusens: this is why the tests fail. So either we drive the logic to set in some other internal module or we
```
from snapcraft.internal.common import is_snap as _is_snap
```
and use it like that.

kalikiana: Ah! I couldn't remember why I didn't use "common" there. I was confused then, dug into the code to find common wasn't the common I'm used to, changed the code, and when you asked I couldn't see it from the code anymore. At least your ugly import is more obvious, haha.

snapcraft/internal/lxd.py
```diff
@@ -139,6 +145,77 @@ def _setup_project(self):
...
+        # CoreLibExecDir
+        installed = os.path.join(os.path.sep, 'var', 'lib', 'snapd', 'snaps',
+                                 filename)
+
+        filepath = os.path.join(rundir, filename)
+        shutil.copyfile(installed, filepath)
+        self._push_file(filepath, os.path.join(self._project_folder, filename))
+        logger.info('Installing {}'.format(filename))
+        cmd = ['snap', 'install', filename]
+        if classic:
```
sergiusens: we probably need a `is_dangerous` to cover for the case where we built our own snapcraft snap

snapcraft/internal/lxd.py
```diff
@@ -139,6 +145,77 @@ def _setup_project(self):
...
+        except requests.exceptions.ConnectionError as e:
+            raise SnapcraftEnvironmentError(
+                'Error connecting to {}'.format(api)) from e
+        if json['type'] == 'error':
+            raise SnapcraftEnvironmentError(
+                'Error querying {!r} snap: {}'.format(
+                    name, json['result']['message']))
+        id = json['result']['id']
+        # Lookup confinement to know if we need to --classic when installing
+        classic = json['result']['confinement'] == 'classic'
```
sergiusens: minor nitpick, but can we call this `is_classic` ?

sergiusens: Can we also retrieve the `dangerous` aspect of this?

snapcraft/internal/lxd.py
kalikiana: Made a separate [PR]( https://github.com/snapcore/snapcraft/pull/1412) for that

snapcraft/tests/commands/test_snap.py
```diff
@@ -305,6 +308,10 @@ def test_snap_containerized_inject_snap_dangerous(self,
             call(['lxc', 'file', 'push',
                   os.path.join(rundir, 'core_123.snap'),
                   '{}/run/core_123.snap'.format(container_name)]),
+            call(['sudo', 'cp', '/var/lib/snapd/snaps/snapcraft_x1.snap',
```
sergiusens: Maybe a logger entry with the command to call `sudo` with here would be good (just so it is no surprise to the random user that decided to install our snap dangerously :wink:)

snapcraft/internal/lxd.py
```diff
@@ -204,10 +204,11 @@ def _inject_snap(self, name):
 
         filepath = os.path.join(rundir, filename)
         shutil.copyfile(installed, filepath)
-        container_filename = os.path.join('run', filename)
+        # os.sep needs to be `/` and on Windows it will be set to `\`
```
sergiusens: I don't think snaps on native windows will be a thing here :smile:

PR #1365, opened 2017-06-15
Add Ruby Plugin

jamesbeedy:
* Added snacraft/plugins/ruby.py
* Installs specified ruby
* Automatically installs 'bundler'
* Installs gems from gemfile via bundler if gemfile exists and or gems from plugin 'gems' option

TODO
* Integration tests


snapcraft/plugins/ruby.py
```diff
@@ -0,0 +1,121 @@
...
+
+class RubyPlugin(BasePlugin):
+
+    @classmethod
+    def schema(cls):
+        schema = super().schema()
+
+        schema['properties']['ruby-version'] = {
+            'type': 'string',
+            'default': "2.3.1"
```
kyrofa: Please use single-quotes throughout (unless the string _includes_ single quotes, in which case double is fine). Just for codebase consistency.

snapcraft/plugins/ruby.py
```diff
@@ -0,0 +1,121 @@
...
+                'type': 'string'
+            },
+            'default': []
+        }
+        return schema
+
+    @classmethod
+    def get_pull_properties(cls):
+        # Inform Snapcraft of the properties associated with pulling. If these
+        # change in the YAML Snapcraft will consider the build step dirty.
```
kyrofa: s/build/pull/. `gems` should also probably be in here.

snapcraft/plugins/ruby.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def __init__(self, name, options, project):
+        super().__init__(name, options, project)
+
+        self._ruby_version = self.options.ruby_version
+        self._rubylib = join(self.installdir, 'lib', 'ruby')
+        self._ruby_dir = join(self.partdir, 'ruby')
+        self._gem_path = \
+            join(self._rubylib, 'gems', self._ruby_version, 'gems')
+        self._gem_home = self._gem_path
```
kyrofa: Given this on my dev machine (with rvm):

```
$ echo $GEM_HOME
/home/kyrofa/.rvm/gems/ruby-2.4.0
$ echo $GEM_PATH
/home/kyrofa/.rvm/gems/ruby-2.4.0:/home/kyrofa/.rvm/gems/ruby-2.4.0@global
```

I think neither of these should include the `gems` dir.

snapcraft/plugins/ruby.py
```diff
@@ -0,0 +1,121 @@
...
+        self._ruby_download_url = \
+            ('https://cache.ruby-lang.org/pub/ruby/ruby-%s.tar.gz' % \
+             self._ruby_version)
```
kyrofa: Please use `format` instead of `%`.

snapcraft/plugins/ruby.py
```diff
@@ -0,0 +1,121 @@
...
+        self._ruby_tar = Tar(self._ruby_download_url, self._ruby_dir)
+        self._gems = self.options.gems
+        self._install_bundler = False
+
+        pkgs = ['gcc', 'g++', 'make', 'zlib1g-dev',
+                'libssl-dev', 'libreadline-dev']
+        for pkg in pkgs:
+            self.build_packages.append(pkg)
```
kyrofa: I bet `self.build_packages.extend(['gcc', 'g++', 'make', 'zlib1g-dev', 'libssl-dev', 'libreadline-dev'])` will bring you closer to enlightenment and pure joy.

snapcraft/plugins/ruby.py
```diff
@@ -0,0 +1,121 @@
...
+        if self._install_bundler:
+            self._bundle_install()
+
+    def _env(self):
+        """Ruby env vars."""
+        env = environ.copy()
+        env['RUBYLIB'] = self._rubylib
+        env['GEM_PATH'] = self._gem_path
+        env['GEM_HOME'] = self._gem_home
+        return env
```
kyrofa: I suggest using the `env` function from the [plugin API](https://github.com/snapcore/snapcraft/blob/master/snapcraft/_baseplugin.py#L155). That would mean you no longer have to use `env=self._env()`, and it will give you this environment at runtime as well (as long as you use `root` instead of hard-coding the installdir). It would look something like this (untested):

```python
def env(self, root):
    rubylib = os.path.join(root, 'lib', 'ruby')
    gem_home = os.path.join(rubylib, 'gems', self._ruby_version)
    
    return [
        'RUBYLIB={}'.format(rubylib),
        'GEM_HOME={}'.format(gem_home),
        'GEM_PATH={}'.format(gem_home)
    ]
```

snapcraft/plugins/ruby.py
```diff
@@ -0,0 +1,121 @@
...
+        self.run(['./configure', '--disable-install-rdoc', '--prefix=/'], cwd=builddir)
+        self.run(['make', '-j{}'.format(self.parallel_build_count)], cwd=builddir)
+        self.run(['make', 'install', 'DESTDIR={}'.format(self.installdir)], cwd=builddir)
+
+    def _gem_install(self):
+        if exists(join(self.builddir, 'Gemfile')):
+            self._install_bundler = True
+            self._gems = self._gems + ['bundler']
+        for gem in self._gems:
+            self.run(['gem', 'install'] + [gem], env=self._env())
```
kyrofa: Why not just `self.run(['gem', 'install', gem])`?

jamesbeedy: I was trying to follow the convention/style from the nodejs plugin. I feel like you also hinted (possibly just for consistent style/readability) to add lists to lists, instead of strings to lists. I'll remove the brackets here since they are not needed.

snapcraft/plugins/ruby.py
jamesbeedy: 👍

snapcraft/plugins/ruby.py
jamesbeedy: 👍

snapcraft/plugins/ruby.py
jamesbeedy: 👍

snapcraft/plugins/ruby.py
jamesbeedy: 👍

snapcraft/plugins/ruby.py
jamesbeedy: 👍

snapcraft/plugins/ruby.py
jamesbeedy: 👍

jamesbeedy:
@kyrofa would you mind giving this a second look over for me? I'm having a hard time getting any insight as to why the latest test is failing.

I know I need to replace the hard coding of the arch with something dynamic in the RUBYLIB path.. Not sure if thats what is causing this to fail, I just know I need to fix it ... looking for the best equivalent to `uname -p` now to use in the plugin code now.

jamesbeedy:
@kyrofa  

I know why its failing (I think).

From what I can tell, the `gem` executable located at `parts/ruby/install/bin/gem`, which is ran in the pull step can't find ruby. The reason I hadn't caught this before is because I was testing this with ruby 2.3.1 which I had installed in my host, when the gem command was ran in the pull step it just used this ruby and it by chance matched version. As soon as I removed os ruby, or changed the ruby-version option to something other then '2.3.1' I started getting hit with this ```-bash: ./gem: //bin/ruby: bad interpreter: No such file or directory
```

**Comment by jamesbeedy** [comment]
Ahh I finally figured it out, it was so silly .... I needed to specify the path to the ruby executable preceding the call to `gem`.

**Comment by jamesbeedy** [comment]
@kyrofa it looks as though the build is passing when I inspect it. Could you help me gather some insight as to why this build is failing?
thx

**Comment by kyrofa** [comment]
Yeah, it looks like you still need to update the list of plugins in the test cases:

```

FAIL: test_list_plugins_large_terminal (snapcraft.tests.commands.test_list_plugins.ListPluginsCommandTestCase)

snapcraft.tests.commands.test_list_plugins.ListPluginsCommandTestCase.test_list_plugins_large_terminal (list-plugins)

testtools.testresult.real._StringException: Empty attachments:

  pythonlogging:''

  File "/home/travis/build/snapcore/snapcraft/snapcraft/tests/commands/test_list_plugins.py", line 57, in test_list_plugins_large_terminal

    self.assertThat(result.output, Contains(self.default_plugin_output))

  File "/usr/local/lib/python3.5/dist-packages/testtools/testcase.py", line 498, in assertThat

    raise mismatch_error

testtools.matchers._impl.MismatchError: 'ant        cmake  go      gulp    kernel  meson   plainbox-provider  python3  scons      \nautotools  copy   godeps  jdk     make    nil     python             qmake    tar-content\ncatkin     dump   gradle  kbuild  maven   nodejs  python2            rust     waf        \n' not in 'ant        cmake  go      gulp    kernel  meson   plainbox-provider  python3  rust         waf\nautotools  copy   godeps  jdk     make    nil     python             qmake    scons      \ncatkin     dump   gradle  kbuild  maven   nodejs  python2            ruby     tar-content\n'
```

Run `./runtests.sh unit` to see these failures locally.

PR #1366, opened 2017-06-15
cli: Add --version command

sparkiegeek:
LP: [#1692752](https://bugs.launchpad.net/snapcraft/+bug/1692752)

Adds a --version flag to the CLI.


snapcraft/tests/commands/test_version.py
```diff
@@ -0,0 +1,25 @@
...
+from testtools.matchers import Equals
+
+from . import CommandBaseTestCase
+
+
+class VersionCommandTestCase(CommandBaseTestCase):
+
+    def test_has_version(self):
+        result = self.run_command(['--version'])
+        self.assertThat(result.exit_code, Equals(0))
```
sparkiegeek: Testing the output itself seems fragile - either we'd have to keep updating the version number as releases are made, or repeat the same poor-person's parsing of `debian/changelog` we see in `setup.py`.

Suggestions welcome though!

snapcraft/__init__.py
```diff
@@ -377,6 +377,22 @@ def find_library(name):
     import ctypes.util
     ctypes.util.find_library = find_library
 
+
+def _get_version():
+    if _os.environ.get('SNAP_NAME') == 'snapcraft':
+        with open(_os.path.join(_os.getenv('SNAP'), 'meta', 'snap.yaml')) as f:
+            snap_yaml = yaml.load(f)
+        return snap_yaml['version']
```
kyrofa: I know you didn't write this originally, but is there a reason we're not simply using `SNAP_VERSION` here as opposed to the yaml dance?

snapcraft/tests/commands/test_version.py
kyrofa: Well, and I'm pretty sure if you're running from source that it'll return `--devel` every time anyway, right?

sparkiegeek: hmm? for me it returns "run, version 2.31\n"

I think the run comes from `python3 -m coverage run` ... and 2.31 comes from the `debian/changelog`

I could test that it contains __version__ but seems like a test that will never fail

snapcraft/__init__.py
sparkiegeek: Good question, I'll switch it to using `SNAP_VERSION`

snapcraft/__init__.py
```diff
@@ -377,6 +377,20 @@ def find_library(name):
     import ctypes.util
     ctypes.util.find_library = find_library
 
+
+def _get_version():
+    if _os.environ.get('SNAP_NAME') == 'snapcraft':
+        return _os.environ["SNAP_VERSION"]
```
elopio: A nit, but please use single quotes on strings for consistency.

snapcraft/__init__.py
```diff
@@ -377,6 +377,20 @@ def find_library(name):
...
+    try:
+        return pkg_resources.require('snapcraft')[0].version
+    except pkg_resources.DistributionNotFound:
+        return 'devel'
+
+
+# Set this early so that the circular imports aren't too painful
+__version__ = _get_version()
```
elopio: Another small detail. Could you please this variable right before the _get_version definition?
I'm trying to get our source code to be readable from top to bottom, and those details help.

snapcraft/tests/commands/test_version.py
```diff
@@ -0,0 +1,25 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2015-2017 Canonical Ltd
```
elopio: This should be only 2017.

snapcraft/__init__.py
sparkiegeek: Sure.

snapcraft/tests/commands/test_version.py
sparkiegeek: Fixed.

snapcraft/__init__.py
sparkiegeek: You missed out a word, but I assume you're asking if I could move the definition of `__version__` above `_get_version`?

I cannot - this is a module level variable, it is evaluated at the time of importing the module, as such it cannot reference a function which hasn't yet been defined.

`_get_version` has to come before assignment to `__version__`

snapcraft/tests/commands/test_version.py
elopio: I would add a little regexp in the check.
Something like self.assertThat(result.output, MatchesRegex('snapcraft, version .*'))

snapcraft/__init__.py
elopio: This is a cool trick, to get it from the env var. I don't see any potential issues, and it simplifies everything, even the tests. +1 to this approach.

snapcraft/__init__.py
elopio: you are right.

elopio:
I saw a comment about circular dependencies when importing snapcraft. Would it make sense to move the version stuff to snapcraft/version.py?

sparkiegeek:
> I saw a comment about circular dependencies when importing snapcraft. Would it make sense to move the version stuff to snapcraft/version.py?

It's tricky because there are multiple things happening at module-import time:

1. `snapcraft.__version__` is being calculated
2. A bunch of `snapcraft.*` modules are imported, in turn importing `snapcraft.cli`
3. The `click` decorators (specifically those on `snapcraft.cli.run`) are being evaluated

We want both 1. and 2. to use the same value, defined in one place. This is why I moved the calculation of `__version__` to before the imports - the prototypical snapcraft namespace is then sufficient to pass 3. and allows Python to do it's normal circular dependency resolution.

Extracting the version bits to `snapcraft.version` would have `snapcraft/__init__.py` and `snapcraft/cli/__init__.py` both importing it, since it's under the snapcraft namespace though, it will trigger the evaluation of `snapcraft/__init__.py` so the circle remains.

snapcraft/tests/commands/test_version.py
sparkiegeek: As per previous comment, that would fail - the guessed name when running tests is 'run', not 'snapcraft'. I could hardcode 'snapcraft' as the name, but then I'd basically be testing the behaviour of `click`, which seems pointless (beyond the point of diminishing returns)

elopio: Well, taking into account that this is the first thing I run when I start testing before making the stable release, it's actually not a big deal to not check the output.

snapcraft/tests/test_init.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class VersionTestCase(tests.TestCase):
+
+    def test_version_from_snap(self):
+        self.useFixture(fixtures.EnvironmentVariable(
+            'SNAP_NAME', 'snapcraft'))
+        self.useFixture(fixtures.EnvironmentVariable(
+            'SNAP_VERSION', '3.14'))
+        self.assertThat(snapcraft._get_version(), Equals('3.14'))
```
sergiusens: a missed opportunity to use `42` here :stuck_out_tongue: 

thanks for the contribution though!

PR #1368, opened 2017-06-19
cli: get_project_options must not discard kwargs

kalikiana:
The function get_project_options aliases certain options. But the remainder of kwargs is ignored so debug=debug coming from snapcraft/cli/lifecycle.py would have no effect.

### DIFF
--- a/snapcraft/cli/_options.py
+++ b/snapcraft/cli/_options.py
@@ -65,4 +65,4 @@ def get_project_options(**kwargs):
         target_deb_arch=kwargs.pop('target_arch'),
     )
 
-    return ProjectOptions(**project_args)
+    return ProjectOptions(**project_args, **kwargs)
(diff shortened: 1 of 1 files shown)


snapcraft/cli/_options.py
```diff
@@ -65,4 +65,4 @@ def get_project_options(**kwargs):
         target_deb_arch=kwargs.pop('target_arch'),
     )
 
-    return ProjectOptions(**project_args)
+    return ProjectOptions(**project_args, **kwargs)
```
elopio: @kalikiana but in here you would be duplicating some arguments, with different names. 
There will be use_geoip and enable_geoip, parallel_builds and no_parallel_builds, target_deb_arch and target_arch.
This is terrible, it would be better if the arguments had the same names. But well, now that we are in this mess, wouldn't it be better to pop the debug argument from kwargs, and get it into project_args?

kalikiana: That's what the code did. But since it wasn't clear enough, I removed the second dict.

kalikiana:
I'll hold off on this one for a bit since I really want it to be tested but I'll have conflicts due to another bug I'm fixing.

PR #1369, opened 2017-06-19
Handle I/O errors in os.link (LP: #1689956).

evandandrea:
Docker for Mac raises an I/O error when trying to hard link:
>[Errno 5] Input/output error: '/Users/ev/Branches/redis/parts/redis/install/bin/redis-sentinel' -> '/Users/ev/Branches/redis/stage/bin/redis-sentinel'

This leaves a file behind, which snapcraft then tries to copy over top of.

LP: [#1689956](https://bugs.launchpad.net/snapcraft/+bug/1689956)


elopio:
You found a workaround! Great, thanks.
The problem is that this is causing tests to fail. And, have you thought about a unit test that lets us reproduce the issue you are getting? Or will it happen only in mac?

evandandrea:
I still need to write a unit test, but I've fixed the test failures (bar the one about me not signing the CLA).

evandandrea:
I've added my email address to my LP account, so if you re-run the tests they should all pass now.

evandandrea:
The CLA check is failing because I haven't publicly disclosed my email addresses on Launchpad (and have no desire to do so). Can you hammer this through?

evandandrea:
Sorry about the delay. Should be good now:

```
☻  cla-check <redacted>
Logging into Launchpad...
Contributor: https://api.launchpad.net/1.0/~ev
The contributor has signed the CLA.
```

elopio:
@evandandrea only Sergio has permission to ignore a failure in the tests and merge. I've added this pr to our plan for 2.33, so it will go out in the next release for sure.

snapcraft/file_utils.py
```diff
@@ -110,6 +110,12 @@ def link_or_copy(source, destination, follow_symlinks=False):
         # symlinks.
         os.link(source_path, destination, follow_symlinks=False)
     except OSError:
+        try:
+            # If os.link raised an I/O error, it may have left a file behind.
+            if not os.path.isdir(destination):
+                os.unlink(destination)
+        except FileNotFoundError:
```
kyrofa: Two points here:

1. I prefer contextlib for this, @elopio do you prefer try/except?
2. You're already `pass`ing one expected exception, why not stop checking to see if it's a dir and pass on `OSError` as well?

elopio: I have no strong opinion about contextlib. I never use it, but it's not because I don't like it, it's because I never remember it exists.

sergiusens: `suppress` would make this more readable too

```python
# If os.link raised an I/O error, it may have left a file behind.
@contextlib.suppress(FileNotFoundError)
os.unlink(destination)
```

snapcraft/file_utils.py
```diff
@@ -110,12 +110,11 @@ def link_or_copy(source, destination, follow_symlinks=False):
         # symlinks.
         os.link(source_path, destination, follow_symlinks=False)
     except OSError:
-        try:
-            # If os.link raised an I/O error, it may have left a file behind.
-            if not os.path.isdir(destination):
-                os.unlink(destination)
-        except FileNotFoundError:
```
kalikiana: Don't we still need to handle the FileNotFoundError? Should both be suppressed?

snapcraft/file_utils.py
```diff
@@ -110,6 +110,11 @@ def link_or_copy(source, destination, follow_symlinks=False):
         # symlinks.
         os.link(source_path, destination, follow_symlinks=False)
     except OSError:
+        # If os.link raised an I/O error, it may have left a file behind.
+        # Skip on OSError in case it doesn't exist or is a directory.
+        with suppress(OSError):
```
kyrofa: You only get an OSError if it's a directory. You still get FileNotFoundError if it  doesn't exist. You likely want `suppress(FileNotFoundError, OSError)` here.

evandandrea: FileNotFound is under OSError in the class hierarchy.

PR #1370, opened 2017-06-19
integrations: add the snapcraft Dockerfile

elopio:
LP: #[1699218](https://bugs.launchpad.net/snapcraft/+bug/1699218)

The source for our snapcore/snapcraft docker image is in a location that probably only @sergiusens knows. It should be in this repo, so we can all find it and touch it.

In particular, the `ENV` entry added in this file is required by snapcraft 2.31, which has just landed into the archive. So we need to hurry and update the image using this as the source.

I've asked @niemeyer to give me access to update the image: https://forum.snapcraft.io/t/in-progress-snapcraft-2-31/682/9


elopio:
This was base on the file made by @didrocks: https://github.com/ubuntu/docker-snapcraft

chihchun:
In case anyone like to use snapcraft and ubuntu-image in docker, here is document and link
* A Docker image for build Ubuntu snaps - https://github.com/chihchun/snapcraft-docker
* snapcraft - Docker Hub - https://hub.docker.com/u/snapcraft/

PR #1372, opened 2017-06-21
cli: Containerbuild clean

kalikiana:
Add a unit test (and fix) for `SNAPCRAFT_CONTAINER_BUILDS=1 clean`.

### DIFF
--- a/snapcraft/internal/lxd.py
+++ b/snapcraft/internal/lxd.py
@@ -219,7 +219,7 @@ class Project(Containerbuild):
 
     def execute(self, step='snap', args=None):
         super().execute(step, args)
-        if step == 'clean' and not args:
+        if step == 'clean' and 'pull' in args:
             print('Deleting {}'.format(self._container_name))
             check_call(['lxc', 'delete', '-f', self._container_name])
 
--- a/snapcraft/internal/lxd.py
+++ b/snapcraft/internal/lxd.py
@@ -219,7 +219,8 @@ class Project(Containerbuild):
 
     def execute(self, step='snap', args=None):
         super().execute(step, args)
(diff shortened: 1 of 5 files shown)


elopio:
This looks correct to me. However, I'm having problems to test it.

I'm failing to build the snap with this error: http://paste.ubuntu.com/24930727/

Then the container is stopped. When I run clean, the first thing it tries to do is to start it, and it fails.

@kalikiana do you see something wrong in my first snapcraft call? And in the second, wouldn't it be better not even trying to start the container? just remove it?

snapcraft/internal/lxd.py
```diff
@@ -219,7 +219,7 @@ def _finish(self):
 
     def execute(self, step='snap', args=None):
         super().execute(step, args)
-        if step == 'clean' and not args:
+        if step == 'clean' and 'pull' in args:
```
elopio: Don't we have to check here if there are no parts in `args`? Maybe, this clean command is to clean only one part, and leave the rest.

kalikiana:
@elopio I have no idea what your error is about. I suggest you open a forum topic for investigation. This branch isn't changing the implementation, only the passing of arguments.
I don't think we can "just" remove the container because then we wouldn't be cleaning anything in the project folder. But, this would also be a nice forum topic. We can maybe find ways to optimize by cleaning outside of the container (and investigate potential problems with that).

elopio:
@kalikiana there's clearly a bug, not introduced by this branch but it prevents me from testing this branch. I reported it here, with simple steps to reproduce: https://bugs.launchpad.net/snapcraft/+bug/1700703

It happens on all my machines, but I made sure to reproduce it in a clean environment. I assigned it to you. Please take a look.

PR #1374, opened 2017-06-21
tests: set up the spread execution in linode

elopio:


elopio:
@fgimenez can you please review this?

spread.yaml
```diff
@@ -8,15 +8,28 @@ environment:
 backends:
   lxd:
     systems: [ubuntu-16.04]
+  linode:
+    key: "$(HOST: echo $SPREAD_LINODE_KEY)"
+    systems:
+      - ubuntu-16.04-64:
```
fgimenez: The default linode kernels don't play well with snapd because of the lack of apparmor support, not sure if this apply here but, because you install snapd later, maybe you should use a specific kernel like we do in snapd's suite https://github.com/snapcore/snapd/blob/master/spread.yaml#L51

spread.yaml
```diff
@@ -8,15 +8,28 @@ environment:
...
+          workers: 3
 
 suites:
  spread_tests/:
    summary: tests of the snapcraft snap
    prepare: |
      apt update
```
fgimenez: We have been suffering lots of problems regarding apt and connection issues from linode hosts to ubuntu mirrors over ipv6 (which is enabled by default), if you hit those, in the form of tests timing out, you could either disable ipv6 or give it a low priority (the latter is giving very good results to us so far https://github.com/snapcore/snapd/blob/master/spread.yaml#L294)

elopio:
Thanks @fgimenez. I've copied your bits.

spread_tests/integration/task.yaml
```diff
@@ -5,5 +5,6 @@ execute: |
   echo "Check that we are using the snapcraft snap"
   snapcraft_path="$(which snapcraft)"
   [ "$snapcraft_path" = "/snap/bin/snapcraft" ]
+  source /snapcraft/venv/bin/activate
```
kyrofa: Is this necessary given that it's already in the suite's prepare step?

elopio: Yes, apparently it happens in a different environment or something. I added it just because without it, it failed.

kyrofa: Huh, okay. Does it need to be in the suite's prepare step, then?

spread.yaml
```diff
@@ -8,15 +8,43 @@ environment:
...
+     apt install --yes squashfuse
+     mkdir -p /snapcraft/venv
+     python3 -m venv /snapcraft/venv
+     # XXX the activate script has an unbound variable.
+     sed -i '1s/^/set +u\n/' /snapcraft/venv/bin/activate
+     source /snapcraft/venv/bin/activate
+     pip install --upgrade pip
+     pip install -r /snapcraft/requirements.txt -r /snapcraft/requirements-devel.txt
+     apt install --yes snapd
      snap install snapcraft --classic --edge
```
kyrofa: Wait... I don't understand what's happening here. You're installing via pip, but also installing the snap. They'll both be in the path. Which one is being tested?

elopio: I'm installing only the requirements. They are all needed to run the integration tests. I didn't add the `.`, and take a look here where we chec that we are using the right command:
https://github.com/snapcore/snapcraft/pull/1374/files#diff-408b0efbb4ace174e50d6bd885ae24baL7

kyrofa: Ahh, stupid periods. So easy to miss. Thanks for the info!

elopio: yay!!!

spread_tests/integration/task.yaml
elopio: Oh, sorry, I merged before reading this question.
Yes, how would I install the requirements in the venv without activating it?

PR #1375, opened 2017-06-22
tests: allow to filter tests in docker

elopio:
In runtests.sh we allow to filter the the name of the test file that will be run. It will be useful to pass that parameter also when running the tests in the docker containers that travis uses.

### DIFF
--- a/tools/travis/run_tests.sh
+++ b/tools/travis/run_tests.sh
@@ -27,6 +27,10 @@ fi
 
 test="$1"
 
+if [ "$#" -gt 1 ]; then
+    pattern="$2"
+fi
+
 if [ "$test" = "static" ]; then
     dependencies="apt install -y python3-pip && python3 -m pip install -r requirements-devel.txt"
 elif [ "$test" = "unit" ]; then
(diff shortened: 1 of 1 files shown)


elopio:
Description added.
About documenting the tests in HACIKING.md, I have a lot to write to get to this point. It's a long lasting technical debt, so it would be great if I can put it in the deliverables of 2.33.

tools/travis/run_tests.sh
```diff
@@ -27,6 +27,10 @@ fi
 
 test="$1"
 
+if [ "$#" -gt 1 ]; then
```
kyrofa: Is there a use-case for having more than two args? Or should this be `-eq 2`?

kyrofa: Why is this protected by an `if` anyway? Passing it to docker exec if it's empty won't do anything anyway... right?

elopio: It can error if greater, or just ignore any additional args. I preferred just to ignore, but I can make it more strict. And I put it in an `if` just because I'm too used to real languages. It's hard for me to keep in mind that `pattern="$2"` would work even when there's no $2. I will make the changes on monday.

kyrofa: Ignoring seems fine, but yeah, I'd just get rid of the `if`.

kalikiana: On a second look, I think we should have

```
if [ "$#" -gt 2]; then
  echo "Usage: "$0" <test> [PATTERN]"
  exit 1
fi
```

And update the message in case of no arguments as well.

elopio:
@kyrofa I removed the if.
@kalikiana I extend the usage error.

This is ready for a new review.

PR #1378, opened 2017-06-23
Integration tests for snap command with target arch

kalikiana:
Raised [in the forum](https://forum.snapcraft.io/t/kernel-cross-compile-problem-missing-snapcraft-target-arch-options/1083/5) the `--target-arch` option seems to be silently ignored depending on where it's specified, specifically if it's specified before the **snap** command.
Additionally the argument never shows up in `--help`.

We're also lacking integration tests for the different methods of invocation.

### DIFF
--- a/snapcraft/cli/_options.py
+++ b/snapcraft/cli/_options.py
@@ -53,6 +53,11 @@ def add_build_options(hidden=False):
 
 
 def get_project_options(**kwargs):
+    ctx = click.get_current_context()
+    for key, value in ctx.parent.params.items():
+        if not kwargs.get(key):
+            kwargs[key] = value
+
     project_args = dict(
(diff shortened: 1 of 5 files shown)


snapcraft/cli/_options.py
```diff
@@ -52,13 +46,18 @@ def _add_build_options(func):
         for name, params in zip(reversed(_BUILD_OPTION_NAMES),
                                 reversed(_BUILD_OPTIONS)):
             option = click.option(name, **params,
-                                  cls=OptionWithHidden, hidden=hidden)
+                                  cls=HiddenOption if hidden else click.Option)
```
sergiusens: The reason it was implemented as it was before as it would be a transparent change once the feature lands in click. That said, this new implementation is more readable and we would need to change code anyways once migrating.

PR #1379, opened 2017-06-23
blank version should not be allowed in snapcraft.yaml

roxyd:
LP: [#1660076](https://bugs.launchpad.net/snapcraft/+bug/1660076)


snapcraft/tests/test_project_loader.py
```diff
@@ -885,6 +885,30 @@ def test_invalid_environment(self, mock_load_plugin):
...
+        self.make_snapcraft_yaml("""name: test
+version:''
+summary: test
+description: test
+confinement: strict
+grade: stable
+parts:
+  part1:
+    plugin: go
+    stage-packages: [fswebcam]
```
elopio: The idea is to keep the test as small as possible, so you can save one line by using `plugin: nil` and removing `stage-packages`.

snapcraft/tests/test_project_loader.py
```diff
@@ -885,6 +885,30 @@ def test_invalid_environment(self, mock_load_plugin):
...
+""")
+        raised = self.assertRaises(
+            errors.SnapcraftSchemaError,
+            project_loader.Config)
+
+        self.assertEqual(
+            raised.message,
+            "'version' can not be an empty string")
```
elopio: I suspect this will fail because we are showing the regexp in the error.
We want to change that in the future to something nicer, but for now, it's ok just to update the expected value here.

elopio:
@roxyd The static tests are failing in travis.

snapcraft/tests/test_project_loader.py:1:1: E265 block comment should start with '# '
snapcraft/tests/test_project_loader.py:912:1: E302 expected 2 blank lines, found 1

snapcraft/tests/test_project_loader.py
```diff
@@ -885,6 +885,30 @@ def test_invalid_environment(self, mock_load_plugin):
             "The 'environment/INVALID' property does not match the required "
             "schema: \[1, 2\].*")
 
+    @unittest.mock.patch('snapcraft.internal.parts.PartsConfig.load_plugin')
+    def test_invalid_yaml_empty_version(self, mock_loadPlugin):
```
elopio: a small nitpick here: the name of the variable should be `mock_load_plugin`

kalikiana:
Looking very good! Thanks for this fix!

PR #1382, opened 2017-06-27
rust plugin: make libc configurable

kalikiana:
rust supports musl as a libc alternative to glibc and it's especially interesting for static linking.


snapcraft/tests/matchers.py
```diff
@@ -18,16 +18,35 @@
 import testtools
 
 
-class HasArchitecture:
-    """Match if the file was built for the expected architecture"""
-
-    def __init__(self, expected_arch):
-        self._expected_arch = expected_arch
+class HasMagic:
```
elopio: Can you please explain this in a comment in the source code?
I'm totally confused because it's called magic.

Ideally, you could find a better name for this that's not the name of the module, which is pretty bad. Something like HasType? HasMIMEType?

snapcraft/tests/matchers.py
```diff
@@ -18,16 +18,35 @@
...
         self._ms = magic.open(magic.NONE)
         self._ms.load()
 
     def __str__(self):
-        return 'HasArchitecture()'
+        return '{}()'.format(self.__name__)
+
+
+class HasLinkage(HasMagic):
+    """Match if the file was built for the expected architecture"""
```
elopio: This comment has to be updated.

integration_tests/plugins/test_rust_plugin.py
```diff
@@ -101,6 +101,18 @@ def test_cross_compiling(self):
         binary = os.path.join(self.parts_dir, 'rust-hello', 'install', 'bin',
                               'rust-hello')
         self.assertThat(binary, HasArchitecture('aarch64'))
+        self.assertThat(binary, HasLinkage('dynamically linked'))
+
+    def test_cross_compiling_musl(self):
```
elopio: Why no integration test for normal compillation?

integration_tests/snaps/rust-musl/snapcraft.yaml
```diff
@@ -0,0 +1,14 @@
+name: rust-musl
```
elopio: what would you think of using the basic rust yaml, and in the test setup append `libc: musl`, instead of adding a new snap?

snapcraft/tests/matchers.py
```diff
@@ -18,16 +18,39 @@
 import testtools
 
 
-class HasArchitecture:
-    """Match if the file was built for the expected architecture"""
-
-    def __init__(self, expected_arch):
-        self._expected_arch = expected_arch
+class HasBinaryFileHeader:
```
kyrofa: This should probably start with an underscore seeing that it can't be used standalone.

snapcraft/tests/plugins/test_rust.py
```diff
@@ -46,6 +48,7 @@ class Options:
             rust_features = []
             rust_revision = ''
             rust_channel = ''
+            libc = self.libc if hasattr(self, 'libc') else 'gnu'
```
kyrofa: Perhaps `libc = getattr(self, 'libc', 'gnu')` would be better.

snapcraft/tests/matchers.py
kalikiana: I changed it so it can be used on its own instead. Also made the subclasses smaller.

integration_tests/plugins/test_rust_plugin.py
kalikiana: I figured the existing ones are enough... I added the linkage assertion to the existing first test now.

integration_tests/snaps/rust-musl/snapcraft.yaml
kalikiana: I guess now that you've also proposed new fixtures, this can be refactored once this lands.

integration_tests/plugins/test_rust_plugin.py
kalikiana: Done

kalikiana:
Apologies for making a mess... I rebased, which locally got me a clean diff and commits on top of master. Can't see how to fix it for GitHub, though.

integration_tests/plugins/test_rust_plugin.py
```diff
@@ -103,6 +107,24 @@ def test_cross_compiling(self):
         binary = os.path.join(self.parts_dir, 'rust-hello', 'install', 'bin',
                               'rust-hello')
         self.assertThat(binary, HasArchitecture('aarch64'))
+        self.assertThat(binary, HasLinkage('dynamically linked'))
```
elopio: It seems to me that the matchers would be clearer called like IsDynamicallyLinked() and IsStaticallyLinked()

snapcraft/plugins/rust.py
```diff
@@ -31,6 +31,9 @@
     - rust-features
       (list of strings)
       Features used to build optional dependencies
+    - libc:
+      (string; default: gnu)
+      Libc to link against. Valid options are: gnu (glibc) and musl
```
elopio: I don't have much clue about this, but why call it gnu instead of glibc?

elopio:
I have a few comments, but no complaints here.
Ideally, this would be reviewed also by somebody who knows rust and musl.

What do you think of my suggestions for the matchers without args? I'll give you a +1 now, but I think it would be clearer that way.

snapcraft/plugins/rust.py
kalikiana: `gnu` and `musl` are the names of the toolchains. I mention glibc there to clarify what it means.

integration_tests/plugins/test_rust_plugin.py
kalikiana: To my mind `HasLinkage` is consistent with `HasArchitecture` in that it's not hiding the basic string-parsing nature of the matcher... but yeah, I added subclasses as per your suggestion now.

sergiusens:
stale

kalikiana:
Why did you close this? I merged master on Thursday to try and get past the false negatives and it had positive reviews.

kalikiana:
False negative... `TypeError: unsupported operand type(s) for -=: 'Retry' and 'int'`

kyrofa:
@kalikiana we can't merge something with false negatives, because it prevents the following tests from running. I know it's annoying, but keep retrying.

kalikiana:
`test_rust_plugin.RustPluginTestCase.test_cross_compiling_musl ... The job exceeded the maximum time limit for jobs, and has been terminated.`

Closing for now, need to investigate how to make this more reliable on Travis.

PR #1383, opened 2017-06-27
autotools: Enable cross-compilation support

kalikiana:
CC and CXX environment variables need to be set accordingly to cross-compile with Autotools.


elopio:
The integration tests are failing here:
configure: error: cannot run C compiled programs.

If you meant to cross compile, use `--host'.

See `config.log' for more details

autoreconf -i

./configure --prefix=

Command '['/bin/sh', '/tmp/tmpnnz1mqq6', './configure', '--prefix=']' returned non-zero exit status 1

kalikiana:
You've been looking at an outdated clone me thinks, I pushed a fix for that several hours before your comment.

elopio:
I was just looking at the latest travis executions. Now they are green, thanks, I will give this a try.

PR #756, opened 2016-08-24
Use block style for multiline YAML values.

josepht:
Make the parts.yaml, 'search', and 'define' output nicer.

* Use block style for multiline strings
* Support OrderedDict for all yaml usage.
* Only show the first line of the description in 'search' output.

LP:#1614651

PR #1384, opened 2017-06-27
Support yaml merge tags

tim-sueberkrueb:
The current yaml loading implementation in snapcraft doesn't support yaml merge tags.

Example:
```yaml
parts:
  base: &base
    plugin: nil
    source: .
  my-part:
    plugin: dump
    <<: *base
```

Snapcraft fails with the following error message:

```
could not determine a constructor for the tag 'tag:yaml.org,2002:merge'
  in "snap/snapcraft.yaml", line 20, column 5
```

Would you like me to create a test case for this? If yes, where? Thanks in advance.

LP: [#1700765](https://bugs.launchpad.net/snapcraft/+bug/1700765)


kalikiana:
Would this be a bug in snapcraft? Or in pyyaml? I'd say a test case is useful, but for me it's the first time I've even heard of merge tags and I wouldn't know what's needed to support them.

tim-sueberkrueb:
@kalikiana This is a bug in Snapcraft. Specifically in the snapcraft implementation of yaml dict constructors. See this [stackoverflow answer](https://stackoverflow.com/questions/5121931/in-python-how-can-you-load-yaml-mappings-as-ordereddicts/21912744#21912744) and [this](https://stackoverflow.com/a/21048064) is what snapcraft uses right now.

### How do reproduce:
snapcraft.yaml
```yaml
name: my-snap-name
version: '0.1'
summary: Single-line elevator pitch for your amazing snap
description: bla bla
grade: devel
confinement:
parts:
  base: &base
    plugin: nil
    source: .
  my-part:
    # See 'snapcraft plugins'
    plugin: dump
    <<: *base
```
See that PyYAML is fine:
```
python3 -c "import yaml; print(yaml.load(open('snap/snapcraft.yaml', 'r')))"
```
Snapcraft bug:
```
could not determine a constructor for the tag 'tag:yaml.org,2002:merge'
  in "snap/snapcraft.yaml", line 20, column 5
```

>  I'd say a test case is useful, but for me it's the first time I've even heard of merge tags and I wouldn't know what's needed to support them.

This pull request should fix it. I'm wondering what kind of test I should create (integration/unit) and where.

tim-sueberkrueb:
This [yaml handling](https://github.com/snapcore/snapcraft/commit/8f31c6bce4d232f8ca5be29fd0920fcf4dd1cad5#diff-9e7eda5d66239bdff81029374f7e97a5R249) was introduced in #756.

tim-sueberkrueb:
@kalikiana @sergiusens sure, thanks very much for reviewing and accepting this fix :)

PR #1385, opened 2017-06-28
lxd: Don't assume user ID to 1000 for raw.idmap

kalikiana:
Fixes: [bug 1700703](https://bugs.launchpad.net/snapcraft/+bug/1700703)


elopio:
@kalikiana I took your branch, but this still fails for me. Do I have to do something special for this? Could you reproduce my error and did this work for you?
Here is my error: http://paste.ubuntu.com/24978212/

kalikiana:
This branch is fixing the UID mapping issue I saw in your logs, your new paste seems to confirm that it worked since the error is gone.

I don't know what your "Failed to run lxc.hook.pre-start" is about, I see no meaningful error message anymore... can you paste /var/log/lxd/snapcraft-my-snap-name/lxc.conf?
If snapcraft-my-snap-name exists,, can you "lxc delete --force snapcraft-my-snap-name" and try again?

elopio:
Here is the .conf file: http://paste.ubuntu.com/24984714/

sergiusens:
this is what I get on DO http://pastebin.ubuntu.com/25011429/

sergiusens:
That is using lxd from the archives on xenial fwiw

PR #1387, opened 2017-07-01
[WIP] tests: reenable the cleanbuild integration test

elopio:
LP: [#1593529](https://bugs.launchpad.net/snapcraft/+bug/1593529)


integration_tests/test_snap.py
```diff
@@ -81,7 +81,10 @@ def test_snap_default(self):
         self.assertThat(snap_file_path, FileExists())
 
     def test_cleanbuild(self):
-        self.skipTest("Fails to run correctly on travis.")
+        subprocess.check_call(['sudo', 'apt', 'install', '--yes', 'lxd'])
```
kalikiana: We use apt-get elsewhere, so for consistency I'd prefer that. Also because apt causes warning spam (`WARNING: apt does not have a stable CLI interface. Use with caution in scripts.`).

integration_tests/test_snap.py
```diff
@@ -81,7 +81,10 @@ def test_snap_default(self):
...
+        subprocess.check_call(['sudo', 'lxd', 'init', '--auto'])
```
kalikiana: Can we please print `/var/log/lxd/lxd.log` if this fails?

elopio:
This one now requires #1415. It is work in progress.

integration_tests/test_snap.py
```diff
@@ -81,8 +81,15 @@ def test_snap_default(self):
...
-        self.run_snapcraft('cleanbuild', 'assemble')
+        subprocess.check_call(['sudo', 'apt', 'install', '--yes', 'lxd'])
+        subprocess.check_call(['sudo', 'lxd', 'init', '--auto'])
+        subprocess.check_call(['sudo', 'lxc', 'network', 'create', 'testbr0'])
+        subprocess.check_call([
+            'sudo', 'lxc', 'network', 'attach-profile', 'testbr0', 'default',
+            'eth0'
+        ])
+
+        self.run_snapcraft('cleanbuild', 'assemble', sudo=True)
```
kalikiana: Why is sudo needed here?

integration_tests/test_snap.py
```diff
@@ -81,8 +81,15 @@ def test_snap_default(self):
         self.assertThat(snap_file_path, FileExists())
 
     def test_cleanbuild(self):
-        self.skipTest("Fails to run correctly on travis.")
-        self.run_snapcraft('cleanbuild', 'assemble')
+        subprocess.check_call(['sudo', 'apt', 'install', '--yes', 'lxd'])
+        subprocess.check_call(['sudo', 'lxd', 'init', '--auto'])
```
kalikiana: Re-iterating my question from before the latest changes, can we print /var/log/lxd/lxd.log here if this fails? Or consider using --logfile to /dev/stderr instead.

elopio:
@kalikiana can you help me here? https://travis-ci.org/snapcore/snapcraft/jobs/272552380#L1557

I don't understand how is it possible that the tar.bz2 file is not there. I could reproduce it today here, but then trying to understand what was going on I updated and reverted lxd, and now I can no longer reproduce it. :pensive: 
Any idea?

integration_tests/containers/test_cleanbuild.py
```diff
@@ -0,0 +1,44 @@
...
+import integration_tests
+
+
+class CleanbuildTestCase(integration_tests.TestCase):
+
+    def run_snapcraft_cleanbuild(self, project_dir):
+        if project_dir:
+            self.copy_project_to_cwd(project_dir)
+
+        command = ['sudo', self.snapcraft_command, '-d', 'cleanbuild']
```
kalikiana: I like the `containers` category... although I'm wondering how this will look in the context of consolidating integration and unit tests. In unit tests, we have `commands` here, since how you run snapcraft is the relevant bit, not what they do in the background. What's your thoughts on this?

These questions are popping up in my head:
* you're not using `run_snapcraft` anymore
* `sudo` is needed?
* you're using `subprocess.Popen` instead of `subprocess.check_output`

First of all, I'd consider it a good thing to make the use of subprocess more obvious in the test case, since I recently confused an integration test with a unit test even after having written several ones... but I also feel logic like this should be re-used. Maybe something like `snapcraft_process` that returns the process?

And comments on why you're manually handling stdout and why sudo is used would be appreciated.

sergiusens:
stale

PR #1389, opened 2017-07-03
Use newer distro module with no tuple

kalikiana:
Currently there's constant deprecation warning spam that can be seen in Travis and elsewhere:

```
/home/travis/build/snapcore/snapcraft/integration_tests/__init__.py:110: PendingDeprecationWarning: dist() and linux_distribution() functions are deprecated in Python 3.5
  self.distro_series = platform.linux_distribution()[2]
```

The code is also hard to read because it uses tuple indices.

We should be using the newer **distro** module with proper method names.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 click>=6.2
+distro==1.0.4
 configparser==3.5.0
 docopt==0.6.2
 file-magic==0.3.0
--- a/integration_tests/__init__.py
+++ b/integration_tests/__init__.py
@@ -14,9 +14,9 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import distro
 import fileinput
 import os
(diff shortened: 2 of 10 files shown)


requirements.txt
```diff
@@ -1,4 +1,5 @@
 click>=6.2
+distro==1.0.4
```
Conan-Kudo: Please don't do `==` unless there's a really good reason to do this.

sergiusens:
For this to go through and the reason it wasn't proposed yet, you will first have to `NEW` it for everything older than `zesty`:

```
sergiusens@mirkwood:~$ rmadison python3-distro
 python3-distro | 1.0.1-2 | zesty/universe  | all
 python3-distro | 1.0.1-2 | artful/universe | all
sergiusens@mirkwood:~$ 
```

sergiusens:
I am going to close this as there is no chance this would land in the following couple of months given SRU timelines.

PR #1390, opened 2017-07-03
meta: bash completion support

sergiusens:
Support in snapcraft and for snapcraft.
LP: #1590767

### DIFF
--- a/snapcraft/internal/meta.py
+++ b/snapcraft/internal/meta.py
@@ -30,7 +30,9 @@ import yaml
 from snapcraft import file_utils
 from snapcraft import shell_utils
 from snapcraft.internal import common
-from snapcraft.internal.errors import MissingGadgetError
+from snapcraft.internal.errors import (
+    MissingGadgetError,
+    SnapcraftPathEntryError)
 from snapcraft.internal.deprecations import handle_deprecation_notice
 from snapcraft.internal.sources import get_source_handler_from_type
 from snapcraft.internal.states import (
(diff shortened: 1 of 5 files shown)


snapcraft/internal/meta.py
```diff
@@ -458,3 +461,14 @@ def _validate_hook(hook_path):
     if not os.stat(hook_path).st_mode & stat.S_IEXEC:
         asset = os.path.basename(hook_path)
         raise CommandError('hook {!r} is not executable'.format(asset))
+
+
+def _verify_app_paths(basedir, apps):
+    for app in apps:
+        path_entries = [i for i in ('desktop', 'completer')
```
kalikiana: What's the reasoning behind the whitelist? Is this something we could take from the schema?

sergiusens: it is part of the schema but cannot be verified at the time of schema loading. What is the required change?

kalikiana: Right. What I mean is, can we use the values from the schema? Since this code looks like something that's very to track down and nothing says what it means and how these entries are meaningful.

sergiusens:
@elopio the integration test would be just like the unit test though, do we really want that?

snapcraft/internal/meta.py
sergiusens: oh, this would be part of a bigger refactor @elopio might be working on as he has shown interest in refactoring the `meta` module.

sergiusens: This would also potentially be a different schema as this is a post facto schema validation validation.

kalikiana: I guess just document it for now then

sergiusens: I thought the code was self documenting at this point as it is a list-comprehension for `path_entries`

PR #1392, opened 2017-07-05
catkin plugin: extract rosdep into new package

### LINKED ISSUE #1464: Extract rosdep into its own package
[design](https://forum.snapcraft.io/t/plugin-design-how-to-organize-components/1287) #1392

kyrofa:
This PR resolves #1464 by extracting rosdep out of the catkin plugin and placing it into `snapcraft.plugins.internal.pullers`. This is paving the way for support of Ament, ROS 2's build system, which also uses rosdep.

Fixes #1464


sergiusens:
If that goes ahead, this code would also need to be able to work if I do

`cp <project>/plugins/ros snap/plugins/`

kyrofa:
> this first requires a forum post discussing deviating changes on what a plugin is.

[Here you go](https://forum.snapcraft.io/t/plugin-design-how-to-organize-components).

> If that goes ahead, this code would also need to be able to work if I do
>
> `cp <project>/plugins/ros snap/plugins/`

It does, but it's a case that should have tests, agreed.

kyrofa:
This PR has been updated to conform to the new proposal in the forum post referenced above. Rosdep is now in the `pullers` package. The loading code has remained modified to skip over packages, but we can make it skip over a specific list of package names if we prefer. Or we could call it `_pullers` and have it skip over packages with internal names like that.

kyrofa:
This PR is now complete. Instead of a warning, as we discussed, it implements a `snapcraft.plugins.internal` package that includes `pullers`, within which `rosdep` is contained.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -146,7 +146,7 @@ def _load_code(self, plugin_name, properties, part_schema,
         plugin = _get_plugin(module)
         if not plugin:
             raise errors.PluginError(
-                'no plugin found in module: {}'.format(plugin_name))
+                'no plugin found in package: {}'.format(plugin_name))
```
sergiusens: this should stay as is, we deal mostly with modules, not packages.

kyrofa: This error message (which is being added in this PR) is hit in two cases: there's a module with the name of the plugin being loaded, or a package with the name of the plugin being loaded, and something inheriting from `BasePlugin` is not found in them. Either language is probably too specific to work for both cases, so I changed it to `no plugin found in {!r}`.

snapcraft/plugins/catkin.py
```diff
@@ -338,9 +339,13 @@ def pull(self):
         compilers.setup()
 
         # Use rosdep for dependency detection and resolution
-        rosdep = _Rosdep(self.options.rosdistro, self._ros_package_path,
-                         self._rosdep_path, self.PLUGIN_STAGE_SOURCES,
-                         self.project)
+        rosdep = _ros.rosdep.Rosdep(
```
sergiusens: thank you!

snapcraft/plugins/_ros/rosdep.py
```diff
@@ -0,0 +1,146 @@
...
+
+class RosdepDependencyNotFoundError(errors.SnapcraftError):
+    fmt = 'rosdep cannot resolve {dependency!r} into a valid dependency'
+
+    def __init__(self, dependency):
+        super().__init__(dependency=dependency)
+
+
+class Rosdep:
+    def __init__(self, *, ros_distro, ros_package_path, rosdep_path,
```
sergiusens: I like starts

PR #1393, opened 2017-07-06
python plugin: output json in pip list

elopio:
LP: [#1683395](https://bugs.launchpad.net/snapcraft/+bug/1683395)

### DIFF
--- a/snapcraft/plugins/python.py
+++ b/snapcraft/plugins/python.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright (C) 2016 Canonical Ltd
+# Copyright (C) 2016-2017 Canonical Ltd
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License version 3 as
@@ -51,6 +51,7 @@ be preferred instead and no interpreter would be brought in through
 `stage-packages` mechanisms.
 """
 
+import json
 import os
 import re
 import shutil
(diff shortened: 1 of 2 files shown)


snapcraft/plugins/python.py
```diff
@@ -438,17 +439,13 @@ def list(self, exec_func=None):
...
-        for line in output.splitlines():
-            line = line.split()
-            m = version_regex.search(line[1])
-            package_listing[line[0]] = m.group(1)
-
-        return package_listing
+        return {
+            package['name']: package['version']
+            for package in json.loads(output)
+        }
```
kalikiana: Looks lovely, much cleaner.

Please add a test case based on the bug, with a config file init that tries to set the format.

elopio: That would be a python integration test, which is very slow.
I'm happy here just checking that the existing tests are still passing, confirming that this change is transparent.

kalikiana: I can't really agree with your points to be honest, it's just no tested at the end of the day.

Perhaps we should run python integration tests daily if it's that bad?

elopio: We are testing that --json returns a list of packages just in the format we require, which is the only thing I'm after here, because it enables me to rely on that for the next branch about asset tracking.

We are not testing anything related to the local config file. And I'm ok with that because the config affects many more things than the output of pip list. As I mentioned on the bug, that's a separate and bigger problem that should probably be solved with your suggestion about the environment variable.

The solution for the slow integration tests is to run them in parallel. Running lxc in travis will allow us to start working on that.

kalikiana: That makes sense. The way you put this PR in the context of the bug report made it seem to me as if it's addressing the bug, which isn't the case. It's just improving the parsing. A follow-up is going to be needed.

PR #1395, opened 2017-07-07
[WIP] python plugin: track the python packages installed during build

elopio:


elopio:
Closing this because we will got with the record/restore methods.

PR #1396, opened 2017-07-07
rust plugin: unset http_proxy for test_cross_compile.

chihchun:
LP: [#1702886](https://bugs.launchpad.net/snapcraft/+bug/1702886)

The default Debian package build system use pybuild[1] set a default http_proxy, which causes the download test failed.

The patch mock download() function in the test.

[1] https://anonscm.debian.org/cgit/dh-python/dh-python.git/tree/pybuild#n55

### DIFF
--- a/snapcraft/tests/plugins/test_rust.py
+++ b/snapcraft/tests/plugins/test_rust.py
@@ -64,7 +64,8 @@ class RustPluginCrossCompileTestCase(tests.TestCase):
         self.env_mock = patcher.start()
         self.addCleanup(patcher.stop)
 
-    def test_cross_compile(self):
+    @mock.patch('snapcraft.internal.sources._script.Script.download')
+    def test_cross_compile(self, mock_download):
         plugin = rust.RustPlugin('test-part', self.options,
                                  self.project_options)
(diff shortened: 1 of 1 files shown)


snapcraft/tests/plugins/test_rust.py
```diff
@@ -65,6 +65,10 @@ class Options:
         self.addCleanup(patcher.stop)
 
     def test_cross_compile(self):
+        # unset http and https proxies.
+        self.useFixture(fixtures.EnvironmentVariable('http_proxy', None))
+        self.useFixture(fixtures.EnvironmentVariable('https_proxy', None))
```
kalikiana: That seems like the wrong approach since it becomes impossible to run the unit test with a proxy...

More importantly, I don't think the unit test needs to download anything. The run_mock ensures rustup and cargo will never be executed - we should just mock download() as well.

chihchun: I have refreshed the patch to mock download() function. Please kindly review.

chihchun:
Refreshed the patch by mock the download function.

PR #1397, opened 2017-07-07
waf: enable cross-compilation support 

kalikiana:
CC and CXX environment variables need to be set accordingly to cross-compile with Waf.

### DIFF
--- a/snapcraft/plugins/waf.py
+++ b/snapcraft/plugins/waf.py
@@ -75,6 +75,8 @@ class WafPlugin(snapcraft.BasePlugin):
         return env
 
     def enable_cross_compilation(self):
+        # Let snapcraft know that this plugin can cross-compile
+        # If the method isn't implemented an exception is raised
         pass
 
     def build(self):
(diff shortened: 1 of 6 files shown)


sergiusens:
@cpaelzer iirc you wrote this plugin originally, mind taking a quick look?

snapcraft/plugins/waf.py
```diff
@@ -65,6 +65,18 @@ def get_build_properties(cls):
...
+    def env(self, root):
+        env = super().env(root)
+        if self.project.is_cross_compiling:
+            env.extend([
+                'CC={}-gcc'.format(self.project.arch_triplet),
+                'CXX={}-g++'.format(self.project.arch_triplet),
+            ])
+        return env
+
+    def enable_cross_compilation(self):
```
cpaelzer: Something still missing - if so drop this and line 160.
Otherwise I think a comment why it is there for people like me would be very kind.

cpaelzer:
Other than the comment above LGTM, I have nothing around to test it.
But the code reads straight forward and I like tests being added.

PR #1398, opened 2017-07-07
tests: fix issues with python 3.6

sergiusens:
LP: #1699009

### DIFF
--- a/snapcraft/tests/fixture_setup.py
+++ b/snapcraft/tests/fixture_setup.py
@@ -112,6 +112,8 @@ class FakeProjectOptions(fixtures.Fixture):
         patcher.start()
         self.addCleanup(patcher.stop)
 
+        # Special handling is required as ProjectOptions attributes are
+        # handled with the @property decorator.
         project_options_t = type(snapcraft.ProjectOptions.return_value)
         for key in self._kwargs:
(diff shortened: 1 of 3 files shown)


snapcraft/tests/fixture_setup.py
```diff
@@ -90,6 +91,32 @@ def setUp(self):
...
+    def setUp(self):
+        super().setUp()
+
+        patcher = mock.patch('snapcraft.ProjectOptions')
+        patcher.start()
+        self.addCleanup(patcher.stop)
+
+        project_options_t = type(snapcraft.ProjectOptions.return_value)
+        for key in self._kwargs:
+            setattr(project_options_t, key, self._kwargs[key])
```
elopio: I'm not sure I'm understanding this. You get the type of the return value, which is a mock?
I might be misunderstanding your idea, but a named tuple can be sued to convert the kwargs dictionary into an object. Also, the kwargs in a Mock are then turned into attributes:

>>> a = mock.Mock(test='a')
>>> a.test
'a'

Maybe one of those can make this code clearer.

sergiusens: I am just going to say that this is because of the decorator usage.

PR #1399, opened 2017-07-07
ruby plugin: new plugin

jamesbeedy:
LP: [#1702838](https://bugs.launchpad.net/snapcraft/+bug/1702838)

This commit extends the current list of default plugins with a new plugin, 'ruby'.
* ruby added to plugins
* integration test added for the ruby plugin
* modify expected default output for 'list-plugins' to include 'ruby'
* added ruby-exists snapcraft.yaml for testing if ruby exists


snapcraft/plugins/ruby.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def build(self):
+        super().build()
+        self._ruby_install(builddir=self._ruby_part_dir)
+        self._gem_install()
+        if self._install_bundler:
+            self._bundle_install()
+
+    def env(self, root):
+        env = {}
```
kalikiana: This should be done a bit differently:
env = super().env(root)
env.append('RUBYPATH={}'.format(join(root, 'bin')))

snapcraft/plugins/ruby.py
```diff
@@ -0,0 +1,144 @@
...
+    - ruby-version:
+      (string)
+      The version of ruby you want this snap to run.
+"""
+import re
+
+import platform
+
+from os import makedirs, environ
+from os.path import exists, join
```
kalikiana: Please just import "os" and use the functions with their namespace. I realize the snapcraft codebase isn't enitrely consistent here, but for new code it's easier to read if one doesn't need to double-check the imports in different files.

snapcraft/plugins/ruby.py
```diff
@@ -0,0 +1,144 @@
...
+
+        self.build_packages.extend(['gcc', 'g++', 'make', 'zlib1g-dev',
+                                    'libssl-dev', 'libreadline-dev'])
+
+        version_map = [('2.%d.[0-9]' % i, '2.%d.0' % i) for i in range(5)]
+
+        for version_regex, version_dir in version_map:
+            if re.compile(version_regex).match(self._ruby_version):
+                self._ruby_version_dir = version_dir
+                break
```
kalikiana: There should probably be a SnapcraftEnvironmentError if the version dir can't be found?

integration_tests/plugins/test_ruby_plugin.py
```diff
@@ -0,0 +1,27 @@
...
+
+import integration_tests
+
+
+class RubyPluginTestCase(integration_tests.TestCase):
+
+    def test_ruby_exists(self):
+        self.run_snapcraft('stage', 'ruby-exists')
+        ruby_exe = os.path.join(self.stage_dir, 'bin', 'ruby')
+        self.assertTrue(os.path.exists(ruby_exe))
```
kalikiana: Me thinks the test should at least use one gem to prove that pulling a gem works. Better yet, a simple hello world that outputs a string so you can see that it runs (in case ruby wasn't put in the snap or in the wrong location).

kyrofa: I'd like to see a running test as well, though not sure if that belongs in integration or snaps tests.

snapcraft/plugins/ruby.py
kyrofa: Indeed. To clarify, `env()` needs to return a list of "VARIABLE=VALUE" strings, not a dict.

snapcraft/plugins/ruby.py
```diff
@@ -0,0 +1,144 @@
...
+            gem_install_cmd = [join(self.installdir, 'bin', 'ruby'),
+                               join(self.installdir, 'bin', 'gem'), 'install']
+            self.run(gem_install_cmd + self._gems,
+                     env=self.env(root=self.installdir))
+
+    def _bundle_install(self):
+        bundle_install_cmd = [join(self.installdir, 'bin', 'ruby'),
+                              join(self.installdir, 'bin', 'bundle'),
+                              'install']
+        self.run(bundle_install_cmd, env=self.env(root=self.installdir))
```
kyrofa: Is setting the `env` here necessary? I'm under the impression that this behavior should happen anyway.

snapcraft/plugins/ruby.py
kyrofa: Wait... patch releases continue using the minor release dir? That's madness! It's probably better to handle this case on the path discovery side, using globbing (since there should only ever be a single option). Otherwise this plugin needs to change if ruby changes versions in any significant way, which sounds fragile.

snapcraft/plugins/ruby.py
jamesbeedy: 👍

snapcraft/plugins/ruby.py
jamesbeedy: @kyrofa yeah .... "It's probably better to handle this case on the path discovery side, using globbing (since there should only ever be a single option)" - could you elaborate on what this would look like possibly? Are there other plugins that use this method?

"Otherwise this plugin needs to change if ruby changes versions in any significant way" - you mean if ruby default install procedure decides to not use the minor release version for the lib dir any longer? 

@kyrofa  @kalikiana I felt creating a map was the most simple and maintainable way forward here ... can you suggest how this might be done differently?

snapcraft/plugins/ruby.py
jamesbeedy: @kyrofa @kalikiana Thanks for your feedback here. Looking into these things now.

jamesbeedy: @kyrofa "Launchpad builders only provide internet access during the pull step, not the build step" - doesn't this somewhat force us to use the pull step then?

integration_tests/plugins/test_ruby_plugin.py
```diff
@@ -33,6 +33,14 @@ def test_ruby_gem_install_rack(self):
 
     def test_ruby_hello(self):
         self.run_snapcraft('stage', 'ruby-hello')
+        os.environ['RUBYPATH'] = os.path.join(self.stage_dir, 'bin')
```
kalikiana: Why did you change it this way? Maybe this needs more discussion. How about opening a forum topic?

The problem I'm seeing with the approach is that it's not testing the real snap anymore. Integration tests should work in a way that you can build, install and run the same test snap - but the variables are set in the test itself, so there's no guarantee it works outside the test.

jamesbeedy: @kalikiana I did this because I couldn't find any other way to pass stage_dir root to create the env vars needed by ruby.

Do you know a better way to do this?

https://forum.snapcraft.io/t/plugins-setting-env-vars-for-stage-test/1343

snapcraft/plugins/ruby.py
jamesbeedy: @kalikiana where would you like to see the SnapcraftEnvironmentError at?

kalikiana: I was thinking if `self._ruby_version_dir` is not set after the `for` loop, raise the exception, So that code using it afterwards doesn't fail with `name ... is not defined`.

jamesbeedy: @kalikiana done

snapcraft/plugins/ruby.py
```diff
@@ -0,0 +1,142 @@
...
+import os
+import re
+import platform
+
+from snapcraft import BasePlugin
+from snapcraft.internal.errors import SnapcraftEnvironmentError
+from snapcraft.sources import Tar
+
+
+class RubyPlugin(BasePlugin):
```
kyrofa: Do we ever want the Ruby headers and manpages to be included into the snap? If not, consider adding a `snap_fileset` function to filter them out so users of the plugin don't need to do it every time. Something like this:

```python
class RubyPlugin(BasePlugin):
    # ...
    def snap_fileset(self):
        fileset = super().snap_fileset()
        fileset.append('-include/')
        fileset.append('-share/')
        return fileset
```

Untested, but should work.

demos/ruby-hello/ruby-hello
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
kyrofa: I'd rather see this be a ruby script. Something like:

```ruby
#!/usr/bin/env ruby

puts "Hello from Ruby!"
```

snapcraft/plugins/ruby.py
```diff
@@ -0,0 +1,142 @@
...
+            builddir, clean_target=False, keep_tarball=True)
+        self.run(['./configure', '--disable-install-rdoc', '--prefix=/'],
+                 cwd=builddir)
+        self.run(['make', '-j{}'.format(self.parallel_build_count)],
+                 cwd=builddir)
+        self.run(['make', 'install', 'DESTDIR={}'.format(self.installdir)],
+                 cwd=builddir)
+
+    def _gem_install(self):
+        if os.path.exists('Gemfile'):
```
kyrofa: I'd like this to be less magical. Perhaps the packager is packaging an upstream that has a Gemfile that's incorrect, and they're specifying gems instead. I suggest encapsulating this functionality in a boolean option called something like `use-bundler`.

demos/ruby-hello/ruby-hello
jamesbeedy: @kyrofa the issue with that is /usr/bin/env doesn't exist ...

snapcraft/plugins/ruby.py
jamesbeedy: @kyrofa "so users of the plugin don't need to do it every time" - what are you thinking happens here? I don't get those dirs unless I specify to bring them forward with the stage/prime filesets ....

snapcraft/plugins/ruby.py
```diff
@@ -0,0 +1,142 @@
...
+        # Raise exception when version_dir not found
+        if not self._ruby_version_dir:
+            raise SnapcraftEnvironmentError(
+                'Ruby version dir not in version_map.')
+
+    def pull(self):
+        super().pull()
+        os.makedirs(self._ruby_part_dir, exist_ok=True)
+        self._ruby_tar.download()
+        self._ruby_install(builddir=self._ruby_part_dir)
```
kyrofa: There's a problem with having this in the pull step: it's installing into the installdir, which is cleaned when the build step is cleaned. So even though we do all this work in the pull step, it can be completely blown away if the build step is cleaned. We need a better solution for this. I think the only real option is to do it all in its own unique area and copy it into the installdir in the pull step as well as the build step.

demos/ruby-hello/ruby-hello
kyrofa: Sure it does. I just tested it.

snapcraft/plugins/ruby.py
kyrofa: You don't get those dirs because you're explicitly filtering them out by providing a whitelist via `stage` and `prime`. Do we want to require our users to do that as well, or just have the plugin blacklist the stuff we know is useless?

snapcraft/plugins/ruby.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def _ruby_install(self, builddir):
+        self._ruby_tar.provision(
+            builddir, clean_target=False, keep_tarball=True)
+        self.run(['./configure', '--disable-install-rdoc', '--prefix=/'],
+                 cwd=builddir)
+        self.run(['make', '-j{}'.format(self.parallel_build_count)],
+                 cwd=builddir)
+        self.run(['make', 'install', 'DESTDIR={}'.format(self.installdir)],
+                 cwd=builddir)
```
kyrofa: I suggest adding a quick cleanup step at the end of this function to make the installed binaries have useful shebangs:

```python
# Fix all shebangs to use the in-snap ruby
file_utils.replace_in_file(self.installdir, re.compile(r''),
                           re.compile(r'^#!.*ruby'),
                           r'#!/usr/bin/env ruby')
```

That allows one to use, say, the `gem` command in the `build` scriptlet. Note that you'll need to import `file_utils` from `snapcraft`.

demos/ruby-hello/ruby-hello
kyrofa: (note that it comes from the core snap)

snapcraft/plugins/ruby.py
kyrofa: I want to redesign this bit, but it can happen after the plugin lands and we can iterate.

demos/ruby-hello/ruby-hello
jamesbeedy: @kyrofa /usr/bin/env doesn't exist in the snap dir... Is it accessing system /usr/bin/env in that case?

jamesbeedy: (note that it comes from the core snap) - what does this mean?

kyrofa: Ah, snaps do not execute all by themselves. They have no bash, no dash, no rootfs. Snaps "execution environment" if you will is the core snap, that's why it exists. It holds all those common things, including `/usr/bin/env`.

snapcraft/plugins/ruby.py
jamesbeedy: @kyrofa makes total sense now. I'll add these bits.

kyrofa:
With the changes we discussed, @elopio this works:

```yaml
name: travis
version: '0.1'
summary: my summary
description: my description

grade: devel
confinement: strict

apps:
  travis:
    command: travis
    plugs: [network, home]

parts:
  my-part:
    plugin: ruby
    build: gem build travis.gemspec
    install: gem install travis-*.gem --env-shebang
```

Eh? Eh?

demos/ruby-hello/ruby-hello
jamesbeedy: ok, I'll give this a closer look. Thanks for explaining

snapcraft/plugins/ruby.py
```diff
@@ -23,6 +23,9 @@
     - gems:
       (list)
       A list of gems to install.
+    - use-bundler
+      (boolean)
+      Use bundler to install gems from a Gemfile.
```
kyrofa: I suggest documenting here that the default is false.

snapcraft/plugins/ruby.py
```diff
@@ -90,13 +97,19 @@ def __init__(self, name, options, project):
...
+        return fileset
+
     def pull(self):
         super().pull()
         os.makedirs(self._ruby_part_dir, exist_ok=True)
         self._ruby_tar.download()
         self._ruby_install(builddir=self._ruby_part_dir)
         self._gem_install()
-        if self._install_bundler:
+        if self._use_bundler:
```
kyrofa: Why not just evaluate `selfl.options.use_bundler` here instead of saving it off?

snapcraft/plugins/ruby.py
```diff
@@ -124,10 +137,12 @@ def _ruby_install(self, builddir):
                  cwd=builddir)
         self.run(['make', 'install', 'DESTDIR={}'.format(self.installdir)],
                  cwd=builddir)
+        # Fix all shebangs to use the in-snap ruby
+        file_utils.replace_in_file(self.installdir, re.compile(r''),
+            re.compile(r'^#!.*ruby'), r'#!/usr/bin/env ruby')
```
kyrofa: This is probably a static error.

snapcraft/plugins/ruby.py
kyrofa: I suggest not worrying about this for now, there are some more core issues here. We can fix it on iteration.

snapcraft/plugins/ruby.py
```diff
@@ -0,0 +1,160 @@
...
+
+class RubyPlugin(BasePlugin):
+
+    @classmethod
+    def schema(cls):
+        schema = super().schema()
+
+        schema['properties']['use-bundler'] = {
+            'type': 'boolean',
+            'default': 'false'
```
kyrofa: This needs to be the python value `False`, not the string `'false'`. I see another plugin that made that error as well... my own :cry: .

snapcraft/plugins/ruby.py
```diff
@@ -0,0 +1,160 @@
...
+            re.compile(r'^#!.*ruby'),
+            r'#!/usr/bin/env ruby')
+
+    def _gem_install(self):
+        if self._use_bundler:
+            self._gems = self._gems + ['bundler']
+        if self._gems:
+            gem_install_cmd = [os.path.join(self.installdir, 'bin', 'ruby'),
+                               os.path.join(self.installdir, 'bin', 'gem'),
+                               'install']
```
kyrofa: You'll need to add `--env-shebang` to the arglist here to make the gems usable (it adds the `/usr/bin/env` shebang instead of hard-coding a path to the part's installdir).

snapcraft/plugins/ruby.py
```diff
@@ -0,0 +1,160 @@
...
+        }
+        return schema
+
+    @classmethod
+    def get_pull_properties(cls):
+        # Inform Snapcraft of the properties associated with pulling. If these
+        # change in the YAML Snapcraft will consider the build step dirty.
+        return ['ruby-version', 'gems']
+
+    def __init__(self, name, options, project):
```
kyrofa: I'd like to tag this plugin as being "in beta" so we can change its options a bit while it gets some use. Would you mind adding a warning (using logger.warn) to the __init__ saying something like "The ruby plugin is currently in beta, its API may break. Use at your own risk".

snapcraft/plugins/ruby.py
```diff
@@ -0,0 +1,160 @@
...
+from snapcraft.sources import Tar
+
+
+class RubyPlugin(BasePlugin):
+
+    @classmethod
+    def schema(cls):
+        schema = super().schema()
+
+        schema['properties']['use-bundler'] = {
```
kyrofa: I'm rethinking this. Bundler supports being pointed to another gemfile path with `--gemfile`. I wonder if we could add a `gemfile` option that defaults to None or an empty string. Then if the user wants to enable bundler, they specify what gemfile they want to use (in most cases I expect it'll simply be 'Gemfile', but by doing it this way we get bundler support and alternative gemfile support with one stone).

I also suggest adding an option to work with `--without`, I imagine lots of users will want to exclude their test and development groups. Perhaps an option called `bundler-without-groups` that accepts a list.

snapcraft/plugins/ruby.py
jamesbeedy: @kyrofa 
```python
    def _bundle_install(self):
        bundle_install_cmd = ['ruby', 'bundle', 'install', '--env-shebang']
        self.run(bundle_install_cmd)
```
like this?

kyrofa: No, bundler does this automatically, I'm talking about when you call `gem install`.

kyrofa:
In order to exercise this PR, I suggest three renditions of the same Travis snap. Test them all by installing them and running `travis`. Make sure none of them blow up.

## Snap 1: Using the existing gem

```yaml
name: travis
version: '0.1'
summary: summary
description: description

grade: devel
confinement: strict

apps:
  travis:
    command: travis
    plugs: [network, home]

parts:
  my-part:
    plugin: ruby
    gems: [travis]
```

## Snap 2: Build/install the gem itself

```yaml
name: travis
version: '0.1'
summary: summary
description: description

grade: devel
confinement: strict

apps:
  travis:
    command: travis
    plugs: [network, home]

parts:
  my-part:
    plugin: ruby
    source: https://github.com/travis-ci/travis.rb.git
    build: gem build travis.gemspec
    install: gem install travis-*.gem --env-shebang
```

## Snap 3: From gemfile

```yaml
name: travis
version: '0.1'
summary: summary
description: description

grade: devel
confinement: strict

apps:
  travis:
    command: travis
    plugs: [network, home]

parts:
  my-part:
    plugin: ruby
    use-bundler: true
```

In the root of the project (alongside the `snap/` dir), create a Gemfile with the following contents:

```
source 'https://rubygems.org'

gem 'travis'
```

snapcraft/plugins/ruby.py
```diff
@@ -0,0 +1,157 @@
...
+                    "its API may break. Use at your own risk")
+
+        self._ruby_version = options.ruby_version
+        self._ruby_part_dir = os.path.join(self.partdir, 'ruby')
+        self._ruby_download_url = \
+            'https://cache.ruby-lang.org/pub/ruby/ruby-{}.tar.gz'.format(
+                self._ruby_version)
+        self._ruby_tar = Tar(self._ruby_download_url, self._ruby_part_dir)
+        self._gems = options.gems or []
+        self._ruby_version_dir = '{}.{}.0'.format(
```
jamesbeedy: @kyrofa how to you feel about this as a step forward here?

kyrofa: It's still pretty hacky. I was okay waiting until later, but if you want to re-design this bit now I've actually pushed a commit here to show you what I had in mind. It uses globbing to discover the version, so no programatic trickery is required. Feel free to delete the commit if you dislike it.

kyrofa:
I'm quite pleased with the state of this plugin. There are still a few changes to be made, but they can be made later too as long as we leave the "beta" flag in place. The only thing preventing my hearty +1 is a lack of unit tests.

elopio:
@kalikiana, please check again, to see if you +1 this. Also @kyrofa, you never left your vote.

kyrofa:
> Also @kyrofa, you never left your vote.

Indeed, we discussed this in chat but I'll bring it back to the PR so as to not leave the reader hanging: I am not a +1 until there are unit tests.

snapcraft/plugins/ruby.py
```diff
@@ -0,0 +1,193 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""The ruby plugin is useful for ruby based parts.
+The plugin uses gem to install dependencies from a `Gemfile`.
```
kalikiana: How about "...uses gem or bundler to install dependencies from a `Gemfile`"?

snapcraft/plugins/ruby.py
kalikiana: How about supporting something like `gems: [spam, -eggs]` to include and exclude gems like with `prime`?

kyrofa: I'm not sure such functionality makes sense: it'd be like adding the ability to use `python-packages: [spam, -eggs]` to the python plugin. Why not just not include the gem in the first place?

snapcraft/plugins/ruby.py
kyrofa: Neither is correct, actually. `gem` is used to install individual gems, `bundler` is used for the Gemfile.

elopio:
closed in favor of #1544, which adds the unit tests.

snapcraft/plugins/ruby.py
kalikiana: You suggested it in the first place, I only came up with an alternative syntax ;-)

PR #1400, opened 2017-07-10
lxd: Distingish FileNotFoundError if not installed

kalikiana:
If LXD is not installed the exception FileNotFoundError will be raised because the 'lxc' executable doesn't exist. The code currently only handles CalledProcessError.

Fixes: [bug 1703344](https://bugs.launchpad.net/snapcraft/+bug/1703344)

### DIFF
--- a/snapcraft/tests/test_lxd.py
+++ b/snapcraft/tests/test_lxd.py
@@ -166,15 +166,15 @@ class LXDTestCase(tests.TestCase):
 
     @patch('snapcraft.internal.lxd.Cleanbuilder._container_run')
     def test_lxc_check_fails(self, mock_run):
-        self.useFixture(tests.fixture_setup.FakeLXD(fail_on_default=True))
+        fake_lxd = tests.fixture_setup.FakeLXD()
+        self.useFixture(fake_lxd)
+        fake_lxd.check_output_mock.side_effect = FileNotFoundError('lxc')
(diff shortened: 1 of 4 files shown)


sergiusens:
I am divided by this, I like keeping the complications in the fixture to make the test more descriptive

snapcraft/tests/test_lxd.py
```diff
@@ -165,15 +165,15 @@ def run_effect(*args, **kwargs):
 
     @patch('snapcraft.internal.lxd.Cleanbuilder._container_run')
     def test_lxc_check_fails(self, mock_run):
-        self.useFixture(tests.fixture_setup.FakeLXD(fail_on_default=True))
+        fake_lxd = tests.fixture_setup.FakeLXD()
+        self.useFixture(fake_lxd)
+        fake_lxd.check_output_mock.side_effect = FileNotFoundError('lxc')
```
sergiusens: There is a coincidence today that the first `check_call` or `check_output` call is for `lxc` but to be complete and future proof we should really be checking if `args[0]` includes `lxc` in the command and if not, not raise the exception.

sergiusens: Random idea, if you don't want to complicate the fixture, pass it a `lambda` or something with conditions to check for first or last.

PR #1402, opened 2017-07-11
Better explain dependency link processing

evandandrea:
As covered in https://github.com/CanonicalLtd/snappy-docs/pull/87#discussion_r126335736, better explain dependency link processing.

### DIFF
--- a/snapcraft/plugins/python.py
+++ b/snapcraft/plugins/python.py
@@ -36,7 +36,8 @@ Additionally, this plugin uses the following plugin-specific keywords:
       Path to a constraints file
     - process-dependency-links:
       (bool; default: false)
-      Enable the processing of dependency links.
+      Enable the processing of dependency links in pip, which allow one
+      project to provide places to look for another project
     - python-packages:
       (list)
       A list of dependencies to get from PyPi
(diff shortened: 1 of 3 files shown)


snapcraft/plugins/python.py
```diff
@@ -36,7 +36,8 @@
       Path to a constraints file
     - process-dependency-links:
       (bool; default: false)
-      Enable the processing of dependency links.
+      Enable the processing of dependency links in pip, which allow one
```
sergiusens: allows?

snapcraft/plugins/python2.py
```diff
@@ -39,7 +39,8 @@
       Path to a constraints file
     - process-dependency-links:
       (bool; default: false)
-      Enable the processing of dependency links.
+      Enable the processing of dependency links in pip, which allow one
```
sergiusens: allows?

snapcraft/plugins/python3.py
```diff
@@ -39,7 +39,8 @@
       Path to a constraints file
     - process-dependency-links:
       (bool; default: false)
-      Enable the processing of dependency links.
+      Enable the processing of dependency links in pip, which allow one
```
sergiusens: allows?

snapcraft/plugins/python.py
evandandrea: You wouldn't say "dependency links allows one project to provide …", you'd say "dependency links allow one project to provide…"

sergiusens: right, I read without the back context.

PR #1403, opened 2017-07-12
lxd: Only remove container if one exists

kalikiana:
If creating a LXD container fails snapcraft always produces two exceptions because it will attempt to remove the non-existing container - this adds a check to avoid that.

Not sure about unit tests... we'd need to know that no exception was raised for 'lxc stop'.

### DIFF
--- a/snapcraft/internal/lxd.py
+++ b/snapcraft/internal/lxd.py
@@ -102,9 +102,17 @@ class Containerbuild:
             self._ensure_container()
             yield
         finally:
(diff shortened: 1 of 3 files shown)


elopio:
This seems good to me. With my mess not being able to run lxc containers in your branches, I kept getting the error message on stop which just added noise.

About tests, you can patch status to not return anything, and check that in that case stop is not called. I know that the lxc mocks are becoming unwieldy, but it's doable. What do you think?

PR #1404, opened 2017-07-12
cleanbuild: ensure we enter a shell on failures on --debug

sergiusens:
No matter how debug was positioned, a shell needs to be entered if a cleanbuild fails.

### DIFF
--- a/snapcraft/cli/__init__.py
+++ b/snapcraft/cli/__init__.py
@@ -89,7 +89,6 @@ class SnapcraftGroup(click.Group):
 @click.option('--debug', '-d', is_flag=True)
 def run(ctx, debug, catch_exceptions=False, **kwargs):
     """Snapcraft is a delightful packaging tool."""
-    ctx.obj['debug'] = debug
     if debug:
         log_level = logging.DEBUG
         click.echo('Starting snapcraft {} from {}.'.format(
--- a/snapcraft/cli/__init__.py
+++ b/snapcraft/cli/__init__.py
@@ -89,6 +89,7 @@ class SnapcraftGroup(click.Group):
 @click.option('--debug', '-d', is_flag=True)
 def run(ctx, debug, catch_exceptions=False, **kwargs):
     """Snapcraft is a delightful packaging tool."""
+    ctx.obj['debug'] = debug
     if debug:
         log_level = logging.DEBUG
         click.echo('Starting snapcraft {} from {}.'.format(
--- a/snapcraft/cli/_options.py
+++ b/snapcraft/cli/_options.py
@@ -59,7 +59,7 @@ def get_project_options(**kwargs):
             kwargs[key] = value
 
     project_args = dict(
-        debug=ctx.obj['debug'] or kwargs.get('debug'),
+        debug=kwargs.pop('debug'),
         use_geoip=kwargs.pop('enable_geoip'),
         parallel_builds=not kwargs.pop('no_parallel_builds'),
         target_deb_arch=kwargs.pop('target_arch'),
--- a/snapcraft/cli/_options.py
+++ b/snapcraft/cli/_options.py
@@ -59,6 +59,7 @@ def get_project_options(**kwargs):
             kwargs[key] = value
 
     project_args = dict(
+        debug=ctx.obj['debug'] or kwargs.get('debug'),
         use_geoip=kwargs.pop('enable_geoip'),
         parallel_builds=not kwargs.pop('no_parallel_builds'),
         target_deb_arch=kwargs.pop('target_arch'),
--- a/snapcraft/tests/commands/__init__.py
+++ b/snapcraft/tests/commands/__init__.py
@@ -74,9 +74,11 @@ class CommandBaseTestCase(tests.TestCase):
     def setUp(self):
         super().setUp()
         self.runner = CliRunner()
+        self.obj = dict()
 
     def run_command(self, args, **kwargs):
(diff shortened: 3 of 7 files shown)


snapcraft/tests/commands/__init__.py
```diff
@@ -74,9 +74,11 @@ class CommandBaseTestCase(tests.TestCase):
     def setUp(self):
         super().setUp()
         self.runner = CliRunner()
+        self.obj = dict()
```
elopio: it's so sad that click is using a name like obj. Even something like `user_data` would be better here. Oh well...

snapcraft/tests/fixture_setup.py
```diff
@@ -479,10 +481,13 @@ def call_effect(*args, **kwargs):
                             }).encode('utf-8')
                 return '[]'.encode('utf-8')
             elif args[0][:2] == ['lxc', 'list'] and self.fail_on_remote:
-                    raise CalledProcessError(returncode=255, cmd=args[0])
+                raise CalledProcessError(returncode=255, cmd=args[0])
             elif args[0][:2] == ['lxc', 'init']:
                 self.name = args[0][3]
                 self.status = 'Stopped'
+            elif ('snapcraft' in args[0] and 'apt-get' not in args[0]
```
elopio: is there a case where we have snapcraft and apt-get in the same command?

snapcraft/cli/_options.py
```diff
@@ -59,6 +59,7 @@ def get_project_options(**kwargs):
             kwargs[key] = value
 
     project_args = dict(
+        debug=ctx.obj['debug'] or kwargs.get('debug'),
```
kalikiana: Why is `ctx.obj['debug']` needed here? We have **kwargs** and **ctx.parent.params**. The existing code already handles  the parent context for options that exist on multiple levels.

snapcraft/tests/fixture_setup.py
sergiusens: yes, `apt-get install snapcraft` :wink:

snapcraft/cli/_options.py
sergiusens: good catch!

snapcraft/tests/fixture_setup.py
```diff
@@ -479,10 +481,15 @@ def call_effect(*args, **kwargs):
...
-                    raise CalledProcessError(returncode=255, cmd=args[0])
+                raise CalledProcessError(returncode=255, cmd=args[0])
             elif args[0][:2] == ['lxc', 'init']:
                 self.name = args[0][3]
                 self.status = 'Stopped'
+            # Fail on an actual snapcraft command and not the command
+            # for the installation of it.
+            elif ('snapcraft' in args[0] and 'apt-get' not in args[0]
+                  and self.fail_on_snapcraft_run):
+                raise CalledProcessError(returncode=255, cmd=args[0])
```
kalikiana: Can you please raise the error directly in the unit test? I'm removing the existing ones here https://github.com/snapcore/snapcraft/pull/1400/files - you can get the same effect by overriding `fake_lxd.check_output_mock.side_effect` and without having to complicate the fixture.

snapcraft/cli/__init__.py
```diff
@@ -89,6 +89,7 @@ def get_command(self, ctx, cmd_name):
 @click.option('--debug', '-d', is_flag=True)
 def run(ctx, debug, catch_exceptions=False, **kwargs):
     """Snapcraft is a delightful packaging tool."""
+    ctx.obj['debug'] = debug
     if debug:
```
kalikiana: So ctx.obj shouldn't be needed then? Same goes for the lifecycle and commands test changes.

snapcraft/tests/fixture_setup.py
sergiusens: I will have to move everything out of the fixture though, this happens late in the game.

snapcraft/cli/__init__.py
sergiusens: this is needed, the obj is originally empty (has nothing).

kalikiana: What's it needed for if we're getting debug from kwargs? I don't see it being used for anything else. It contained the ProjectOptions before and even in master I don't see anything using obj. Is it indirectly being used some other way?

sergiusens: it is needed for the case `snapcraft --debug cleanbuild` the sole reason for this. If `--debug` comes before the command it is not passed on to the sucommand

snapcraft/tests/fixture_setup.py
sergiusens: I tried, it is just moving the complication and making the test harder to read, which is the reason we use fixtures in the first place, so a no on this, at least on this branch.

snapcraft/cli/__init__.py
sergiusens: Ok, you go what you wanted, I was mistaken on the behavior of `get_current_context`.

Please validate it all still works.

kalikiana: But you removed the use of obj in get_project_options... I thought it works like other options (geoip, parallel_builds, target_arch) now via params?

snapcraft/tests/fixture_setup.py
kalikiana: Aye, it was straight-forward fpr the existing errors, let's leave it for now.

PR #1406, opened 2017-07-14
recording: rename the file to manifest.yaml

elopio:


snapcraft/internal/meta.py
```diff
@@ -201,7 +201,7 @@ def write_snap_directory(self):
 
                     file_utils.link_or_copy(source, destination)
 
-        self._record_snapcraft()
+        self._record_manifest()
```
kalikiana: I like the renaming. Actually makes it easier to ack the code for manifest and find relevant test cases as well.

sergiusens: :+1:

PR #1407, opened 2017-07-14
 recording: record the original snapcraft.yaml

elopio:


elopio:
Requires #1406

snapcraft/tests/test_lifecycle.py
```diff
@@ -678,6 +686,36 @@ def test_prime_with_build_info_records_snapcraft_yaml(self):
 description: test
 confinement: strict
 grade: stable
+
+
```
kalikiana: Any idea where that extra space is coming from? I was going to suggest fixing it before I realized it's the expected output...

snapcraft/tests/test_lifecycle.py
```diff
@@ -689,7 +727,7 @@ def test_prime_with_build_info_records_snapcraft_yaml(self):
 build-packages: []
 """.format(self.project_options.deb_arch))
         self.assertThat(
-            os.path.join('prime', 'snap', 'snapcraft.yaml'),
+            os.path.join('prime', 'snap', 'manifest.yaml'),
```
kalikiana: Should we be checking that manifest.yaml is a subset of snapcraft.yaml?

snapcraft/tests/test_lifecycle.py
elopio: This is because the make_snapcraft_yaml helper leaves room in case you want to set the type. But when you don't set the type, it still leaves the empty line. I could fix it, but that would make the helper more complicated, and this is the only place where we actually check the full contents, so I would prefer not to.

elopio: oh, I know. I set the type, and then it will look nicer.

snapcraft/internal/meta.py
```diff
@@ -134,15 +136,21 @@ def setup_assets(self):
...
+        manifest_file_path = os.path.join(prime_snap_dir, 'manifest.yaml')
         if os.path.isfile(manifest_file_path):
             os.unlink(manifest_file_path)
 
         # FIXME hide this functionality behind a feature flag for now
         if os.environ.get('SNAPCRAFT_BUILD_INFO'):
-            os.makedirs(manifest_dir, exist_ok=True)
+            os.makedirs(prime_snap_dir, exist_ok=True)
+            shutil.copy2(
+                self._snapcraft_yaml_path, recorded_snapcraft_yaml_path)
```
elopio: There were a few issues here.
Previously, we were recording the snapcraft.yaml even when the flag is not set, which is leaking information. But that happened only when the yaml was in snap/snapcraft.yaml. The other two locations were ignored.
I saw four options to solve this. One, find the snapcraft.yaml in meta. That's basically, do the same thing we do in load, once at the start and once at the end. The second, save in the project options the path of the loaded yaml. Third, make public the snapcraft_yaml_path attribute that we have in config, and pass it to the function that creates meta at the end. Fourth, pass to meta the whole config, not just the loaded data.
I hated all those options, so I went with the easiest, which was #3.

snapcraft/tests/test_lifecycle.py
elopio: It's the other way around. manifest will include more information.
But that's what all the tests are checking. I added here one test to check that the recorded snapcraft.yaml is a pristine copy, and all the tests check pieces of information that are added to the original and then recorded in manifest.yaml

snapcraft/tests/test_lifecycle.py
```diff
@@ -649,9 +649,14 @@ def test_clean_removes_global_state(self):
             Not(DirExists()))
 
 
-class RecordManifestTestCase(BaseLifecycleTestCase):
+class BuildInfoFlagUnsetTestCase(BaseLifecycleTestCase):
 
-    def test_prime_without_build_info_does_not_record_manifest(self):
+    scenarios = (
```
sergiusens: I don't understand why this needs to be a scenario instead of a check for each at the end of the test. Can you explain?

elopio: You are right. This just makes it slower. I'm changing it.

PR #1408, opened 2017-07-14
lxd: Stop setting $HOME in containers

kalikiana:
Because the project folder is mounted into the container and `$HOME` points at it, .cache ends up in the project folder. We need to stop setting `$HOME`.

This addresses one aspect of bug [1703642](https://bugs.launchpad.net/snapcraft/+bug/1703642).

### DIFF
--- a/snapcraft/internal/lxd.py
+++ b/snapcraft/internal/lxd.py
@@ -190,14 +190,17 @@ class Project(Containerbuild):
         if not self._get_container_status():
             check_call([
                 'lxc', 'init', self._image, self._container_name])
+        if self._get_container_status()['status'] == 'Stopped':
             check_call([
                 'lxc', 'config', 'set', self._container_name,
                 'environment.SNAPCRAFT_SETUP_CORE', '1'])
+            check_call([
+                'lxc', 'config', 'set', self._container_name,
+                'environment.XDG_CACHE_HOME', '/run/'])
             # Map host user to root inside container
             check_call([
                 'lxc', 'config', 'set', self._container_name,
                 'raw.idmap', 'both {} 0'.format(os.getuid())])
-        if self._get_container_status()['status'] == 'Stopped':
             check_call([
                 'lxc', 'start', self._container_name])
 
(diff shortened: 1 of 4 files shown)


elopio:
Saviq left a question in the bug, please reply that one. And I have another.

Why are we setting home to be the mounted project directory? 

I find this can cause more confusions along the way. If I want to debug my container, I would expect the cache to be in /root/.cache, I wouldn't think about finding it in /run/.

So why don't we leave $HOME to /root, just not touching it. Then we cd into /project-dir, or even /root/project-dir to execute the snapcraft command? I might be missing something obvious here, but that's what I would look for if I bash into the container to debug.

Saviq:
A good first step :)

snapcraft/internal/lxd.py
```diff
@@ -74,16 +75,20 @@ def _get_remote_info(self):
...
-        check_call(['lxc', 'exec', self._container_name, '--env',
-                   'HOME=/{}'.format(self._project_folder), '--'] + cmd)
+    def _container_run(self, cmd, cwd=None):
+        if cwd:
+            check_call(['lxc', 'exec', self._container_name, '--',
+                        'bash', '-c', 'cd {}; {}'.format(
+                            cwd,
+                            ' '.join(pipes.quote(arg) for arg in cmd))])
+        else:
+            check_call(['lxc', 'exec', self._container_name, '--'] + cmd)
```
elopio: I would like this more with only one check_call. Like:
```
if cwd:
    cmd = ['bash', '-c', 'cd {}; {}'.format(cwd, ...)]
check_call(['lxc', ...] + cmd)
```

snapcraft/internal/lxd.py
```diff
@@ -137,7 +142,8 @@ def _setup_project(self):
                            os.path.basename(tar_filename))
         self._container_run(['mkdir', self._project_folder])
         self._push_file(tar_filename, dst)
-        self._container_run(['tar', 'xvf', os.path.basename(tar_filename)])
+        self._container_run(['tar', 'xvf', os.path.basename(tar_filename)],
+                            cwd=self._project_folder)
```
elopio: I learned that in the upcoming lxc version, it's possible to just push the directory.
Not very useful though, as we have to support the version in trusty. Just fyi.

PR #1410, opened 2017-07-14
tests: remove download timeout workaround

elopio:
The windowing issue was resolved by the CDN provider.


sergiusens:
shouldn't this be `git revert` on the commit?

sergiusens:
Closing in favour of #1423 which is the result of running `git revert <commit>`

PR #1411, opened 2017-07-15
tests: build the snapcraft snap in travis tests

elopio:


elopio:
Ah, damn, something obvious here that will make it fail: we can't install the core snap in the snapcore/snapcraft docker image.
I'm talking to travis support to be able to ssh into the builder, and debug the issues with the lxc networking. It seems we better wait in here for that lxc container, too.

elopio:
This now requires #1415

sergiusens:
the best next step would be to run spread against this using the lxd backend

PR #1413, opened 2017-07-18
core: minimal windows support.

sergiusens:
Add minimal windows support.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,8 +18,11 @@
 import codecs
 import os
 import re
+import sys
 
 from setuptools import setup
+if sys.platform == 'win32':
+    import py2exe
 
 version = 'devel'
 # look/set what version we have
@@ -51,7 +54,13 @@ setup(
               'snapcraft.plugins',
               'snapcraft.storeapi'],
     package_data={'snapcraft.internal.repo': ['manifest.txt']},
-    scripts=['bin/snapcraft', 'bin/snapcraft-parser'],
+    scripts=['bin/snapcraft-parser'],
+    entry_points={
+        'console_scripts': [
+            'snapcraft = snapcraft.cli.__main__:run',
+            'snapcraft-parser = snapcraft.internal.parser:main',
+        ],
+    },
     data_files=[
         ('share/snapcraft/schema',
             ['schema/' + x for x in os.listdir('schema')]),
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright (C) 2015 Canonical Ltd
+# Copyright (C) 2015-2017 Canonical Ltd
 #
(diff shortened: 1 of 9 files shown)


snapcraft/_options.py
```diff
@@ -242,7 +243,16 @@ def _set_machine(self, target_deb_arch):
...
+        if sys.platform == 'linux':
+            self.__machine_info = _ARCH_TRANSLATIONS[self.__target_machine]
+        else:
+            self.__machine_info = {
+                'kernel': 'NOBUILD',
+                'deb': 'NOBUILD',
+                'uts_machine': 'NOBUILD',
+                'triplet': 'NOBUILD',
+                'core-dynamic-linker': 'NOBUILD',
+            }
```
kalikiana: With the placeholders it looks a little cryptic... can you add a comment here to clarify what the code does?

Maybe `# Provider fake machine info to allow container and store commands to work on non-Linux platforms`.

setup.py
```diff
@@ -51,7 +51,12 @@
               'snapcraft.plugins',
               'snapcraft.storeapi'],
     package_data={'snapcraft.internal.repo': ['manifest.txt']},
-    scripts=['bin/snapcraft', 'bin/snapcraft-parser'],
+    entry_points={
+        'console_scripts': [
+            'snapcraft = snapcraft.cli.__main__:run',
+            'snapcraft-parser = snapcraft.internal.parser:main',
+        ],
+    },
```
kalikiana: I have no idea what difference this makes... can you add a comment? Like `scripts isn't good enough because...`.

elopio: Can we delete the files in bin now?
Also, they were doing a very ugly thing calling snapcraft.internal.dirs.setup_dirs. Do we no longer need that? I would be very happy if so.

elopio:
If you get this tested in a windows machine, I'm happy.
I left a question about the ugly things that we are doing in bin/snapcraft.

sergiusens:
@elopio this was developed and tested on Windows :wink:

setup.py
sergiusens: @kalikiana this is just standard python, it is a portable way to deliver.
@elopio not ready to delete the scripts in `bin` until we say we don't care about local development anymore.

kalikiana: If it's standard, why didn't we do it this way before? That's what I mean.

sergiusens: Oh, because the person that did this originally wasn't a python developer and we never got to change it.

kyrofa: `scripts` are the old distutils way of doing things, `entry_points` are setuptools. They're the [recommended way of doing this](https://packaging.python.org/tutorials/distributing-packages/#scripts), particularly when you're talking [cross-platform stuff](https://setuptools.readthedocs.io/en/latest/setuptools.html#automatic-script-creation).

elopio:
+1 for me. I'm going to trigger the autopkgtests for xenial, just in case.
And it would be nice to comment why on windows a new pymacaroons is being installed.

PR #1414, opened 2017-07-18
cmake plugin: call the cmake using build dir as source

elopio:
LP: [#1704933](https://bugs.launchpad.net/snapcraft/+bug/1704933)

### DIFF
--- a/snapcraft/plugins/cmake.py
+++ b/snapcraft/plugins/cmake.py
@@ -64,22 +64,15 @@ class CMakePlugin(snapcraft.plugins.make.MakePlugin):
         self.build_packages.append('cmake')
 
     def build(self):
-        if os.path.exists(self.builddir):
-            shutil.rmtree(self.builddir)
-        os.mkdir(self.builddir)
-
-        source_subdir = getattr(self.options, 'source_subdir', None)
-        if source_subdir:
-            sourcedir = os.path.join(self.sourcedir, source_subdir)
-        else:
-            sourcedir = self.sourcedir
-
         env = self._build_environment()
+        # It's a good practice to call cmake from an empty directory, not part
+        # of the project source. Some projects even enforce this.
+        cmake_builddir = os.path.join(self.builddir, 'cmake_build')
+        os.mkdir(cmake_builddir)
+        self.run(['cmake', self.builddir, '-DCMAKE_INSTALL_PREFIX='] +
+                 self.options.configflags, env=env, cwd=cmake_builddir)
 
-        self.run(['cmake', sourcedir, '-DCMAKE_INSTALL_PREFIX='] +
-                 self.options.configflags, env=env)
-
-        self.make(env=env)
+        self.make(env=env, cwd=cmake_builddir)
 
     def _build_environment(self):
         env = os.environ.copy()
--- a/snapcraft/plugins/cmake.py
+++ b/snapcraft/plugins/cmake.py
@@ -31,7 +31,6 @@ Additionally, this plugin uses the following plugin-specific keywords:
 """
 
 import os
-import shutil
 
 import snapcraft.plugins.make
 
--- a/integration_tests/__init__.py
+++ b/integration_tests/__init__.py
@@ -103,6 +103,7 @@ class TestCase(testtools.TestCase):
         # they might be created after calling the snapcraft command on the
         # project dir.
         self.parts_dir = 'parts'
+        self.build_dir = 'build'
         self.stage_dir = 'stage'
         self.prime_dir = 'prime'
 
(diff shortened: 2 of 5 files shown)


snapcraft/plugins/cmake.py
```diff
@@ -64,22 +64,15 @@ def __init__(self, name, options, project):
         self.build_packages.append('cmake')
 
     def build(self):
-        if os.path.exists(self.builddir):
-            shutil.rmtree(self.builddir)
-        os.mkdir(self.builddir)
```
elopio: I have no clue why we were deleting the builddir. A step before this had just populated it with a copy of the sources.

snapcraft/plugins/cmake.py
```diff
@@ -64,22 +64,15 @@ def __init__(self, name, options, project):
...
-
-        source_subdir = getattr(self.options, 'source_subdir', None)
-        if source_subdir:
-            sourcedir = os.path.join(self.sourcedir, source_subdir)
```
elopio: This is not needed, because the base class sets builddir to self.build_basedir + self.source_subdir. It is safe to just use self.builddir.

snapcraft/plugins/cmake.py
kyrofa: Because the pluginhandler copies the source to the builddir before handing control off to the plugin. Since cmake knows it's about to do an out-of-source build using the sourcedir, it empties the builddir so it has a clean area in which to build.

kyrofa:
One of the reasons we copy the source to the builddir in the first place is for the many build systems we support that don't provide a method for building out-of-source. We essentially are "faking" an out-of-source build for those systems, leaving the initial code untouched in `<partdir>/src`. Cool. Except that `cmake` actually _does_ support a real out-of-source build, which is what we're utilizing here. There's no need to have multiple copies of the source here, because cmake supports using the source we already have an building into our builddir. It's a perfect fit for what we're doing. This PR stops taking advantage of that, which tickles my OCD in a negative way :stuck_out_tongue: .

It's my impression from the bug that the whole motivation behind this is supporting patching the source in `prepare`, which in my opinion is wrong. Taking in the above paragraph, you see that patching in `prepare` really works as a _side effect_ rather than by design. That's proven by the fact that not all plugins work that way. If we really want to support patching, we should add scriptlets surrounding the `pull` step. Then the patching can be done in the scriptlet that runs after `pull`, and modify the source in `src`, not in the builddir. This seems a much more suitable solution that modifying our plugins so that the "side effect" being utilized by patching in `prepare` applies across the board.

elopio:
Thanks for the review @kyrofa. I see your point, and indeed, a post-pull step would solve all the two cases I have.

However, my OCD is tickled the other way. Every plugin starts by copying the src into build. Then I would expect cmake to do the same, and I find it pretty weird that the first step is to delete contents that were just copied. To me, because I'm used to expect a copy of the src in build, and that every command I run in prepare, build and install will use build as cwd, the cmake plugin behaviour is unexpected.

kyrofa:
> However, my OCD is tickled the other way. Every plugin starts by copying the src into build. Then I would expect cmake to do the same, and I find it pretty weird that the first step is to delete contents that were just copied.

Well, remember that the plugin has nothing to do with copying src into build. That's done in the pluginhandler, before the plugin's `build` has been called. That's why cmake then has to delete it. I agree, it's not beautiful.

> To me, because I'm used to expect a copy of the src in build, and that every command I run in prepare, build and install will use build as cwd, the cmake plugin behaviour is unexpected.

Well, every command you run in `prepare` `build` and `install` DOES use build as cwd, right? It's just that build doesn't always contain the src. Because prepare/build/install should be concerned with building, not messing with the src.

I really think having scriptlets for messing with the src is the solution here, but if we can't get buy-off on that then I'm okay with this.

elopio:
@kyrofa what do you think of the case in mumble, where we found a problem originally?

Take a look here: https://github.com/mumble-voip/mumble/blob/19ecc86f17e8436d806b43993c089046b4cc0bbd/src/mumble_proto/mumble_proto.pro#L19

In there, protoc is called with `-I..`. That means that it will search for the protobuf file in the parent directory. Let me try to explain what happens, because it's weird and I'm trying to understand the details.

If you define `project_files`, snapcraft will take them from the source dir but execute with build dir as cwd (https://github.com/snapcore/snapcraft/blob/master/snapcraft/plugins/qmake.py#L112)
So, the assumption that the protobuf file is in `..` is broken. The file is actually in `../../src`.

I think it might be a bug in mumble to locate the protobuf in that way, maybe there's a smarter way to do it. But it doesn't seem completely unreasonable, and it might be a common thing.

kyrofa:
> @kyrofa what do you think of the case in mumble, where we found a problem originally?

I think that an example of a .pro file that's written incorrectly. It seems that it prevents any sort of out-of-source build. There are other ways to use protobuf, like [this](https://vilimpoc.org/blog/2013/06/09/using-google-protocol-buffers-with-qmake/) or using paths relative to PWD, etc.

elopio:
I'll close this one for now. We have the bug open in case somebody else is affected, and it links to this PR, so it won't be lost.

PR #1415, opened 2017-07-19
tests: run the tests in travis using LXD containers

elopio:


tools/travis/run_docker_container.sh
```diff
@@ -1,55 +0,0 @@
...
-  --env LC_ALL=C.UTF-8 \
-  --volume "$project_path":"$project_path" \
-  --workdir "$project_path" \
-  --tty \
-  --detach \
-  ubuntu:xenial
-
-echo "Set up the main archive."
-docker exec -i "$name" sed -i s/archive.ubuntu.com/us.archive.ubuntu.com/g /etc/apt/sources.list
-docker exec -i "$name" apt update
```
kalikiana: Would it be worth keeping the docker workflow around to catch docker-related regressions?  Maybe run less frequently. Just a thought.

elopio: yup, I didn't want to remove it yet. That was a mistake. Some of the jobs are still using docker, I'll update them later.
But when we are done migrating to lxc, I plan to remove it.
I think we should add a new suite instead, for testing the docker snapcore/snapcraft container. It would build snaps though, not run unit or integration tests.

tools/travis/run_lxc_container.sh
```diff
@@ -0,0 +1,56 @@
...
+name="$1"
+
+lxc=/snap/bin/lxc
+
+echo "Start the LXC container."
+sudo $lxc launch --ephemeral ubuntu:xenial "$name"
+# This is likely needed to wait for systemd in the container to start and get
+# an IP, configure DNS. First boot is always a bit slow because cloud-init
+# needs to run too.
+sleep 10
```
elopio: I tried removing the sleep and leaving just the ping, but it wasn't enough. apt update still failed.
I'm trying now with this loop plus a 5s sleep.

PR #1416, opened 2017-07-19
core: a clean command should clean.

sergiusens:
When running clean it sometimes does not clean to avoid cleaning some things
you might not want to clean but that has proven to be a cumbersome user
facing feature so now it would just clean everything in the path.

### DIFF
--- a/snapcraft/internal/lifecycle.py
+++ b/snapcraft/internal/lifecycle.py
@@ -465,7 +465,7 @@ def _verify_dependents_will_be_cleaned(part_name, clean_part_names, step,
                 additional_dependents.append(part_name)
 
                 logger.warning(
-                    'Requested clean of {!r} which require also cleaning '
+                    'Requested clean of {!r} which requires also cleaning '
                     'the part{} {}'.format(part_name,
                                            '' if len(dependents) == 1 else 's',
                                            humanized_parts))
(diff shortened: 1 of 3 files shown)


kyrofa:
This is different from what we discussed at the sprint:

```
### A better clean story

- Make it interactive with reasons.
- Add a `--yes`.
- Add `snapcraft config` for automation.
```

snapcraft/internal/lifecycle.py
```diff
@@ -454,27 +454,30 @@ def _verify_dependents_will_be_cleaned(part_name, clean_part_names, step,
...
-                raise errors.SnapcraftEnvironmentError(
-                    'Requested clean of {!r} but {} depend{} upon it. Please '
-                    "add each to the clean command if that's what you "
-                    'intended.'.format(part_name, humanized_parts,
-                                       's' if len(dependents) == 1 else ''))
+                logger.warning(
+                    'Requested clean of {!r} which require also cleaning '
+                    'the part{} {}'.format(part_name,
+                                           '' if len(dependents) == 1 else 's',
+                                           humanized_parts))
```
kalikiana: So the exception becomes a warning? Does it mean it'll go ahead and clean parts or does it still skip?

sergiusens:
Basically from what I read on the bug again and the feedback I got at the sprint here.

sergiusens:
Let's see how it does in the bug reporing camp @elopio 
The more I think of this, the more I think that this is the default everywhere, even on an android build when cleaning a target and those builds are huge.

PR #1417, opened 2017-07-20
kbuild plugin: move over the cross-compiling logic from the kernel plugin

kalikiana:
Move the cross-compiling bits from the kernel plugin to the kbuild plugin, taking advantage of the fact that kbuild's being used as a superclass.

Note: The test doesn't include x86 32 or 64bit because snapcraft doesn't support it (see _options.py).


snapcraft/plugins/kbuild.py
```diff
@@ -122,6 +122,21 @@ def __init__(self, name, options, project):
         if logger.isEnabledFor(logging.DEBUG):
             self.make_cmd.append('V=1')
 
+    def enable_cross_compilation(self):
+        logger.info('Cross compiling kernel target {!r}'.format(
```
kyrofa: Not necessarily a kernel target anymore, right?

kalikiana: It's still the kernel architecture...
Alternatively we can drop it entirely and just rely on the regular message that refers to the debian target architecture.

kubiko:
this might be to much ask for this pull request, but issues I hit when building u-boot with kbuild plugin:
a) some way to override install step, as calling make install on uboot build fails miserably
b) can't say if this is something common for u-boot on arm64, but I had to use arm64 toolchain (cross or native), but override ARCH to "arm"
So to fully cover dragon board use case we would also need option to override ARCH value if not matching to target arch.
For reference this was custom plugin I used:
https://github.com/kubiko/dragonboard-gadget/blob/master/snap/plugins/x-uboot.py

snapcraft/plugins/kbuild.py
sergiusens: Yeah, either drop or move just this bit to the `kernel plugin` so that in that plugin we see

```
def enable_cross_compilation(self):
    logger.info('Cross compiling kernel target {!r}'.format(
        self.project.kernel_arch))
    super().enable_cross_compilation()
```

And no message in this plugin. This is just here as the kernel makefiles take a long time to load and as a user it was satisfactory to see that you were doing the right thing.

PR #1418, opened 2017-07-25
nodejs plugin: copy the content of out of tree symlinks

sergiusens:
LP: [#1702661](https://bugs.launchpad.net/snapcraft/+bug/1702661)

### DIFF
--- a/snapcraft/plugins/nodejs.py
+++ b/snapcraft/plugins/nodejs.py
@@ -220,7 +220,7 @@ def get_nodejs_release(node_engine, arch):
 
 def _fix_symlinks(modulesdir):
     """Fix symlinks living in the part's build so that they live in install.
-    
+
     When running newer versions of npm, symlinks to the local tree are
     created from the part's installdir to the root of the builddir of the
     part (this only affects some build configurations in some projects)
--- a/snapcraft/plugins/nodejs.py
+++ b/snapcraft/plugins/nodejs.py
@@ -149,9 +149,10 @@ class NodePlugin(snapcraft.BasePlugin):
         super().build()
         if self.options.node_package_manager == 'npm':
             self._npm_install(rootdir=self.builddir)
-            # Fix the symlink to the build directory LP: #1702661
+            # Copy the content of the symlink to the build directory
+            # LP: #1702661
             modules_dir = os.path.join(self.installdir, 'lib', 'node_modules')
-            _fix_symlinks(modules_dir)
+            _copy_symlinked_content(modules_dir)
         else:
             self._yarn_install(rootdir=self.builddir)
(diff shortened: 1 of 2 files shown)


sergiusens:
This specifically affects the heroku build reported in the bug where using a combination of local modules and remote ones causes the link farm issue. This should be added to the _archive rebuild_ job

snapcraft/plugins/nodejs.py
```diff
@@ -212,3 +216,26 @@ def _get_nodejs_base(node_engine, machine):
...
+    """Fix symlinks living in the part's build so that they live in install.
+
+    When running newer versions of npm, symlinks to the local tree are
+    created from the part's installdir to the root of the builddir of the
+    part (this only affects some build configurations in some projects)
+    which is valid when running from the context of the part but invalid
+    as soon as the artifacts migrate across the stages,
+    i.e.; stage and prime.
+
+    If modulesdir does not exit we simply return.
```
kyrofa: s/exit/exist/

snapcraft/plugins/nodejs.py
```diff
@@ -212,3 +216,26 @@ def _get_nodejs_base(node_engine, machine):
...
+    """
+    if not os.path.exists(modulesdir):
+        return
+    modules = [os.path.join(modulesdir, d) for d in os.listdir(modulesdir)]
+    symlinks = [l for l in modules if os.path.islink(l)]
+    for link_path in symlinks:
+        print('Fixing', link_path)
```
kyrofa: Perhaps this should be `echo.debug` or something similar?

snapcraft/plugins/nodejs.py
```diff
@@ -212,3 +216,26 @@ def _get_nodejs_base(node_engine, machine):
...
+        link_target = os.path.realpath(link_path)
+        os.unlink(link_path)
+        link_or_copy_tree(link_target, link_path)
```
kyrofa: The title of this PR as well as the function name makes me think you're simply rewriting symlinks ("fixing" them) here, but you're not, you're following it and copying the target. It would be clearer to rename things.

sergiusens: suggestions?

snapcraft/plugins/nodejs.py
sergiusens: Ah, this was just a testing thing of some wierdness when I was trying to get it working as a local project plugin

snapcraft/plugins/nodejs.py
kyrofa: Perhaps `_follow_symlinks(modulesdir)`?

PR #1419, opened 2017-07-26
meta: add `base` as a type and top level property

mvo5:
Thanks for helping us make a better Snapcraft!

LP: #1706564


sergiusens:
Thanks for the contribution. The tests failed because your commit in the email has not signed the CLA, you can either sign it with your @ubuntu.com account or use your @canonical.com email to author the commit.

mvo5:
The contributor agreement thing is hopefully sorted now.

mvo5:
Thanks for the feedback and sorry for the delay, I was on vacation. 

I added the "base" to the support types now and also added tests to snaps_tests as suggested. This will validate that the snaps can be build and installed. If you need more tests, please let me know.

PR #1420, opened 2017-07-26
add new "no-wrapper" property to apps

mvo5:
A RFC branch to fix https://bugs.launchpad.net/snapcraft/+bug/1706575

This allows to specify that no shell wrapper for apps should be generated. This is useful for e.g. statically linked binaries that don't need this anyway.

LP: 1706575


schema/snapcraft.yaml
```diff
@@ -190,6 +190,10 @@ properties:
                 - type: string
                   minLength: 1
                 - type: number
+          no-wrapper:
```
zyga: Suggestion: make this "wrappers" enum that has possible values "legacy" and "none" as we may be able to provide some kind of useful wrappers later on.

sergiusens: Agreed, this should be an array of unique elements like `build-attributes` in parts.

integration_tests/snaps/assemble/snapcraft.yaml
```diff
@@ -10,6 +10,8 @@ apps:
     command: binary1
   binary2:
     command: subdir/binary2
+  binary-no-wrapper:
+    command: subdir/binary3
```
elopio: This is missing:
`wrapper: none`

integration_tests/test_snap.py
```diff
@@ -73,6 +73,10 @@ def test_snap(self):
             os.path.join(self.prime_dir, 'bin', 'not-wrapped.wrapper'),
             Not(FileExists()))
 
+        self.assertThat(
+            os.path.join(self.prime_dir, 'bin', 'command-binary3.wrapper'),
```
elopio: The name of the file should be: `command-binary-no-wrapper.wrapper`.

mvo5:
Thanks for the feedback and sorry for my slow reply. I addressed the review feedback now. Please let me know if I can help in any other way.

schema/snapcraft.yaml
```diff
@@ -190,6 +190,13 @@ properties:
                 - type: string
                   minLength: 1
                 - type: number
+          wrapper:
```
zyga: Nice :-)

snapcraft/internal/meta.py
```diff
@@ -371,7 +371,11 @@ def _wrap_apps(self, apps):
             if os.path.splitext(f)[1] == '.desktop':
                 os.remove(os.path.join(gui_dir, f))
         for app in apps:
-            self._wrap_app(app, apps[app])
+            wrapper = "legacy"
+            if "wrapper" in apps[app]:
+                wrapper = apps[app]["wrapper"]
```
zyga: `wrapper = apps[app].get("wrapper", legacy)` is shorter

Not sure if snapcraft uses anything that can derive the default from the schema.

kalikiana: Yes. `default: legacy` does exactly that, so no need to do it in code.

sergiusens:
Hi, thanks for the changes, I was thinking of something more like:

```
apps:
    no-wrapper-app:
        command: app
        app-constraints: [no-wrapper]
    wrapped-app
        command: app
```

wdyt?

sergiusens:
Now that I think of it, it `no-wrapper` might even be `no-environment` as we already have plans to get rid of wrappers and make use of `env` for `apps` a lot more.

mvo5:
I updated the PR now to follow the suggestion of Gustavo in https://forum.snapcraft.io/t/telling-snapcraft-to-skip-generating-wrapper-scripts/1635/3 - i.e. it is using: `adapter: {"", "none"}` now.

mvo5:
How are things looking? I updated the code based on the feedback that I got from @niemeyer. Please let me know if there is anything else I can do to help this PR along, I need it for https://github.com/snapcore/snapd/pull/3625 so it woudl be great if it could make it to the next snapcraft release.

kyrofa:
Ping on this one.

mvo5:
Thanks, updated and removed the "" from the enum.

PR #1425, opened 2017-07-27
options: properly handle missing compiler prefix

kyrofa:
Otherwise the attempted cross-compile fails with: 'ProjectOptions' object has no attribute '_ProjectOptions__machine_target'.


kyrofa:
> [...] you might want to take the opportunity to change this exception

Done. Took the opportunity to fix the others in options as well.

PR #1427, opened 2017-07-28
Add cx_Freeze options targeting bin/snapcraft

alextnewman:
This adds minimal win32 cx_Freeze support and refactors setup.py to account for incompatibilities in `setup()` while eliminating redundant packaging data. Ideally, this should not impact the current setup.py behavior beyond using cx_Freeze. 

This work is not exhaustively tested but should produce a working snapcraft.exe. I also tested `bdist_msi` and found that this can be deployed with an installer.

At this time, I'm targeting `bin/snapcraft` despite the fact that this is supposedly deprecated because it is far easier to pass to cx_Freeze as an entry point. I'd welcome a more future-proof approach.


alextnewman:
I've actually introduced some github metadata problems into this branch/PR that I think would be simpler to address on a clean PR. I'm going to close this and re-attempt with a clean branch as I believe the technical issues are solved.

PR #1428, opened 2017-07-28
core: reexec as root for `os` snaps if necessary

sergiusens:

### DIFF
--- a/snapcraft/cli/lifecycle.py
+++ b/snapcraft/cli/lifecycle.py
@@ -13,6 +13,7 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+import subprocess
 import sys
 
 import click
@@ -23,6 +24,14 @@ from . import echo
 from . import env
 
 
+def _reexec_as_root():
+    click.echo('Reexecuting the build as root')
+    try:
+        subprocess.check_call(['sudo'] + sys.argv)
+    except subprocess.CalledProcessError:
+        sys.exit(1)
+
+
 def _execute(command, parts, **kwargs):
     project_options = get_project_options(**kwargs)
 
@@ -31,6 +40,9 @@ def _execute(command, parts, **kwargs):
     else:
         try:
             lifecycle.execute(command, project_options, parts)
+        except errors.SnapcraftUserPermissionError as user_error:
+            if user_error.user == 'root':
+                _reexec_as_root()
         except Exception as e:
             echo.error(e)
             sys.exit(1)
@@ -140,10 +152,13 @@ def snap(directory, output, **kwargs):
         try:
             snap_name = lifecycle.snap(
                 project_options, directory=directory, output=output)
+            echo.info('Snapped {}'.format(snap_name))
+        except errors.SnapcraftUserPermissionError as user_error:
+            if user_error.user == 'root':
+                _reexec_as_root()
         except Exception as e:
             echo.error(e)
             sys.exit(1)
-        echo.info('Snapped {}'.format(snap_name))
 
 
 @lifecyclecli.command()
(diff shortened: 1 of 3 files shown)


snapcraft/cli/lifecycle.py
```diff
@@ -13,6 +13,7 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+import subprocess
```
elopio: this is missing an empty line above.

snapcraft/cli/lifecycle.py
```diff
@@ -23,6 +24,14 @@
 from . import env
 
 
+def _reexec_as_root():
+    click.echo('Reexecuting the build as root')
+    try:
+        subprocess.check_call(['sudo'] + sys.argv)
+    except subprocess.CalledProcessError:
+        sys.exit(1)
```
elopio: Why are you catching this exception here, instead of letting the caller handle it?
It already has a sys.exit(1), plus echo of the error.

snapcraft/cli/lifecycle.py
```diff
@@ -31,6 +40,9 @@ def _execute(command, parts, **kwargs):
     else:
         try:
             lifecycle.execute(command, project_options, parts)
+        except errors.SnapcraftUserPermissionError as user_error:
+            if user_error.user == 'root':
+                _reexec_as_root()
```
kyrofa: It's only build that needs this, right? Doesn't this effectively make _every_ lifecycle step run as root? Is that necessary, or can we limit it to build? I suppose everything would be owned by root at that point...

snapcraft/internal/errors.py
```diff
@@ -222,6 +222,11 @@ def __init__(self, message, snapcraft_yaml='snapcraft.yaml'):
         super().__init__(message=message, snapcraft_yaml=snapcraft_yaml)
 
 
+class SnapcraftUserPermissionError(SnapcraftError):
+
+    fmt = '{user} privilidges are required to continue'
```
kyrofa: s/privilidges/privileges/

snapcraft/internal/lifecycle.py
```diff
@@ -595,3 +602,11 @@ def clean(project_options, parts, step=None):
     _clean_parts(parts, step, config, staged_state, primed_state)
 
     _cleanup_common_directories(config, project_options)
+
+
+_TYPE_REQUIRES_ROOT = ['os']
+
+
+def _ensure_user(snap_type):
+    if (snap_type in _TYPE_REQUIRES_ROOT and os.getuid() != 0):
```
kyrofa: `os.geteuid()` might be a better fit here.

snapcraft/cli/lifecycle.py
sergiusens: prime needs it too

snapcraft/cli/lifecycle.py
sergiusens: we have a mix in our code base for this

snapcraft/cli/lifecycle.py
ogra1: build actually has the sudo calls in the build scripts of core itself now via https://github.com/snapcore/core/pull/51 ... what failed hard was prime even after this change ...

kyrofa: @ogra1 so does build not actually need this anymore? I'm leaning more and more toward having an environment variable or build attribute that says prime needs to be done with sudo rather than automatically trying it.

ogra1: exactly, build should be fine with the above "hack" to run all the live-build steps with sudo, when we tried it with the already deployed change in LP (before it got rolled back) only prime failed (note though that mksquashfs uses special options for os snaps and since we failed in prime i dont know if there would have been any subsequent errors in the "snap" step or if the filesystem permissions in teh resulting snap would have been broken)

niemeyer:
Please see the discussion in https://forum.snapcraft.io/t/run-snapcraft-in-container-as-a-user/1438

kyrofa:
This has gotten pretty stale, especially after the CLI error handling changes. I'll close for now.

PR #1429, opened 2017-07-28
cli: properly handle exceptions in lifecycle

kyrofa:
The exception handling proposed here:

When debug=False:
- If exception is a `SnapcraftError`, just display a nice error and exit 1.
- If exception is NOT a `SnapcraftError`, raise so traceback is shown.

When debug=True:
- Raise any exception so traceback is shown.

**Note:** This PR is limited to the lifecycle CLI, but the same logic should probably be applied to the store CLI in a separate PR.


kyrofa:
> we need to do a sweep of exceptions.

Nothing would please my OCD more!

snapcraft/cli/_errors.py
```diff
@@ -0,0 +1,61 @@
...
+@contextlib.contextmanager
+def exception_catcher(debug):
+    try:
+        yield
+    except snapcraft.internal.errors.SnapcraftError as e:
+        if debug:
+            raise
+        else:
+            echo.error(str(e))
+            sys.exit(1)
```
kalikiana: How about taking the error code from the exception? Default would be 1. That'd make it really easy to let consumers of the CLI distinguish errors.

kyrofa: Excellent idea. However, since the traceback ends up exiting 1, I defaulted to 2. That way we can at least distinguish between non-snapcraft errors and snapcraft errors. We can further flesh this out by reorganizing our errors a bit better.

sergiusens:
Closing in favour of #1436

PR #1430, opened 2017-07-29
tests: enable the snaps installation in armhf

elopio:


kyrofa:
This has gotten a bit stale, so I'll close for now.

PR #1431, opened 2017-07-30
tests: do not check the opencv command if it wasn't installed

elopio:


elopio:
Yes, this is a sad reminder of past times. But now that we can (in theory) install snaps in containers, we can rethink our snaps suite.
However, the installation in armhf is still failing. I need to ask about this.

PR #1432, opened 2017-08-01
kbuild plugin: support Makefile without install target

kalikiana:
Projects using **KBuild** often have no install target in the Makefile.

### DIFF
--- a/integration_tests/plugins/test_kbuild_plugin.py
+++ b/integration_tests/plugins/test_kbuild_plugin.py
@@ -23,9 +23,6 @@ import integration_tests
 
 class KBuildPluginTestCase(integration_tests.TestCase):
 
-    def setUp(self):
-        super().setUp()
-
     def test_stage(self):
         self.run_snapcraft('stage', 'kbuild-hello')
 
--- a/integration_tests/plugins/test_kbuild_plugin.py
+++ b/integration_tests/plugins/test_kbuild_plugin.py
@@ -1,0 +1,34 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2016-2017 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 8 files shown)


integration_tests/plugins/test_kbuild_plugin.py
```diff
@@ -0,0 +1,34 @@
...
+
+from testtools.matchers import FileExists
+
+import integration_tests
+
+
+class KBuildPluginTestCase(integration_tests.TestCase):
+
+    def setUp(self):
+        super().setUp()
```
sparkiegeek: Redundant method is redundant, KIWF :fire:

snapcraft/plugins/kbuild.py
```diff
@@ -43,6 +43,12 @@
       definitions.  If you don't want default for one or more implicit configs
       coming out of these, just add them to this list as well.
 
+    - build-attributes
+      (list of strings)
+      list of boolean build options:
```
sergiusens: we can improve this text in future PRs as this is a core attribute that the plugin understands instead of being an attribute of the plugin itself.

snapcraft/plugins/kbuild.py
```diff
@@ -103,13 +109,24 @@ def schema(cls):
             'default': None,
         }
 
+        schema['properties']['build-attributes'] = {
```
sergiusens: this is already in the main schema and not really needed here (in `schema/snapcraft.yaml`)

schema/snapcraft.yaml
```diff
@@ -288,6 +288,7 @@ properties:
             uniqueItems: true
             items:
               type: string
+              pattern: "(no-system-libraries|no-install)"
```
sergiusens: As we add values here, it might be better to use an `enum` like on https://github.com/snapcore/snapcraft/pull/1432/commits/6394486535ef6ed6c7ba0d086bfad056870a77f5#diff-fa3e006298d99e14b7fb303c7883c895R252

PR #1433, opened 2017-08-01
core: cache FileBase entries when a checksum is provided

sergiusens:


snapcraft/file_utils.py
```diff
@@ -239,3 +239,18 @@ def calculate_sha3_384(path):
                 break
             hasher.update(buf)
         return hasher.hexdigest()
+
+
+def calculate_hash(path, *, algorithm):
+    """Calculate the hash for path with algorithm."""
+    blocksize = 2**20
+    # This will raise an AttributeError if algorithm is unsupported
+    hasher = getattr(hashlib, algorithm)()
```
elopio: nit: the potential error should go first. That way we avoid the wasted allocation of blocksize in case of error.

snapcraft/internal/cache/_file.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def __init__(self):
+        """Create a FileCache."""
+        super().__init__()
+        self.file_cache = os.path.join(self.cache_root, 'files')
+
+    def cache(self, *, filename, algorithm, hash):
+        """Cache a file revision with hash in XDG cache, unless it already exists.
+        :param str filename: path to the file to cache.
+        :param str algorithm: algorithm used to calculate has as understood
```
elopio: typo: has/hash

snapcraft/internal/cache/_file.py
```diff
@@ -0,0 +1,75 @@
...
+                shutil.copyfile(filename, cached_file_path)
+        except OSError:
+            logger.warning(
+                'Unable to cache file {}.'.format(cached_file_path))
+        return cached_file_path
+
+    def get(self, *, algorithm, hash):
+        """Get the filepath which matches the hash calculated with algorithm.
+
+        :param str algorithm: algorithm used to calculate has as understood
```
elopio: typo: has/hash

snapcraft/internal/sources/errors.py
```diff
@@ -31,8 +31,8 @@ def __init__(self, message):
 
 class DigestDoesNotMatchError(errors.SnapcraftError):
 
-    fmt = 'Expected the digest for source to be {expected}, '
-    'but it was {calculated}'
+    fmt = ('Expected the digest for source to be {expected}, '
+           'but it was {calculated}')
```
elopio: ^_^

sergiusens: yeah, this was driving me nuts

snapcraft/file_utils.py
sergiusens: I just brought this in from the implementation in `sources`, oh, I see what you mean.

sergiusens:
Thanks for the review, I want to introduce the `clean-cache` subcommand in order for this to be a thing.

snapcraft/internal/cache/_file.py
```diff
@@ -0,0 +1,75 @@
...
+        try:
+            if not os.path.isfile(cached_file_path):
+                # this must not be hard-linked, as rebuilding a snap
+                # with changes should invalidate the cache, hence avoids
+                # using fileutils.link_or_copy.
+                shutil.copyfile(filename, cached_file_path)
+        except OSError:
+            logger.warning(
+                'Unable to cache file {}.'.format(cached_file_path))
+        return cached_file_path
```
kalikiana: What good is returning the cached_file_path if it failed? Due to the try/ except it seems like it may not exist and lead to surprising failure in the consumer of the method.

snapcraft/internal/sources/_base.py
```diff
@@ -45,12 +46,43 @@ def __init__(self, source, source_dir, source_tag=None, source_commit=None,
...
+        # First check if we already have the source file cached.
+        if self.source_checksum:
+            algorithm, hash = split_checksum(self.source_checksum)
+            cache_file = file_cache.get(algorithm=algorithm, hash=hash)
+            if cache_file:
+                source_file = os.path.join(self.source_dir,
+                                           os.path.basename(cache_file))
+                # We make this copy as the provisioning logic can delete
+                # this file and we don't want that.
+                shutil.copy2(cache_file, source_file)
```
kalikiana: I feel this might be cleaner if file_cache.get could just take source_checksum even if it's None. Though you can probably argue that more explicit is clearer, so feel free to leave it.

snapcraft/internal/sources/_base.py
```diff
@@ -45,12 +46,43 @@ def __init__(self, source, source_dir, source_tag=None, source_commit=None,
...
+            shutil.copy2(self.source, source_file)
+
+        # Before provisioning we verify the file if source_checksum is defined
+        # and we cache the file for future reuse.
+        if self.source_checksum:
+            algorithm, digest = verify_checksum(
+                self.source_checksum, source_file)
+            file_cache.cache(filename=source_file,
+                             algorithm=algorithm,
+                             hash=hash)
```
kalikiana: Same as above, if cache took source_check_sum this could be handled in a single place.

snapcraft/tests/cache/test_file.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def test_cache_not_possible(self):
+        with open('hash_file', 'w') as f:
+            f.write('random stub data')
+
+        bad_calculated_hash = '1'
+        file = self.file_cache.cache(filename='hash_file',
+                                     algorithm=self.algo,
+                                     hash=bad_calculated_hash)
+        self.assertThat(file, Is(None))
```
kalikiana: I don't see the case where caching fials but a filename is returned, cf. my first comment.

sergiusens: Good catch on this and issue 1, will fix this.

snapcraft/internal/sources/_base.py
sergiusens: I don't understand this comment in the context of `shutil.copy2`, can you expand?

snapcraft/internal/sources/_base.py
sergiusens: ah... I want to keep hash and algorithm separate from how it is represented in a source file. Maybe the better way would be to split at `__init__`.
This is because this is not the only place we check for hashes, what is sadder is that the implementation that does the checksum here does it wrong as it does not chunk it correctly :-/

sergiusens: the new on in `file_utils` does though

snapcraft/internal/sources/_base.py
kalikiana: Note that I'm referring to file_cache.get here. I apparently find to pick the best line for context.

snapcraft/internal/cache/_file.py
```diff
@@ -0,0 +1,76 @@
...
+                           'provided'.format(filename))
+            return None
+        cached_file_path = os.path.join(self.file_cache, algorithm, hash)
+        os.makedirs(os.path.dirname(cached_file_path), exist_ok=True)
+        try:
+            if not os.path.isfile(cached_file_path):
+                # this must not be hard-linked, as rebuilding a snap
+                # with changes should invalidate the cache, hence avoids
+                # using fileutils.link_or_copy.
+                shutil.copyfile(filename, cached_file_path)
```
elopio: wait, shouldn't this get an `else return None`?

If the path is not a file, it won't be copied but you still return cached_file_path.

elopio: Ah, nevermind. Too early for me. If the file exists, it's already in the cache and there's nothing to copy. I think it would be clearer like this:
```
cached_file_path = os.path.join(self.file_cache, algorithm, hash)
if not os.path.isfile(cached_file_path):
    os.makedirs(os.path.dirname(cached_file_path), exist_ok=True)
    try:
       shutil.copyfile(filename, cached_file_path)
    except ...
```
The try would be smaller too, which is always good.

sergiusens:
this is failing on
```
Starting new HTTP connection (1): pyyaml.org

('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))

}}}
```
Going to ignore it.

PR #1436, opened 2017-08-04
cli: properly handle exceptions

kyrofa:
This PR is an alternative to #1429. Instead of using click-specific decorators for every command, it simply installs a `sys.excepthandler` to deal properly with exceptions.

Advantages over #1429:
- Simpler
- Covers the store CLI with no additional work
- Covers future CLI additions automatically, nothing special needed

Disadvantages compared to #1429:
- Can no longer unit test CLI exit codes since the CLI is run within the same python process, the test catches the exception, and `excepthook` is never called. Integrations tests are needed instead.
- Global behavior can cause unexpected things to happen (i.e. the simple fact that some exceptions are masked by a global handler may be unexpected a year down the road when we add a new command)

### DIFF
--- a/snapcraft/_store.py
+++ b/snapcraft/_store.py
@@ -35,10 +35,6 @@ from tabulate import tabulate
 
 from snapcraft.file_utils import calculate_sha3_384
 from snapcraft import storeapi
-from snapcraft.storeapi.errors import (
-    StoreDeltaApplicationError,
-    StoreAssertionError,
-)
 from snapcraft.internal import (
     cache,
     deltas,
@@ -236,9 +232,7 @@ def _export_key(name, account_id):
 
 def list_keys():
     if not repo.Repo.is_package_installed('snapd'):
-        raise EnvironmentError(
-            'The snapd package is not installed. In order to use `list-keys`, '
-            'you must run `apt install snapd`.')
+        raise storeapi.errors.MissingSnapdError('list-keys')
     keys = list(_get_usable_keys())
     store = storeapi.StoreClient()
     with _requires_login():
@@ -258,9 +252,7 @@ def list_keys():
 
 def create_key(name):
     if not repo.Repo.is_package_installed('snapd'):
-        raise EnvironmentError(
-            'The snapd package is not installed. In order to use '
-            '`create-key`, you must run `apt install snapd`.')
+        raise storeapi.errors.MissingSnapdError('create-key')
     if not name:
         name = 'default'
     keys = list(_get_usable_keys(name=name))
@@ -268,7 +260,7 @@ def create_key(name):
         # `snap create-key` would eventually fail, but we can save the user
         # some time in this obvious error case by not bothering to talk to
         # the store first.
-        raise RuntimeError('You already have a key named "{}".'.format(name))
+        raise storeapi.errors.KeyAlreadyRegisteredError(name)
     store = storeapi.StoreClient()
     try:
         account_info = store.get_account_information()
@@ -281,8 +273,7 @@ def create_key(name):
         # yet.
         enabled_names = set()
     if name in enabled_names:
-        raise RuntimeError(
-            'You have already registered a key named "{}".'.format(name))
+        raise storeapi.errors.KeyAlreadyRegisteredError(name)
     subprocess.check_call(['snap', 'create-key', name])
 
 
@@ -290,25 +281,19 @@ def _maybe_prompt_for_key(name):
     keys = list(_get_usable_keys(name=name))
     if not keys:
         if name is not None:
-            raise RuntimeError(
-                'You have no usable key named "{}".\nSee the keys available '
-                'in your system with `snapcraft keys`.'.format(name))
+            raise storeapi.errors.NoSuchKeyError(name)
         else:
-            raise RuntimeError(
-                'You have no usable keys.\nPlease create at least one key '
-                'with `snapcraft create-key` for use with snap.')
+            raise storeapi.errors.NoKeysError
     return _select_key(keys)
 
 
 def register_key(name):
     if not repo.Repo.is_package_installed('snapd'):
-        raise EnvironmentError(
-            'The snapd package is not installed. In order to use '
-            '`register-key`, you must run `apt install snapd`.')
+        raise storeapi.errors.MissingSnapdError('register-key')
     key = _maybe_prompt_for_key(name)
     store = storeapi.StoreClient()
     if not _login(store, acls=['modify_account_key'], save=False):
-        raise RuntimeError('Cannot continue without logging in successfully.')
+        raise storeapi.errors.LoginRequiredError()
     logger.info('Registering key ...')
     account_info = store.get_account_information()
     account_key_request = _export_key(key['name'], account_info['account_id'])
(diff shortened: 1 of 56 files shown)


kyrofa:
The tests will not pass here. There's some more work to do to move exit-code-testing unit tests into integration, but I want to make sure we like this direction before I continue.

kyrofa:
@sergiusens says he prefers #1429. @elopio, @kalikiana thoughts? I've tried to outline the pros/cons here. Either way makes me happy, which makes you three a quorum.

elopio:
I don't have a preference. Both approaches look perfectly valid to me.

Just one detail here: don't turn all the unit tests for exit codes into integration tests. We should have one or two integration tests for exit codes and traces, unit tests (probably patching `sys.exit`) for the exception handler, and then full coverage of unit tests for all the commands starting one level down. So IMO, most unit tests shouldn't check exit codes, just that the right exception was raised after calling a function.

kyrofa:
Haha, since @elopio lamely backed out of having an opinion, I will state my slight preference for this PR. I think it's more likely that we add new commands without using the right decorators in the other PR (thus resulting in some commands that oddly don't handle debugging exceptions appropriately) than this PR's disadvantages biting us.

kalikiana:
Compared to #1429 I'm finding the exception_handler is nice and straight-forward compared to the cryptic click-specific approach and
this is actually a standard Python feature. For what it's worth I don't think the "unexpected things" are a concern here because everything goes through one code path, without explicitly having to add something (decorators) to certain places, requiring new test cases that are hopefully written correctly. Overall I think this is less error-prone than the alternative. So, I prefer this one.

I second @elopio regarding tests, we should only have integration tests specifically verifying return codes and traceback behavior. Anything else can basically drop tests for the exit code, because we can rely on the exception behavior without being paranoid about it.

kyrofa:
Thanks for the review @kalikiana, I agree. @sergiusens please provide your thoughts unfiltered by me. The final decision is of course up to you.

sergiusens:
If everyone prefers this, let's go with it.

sergiusens:
I just think it goes against one item in the zen of python a bit
_Explicit is better than implicit._

snapcraft/tests/test_cli_errors.py
```diff
@@ -0,0 +1,93 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
```
sergiusens: why isn't this module under `snapcraft.tests.cli.test_cli_errors` ?

snapcraft/internal/errors.py
```diff
@@ -41,13 +41,24 @@ def __init__(self, **kwargs):
     def __str__(self):
         return self.fmt.format([], **self.__dict__)
 
+    @property
```
sergiusens: this doesn't necessarily need to be a property, it might as well be a class attribute, yes you can change it, but someone can override this behavior too.

I am less inclined about properties these days.

kyrofa:
Ignore the last push, throwing to CI while I hop in the van. Back later to clean up.

snapcraft/internal/errors.py
kyrofa: All I really care about here is API design: this is designed to be reimplemented by subclasses. I find that to be clear given a function API, making it an attribute makes that relationship less clear and more difficult to use. I'd prefer having this be either a property (function) or a real function.

snapcraft/tests/test_cli_errors.py
kyrofa: Good question. I'm not sure. It is now!

kyrofa:
Alright all done. This is a large PR since it also includes an exception sweep, but it's broken down into nice reviewable chunks via commits.

snapcraft/internal/errors.py
sergiusens: If it makes it clearer in your mind, that is all I care about.
Classes and inheritance in python is weird at times where there is no one way to do things and in general people prefer not using `property` decorators. But :+1:

snapcraft/internal/errors.py
```diff
@@ -166,6 +166,14 @@ class MissingGadgetError(SnapcraftError):
         'https://github.com/snapcore/snapd/wiki/Gadget-snap')
 
 
+class PluginOutdatedError(SnapcraftError):
+
+    fmt = 'This plugin is outdated: {message}'
+
+    def __init__(self, message):
```
sergiusens: you only really need to define `fmt` in these cases, `__init__` would do the right thing given `kwargs` in the super class.

snapcraft/cli/parts.py
```diff
@@ -13,11 +13,10 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
-import sys
+
```
sergiusens: these new lines are unnecessary ;)

snapcraft/_store.py
```diff
@@ -281,34 +274,27 @@ def create_key(name):
...
 
 
 def _maybe_prompt_for_key(name):
     keys = list(_get_usable_keys(name=name))
     if not keys:
         if name is not None:
-            raise RuntimeError(
-                'You have no usable key named "{}".\nSee the keys available '
-                'in your system with `snapcraft keys`.'.format(name))
+            raise storeapi.errors.NoSuchKeyError(name)
```
sergiusens: why storeapi.errors and in other just errors? Maybe consolidate with storeapi.errors

snapcraft/tests/commands/test_sign_build.py
```diff
@@ -57,15 +57,30 @@ def test_sign_build_snapd_not_installed(self, mock_installed,
...
+            ['sign-build', self.snap_test.snap_path, '--local'])
 
-        self.assertThat(result.exit_code, Equals(1))
-        self.assertThat(result.output, Contains(
+        self.assertThat(str(raised), Contains(
             'The snapd package is not installed.'))
         mock_installed.assert_called_with('snapd')
         self.assertEqual(0, mock_check_output.call_count)
 
+    # @mock.patch('subprocess.check_output')
```
sergiusens: what happened here?

snapcraft/internal/errors.py
kyrofa: Yeah, but I have two problems with using the super class's `__init__` directly:

1. When there's only one or two options (both of which are required) it doesn't add meaning, just typing.
2. Typos in the kwargs doesn't result in even a linter error, but completely borks the exception. It's bitten me a few times.

I find subclassing and using positional (or kwargs the linter can pick up on) to be safer. If you disagree I can fix it.

snapcraft/cli/parts.py
kyrofa: Fixed.

snapcraft/_store.py
kyrofa: Good call, done.

snapcraft/tests/commands/test_sign_build.py
kyrofa: Whoops, ignore this! It's gone.

kyrofa:
> try and amend the changes into this commit, this seems like a commit worthy of just a rebase on top of master instead of a squash.

Yeah good idea. The review has been trivial enough so far I've just been amending the commits in question, but even if we get something more involved I can always reorganize commits at the end if we want to merge instead of squash.

sergiusens:
The tests had failed as you changed `exit_code` to a `property` but didn't update the fake error in the test. I reverted this to a `property` as if not it should have been named `get_exit_code`

snapcraft/internal/errors.py
kalikiana: What kind of errors are you referring to here? Maybe we can add some validation to the base class?

PR #1447, opened 2017-08-09
add support for the "contact" field in snapcraft

mvo5:
This allows a snapcraft.yaml to have a new "contact" field. This is part of a wider meta-data improvement, outlined in https://forum.snapcraft.io/t/development-sprint-june-26th-2017/415/48. The idea is that the metadata is part of snapcraft.yaml and that snapcraft pushes the metadata via some new store API to the store. The idea is that the metadata can be changed on the store side (e.g. to add new translations without having to push a new snap revision) but to still make it transparent for people who do not use the web-ui. A conflict mechanism is also planned. 

This PR is one (important) first step because right now the other metadata (summary, description) can already be part of snapcraft.yaml but `contact` is currently not supported.

LP: #1624829 (partly)

### DIFF
--- a/demos/godd/snap/snapcraft.yaml
+++ b/demos/godd/snap/snapcraft.yaml
@@ -5,6 +5,7 @@ description:
  Written in go with support for device auto-detection via libgudev,
  you would need to use hw-assign to access devices.
 confinement: strict
+contact: mailto:michaelvogt+godd@imap.cc
 
 apps:
   godd:
--- a/schema/snapcraft.yaml
+++ b/schema/snapcraft.yaml
@@ -116,6 +116,9 @@ properties:
         - type: string
           minLength: 1
         - type: number
+  contact:
+    type: string
+    description: contact information for the given snap
   apps:
     type: object
     additionalProperties: false


demos/godd/snap/snapcraft.yaml
```diff
@@ -5,6 +5,7 @@ description:
  Written in go with support for device auto-detection via libgudev,
  you would need to use hw-assign to access devices.
 confinement: strict
+contact: mailto:michaelvogt+godd@imap.cc
```
kalikiana: Does use of the mailto scheme suggest that there could be others? Like tel or http? If that's the case maybe this should be a list?

mvo5: Yeah, I think we want to support http: etc as well.

sergiusens:
Hi, thanks for adding this but I need to ask: What advantage do we have in introducing this now given the current state of affairs where changes in the metadata are not reflected store side?

kyrofa:
> What advantage do we have in introducing this now given the current state of affairs where changes in the metadata are not reflected store side?

Probably none, but the idea as I understand it is to eventually have snapcraft provide this information to the store out-of-band, via an API. It needs to have this information in the YAML before that can happen. Do you want to hold off?

kyrofa:
Where does this PR stand?

mvo5:
@kyrofa If there is a chance that this gets merged I am happy to update it. It depends a bit on how things are going with the new store API but my gut-feeling is that this is still useful.

sergiusens:
`contact` is currently being discussed on the forum, once that is sorted we can reopen or redo this PR.

Thanks and sorry for the wait.

PR #1471, opened 2017-08-09
catkin plugin: support passing args to cmake

kyrofa:
This PR resolves LP: [#1709640](https://bugs.launchpad.net/snapcraft/+bug/1709640) by adding a new plugin option called `cmake-args` that accepts a list of cmake parameters. If provided, they are appended at the end of the `catkin_make_isolated` invocation.

### DIFF
--- a/demos/ros/snap/snapcraft.yaml
+++ b/demos/ros/snap/snapcraft.yaml
@@ -17,4 +17,3 @@ parts:
       - talker
       - listener
     include-roscore: true
-    build-attributes: [debug]
--- a/snapcraft/plugins/catkin.py
+++ b/snapcraft/plugins/catkin.py
@@ -38,6 +38,9 @@ Additionally, this plugin uses the following plugin-specific keywords:
       (list of strings)
       List of rosinstall files to merge while pulling. Paths are relative to
       the source.
+    - catkin-cmake-args:
+      (list of strings)
+      Configure flags to pass onto the cmake invocation from catkin.
     - underlay:
       (object)
       Used to inform Snapcraft that this snap isn't standalone, and is actually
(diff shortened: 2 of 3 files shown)


kyrofa:
> We have configflags for the other plugins and the first words in the description allude to that, but we said we'd prepend the plugin name to the keyword for the entry; which isn't cmake, and even in cmake we have config flags, so maybe that is the better term here.

Yeah it's tricky. These options aren't for catkin, they're for cmake. Catkin users will be familiar with this functionality being available via the `--cmake-args` parameter on catkin, hence the naming choice on my part. Calling it `configflags` like cmake makes me think the flags are being consumed by catkin, which isn't the case. There may not be a perfect rule we can follow on this one.

kyrofa:
> Can you limit this PR to just the addition of being able to pass cmake parameters and put the other stuff in a different PR please?

Yes, but the test changes here (other than a one-liner) are because adding the new schema check made the schema function too complex, so the changes are still due to this PR. Would you still like me to split them out knowing that the schema check WILL get more complex?

sergiusens:
On mié, 9 ago, 2017 at 5:45 PM, Kyle Fazzari 
>> We have configflags for the other plugins and the first words in the 
>> description allude to that, but we said we'd prepend the plugin name 
>> to the keyword for the entry; which isn't cmake, and even in cmake 
>> we have config flags, so maybe that is the better term here.
>> 
> Yeah it's tricky. These options aren't for catkin, they're for cmake. 
> Catkin users will be familiar with this functionality being available 
> via the --cmake-args parameter on catkin, hence the naming choice on 
> my part. Calling it configflags like cmake makes me think the flags 
> are being consumed by catkin, which isn't the case. There may not be 
> a perfect rule we can follow on this one.
> 
How about `catkin-cmake-args` then ?

kyrofa:
> How about `catkin-cmake-args` then ?

Done!

PR #1474, opened 2017-08-10
kbuild plugin: use ARCH from environment if set

kalikiana:
In some cases the required ARCH doesn't match the kernel architecture, for example **arm** versus **arm64** with [uboot](https://github.com/kubiko/dragonboard-gadget).

With this change `ARCH=arm snapcraft --target-arch=arm64` [does the right thing](https://github.com/kubiko/dragonboard-gadget/pull/1).

Discussion: https://forum.snapcraft.io/t/custom-environment-variables-for-kbuild-and-other-plugins/1639


kubiko:
I think better way here would be way to set env variable in snapcraft.yaml
As this change would not fix the build in launchpad

PR #1478, opened 2017-08-10
Add cx_Freeze options targeting bin/snapcraft

alextnewman:
This is a clean repeat of #1427 in order to address Github metadata issues and eliminate overall noise. 

This adds minimal win32 cx_Freeze support and refactors setup.py to account for incompatibilities in `setup()` while eliminating redundant packaging data. Ideally, this should not impact the current setup.py behavior beyond using cx_Freeze. 

This work is not exhaustively tested but should produce a working snapcraft.exe. I also tested `bdist_msi` and found that this can be deployed with an installer.

At this time, I'm targeting `bin/snapcraft` despite the fact that this is supposedly deprecated because it is far easier to pass to cx_Freeze as an entry point. I'd welcome a more future-proof approach.

### DIFF
--- a/bin/snapcraft
+++ b/bin/snapcraft
@@ -19,7 +19,11 @@ import os
 import sys
 
 # make running from bzr snapshot easy
-topdir = os.path.abspath(os.path.join(__file__, '..', '..'))
+if getattr(sys, 'frozen', False):
+	topdir = os.path.dirname(sys.executable)
+else:
+	topdir = os.path.abspath(os.path.join(__file__, '..', '..'))
+	
 if os.path.exists(os.path.join(topdir, 'setup.py')):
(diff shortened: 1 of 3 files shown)


alextnewman:
@sergiusens It should be un-conflicting now. Remind me never to attempt to use the GitHub web-based editor again...

sergiusens:
This worked great when testing. Excellent and thank you very much!

sergiusens:
On vie, 11 ago, 2017 at 10:06 AM, Alex T Newman 
> @sergiusens It should be un-conflicting now. Remind me never to 
> attempt to use the GitHub web-based editor again...
> 
Yeah, that is only easy for really really trivial commits :wink:
>

PR #1481, opened 2017-08-11
cli: better error message for missing mksquashfs

sergiusens:

### DIFF
--- a/snapcraft/internal/errors.py
+++ b/snapcraft/internal/errors.py
@@ -138,6 +138,8 @@ class MissingCommandError(SnapcraftError):
     )
 
     def __init__(self, required_commands):
+        if isinstance(required_commands, list):
+            required_commands = required_commands.join(' ')
         super().__init__(required_commands=required_commands)
 
 
--- a/snapcraft/internal/errors.py
+++ b/snapcraft/internal/errors.py
@@ -138,8 +138,6 @@ class MissingCommandError(SnapcraftError):
     )
 
     def __init__(self, required_commands):
-        if isinstance(required_commands, list):
-            required_commands = required_commands.join(' ')
         super().__init__(required_commands=required_commands)
 
 
--- a/snapcraft/internal/lifecycle.py
+++ b/snapcraft/internal/lifecycle.py
@@ -368,6 +368,9 @@ def _snap_data_from_dir(directory):
 
 
 def snap(project_options, directory=None, output=None):
+    # Check for our prerequesite external command early
+    repo.check_for_command('mksquashfs')
+
     if directory:
         prime_dir = os.path.abspath(directory)
         snap = _snap_data_from_dir(prime_dir)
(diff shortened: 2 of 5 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+// Place your settings in this file to overwrite default and user settings.
```
kyrofa: Is this supposed to be here?

snapcraft/tests/test_lifecycle.py
```diff
@@ -1061,3 +1062,14 @@ def _create_core_snap(self):
             print('description: description', file=f)
 
         return lifecycle.snap(self.project_options, directory=core_path)
+
+
+class SnapErrorsTestCase(BaseLifecycleTestCase):
+
+    def test_mksquashfs_missing(self):
+        with mock.patch('shutil.which') as which_mock:
```
kyrofa: Why this instead of as a decorator?

snapcraft/internal/errors.py
```diff
@@ -138,6 +138,8 @@ class MissingCommandError(SnapcraftError):
     )
 
     def __init__(self, required_commands):
+        if isinstance(required_commands, list):
+            required_commands = required_commands.join(' ')
```
kyrofa: It looks like you're no longer calling this exception with a list. Is this logic still required?

.vscode/settings.json
sergiusens: uh, no

snapcraft/tests/test_lifecycle.py
sergiusens: in this case I just felt like it, if the test were more complex I'd have more of a reason to

sergiusens:
I do want to add that I want to extend this exception with more guided help depending on the OS you are on to tell you how to get the command we need.

snapcraft/internal/repo/__init__.py
```diff
@@ -30,7 +30,7 @@
 
 def check_for_command(command):
     if not shutil.which(command):
-        raise MissingCommandError([command])
+        raise MissingCommandError(command)
```
kalikiana: Nice one cleaning this up!

snapcraft/internal/lifecycle.py
```diff
@@ -368,6 +368,9 @@ def _snap_data_from_dir(directory):
 
 
 def snap(project_options, directory=None, output=None):
+    # Check for our prerequesite external command early
+    repo.check_for_command('mksquashfs')
```
kalikiana: This might become rather complex very easily, but what about something like ubuntu_packages['squashfs-tools'] and respective options for other platforms?

sergiusens: I think that implementing in the generic `Repo` would help, so that

```python
>>> print(Ubuntu.get_installable('mksquashfs'))
Installable(package='squashfs-tools', tool='apt')
>>> print(Homebrew.get_installable('mksquashfs'))
Installable(package='squashfs', tool='brew')
>>> print(Ubuntu.get_installable('lxd'))
Installable(package='lxd', tool='snap')
```

Names and attributes are all just made up on the fly, we only need to support the tools snapcraft requires and it isn't as much of a deal on an installation from a package either.

PR #1483, opened 2017-08-11
lxd: path cannot have extra forward slashes

kalikiana:
After double-checking that the logic for mounting the project folder on the Snapcraft side I ended up finding that LXD can end up claiming the device exists without actually mounting it. In that case `lxc device remove` aborts with internal errors.
Apparently the LXD issue can be avoided by getting rid of a leading / in the mounted path.

This fixes [Could not find snap/snapcraft.yaml](https://forum.snapcraft.io/t/repeated-container-builds-fail-to-find-snapcraft-yaml/1649).

Notes:
* I'm not especially happy with the test side of this, it's a virtually invisible change. Suggestions would be welcome.
* Maybe we should be moving to using `lxc edit` for configuration as it can be used to fix containers that would be stuck using dedicated commands.

### DIFF
--- a/snapcraft/internal/lxd.py
+++ b/snapcraft/internal/lxd.py
@@ -317,7 +317,7 @@ class Project(Containerbuild):
             check_call([
                 'lxc', 'config', 'device', 'add', self._container_name,
                 destination, 'disk', 'source={}'.format(source),
-                'path=/{}'.format(destination)])
+                'path={}'.format(destination)])
 
     def _finish(self):
         # Nothing to do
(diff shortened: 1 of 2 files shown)


sergiusens:
Once @elopio's branch lands for container testing we can extend these tests, just add a personal note or bug in launchpad to keep track of it.

elopio:
@kalikiana I see that path still has two slashes:

`'path=//root/build_basic'`

Could that be a problem?

I was able to build the snap in containers many times, so +1, just pending on the answer to that question.

kalikiana:
@elopio If by **still** you mean an existing container, yes, the bug won't be fixed for that container. As mentioned in the description, `lxc config edit` seems to be the only way to fix that. I'm looking into that, but it'll be more invasive changes.

PR #1484, opened 2017-08-11
lxd: configure user in container

kalikiana:
This branch enables a distinction between building with a user inside the container and root (as currently is the only option). Setting up the user needs some care with regard to file ownership, it includes knowing what commands to run as root (`apt-get`) and as the user (`mkdir`).

Note this is largely a proof of concept, [the discussion](https://forum.snapcraft.io/t/run-snapcraft-in-container-as-a-user/1438/5) on the exact behavior is still on-going.


kyrofa:
Given the forum discussion, where does this PR stand?

PR #1486, opened 2017-08-14
core: improve source caching logic

sergiusens:
We only really need to cache downloaded files and not local files. This also transparently enables caching for plugins that don't call pull but instead use `download` and `provision`.


sergiusens:
@kalikiana the unit test does that by simple merit of checking that `requests.get` is not called on the second call to pull.

Issue #1454, opened 2017-08-09
python plugin recording


sergiusens:
[design](https://forum.snapcraft.io/t/record-build-details-of-python-snaps/1640)

sergiusens:
#1487 implements this

sergiusens:
#1521 added to this

PR #1487, opened 2017-08-14
python plugin: record manifest

elopio:


sergiusens:
Fixes #1454

snapcraft/tests/plugins/test_python.py
```diff
@@ -121,26 +132,33 @@ def test_env(self):
...
                                      self.project_options)
         setup_directories(plugin, self.options.python_version)
         plugin.pull()
         self.assertFalse(mock_run.called)
 
+    @mock.patch.object(python.PythonPlugin, 'run_output')
     @mock.patch.object(python.PythonPlugin, 'run')
-    def test_pull_with_nothing(self, mock_run):
+    def test_pull_with_nothing(self, mock_run, mock_run_output):
+        mock_run_output.side_effect = fake_empty_pip_list
```
sparkiegeek: This can/should be done in the `@mock.patch.object` line - e.g. `@mock.patch.object(python.PythonPlugin, 'run_output', side_effect=fake_empty_pip_list)`

Makes a clearer distinction between your fixture and the actual test

sergiusens:
It feels like storing the requirements file is almost informational now, isn't it?

Given that we record the exact packages to install and can potentially use constraints to avoid some weird behaviour (unbeknownst to us) of pip picking packages outside of what was locked down. I guess the reason we store requirements files is in case it is something external to the sources of the project.

I don't have any critique here, just thinking out loud as this has prevented me from reviewing properly since I first saw the PR Monday night.

snapcraft/tests/plugins/test_python.py
sergiusens: This looks like sane advice. I just reviewed another PR where I made a comment about it getting complicated on reviewing what we want to test versus the setup for that and the resulting verification. This is a big thing for tests that have complicated fixtures to setup.

sergiusens:
Since testing was brought up, now that we are using testtools, I much prefer
```python
self.assertThat(output, Equals(expected)
```
instead of
```python
self.assertEqual(output, expected)
```

since in the latter when writing (made easier with good variable naming) or executing it is never clear what is what you wanted versus what you got.

kyrofa:
> I much prefer
> ```python
> self.assertThat(output, Equals(expected)
> ```
> instead of
> ```python
> self.assertEqual(output, expected)
> ```

> since in the latter when writing (made easier with good variable naming) or executing it is never clear what is what you wanted versus what you got.

Agreed, especially when you consider that, with testtools, it's `assertEqual(expected, actual)` where order matters.

snapcraft/tests/plugins/test_python.py
elopio: That's a good suggestion. Thanks!

elopio:
Alright, thanks for the reviews!
I updated the decorator, assertThat, and modified the code to only record the file contents if they are remote.
Please take another look.

snapcraft/_baseplugin.py
```diff
@@ -131,6 +131,10 @@ def clean_build(self):
         """Clean the artifacts that resulted from building this part."""
         pass
 
+    def get_manifest(self):
+        """Return the information to record after the build of this part."""
```
kyrofa: Does the return value here need to be a dict? Documenting expectations here would be nice.

elopio: Yes, that would be very good.
We haven't finished discussing about docstrings. How do you imagine this? As prose, or just an `rtype` sphinx tag?
I added `:rtype: dict`, let me know what do you think.

kyrofa: The prose you have in addition to rtype is excellent.

snapcraft/plugins/python.py
```diff
@@ -346,7 +351,19 @@ def build(self):
 
         setup_file = os.path.join(self.builddir, 'setup.py')
         with simple_env_bzr(os.path.join(self.installdir, 'bin')):
-            self._run_pip(setup_file)
+            installed_pipy_packages = self._run_pip(setup_file)
+        # We record the requirements and constraints files only if they are
+        # remote. If they are local, they are already tracked with the source.
+        if self.options.requirements and isurl(self.options.requirements):
```
sergiusens: this check need improvement as != url does not imply it is part of the source. You can do this in a follow up PR though.

sergiusens:
Fixes #1454

PR #1488, opened 2017-08-16
lxd: always remove existing device for project folder

kalikiana:
As a follow-up on #1483 this change removes an existing device for the project folder.


kalikiana:
@sergiusens I think you may have misunderstood how the test works. FakeLXD gets values like name and status from the mock commands that are executed. So there's no need to set anything. However, this particular test fakes an existing state of FakeLXD without any commands. If you prefer, the test could actually execute commands to create a container.

kyrofa:
@kalikiana I think what @sergiusens means is that there are a lot of container tests in here all mocking similar things. Perhaps this would be a good opportunity to refactor these tests to have a common base class that extracts the common mocks, which will increase readability for all of them.

PR #1490, opened 2017-08-17
ant plugin, gradle plugin: add support for authenticated proxies

### LINKED ISSUE #1485: Make script aware of 407 when downloading gradle
builds.snapcraft.io rejects downloading gradle.

```
Downloading https://services.gradle.org/distributions/gradle-4.0.2-bin.zip

Exception in thread "main" java.io.IOException: Unable to tunnel through proxy. Proxy returns "HTTP/1.1 407 Proxy Authentication Required"
```

Source: https://build.snapcraft.io/user/JabRef/jabref/67152

An issue has been submitted - see https://github.com/canonical-websites/build.snapcraft.io/issues/918. I don't know, if this will be fixed or if it is easier to quickyl adapt `gradle.py`.

I assume, gradle.py has to be adapted accordingly? https://github.com/snapcore/snapcraft/blob/master/snapcraft/plugins/gradle.py#L88; if gradlew exists, execute it. If that fails with proxy 407 error (grepping stdout?), execute gradle

sparkiegeek:
LP: [#1682534](https://bugs.launchpad.net/snapcraft/+bug/1682534)

ant plugin, gradle plugin: Add support for authenticated proxies

Bring ant and gradle up to par with maven plugin wrt. authenticated proxies. Fixes #1485 and re-fixes the linked LP bug (AFAICT it doesn't need the unauthenticated proxy that @cjwatson suggested in the LP bug)

### DIFF
--- a/snapcraft/plugins/ant.py
+++ b/snapcraft/plugins/ant.py
@@ -105,6 +105,10 @@ class AntPlugin(snapcraft.plugins.jdk.JdkPlugin):
                 yield '-D{}.proxyHost={}'.format(scheme, parsed.hostname)
             if parsed.port is not None:
                 yield '-D{}.proxyPort={}'.format(scheme, parsed.port)
+            if parsed.username is not None:
+                yield '-D{}.proxyUser={}'.format(scheme, parsed.username)
+            if parsed.password is not None:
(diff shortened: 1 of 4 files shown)


evandandrea:
Looks good! Thank you.

PR #1493, opened 2017-08-18
lxd: only pass target arch if specified explicitly

kalikiana:
As [discussed in the forum discussion](https://forum.snapcraft.io/t/cleanbuild-remote-on-pi-grabs-wrong-arch-lxc-image/691/7) container builds should default to the architecture of the container rather than the client.

~~Alternative approach to #1498~~.

### DIFF
--- a/snapcraft/_options.py
+++ b/snapcraft/_options.py
@@ -150,6 +150,10 @@ class ProjectOptions:
         return self.__target_machine != self.__platform_arch
 
     @property
+    def target_arch(self):
+        return self.__target_arch
+
+    @property
     def cross_compiler_prefix(self):
         try:
             return self.__machine_info['cross-compiler-prefix']
@@ -248,6 +252,7 @@ class ProjectOptions:
 
     def _set_machine(self, target_deb_arch):
         self.__platform_arch = _get_platform_architecture()
+        self.__target_arch = target_deb_arch
         if not target_deb_arch:
             self.__target_machine = self.__platform_arch
         else:
--- a/snapcraft/internal/lxd.py
+++ b/snapcraft/internal/lxd.py
@@ -75,7 +75,6 @@ class Containerbuild:
         if not deb_arch:
             raise ContainerConnectionError(
                 'Unrecognized server architecture {}'.format(kernel))
-        self._host_arch = deb_arch
         self._image = 'ubuntu:xenial/{}'.format(deb_arch)
         # Use a temporary folder the 'lxd' snap can access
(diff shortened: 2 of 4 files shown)


snapcraft/_options.py
```diff
@@ -150,6 +150,10 @@ def is_cross_compiling(self):
         return self.__target_machine != self.__platform_arch
 
     @property
+    def target_arch_specified(self):
+        return self.__target_arch_specified
```
kyrofa: Looking at this from an external point of view, this seems a slightly odd property to make part of the class's public API. It begs the question: "Target arch specified for what?" Why don't we just save the target arch? Then if it's `None` we can start making assumptions about it.

snapcraft/internal/lxd.py
```diff
@@ -141,7 +140,7 @@ def execute(self, step='snap', args=None):
             command = ['snapcraft', step]
             if step == 'snap':
                 command += ['--output', self._snap_output]
-            if self._host_arch != self._project_options.deb_arch:
+            if self._project_options.target_arch_specified:
```
kyrofa: Looking at this further, why aren't we just passing the cli args used to actually call snapcraft on to the container? Why try to reproduce them by looking back at the options, which essentially represent the already-parsed arguments?

snapcraft/_options.py
kalikiana: I didn't go for just saving because `deb_arch` right now means either client or target arch. No strong opinion, though, so I replaced `target_arch_specified` with `target_arch` now which can be **None**.

snapcraft/internal/lxd.py
kalikiana: I was going to argue that it wouldn't solve the problem, but then I [made another PR](https://github.com/snapcore/snapcraft/pull/1498).

kyrofa:
Does https://github.com/snapcore/snapcraft/pull/1498 actually solve this problem?

kalikiana:
Yes, it does. You can see the same test cases for the architecture in both PRs ;-)

kyrofa:
> This was my response (as stated in the comment) to your question [...]

To be clear, your original comment (the one I read) was: "I was going to argue that it wouldn't solve the problem, but then I made another PR." That's not at all clearly indicating that it was an alternative in any way. I should not have to compare code to determine that.

kalikiana:
@kyrofa Sorry about the confusion. I edited the comment in an attempt to clarify, but also failed to point it out in the PR description. I'll make an effort to be more verbose.

And thank you for testing this!

PR #1494, opened 2017-08-18
Parameter forwarding for java object was blocked when using wrapper script

fmanea:
Removed reduntant code (Make plugin, makefile and wrapper)
This is merely an improvement, and not a bug.
Signed-off-by: Flavian Manea <flavian.manea@rinftech.com>

Thanks for helping us make a better Snapcraft!

Checklist:

- Have you signed the contributor licence agreement?
  I have signed the licence agreement.

- Is there a reported a bug for the problem you are fixing?
  There is no bug for this problem, I just stumbled upon this when developing more complex examples, but it would have helped if this example would have been in the form in which I modified it, it could save other a little headache.

- Have you read our contribution guide?
  I have read the contribution guide.


kyrofa:
That demo doesn't _accept_ any args, does it? That said, I'm not sure why a wrapper was being used anyway, perhaps it was historical. So while the reason for this change doesn't make sense to me, the change itself seems fine.

Can you please take a look at the template for the PR, though? Those questions are important (though you can probably ignore the LP bug these days).

fmanea:
I took a look at the template and edited my comment.

fmanea:
I have received also "Sally Coome (sallycoome) added you as a member of Canonical Contributor
Agreement (contributor-agreement-canonical).", so I guess, everything is okay now.

PR #1497, opened 2017-08-19
ci: release snap to a branch for every PR

sergiusens:


elopio:
:)

sergiusens:
@cprov can you help me make this work? Or is this a lost cause given the forks?

elopio:
@sergiusens I need this to happen very badly. Any updates? Can I help?

elopio:
Now it's clear that this won't fly because it would need a secret environment variable in public pull requests.

We should insist on build.snapcraft.io to support this use case. And in the meantime, we can use transfer to publish the generated snap.

kyrofa:
I just want to throw this out there: circle ci supports this. You just need to check a checkbox.

elopio:
I personally prefer Travis, despite the shortcomings. But I wouldn't be strongly opposed to migrate to circle CI if it works better for the rest. Your call.

kyrofa:
Nah, just something I recently learned about that I wanted to share. Migrating the whole thing would be a pain. I would suggest just putting the snap build there, but I know you're wanting to use the resulting snap in later tests.

PR #1498, opened 2017-08-21
lxd: pass original CLI arguments down to container

kalikiana:
Originally trying to fix the --target-arch default as [discussed in the forum](https://forum.snapcraft.io/t/cleanbuild-remote-on-pi-grabs-wrong-arch-lxc-image/691/7) so container builds default to the architecture of the container rather than the client, I followed up [on a suggestion](https://github.com/snapcore/snapcraft/pull/1493) to pass the original commands instead of picking options we need by hand.
This also addresses [`--no-parallel-build` being ignored when building in a container](https://forum.snapcraft.io/t/cleanbuild-no-parallel-build-exclusive/614/4).

**Alternative approach to #1493**.

### DIFF
--- a/snapcraft/_options.py
+++ b/snapcraft/_options.py
@@ -150,8 +150,8 @@ class ProjectOptions:
         return self.__target_machine != self.__platform_arch
 
     @property
-    def target_arch_specified(self):
-        return self.__target_arch_specified
+    def target_arch(self):
+        return self.__target_arch
 
     @property
     def cross_compiler_prefix(self):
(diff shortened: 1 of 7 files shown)


kyrofa:
Perhaps a dumb question, but... can you not just use `sys.argv` here?

kalikiana:
sys.argv would mean arguments have different names like -`d` versus `--debug`, which I have to check in `lxd.py`.

kyrofa:
> sys.argv would mean arguments have different names like -`d` versus `--debug`, which I have to check in `lxd.py`.

Just to make sure `--debug` is passed first? Doesn't passing `sys.argv` guarantee that?

kalikiana:
> Just to make sure --debug is passed first? Doesn't passing sys.argv guarantee that?

Only if you run it as `snapcraft -d ...`. Older versions, such as what's in the archives right now, will error out if you do `snapcraft ... --debug`. So you'd be in a situation where some arguments don't work the same **if** you're building in a container and not using the snap.

Issue #1492, opened 2017-08-17
Support for advanced filtering for build-packages


sergiusens:
[design](https://forum.snapcraft.io/t/specifying-architecture-on-cross-compilations/1641/20?u=sergiusens)

sergiusens:
sources has it's own issue now (#1687), let's consider this one done

PR #1500, opened 2017-08-22
grammar: move into project_loader

kyrofa:
This PR makes progress on #1492 by extracting the stage package grammar from the pluginhandler and placing it in project_loader. This is the first step toward making the grammar more generic and used in more of the YAML. Other than moving the grammar around (and the small import tweaks necessary to continue working with slightly more generic names) this PR makes no changes.


elopio:
I like the new location. But I don't like the name `grammar` for the package. IMO, even something like `package_grammar` would be better. Ideally we would find a word for `stage_or_build_package_grammar`, but I can't think of one that's perfect. Maybe `distro_package_grammar`?

kyrofa:
The problem with that is that we want to start using this grammar for more of the YAML beyond just package lists. For example, after we finish making it work for build packages the next item up is `source`. "Grammar" was the only thing I could think of besides "snapcraft grammar" which seems redundant :P . Totally open to suggestions though.

elopio:
right, that's a problem. `keyword_grammar`, `snaprcart_field_grammar` come to mind. But I'm not sure if they are better. Grammar is good, and we can always change it.

snapcraft/internal/grammar/errors.py
```diff
@@ -17,15 +17,20 @@
...
 
 
-class StagePackageSyntaxError(errors.SnapcraftError):
+class GrammarError(errors.SnapcraftError):
+    """Base class for grammar-related errors."""
+    pass
 
-    fmt = 'Invalid syntax for stage packages: {message}'
+
+class GrammarSyntaxError(GrammarError):
```
kalikiana: That reads a little funny. Makes me wonder if it's a syntax error or wrong grammar ;-)

Generally the name grammar makes sense to me since we want to use it in more places and add new selectors.

sergiusens:
Looks good, wrt the package name `grammar`, we can move it later once we have a clear `project_loader` refactor and insert it there where we can apply the logic on project load.

kyrofa:
> Looks good, wrt the package name grammar, we can move it later once we have a clear project_loader refactor and insert it there where we can apply the logic on project load.

Yeah I can do that now. I'll submit a PR refactoring project_loader to be a package, then the grammar can move into there.

sergiusens:
On mar, 22 ago, 2017 at 6:24 PM, Kyle Fazzari 
>> Looks good, wrt the package name grammar, we can move it later once 
>> we have a clear project_loader refactor and insert it there where we 
>> can apply the logic on project load.
>> 
> Yeah I can do that now. I'll submit a PR refactoring project_loader 
> to be a package, then the grammar can move into there.
> 
Thanks for working on this, I believe this is the best path for some 
good separation of concerns.
>

kyrofa:
> Thanks for working on this, I believe this is the best path for some good separation of concerns.

Agreed. This PR has now been rebased on top of master, and now moves the grammar into the project_loader. Which gives more meaning to "grammar" I think.

PR #1501, opened 2017-08-22
tests: use assertThat instead of assertEqual

elopio:
assertThat from testtools offers a better error message in casee of failure, and makes it easier to use a consistent order in the actual and expected arguments.

### DIFF
--- a/CODE_STYLE.md
+++ b/CODE_STYLE.md
@@ -1,4 +1,4 @@
-# Snapcraft code style guidelines
+# Snapcraft coding style guidelines
 
 When writing code for snapcraft, we try to follow a set of rules that will lead
 to consistency and readability.
--- a/CODE_STYLE.md
+++ b/CODE_STYLE.md
@@ -1,0 +1,28 @@
+# Snapcraft code style guidelines
+
+When writing code for snapcraft, we try to follow a set of rules that will lead
+to consistency and readability.
+
+This is a permanent work in progress, and sometimes being too strict with the
+rules can end up making things actually less readable. So when you disagree
+with one of the rules, please talk to us and help us making it better.
+
+Some of the rules are enforced with static tests. You can read the [TESTING][1]
+document for more information and details about how to run the static suite of
+tests. Some other rules are only socially enforced during code reviews.
+
+## PEP 8
+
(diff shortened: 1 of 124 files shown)


elopio:
To my dear reviewers, this huge PR is for you <3
On friday I was too tired to think, and I had to find something to do in the 4 hours for the flight back. So, I made this big refactor.

CONTRIBUTING.md
```diff
@@ -19,10 +19,13 @@ give us permission to use your contributions.
    Make a fork of Snapcraft, and create a branch named specifically for the
    feature on which you'd like to work. Make your changes there, adding new
    tests as needed, and make sure the existing tests continue to pass when your
-   changes are complete (for information about running the tests, see the
-   [HACKING][3] document).
+   changes are complete (for information see the [HACKING][3] and [TESTING][4]
+   documents).
 
-3. Squash commits into one, well-formatted commit. If you really feel like there
+3. We try to follow a consistent and readable code style. Read the
```
kyrofa: Step 2 discusses making the changes necessary. Should "make your changes while following the style guide" just be part of step 2? Or does it make that step too big?

kyrofa:
Blech, this was mind-numbingly boring to review, I can only imagine what it was like to write. However, it is much appreciated!

@kalikiana a while back we switched to testtools, which still has an `assertEqual`. However, whereas previously the parameters were order-independent, testtools uses `assertEqual(expected, actual)` and presents failures as such. Of course, we were all over the place with our ordering, so the failure messages were often misleading. Using `assertThat(actual, Equals(expected))` is clear from the code which is actual and which is expected, and also results in a helpful failure.

CONTRIBUTING.md
elopio: I prefer this in a separate step, on it's own. But I don't have a strong opinion.

elopio:
@kalikiana I added the description.

CODE_STYLE.md
```diff
@@ -0,0 +1,28 @@
+# Snapcraft code style guidelines
```
kalikiana: "Coding style" is the term I would expect here

PR #1503, opened 2017-08-22
ci: speedup the CLA check

sergiusens:
Verify an email only once and don't verify commits that are already in master


elopio:
Thank you for this!

It is failing with: error: unknown option `format="%(refname)"'

tools/cla_check.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python
+
+# XXX copied from https://github.com/kyrofa/cla-check
+import collections
+import os
+import subprocess
+import sys
+
+Commit = collections.namedtuple('Commit', ['email', 'hash'])
```
elopio: This should go after all the imports.

sergiusens:
On mié, 23 ago, 2017 at 2:45 AM, Leo Arias <notifications@github.com> 
wrote:
> Thank you for this!
> 
> It is failing with: error: unknown option `format="%(refname)"'
> 

Yeah, seems the usable git on that docker image is not the latest and 
greatest :-)
>

tools/cla_check.py
```diff
@@ -0,0 +1,75 @@
...
+    split_output = (i.split('|') for i in output.split('\n') if i != '')
+    commits = [Commit(email=i[0], hash=i[1]) for i in split_output]
+    return commits
+
+
+def get_emails_from_commits(commits):
+    emails = set()
+    for c in commits:
+        output = check_output(['git', 'branch', '--contains', c.hash])
+        if '* master' not in output.split('\n'):
```
kyrofa: This of course means that the user is _on_ master. Is that really a requirement?

sergiusens: no, not the user, the commit

kyrofa: The output of this commit is a list of branches containing the commit in question. An asterisk is next to the branch the user is currently on.

sergiusens: oh ic `*`

PR #1505, opened 2017-08-24
errors: introduce ContainerError

kalikiana:
As suggested during the review of #1302 container-related errors should use their own error class.


kalikiana:
I gave it a shot. Also split out Snapd errors.

kyrofa:
Yeah that's better. Ideally we'd refactor the messages to be more in the error class rather than the caller, but this is an improvement.

sergiusens:
Good point @kyrofa, let's see if that can be done

PR #1507, opened 2017-08-25
many: simplify plugin loading

kyrofa:
Currently, loading a plugin goes through many layers of redirection, which is confusing and causes components to be integrated more tightly than would be ideal. This PR simplifies it by no longer loading plugins within PluginHandler, but rather accepting an already-loaded plugin. That plugin is then loaded in PartsConfig, where it is then handed to the PluginHandler.

Beyond an improved design, this also gives us the ability to extract grammar processing completely from the PluginHandler, which will be done in a follow-up PR.

### DIFF
--- a/snapcraft/internal/pluginhandler/_plugin_loader.py
+++ b/snapcraft/internal/pluginhandler/_plugin_loader.py
@@ -1,0 +1,217 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2017 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import contextlib
+import importlib
+import logging
+import sys
+
+import jsonschema
+
+import snapcraft
+from snapcraft.internal.project_loader.errors import YamlValidationError
+from snapcraft.internal import (
+    errors,
+    sources,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def load_plugin(plugin_name, part_name, project_options, properties,
+                part_schema, definitions_schema):
+    module_name = plugin_name.replace('-', '_')
+    module = _load_module(module_name, plugin_name, project_options)
+    plugin_class = _get_plugin(module)
+    if not plugin_class:
+        raise errors.PluginError(
+            'no plugin found in module {!r}'.format(plugin_name))
+
+    _validate_pull_and_build_properties(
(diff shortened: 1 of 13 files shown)


snapcraft/internal/project_loader/_parts_config.py
```diff
@@ -149,16 +149,47 @@ def validate(self, part_names):
...
+            part_name=part_name,
+            properties=part_properties,
             project_options=self._project_options,
             part_schema=self._validator.part_schema,
             definitions_schema=self._validator.definitions_schema)
 
-        self.build_tools += part.code.build_packages
+        part_schema = {}
+        definitions_schema = {}
+        project_options = snapcraft.ProjectOptions()
```
sergiusens: why not `self._project_options`?

snapcraft/internal/pluginhandler/_plugin_loader.py
```diff
@@ -0,0 +1,217 @@
...
+from snapcraft.internal.project_loader.errors import YamlValidationError
+from snapcraft.internal import (
+    errors,
+    sources,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def load_plugin(plugin_name, part_name, project_options, properties,
```
sergiusens: this should probably be split into `get_plugin_for(plugin_name)` and `init_plugin(...)`

snapcraft/internal/project_loader/_parts_config.py
kyrofa: It actually is, [below](https://github.com/snapcore/snapcraft/pull/1507/files/7cdd84e2d81c85871c990a0d6f7c60b8a7bad5b7#diff-e13dcb301fd4acd6302bee50b3091b03R178). This is keeping the same checks as was present before the refactor. I wasn't sure why there were there, but they seemed relatively harmless and removing them made me nervous. We can try though, and see what the integration tests think.

snapcraft/internal/pluginhandler/_plugin_loader.py
kyrofa: To summarize our conversation on rocket: yes, but the init function will be in PartsConfig to do things like process the grammar. I'll do that in future PRs where the grammar is actually processed in PartsConfig.

snapcraft/internal/project_loader/_parts_config.py
sergiusens: It should just work, this is probably to avoid unit test changes which I don't think would fail today. Please give it a try and remove that if below

snapcraft/internal/project_loader/_parts_config.py
```diff
@@ -149,16 +149,47 @@ def validate(self, part_names):
...
+
+        logger.debug('Setting up part {!r} with plugin {!r} and '
+                     'properties {!r}.'.format(part_name,
+                                               plugin_name,
+                                               part_properties))
+
+        part = pluginhandler.PluginHandler(
+            plugin=plugin,
+            part_properties=part_properties,
+            project_options=project_options,
```
sergiusens: so, `project_options=self._project_options,`

kyrofa: Done.

snapcraft/internal/project_loader/_parts_config.py
kyrofa: Yeah I was able to get rid of the other checks as well.

PR #1509, opened 2017-08-25
project_loader: process stage package grammar

kyrofa:
This PR makes more progress on #1492 by extracting the processing of the stage package grammar from the pluginhandler and placing it into the project_loader. This is done by creating an encapsulating grammar processor dedicated to the grammar in the part (with the vision of adding grammar support to more part properties in the future, e.g. `source`). The pluginhandler then uses this processor to extract grammar-powered properties, leaving the raw YAML alone for state tracking.


snapcraft/_baseplugin.py
```diff
@@ -106,6 +106,8 @@ def __init__(self, name, options, project=None):
 
         self.sourcedir = os.path.join(self.partdir, 'src')
         self.installdir = os.path.join(self.partdir, 'install')
+        self.statedir = os.path.join(self.partdir, 'state')
+        self.packagedir = os.path.join(self.partdir, 'ubuntu')
```
sergiusens: should probably be called `osrepodir`

snapcraft/internal/errors.py
```diff
@@ -213,3 +213,11 @@ class InvalidBuildPropertiesError(SnapcraftError):
 
     def __init__(self, plugin_name, properties):
         super().__init__(plugin_name=plugin_name, properties=properties)
+
+
+class StagePackageDownloadError(SnapcraftError):
```
sergiusens: shouldn't this be in `snapcraft.internal.repo.errors`?

sergiusens: oh, maybe not after a look on how it is used.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -64,24 +63,26 @@ def installdir(self):
...
         self.deps = []
 
         self.stagedir = project_options.stage_dir
         self.primedir = project_options.prime_dir
 
-        parts_dir = project_options.parts_dir
-        self.ubuntudir = os.path.join(parts_dir, self.name, 'ubuntu')
-        self.statedir = os.path.join(parts_dir, self.name, 'state')
-        self.sourcedir = os.path.join(parts_dir, self.name, 'src')
+        self.packagedir = plugin.packagedir
```
sergiusens: why not just always refer to `self.plugin.<>` instead of `self.<>`?

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -227,14 +222,21 @@ def mark_cleaned(self, step):
             os.rmdir(self.statedir)
 
     def _fetch_stage_packages(self):
-        try:
-            self.stage_packages = self._stage_package_handler.fetch()
-        except repo.errors.PackageNotFoundError as e:
-            raise RuntimeError("Error downloading stage packages for part "
-                               "{!r}: {}".format(self.name, e.message))
+        stage_packages = self._grammar_processor.get_stage_packages()
```
sergiusens: I would of loved to have this processed already by the `project_loader` but I guess that is hard due to the repo sources thing :-/

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -227,14 +222,21 @@ def mark_cleaned(self, step):
...
+            logger.debug('Fetching stage-packages {!r}'.format(stage_packages))
+            try:
+                self.stage_packages = self._stage_packages_repo.get(
+                    stage_packages)
+            except repo.errors.PackageNotFoundError as e:
+                raise errors.StagePackageDownloadError(self.name, e.message)
 
     def _unpack_stage_packages(self):
-        self._stage_package_handler.unpack(self.installdir)
+        stage_packages = self._grammar_processor.get_stage_packages()
```
sergiusens: you can process these only once with a `@property` trick or using `get_stage_packages`, in both cases storing the result

snapcraft/internal/project_loader/_parts_config.py
```diff
@@ -169,12 +172,25 @@ def load_part(self, part_name, plugin_name, part_properties):
                                                plugin_name,
                                                part_properties))
 
+        sources = getattr(plugin, 'PLUGIN_STAGE_SOURCES', None)
```
sergiusens: if set to None in baseplugin we can get away with `getattr`

snapcraft/internal/project_loader/_parts_config.py
```diff
@@ -169,12 +172,25 @@ def load_part(self, part_name, plugin_name, part_properties):
...
+        stage_packages_repo = repo.Repo(
+            plugin.packagedir, sources=sources,
+            project_options=self._project_options)
+
+        grammar_processor = grammar_processing.PartGrammarProcessor(
```
sergiusens: ah I see now, you just execute the logic once in the plugin. There are benefits to this I guess as the repo is not setup during project load, I guess this is why you went down this path now.

sergiusens: This should be a pattern that everything using grammar follows though.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -64,24 +63,26 @@ def installdir(self):
         return self.plugin.installdir
 
     def __init__(self, *, plugin, part_properties, project_options,
-                 part_schema, definitions_schema):
+                 part_schema, definitions_schema, stage_packages_repo,
+                 grammar_processor):
```
elopio: wooo, dependency injection in the processor. I like it.

snapcraft/tests/pluginhandler/test_stage_package_handler.py
```diff
@@ -1,89 +0,0 @@
...
-        self.addCleanup(patcher.stop)
-
-        self.get_mock = self.repo_mock.return_value.get
-        self.unpack_mock = self.repo_mock.return_value.unpack
-        self.is_valid_mock = self.repo_mock.return_value.is_valid
-
-        self.cache_dir = os.path.join(os.getcwd(), 'cache')
-        self.unpack_dir = os.path.join(os.getcwd(), 'unpack')
-
-    def test_fetch(self):
```
kalikiana: test_fetch and test_unpack don't seem to have moved into the new world? Or did I miss it?

snapcraft/internal/pluginhandler/__init__.py
sergiusens: nvm, this might not be necessary at all

snapcraft/tests/pluginhandler/test_stage_package_handler.py
sergiusens: @kyrofa please look into this.

Given the refactor I think the things tested aren't relevant.

kyrofa: Indeed, this entire class was blown away. Now we're fetching/unpacking directly with the repo again.

PR #1510, opened 2017-08-26
ci: disable the travis deploy stage for docs

sergiusens:
Docs generation is broken causing all tests against master to
be broken as well giving us little certainty about the current
code quality.


kyrofa:
Why delete the scripts? Don't we want this fixed at some point?

sergiusens:
@kyrofa I would rather re do them when the time comes just like cla-check. I want out of script maze :wink:

kalikiana:
What did break the docs? This isn't a revert of any kind.

sergiusens:
@kalikiana it was always temporary, credentials expired, and they weren't taken care of. I'd rather have nothing than something broken.

Not sure about you, but I get the travis runs for all things that go into master and this thing being red just causes too much noise.

PR #1511, opened 2017-08-26
project_loader: support grammar on build-packages

kyrofa:
Currently Snapcraft only supports grammar-powered stage-packages. This PR makes another good chunk of progress on #1492 by adding support for grammar-powered build-packages as well, both at the part-level and global.

Note that this PR depends upon #1509, but was proposed anyway as I'm leaving for a few days. It also may give the other PR more context.


kyrofa:
> Very nicely done, it does seem that build tools is all over the place though, would be good if we the processing for it in some internal lifecycle step in some future.

Yeah, that would be a nice simplification indeed.

PR #1513, opened 2017-08-28
lifecycle: outdated step should raise SnapcraftError

kalikiana:
Outdated steps should raise derivatives of SnapcraftError

### DIFF
--- a/snapcraft/internal/errors.py
+++ b/snapcraft/internal/errors.py
@@ -14,6 +14,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+from snapcraft import formatting_utils
+
 
 class SnapcraftError(Exception):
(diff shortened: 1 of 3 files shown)


snapcraft/internal/errors.py
```diff
@@ -45,6 +47,57 @@ def __init__(self, step):
...
+
+    fmt = (
+        'The {step!r} step of {part!r} is out of date:\n'
+        '{report}'
+        'In order to continue, please clean that part\'s {step!r} step '
+        'by running:\n'
+        'snapcraft clean {parts_names} -s {step}\n'
+    )
+
+    def __init__(self, step, part,
```
sergiusens: please add an `*` after self to force all of these to be kwargs (API clarity in the long wrong)

kalikiana: Hmm I had it in there before but it didn't seem to work... I guess it was unrelated. Added the * back!

PR #1515, opened 2017-08-29
 tests: use a fake pip, instead of mocking everything

elopio:
I think we are using too many mocks for subprocess. If we just mock every call to subprocess, the tests become less clear, and they don't really cover everything that might be happening.
This is challenging, because things are order dependent. So I've been experimenting with mocks and fakes.

The idea is that we will mock only the specific command that we are testing, letting everything else continue calling the real thing. We replace the command with a fake implementation that will record some values in memory when required, and return the same things that the real would return. In addition, we keep a mock where we can assert the order of the calls.

Before going any further, I would like to put this for discussion. This approach also has drawbacks, but I would like to implement it with all our plugins and with LXD/LXC.

### DIFF
--- a/snapcraft/tests/plugins/test_python.py
+++ b/snapcraft/tests/plugins/test_python.py
@@ -149,7 +149,7 @@ class FakePip(fixtures.Fixture):
                 self.download['constraints'] = constraint.read().splitlines()
             args.remove(args[index])
             args.remove(args[index])
-        if args[0] == '--requirement':
+        if '--requirement' in args:
             index = args.index('--requirement')
             with open(args[index + 1]) as requirement:
                 self.download['requirements'] = requirement.read().splitlines()
--- a/snapcraft/tests/plugins/test_python.py
+++ b/snapcraft/tests/plugins/test_python.py
@@ -15,13 +15,15 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import collections
+import json
 import os
+import subprocess
 import tempfile
 from glob import glob
 from unittest import mock
 
 import fixtures
-from testtools.matchers import Equals, FileContains, HasLength
+from testtools.matchers import Contains, Equals, FileContains, HasLength, Not
 
 import snapcraft
 from snapcraft import tests
@@ -29,7 +31,7 @@ from snapcraft.tests import fixture_setup
 from snapcraft.plugins import python
 
 
-def setup_directories(plugin, python_version):
+def setup_directories(plugin, python_version, setup_file=True):
     version = '2.7' if python_version == 'python2' else '3.5'
     os.makedirs(plugin.sourcedir)
     os.makedirs(plugin.builddir)
@@ -40,7 +42,8 @@ def setup_directories(plugin, python_version):
 
     os.makedirs(os.path.join(python_lib_path, 'dist-packages'))
     os.makedirs(python_include_path)
-    open(os.path.join(plugin.sourcedir, 'setup.py'), 'w').close()
+    if setup_file:
+        open(os.path.join(plugin.sourcedir, 'setup.py'), 'w').close()
 
     site_path = os.path.join(plugin.installdir, 'lib', 'python' + version,
                              'site-packages')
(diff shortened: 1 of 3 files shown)


kalikiana:
You seem to be doing two things here in terms of the test API:
* `call(['pip', 'install', 'foo']` -> `fake_pip.install['installed']`
    How about a proper API here? Like `fake_pip.installed_packages`
* `----disable-pip-version-check` -> `fake_pip.install['disable_pip_version_check']`
    How about `fake_pip.args`? Since these are literally the CLI args used that you want to check here and it'll be clearer from the test what's a verbatim argument and what's a result value.

For comparison, with `FakeLXD` values like `.name`, `.status` and mocked JSON reflect the mocked commands that were run in the test, like `fake_pip.installed` does. But tests still assert mocked calls as a "safety guard", for example if an `apt-get` call is accidentally missing, you will see it - I'm not sure what the equivalent would be in FakeLXD terms.
Would you mind focussing your proposal on designing the test API for FakePIP and FakeLXD with a couple of examples? That seems more crucial for me to form an opinion than the implementation behind it. Maybe we can discuss it in a hangout.

snapcraft/tests/plugins/test_python.py
```diff
@@ -61,6 +64,123 @@ def fake_empty_pip_list(*args, **kwargs):
         return ''
 
 
+class FakePip(fixtures.Fixture):
```
sergiusens: put this in `snapcraft.tests.plugins.fakes` or something.

snapcraft/tests/plugins/test_python.py
```diff
@@ -86,7 +211,93 @@ class Options:
...
+        plugin = python.PythonPlugin('test-part', self.options,
+                                     self.project_options)
+        setup_directories(
+            plugin, self.options.python_version, setup_file=False)
+        with FakePip() as fake_pip:
+            plugin.pull()
+            # XXX Just a safeguard to make sure that the mock doesn't get out
+            # of sync.
+            self.assertThat(
+                fake_pip.mock_call.mock_calls[2][1][0],
```
sergiusens: sorry, but `[2][1][0]` is even harder to read.

snapcraft/tests/plugins/test_python.py
```diff
@@ -86,7 +211,93 @@ class Options:
...
+    def test_pull_with_nothing(self):
+        plugin = python.PythonPlugin('test-part', self.options,
+                                     self.project_options)
+        setup_directories(plugin, self.options.python_version)
+
+        with FakePip() as fake_pip:
+            plugin.pull()
+            self.assertThat(fake_pip.download['constraints'], Equals([]))
+            self.assertThat(fake_pip.download['requirements'], Equals([]))
+            self.assertThat(fake_pip.download['downloaded'], Equals([]))
```
sergiusens: just by reading this and not the fake, this check really doesn't make sense given the key names used here.

sergiusens:
Good first start, but needs a twist of knobs to make it scale into the future :-)

snapcraft/tests/plugins/test_python.py
kyrofa: Yeah this is tough to parse when trying to figure out what we're testing. Is this verifying that `download` isn't called without `'.'`? Or `install`? Or `wheel`?

snapcraft/tests/plugins/test_python.py
elopio: I've moved it to snapcraft/tests/fixture_setup/fake_pip.py, imported as snapcraft.tests.fixture_setup.FakePip. How about that?

snapcraft/tests/plugins/test_python.py
elopio: I think it was a mistake to add this for discussion. This is a [spy](http://xunitpatterns.com/Test%20Spy.html), and I'm not yet happy with any spy implementations I've tried.  
This is an unnecessary assertion, because if the setup.py doesn't exist and '.' is passed to the fake, it will raise an exception. So the fact that it doesn't fail is a good test. However, that's opaque and could lead to a fake that gets out of sync.

I've changed this to store in a boolean in a variable if '.' is passed to the fake. So it's more readable, but it's still opaque, it's not clear what leads to that variable being true by just reading at the test. I think that's better, but not perfect either.

And the third option would be to keep the `assert_has_calls` just as before, maybe replacing some already tested values with mock.ANY to reduce the burden of maintaining that. However, if we use `assert_has_calls` a lot, there's no reason to keep the fake.

So three options:
 * Rely on the fake raising an exception as the real would do.
 * If '.' is received as an argument, save that info in a variable.
 * Keep using `assert_has_calls`.

What do you think?

snapcraft/tests/plugins/test_python.py
elopio: I've added a get_downloaded method to the fake. Please take a look and let me know if it's better.

PR #1518, opened 2017-08-30
project: introduce build-snaps

sergiusens:
Introduce build-snaps with support for advanced grammar.

### DIFF
--- a/integration_tests/test_build_snap_grammar.py
+++ b/integration_tests/test_build_snap_grammar.py
@@ -1,0 +1,105 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2017 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+import fileinput
+import subprocess
+
+import testscenarios
+from testtools.matchers import Contains, Equals
+
+import integration_tests
+import snapcraft
+
+
+def _construct_scenarios():
+    main_scenarios = {
+        'build-snap-grammar': True,
+        'build-snap-grammar-try': True,
+        'build-snap-grammar-try-skip': False,
+        'build-snap-grammar-try-else': True,
+        'build-snap-grammar-on': False,
+        'build-snap-grammar-on-else': True,
+    }
+
+    # Just some combinations
+    channel_scenarios = [
+        '', '/stable', '/latest/stable']
+
+    all_scenarios = []
+    for project, expected_install in main_scenarios.items():
+        for channel in channel_scenarios:
+            d = dict(project=project, hello_installed=expected_install,
+                     channel=channel)
+            scenario = ('{}{}'.format(project, channel), d)
+            all_scenarios.append(scenario)
+
+    return all_scenarios
+
+
+class BuildSnapGrammarTestCase(testscenarios.WithScenarios,
+                               integration_tests.TestCase):
+
+    scenarios = _construct_scenarios()
+
+    def setUp(self):
+        super().setUp()
+        if self._hello_is_installed():
+            self.fail(
+                'This integration test cannot run if you already have the '
+                "'hello' package installed. Please uninstall it before "
+                'running this test.')
+
+    def tearDown(self):
+        super().tearDown()
+
+        # Remove hello. This is safe since the test fails if hello was already
+        # installed.
+        try:
+            subprocess.check_output(
+                ['sudo', 'snap', 'remove', 'hello'],
+                stderr=subprocess.STDOUT)
+        except subprocess.CalledProcessError as e:
+            self.fail("unable to remove 'hello': {}".format(e.output))
+
+    def _hello_is_installed(self):
+        return snapcraft.repo.snaps.SnapPackage.is_snap_installed('hello')
+
+    def _add_channel_information_to_hello(self):
+        replacement = '- hello{}'.format(self.channel)
+        with fileinput.input('snapcraft.yaml', inplace=True) as input_file:
+            for line in input_file:
+                print(line.replace('- hello', replacement), end='')
+
+    def test_grammar(self):
+        self.copy_project_to_cwd(self.project)
+        self._add_channel_information_to_hello()
+
+        self.run_snapcraft('pull')
+
+        self.assertThat(self._hello_is_installed(),
+                        Equals(self.hello_installed))
+
+
+class BuildSnapGrammarErrorsTestCase(integration_tests.TestCase):
+
+    def test_on_other_arch_else_fail(self):
+        """Test that 'on' fails with an error if it hits an 'else fail'."""
+
+        exception = self.assertRaises(
+            subprocess.CalledProcessError, self.run_snapcraft,
+            ['pull'], 'build-snap-grammar-fail')
+
+        self.assertThat(exception.output, Contains(
+            "Unable to satisfy 'on other-arch', failure forced"))
--- a/integration_tests/test_build_snap_grammar.py
+++ b/integration_tests/test_build_snap_grammar.py
@@ -58,8 +58,8 @@ class BuildSnapGrammarTestCase(testscenarios.WithScenarios,
         if self._hello_is_installed():
             self.fail(
                 'This integration test cannot run if you already have the '
-                "'hello' package installed. Please uninstall it before "
-                'running this test.')
+                "'hello' snap installed. Please uninstall it "
+                "by running 'sudo snap remove hello'.")
 
     def tearDown(self):
         super().tearDown()
--- a/schema/snapcraft.yaml
+++ b/schema/snapcraft.yaml
@@ -276,6 +276,9 @@ properties:
           stage-packages:
             $ref: "#/definitions/grammar-array"
             default: [] # For some reason this doesn't work if in the ref
+          build-snaps:
+            $ref: "#/definitions/grammar-array"
+            default: [] # For some reason this doesn't work if in the ref
           build-packages:
             $ref: "#/definitions/grammar-array"
             default: [] # For some reason this doesn't work if in the ref
--- a/snapcraft/internal/project_loader/_config.py
+++ b/snapcraft/internal/project_loader/_config.py
@@ -89,6 +89,7 @@ class Config:
         if project_options is None:
             project_options = snapcraft.ProjectOptions()
 
+        self.build_snaps = set()
         self.build_tools = []
         self._project_options = project_options
 
@@ -118,6 +119,7 @@ class Config:
         self.parts = PartsConfig(self.data,
                                  self._project_options,
                                  self._validator,
+                                 self.build_snaps,
                                  self.build_tools,
                                  self.snapcraft_yaml_path)
 
--- a/snapcraft/internal/project_loader/_config.py
+++ b/snapcraft/internal/project_loader/_config.py
@@ -116,12 +116,12 @@ class Config:
         self.build_tools = grammar_processor.get_build_packages()
         self.build_tools |= set(project_options.additional_build_packages)
 
-        self.parts = PartsConfig(self.data,
-                                 self._project_options,
-                                 self._validator,
-                                 self.build_snaps,
-                                 self.build_tools,
-                                 self.snapcraft_yaml_path)
+        self.parts = PartsConfig(parts=self.data,
+                                 project_options=self._project_options,
+                                 validator=self._validator,
+                                 build_snaps=self.build_snaps,
+                                 build_tools=self.build_tools,
+                                 snapcraft_yaml=self.snapcraft_yaml_path)
 
         if 'architectures' not in self.data:
             self.data['architectures'] = [self._project_options.deb_arch]
--- a/snapcraft/internal/project_loader/_parts_config.py
+++ b/snapcraft/internal/project_loader/_parts_config.py
@@ -37,13 +37,14 @@ logger = logging.getLogger(__name__)
 
 class PartsConfig:
 
-    def __init__(self, parts, project_options, validator, build_tools,
-                 snapcraft_yaml):
+    def __init__(self, parts, project_options, validator,
+                 build_snaps, build_tools, snapcraft_yaml):
         self._snap_name = parts['name']
         self._confinement = parts['confinement']
         self._parts_data = parts.get('parts', {})
         self._project_options = project_options
         self._validator = validator
+        self.build_snaps = build_snaps
         self.build_tools = build_tools
         self._snapcraft_yaml = snapcraft_yaml
 
@@ -192,6 +193,7 @@ class PartsConfig:
             stage_packages_repo=stage_packages_repo,
             grammar_processor=grammar_processor)
 
+        self.build_snaps |= grammar_processor.get_build_snaps()
         self.build_tools |= grammar_processor.get_build_packages()
 
         if part.source_handler and part.source_handler.command:
--- a/snapcraft/internal/project_loader/_parts_config.py
+++ b/snapcraft/internal/project_loader/_parts_config.py
@@ -37,7 +37,7 @@ logger = logging.getLogger(__name__)
 
 class PartsConfig:
 
-    def __init__(self, parts, project_options, validator,
+    def __init__(self, *, parts, project_options, validator,
                  build_snaps, build_tools, snapcraft_yaml):
         self._snap_name = parts['name']
         self._confinement = parts['confinement']
--- a/snapcraft/internal/repo/snaps.py
+++ b/snapcraft/internal/repo/snaps.py
@@ -36,7 +36,7 @@ class SnapPackage:
     and remote stores to obtain information about the snap, such as its
     confinment value and channel availability.
 
-    This information can also be used determine if a snap should be
+    This information can also be used to determine if a snap should be
     installed or refreshed.
 
     There are risks of the data falling out of date between the query and the
@@ -65,7 +65,7 @@ class SnapPackage:
         self._store_snap_info = None
 
         self._is_installed = None
-        self._is_on_store = None
+        self._is_in_store = None
 
     @property
     def installed(self):
@@ -74,10 +74,10 @@ class SnapPackage:
         return self._is_installed
 
     @property
-    def on_store(self):
-        if self._is_on_store is None:
-            self._is_on_store = self.get_store_snap_info() is not None
-        return self._is_on_store
+    def in_store(self):
+        if self._is_in_store is None:
+            self._is_in_store = self.get_store_snap_info() is not None
+        return self._is_in_store
 
     def get_local_snap_info(self):
         """Returns a local payload for the snap.
@@ -92,14 +92,14 @@ class SnapPackage:
         """Returns a store payload for the snap.
 
         Validity of the results are determined by checking self.remote."""
-        if self._is_on_store is None:
+        if self._is_in_store is None:
             with contextlib.suppress(exceptions.HTTPError):
                 self._store_snap_info = _get_store_snap_info(self.name)
         return self._store_snap_info
 
     def _get_store_channels(self):
         snap_store_info = self.get_store_snap_info()
-        if not self.on_store:
+        if not self.in_store:
             return dict()
 
         return snap_store_info['channels']
@@ -119,7 +119,7 @@ class SnapPackage:
 
     def is_valid(self):
         """Check if the snap is valid."""
-        if not self.on_store:
+        if not self.in_store:
             return False
         store_channels = self._get_store_channels()
         return self.channel in store_channels.keys()
@@ -162,7 +162,7 @@ def install_snaps(snaps_list):
     """Install snaps of the format <snap-name>/<channel>."""
     for snap in snaps_list:
         snap_pkg = SnapPackage(snap)
-        if not snap_pkg.installed and snap_pkg.on_store:
+        if not snap_pkg.installed and snap_pkg.in_store:
             snap_pkg.install()
         elif snap_pkg.get_current_channel() != snap_pkg.channel:
             snap_pkg.refresh()
--- a/snapcraft/internal/repo/snaps.py
+++ b/snapcraft/internal/repo/snaps.py
@@ -25,6 +25,7 @@ from requests import exceptions
 from . import errors
 
 
+_CHANNEL_RISKS = ['stable', 'candidate', 'beta', 'edge']
 logger = logging.getLogger(__name__)
 
 
@@ -108,7 +109,8 @@ class SnapPackage:
         if self.installed:
             local_snap_info = self.get_local_snap_info()
             current_channel = local_snap_info['channel']
-            if current_channel in ('stable', 'candidate', 'beta', 'edge'):
+            if any([current_channel.startswith(risk)
+                    for risk in _CHANNEL_RISKS]):
                 current_channel = 'latest/{}'.format(current_channel)
         return current_channel
 
--- a/snapcraft/internal/repo/snaps.py
+++ b/snapcraft/internal/repo/snaps.py
@@ -133,6 +133,7 @@ class SnapPackage:
         if self._original_channel:
             snap_install_cmd.extend(['--channel', self._original_channel])
         if self.is_classic:
+            # TODO make this a user explicit choice
             snap_install_cmd.extend(['--classic'])
         try:
             check_call(snap_install_cmd)
@@ -147,6 +148,9 @@ class SnapPackage:
             snap_refresh_cmd = ['sudo']
         snap_refresh_cmd.extend(['snap', 'refresh', self.name,
                                  '--channel', self.channel])
+        if self.is_classic:
+            # TODO make this a user explicit choice
+            snap_refresh_cmd.extend(['--classic'])
         try:
             check_call(snap_refresh_cmd)
         except CalledProcessError as install_error:
--- a/snapcraft/internal/repo/snaps.py
+++ b/snapcraft/internal/repo/snaps.py
@@ -133,7 +133,7 @@ class SnapPackage:
             snap_install_cmd.extend(['--channel', self._original_channel])
         if self.is_classic():
             # TODO make this a user explicit choice
-            snap_install_cmd.extend(['--classic'])
+            snap_install_cmd.append('--classic')
         try:
             check_call(snap_install_cmd)
         except CalledProcessError as install_error:
@@ -149,7 +149,7 @@ class SnapPackage:
                                  '--channel', self.channel])
         if self.is_classic():
             # TODO make this a user explicit choice
-            snap_refresh_cmd.extend(['--classic'])
+            snap_refresh_cmd.append('--classic')
         try:
             check_call(snap_refresh_cmd)
         except CalledProcessError as install_error:
@@ -210,7 +210,7 @@ def _get_local_snap_info(snap_name):
 def _get_store_snap_info(snap_name):
     # This logic uses /v2/find returns an array of results, given that
     # we do a strict search either 1 result or a 404 will be returned.
-    slug = '{}{}'.format('find?name=', urllib.parse.quote(snap_name, safe=''))
+    slug = 'find?name={}'.format(urllib.parse.quote(snap_name, safe=''))
     url = get_snapd_socket_path_template().format(slug)
     with requests_unixsocket.Session() as session:
         snap_info = session.get(url)
--- a/snapcraft/internal/repo/snaps.py
+++ b/snapcraft/internal/repo/snaps.py
@@ -186,8 +186,9 @@ def _snap_command_requires_sudo():
 
 def _get_parsed_snap(snap):
     if '/' in snap:
-        snap_name = snap[:snap.find('/')]
-        snap_channel = snap[snap.find('/')+1:]
+        sep_index = snap.find('/')
+        snap_name = snap[:sep_index]
+        snap_channel = snap[sep_index+1:]
     else:
         snap_name = snap
         snap_channel = ''
--- a/snapcraft/internal/repo/snaps.py
+++ b/snapcraft/internal/repo/snaps.py
@@ -34,7 +34,7 @@ class SnapPackage:
 
     It uses information provided by snapd implicitly referring to the local
     and remote stores to obtain information about the snap, such as its
-    confinment value and channel availability.
+    confinement value and channel availability.
 
     This information can also be used to determine if a snap should be
     installed or refreshed.
--- a/snapcraft/internal/repo/snaps.py
+++ b/snapcraft/internal/repo/snaps.py
@@ -16,8 +16,8 @@
 import contextlib
 import logging
 import sys
-import urllib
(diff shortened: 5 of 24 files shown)


snapcraft/internal/project_loader/_config.py
```diff
@@ -89,6 +89,7 @@ def __init__(self, project_options=None):
         if project_options is None:
             project_options = snapcraft.ProjectOptions()
 
+        self.build_snaps = set()
         self.build_tools = []
```
elopio: have I said how much I hate this build_tools? :D
I'm just saying it again to put myself pressure to fix it.

snapcraft/internal/repo/snaps.py
```diff
@@ -0,0 +1,196 @@
...
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
```
elopio: missing empty line

snapcraft/internal/repo/snaps.py
```diff
@@ -0,0 +1,196 @@
...
+
+logger = logging.getLogger(__name__)
+_SNAPD_URI_ROOT_TEMPLATE = 'http+unix://%2Frun%2Fsnapd.socket/v2/{}'
+
+
+class SnapPackage:
+
+    @classmethod
+    def is_valid_snap(cls, snap):
+        return cls(snap).is_valid()
```
elopio: I consider this very weird. A class method that makes an instance. I haven't seen it before, is it an accepted pattern?

I would prefer the caller to make the instance. Or, a little less preferable, to make is_valid_snap a module function instead of a class method.

snapcraft/internal/repo/snaps.py
```diff
@@ -0,0 +1,196 @@
...
+        self._store_snap_info = None
+
+        self._is_installed = None
+        self._is_on_store = None
+
+    @property
+    def installed(self):
+        if self._is_installed is None:
+            self.get_local_snap_info()
+        return self._is_installed
```
elopio: I think it's the property the one that should assign _is_installed, instead of doing it in get_local_snap_info

snapcraft/internal/repo/snaps.py
```diff
@@ -0,0 +1,196 @@
...
+
+    @property
+    def on_store(self):
+        if self._is_on_store is None:
+            self.get_store_snap_info()
+        return self._is_on_store
+
+    def get_local_snap_info(self):
```
elopio: Any reason for not putting an _ in front of this one?

snapcraft/internal/repo/snaps.py
```diff
@@ -0,0 +1,196 @@
...
+    @property
+    def installed(self):
+        if self._is_installed is None:
+            self.get_local_snap_info()
+        return self._is_installed
+
+    @property
+    def on_store(self):
+        if self._is_on_store is None:
+            self.get_store_snap_info()
```
elopio: same comment about assigning _is_on_store

snapcraft/internal/repo/snaps.py
elopio: Another question, can the installed status of the snap change after this property was called the first time? This returns the value of the first call, and never checks again. It feels weird, because the snap can be installed or uninstalled before the second call.

snapcraft/internal/repo/snaps.py
```diff
@@ -0,0 +1,196 @@
...
+            except exceptions.HTTPError:
+                self._is_on_store = False
+        return self._store_snap_info
+
+    def _get_store_channels(self):
+        snap_store_info = self.get_store_snap_info()
+        if not self.on_store:
+            return dict()
+
+        return snap_store_info['result'][0]['channels']
```
elopio: Is that `0` because multiple results can be returned? It would be nice to explain that in a comment.

snapcraft/internal/repo/snaps.py
```diff
@@ -0,0 +1,196 @@
...
+    def install(self):
+        """Installs the snap onto the system."""
+        snap_install_cmd = []
+        if _snap_command_requires_sudo():
+            snap_install_cmd = ['sudo']
+        snap_install_cmd.extend(['snap', 'install', self.name])
+        if self._original_channel:
+            snap_install_cmd.extend(['--channel', self._original_channel])
+        if self.is_classic:
+            snap_install_cmd.extend(['--classic'])
```
elopio: I now this would complicate the snapcraft.yaml, but I would prefer the user to declare if she wants to install the snap as classic.

snapcraft/internal/repo/snaps.py
```diff
@@ -0,0 +1,196 @@
...
+        """Refreshes a snap onto a channel on the system."""
+        snap_refresh_cmd = []
+        if _snap_command_requires_sudo():
+            snap_refresh_cmd = ['sudo']
+        snap_refresh_cmd.extend(['snap', 'refresh', self.name,
+                                 '--channel', self.channel])
+        try:
+            check_call(snap_refresh_cmd)
+        except CalledProcessError as install_error:
+            raise errors.SnapInstallError(snap_name=self.name,
```
elopio: A Refresh error would be more appropriate here.

snapcraft/internal/repo/snaps.py
sergiusens: This is a common pattern, yes

snapcraft/internal/repo/snaps.py
sergiusens: it can change even while the call is taking place even if we iteratively call over and over again. It is just like query and install with `apt`

snapcraft/internal/repo/snaps.py
sergiusens: This means I also need ancillary values for the actual info which I was trying to avoid.

snapcraft/internal/repo/snaps.py
sergiusens: yes, I had it in a comment somewhere else before moving this to a class.

snapcraft/internal/repo/snaps.py
sergiusens: this is not what was agreed to on the forum :wink:

snapcraft/internal/repo/snaps.py
sergiusens: this I will indeed ignore ;-)

snapcraft/internal/repo/snaps.py
sergiusens: got it

snapcraft/internal/repo/snaps.py
sergiusens: I have no reason for the fact that it can be used outside even though it isn't right now.

snapcraft/internal/project_loader/_parts_config.py
```diff
@@ -37,13 +37,14 @@
 
 class PartsConfig:
 
-    def __init__(self, parts, project_options, validator, build_tools,
-                 snapcraft_yaml):
+    def __init__(self, parts, project_options, validator,
+                 build_snaps, build_tools, snapcraft_yaml):
```
kalikiana: Can you add a * here and use keywords?

snapcraft/internal/repo/snaps.py
```diff
@@ -0,0 +1,210 @@
...
+
+
+def install_snaps(snaps_list):
+    """Install snaps of the format <snap-name>/<channel>."""
+    for snap in snaps_list:
+        snap_pkg = SnapPackage(snap)
+        if not snap_pkg.installed and snap_pkg.on_store:
+            snap_pkg.install()
+        elif snap_pkg.get_current_channel() != snap_pkg.channel:
+            snap_pkg.refresh()
```
kalikiana: This should also check if is_classic is the same and requires a refresh.

integration_tests/test_build_snap_grammar.py
```diff
@@ -0,0 +1,105 @@
...
+
+    scenarios = _construct_scenarios()
+
+    def setUp(self):
+        super().setUp()
+        if self._hello_is_installed():
+            self.fail(
+                'This integration test cannot run if you already have the '
+                "'hello' package installed. Please uninstall it before "
+                'running this test.')
```
kalikiana: This should say "if you already have the 'hello' **snap** installed", otherwise it makes you think of apt or dnf. Also, print the command that does that.

snapcraft/internal/repo/snaps.py
```diff
@@ -0,0 +1,210 @@
...
+        if self.is_classic:
+            snap_install_cmd.extend(['--classic'])
+        try:
+            check_call(snap_install_cmd)
+        except CalledProcessError as install_error:
+            raise errors.SnapInstallError(snap_name=self.name,
+                                          snap_channel=self.channel)
+
+    def refresh(self):
+        """Refreshes a snap onto a channel on the system."""
```
kalikiana: This should be `Refreshes a snap to the latest version, channel and confinement type` because 1) it will update the snap if there's a new version 2) as per my comment on `install_snaps` confinement may have changed.

snapcraft/internal/repo/snaps.py
```diff
@@ -0,0 +1,210 @@
...
+    """Install snaps of the format <snap-name>/<channel>."""
+    for snap in snaps_list:
+        snap_pkg = SnapPackage(snap)
+        if not snap_pkg.installed and snap_pkg.on_store:
+            snap_pkg.install()
+        elif snap_pkg.get_current_channel() != snap_pkg.channel:
+            snap_pkg.refresh()
+
+
+def _snap_command_requires_sudo():
```
kalikiana: Can we have a unit test for this?

snapcraft/internal/repo/snaps.py
```diff
@@ -0,0 +1,210 @@
...
+
+    def install(self):
+        """Installs the snap onto the system."""
+        snap_install_cmd = []
+        if _snap_command_requires_sudo():
+            snap_install_cmd = ['sudo']
+        snap_install_cmd.extend(['snap', 'install', self.name])
+        if self._original_channel:
+            snap_install_cmd.extend(['--channel', self._original_channel])
+        if self.is_classic:
```
sergiusens: this is missing the TODO comment to make it a user choice on non ci/container environments

snapcraft/internal/repo/snaps.py
sergiusens: ah, right, for people who bounce back and forth; in general terms if installed once with classic you don't need to pass it on for subsequent installs

snapcraft/internal/repo/snaps.py
sergiusens: Yeah, unittest for this whole module coming today :-)

snapcraft/internal/repo/snaps.py
kalikiana: To be clear, I'm talking about the case where snap foo is has strict confinement and a new revision is a classic snap, or vice versa. Afair snapd doesn't change it automatically.

schema/snapcraft.yaml
```diff
@@ -276,6 +276,9 @@ properties:
           stage-packages:
             $ref: "#/definitions/grammar-array"
             default: [] # For some reason this doesn't work if in the ref
+          build-snaps:
+            $ref: "#/definitions/grammar-array"
+            default: [] # For some reason this doesn't work if in the ref
```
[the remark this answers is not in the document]
kyrofa: When I initially wrote the grammar back in the day I wasn't sure what was happening, but I think I know now so we can update the comment: a jsonschema ref isn't expanded in any fancy way by the YAML (which is why I'd like to use a recursive YAML anchor for this, but jsonschema causes things to explode that way since it wants to walk the whole thing). This means that we can't actually access the chunk of YAML representing that bit of the schema very easily in code. Like we do [here](https://github.com/snapcore/snapcraft/blob/master/snapcraft/internal/pluginhandler/__init__.py#L628) to expand defaults. So this is kind of an ugly hack to allow specifying the defaults to continue actually resulting in defaults.

snapcraft/internal/repo/snaps.py
```diff
@@ -0,0 +1,214 @@
...
+
+class SnapPackage:
+    """SnapPackage acts as a mediator to install or refresh a snap.
+
+    It uses information provided by snapd implicitly referring to the local
+    and remote stores to obtain information about the snap, such as its
+    confinment value and channel availability.
+
+    This information can also be used determine if a snap should be
+    installed or refreshed.
```
kyrofa: There should be a "to" in here: "This information can also be used _to_ determine[...]"

snapcraft/internal/repo/snaps.py
```diff
@@ -0,0 +1,214 @@
...
+        self._is_on_store = None
+
+    @property
+    def installed(self):
+        if self._is_installed is None:
+            self._is_installed = self.get_local_snap_info() is not None
+        return self._is_installed
+
+    @property
+    def on_store(self):
```
kyrofa: **On** the store, or **in** the store?

snapcraft/internal/repo/snaps.py
```diff
@@ -0,0 +1,214 @@
...
+            return False
+        store_channels = self._get_store_channels()
+        return self.channel in store_channels.keys()
+
+    def install(self):
+        """Installs the snap onto the system."""
+        snap_install_cmd = []
+        if _snap_command_requires_sudo():
+            snap_install_cmd = ['sudo']
+        snap_install_cmd.extend(['snap', 'install', self.name])
```
kyrofa: Is there an advantage to shelling out as opposed to using the snapd API for this?

snapcraft/internal/repo/snaps.py
```diff
@@ -0,0 +1,214 @@
...
+        return store_channels[self.channel]['confinement'] == 'classic'
+
+    def is_valid(self):
+        """Check if the snap is valid."""
+        if not self.on_store:
+            return False
+        store_channels = self._get_store_channels()
+        return self.channel in store_channels.keys()
+
+    def install(self):
```
kyrofa: What if the snap I want to use to build requires devmode? Are we choosing to not support that?

snapcraft/internal/repo/snaps.py
```diff
@@ -0,0 +1,214 @@
...
+    if requires_root:
+        logger.warning('snapd is not logged in, snap install '
+                       'commands will use sudo')
+    return requires_root
+
+
+def _get_parsed_snap(snap):
+    if '/' in snap:
+        snap_name = snap[:snap.find('/')]
+        snap_channel = snap[snap.find('/')+1:]
```
kyrofa: The two `snap.find('/')`s here are bothering me-- perhaps it should be saved in a variable?

tools/travis/run_lxd_container.sh
```diff
@@ -32,7 +32,7 @@ name="$1"
 lxc="/snap/bin/lxc"
 
 echo "Starting the LXD container."
-$lxc launch --ephemeral ubuntu:xenial "$name"
+$lxc launch --ephemeral --config security.privileged=true ubuntu:xenial "$name"
```
kyrofa: Why do we need privileged containers here?

snapcraft/internal/repo/snaps.py
sergiusens: not at the moment, check the forum post on the subject

snapcraft/internal/repo/snaps.py
sergiusens: it is much more work as we'd need to take care of "uploading" the snap.

sergiusens: Oh wait, not for these snaps. Maybe not, let me look

snapcraft/internal/repo/snaps.py
sergiusens: ok, fair enough

tools/travis/run_lxd_container.sh
sergiusens: The commit that added this has the info, but here it is: https://bugs.launchpad.net/snapd/+bug/1709536

kyrofa: Ah yes, I hit that myself. I fixed it by removing the Nice value from the systemd unit file though as opposed to making the container privileged. It's probably not a big deal, let's just keep an eye on the bug and change back to unprivileged once it's fixed.

snapcraft/internal/repo/snaps.py
sergiusens: yeah, that is what I meant by _bounce back and forth_ :wink:

snapcraft/internal/repo/snaps.py
sergiusens: I'll leave this for future work, I don't even know how stable that API is.

sergiusens: nor where it is documented

kyrofa: Yeah, fair enough.

snapcraft/internal/repo/snaps.py
sergiusens: this is done in the `is_classic` check now

snapcraft/internal/repo/snaps.py
```diff
@@ -0,0 +1,215 @@
...
+    with requests_unixsocket.Session() as session:
+        snap_info = session.get(url)
+    snap_info.raise_for_status()
+    return snap_info.json()['result']
+
+
+def _get_store_snap_info(snap_name):
+    # This logic uses /v2/find returns an array of results, given that
+    # we do a strict search either 1 result or a 404 will be returned.
+    slug = '{}{}'.format('find?name=', urllib.parse.quote(snap_name, safe=''))
```
sparkiegeek: huh? why are you interpolating a fixed string?

'find?name={}'.format(...)` would be more intuitive

snapcraft/internal/repo/snaps.py
```diff
@@ -109,8 +108,8 @@ def get_current_channel(self):
         if self.installed:
             local_snap_info = self.get_local_snap_info()
             current_channel = local_snap_info['channel']
-            if current_channel == 'stable':
-                current_channel = 'latest/stable'
+            if current_channel in ('stable', 'candidate', 'beta', 'edge'):
+                current_channel = 'latest/{}'.format(current_channel)
```
kyrofa: This check won't catch branches, think `build-snaps: go/hotfix-211`. Perhaps instead of a whitelist of risks, just check to see if a track is specified (e.g. searching for `/`)?

snapcraft/internal/repo/snaps.py
kyrofa: I did. It says nothing about devmode, thus my question.

snapcraft/tests/repo/test_snaps.py
```diff
@@ -0,0 +1,365 @@
...
+        cmd = self._get_snap_cmd(cmd)
+        if cmd == 'install' and not self.install_success:
+            raise subprocess.CalledProcessError(returncode=1, cmd=cmd)
+        if cmd == 'refresh' and not self.refresh_success:
+            raise subprocess.CalledProcessError(returncode=1, cmd=cmd)
+        if cmd == 'whoami':
+            return 'email: -'.encode()
+
+
+class FakeSnapdServer(BaseHTTPRequestHandler):
```
kyrofa: Having @chipaca look this fake over to make sure everything is sane would be helpful.

sergiusens: I sort of asked him the necessary questions about API and showed him the original implementation so I don't think he is needed her @kyrofa

snapcraft/internal/repo/snaps.py
sergiusens: I don't know :upside_down_face:

chipaca: any reason for not using `urllib.parse.urlencode` here?

snapcraft/tests/repo/test_snaps.py
chipaca: can confirm, was sort of asked.

snapcraft/internal/repo/snaps.py
chipaca: FWIW using `urllib.parse.quote` like that results in unnecessary %-encoding of characters that don't need it. If you'd rather not use `urlencode` (maybe building a dict isn't to your taste?), just use `urllib.parse.quote_plus` which is the quote function for the query part of a URL.

schema/snapcraft.yaml
```diff
@@ -276,6 +276,9 @@ properties:
           stage-packages:
             $ref: "#/definitions/grammar-array"
             default: [] # For some reason this doesn't work if in the ref
+          build-snaps:
```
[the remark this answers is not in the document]
wandrewkeech: go 1.9 now, already...

snapcraft/internal/repo/snaps.py
sergiusens: `requests_unixsockets` doesn't support the dict as params, but I guess you are implying that urlencode can do that so i'll take a look. The only one I sort of have problems with is `urljoin`.

sergiusens: switched to `urlencode` on bec56ed

snapcraft/internal/repo/snaps.py
sergiusens: right, so it would need a discussion first, right now it shouldn't work. In all fairness, if we discuss this, I'd say it shouldn't just like we are not auto connecting interfaces.

snapcraft/tests/repo/test_snaps.py
```diff
@@ -0,0 +1,377 @@
...
+    def get_request(self):
+        request, client_address = self.socket.accept()
+        # BaseHTTPRequestHandler expects a tuple with the client address at
+        # index 0, so we fake one
+        if len(client_address) == 0:
+            client_address = (self.server_address,)
+        return (request, client_address)
+
+
+class FakeSnapd(fixtures.Fixture):
```
kalikiana: I like your new FakeSnapd a lot. But we already have a FakeSnapd fixture in `snapcraft/tests/fixture_setup.py` so please merge the two.

snapcraft/tests/repo/test_snaps.py
```diff
@@ -0,0 +1,377 @@
...
+            ['snap', 'whoami'],
+            ['sudo', 'snap', 'refresh', 'fake-snap',
+             '--channel', 'classic/stable', '--classic']]))
+
+    def test_refresh_fails(self):
+        snap_pkg = snaps.SnapPackage('fake-snap/strict/stable')
+        self.fake_snap_command.refresh_success = False
+        self.assertRaises(errors.SnapRefreshError, snap_pkg.refresh)
+
+    def test_bundle_install(self):
```
kalikiana: Nit: the term "bundle" confused me. I think `test_install_multiple` would be much clearer.

snapcraft/tests/repo/test_snaps.py
```diff
@@ -0,0 +1,377 @@
...
+         dict(snap='missing-snap/latest/stable',
+              expected=False)),
+    ]
+
+    def test_is_in_store(self):
+        snap_pkg = snaps.SnapPackage(self.snap)
+        self.assertThat(snap_pkg.in_store, Is(self.expected))
+
+
+class SnapPackageIsClasicTest(SnapPackageBaseTestCase):
```
kalikiana: Typo: SnapPackageIsClas**s**icTest should have two **s**.

snapcraft/tests/repo/test_snaps.py
```diff
@@ -0,0 +1,377 @@
...
+    def setUp(self):
+        super().setUp()
+        self.fake_snap_command = FakeSnapCommand()
+        self.useFixture(self.fake_snap_command)
+
+    def test_install_classic(self):
+        snap_pkg = snaps.SnapPackage('fake-snap/classic/stable')
+        snap_pkg.install()
+        self.assertThat(self.fake_snap_command.calls, Equals([
+            ['snap', 'whoami'],
```
kalikiana: There's a test missing for the case of a logged in user where `sudo` won't be used.

snapcraft/tests/repo/test_snaps.py
sergiusens: I would rather not do it here as it would require substantial changes. I'll do it in a follow up PR though.

snapcraft/internal/repo/snaps.py
```diff
@@ -0,0 +1,220 @@
...
+_CHANNEL_RISKS = ['stable', 'candidate', 'beta', 'edge']
+logger = logging.getLogger(__name__)
+
+
+class SnapPackage:
+    """SnapPackage acts as a mediator to install or refresh a snap.
+
+    It uses information provided by snapd implicitly referring to the local
+    and remote stores to obtain information about the snap, such as its
+    confinment value and channel availability.
```
elopio: typo: confinement.

integration_tests/test_build_snap_grammar.py
```diff
@@ -0,0 +1,105 @@
...
+        # Remove hello. This is safe since the test fails if hello was already
+        # installed.
+        try:
+            subprocess.check_output(
+                ['sudo', 'snap', 'remove', 'hello'],
+                stderr=subprocess.STDOUT)
+        except subprocess.CalledProcessError as e:
+            self.fail("unable to remove 'hello': {}".format(e.output))
+
+    def _hello_is_installed(self):
```
elopio: `_is_hello_installed` should be the name of this function.

snapcraft/tests/repo/test_snaps.py
```diff
@@ -0,0 +1,399 @@
...
+
+        def side_effect_check_output(cmd, *args, **kwargs):
+            return side_effect(original_check_output, cmd, *args, **kwargs)
+
+        def side_effect(original, cmd, *args, **kwargs):
+            if self._is_snap_command(cmd):
+                self.calls.append(cmd)
+                return self._fake_snap_command(cmd, *args, **kwargs)
+            else:
+                return original(cmd, *args, **kwargs)
```
elopio: Hey, I like this! ;)

snapcraft/tests/repo/test_snaps.py
```diff
@@ -0,0 +1,399 @@
...
+        return self._get_snap_cmd(cmd) in ['install', 'refresh', 'whoami']
+
+    def _fake_snap_command(self, cmd, *args, **kwargs):
+        cmd = self._get_snap_cmd(cmd)
+        if cmd == 'install' and not self.install_success:
+            raise subprocess.CalledProcessError(returncode=1, cmd=cmd)
+        elif cmd == 'refresh' and not self.refresh_success:
+            raise subprocess.CalledProcessError(returncode=1, cmd=cmd)
+        elif cmd == 'whoami':
+            return 'email: {}'.format(self._email).encode()
```
elopio: I've been reflecting upon these fakes. I think things would be a lot easier if we always called check_output, even when we don't care about the output. That way we only have to make one patch. And if there is only one patch, and in good theory the test should only call the patched function once, then the test itself could inject the exception to raise. That would be clearer to me. But well, all random thoughts for future improvement, we can discuss about the fake tomorrow.

snapcraft/tests/repo/test_snaps.py
```diff
@@ -0,0 +1,399 @@
...
+
+
+class FakeSnapdServer(BaseHTTPRequestHandler):
```
elopio: I would move this fake to snapcraft/tests/fake_servers/snapd.

snapcraft/tests/repo/test_snaps.py
```diff
@@ -0,0 +1,399 @@
...
+            params = {'channel': 'candidate/branch'}
+        elif parsed_url.path.endswith('/fake-snap-track-stable'):
+            status_code = 200
+            params = {'channel': 'track/stable'}
+        elif parsed_url.path.endswith('/fake-snap-track-stable-branch'):
+            status_code = 200
+            params = {'channel': 'track/stable/branch'}
+        elif parsed_url.path.endswith('/fake-snap-edge'):
+            status_code = 200
+            params = {'channel': 'edge'}
```
elopio: This unending stream of elifs caused many problems when we left the fake servers evolve. That's why on the new ones we are using pyramid to route the requests. This is not a lot better, just a little, but it will help for when we are finally able to use https://github.com/pretenders/pretenders

tools/travis/build_snapcraft_snap.sh
```diff
@@ -25,9 +25,12 @@ lxc="/snap/bin/lxc"
 
 "$script_path/setup_lxd.sh"
 "$script_path/run_lxd_container.sh" snap-builder
+# Workaround for
+# - Setup snap "core" (2462) security profiles (cannot reload udev rules: exit status 2
```
elopio: is there a bug for this one?

snapcraft/tests/repo/test_snaps.py
```diff
@@ -0,0 +1,399 @@
...
+         dict(snap='fake-snap-stable/stable',
+              expected='latest/stable')),
+        ('latest/stable',
+         dict(snap='fake-snap-stable/latest/stable',
+              expected='latest/stable')),
+        ('candidate/branch',
+         dict(snap='fake-snap-branch/candidate/branch',
+              expected='latest/candidate/branch')),
+        ('track/stable/branch',
+         dict(snap='fake-snap-track-stable-branch/trakc/stable/branch',
```
elopio: There is a typo here: trakc/track.
This is one of the good reasons to migrate out of those hard-coded fake servers.

integration_tests/test_build_snap_grammar.py
sergiusens: I replicated what was in `test_build_package_grammar`. Let's change this in a different PR.

integration_tests/test_build_snap_grammar.py
```diff
@@ -0,0 +1,105 @@
...
+            scenario = ('{}{}'.format(project, channel), d)
+            all_scenarios.append(scenario)
+
+    return all_scenarios
+
+
+class BuildSnapGrammarTestCase(testscenarios.WithScenarios,
+                               integration_tests.TestCase):
+
+    scenarios = _construct_scenarios()
```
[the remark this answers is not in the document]
sergiusens: Yeah, I saw multiply, this feels much more readable IMO, the other options was as @sparkiegeek mentioned, `subTest` but iirc, `testscenarios` had issues with that, right?

snapcraft/tests/repo/test_snaps.py
sergiusens: Using `check_output` only will not stream the output to `stdout` which is a feature. If you must, we should migrate to our own runner based on `Popen`

snapcraft/tests/repo/test_snaps.py
sergiusens: My next step after this PR is to refactor `fixtures_setup` and others to make it more readable, I was lost inside those modules while trying to make out some structure from it

snapcraft/tests/repo/test_snaps.py
sergiusens: I don't like complicating things at the start though.

tools/travis/build_snapcraft_snap.sh
sergiusens: no idea, I was told it was known

snapcraft/tests/repo/test_snaps.py
sergiusens: Fixed, but in the case of this test fixture the request for the _future_ is never made, if I had called it, I would of gotten an error (like `install` or `is_classic`)

integration_tests/test_build_snap_grammar.py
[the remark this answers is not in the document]
elopio: right: https://github.com/testing-cabal/testtools/issues/235

PR #1519, opened 2017-08-31
lxd: use a unique temporary folder

kalikiana:
See the [problem report in the forum](https://forum.snapcraft.io/t/cant-do-two-cleanbuilds-at-once/1883). `snapcraft cleanbuild` or `SNAPCRAFT_CONTAINER_BUILDS=1 snapcraft` can currently fails when used multiple times in parallel and the snapcraft snap is used.

I'm moving the temporary folder to the constructor so that it becomes accessible to tests. It's also a prerequisite to #1302 using folder in the non-snap case so I dropped the `is_snap` condition. Also, I moved the tests from commands.test_snap because they're testing behavior, not command usage.


snapcraft/internal/lxd.py
```diff
@@ -76,6 +77,11 @@ def __init__(self, *, output, source, project_options,
                 'Unrecognized server architecture {}'.format(kernel))
         self._host_arch = deb_arch
         self._image = 'ubuntu:xenial/{}'.format(deb_arch)
+        # Use a temporary folder the 'lxd' snap can access
+        lxd_common = os.path.expanduser(
+            os.path.join('~', 'snap', 'lxd', 'common'))
+        os.makedirs(lxd_common, exist_ok=True)
+        self.tmp = tempfile.mkdtemp(prefix='snapcraft', dir=lxd_common)
```
sergiusens: please rename `tmp` to `tmp_dir` for clarity.
While we are at it, maybe even rename `lxd_common` to `lxd_common_dir`

snapcraft/tests/test_lxd.py
```diff
@@ -186,3 +191,148 @@ def test_remote_does_not_exist(self, mock_run):
         with ExpectedException(ContainerConnectionError,
                                'There are either.*{}.*'.format(self.remote)):
             self.make_cleanbuilder()
+
+    def test_parallel_invocation(self):
+        builder1 = self.make_cleanbuilder()
+        builder2 = self.make_cleanbuilder()
+        builder1.execute()
+        builder2.execute()
+
```
sergiusens: Maybe, maybe add the adapted assertion `self.fake_filesystem.rmtree_mock.assert_has_calls([call(builder.tmp)])` for each cleanbuilder instance at the end

snapcraft/tests/test_lxd.py
```diff
@@ -186,3 +191,148 @@ def test_remote_does_not_exist(self, mock_run):
...
+        mock_is_snap.side_effect = lambda: True
+
+        fake_snapd = tests.fixture_setup.FakeSnapd()
+        self.useFixture(fake_snapd)
+
+        builder1 = self.make_cleanbuilder()
+        builder2 = self.make_cleanbuilder()
+        builder1.execute()
+        builder2.execute()
+
```
sergiusens: Maybe, maybe add the adapted assertion `self.fake_filesystem.rmtree_mock.assert_has_calls([call(builder.tmp)])` for each cleanbuilder instance at the end

PR #1521, opened 2017-08-31
python plugin: always record constraints and requirements contents

elopio:

-----
Sergio suggested an existential question about the usefulness of recording these files if they are already tracked in the source. I made a naive implementation of ignoring the file if it's not from a remote server, but this brings more questions about when we should and when we shouldn't record the contents. If you have an opinion, please comment on the forum:

https://forum.snapcraft.io/t/record-build-details-of-python-snaps/1640/4

For now, I'm just recording it always.


sergiusens:
you should check all items as you have verified that is the case.

PR #1522, opened 2017-08-31
catkin plugin: only append PYTHONPATH if set

kyrofa:
This PR fixes LP: [#1714333](https://bugs.launchpad.net/snapcraft/+bug/1714333) by only including the original PYTHONPATH if it's actually set. Otherwise, Python interprets empty PYTHONPATH segments as the current working directory, which in the case of the Catkin plugin is not desired.

### DIFF
--- a/snapcraft/tests/plugins/test_catkin.py
+++ b/snapcraft/tests/plugins/test_catkin.py
@@ -14,12 +14,16 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import ast
 import builtins
 import os
 import os.path
 import re
 import subprocess
 import shutil
+import sys
+import tempfile
+import textwrap
 
 from unittest import mock
 import testtools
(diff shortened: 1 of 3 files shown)


elopio:
@kyrofa I think that a functional test would be nicer here. Like test with and without PYTHONPATH in env, and check that no ":" is left at the end. Do you think it would be worth the extra work?

kyrofa:
@elopio I agree, but such a thing would require the snap to be made and installed, i.e. a demo test. Do you want a demo just for this?

Which leads me to: it would be awesome to have a class of integration tests that get installed in an environment as opposed to only doing that for our demos.

kyrofa:
@elopio alright, here's a test that evaluates the environment by writing it to a file and executing it. It then verified that there are no empty segments in PYTHONPATH. Thoughts?

elopio:
I like it @kyrofa!

snapcraft/tests/plugins/test_catkin.py
```diff
@@ -629,6 +633,59 @@ def test_run_environment_with_underlay(self, run_mock):
...
+
+        self._assert_pythonpath_no_empty_segments(environment)
+
+    def test_pythonpath_if_not_empty(self):
+        environment = self._evaluate_environment(textwrap.dedent("""
+            export PYTHONPATH=foo
+        """))
+
+        # Verify that the PYTHONPATH contains no empty segments
+        segments = self._assert_pythonpath_no_empty_segments(environment)
```
sergiusens: reading this is not straightforward, ideally you have a helper that returns segments.

kyrofa: Yeah that was a little odd. Fixed.

Issue #1453, opened 2017-08-09
nodejs plugin recording


sergiusens:
[design](https://forum.snapcraft.io/t/record-build-details-of-nodejs-snaps/1925)

PR #1524, opened 2017-09-01
node plugin: record the yarn.lock file

elopio:

-----
This is part of #1453.
Requires #1523

### DIFF
--- a/snapcraft/tests/plugins/test_nodejs.py
+++ b/snapcraft/tests/plugins/test_nodejs.py
@@ -397,6 +397,32 @@ class NodePluginManifestTestCase(NodePluginBaseTestCase):
                      ['testpackage1=1.0', 'testpackage2=1.2']})))
 
 
+class NodePluginYarnLockManifestTestCase(NodePluginBaseTestCase):
+
+    def test_get_manifest_with_yarn_lock_file(self):
+        class Options:
+            source = '.'
+            node_packages = []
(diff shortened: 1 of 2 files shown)


snapcraft/tests/plugins/test_nodejs.py
```diff
@@ -291,6 +298,76 @@ class Options:
...
+                plugin.sourcedir,
+                'yarn.lock'), 'w') as yarn_lock_file:
+            yarn_lock_file.write('test yarn lock contents')
+
+        plugin.build()
+
+        self.assertThat(
+            plugin.get_manifest(), Equals(
+                collections.OrderedDict(
+                    {'yarn-lock-contents': 'test yarn lock contents',
```
kalikiana: The fact that the test can get away with arbitrary data makes me think the plugin should at least do basic validation - but then it took all of 5 minutes in DuckDuckGo to find that it's not YAML but some 500 lines of custom parsing...

So, I think some validation would be nice, whether that's a regex or nodejs command, before saving the lock contents in the manifest. And using a real-looking example in the test. But I'll leave this as a would-be-nice suggestion.

elopio: If the plugin can build the package with the contents of the yarn.lock, we should record whatever was used despite of the specific value. If the value is not valid, the plugin should fail before recording anything.

PR #1526, opened 2017-09-01
catkin plugin: don't assume catkin is in underlay

kyrofa:

-----

Currently the Catkin plugin pulls its own Catkin off to the side in order to search for dependencies in the underlay, but then doesn't use it in isolation, using Catkin components out of the underlay. This breaks if the underlay doesn't actually include Catkin, and makes no sense considering that Catkin is already pulled off to the side. This PR fixes LP: [#1696014](https://bugs.launchpad.net/snapcraft/+bug/1696014) by using that version completely.

### DIFF
--- a/demos/shared-ros/Makefile
+++ b/demos/shared-ros/Makefile
@@ -1,0 +1,16 @@
+all: ros-app/*.snap
+
+ros-base/*.snap:
+	cd ros-base && snapcraft
+
+ros-app/ros-base.tar.bz2: ros-base/*.snap
+	tar czf ros-app/ros-base.tar.bz2 -C ros-base stage/
+
+ros-app/*.snap: ros-app/ros-base.tar.bz2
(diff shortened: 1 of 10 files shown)


sergiusens:
Looks good, the naming scheme under `integration_tests/snaps` for plugins should start with the plugin name for faster identification.

demos/shared-ros/Makefile
```diff
@@ -0,0 +1,16 @@
+all: ros-app/*.snap
```
sergiusens: this rogued itself in, right?

kyrofa: Actually no, I got tired of the manual steps to build the consumer, so decided to wrap it in a Makefile. I doubt this is the last time it'll be useful.

kyrofa:
> start with `catkin` for the integration tests dirnames and snap names

Done.

PR #1527, opened 2017-09-03
repo: Use os-release(5) to detect supported Linux distributions

Conan-Kudo:
All supported Linux distributions provide an `/etc/os-release` file that contains the necessary information for determining supported repository modules to use.

Thus, we can rely on it instead of the deprecated `linux_distribution` method from the `platform` module that will be removed in Python 3.7.

This will also make it much easier to extend to support RPM-based Linux distributions.

Signed-off-by: Neal Gompa <ngompa13@gmail.com>

-----

### DIFF
--- a/snapcraft/internal/common.py
+++ b/snapcraft/internal/common.py
@@ -263,3 +263,12 @@ def get_pkg_config_paths(root, arch_triplet):
     ]
 
     return [p for p in paths if os.path.exists(p)]
+
+
+def get_os_release_info():
+    with open("/etc/os-release") as os_release_file:
+        os_release_dict = {}
+        for line in os_release_file:
+            key, value = line.rstrip().split('=')
+            os_release_dict[key] = value.strip('"')
+    return os_release_dict
--- a/integration_tests/__init__.py
+++ b/integration_tests/__init__.py
@@ -16,7 +16,6 @@
 
 import fileinput
 import os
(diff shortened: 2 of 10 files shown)


Conan-Kudo:
@kyrofa @sergiusens @elopio So, I know what's wrong here, but I don't really want to duplicate myself over half a dozen other places. How should I make this used across the board?

kalikiana:
@Conan-Kudo How about introducing `_get_distro` and moving those into `snapcraft/internal/common.py`?

Conan-Kudo:
@kalikiana I left the name as `_get_os_release_info()` as it's technically not a "distro" specific thing and it's clear on what it's called, but I did move it into common.

Let's see how this goes...

Conan-Kudo:
@kalikiana We're all gravy now!

snapcraft/internal/common.py
```diff
@@ -263,3 +263,12 @@ def get_pkg_config_paths(root, arch_triplet):
     ]
 
     return [p for p in paths if os.path.exists(p)]
+
+
+def _get_os_release_info():
```
elopio: We use the underscore in front of the name to mark something as internal to the file.
This function is called from outside of this file, so please remove the underscore.

Conan-Kudo:
@elopio Done.

Issue #1452, opened 2017-08-09
machine information


sergiusens:
[design](https://forum.snapcraft.io/t/record-machine-information-in-the-manifest/1961)

Issue #36, opened 2015-10-29
Save test results as subunit


sergiusens:
Subunit is the test result stream that we are using in many places. It allows parallel execution and to export to many other formats.
It would be nice if the plainbox results could be streamed to subunit.

https://launchpad.net/subunit

Launchpad Details: [#LP1484597](https://bugs.launchpad.net/bugs/1484597) Leo Arias - 2015-08-13 13:19:11 -0300

sergiusens:
I had a look at subunit but I need more time to prepare everything, marking as triaged. I think it can happen by the next release, after the imminent release of 0.23 (so 0.24)

Launchpad Details: [#LPC](https://bugs.launchpad.net/snapcraft/+bug/1484597/comments/1) Zygmunt Krynicki - 2015-08-20 06:56:44 -0300

sergiusens:
this is now doable with SessionAssistant API for plainbox

Launchpad Details: [#LPC](https://bugs.launchpad.net/snapcraft/+bug/1484597/comments/2) Zygmunt Krynicki - 2015-09-21 20:38:56 -0300

PR #1528, opened 2017-09-04
recording: record the machine information collected by uname

elopio:

-----

Updates: #1452.

### DIFF
--- a/integration_tests/test_asset_recording.py
+++ b/integration_tests/test_asset_recording.py
@@ -17,6 +17,7 @@
 import filecmp
 import os
 import subprocess
+import sys
 import yaml
 
 import fixtures
@@ -54,6 +55,20 @@ class SnapcraftYamlRecordingTestCase(AssetRecordingBaseTestCase):
 
 
 class ManifestRecordingTestCase(AssetRecordingBaseTestCase):
+
+    def test_prime_records_uname(self):
+        self.run_snapcraft('prime', project_dir='basic')
+
+        recorded_yaml_path = os.path.join(
+            self.prime_dir, 'snap', 'manifest.yaml')
+        with open(recorded_yaml_path) as recorded_yaml_file:
+            recorded_yaml = yaml.load(recorded_yaml_file)
+
+        expected_uname = subprocess.check_output(
+            ['uname', '-srvmpio']).decode(sys.getfilesystemencoding()).strip()
+        self.assertThat(
+            recorded_yaml['parts']['dummy-part']['uname'],
+            Equals(expected_uname))
 
     def test_prime_with_architectures(self):
         """Test the recorded manifest for a basic snap
--- a/snapcraft/internal/pluginhandler/__init__.py
+++ b/snapcraft/internal/pluginhandler/__init__.py
@@ -335,10 +335,14 @@ class PluginHandler:
     def mark_build_done(self):
         build_properties = self.plugin.get_build_properties()
         plugin_manifest = self.plugin.get_manifest()
+        machine_manifest = self._get_machine_manifest()
 
         self.mark_done('build', states.BuildState(
-            build_properties, self._part_properties,
-            self._project_options, plugin_manifest))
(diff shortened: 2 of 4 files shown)


snapcraft/tests/test_lifecycle.py
```diff
@@ -724,6 +742,7 @@ def test_prime_with_build_info_records_manifest(self):
     prime: []
     stage: []
     stage-packages: []
+    uname: test uname
```
kalikiana: Out of context this looks like a scriptlet - please use something that resembles the output of uname to avoid confusion.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -333,10 +333,14 @@ def ignore(directory, files):
...
+        machine_manifest = self._get_machine_manifest()
 
         self.mark_done('build', states.BuildState(
-            build_properties, self._part_properties,
-            self._project_options, plugin_manifest))
+            build_properties, self._part_properties, self._project_options,
+            plugin_manifest, machine_manifest))
+
+    def _get_machine_manifest(self):
+        return {'uname': common.run_output(['uname', '-a'])}
```
kalikiana: I'm wondering how flexible this is going to be in the future. Would it be worth going for separate fields like `kernel-name`, `kernel-version` and so on?

snapcraft/tests/test_lifecycle.py
elopio: uname -a outputs something like this:
Linux juju-44eb60-remote-devel-0 4.10.0-32-generic #36~16.04.1-Ubuntu SMP Wed Aug 9 09:19:02 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

I would need to put the same string on every test here. So, shorter is better. And in many tests, we have "test-x" as the fake value for x.

snapcraft/internal/pluginhandler/__init__.py
elopio: I asked the security team in the forum if uname -a was appropriate. No reply so far, but pretty easy to change later.

snapcraft/tests/test_lifecycle.py
kalikiana: I said "something that resembles" for that reason.

My point was, `test uname` looks like a shell command. "test-x" doesn't look like a command.

elopio: ok, I've replaced it with Linux test uname 4.10 x86_64.

integration_tests/test_asset_recording.py
```diff
@@ -55,6 +55,20 @@ def test_prime_records_snapcraft_yaml_copy(self):
...
+    def test_prime_records_uname(self):
+        self.run_snapcraft('prime', project_dir='basic')
+
+        recorded_yaml_path = os.path.join(
+            self.prime_dir, 'snap', 'manifest.yaml')
+        with open(recorded_yaml_path) as recorded_yaml_file:
+            recorded_yaml = yaml.load(recorded_yaml_file)
+
+        expected_uname = subprocess.check_output(
+            ['uname', '-a']).decode('utf-8').strip()
```
kyrofa: We've started a pattern of using `.decode(sys.getfilesystemencoding())` for this type of thing. Should we be using that here as well?

snapcraft/internal/pluginhandler/__init__.py
kyrofa: Having the hostname in there is somewhat identifying, and seemingly unnecessary to track. I suggest we use e.g. `uname -srvmpio` instead.

elopio:
Alright, I've pushed changes with your suggestions.

@kalikiana is there anything else still holding your +1?

sergiusens:
If this is part of #1452 mind updating the commit, look at the latest from @kyrofa in master for syntax

elopio:
@sergiusens amended the commit.

PR #1529, opened 2017-09-04
recording: record the packages installed in the machine

### LINKED ISSUE #1452: machine information

elopio:

-----
This closes #1452.
Requires #1528.

### DIFF
--- a/integration_tests/test_asset_recording.py
+++ b/integration_tests/test_asset_recording.py
@@ -20,9 +20,10 @@ import subprocess
 import sys
 import yaml
 
+import apt
 import fixtures
 import testscenarios
-from testtools.matchers import Equals
+from testtools.matchers import Contains, Equals
 
 import snapcraft
 import integration_tests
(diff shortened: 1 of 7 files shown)


integration_tests/test_asset_recording.py
```diff
@@ -55,6 +56,35 @@ def test_prime_records_snapcraft_yaml_copy(self):
...
+        self.run_snapcraft('prime', project_dir='basic')
+
+        recorded_yaml_path = os.path.join(
+            self.prime_dir, 'snap', 'manifest.yaml')
+        with open(recorded_yaml_path) as recorded_yaml_file:
+            recorded_yaml = yaml.load(recorded_yaml_file)
+
+        with apt.Cache() as apt_cache:
+            expected_package = 'python3={}'.format(
+                apt_cache['python3'].installed.version)
```
kalikiana: How about checking repo.get_installed_packages as well? My gut feeling is we should also explicitly test that it works correctly (even if the manifest uses it indirectly) - or if not here, in a unit test.

snapcraft/tests/fixture_setup.py
```diff
@@ -754,16 +753,20 @@ def setUp(self):
                     'internal', 'repo', 'manifest.txt'))) as manifest_file:
             self.add_packages([line.strip() for line in manifest_file])
 
+    def add_package(self, package):
+        package.temp_dir = self.path
+        self.cache[package.name] = package
```
kalikiana: Can you create the FakeaptCachePackage here instead of every call to the function?

integration_tests/test_asset_recording.py
elopio: I added a unit test in snapcraft/tests/repo/test_deb.py

snapcraft/tests/fixture_setup.py
elopio: I wouldn't want the add_package method to receive all the arguments that FakeAptCachePackage. If you want to just add a package by name you can `self.add_packages([test])`. If you need more control than the name, I think it's ok that you instantiate your own.

Maybe that other method should be called `add_packages_by_name`.

sergiusens:
@elopio what is the last part of #1452? Doesn't this fix it completely?

PR #1532, opened 2017-09-06
tests: update the meson test to latest meson requirements

sergiusens:
The latest meson does not allow targets to be prefixed with
`meson-`.

-----


kyrofa:
PR looks fine, although aren't these demo tests being run in the nightly? How did we miss it until now?

sergiusens:
@kyrofa because the nightly is something we all need to wake up to and look at.
This may also be because this only manifests itself in adt from package building which enables proposed, I don't think we want to deal with that though.

PR #1539, opened 2017-09-08
nodejs plugin: expose hidden bin path when using yarn

sergiusens:
Expose the path for `yarn run` to find the correct binaries
given that it is now not exposed to the bin directory in the
part's install directory.

-----

### DIFF
--- a/integration_tests/plugins/test_nodejs_plugin.py
+++ b/integration_tests/plugins/test_nodejs_plugin.py
@@ -60,3 +60,9 @@ class NodeJSPluginTestCase(testscenarios.WithScenarios,
                         FileExists())
         self.assertThat(os.path.join(part_builddir, 'command-two-run'),
                         FileExists())
+
+        # Ensure the bin entry makes it to bin in the part's install directory
+        part_installdir = os.path.join(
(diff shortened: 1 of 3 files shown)


sergiusens:
One comment here, this only fixes the tests. We need to research more into figuring out what other potential issues the new yarn brings and fix that with proper integration tests.

kyrofa:
I'm just using our shout demo to test this, and neither master nor this PR succeed in building it, both failing during the `prime` step with: `FileNotFoundError: [Errno 2] No such file or directory: '/home/ubuntu/snapcraft/demos/shout/prime/bin/shout'`. Indeed, `shout` is now in `node_modules/.bin/shout`. Should this actually be put into the plugin's `env()`, then?

elopio:
Yes, the problem is not that we don't find bins during build. That's a separate problem: https://bugs.launchpad.net/snapcraft/+bug/1685551

The problem here is that the bin is installed to `prime/node_modules/.bin`, and we are searching for it in `bin` in the app command.

kyrofa:
Verified with the new push: the shout demo can again be snapped.

Issue #1444, opened 2017-08-08
catkin pip support


sergiusens:
[design](https://forum.snapcraft.io/t/proposal-catkin-pip-support/2083)

sergiusens:
#1479 and LP: #1683036 make progress on by adding the ability for rosdep to parse multiple dependency types. It also updates the catkin plugin to support the new API, but it still only supports debs right now (it'll raise an exception if the dependency type is anything else). This is paving the way for support of pip dependencies coming from rosdep. Actually supporting this will require some experimentation, and of course, a design proposal.

kyrofa:
More progress in #1540, where some helpers are being moved out of the Python plugin into a place where Catkin can also use them.

kyrofa:
After the pip extraction lands, there's only one more PR as part of this. I expect this feature to be done this week.

PR #1540, opened 2017-09-08
plugins: extract python finder functions

kyrofa:

-----

This PR is the first in a series that extracts pip from the python plugin, thus making progress on LP: [#1683036](https://bugs.launchpad.net/snapcraft/+bug/1683036). It makes no changes to the Python plugin just yet, just copies functionality out into a more accessible place.

### DIFF
--- a/snapcraft/plugins/_python/_python_finder.py
+++ b/snapcraft/plugins/_python/_python_finder.py
@@ -1,0 +1,78 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2017 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+import glob
+
+from . import errors
+
+
+def get_python_command(python_major_version, *, stage_dir, install_dir):
+    """Find the python command to use, preferring staged over the part.
+
+    We prefer the staged python as opposed to the in-part python in order to
+    support one part that supplies python, with another part built `after` it
+    wanting to use its python.
+
+    :param str python_major_version: The python major version to find (2 or 3)
+    :param str stage_dir: Path to the staging area
+    :param str install_dir: Path to the part's install area
+
+    :return: Path to the python command that was found
+    :rtype: str
+
+    :raises MissingPythonCommandError: If no python could be found in the
+                                       staging or part's install area.
+    """
+    python_command_name = 'python{}'.format(python_major_version)
+    python_command = os.path.join('usr', 'bin', python_command_name)
+    staged_python = os.path.join(stage_dir, python_command)
+    part_python = os.path.join(install_dir, python_command)
+
+    if os.path.exists(staged_python):
+        return staged_python
+    elif os.path.exists(part_python):
+        return part_python
+    else:
+        raise errors.MissingPythonCommandError(
+            python_command_name, [stage_dir, install_dir])
+
+
+def get_python_headers(python_major_version, *, stage_dir):
+    """Find the python headers to use, if any, preferring staged over the host.
+
+    We want to make sure we use the headers from the staging area if available,
+    or we may end up building for an older python version than the one we
+    actually want to use.
+
+    :param str python_major_version: The python version to find (2 or 3)
+    :param str stage_dir: Path to the staging area
+
+    :return: Path to the python headers that were found ('' if none)
+    :rtype: str
+    """
+    python_command_name = 'python{}'.format(python_major_version)
+    base_match = os.path.join('usr', 'include', '{}*'.format(
+        python_command_name))
+    staged_python = glob.glob(os.path.join(stage_dir, base_match))
+    host_python = glob.glob(os.path.join(os.path.sep, base_match))
+
+    if staged_python:
+        return staged_python[0]
+    elif host_python:
+        return host_python[0]
+    else:
+        return ''
--- a/snapcraft/tests/plugins/python/test_python_finder.py
+++ b/snapcraft/tests/plugins/python/test_python_finder.py
@@ -1,0 +1,181 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2017 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+import re
+
+from unittest import mock
+from testtools.matchers import (
+    Equals,
+    MatchesRegex,
+)
+
+from snapcraft.plugins._python import (
+    errors,
+    _python_finder,
+)
+
+from snapcraft import (
+    tests
+)
+
+
+def _create_python_binary(base_dir):
+    python_command_path = os.path.join(
+        base_dir, 'usr', 'bin', 'pythontest')
+    os.makedirs(os.path.dirname(python_command_path))
+    open(python_command_path, 'w').close()
+
+
+class GetPythonCommandTestCase(tests.TestCase):
+
+    def test_staged(self):
+        """get_python_command should support staged python"""
+
+        stage_dir = 'stage_dir'
+        install_dir = 'install_dir'
+
+        # Create staged binary
+        _create_python_binary(stage_dir)
+
+        python_command = _python_finder.get_python_command(
+            'test', stage_dir=stage_dir, install_dir=install_dir)
+        self.assertThat(
+            python_command, Equals(os.path.join(
+                stage_dir, 'usr', 'bin', 'pythontest')))
+
+    def test_in_part(self):
+        """get_python_command should support in-part python"""
+
+        stage_dir = 'stage_dir'
+        install_dir = 'install_dir'
+
+        # Create installed binary
+        _create_python_binary(install_dir)
+
+        python_command = _python_finder.get_python_command(
+            'test', stage_dir=stage_dir, install_dir=install_dir)
+        self.assertThat(
+            python_command, Equals(os.path.join(
+                install_dir, 'usr', 'bin', 'pythontest')))
+
+    def test_staged_and_in_part(self):
+        """get_python_command should prefer staged python over in-part"""
+
+        stage_dir = 'stage_dir'
+        install_dir = 'install_dir'
+
+        # Create both staged and installed binaries
+        _create_python_binary(stage_dir)
+        _create_python_binary(install_dir)
+
+        python_command = _python_finder.get_python_command(
+            'test', stage_dir=stage_dir, install_dir=install_dir)
+        self.assertThat(
+            python_command, Equals(os.path.join(
+                stage_dir, 'usr', 'bin', 'pythontest')))
+
+    def test_missing_raises(self):
+        """get_python_command should raise if no python can be found"""
+
+        stage_dir = 'stage_dir'
+        install_dir = 'install_dir'
+
+        raised = self.assertRaises(
+            errors.MissingPythonCommandError,
+            _python_finder.get_python_command, 'test', stage_dir=stage_dir,
+            install_dir=install_dir)
+        self.assertThat(str(raised), Equals(
+            'Unable to find pythontest, searched: stage_dir:install_dir'))
+
+
+class GetPythonHeadersTestCase(tests.TestCase):
+
+    def setUp(self):
+        super().setUp()
+
+        patcher = mock.patch('glob.glob')
+        self.mock_glob = patcher.start()
+        self.addCleanup(patcher.stop)
+
+    def test_staged(self):
+        """get_python_headers should support staged headers"""
+
+        stage_dir = 'stage_dir'
+
+        # Fake out glob so it looks like the headers are in the staging area
+        def _fake_glob(pattern):
+            if pattern.startswith(stage_dir):
+                return [
+                    os.path.join(stage_dir, 'usr', 'include', 'pythontest')]
+            return []
+        self.mock_glob.side_effect = _fake_glob
+
+        self.assertThat(
+            _python_finder.get_python_headers('test', stage_dir=stage_dir),
+            Equals(os.path.join(stage_dir, 'usr', 'include', 'pythontest')))
+
+    def test_host(self):
+        """get_python_headers should support the hosts's headers"""
+
+        stage_dir = 'stage_dir'
+
+        # Fake out glob so it looks like the headers are installed on the host
+        def _fake_glob(pattern):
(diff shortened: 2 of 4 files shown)


snapcraft/plugins/_python/_python_finder.py
```diff
@@ -0,0 +1,44 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+import glob
+
+
+def get_python_command(python_version, stage_dir, install_dir):
```
elopio: can you please add docstrings to these two functions?

snapcraft/plugins/_python/_python_finder.py
```diff
@@ -0,0 +1,44 @@
...
+    python_command = os.path.join('usr', 'bin', python_version)
+
+    # staged as in the stage dir, not from stage-packages
+    staged_python = os.path.join(stage_dir, python_command)
+    unstaged_python = os.path.join(install_dir, python_command)
+
+    if os.path.exists(staged_python):
+        return staged_python
+    else:
+        return unstaged_python
```
elopio: what about having a safeguard here in case unstaged python doesn't exist? like in the next func.

snapcraft/plugins/_python/_python_finder.py
```diff
@@ -0,0 +1,44 @@
...
+    base_match = os.path.join('usr', 'include', '{}*'.format(python_version))
+    unstaged_python = glob.glob(os.path.join(os.path.sep, base_match))
+    staged_python = glob.glob(os.path.join(stage_dir, base_match))
+
+    if staged_python:
+        return staged_python[0]
+    elif unstaged_python:
+        return unstaged_python[0]
+    else:
+        return ''
```
elopio: Wouldn't it be more pythonic to raise an exception?

snapcraft/tests/plugins/python/test_python_finder.py
```diff
@@ -0,0 +1,114 @@
...
+
+def _create_python_binary(base_dir):
+    python_command_path = os.path.join(
+        base_dir, 'usr', 'bin', 'python2')
+    os.makedirs(os.path.dirname(python_command_path))
+    open(python_command_path, 'w').close()
+
+
+def _create_python_headers(base_dir):
+    os.makedirs(os.path.join(base_dir, 'usr', 'include', 'python2'))
```
elopio: I prefer to put the helpers at the start of the file the things that are global. I thought that was pep8, but can't find a specific mention to this. Why do you like them at the bottom?

snapcraft/tests/plugins/python/test_python_finder.py
```diff
@@ -0,0 +1,114 @@
...
+        install_dir = 'install_dir'
+
+        # Create installed binary
+        _create_python_binary(install_dir)
+
+        python_command = _python_finder.get_python_command(
+            'python2', stage_dir, install_dir)
+        self.assertThat(
+            python_command, Equals(os.path.join(
+                install_dir, 'usr', 'bin', 'python2')))
```
elopio: A nit, but I would use 'python3' in all our test data, even if the name is not really relevant.
Or maybe, make it clearer that the name is not relevant by naming it `test-python-version`.

snapcraft/plugins/_python/_python_finder.py
```diff
@@ -0,0 +1,44 @@
...
+    if os.path.exists(staged_python):
+        return staged_python
+    else:
+        return unstaged_python
+
+
+def get_python_headers(python_version, stage_dir):
+    base_match = os.path.join('usr', 'include', '{}*'.format(python_version))
+    unstaged_python = glob.glob(os.path.join(os.path.sep, base_match))
+    staged_python = glob.glob(os.path.join(stage_dir, base_match))
```
elopio: nit: revert the order of these two declarations so they match the order in the if statements.

snapcraft/tests/plugins/python/test_python_finder.py
```diff
@@ -0,0 +1,114 @@
...
+        # Create staged binary
+        _create_python_binary(stage_dir)
+
+        python_command = _python_finder.get_python_command(
+            'python2', stage_dir, install_dir)
+        self.assertThat(
+            python_command, Equals(os.path.join(
+                stage_dir, 'usr', 'bin', 'python2')))
+
+    def test_installed(self):
```
elopio: Here you are using the word `installed`. I personally prefer that one over unstaged. Why not use it everywhere?

snapcraft/tests/plugins/python/test_python_finder.py
```diff
@@ -0,0 +1,114 @@
...
+        _create_python_binary(install_dir)
+
+        python_command = _python_finder.get_python_command(
+            'python2', stage_dir, install_dir)
+        self.assertThat(
+            python_command, Equals(os.path.join(
+                stage_dir, 'usr', 'bin', 'python2')))
+
+
+class GetPythonHeadersTestCase(tests.TestCase):
```
elopio: no test for when both exist and staged is preferred?
no test for when '' is returned?

kyrofa:
> I'm sorry, I left a lot of nit comments. The only real thing I want to see are docstrings.

Ha! Those are my favorite, it means I don't need to change much ;) . Thanks for the thorough review!

snapcraft/plugins/_python/_python_finder.py
kyrofa: Yes, definitely. We should try to start making that a habit, at least on public functions. Agreed?

elopio: I'm +1. Please point it out when I forget it.

snapcraft/tests/plugins/python/test_python_finder.py
kyrofa: Haha, you'll love this explanation: `installed` means "in the part." `unstaged` means actually installed on the host. Partway through this refactor I switched to using that language, I'll make that change here as well.

snapcraft/tests/plugins/python/test_python_finder.py
kyrofa: The first test actually exists, because there's no way to NOT have the headers installed on the host for these tests. Which also means I don't know of a way to test the `''` case. I can probably mock `os.exists` or something, though, but that sounds a little anti-Leo. Thoughts?

snapcraft/tests/plugins/python/test_python_finder.py
kyrofa: You're right, I'm sorry, I keep forgetting about this. I like them at the bottom because it seems inherently natural that the public functions are more interesting to the reader, thus should be closer to the top. It's like designing a web-page: put the stuff people want to see front and center. But I'm happy to switch the order.

snapcraft/plugins/_python/_python_finder.py
kyrofa: You got it.

snapcraft/plugins/_python/_python_finder.py
kyrofa: I would agree if it was actually an error, but it's not unexpected in the code. If I raised an exception here, the usage of this function would look like:

```
try:
    headers = get_python_headers([...])
except MissingPythonHeadersError:
    pass
else:
    # Do something with the headers
```

Right now it's just an if, which feels a little cleaner.

snapcraft/tests/plugins/python/test_python_finder.py
kyrofa: Ha, and it's not even consistent here either. How dreadful.

snapcraft/plugins/_python/_python_finder.py
kyrofa: Yeah the usage of this function is a little odd, and perhaps it doesn't make as much sense in isolation. Check [this](https://github.com/snapcore/snapcraft/blob/master/snapcraft/plugins/python.py#L162) out, which uses it before any stage packages are pulled (i.e. before the part's python has even been fetched). This function is has two purposes: to figure out a python command to use, yes, but also to simply determine which area WILL contain python: staging, or the part. The latter assumption is the plugin's leaking into this module. If we modify this function to simply be the former (find an existing python), the Python plugin can catch the exception and make whatever assumptions it wants to make, which is a better separation of concerns, I think.

snapcraft/tests/plugins/python/test_python_finder.py
kyrofa: Done.

snapcraft/tests/plugins/python/test_python_finder.py
kyrofa: Fixed.

snapcraft/plugins/_python/_python_finder.py
kyrofa: Sure thing, done.

snapcraft/plugins/_python/_python_finder.py
kyrofa: Done.

snapcraft/plugins/_python/_python_finder.py
kyrofa: Done.

snapcraft/tests/plugins/python/test_python_finder.py
kyrofa: I went with `python3` so I could continue testing the headers on the host.

snapcraft/tests/plugins/python/test_python_finder.py
elopio: I vote for anti-Leo, mocking os.exists.
But, now I found a middle-ground, like in my proposal to mock pip. You can mock os.exists only if it receives the argument you are sending, and use te real os.exists in all the other cases. I'm starting to like that pattern, please let me know your thoughts because I think I'll start using it everywhere.

snapcraft/plugins/_python/_python_finder.py
```diff
@@ -0,0 +1,69 @@
...
+    """Find the python command to use, preferring staged over the part.
+
+    :param str python_version: The python version to find (python2 or python3)
+    :param str stage_dir: Path to the staging area
+    :param str install_dir: Path to the part's install area
+
+    :return: Path to the python command that was found
+    :rtype: str
+
+    :raises MissingPythonCommandError: If no python could be find in the
```
elopio: isn't it: "could be found"? It's scary when I get english right and you don't :p

snapcraft/plugins/_python/_python_finder.py
```diff
@@ -0,0 +1,69 @@
...
+                                       staging or part's install area.
+    """
+
```
elopio: I think we need to agree on this empty line. I find it unnecessary.

snapcraft/tests/plugins/python/test_python_finder.py
elopio: That's what had me confused. I like that rule of more public first, and then group the helpers after they are used. Maybe, they removed the pep8 part that talked about organizing the functions in the module, and now I have an internal conflict :)

elopio:
Thanks for all the changes and replies. I would like to see that test for the '' case, but not a big deal. I'll +1 this now and let you decide if you want to add it.

snapcraft/plugins/_python/_python_finder.py
kyrofa: Yikes. Say it with a strong southern accent.

snapcraft/plugins/_python/_python_finder.py
kyrofa: I find it annoying on one-liner dockstrings, but unnecessary on multi-line. No strong preference, I'll remove them.

snapcraft/tests/plugins/python/test_python_finder.py
kyrofa: Anti-Leo, done. That should be completely covered in tests, now.

snapcraft/plugins/_python/_python_finder.py
kyrofa: Fixed.

snapcraft/plugins/_python/_python_finder.py
kyrofa: Done.

snapcraft/tests/plugins/python/test_python_finder.py
kyrofa: Ha! Well, think on it. Super easy to flip around.

kyrofa:
> I would like to see that test for the '' case, but not a big deal.

Done, thanks man :smile: .

snapcraft/plugins/_python/_python_finder.py
```diff
@@ -0,0 +1,67 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+import glob
+
+from . import errors
+
+
+def get_python_command(python_version, stage_dir, install_dir):
```
sergiusens: how about

```python
def get_python_command(python_major_version, *, stage_dir, install_dir)
```

Where `python_major_version` can be 2 or 3? You can expand the expected command from a dict if needed.

snapcraft/plugins/_python/_python_finder.py
```diff
@@ -0,0 +1,67 @@
...
+    if os.path.exists(staged_python):
+        return staged_python
+    elif os.path.exists(part_python):
+        return part_python
+    else:
+        raise errors.MissingPythonCommandError(
+            python_version, [stage_dir, install_dir])
+
+
+def get_python_headers(python_version, stage_dir):
```
sergiusens:
```python
def get_python_headers(python_version, stage_dir):
```

to

```python
def get_python_headers(python_major_version, *, preferred_dir):
```

snapcraft/plugins/_python/_python_finder.py
sergiusens: how about

```python
def get_python_command(python_major_version, *, stage_dir, install_dir)
```

Where `python_major_version` can be 2 or 3? You can expand the expected command from a dict if needed.

snapcraft/plugins/_python/_python_finder.py
sergiusens: Although to keep consistency, it should still be `stage_dir`

snapcraft/tests/plugins/python/test_python_finder.py
```diff
@@ -0,0 +1,181 @@
...
+        _create_python_binary(install_dir)
+
+        python_command = _python_finder.get_python_command(
+            'python-test', stage_dir, install_dir)
+        self.assertThat(
+            python_command, Equals(os.path.join(
+                install_dir, 'usr', 'bin', 'python-test')))
+
+    def test_staged_and_in_part(self):
+        """get_python_command should prefer staged python over in-part"""
```
sergiusens: The why would be good to explain here.

kyrofa:
> I take that the next PR migrates to using this, right?

Partially. I've tried to take the path of easiest review. There are three PRs (of which this is the first) where functionality is only copied from the Python plugin. The next PR will indeed use this, but it itself will not be used. On the fourth PR, in one go the Python plugin is updated to use this. That's the one that will need heavy testing. In parallel with that review, Catkin support can also be added.

snapcraft/plugins/_python/_python_finder.py
kyrofa: There would be some gymnastics in the Python plugin to turn `python2` into `2` and `python3` into `3` (not difficult, of course, but unfortunate). It also then feels like this function should support both strings and ints. I'll think on this.

kyrofa: Yeah, +1 on this. It's just an unclear API otherwise.

snapcraft/tests/plugins/python/test_python_finder.py
kyrofa: I added a blurb to the finder functions instead of the tests.

snapcraft/plugins/_python/_python_finder.py
kyrofa: Done.

kyrofa:
Squashed.

sergiusens:
@kyrofa aren't you going to reference #1444 in the PR?

kyrofa:
> @kyrofa aren't you going to reference #1444 in the PR?

Ah right-- put it in the commit.

kyrofa:
All green here.

PR #1541, opened 2017-09-08
tests: fix the TEST_STORE environment variable 

elopio:

-----
The environment variable is passed to the containers as ''. The setup of integration tests takes care of translating that to 'fake', but some tests are calling the skip before the setup, so weird things happen.

This is a patch for now. The right solution is to not skip any tests, which is almost always doable.

### DIFF
--- a/integration_tests/__init__.py
+++ b/integration_tests/__init__.py
@@ -366,6 +366,12 @@ class StoreTestCase(TestCase):
         self.test_store = fixture_setup.TestStore()
         self.useFixture(self.test_store)
 
+    def is_store_fake(self):
+        return (os.getenv('TEST_STORE') or 'fake') == 'fake'
+
+    def is_store_staging(self):
+        return os.getenv('TEST_STORE') == 'staging'
+
     def login(self, email=None, password=None, expect_success=True):
         email = email or self.test_store.user_email
         password = password or self.test_store.user_password
--- a/integration_tests/store/test_store_revisions.py
+++ b/integration_tests/store/test_store_revisions.py
@@ -14,10 +14,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import os
 import re
 import subprocess
-import unittest
 
 from testtools.matchers import Contains, Equals, FileExists, MatchesRegex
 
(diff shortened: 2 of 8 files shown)


elopio:
@sergiusens sounds good. But I don't really like skip decorators. Any complaints if I turn all of them into calls to the is_store_fake method?

integration_tests/__init__.py
```diff
@@ -366,6 +366,12 @@ def setUp(self):
         self.test_store = fixture_setup.TestStore()
         self.useFixture(self.test_store)
 
+    def is_store_fake(self):
+        return (os.getenv('TEST_STORE') or 'fake') == 'fake'
```
kyrofa: This causes a bit of a double-take to determine that you actually want to default to the fake store. I'd prefer using a more explicit:

```python
def is_store_fake(self):
    test_store = os.getenv('TEST_STORE')
    if test_store and test_store != 'fake':
        return False
    return True
```

kyrofa: Ah heck, nevermind. That's no more clear.

kyrofa: `os.getenv('TEST_STORE', 'fake')` would look nicer, I think.

integration_tests/store/test_store_revisions.py
```diff
@@ -64,9 +62,10 @@ def test_revisions_with_login_bad_snap_with_arch(self):
             "Snap 'mysnap' for 'i386' was not found in '16' series.",
             str(error.output))
 
-    @unittest.skipUnless(
-        os.getenv('TEST_STORE', 'fake') == 'fake', 'Skip fake store.')
     def test_revisions_fake_store(self):
+        if not self.is_store_fake():
```
sergiusens: if these 3 tests go into their on `FakeStoreRevisionsTestCase` (or whatever better name) you can just do this in `setUp`

sergiusens:
squash time! (with a maybe a prior knitting session)

integration_tests/__init__.py
kalikiana: This was part of the bug. It would return '' if TEST_STORE is set to ''. Perhaps counter intuitively, getenv only uses the default if a variable is unset.

kyrofa: Ah interesting, I suppose that makes sense. Ignore me, then.

PR #1542, opened 2017-09-08
tests: add integration tests for build snaps

elopio:

-----
There are two missing scenarios because of bugs:
https://bugs.launchpad.net/snapcraft/+bug/1716054
https://bugs.launchpad.net/snapcraft/+bug/1716055

### DIFF
--- a/snapcraft/tests/fixture_setup.py
+++ b/snapcraft/tests/fixture_setup.py
@@ -20,6 +20,7 @@ import copy
 import io
 import os
 import string
+import subprocess
 import sys
 import threading
 import urllib.parse
@@ -31,6 +32,7 @@ from subprocess import CalledProcessError
 
 import fixtures
 import xdg
+import yaml
 
 import snapcraft
 from snapcraft.tests import fake_servers
@@ -905,3 +907,60 @@ class FakeAptCachePackage():
 
     def get_dependencies(self, _):
         return []
+
+
+class WithoutSnapInstalled(fixtures.Fixture):
+    """Assert that a snap is not installed and remove it on clean up.
+
+    :raises: AssertionError: if the snap is installed when this fixture is
+        set up.
+    """
+
+    def __init__(self, snap_name):
+        super().__init__()
+        self.snap_name = snap_name.split('/')[0]
+
+    def setUp(self):
+        super().setUp()
+        if snapcraft.repo.snaps.SnapPackage.is_snap_installed(self.snap_name):
+            raise AssertionError(
+                "This test cannot run if you already have the {snap!r} snap "
+                "installed. Please uninstall it by running "
+                "'sudo snap remove {snap}'.".format(snap=self.snap_name))
+
+        self.addCleanup(self._remove_snap)
+
+    def _remove_snap(self):
+        try:
+            subprocess.check_output(
+                ['sudo', 'snap', 'remove', self.snap_name],
+                stderr=subprocess.STDOUT)
+        except subprocess.CalledProcessError as e:
+            RuntimeError("unable to remove {!r}: {}".format(
+                self.snap_name, e.output))
+
+
+class SnapcraftYaml(fixtures.Fixture):
+
+    def __init__(
+            self, path, name='test-snap', version='test-version',
+            summary='test-summary', description='test-description'):
+        super().__init__()
+        self.path = path
+        self.data = {
+            'name': name,
+            'version': version,
+            'summary': summary,
+            'description': description,
+            'parts': {}
+        }
+
+    def update_part(self, name, data):
+        part = {name: data}
+        self.data['parts'].update(part)
+
+    def setUp(self):
+        super().setUp()
+        with open(os.path.join(self.path, 'snapcraft.yaml'),
+                  'w') as snapcraft_yaml_file:
+            yaml.dump(self.data, snapcraft_yaml_file)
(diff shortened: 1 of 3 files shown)


snapcraft/tests/fixture_setup.py
```diff
@@ -875,3 +877,57 @@ def fetch_binary(self, dir_, progress):
 
     def get_dependencies(self, _):
         return []
+
+
+class WithSnapNotInstalled(fixtures.Fixture):
```
kalikiana: My first guess from the name was that it's mocking that the snap is not installed, cf. `FakeSnapd`. Seeing that it's actually asserting that the snap isn't installed yet, it would obviously have to be an `assertSnapNotInstalled` rather than a fixture. Then I realized it's also removing the snap in the cleanup, which explains why you went for a fixture... to my mind, `SnapInstalledDuringTheTest` would convey better what it's actually doing.

snapcraft/tests/fixture_setup.py
```diff
@@ -875,3 +877,57 @@ def fetch_binary(self, dir_, progress):
...
+        self.data = {
+            'name': name,
+            'version': version,
+            'summary': summary,
+            'description': description,
+        }
+
+    def add_part(self, name, data):
+        part = {name: data}
+        if 'parts' in self.data:
```
kalikiana: The smarts to update an existing part aren't obvious from the name. It would be clearer to my mind to have an explicit `extend_part` method.

snapcraft/tests/fixture_setup.py
sergiusens: I don't like either, but prefer @elopio's... @kyrofa you are the tie breaker!

snapcraft/tests/fixture_setup.py
sergiusens: I would go the opposite path, no where in this code base is there a need to update a part and given it comes from a clean fixture we should raise an error if the part name is already added.

snapcraft/tests/fixture_setup.py
```diff
@@ -875,3 +877,57 @@ def fetch_binary(self, dir_, progress):
...
+        super().__init__()
+        self.path = path
+        self.data = {
+            'name': name,
+            'version': version,
+            'summary': summary,
+            'description': description,
+        }
+
+    def add_part(self, name, data):
```
sergiusens: Add a `*` after `self`.

snapcraft/tests/fixture_setup.py
kyrofa: Yeah there's a lot going on here that's difficult to convey in a simple class name. Where's my docstring?!

Let me think this through:

- This fixture doesn't care that the snap was actually installed during the test. Indeed, `snap remove nonexisting-snap` exits 0. As such, `SnapInstalledDuringTheTest` does not make sense.
- This fixture DOES care if the snap is installed before the test by failing in that case. So @elopio's direction is the right one.

So the real question is: which is more important? The assertion at the beginning, or the cleanup step? I think it's more important that it fails is the snap was already installed, because otherwise the test isn't testing the right thing. Perhaps `WithoutSnapInstalled` would be better, though?

snapcraft/tests/fixture_setup.py
kyrofa: Maybe I'm missing an edit here, but this doesn't handle the case of when a specific part already exists, but merely when _a_ part exists (in which case we want to add another, not blow away the one that already exists).

snapcraft/tests/fixture_setup.py
```diff
@@ -875,3 +877,57 @@ def fetch_binary(self, dir_, progress):
...
+        part = {name: data}
+        if 'parts' in self.data:
+            self.data['parts'].update(part)
+        else:
+            self.data['parts'] = part
```
kyrofa: Wait... shouldn't this be a list of dicts?

snapcraft/tests/fixture_setup.py
kalikiana: I can't parse your phrasing... I'd expect `add_part("foo", {})` called twice to be an error. If there's a use case for modifying an existing part 'foo' it should be separate from this method.

elopio: I've renamed this to update_part and copied the behaviour of a dict.

snapcraft/tests/fixture_setup.py
elopio: @sergiusens I'm trying to understand the rule we are following for named arguments.
`add_part('my-part', {plugin: 'nil'}` seems clear to me.

`add_part(name='my-part', data={plugin: 'nil'}` doesn't add much, IMO. I'm not against adding it, not a big issue to me. I just want to understand when I should and when I shouldn't.

snapcraft/tests/fixture_setup.py
```diff
@@ -905,3 +907,60 @@ def fetch_binary(self, dir_, progress):
...
+        self.path = path
+        self.data = {
+            'name': name,
+            'version': version,
+            'summary': summary,
+            'description': description,
+            'parts': {}
+        }
+
+    def update_part(self, name, data):
```
sergiusens: wait, this now updates the whole of the `parts` definition, is this what you wanted?

elopio: no, I made a mistake. Thanks for catching it.
I want to call `update` in the `parts` item. Replace a part if it exists, create it if it doesn't.

PR #1412, opened 2017-07-17
lxd: snapcraft refresh in containers

kalikiana:
- Only update apt cache and install snapcraft once
- Allow snapcraft update to upgrade packages and snaps

See the [forum proposal](https://forum.snapcraft.io/t/update-persistent-lxd-containers/1648/3).

**Note**: The other part is implemented in #1546.

PR #1546, opened 2017-09-12
cli: update parts cache in the container

kalikiana:

-----

- Pass snapcraft update through to the container
See the [forum proposal](https://forum.snapcraft.io/t/update-persistent-lxd-containers/1648/3).

**Note**: The other part is implemented in #1412.


snapcraft/tests/commands/test_update.py
```diff
@@ -106,3 +132,28 @@ def test_update_with_no_content_length_is_supported(self):
...
+                                                 mock_container_run):
+        mock_container_run.side_effect = lambda cmd, **kwargs: cmd
+        mock_getuid.return_value = 1234
+        fake_lxd = fixture_setup.FakeLXD()
+        self.useFixture(fake_lxd)
+        # Container was created before and is running
+        fake_lxd.name = 'local:snapcraft-snap-test'
+        fake_lxd.status = 'Running'
+        fake_logger = fixtures.FakeLogger(level=logging.INFO)
+        self.useFixture(fake_logger)
```
sergiusens: I don't think you need this fake logger setup here; or what is it intended for as there are no checks for it. `self.run_command` would return an object with an `output` and `exit_code` attribute.

kalikiana: You're right! Thanks for pointing it out! I'd copied it because it's used in other tests, but actually the output is the same - we could probably simplify a few other tests this way at some point.
I dropped the logger and added a check for exit code (not sure if it's really necessary here, but can't hurt to have it).

snapcraft/tests/commands/test_update.py
```diff
@@ -20,19 +20,44 @@
...
 
 class UpdateCommandTestCase(CommandBaseTestCase, TestWithFakeRemoteParts):
 
+    yaml_template = dedent("""\
+        name: snap-test
+        version: 1.0
+        summary: test snapping
+        description: if snap is succesful a snap package will be available
+        architectures: ['amd64']
+        type: {}
```
kyrofa: When using an external template like this, named parameters are far more clear and scalable than positional ones.

kalikiana:
@kyle Note that the PR description explicitly mentions the part that this PR implements. This one is **update**. #1412 covers the refresh/ upgrade.

snapcraft/tests/commands/test_update.py
kalikiana: I just copied it from the snap test :-D but yeah, I changed it to {snap_type}

snapcraft/tests/commands/test_update.py
```diff
@@ -20,19 +20,44 @@
...
+                plugin: nil
+        """)
+
     def setUp(self):
         super().setUp()
         self.parts_dir = os.path.join(BaseDirectory.xdg_data_home, 'snapcraft')
         self.parts_yaml = os.path.join(self.parts_dir, 'parts.yaml')
         self.headers_yaml = os.path.join(self.parts_dir, 'headers.yaml')
 
+    def make_snapcraft_yaml(self, n=1, snap_type='app', snapcraft_yaml=None):
```
sergiusens: There is a `make_snapcraft_yaml` in the parent class for tests which you can just use, right?

kalikiana: I basically went with the style used in other tests there, which have their own customized version of the method for repeated use. But you're right, in this case I can keep it really simple. So, done.

kalikiana:
Tests are green again!

snapcraft/cli/parts.py
```diff
@@ -25,8 +28,13 @@ def partscli(ctx):
 
 @partscli.command()
 @click.pass_context
-def update(ctx):
+def update(ctx, **kwargs):
     """Updates the parts listing from the cloud."""
+    container_config = env.get_container_config()
+    if container_config.use_container:
+        project_options = get_project_options(**kwargs)
+        lifecycle.containerbuild('update', project_options, container_config)
```
elopio: This looks good to me, but I'm wondering why do we have to update the parts in the host?

kalikiana: Because `snapcraft define` and `snapcraft search` operate on the host. There's little reason to start a container for those... or do you think we should?

sergiusens:
On mar, oct 24, 2017 at 2:29 AM, Leo Arias <notifications@github.com> 
wrote:
> @elopio commented on this pull request.
> 
> In snapcraft/cli/parts.py:
> 
> >      """Updates the parts listing from the cloud."""
> +    container_config = env.get_container_config()
> +    if container_config.use_container:
> +        project_options = get_project_options(**kwargs)
> +        lifecycle.containerbuild('update', project_options, 
> container_config)
> This looks good to me, but I'm wondering why do we have to update the 
> parts in the host?
> 

If you are outside of a project directory these commands should still 
work, and as such define and search should probably always run from the 
context of the host while update needs to happen in both places.

A note of that ^ should be made in the code so it doesn't fall through 
the cracks in the future.

PR #1549, opened 2017-09-14
plugins: extract pip from python plugin

kyrofa:

-----

At long last, making more progress on LP: [#1683036](https://bugs.launchpad.net/snapcraft/+bug/1683036) and #1444, here is the final bit of extraction necessary from the Python plugin: the Pip wrapper itself.

Note that unlike the other PRs, this is a bit of a refactor. The previous pip wrapper was a very private implementation detail of the Python plugin, and didn't need much of an API. This PR introduces a bit more of an API to it in the hopes of making it a little easier to use in other plugins. It also makes it easier to test.

This PR doesn't modify the Python plugin in any way just yet (that's the next PR).

### DIFF
--- a/snapcraft/plugins/_python/__init__.py
+++ b/snapcraft/plugins/_python/__init__.py
@@ -14,5 +14,6 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+from ._pip import Pip  # noqa
 from ._python_finder import get_python_command  # noqa
 from ._sitecustomize import generate_sitecustomize  # noqa
--- a/snapcraft/plugins/_python/_pip.py
+++ b/snapcraft/plugins/_python/_pip.py
@@ -1,0 +1,349 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2016-2017 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import collections
+import contextlib
+import json
+import logging
+import os
+import shutil
+import subprocess
+import sys
+import tempfile
+
+import snapcraft
+from snapcraft import file_utils
+from ._python_finder import (
+    get_python_command,
+    get_python_headers,
+    get_python_home,
+)
+from . import errors
+
+logger = logging.getLogger(__name__)
+
+
+def _process_common_args(*, packages, constraints,
+                         requirements, process_dependency_links):
+    args = []
+    if constraints:
+        for constraint in constraints:
+            args.extend(['--constraint', constraint])
+
+    if requirements:
+        for requirement in requirements:
+            args.extend(['--requirement', requirement])
+
+    if process_dependency_links:
+        args.append('--process-dependency-links')
+
+    if packages:
+        args.extend(packages)
+
+    return args
+
+
+class Pip:
+    """Wrapper for pip abstracting the args necessary for use in a part.
+
+    This class takes care of fetching pip, setuptools, and wheel, and then
+    simply shells out to pip with the magical arguments necessary to install
+    packages into a part.
+
+    Of particular importance: packages must be downloaded (via download())
+    before they can be installed or have wheels built.
+    """
+
+    def __init__(self, *, python_major_version, part_dir, install_dir,
+                 stage_dir):
+        """Initialize pip.
+
+        Check to see if pip has already been installed. If not, fetch pip,
+        setuptools, and wheel, and install them so they can be used.
+
+        :param str python_major_version: The python major version to find (2 or
+                                         3)
+        :param str part_dir: Path to the part's working area
+        :param str install_dir: Path to the part's install area
+        :param str stage_dir: Path to the staging area
+
+        :raises MissingPythonCommandError: If no python could be found in the
+                                           staging or part's install area.
+        """
+        self._python_major_version = python_major_version
+        self._part_dir = part_dir
+        self._install_dir = install_dir
+        self._stage_dir = stage_dir
+
+        self._python_package_dir = os.path.join(
+            self._part_dir, 'python-packages')
+        os.makedirs(self._python_package_dir, exist_ok=True)
+
+        self._python_command = get_python_command(
+            self._python_major_version, stage_dir=self._stage_dir,
+            install_dir=self._install_dir)
+        self._python_home = get_python_home(
+            self._python_major_version, stage_dir=self._stage_dir,
+            install_dir=self._install_dir)
+
+        self._setup()
+
+    def _setup(self):
+        # Check to see if we have our own pip, yet. If not, we need to use the
+        # pip on the host (installed via build-packages) to grab our own.
+        if not self._is_pip_installed():
+            logger.info('Fetching pip, setuptools, and wheel...')
+
+            real_python_home = self._python_home
+
+            # Make sure we're using pip from the host. Wrapping this operation
+            # in a try/finally to make sure we revert away from the host's
+            # python at the end.
+            try:
+                self._python_home = os.path.join(os.path.sep, 'usr')
+
+                # Using the host's pip, install our own pip and other tools we
+                # need.
+                self.download({'pip', 'setuptools', 'wheel'})
+                self.install(
+                    {'pip', 'setuptools', 'wheel'}, ignore_installed=True)
+            finally:
+                # Now that we have our own pip, reset the python home
+                self._python_home = real_python_home
+
+    def _is_pip_installed(self):
+        try:
+            # We're expecting an error here at least once complaining about
+            # pip not being installed. In order to verify that the error is the
+            # one we think it is, we need to process the stderr. So we'll
+            # redirect it to stdout. If it's not the error we expect, something
+            # is wrong, so re-raise it.
+            self._run([], stderr=subprocess.STDOUT)
+        except subprocess.CalledProcessError as e:
+            output = e.output.decode(sys.getfilesystemencoding()).strip()
+            if 'no module named pip' in output.lower():
+                return False
+            else:
+                raise e
+
+        return True
+
+    def download(self, packages, *, setup_py_dir=None, constraints=None,
+                 requirements=None, process_dependency_links=False):
+        """Download packages into cache, but do not install them.
+
+        :param iterable packages: Packages to download from index.
+        :param str setup_py_dir: Directory containing setup.py.
+        :param iterable constraints: Collection of paths to constraints files.
+        :param iterable requirements: Collection of paths to requirements
+                                      files.
+        :param boolean process_dependency_links: Enable the processing of
+                                                 dependency links.
+        """
+        args = _process_common_args(
+            process_dependency_links=process_dependency_links,
+            packages=packages, constraints=constraints,
+            requirements=requirements)
+
+        cwd = None
+        if setup_py_dir:
+            args.append('.')
+            cwd = setup_py_dir
+
+        if not args:
+            return  # No operation was requested
+
+        # Using pip with a few special parameters:
+        #
+        # --disable-pip-version-check: Don't whine if pip is out-of-date with
+        #                              the version on pypi.
+        # --dest: Download packages into the directory we've set aside for it.
+        self._run(['download', '--disable-pip-version-check', '--dest',
+                   self._python_package_dir] + args, cwd=cwd)
+
+    def install(self, packages, *, setup_py_dir=None, constraints=None,
+                requirements=None, process_dependency_links=False,
+                upgrade=False, install_deps=True, ignore_installed=False):
+        """Install packages from cache.
+
+        The packages should have already been downloaded via `download()`.
+
+        :param iterable packages: Packages to install from cache.
+        :param str setup_py_dir: Directory containing setup.py.
+        :param iterable constraints: Collection of paths to constraints files.
+        :param iterable requirements: Collection of paths to requirements
+                                      files.
+        :param boolean process_dependency_links: Enable the processing of
+                                                 dependency links.
+        :param boolean upgrade: Recursively upgrade packages.
+        :param boolean install_deps: Install package dependencies.
+        :param boolean ignore_installed: Reinstall packages if they're already
+                                         installed
+        """
+        args = _process_common_args(
+            process_dependency_links=process_dependency_links,
+            packages=packages, constraints=constraints,
+            requirements=requirements)
+
+        if upgrade:
+            args.append('--upgrade')
+
+        if not install_deps:
+            args.append('--no-deps')
+
+        if ignore_installed:
+            args.append('--ignore-installed')
+
+        cwd = None
+        if setup_py_dir:
+            args.append('.')
+            cwd = setup_py_dir
+
+        if not args:
+            return  # No operation was requested
+
+        # Using pip with a few special parameters:
+        #
+        # --user: Install packages to PYTHONUSERBASE, which we've pointed to
+        #         the installdir.
+        # --no-compile: Don't compile .pyc files. FIXME: This is legacy, and
+        #               should be removed once this refactor has been
+        #               validated.
+        # --no-index: Don't hit pypi, assume the packages are already
+        #             downloaded (i.e. by using `self.download()`)
+        # --find-links: Provide the directory into which the packages should
+        #               have already been fetched
+        self._run(['install', '--user', '--no-compile', '--no-index',
+                   '--find-links', self._python_package_dir] + args, cwd=cwd)
+
+    def wheel(self, packages, *, setup_py_dir=None, constraints=None,
+              requirements=None, process_dependency_links=False):
+        """Build wheels of packages in the cache.
+
+        The packages should have already been downloaded via `download()`.
+
+        :param iterable packages: Packages in cache for which to build wheels.
+        :param str setup_py_dir: Directory containing setup.py.
+        :param iterable constraints: Collection of paths to constraints files.
+        :param iterable requirements: Collection of paths to requirements
+                                      files.
+        :param boolean process_dependency_links: Enable the processing of
+                                                 dependency links.
+
+        :return: List of paths to each wheel that was built.
+        :rtype: list
+        """
+        args = _process_common_args(
+            process_dependency_links=process_dependency_links,
+            packages=packages, constraints=constraints,
+            requirements=requirements)
+
+        cwd = None
+        if setup_py_dir:
+            args.append('.')
+            cwd = setup_py_dir
+
+        if not args:
+            return  # No operation was requested
+
+        wheels = []
+        with tempfile.TemporaryDirectory() as temp_dir:
+
+            # Using pip with a few special parameters:
+            #
+            # --no-index: Don't hit pypi, assume the packages are already
+            #             downloaded (i.e. by using `self.download()`)
+            # --find-links: Provide the directory into which the packages
+            #               should have already been fetched
+            # --wheel-dir: Build wheels into a temporary working area rather
+            #              rather than cwd. We'll copy them over. FIXME: We can
+            #              probably get away just building them in the package
+            #              dir. Try that once this refactor has been validated.
+            self._run(['wheel', '--no-index', '--find-links',
+                       self._python_package_dir, '--wheel-dir',
+                       temp_dir] + args, cwd=cwd)
+            wheels = os.listdir(temp_dir)
+            for wheel in wheels:
+                file_utils.link_or_copy(
+                    os.path.join(temp_dir, wheel),
+                    os.path.join(self._python_package_dir, wheel))
+
+        return [os.path.join(self._python_package_dir, wheel)
+                for wheel in wheels]
+
+    def list(self):
+        """Determine which packages have been installed.
+
+        :return: Dict of installed python packages and their versions
+        :rtype: dict
+        """
+        output = self._run(['list', '--format=json'])
+        packages = collections.OrderedDict()
+        try:
+            json_output = json.loads(
+                    output, object_pairs_hook=collections.OrderedDict)
+        except json.decoder.JSONDecodeError as e:
+            raise errors.PipListInvalidJsonError(output) from e
+
+        for package in json_output:
+            if 'name' not in package:
+                raise errors.PipListMissingFieldError('name', output)
+            if 'version' not in package:
+                raise errors.PipListMissingFieldError('version', output)
+            packages[package['name']] = package['version']
+        return packages
+
+    def clean_packages(self):
+        """Remove the package cache."""
+        with contextlib.suppress(FileNotFoundError):
+            shutil.rmtree(self._python_package_dir)
+
+    def env(self):
+        """The environment used by pip.
+
+        This function is only useful if you happen to need to call into pip's
+        environment without using the API otherwise made available here (e.g.
+        calling the setup.py directly instead of with pip).
+
+        :return: Dict of the environment necessary to use the pip contained
+                 here.
+        :rtype: dict
+        """
+        env = os.environ.copy()
+        env['PYTHONUSERBASE'] = self._install_dir
+        env['PYTHONHOME'] = self._python_home
+
+        env['PATH'] = '{}:{}'.format(
+            os.path.join(self._install_dir, 'usr', 'bin'),
+            os.path.expandvars('$PATH'))
+
+        headers = get_python_headers(
+            self._python_major_version, stage_dir=self._stage_dir)
+        if headers:
+            current_cppflags = env.get('CPPFLAGS', '')
+            env['CPPFLAGS'] = '-I{}'.format(headers)
+            if current_cppflags:
+                env['CPPFLAGS'] = '{} {}'.format(
+                    env['CPPFLAGS'], current_cppflags)
+
+        return env
+
+    def _run(self, args, **kwargs):
+        env = self.env()
+
+        return snapcraft.internal.common.run_output(
+            [self._python_command, '-m', 'pip'] + list(args), env=env,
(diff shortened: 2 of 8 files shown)


snapcraft/plugins/_python/_pip.py
```diff
@@ -0,0 +1,263 @@
...
+                          example, or the path to a dir containing a setup.py.
+        """
+        if not args:
+            return  # No packages to install
+
+        # Using pip with a few special parameters:
+        #
+        # --user: Install packages to PYTHONUSERBASE, which we've pointed to
+        #         the installdir.
+        # --no-compile: Don't compile .pyc files. FIXME: Not sure why
```
kyrofa: Although this is a refactor, I wanted to make sure I didn't change any of the args we were using. However, I wanted to carefully document why we were using these args, and I didn't know the answer to some of them. @sergiusens, can you help me fill this out by answering my FIXMEs? There are some here in `install()` as well as below in `wheel()`.

sergiusens: `--no-compile` is legacy and I think we should just remove it, it avoids needed to use `stage` and `prime` when using two parts, we even have a bug to remove it, so I say we just don't use it in this refactor.

snapcraft/plugins/_python/_pip.py
```diff
@@ -0,0 +1,263 @@
...
+        # --no-index: Don't hit pypi, assume the packages are already
+        #             downloaded (i.e. by using `self.download()`)
+        # --find-links: Provide the directory into which the packages should
+        #               have already been fetched
+        # --ignore-installed: If the package is already installed, reinstall
```
sergiusens: this is wrong, we only use it to install the latest `pip`, `wheels` and `setuptools`

```python
        args = ['pip', 'setuptools', 'wheel']

        pip_command = [self._get_python_command(), '-m', 'pip']

        pip = _Pip(exec_func=subprocess.check_call,
                   runnable=pip_command,
                   package_dir=self._python_package_dir, env=env,
                   extra_install_args=['--ignore-installed'])
```

And this is beacuse we are using python on the host to install these tools in the """venv"""

kyrofa: Ah! Good catch, that makes more sense indeed.

snapcraft/plugins/_python/_pip.py
kyrofa: Okay, I'll drop it and give it a test.

snapcraft/plugins/_python/_pip.py
```diff
@@ -0,0 +1,263 @@
...
+        with tempfile.TemporaryDirectory() as temp_dir:
+
+            # Using pip with a few special parameters:
+            #
+            # --no-index: Don't hit pypi, assume the packages are already
+            #             downloaded (i.e. by using `self.download()`)
+            # --find-links: Provide the directory into which the packages
+            #               should have already been fetched
+            # --wheel-dir: Build wheels into a temporary working area rather
+            #              rather than cwd. We'll copy them over. FIXME: Why
```
sergiusens: to have a central "across steps" repo, for which install would prefer the prebuilt wheel, this is debatable though. If you see benefits to not doing this and nothing breaks I am all for it.

Take into account that this builds wheels for the dependencies and not just what is pointed by `source`.

snapcraft/plugins/_python/_pip.py
```diff
@@ -0,0 +1,263 @@
...
+            install_dir=self._install_dir)
+
+        self._setup()
+
+    def _setup(self):
+        # Check to see if we have our own pip, yet. If not, we need to use the
+        # pip on the host (installed via build-packages) to grab our own.
+        try:
+            self._run([], stderr=subprocess.STDOUT)
+        except subprocess.CalledProcessError as e:
```
elopio: I know python is all about asking forgiveness and not permission, but this would look nicer to me with a function _is_pip_installed.

snapcraft/plugins/_python/_pip.py
```diff
@@ -0,0 +1,263 @@
...
+            output = e.output.decode(sys.getfilesystemencoding()).strip()
+            if 'no module named pip' in output.lower():
+                logger.info('Fetching pip, setuptools, and wheel...')
+                env = {'PYTHONHOME': '/usr'}
```
elopio: I think this will bite us in the future when there is no pip in /usr.
But it will be easy to spot and fix in that future.

snapcraft/plugins/_python/_pip.py
kyrofa: Fixed. `--ignore-installed` is now only used when pip etc. are being installed.

snapcraft/plugins/_python/_pip.py
```diff
@@ -0,0 +1,263 @@
...
+
+    def download(self, args, **kwargs):
+        """Using the requested args, download packages into a cache.
+
+        Simple wrapper for `pip download`. Args are appended at the end, and
+        any keyword args are passed along to `common.run_output`. Note that
+        this function does not install the fetched packages.
+
+        :param list args: Arguments for `pip download`. Package names, for
+                          example, or the path to a dir containing a setup.py.
```
elopio: Could we split package names from other flags we want to pass to download?

snapcraft/plugins/_python/_pip.py
kyrofa: I updated the FIXME to indicate that we should try building in the package_dir once the refactor has been validated.

snapcraft/plugins/_python/_pip.py
kyrofa: We agreed on chat that it makes more sense to leave this alone until the refactor has been validated, so I updated the FIXME to indicate such.

snapcraft/plugins/_python/_pip.py
```diff
@@ -0,0 +1,263 @@
...
+            json_output = json.loads(
+                    output, object_pairs_hook=collections.OrderedDict)
+        except json.decoder.JSONDecodeError as e:
+            raise errors.PipListInvalidJsonError(output) from e
+
+        for package in json_output:
+            if 'name' not in package:
+                raise errors.PipListMissingFieldError('name', output)
+            if 'version' not in package:
+                raise errors.PipListMissingFieldError('version', output)
```
elopio: Wait, is that possible? Shouldn't we just panic because pip is broken?

snapcraft/plugins/_python/_pip.py
```diff
@@ -0,0 +1,263 @@
...
+        """Remove the package cache."""
+        with contextlib.suppress(FileNotFoundError):
+            shutil.rmtree(self._python_package_dir)
+
+    def env(self):
+        """The environment used by pip.
+
+        This function is only useful if you happen to need to call into pip's
+        environment without using the API otherwise made available here (e.g.
+        calling the setup.py directly instead of with pip).
```
elopio: I got lost here. This sounds like an internal method _env. If somebody needs to call setup.py without pip, why would he use the pip class?

snapcraft/plugins/_python/_pip.py
```diff
@@ -0,0 +1,263 @@
...
+        :return: Dict of the environment necessary to use the pip contained
+                 here.
+        :rtype: dict
+        """
+        env = os.environ.copy()
+        env['PYTHONUSERBASE'] = self._install_dir
+        if self._python_command.startswith(self._stage_dir):
+            env['PYTHONHOME'] = os.path.join(self._stage_dir, 'usr')
+        else:
+            env['PYTHONHOME'] = os.path.join(self._install_dir, 'usr')
```
elopio: what about adding get_python_home to python_finder?

snapcraft/plugins/_python/errors.py
```diff
@@ -47,3 +47,19 @@ class MissingSitePyError(PythonPluginError):
 
     def __init__(self, site_py_glob):
         super().__init__(site_py_glob=site_py_glob)
+
+
+class PipListInvalidJsonError(PythonPluginError):
+
+    fmt = "Pip packages output isn't valid json: {json!r}"
```
elopio: Here again, it feels like you are being extremely cautious. If pip list --json doesn't return a json, I think it's time to panic because everything is likely to be broken. In many places, we are not wrapping this unexpected errors, and I'm not sure if we should. But that's a good discussion to have.

snapcraft/tests/plugins/python/test_pip.py
```diff
@@ -0,0 +1,412 @@
...
+    _pip,
+    errors,
+)
+
+from snapcraft import (
+    tests
+)
+
+
+def _create_python_binary(base_dir):
```
elopio: don't you have this same method three times now?
Why not put it in a PythonBaseTestCase class?

snapcraft/tests/plugins/python/test_pip.py
```diff
@@ -0,0 +1,412 @@
...
+
+
+class PipRunTestCase(tests.TestCase):
+
+    def setUp(self):
+        super().setUp()
+
+        patcher = mock.patch('snapcraft.internal.common.run_output')
+        self.mock_run_output = patcher.start()
+        self.addCleanup(patcher.stop)
```
elopio: Your tests are very nice, but I still prefer my implementation with a fake pip fixture.
As I said during our discussion, I'm not sure what's best, so I'll let Sergio break the tie.

snapcraft/plugins/_python/_pip.py
sergiusens: indeed

snapcraft/plugins/_python/_pip.py
sergiusens: +1

snapcraft/plugins/_python/_pip.py
```diff
@@ -98,7 +98,9 @@ def _setup(self):
                 # Using the host's pip, install our own pip and other tools we
                 # need.
                 self.download({'pip', 'setuptools', 'wheel'}, env=env)
-                self.install({'pip', 'setuptools', 'wheel'}, env=env)
+                self.install(
+                    {'pip', 'setuptools', 'wheel', '--ignore-installed'},
```
sergiusens: what is the line here `--ignore-installed` is as much a parameter as `install` is (which is a subcommand).
I would like this to be some sort of keyword argument (option list, true/false, or like `build-attributes`)

snapcraft/plugins/_python/_pip.py
```diff
@@ -0,0 +1,263 @@
...
+    simply shells out to pip with the magical arguments necessary to install
+    packages into a part. The important functions are:
+
+    download(): Fetch the provided list of packages (but do not install them)
+    install(): Install the provided list of packages (but do not fetch them)
+    wheel(): Build wheels for the provided list of packages (but do not fetch
+             them)
+    list(): List the installed packages
+    clean_packages(): Remove the package cache
+    env(): Obtain the environment required when calling pip
```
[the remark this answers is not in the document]
kyrofa: Agreed, it could easily get out of date. I'll remove it.

snapcraft/plugins/_python/_pip.py
kyrofa: Hey, we're calling an external command, anything is possible :stuck_out_tongue: . In all seriousness, no, I don't see this happening. But if it does, I think a bug report that says:

> I ran snapcraft and got "Pip packages json missing 'name' field"

Would be easier to triage than:

> I ran snapcraft and got this traceback:
>
> ```
> Traceback (most recent call last):
>   File "/snapcraft/bin/snapcraft", line 36, in <module>
>     obj=dict(project=snapcraft.ProjectOptions()))
>   File "/usr/lib/python3/dist-packages/click/core.py", line 716, in __call__
>     return self.main(*args, **kwargs)
>   File "/usr/lib/python3/dist-packages/click/core.py", line 696, in main
>     rv = self.invoke(ctx)
>   File "/usr/lib/python3/dist-packages/click/core.py", line 1037, in invoke
>     return Command.invoke(self, ctx)
>   File "/usr/lib/python3/dist-packages/click/core.py", line 889, in invoke
>     return ctx.invoke(self.callback, **ctx.params)
>   File "/usr/lib/python3/dist-packages/click/core.py", line 534, in invoke
>     return callback(*args, **kwargs)
>   File "/usr/lib/python3/dist-packages/click/decorators.py", line 17, in new_func
>     return f(get_current_context(), *args, **kwargs)
>   File "/snapcraft/snapcraft/cli/__init__.py", line 110, in run
>     ctx.forward(lifecyclecli.commands['snap'])
>   File "/usr/lib/python3/dist-packages/click/core.py", line 552, in forward
>     return self.invoke(cmd, **kwargs)
>   File "/usr/lib/python3/dist-packages/click/core.py", line 534, in invoke
>     return callback(*args, **kwargs)
>   File "/snapcraft/snapcraft/cli/lifecycle.py", line 135, in snap
>     project_options, directory=directory, output=output)
>   File "/snapcraft/snapcraft/internal/lifecycle.py", line 360, in snap
>     execute('prime', project_options)
>   File "/snapcraft/snapcraft/internal/lifecycle.py", line 130, in execute
>     _Executor(config, project_options).run(step, part_names)
>   File "/snapcraft/snapcraft/internal/lifecycle.py", line 225, in run
>     self._run_step(step, part, part_names)
>   File "/snapcraft/snapcraft/internal/lifecycle.py", line 262, in _run_step
>     getattr(part, step)()
>   File "/snapcraft/snapcraft/internal/pluginhandler/__init__.py", line 252, in pull
>     self.plugin.pull()
>   File "/snapcraft/snapcraft/plugins/python.py", line 180, in pull
>     self._run_pip(setup, download=True)
>   File "/snapcraft/snapcraft/plugins/python.py", line 340, in _run_pip
>     return pip.list(self.run_output)
>   File "/snapcraft/snapcraft/plugins/python.py", line 477, in list
>     packages[package['name']] = package['version']
> KeyError: 'name'
> ```

In fact, since the error is probably on their system, that helpful error might lead them to their problem, whereas the traceback is opaque. I guess my point is, it'd be nice to add some context to such errors if we have it. But indeed, this may be overly cautious. Shall I remove these exceptions?

snapcraft/plugins/_python/_pip.py
```diff
@@ -0,0 +1,263 @@
...
+    get_python_command,
+    get_python_headers,
+)
+from . import errors
+
+logger = logging.getLogger(__name__)
+
+
+class Pip:
+    """Wrapper for pip abstracting the nasty args necessary for use in a part.
```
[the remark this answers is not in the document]
kyrofa: Hahaha, fiiiine.

snapcraft/plugins/_python/_pip.py
kyrofa: Yeah this seems weird when reviewing in isolation, so to explain allow me to point you to what WILL use it. The Python plugin uses pip to download and install packages from `python-packages`, the setup.py, requirements.txt, etc. However, in addition to that, it also runs the setup.py manually, [here](https://github.com/snapcore/snapcraft/blob/master/snapcraft/plugins/python.py#L292). That needs to execute within the same environment, and yet as it explicitly doesn't use pip, this refactor does not extract it from the plugin.

Do you see a cleaner way to do that?

snapcraft/plugins/_python/_pip.py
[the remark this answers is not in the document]
kyrofa: Actually, I just removed the function overviews. I left the high-level purpose and download-before-install note.

snapcraft/plugins/_python/_pip.py
```diff
@@ -0,0 +1,263 @@
...
+    download(): Fetch the provided list of packages (but do not install them)
+    install(): Install the provided list of packages (but do not fetch them)
+    wheel(): Build wheels for the provided list of packages (but do not fetch
+             them)
+    list(): List the installed packages
+    clean_packages(): Remove the package cache
+    env(): Obtain the environment required when calling pip
+
+    Of particular importance: packages must be downloaded (via download())
+    before they can be installed or have wheels built.
```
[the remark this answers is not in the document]
kyrofa: Yeah the API seems error-prone, I agree. Especially because I'm not sure how easy it would be to accurately detect if one is trying to install a package that has not been downloaded. However, it's done this way because the download should only be done in the pull step, but the install needs to happen in both pull and build.

snapcraft/plugins/_python/_pip.py
```diff
@@ -0,0 +1,263 @@
...
+
+        :raises MissingPythonCommandError: If no python could be found in the
+                                           staging or part's install area.
+        """
+        self._python_major_version = python_major_version
+        self._part_dir = part_dir
+        self._install_dir = install_dir
+        self._stage_dir = stage_dir
+
+        self._python_package_dir = os.path.join(self._part_dir, 'packages')
```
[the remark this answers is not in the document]
kyrofa: Good call, done.

snapcraft/plugins/_python/_pip.py
```diff
@@ -0,0 +1,263 @@
...
+
+    Of particular importance: packages must be downloaded (via download())
+    before they can be installed or have wheels built.
+    """
+
+    def __init__(self, *, python_major_version, part_dir, install_dir,
+                 stage_dir):
+        """Initialize pip.
+
+        First check to see if pip has already been installed. If not, fetch
```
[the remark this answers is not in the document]
kyrofa: Fixed.

snapcraft/plugins/_python/_pip.py
kyrofa: Done, thank you, that looks much better indeed.

snapcraft/plugins/_python/_pip.py
kyrofa: Good call, done.

snapcraft/tests/plugins/python/test_pip.py
kyrofa: Ah,  yes of course, good catch.

snapcraft/plugins/_python/_pip.py
kyrofa: Agreed. I just pushed a commit that makes the API much more explicit via kwargs. Please let me know what you think.

snapcraft/tests/plugins/python/test_pip.py
kyrofa: After our discussion today, I think I agree that the fake fixture is best. However, I'd rather have that as a follow-up.

snapcraft/plugins/_python/_pip.py
```diff
@@ -0,0 +1,339 @@
...
+        # --no-index: Don't hit pypi, assume the packages are already
+        #             downloaded (i.e. by using `self.download()`)
+        # --find-links: Provide the directory into which the packages should
+        #               have already been fetched
+        self._run(['install', '--user', '--no-compile', '--no-index',
+                   '--find-links', self._python_package_dir] + args, cwd=cwd)
+
+    def wheel(self, packages, *, setup_py_dir=None, constraints=None,
+              requirements=None, process_dependency_links=False):
+        """Bbuild wheels of packages in the cache.
```
elopio: Typo: Bbuild/Build

snapcraft/plugins/_python/_pip.py
elopio: Got it.

snapcraft/plugins/_python/_pip.py
kyrofa: Ugh, thank you.

snapcraft/plugins/_python/__init__.py
```diff
@@ -14,5 +14,6 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+from ._pip import Pip  # noqa
 from ._python_finder import get_python_command  # noqa
```
sergiusens: do these two (from a previous PR) necessarily need to be exported?

snapcraft/plugins/_python/_pip.py
```diff
@@ -0,0 +1,339 @@
...
+    get_python_command,
+    get_python_headers,
+    get_python_home,
+)
+from . import errors
+
+logger = logging.getLogger(__name__)
+
+
+def _process_common_args(args, *, packages, constraints,
```
sergiusens: I like this, but even though it is a list, it would be better to return `args` as well.

snapcraft/plugins/_python/_pip.py
```diff
@@ -0,0 +1,339 @@
...
+
+    def _setup(self):
+        # Check to see if we have our own pip, yet. If not, we need to use the
+        # pip on the host (installed via build-packages) to grab our own.
+        if not self._is_pip_installed():
+            logger.info('Fetching pip, setuptools, and wheel...')
+            real_python_home = self._python_home
+
+            # Make sure we're using pip from the host
+            self._python_home = os.path.join(os.path.sep, 'usr')
```
sergiusens: Although everything falls apart if this fails, the fact that there is a `defer` equivalent to `go`'s more apparent you either should use a `contextmanager` or a simple `try`/`catch` (with no `catch` :wink:):
```python
 # Make sure we're using pip from the host
try:
    self._python_home = os.path.join(os.path.sep, 'usr')
    # Using the host's pip, install our own pip and other tools we
    # need.
    self.download({'pip', 'setuptools', 'wheel'})
    self.install({'pip', 'setuptools', 'wheel'}, ignore_installed=True)
finally:
````

File: snapcraft/plugins/_python/_pip.py [hunk 23]
```diff
@@ -0,0 +1,339 @@
...
+            # need.
+            self.download({'pip', 'setuptools', 'wheel'})
+            self.install({'pip', 'setuptools', 'wheel'}, ignore_installed=True)
+
+            # Now that we have our own pip, reset the python home
+            self._python_home = real_python_home
+
+    def _is_pip_installed(self):
+        try:
+            self._run([], stderr=subprocess.STDOUT)
```
sergiusens: please add a comment, I had to look at the implementation of `_run` to figure out what this would do.

snapcraft/plugins/_python/_pip.py
```diff
@@ -0,0 +1,339 @@
...
+        :param boolean process_dependency_links: Enable the processing of
+                                                 dependency links.
+        """
+        args = []
+        cwd = None
+        if setup_py_dir:
+            args.append('.')
+            cwd = setup_py_dir
+
+        _process_common_args(
```
sergiusens: this would convey more readability if in the end we assign `args` to what this returns.

snapcraft/plugins/_python/_pip.py
```diff
@@ -0,0 +1,339 @@
...
+            args, process_dependency_links=process_dependency_links,
+            packages=packages, constraints=constraints,
+            requirements=requirements)
```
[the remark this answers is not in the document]
sergiusens: oh, I agree.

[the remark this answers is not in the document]
sergiusens: this is smart, but not obvious :wink:

snapcraft/plugins/_python/_pip.py
```diff
@@ -0,0 +1,339 @@
...
+        :param boolean ignore_installed: Reinstall packages if they're already
+                                         installed
+        """
+        args = []
+        cwd = None
+        if setup_py_dir:
+            args.append('.')
+            cwd = setup_py_dir
+
+        _process_common_args(
```
sergiusens: ditto about `args`

snapcraft/plugins/_python/_pip.py
```diff
@@ -0,0 +1,339 @@
...
+                                                 dependency links.
+        :param boolean upgrade: Recursively upgrade packages.
+        :param boolean install_deps: Install package dependencies.
+        :param boolean ignore_installed: Reinstall packages if they're already
+                                         installed
+        """
+        args = []
+        cwd = None
+        if setup_py_dir:
+            args.append('.')
```
sergiusens: this should be the last arg so that `python-packages` and `requirements.txt` can potentially satisfy `install_requires`

snapcraft/plugins/_python/_pip.py
```diff
@@ -0,0 +1,339 @@
...
+        :param boolean process_dependency_links: Enable the processing of
+                                                 dependency links.
+
+        :return: List of paths to each wheel that was built.
+        :rtype: list
+        """
+        args = []
+        cwd = None
+        if setup_py_dir:
+            args.append('.')
```
sergiusens: ditto on order.

snapcraft/plugins/_python/_pip.py
sergiusens: Might want to add something like, e.g. when used in scriptlets.

kyrofa: @sergiusens does this help in scriptlets? You still can't just run `pip` and expect it to work. We need some sort of binary wrapper for that, no?

snapcraft/plugins/_python/__init__.py
kyrofa: Yeah, they're still needed in the Python plugin.

snapcraft/plugins/_python/_pip.py
kyrofa: With the switch in arg ordering, it doesn't need to be passed args at all, which is a nice simplification.

snapcraft/plugins/_python/_pip.py
kyrofa: Ah yes indeed, thank you for catching that.

snapcraft/plugins/_python/_pip.py
kyrofa: Ah ha, good to know! Thank you :) .

snapcraft/plugins/_python/_pip.py
kyrofa: Done.

snapcraft/plugins/_python/_pip.py
kyrofa: Done.

snapcraft/plugins/_python/_pip.py
kyrofa: Not a problem, done.

snapcraft/plugins/_python/_pip.py
kyrofa: Yeah, stopped passing args at all, and assigned.

snapcraft/plugins/_python/_pip.py
kyrofa: Done.

snapcraft/plugins/_python/_pip.py
kyrofa: Fixed.

snapcraft/plugins/_python/_pip.py
kyrofa: Done.

PR #1550, opened 2017-09-14
demos: remove the py?-project demos

sergiusens:
Not only were these bad demos, they test little to nothing.

-----


kyrofa:
This is fine, although I've been using these demos to validate my pip refactors. Which leads me to two questions:

1. Are there good integration test snaps that replace them?
2. More philosophical: when I knew I was going to potentially break the python plugin, these demos were super easy to discover as "the snaps that will test the python plugin." That will be harder with the slew of integration test snaps we have.

sergiusens:
@kyrofa that is why we conveniently prefix the test names with the plugin name, these two demos hardly test the python plugin at all, they just dump a script.

btw, your test of fire is to build snapcraft with your changes which will happen when you propose anyways.

kyrofa:
> btw, your test of fire is to build snapcraft with your changes which will happen when you propose anyways.

*shudder* don't I know it.

PR #1551, opened 2017-09-14
dirs: set plugin, schema, and library dir for snap

kyrofa:

-----

The default plugin, schema, and library dir are relative to `sys.prefix`, which works nicely for Snapcraft as a deb or in a venv. However, Snapcraft's Python plugin installs packages in `$SNAP/` as a prefix, while Python itself is contained in `$SNAP/usr/`. As a result, using `sys.prefix` (which is `/usr`) to find these files when Snapcraft is a snap ends up looking in the wrong place.

Treat running Snapcraft as a snap as a special case, similar to how running from git is supported, and set the directories appropriately.

It's worth pointing out: when making this change, I realized nothing uses `common.get_plugindir()`. The fact that it points to an invalid place (as it always does, even in a deb or venv) doesn't have any side-effects. This PR sets it anyway for consistency, but we should consider getting rid of it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ demos/*/parts/
 demos/*/prime/
 demos/*/stage/
 demos/**/*.snap
+snap/.snapcraft/
 tests/unit/parts/
 tests/unit/snap/
 tests/unit/stage/
--- a/snapcraft/__init__.py
+++ b/snapcraft/__init__.py
@@ -456,8 +456,3 @@ def str_presenter(dumper, data):
 yaml.add_representer(str, str_presenter)
 yaml.add_representer(OrderedDict, dict_representer)
 yaml.add_constructor(_mapping_tag, dict_constructor)
-
-from snapcraft.internal import common as _common # noqa
-if _common.is_snap():
-    snap = _os.environ.get('SNAP')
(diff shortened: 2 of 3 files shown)


kyrofa:
How I tested this:

Before this change:

```
$ cd demos/opencv/
$ snapcraft -d
<snip>
No libraries to exclude from this release
<snip>
```

After this change:

$ cd demos/opencv/
$ snapcraft clean && snapcraft -d
<snip, nothing interesting>
```

To be explicit: the second run does _not_ say "No libraries to exclude from this release."

sergiusens:
Was this tested from the snap itself?

kyrofa:
Indeed,  yes. I updated to reflect.

kyrofa:
Yes, but only if we get rid of pluginsdir, where the subpath differs.

kalikiana:
I'm guessing we can't easily test this since we can't build a snap from a snap of the same PR yet?

Giving it a manual go right now. Code looks sensible.

kyrofa:
Actually, we could use our spread suite, that runs for the snap, right? I'm not familiar with that at all, though.

kalikiana:
Confirmed that it works (tried before and after as suggested).

PR #1552, opened 2017-09-15
tests: replace the first batch of demo tests with snapd integration tests

elopio:

-----
The mosquitto demo is not showing anything interesting, and it's test is not going to catch any bugs.
Instead of that, I have added a new suite of tests for the integration with snapd, that will execute a python hello world, showing that a python snap built with snapcraft can be executed.

The same with the others. I replaced them with installing and running a hello world. 

To this new suite now we can add more interesting things like hooks, and config. With the added benefit that it will run on travis, so we will catch bugs earlier.


sergiusens:
Ok, I did my first round, the thing that irks me a bit is that I would of thought the integration tests that matter would get an extra installation assertion instead of creating a completely new suite. Others, please weigh in

integration_tests/snapd/test_autotools_snap.py
```diff
@@ -0,0 +1,34 @@
...
+
+class AutotoolsTestCase(integration_tests.SnapdIntegrationTestCase):
+
+    def test_install_and_execution(self):
+        with fixture_setup.WithoutSnapInstalled('autotools-hello'):
+            self.run_snapcraft(project_dir='autotools-hello')
+            self.install_snap()
+            self.assertThat(
+                subprocess.check_output(
+                    ['autotools-hello'], universal_newlines=True),
```
kyrofa: So `universal_newlines` means you don't need to decode the output?

kyrofa:
I actually like splitting this out into a new suite. These will typically take a bit longer to run, and it gives a well-defined place to put new tests that also happen to get installed. It makes it easy to copy a template instead of "I need this snap to be installed... let me find one that IS installed so I can copy it. No, not that one..."

demos/opencv/snap/snapcraft.yaml
```diff
@@ -1,16 +0,0 @@
-name: opencv-example
-version: 1.0
-summary: Use OpenCV and OpenGL
-description: A simple OpenCV example
-confinement: strict
-
-build-packages: [gcc, g++, libc6-dev, libopencv-dev]
```
kyrofa: This was a good demo to ensure that library crawling continued working correctly. Does one of the "hello" snaps cover that case?

sergiusens:
Ok @kyrofa point taken.
@elopio please just re-add the opencv one for now until you think of a good solution and let's move this forward as a first batch

elopio:
Thanks for the catch @kyrofa.
@sergiusens I reverted opencv.

kyrofa:
Land this baby.

PR #1554, opened 2017-09-19
store: handle revoked developers

sergiusens:
Add logic to handle revoking developers from the collaboration feature.
This work includes a refactor on how assertions are presented to
the components that handle the CLI.

-----


snapcraft/storeapi/__init__.py
```diff
@@ -643,11 +640,20 @@ def push_assertion(self, snap_id, assertion, endpoint):
                 'snap_developer': assertion.decode('utf-8'),
             }
         auth = _macaroon_auth(self.conf)
+
+        url = 'snaps/{}/{}'.format(snap_id, endpoint)
+
+        # For `snap-developer`, revoking developers will require their uploads
+        # to be invalidated.
+        if force:
+            url = url + '?ignore_revoked_uploads'
```
sergiusens: hmm, this I did not double check, should probably be passed on using a dict for on `self.put`

sergiusens:
Take into account this is missing integration tests and unit tests for the assertions module

snapcraft/storeapi/assertions.py
```diff
@@ -0,0 +1,174 @@
...
+        if snap_sign.returncode != 0:
+            err = err.decode('ascii', errors='replace')
+            raise errors.StoreAssertionError(
+                snap_name=self._snap_name,
+                endpoint=self._assertion_type, error=err)
+
+        self.signed_assertion = assertion
+        return assertion
+
+    def push(self, force=False):
```
elopio: Please add a docstring for the force arg, explaining that it can remove colaborators.

snapcraft/storeapi/assertions.py
```diff
@@ -0,0 +1,174 @@
...
+        self._account_info = None
+        self.assertion = None
+        self.signed_assertion = None
+
+    def get(self):
+        self.assertion = self._store_client.get_assertion(
+            self.snap_id, self._assertion_type)
+        return self.assertion
+
+    def sign(self):
```
elopio: I would love a docstring explaning what is sign and what it does.

snapcraft/storeapi/assertions.py
```diff
@@ -0,0 +1,174 @@
...
+        cmdline = ['snap', 'sign']
+        if self._signing_key:
+            cmdline += ['-k', self._signing_key]
+        snap_sign = subprocess.Popen(
+            cmdline, stdin=subprocess.PIPE, stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE)
+        data = json.dumps(self.assertion).encode('utf8')
+        assertion, err = snap_sign.communicate(input=data)
+        if snap_sign.returncode != 0:
+            err = err.decode('ascii', errors='replace')
```
elopio: why ascii here?

snapcraft/storeapi/assertions.py
```diff
@@ -0,0 +1,174 @@
...
+            raise errors.SnapNotFoundError(self._snap_name)
+
+    def __init__(self, *, snap_name, signing_key):
+        self._store_client = StoreClient()
+        self._snap_name = snap_name
+        self._signing_key = signing_key
+        self._release = constants.DEFAULT_SERIES
+        self._account_info = None
+        self.assertion = None
+        self.signed_assertion = None
```
elopio: It would be nice to have docstring for these two attributes.

snapcraft/storeapi/assertions.py
```diff
@@ -0,0 +1,174 @@
...
+        # Reference the already fetched information
+        new_instance._account_info = self._account_info
+        new_instance.assertion = new_assertion
+        return new_instance
+
+    def get(self):
+        """Returns a dict containing the developer assertion for snap_name.
+
+        If time_format is 'human' the time will be reformatted into a human
+        friendly format.
```
elopio: Where do you set time_format?

sergiusens:
Thanks @elopio, I am working on the docstrings and also need to update some that got left over live `time_format` (which is why I decided to stop writing them until I had a good refactor).

snapcraft/storeapi/assertions.py
sergiusens: This comes from #813

snapcraft/storeapi/assertions.py
sergiusens: done

snapcraft/storeapi/assertions.py
sergiusens: done

snapcraft/storeapi/assertions.py
```diff
@@ -0,0 +1,262 @@
...
+
+class _BaseAssertion:
+    """Private Base class to handle assertions.
+
+    Implementations are supposed to define a class attribute to determine
+    the assertion type.
+
+    :cvar str _assertion_type: the assertion type, also treated as the endpoint
+                               for the assertion on the store and the payload
+                               header key for the returned data.
```
elopio: I'm not sure about the rules, but I'd say that if the attribute must be manipulated by subclasses, it shouldn't be marked as internal with the _ in the front.

snapcraft/storeapi/assertions.py
elopio: I searched on the comments and commits there and couldn't find an explanation. I missed it during that review, but now that ralsina is not around, any idea who could explain it?
Does it make sense to you? To me, utf-8 sounds better.

snapcraft/storeapi/assertions.py
```diff
@@ -0,0 +1,262 @@
...
+            self.sign()
+        self._store_client.push_assertion(
+            self.snap_id, self.signed_assertion, self._assertion_type,
+            force=force)
+
+
+class DeveloperAssertion(_BaseAssertion):
+    """Implementation of a developer assertion.
+    The assertion is used to enable collaboration for a given snap_name by
+    means of operating agains the snap-id's developer assertion store endpoint.
```
elopio: I would remove "by means of"
And there's a typo: agains/against

snapcraft/storeapi/assertions.py
```diff
@@ -0,0 +1,262 @@
...
+
+    """
+
+    _assertion_type = 'developers'
+
+    def new_assertion(self, *, developers):
+        """Create a new assertion with developers based out of the current one.
+
+        The new assertion has its assertion's authority-id normalized to the
+        assertion's publisher_id and the assertion's revision increaded by 1.
```
elopio: s/increaded/increased

snapcraft/storeapi/assertions.py
```diff
@@ -0,0 +1,262 @@
...
+
+        new_instance = DeveloperAssertion(snap_name=self._snap_name,
+                                          signing_key=self._signing_key)
+        # Reference the already fetched information
+        new_instance._account_info = self._account_info
+        new_instance.assertion = new_assertion
+        return new_instance
+
+    def get(self):
+        """Returns a dict containing the developer assertion for snap_name.
```
elopio: s/Returns/Return, per pep 257.

snapcraft/storeapi/assertions.py
```diff
@@ -0,0 +1,262 @@
...
+    def get_developers(self):
+        """Return a copy of the current developers listed in the assertion.
+
+        :returns: a list of developers.
+        :rtype: list
+        """
+        return self.get()['developers'].copy()
+
+    def is_equal(self, other_assertion):
+        """Determines equality of developer lists in each assertion.
```
elopio: -> Determine, per pep257.

snapcraft/storeapi/assertions.py
```diff
@@ -0,0 +1,262 @@
...
+
+        Returns True if the list of developers in this instances assertion list
+        is equal to the list from other_assertion.
```
elopio: I would put this in :returns:, but just a personal preference, not requesting a change.

snapcraft/cli/assertions.py
```diff
@@ -13,10 +13,34 @@
...
+    # Sample entry:
+    #
+    # developers:
+    #   - developer-id: "dev-one"      # Which developer
+    #     since: "2017-02-10 08:35:00" # When contributions started
+    #     until: "2018-02-10 08:35:00" # When contributions ceased (optional)
+    #
+    # All timestamps are UTC, and the "now" special string will be replaced by
+    # the current time. Do not remove entries or use an until time in the past
+    # unless you want existing snaps provided by the developer to stop working.""")  # noqa
```
atomatt: "stop working" is interesting. It might be that something continues to work but cannot be installed. However, I'm not sure there's a better description until things start using snap-developer.

snapcraft/cli/assertions.py
```diff
@@ -77,3 +101,87 @@ def validate(snap_name, validations, key_name):
...
+    if new_dev_assertion.is_equal(dev_assertion):
+        echo.warning('Aborting due to unchanged collaborators list.')
+        return
+
+    try:
+        new_dev_assertion.push()
+    except storeapi.errors.StoreValidationError as store_error:
+        if store_error.error_list[0]['code'] != 'revoked-uploads':
+            raise store_error
+        click.echo('This will revoke the following collaborators: {!r}'.format(
```
atomatt: The response really lists uploads/revisions of the snap, not collaborators, although the upload does include info about its developer.

snapcraft/storeapi/assertions.py
```diff
@@ -0,0 +1,262 @@
...
+                  provided developers list.
+        :rtype: DeveloperAssertion
+        """
+        new_assertion = deepcopy(self.assertion)
+        new_assertion['developers'] = developers
+
+        # The revision should be incremented, to avoid `invalid-revision`
+        # errors.
+        new_assertion['revision'] = str(int(
+            self.assertion.get('revision', '0'))+1)
```
atomatt: By convention, the first revision of an assertion is zero so you could probably pass `get` a '-1' default ... but things will still work like this.

snapcraft/storeapi/assertions.py
```diff
@@ -0,0 +1,262 @@
...
+        new_instance.assertion = new_assertion
+        return new_instance
+
+    def get(self):
+        """Returns a dict containing the developer assertion for snap_name.
+
+        The data that comes from the store query looks as follows:
+        {'snap_developer': {
+             'type': 'snap-developer',
+             'authority-id': <account_id of the publisher>,
```
atomatt: Small and mostly irrelevant point, but this could also be the store's authority, typically "canonical".

snapcraft/storeapi/assertions.py
```diff
@@ -0,0 +1,262 @@
...
+        :rtype: dict
+        """
+        try:
+            self.assertion = super().get()
+        except errors.StoreValidationError as e:
+            if e.error_list[0]['code'] != 'snap-developer-not-found':
+                raise
+            self.assertion = {
+                'type': 'snap-developer',
+                'authority-id': self.publisher_id,
```
atomatt: w.r.t. comment above about the "canonical" authority, and just to be super clear ... This bit is correct. If the assertion is generated locally then it should be whoever authenticated with snapcraft.

snapcraft/storeapi/assertions.py
sergiusens: it is an error after all; I agree, we won't be inadvertently breaking assertions so I am ok with changing it.

snapcraft/cli/assertions.py
sergiusens: it would be interesting to record this

snapcraft/cli/assertions.py
```diff
@@ -77,3 +101,87 @@ def validate(snap_name, validations, key_name):
...
+            ' '.join(store_error.error_list[0]['extra'])))
```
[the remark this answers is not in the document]
sergiusens: if we get documentation on the error format we can work with that. I would do this in a future PR

snapcraft/storeapi/assertions.py
sergiusens: I am treating it as C++, `protected`, I've seen popular projects do this.

Like 'Creating Fixtures' here https://pypi.python.org/pypi/fixtures

Since python does not enforce anything it really doesn't matter.

PR #1557, opened 2017-09-19
cleanbuild: add a --image option to build in a different image

mwhudson:
For my own nefarious purposes I want to build snaps in an artful lxd which you can't do using cleanbuild today.

### DIFF
--- a/snapcraft/internal/lxd.py
+++ b/snapcraft/internal/lxd.py
@@ -53,7 +53,7 @@ _STORE_KEY = (
 class Containerbuild:
 
     def __init__(self, *, output, source, project_options,
-                 metadata, container_name, remote=None):
+                 metadata, container_name, remote=None, image=None):
         if not output:
             output = common.format_snap_name(metadata)
         self._snap_output = output
@@ -72,11 +72,14 @@ class Containerbuild:
             kernel = server_environment['kernel_architecture']
         except KeyError:
             kernel = server_environment['kernelarchitecture']
-        deb_arch = _get_deb_arch(kernel)
-        if not deb_arch:
-            raise ContainerConnectionError(
-                'Unrecognized server architecture {}'.format(kernel))
-        self._image = 'ubuntu:xenial/{}'.format(deb_arch)
+        if image is None:
+            deb_arch = _get_deb_arch(kernel)
+            if not deb_arch:
+                raise ContainerConnectionError(
+                    'Unrecognized server architecture {}'.format(kernel))
+            self._image = 'ubuntu:xenial/{}'.format(deb_arch)
+        else:
+            self._image = image
         # Use a temporary folder the 'lxd' snap can access
         lxd_common_dir = os.path.expanduser(
             os.path.join('~', 'snap', 'lxd', 'common'))
@@ -282,11 +285,11 @@ class Containerbuild:
 class Cleanbuilder(Containerbuild):
 
     def __init__(self, *, output=None, source, project_options,
-                 metadata=None, remote=None):
+                 metadata=None, remote=None, image=None):
         container_name = petname.Generate(3, '-')
         super().__init__(output=output, source=source,
                          project_options=project_options, metadata=metadata,
-                         container_name=container_name, remote=remote)
+                         container_name=container_name, remote=remote, image=image)
 
 
 class Project(Containerbuild):
--- a/snapcraft/internal/lxd.py
+++ b/snapcraft/internal/lxd.py
@@ -289,7 +289,8 @@ class Cleanbuilder(Containerbuild):
         container_name = petname.Generate(3, '-')
         super().__init__(output=output, source=source,
                          project_options=project_options, metadata=metadata,
-                         container_name=container_name, remote=remote, image=image)
+                         container_name=container_name, remote=remote,
+                         image=image)
 
 
 class Project(Containerbuild):
--- a/snapcraft/cli/lifecycle.py
+++ b/snapcraft/cli/lifecycle.py
@@ -171,8 +171,9 @@ def clean(parts, step, **kwargs):
 @click.option('--remote', metavar='<remote>',
               help='Use a specific lxd remote instead of a local container.')
 @click.option('--image', metavar='<image>', default=None,
-              help=('Build in an instance of this image rather than ubuntu:xenial/$arch. '
-                    'Do not use this unless you know what you are doing.'))
+              help=('Build in an instance of this image rather than '
+                    'ubuntu:xenial/$arch. Do not use this unless you know '
+                    'what you are doing.'))
 @click.option('--debug', is_flag=True,
               help='Shells into the environment if the build fails.')
 def cleanbuild(remote, image, debug, **kwargs):
--- a/snapcraft/cli/lifecycle.py
+++ b/snapcraft/cli/lifecycle.py
@@ -170,9 +170,12 @@ def clean(parts, step, **kwargs):
 @add_build_options()
 @click.option('--remote', metavar='<remote>',
               help='Use a specific lxd remote instead of a local container.')
+@click.option('--image', metavar='<image>', default=None,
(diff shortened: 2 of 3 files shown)


snapcraft/internal/lxd.py
```diff
@@ -72,11 +72,14 @@ def __init__(self, *, output, source, project_options,
...
-                'Unrecognized server architecture {}'.format(kernel))
-        self._image = 'ubuntu:xenial/{}'.format(deb_arch)
+        if image is None:
+            deb_arch = _get_deb_arch(kernel)
+            if not deb_arch:
+                raise ContainerConnectionError(
+                    'Unrecognized server architecture {}'.format(kernel))
+            self._image = 'ubuntu:xenial/{}'.format(deb_arch)
+        else:
+            self._image = image
```
kyrofa: This provides the ability to build not only on a different release, but a completely different arch. Is that intended?

kyrofa:
I've tested this PR and can verify that it works, but I would prefer more consistent architecture handling unless being able to specify the arch is required.

snapcraft/internal/lxd.py
mwhudson: This is true and indeed will probably break horribly. It will also let you specify a fedora container which probably won't work either. But otoh it provides an easy mapping from what you pass to lxd launch to what you pass here. If the user does something silly and it breaks, they get to keep both pieces.

My original version just allowed "ubuntu:$series/$deb_arch" and had a --series argument, but that didn't actually work for my use case of building on artful because artful is only available on the ubuntu-daily: remote, not ubuntu:.

kyrofa: Fair enough. I have no strong opinions about the arch, and I think the overall functionality is nice to have. I'll let the others weigh in.

kalikiana: Specifying the arch is optional afair - how about checking that the arch is supported by the server if it was specified and a nice error otherwise?

mwhudson: I'm sorry, I don't understand what you mean here.

kalikiana: In this method `deb_arch` is the image architecture ie. `/amd64` mapped from the kernel architecture reported by the LXD server. We could 1. add that suffix if the image name is eg. `ubuntu:zesty` 2. or check a given image name eg. `ubuntu:zesty/amd64` to see if the last component is an architecture supported by the server.

mwhudson: lxd appears to already to do 1. (e.g. lxc launch ubuntu:artful works).

I admit I don't really get the desire to protect the user from themselves here. If I pass something stupid, I assume it fails in a pretty obvious way. Maybe I should try that...

mwhudson: Here's what happens:

{{{
(mwhudson/scrollbars)mwhudson@aeglos:~/src/subiquity$ snapcraft cleanbuild --image ubuntu:artful/ppc64el
Creating snapcraft-songfully-corneous-pearl
error: Requested architecture isn't supported by this host
  File "/snap/snapcraft/x5/bin/snapcraft", line 11, in <module>
    load_entry_point('snapcraft==2.34', 'console_scripts', 'snapcraft')()
  File "/snap/snapcraft/x5/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 565, in load_entry_point
  File "/snap/snapcraft/x5/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2631, in load_entry_point
    return ep.load()
  File "/snap/snapcraft/x5/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2291, in load
    return self.resolve()
  File "/snap/snapcraft/x5/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2297, in resolve
  File "/snap/snapcraft/x5/lib/python3.6/site-packages/snapcraft/cli/__main__.py", line 19, in <module>
    run(prog_name='snapcraft')
  File "/snap/snapcraft/x5/lib/python3.6/site-packages/click/core.py", line 722, in __call__
  File "/snap/snapcraft/x5/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/snap/snapcraft/x5/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/snap/snapcraft/x5/lib/python3.6/site-packages/click/core.py", line 895, in invoke
  File "/snap/snapcraft/x5/lib/python3.6/site-packages/click/core.py", line 535, in invoke
  File "/snap/snapcraft/x5/lib/python3.6/site-packages/snapcraft/cli/lifecycle.py", line 196, in cleanbuild
    lifecycle.cleanbuild(project_options, remote, image)
  File "/snap/snapcraft/x5/lib/python3.6/site-packages/snapcraft/internal/lifecycle.py", line 337, in cleanbuild
    metadata=config.get_metadata(), remote=remote, image=image).execute()
  File "/snap/snapcraft/x5/lib/python3.6/site-packages/snapcraft/internal/lxd.py", line 149, in execute
    with self._ensure_started():
  File "/snap/snapcraft/x5/usr/lib/python3.6/contextlib.py", line 82, in __enter__
    return next(self.gen)
  File "/snap/snapcraft/x5/lib/python3.6/site-packages/snapcraft/internal/lxd.py", line 133, in _ensure_started
    self._ensure_container()
  File "/snap/snapcraft/x5/lib/python3.6/site-packages/snapcraft/internal/lxd.py", line 121, in _ensure_container
    'lxc', 'launch', '-e', self._image, self._container_name])
  File "/snap/snapcraft/x5/usr/lib/python3.6/subprocess.py", line 291, in check_call
subprocess.CalledProcessError: Command '['lxc', 'launch', '-e', 'ubuntu:artful/ppc64el', 'local:snapcraft-songfully-corneous-pearl']' returned non-zero exit status 1.
}}}

This seems clear enough to me? It would be better if snapcraft didn't dump the traceback out but that's another issue...

kalikiana: That's exactly what I'd like to avoid here. The traceback is what happens by default on unhandled errors.

mwhudson: So the fix for that is to handle lxc launch failing better? There are other reasons it might fail (running out of disk space being an obvious one, I guess).

kalikiana: That's how we handle it because the return code from the `lxc` command doesn't reflect the type of failure.

snapcraft/internal/lxd.py
```diff
@@ -72,11 +72,14 @@ def __init__(self, *, output, source, project_options,
             kernel = server_environment['kernel_architecture']
         except KeyError:
             kernel = server_environment['kernelarchitecture']
-        deb_arch = _get_deb_arch(kernel)
-        if not deb_arch:
-            raise ContainerConnectionError(
-                'Unrecognized server architecture {}'.format(kernel))
-        self._image = 'ubuntu:xenial/{}'.format(deb_arch)
```
xnox: Imho, old code is fine and should be tweaked slightly. Make ubuntu/xenial the default value for image and subtitue it in:
```
def __init__(self, ..., image='ubuntu:xenial'):
...
self._image = '{}/{}'.format(image, deb_arch)
```

And that's it. This way existing architecture handling is preserved but one can override the image to be "ubuntu:a" or "ubuntu-daily:b" etc.

snapcraft/internal/lxd.py
kalikiana: Thinking on it a little more, I guess you were suggesting to disable the traceback if the command failed at all? That's actually a good idea!
So: we can still avoid the actual lxc error here. But have another PR to catch `lxc init`, `lxc launch` and `lxc exec` respectively failing and (use a SnapcraftError subclass to) provide a clean error message.

mwhudson:
Replaced by https://github.com/snapcore/snapcraft/pull/1769

PR #1565, opened 2017-09-21
cli: add the assemble command

sergiusens:
This is what once was supposed to be the snap-dir command.

LP: #1707095

-----


integration_tests/test_snap.py
```diff
@@ -105,6 +105,19 @@ def test_snap_directory(self):
...
 
+    def test_assemble_directory(self):
+        self.copy_project_to_cwd('assemble')
+        self.run_snapcraft('snap')
+
+        snap_file_path = 'assemble_1.0_{}.snap'.format(self.deb_arch)
+        os.remove(snap_file_path)
+
+        # Verify that Snapcraft can snap its own snap directory (this will make
+        # sure `snapcraft snap` and `snapcraft snap <directory>` are always in
```
kalikiana: This should probably say `snapcraft assemble <directory>`?

snapcraft/cli/lifecycle.py
```diff
@@ -137,6 +140,24 @@ def snap(directory, output, **kwargs):
 
 
 @lifecyclecli.command()
+@click.argument('directory')
+@click.option('--output', '-o', help='path to the resulting snap.')
+def assemble(directory, output, **kwargs):
+    """Assemble a snap from a directory layout.
+
+    The layout of <directory> should contain a valid meta/snap.yaml
```
kalikiana: How about mentioning `prime` here?

sergiusens: this is a no, `prime` could be a foreign concept to the assembly. Any other wording is welcome

integration_tests/test_snap.py
sergiusens: done

snapcraft/cli/lifecycle.py
kalikiana: Not sure what you mean by foreign concept... but maybe more explicit is better. How about

`[...], optionally meta/gui and meta/manifest.yaml
as well as any other files making up the filesystem of the snap image.`

sergiusens: foreign as in a user that has no idea will just suffer with those concepts. I would rather say _The target <directory> should contain a valid snap layout, which would minimally have a valid meta/snap.yaml_ (and I would remove what follows the comma).

I am not for listing only *some* optional entries and folks should be introduced to this as they go, much less for this assemble command which is already an advanced command.

@evandandrea your thoughts would be appreciated.

kyrofa: `meta/snap.yaml` is all that's required to make something a snap, whether the directory was created with `snapcraft prime` or not. I think the description is fine as-is.

snapcraft/internal/deprecations.py
```diff
@@ -31,6 +31,8 @@
     'dn4': "The 'history' command has been replaced by 'list-revisions'.",
     'dn5': "Aliases are now handled by the store, and shouldn't be declared "
            "in the snap.",
+    'dn6': "Use of the 'snap' command with a directory has been deprecated "
+           "in favour of the assemble command.",
```
kyrofa: I suggest quoting `'assemble'` here for consistency.

snapcraft/internal/lifecycle.py
```diff
@@ -351,15 +351,18 @@ def snap(project_options, directory=None, output=None):
     # Check for our prerequesite external command early
     repo.check_for_command('mksquashfs')
```
kyrofa: I think you can remove this now, no?

snapcraft/cli/lifecycle.py
```diff
@@ -137,6 +140,24 @@ def snap(directory, output, **kwargs):
 
 
 @lifecyclecli.command()
+@click.argument('directory')
+@click.option('--output', '-o', help='path to the resulting snap.')
+def assemble(directory, output, **kwargs):
+    """Assemble a snap from a directory layout.
```
kyrofa: I'm not sure the word "layout" adds any meaning here.

sergiusens:
thanks @kyrofa, `assemble` has been renamed to `pack` as well

kyrofa:
> thanks @kyrofa, assemble has been renamed to pack as well

No problem, I'll dismiss for now and take another pass once you've made the change.

snapcraft/internal/lifecycle.py
sergiusens: It would be best to check really early in the game still for the default "run" case.

integration_tests/test_snap.py
```diff
@@ -110,6 +110,19 @@ def test_snap_directory(self):
...
 
+    def test_assemble_directory(self):
+        self.copy_project_to_cwd('pack')
+        self.run_snapcraft('snap')
+
+        snap_file_path = 'assemble_1.0_{}.snap'.format(self.deb_arch)
+        os.remove(snap_file_path)
+
+        # Verify that Snapcraft can snap its own snap directory (this will make
+        # sure `snapcraft snap` and `snapcraft assemble <directory>` are always
```
kyrofa: Check your docs here: no longer assemble.

snapcraft/internal/deprecations.py
```diff
@@ -31,6 +31,8 @@
     'dn4': "The 'history' command has been replaced by 'list-revisions'.",
     'dn5': "Aliases are now handled by the store, and shouldn't be declared "
            "in the snap.",
+    'dn6': "Use of the 'snap' command with a directory has been deprecated "
+           "in favour of the pack command.",
```
kyrofa: All of the commands in this list are quoted except for `pack`.

snapcraft/internal/lifecycle.py
kyrofa: I don't quite understand, but as long as you've considered it, that's good enough for me.

integration_tests/test_snap.py
kyrofa: Ha! You're so sneaky...

snapcraft/internal/deprecations.py
sergiusens: quoting is indeed something good to have

PR #1574, opened 2017-09-28
Add Dotnet plugin and a sample snap for integration test

rakeshsinghranchi:

-----

### DIFF
--- a/snapcraft/plugins/dotnet.py
+++ b/snapcraft/plugins/dotnet.py
@@ -1,0 +1,200 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2017 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""The dotnet plugin is used to build dotnet core runtime parts.'
+
+The plugin uses the dotnet SDK to install dependencies from nuget
+and follows standard semantics from a dotnet core project.
+
+This plugin uses the common plugin keywords as well as those for "sources".
+For more information check the 'plugins' topic for the former and the
+'sources' topic for the latter.
+
+The plugin will take into account the following build-attributes:
+
+    - debug: builds using a Debug configuration.
+    - no-runtime: does not setup the runtime. This is of use when the
+                  runtime is provided in another snap and exposed through
+                  the content interface or the build creates a standalone
+                  executable.
+
+Additionally, this plugin uses the following plugin-specific keywords:
+
+    - dotnet-runtime:
+      (string)
+      The runtime to embed in the resulting snap.
+"""
+import os
+import shutil
+import fnmatch
+
+import snapcraft
+from snapcraft import sources
+
+
+_RUNTIME_DEFAULT = '2.0.0'
+_SDK_DEFAULT = '2.0.0'
+# TODO extend for more than xenial
+_SDKS_AMD64 = {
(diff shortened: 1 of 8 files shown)


rakeshsinghranchi:
@elopio , What's the next step for this PR ? I see CI failing with test_list_plugins_large_terminal  test.

snapcraft/plugins/dotnet.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def __init__(self, name, options, project):
+        super().__init__(name, options, project)
+
+        self._dotnet_dir = os.path.join(self.partdir, 'dotnet')
+        self._dotnet_sdk_dir = os.path.join(self._dotnet_dir, 'sdk')
+
+        # These are required for the runtime
+        if 'no-runtime' in options.build_attributes:
+            self._runtime = _DummyTar()
```
elopio: this self._runtime is not being used anywhere.

snapcraft/plugins/dotnet.py
```diff
@@ -0,0 +1,200 @@
...
+        try:
+            sdk_arch = _SDK_DICT_FOR_ARCH[self.project.deb_arch]
+        except KeyError as missing_arch:
+            raise NotImplementedError(
+                'This plugin does not support architecture '
+                '{}'.format(missing_arch))
+        try:
+            # TODO support more SDK releases
+            sdk_version = sdk_arch['2.0.0']
+        except KeyError as missing_version:
```
elopio: This is an exception that will never be reached, because the '2.0.0' is hardcoded.

snapcraft/plugins/dotnet.py
```diff
@@ -0,0 +1,200 @@
...
+}
+
+
+class DotNetPlugin(snapcraft.BasePlugin):
+
+    @classmethod
+    def schema(cls):
+        schema = super().schema()
+
+        schema['properties']['dotnet-runtime'] = {
```
elopio: This property is never used in the plugin.

snapcraft/plugins/dotnet.py
```diff
@@ -0,0 +1,200 @@
...
+        publish_cmd = [self._dotnet_cmd, 'publish', '-c', configuration,
+                       '-o', self.installdir]
+        if 'no-runtime' not in self.options.build_attributes:
+            # Build command for self-contained application
+            publish_cmd += ['--self-contained', '-r', 'linux-x64']
+        self.run(publish_cmd)
+
+        if 'no-runtime' in self.options.build_attributes:
+            os.makedirs(os.path.join(self.installdir, 'dotnet-runtime'),
+                        exist_ok=True)
```
elopio: no-runtime means only that a dir has to be created? That sounds weird.

snapcraft/plugins/dotnet.py
```diff
@@ -0,0 +1,200 @@
...
+        else:
+            # Workaround to set the right permission for the executable.
+            appname = os.path.join(self.installdir, self._get_appname())
```
elopio: if _get_appname doesn't return anything, this call will fail.

elopio:
Hello @rakeshsinghranchi!

I was still traveling yesterday. I've just pushed the tests I wrote, and left a few comments in the PR for you and @sergiusens to look at. It seems to me that it will be easier now to remove everything related to 'no-runtime' in this PR, and add it in a follow up, with an integration test.

snapcraft/plugins/dotnet.py
```diff
@@ -0,0 +1,199 @@
...
+                os.chmod(appname, 0o755)
+
+    def _get_appname(self):
+        for file in os.listdir(self.builddir):
+            if fnmatch.fnmatch(file, '*.??proj'):
+                return os.path.splitext(file)[0]
+                break
+
+    def env(self, root):
+        env = super().env(root)
```
elopio: It seems to me that this env is never used by the plugin.

elopio:
To make this easier to land asap, I've pushed a new commit that removes the unused code, mostly references to runtime. In follow up PRs we can add the option to select a dotnet version, and no-runtime.

This is the commit with the removed code, so we don't lose it:
https://github.com/snapcore/snapcraft/pull/1574/commits/33d9a62079bf3cccb104984e07770ab9552c87da

rakeshsinghranchi:
Thanks so much @elopio and @sergiusens for helping throughout this PR. @sergiusens , this PR takes care of building Self-Contained application. We would need "no-runtime" ( or equivalent enum ) build attribute or plugin specific keyword to support FDD ( Framework dependent deployment) scenario too. 

Will work on picking up SDK from releases json file and default to dotnet-sdk snap as build-snap.

PR #1577, opened 2017-09-29
lxd: don't inject local snaps on a different arch

kalikiana:

-----
As [brought up in the forum](https://forum.snapcraft.io/t/cleanbuild-remote-on-pi-grabs-wrong-arch-lxc-image/691/15) when running the Snapcraft snap with a remote that has a different architecture it will try to push local snaps that can't be installed. For instance, local x86-64 machine, remote Raspberry Pi, which is armhf. In that case we have to install the snaps from the store instead.


snapcraft/tests/test_lxd.py
```diff
@@ -367,16 +376,21 @@ def test_inject_snap_dangerous(self,
...
              'id': '',
+             'channel': 'edge',
              'revision': 'x1'},
         ]
 
         builder = self.make_cleanbuilder()
 
         builder.execute()
+        if hasattr(self, 'cross') and self.cross:
+            return
```
elopio: I don't understand this return. Should it be a skip? A test without assertion is wrong.

snapcraft/internal/lxd.py
```diff
@@ -218,6 +218,25 @@ def _inject_snap(self, name):
...
+            cmd = ['snap', 'install', name]
+            if is_classic:
+                cmd.append('--classic')
+            cmd.append('--channel')
+            cmd.append(json['result']['channel'])
+            self._container_run(cmd)
+            # Switch channel if needed
+            cmd[1] = 'refresh'
+            self._container_run(cmd)
+            return
```
elopio: I think it might be clearer if you split wrap this into a function.

snapcraft/tests/test_lxd.py
kalikiana: I was thinking the existing assertions didn't make sense... I added different ones now.

snapcraft/internal/lxd.py
kalikiana: Good idea actually. I refactored it now, and the file-based install also uses it.

snapcraft/internal/lxd.py
```diff
@@ -246,13 +253,27 @@ def _inject_snap(self, name):
...
+        if is_dangerous:
+            args.append('--dangerous')
         if is_classic:
-            cmd.append('--classic')
-        self._container_run(cmd)
+            args.append('--classic')
+        self._container_run(['snap', 'install', name] + args)
+        if channel:
+            # Switch channel if needed
+            self._container_run(['snap', 'refresh', name] + args)
```
kyrofa: I'm curious why we're doing this in a step separated from the install. I guess I have two questions:

1. Why not just use `--channel=<cannel>` in the install?
2. What do we expect to happen if we refresh to (or install with) a channel when installing with `--dangerous`?

snapcraft/tests/test_lxd.py
kyrofa: The return still seems to be here, though.

kalikiana: I figured it's redundant since it's the same in both cases... which in hindsight I should've said in the comment. I added it now anyway.

snapcraft/internal/lxd.py
kalikiana: I updated the comments in an attempt to clarify: `snap install` will do nothing if the snap's already installed, even if the channel is different. So the refresh is needed to switch it - it won't do anything if the channel is the same. That's just how the command works...
There is no case where `--dangerous` is used with a channel - if it's a different arch and we `snap install` by name plus channel, there's no reasonable way to push an x-revisioned snap then, so we use the regular one. Does this make sense to you?

kalikiana:
Failing with a false negative `any.whl/urllib3/util/retry.py", line 228, in increment
    total -= 1
TypeError: unsupported operand type(s) for -=: 'Retry' and 'int'`

snapcraft/internal/lxd.py
kyrofa: Ah, I see-- I actually missed that you were also passing the channel to install. Carry on.

snapcraft/internal/lxd.py
```diff
@@ -218,6 +218,13 @@ def _inject_snap(self, name):
         id = json['result']['id']
         # Lookup confinement to know if we need to --classic when installing
         is_classic = json['result']['confinement'] == 'classic'
+
+        # If the server has a different arch we can't inject local snaps
+        if (self._project_options.target_arch
+                and self._project_options.target_arch != self._server_arch):
+            channel = json['result']['channel']
+            return self._install_snap(name, channel, is_classic=is_classic)
```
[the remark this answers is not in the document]
sergiusens: eventually this code would need some refactoring, these return statements in the middle will bite us back later

PR #1578, opened 2017-09-29
project_loader: quote more environment variable values

malept:
- ~~If this is a new feature. Have you discussed the design on the [forum](https://forum.snapcraft.io)?~~

-----

Fixes https://bugs.launchpad.net/snapcraft/+bug/1720230

### DIFF
--- a/snapcraft/internal/meta.py
+++ b/snapcraft/internal/meta.py
@@ -306,7 +306,8 @@ class _SnapPackaging:
 
     def _write_wrap_exe(self, wrapexec, wrappath,
                         shebang=None, args=None, cwd=None):
-        args = ' '.join(args) + ' "$@"' if args else '"$@"'
+        quoted_args = ['"{}"'.format(arg) for arg in args]
+        args = ' '.join(quoted_args) + ' "$@"' if args else '"$@"'
         cwd = 'cd {}'.format(cwd) if cwd else ''
 
         # If we are dealing with classic confinement it means all our
--- a/snapcraft/internal/meta.py
+++ b/snapcraft/internal/meta.py
@@ -306,7 +306,10 @@ class _SnapPackaging:
 
     def _write_wrap_exe(self, wrapexec, wrappath,
                         shebang=None, args=None, cwd=None):
-        quoted_args = ['"{}"'.format(arg) for arg in args]
+        if args:
+            quoted_args = ['"{}"'.format(arg) for arg in args]
+        else:
+            quoted_args = []
         args = ' '.join(quoted_args) + ' "$@"' if args else '"$@"'
         cwd = 'cd {}'.format(cwd) if cwd else ''
 
--- a/snapcraft/internal/project_loader/_config.py
+++ b/snapcraft/internal/project_loader/_config.py
@@ -201,8 +201,8 @@ class Config:
 
     def project_env(self):
         return [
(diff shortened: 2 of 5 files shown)


malept:
I think one of the unit test builds failed due to network issues?

sergiusens:
indeed it has @malept 
retrying now

snapcraft/internal/meta.py
```diff
@@ -306,7 +306,11 @@ def _get_version(self, version, version_script=None):
 
     def _write_wrap_exe(self, wrapexec, wrappath,
                         shebang=None, args=None, cwd=None):
-        args = ' '.join(args) + ' "$@"' if args else '"$@"'
+        if args:
+            quoted_args = ['"{}"'.format(arg) for arg in args]
```
sergiusens: May I recommend:
```
In [1]: import shlex

In [2]: args = [shlex.quote(arg) for arg in ['a a', 'b', '--open', '--close file=file-path']]

In [3]: args
Out[3]: ["'a a'", 'b', '--open', "'--close file=file-path'"]
```

sergiusens: shlex is already imported in this module btw,

sergiusens: hmm, maybe do not do this as environment variables will be single quoted

malept: Maybe only run `shlex.quote` if there's a space in the arg?

sergiusens: shlex.quote would be good, but we might have cases where we have something like

```
command: my-command --wordir $SNAP
```
which would end up with us having `'$SNAP' in single quotes which means it wouldn't be expanded by the shell.

malept: This might not even be the correct place to fix this, but what I was seeing was something to the effect of `desktop-launch` being generated with this line:

```sh
exec path/to/Test App/testapp "$@"
```

Which would obviously fail. _(I'll take a look later tonight at the exact line I saw, this is from memory.)_

malept: My mistake, it was `command-$APPNAME.wrapper`, the last line looks like this:

```
exec "desktop-launch" $SNAP/testapp/Test App "$@"
```

malept: Been busy, I'll try to paste an example this weekend.

malept: Here's a fuller example (from Electron Forge tests) about what happens with the current version of snapcraft, not this PR: https://gist.github.com/malept/089081b34271ff6e9ddd2f8faef7f536

I don't know if that behavior with the quotes as a subset of the string is a specified behavior in YAML. [This is what happens](http://nodeca.github.io/js-yaml/#yaml=YXBwczoKICB0ZXN0YXBwOgogICAgY29tbWFuZDogZGVza3RvcC1sYXVuY2ggJyRTTkFQL3Rlc3RhcHAvVGVzdCBBcHAn) when the YAML library I'm using deserializes it.

A workaround may look like this:

```yaml
apps:
  testapp:
     command: "desktop-launch '$SNAP/testapp/Test App'"
```

However, I'd have to see whether there's even that level of control over how the YAML is serialized.

kyrofa: @malept ah ha! Okay, very good. And yes, I can verify that your change fixes that case.

PR #1581, opened 2017-10-02
catkin plugin: support rosdep pip dependencies

kyrofa:

-----

The Catkin plugin uses rosdep to resolve dependencies, however, the only type of dependencies it currently supports are apt-based. This PR fixes LP: [#1683036](https://bugs.launchpad.net/snapcraft/+bug/1683036) by adding support for pip-based dependencies.

### DIFF
--- a/snapcraft/plugins/_python/_pip.py
+++ b/snapcraft/plugins/_python/_pip.py
@@ -20,6 +20,7 @@ import json
 import logging
 import os
 import shutil
+import stat
 import subprocess
 import sys
 import tempfile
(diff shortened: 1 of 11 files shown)


snapcraft/plugins/_python/_pip.py
```diff
@@ -99,9 +98,12 @@ def __init__(self, *, python_major_version, part_dir, install_dir,
             self._python_major_version, stage_dir=self._stage_dir,
             install_dir=self._install_dir)
 
-        self._setup()
```
sergiusens: Good call, this makes testing easier :-)

kyrofa: Yeah. I liked having an init that made sure the class was immediately usable, but then it was impossible to determine whether or not things were setup without... setting them up :stuck_out_tongue: .

snapcraft/plugins/_python/_pip.py
```diff
@@ -124,6 +126,10 @@ def _setup(self):
                 # Now that we have our own pip, reset the python home
                 self._python_home = real_python_home
 
+    def is_setup(self):
```
sergiusens: Given the docstring, this should probably be renamed to `is_pip_setup`

sergiusens: And given what it does, I would just make `_is_pip_installed` public and use that instead

sergiusens: And if so, this is probably missing a test, please double check.

kyrofa: I agree with promoting `_is_pip_installed` to public, but I'd rather call it `is_setup()` so we can have the `setup`/`is_setup` logical pair.

kyrofa: Done.

Issue #1443, opened 2017-08-08
new ament plugin (ros2)


sergiusens:
[design](https://forum.snapcraft.io/t/proposal-ros-2-support/1087)

kyrofa:
I'm spiking this now, but allowing for time for OSRF review I'll say a few weeks.

PR #1582, opened 2017-10-02
plugins: add ros2 boostrapper

kyrofa:

-----

Pave the way for an Ament plugin: make progress on LP: [#1686850](https://bugs.launchpad.net/snapcraft/+bug/1686850) (and #1443) by adding code to bootstrap ROS2 underlay from source.

Note that a Debian archive containing ros2 exists, but according to the docs it's temporary. Until it stabilizes, we'll build from source. This PR is essentially the [documentation for building from source](https://github.com/ros2/ros2/wiki/Linux-Development-Setup) in Python form.

A follow-up PR will add an Ament plugin that utilizes the bootstrapper.

### DIFF
--- a/snapcraft/plugins/_ros/ros2.py
+++ b/snapcraft/plugins/_ros/ros2.py
@@ -73,10 +73,9 @@ class Bootstrapper:
         # Dependencies for the rest of ros2
         packages.extend([
             'libopencv-dev', 'libpoco-dev', 'libpocofoundation9v5',
-            'libpocofoundation9v5-dbg', 'python-empy', 'python3-dev',
-            'python3-empy', 'python3-nose', 'python3-pip',
-            'python3-setuptools', 'python3-yaml', 'libtinyxml-dev',
-            'libeigen3-dev'])
+            'libpocofoundation9v5-dbg', 'python3-dev', 'python3-empy',
+            'python3-nose', 'python3-pip', 'python3-setuptools',
+            'python3-yaml', 'libtinyxml-dev', 'libeigen3-dev'])
 
         return packages
 
--- a/snapcraft/plugins/_ros/ros2.py
+++ b/snapcraft/plugins/_ros/ros2.py
@@ -86,8 +86,8 @@ class Bootstrapper:
         """Download the ROS2 underlay.
 
         This first downloads the tools required to fetch the underlay, and then
-        uses those tools fetch the underlay. Both steps are run in a manner
-        such that they are only run once, even if subsequent steps fail.
+        uses those tools to fetch the underlay. Both steps are only run once,
+        even if subsequent steps fail.
         """
         self._run_step(
             self._install_tools,
--- a/snapcraft/plugins/_ros/ros2.py
+++ b/snapcraft/plugins/_ros/ros2.py
@@ -65,19 +65,16 @@ class Bootstrapper:
     def get_build_packages(self):
         """Return the packages required for building the underlay."""
 
-        packages = ['cmake']
+        return [
+            # Dependencies for FastRTPS
+            'libasio-dev', 'libtinyxml2-dev',
 
-        # Dependencies for FastRTPS
-        packages.extend(['libasio-dev', 'libtinyxml2-dev'])
-
-        # Dependencies for the rest of ros2
-        packages.extend([
-            'libopencv-dev', 'libpoco-dev', 'libpocofoundation9v5',
+            # Dependencies for the rest of ros2
+            'cmake', 'libopencv-dev', 'libpoco-dev', 'libpocofoundation9v5',
             'libpocofoundation9v5-dbg', 'python3-dev', 'python3-empy',
             'python3-nose', 'python3-pip', 'python3-setuptools',
-            'python3-yaml', 'libtinyxml-dev', 'libeigen3-dev'])
-
-        return packages
+            'python3-yaml', 'libtinyxml-dev', 'libeigen3-dev'
+        ]
 
     def get_stage_packages(self):
         """Return the packages required for running the underlay."""
--- a/snapcraft/plugins/_ros/ros2.py
+++ b/snapcraft/plugins/_ros/ros2.py
@@ -130,7 +130,7 @@ class Bootstrapper:
 
         subprocess.check_call(command, env=env)
 
-    def _step_done(self, step):
+    def _is_step_done(self, step):
         return os.path.isfile(os.path.join(self._state_dir, step))
 
     def _set_step_done(self, step):
@@ -138,7 +138,7 @@ class Bootstrapper:
         open(os.path.join(self._state_dir, step), 'w').close()
 
     def _run_step(self, callable, *, step, skip_message=None):
-        if self._step_done(step):
+        if self._is_step_done(step):
             if skip_message:
                 logger.debug(skip_message)
         else:
--- a/snapcraft/plugins/_ros/ros2.py
+++ b/snapcraft/plugins/_ros/ros2.py
@@ -1,0 +1,182 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2017 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import contextlib
+import os
+import logging
+import shutil
+import subprocess
+
+from snapcraft import (
+    repo,
+    sources,
+)
+
+logger = logging.getLogger(__name__)
+
+_ROS2_URL_TEMPLATE = (
+    'https://raw.githubusercontent.com/ros2/ros2/{version}/ros2.repos'
+)
+
+
+_INSTALL_TOOLS_STEP = 'install-tools'
+_FETCH_ROS2_STEP = 'fetch-ros2'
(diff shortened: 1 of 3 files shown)


kyrofa:
@nuclearsandwich, @tfoote, @wjwwood: a review from one or more of you guys would be greatly appreciated. Just want to make sure the ROS2 bootstrap process looks sane.

tfoote:
I'm not super familiar with the snapcraft process, but it looks like you have the right flow for building ros2.

kyrofa:
@tfoote Indeed, I don't expect you to be intimately familiar with Snapcraft, just want to double-check the ros2 bit. Thank you for taking a look!

snapcraft/plugins/_ros/ros2.py
```diff
@@ -0,0 +1,182 @@
...
+
+        packages = ['cmake']
+
+        # Dependencies for FastRTPS
+        packages.extend(['libasio-dev', 'libtinyxml2-dev'])
+
+        # Dependencies for the rest of ros2
+        packages.extend([
+            'libopencv-dev', 'libpoco-dev', 'libpocofoundation9v5',
+            'libpocofoundation9v5-dbg', 'python-empy', 'python3-dev',
```
nuclearsandwich: The python-empy dependency jumped out at me as I didn't think it was necessary. It's listed on our [source build](https://github.com/ros2/ros2/wiki/Linux-Development-Setup) instructions as well which is likely where it comes from and a brief discussion with coworkers suggests it is likely no longer necessary.

It might have been necessary for the ROS 1 bridge at some point but doesn't appear to be needed anymore. I think it can be removed from this dependency list while we discuss it further. And I'll update again if I turn out to be incorrected.

kyrofa: Thank you! I'll remove it and see if things still build.

kyrofa: Verified: `python-empy` is not necessary. Removed.

snapcraft/plugins/_ros/ros2.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def __init__(self, *, version, bootstrap_path, ubuntu_sources, project):
+        """Initialize bootstrapper.
+
+        :param str version: The ROS2 version to boostrap (e.g. release-beta3)
+        :param str bootstrap_path: Working directory for the bootstrap process
+        :param str ubuntu_sources: Ubuntu repositories from which dependencies
+                                   will be fetched.
+        :param project: Instance of ProjectOptions for project-wide settings
+        :type project: snapcraft.ProjectOptions
```
elopio: why not call it project_options?

snapcraft/plugins/_ros/ros2.py
```diff
@@ -0,0 +1,181 @@
...
+
+        # Dependencies for FastRTPS
+        packages.extend(['libasio-dev', 'libtinyxml2-dev'])
+
+        # Dependencies for the rest of ros2
+        packages.extend([
+            'libopencv-dev', 'libpoco-dev', 'libpocofoundation9v5',
+            'libpocofoundation9v5-dbg', 'python3-dev', 'python3-empy',
+            'python3-nose', 'python3-pip', 'python3-setuptools',
+            'python3-yaml', 'libtinyxml-dev', 'libeigen3-dev'])
```
elopio: Save a few statements by writing it like this:
```
packages = [
    'cmake',
    # Dependencies for FastRTPS
    'libasio-dev', 'libtinyxml2-dev',
    # Dependencies for the rest of ros2
    ...
]

File: snapcraft/plugins/_ros/ros2.py (hunk 2 above)
Comment (kyrofa) [reply snapcraft/plugins/_ros/ros2.py]: I can, although we're not consistent on this. It's called `project` in the plugin, as well as rosdep, and I'm sure elsewhere. `project` is shorter and has the same meaning to me; do you feel that `project_options` is clearer?

File: snapcraft/plugins/_ros/ros2.py (hunk 3 above)
Comment (kyrofa) [reply snapcraft/plugins/_ros/ros2.py]: Sure thing.

File: snapcraft/plugins/_ros/ros2.py [hunk 4]
```diff
@@ -0,0 +1,181 @@
...
+        """Return the packages required for running the underlay."""
+
+        # Ament is a python3 tool, and it requires setuptools at runtime
+        return ['python3', 'python3-setuptools']
+
+    def pull(self):
+        """Download the ROS2 underlay.
+
+        This first downloads the tools required to fetch the underlay, and then
+        uses those tools fetch the underlay. Both steps are run in a manner
```
elopio: typo: to fetch.

snapcraft/plugins/_ros/ros2.py
```diff
@@ -0,0 +1,181 @@
...
+        such that they are only run once, even if subsequent steps fail.
```
elopio: s/: Both steps are run only once, even if subsequent steps fail.

kyrofa: Haha, yes that was rather verbose, wasn't it?

snapcraft/plugins/_ros/ros2.py
elopio: I do, because the object is not Project, it's ProjectOptions. Maybe file a bitesize bug for later to make it consistent in all plugins?

snapcraft/plugins/_ros/ros2.py
```diff
@@ -0,0 +1,181 @@
...
+        """
+        self._run_step(
+            self._install_tools,
+            step=_INSTALL_TOOLS_STEP,
+            skip_message='Tools already installed. Skipping...')
+        self._run_step(
+            self._fetch_ros2,
+            step=_FETCH_ROS2_STEP,
+            skip_message='ros2 already fetched. Skipping...')
```
elopio: Maybe, the `Skipping...` part could be added by _run_step. That way it will be harder to fail to follow the convention.

snapcraft/plugins/_ros/ros2.py
```diff
@@ -0,0 +1,181 @@
...
+    def _run(self, command):
+        env = os.environ.copy()
+        env['PATH'] = env['PATH'] + ':' + os.path.join(
+            self._tool_dir, 'usr', 'bin')
+        env['PYTHONPATH'] = os.path.join(
+            self._tool_dir, 'usr', 'lib', 'python3', 'dist-packages')
+
+        subprocess.check_call(command, env=env)
+
+    def _step_done(self, step):
```
elopio: _is_step_done, or _was_step_done.

snapcraft/plugins/_ros/ros2.py
```diff
@@ -0,0 +1,181 @@
...
+        return os.path.isfile(os.path.join(self._state_dir, step))
+
+    def _set_step_done(self, step):
+        os.makedirs(self._state_dir, exist_ok=True)
+        open(os.path.join(self._state_dir, step), 'w').close()
```
elopio: Did you have a discussion about this in the forum? I'm ok with it, but sounds like the kind of things that we are told later to rewrite. Is Sergio ok with this too?

snapcraft/plugins/_ros/ros2.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def _run_step(self, callable, *, step, skip_message=None):
+        if self._step_done(step):
+            if skip_message:
+                logger.debug(skip_message)
+        else:
+            callable()
+            self._set_step_done(step)
+
+    def _install_tools(self):
```
elopio: What do you think about _install_vcstool instead of _install_tools?

kyrofa: It's too specific: this class needs to be able to grow as ros2 changes.

elopio:
This seems pretty good to me. I left a few comments about typos and nit questions. Are you going to add an integration test for ros2?

snapcraft/plugins/_ros/ros2.py
sergiusens: Leave it as project and rename the class (with backwards compat).

snapcraft/plugins/_ros/ros2.py
sergiusens: I was going to propose this change as well, but it is somewhat more readable the original way and those statements don't really add up to the slowness of the build in comparison

snapcraft/plugins/_ros/ros2.py
kyrofa: I did not. This is pretty basic state tracking in its own working area, nothing heavy, and it's pretty required. Building the underlay takes a good chunk of time; it would be dreadful to have a build error in one's own code result in needing to rebuild it.

kyrofa:
> Are you going to add an integration test for ros2?

In the follow-up PR, yes. Then we actually have something to run/test.

snapcraft/plugins/_ros/ros2.py
kyrofa: Note that if we develop a way to track state more granularly, this would be easy to change over.

snapcraft/plugins/_ros/ros2.py
kyrofa: I'd rather keep it simple, otherwise I need to start asking "should it add the period at the end as well?" And then it's not at all clear how the function is supposed to be used, either.

snapcraft/plugins/_ros/ros2.py
kyrofa: Alright, I left it.

snapcraft/plugins/_ros/ros2.py
kyrofa: I have no strong opinions here, but I already made the change and think the readability wasn't harmed. Please let me know if you feel otherwise and I can revert that commit.

snapcraft/plugins/_ros/ros2.py
kyrofa: Fixed.

snapcraft/plugins/_ros/ros2.py
kyrofa: Fixed.

snapcraft/plugins/_ros/ros2.py
kyrofa: Went with `_is_step_done`, thank you.

PR #1583, opened 2017-10-03
ament plugin: new plugin

kyrofa:

-----

As ROS2 nears a stable release, it's a good time to introduce support for it by way of a beta plugin. This PR resolves LP: [#1686850](https://bugs.launchpad.net/snapcraft/+bug/1686850) (and #1443) by doing exactly that. We're starting simple, just bootstrapping ROS2 and using it to build the provided source. ROS2 doesn't have a lot of tooling just yet, so this plugin will need to grow with it.

Note that the high line count is due to an integration test: the plugin itself is relatively small.

### DIFF
--- a/snapcraft/plugins/ament.py
+++ b/snapcraft/plugins/ament.py
@@ -1,0 +1,197 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2017 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""The ament plugin is useful for building ROS2 parts.
+
+This plugin uses the common plugin keywords as well as those for "sources".
+For more information check the 'plugins' topic for the former and the
+'sources' topic for the latter.
+
+Additionally, this plugin uses the following plugin-specific keywords:
+
+    - version:
+      (string)
+      The ROS2 version required by this system. This relates to the ros2 tags.
+      Defaults to 'release-beta3'.
+"""
+
+import logging
+import os
+import re
+import textwrap
+
+import snapcraft
+from . import _ros
+
+logger = logging.getLogger(__name__)
+
+
+class AmentPlugin(snapcraft.BasePlugin):
+
+    @property
+    def PLUGIN_STAGE_SOURCES(self):
+        """Specify the sources used for stage-packages.
+
+        Note the ROS archive in particular. Build-packages do not come from
+        here, but stage-packages do.
+        """
+
+        return """
+deb http://packages.ros.org/ros/ubuntu/ {0} main
+deb http://${{prefix}}.ubuntu.com/${{suffix}}/ {0} main universe
+deb http://${{prefix}}.ubuntu.com/${{suffix}}/ {0}-updates main universe
+deb http://${{prefix}}.ubuntu.com/${{suffix}}/ {0}-security main universe
+deb http://${{security}}.ubuntu.com/${{suffix}} {0}-security main universe
+""".format('xenial')
+
+    @classmethod
+    def schema(cls):
+        """Specify the jsonschema of the supported options.
+
+        Snapcraft will use this schema to validate the YAML when this plugin is
+        used.
+        """
+
+        schema = super().schema()
+
+        schema['properties']['version'] = {
+            'type': 'string',
+            'default': 'release-beta3'
+        }
+
+        return schema
+
+    @classmethod
+    def get_pull_properties(cls):
+        """Inform Snapcraft of the properties associated with pulling.
+
+        If these change in the YAML, Snapcraft will consider the pull step
+        dirty.
+        """
+        return ['version']
+
+    def __init__(self, name, options, project):
+        """Initialize a new AmentPlugin.
+
+        :param str name: The name of this part
+        :param object options: Objectified version of the properties for this
+                               part
+        :param project: Instance of ProjectOptions for project-wide settings
+        :type project: snapcraft.ProjectOptions
+        """
+        super().__init__(name, options, project)
+
+        # FIXME: Remove this warning once the plugin (and indeed ROS2) is
+        # considered stable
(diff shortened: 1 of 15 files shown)


kyrofa:
@elopio we've got a problem on the snapd integration tests: if the test takes longer than 10 minutes (like the new ros2 one does), since the test shows no output Travis thinks it's hung and kills it. Can we show a spinner or something?

elopio:
@kyrofa take a look at this:
https://github.com/snapcore/snapcraft/pull/1387/files#diff-ddb7b19b1787aaa59386a4b894b4a41aR31

I think we need to capture the output, but also print it.

kyrofa:
@elopio alright, I took a crack at that in #1591.

kyrofa:
@elopio I'm afraid travis_wait doesn't work here either:

1) It masks all output completely, which sucks
2) The ROS2 test simply takes too long anyway (on my computer it takes 15 minutes).

How can we incorporate this test without killing Travis?

elopio:
@kyrofa mark it as a slow test and run it only on autopkgtest?

We can make it a little smarter. Like comment in the commit with +autopkgtest, and the bot will run them for you.

kyrofa:
After discussion in IRC, this PR is currently blocked on the implementation of "slow" tests.

kalikiana:
The PR implementing the "slow" tests is https://github.com/snapcore/snapcraft/pull/1602

kyrofa:
Alright, this PR has been cleaned up, utilizing the slow tests.

snapcraft/plugins/ament.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def _build_env(self):
+        env = os.environ.copy()
+        env['PYTHONPATH'] = os.path.join(
+            os.path.sep, 'usr', 'lib', 'python3', 'dist-packages')
+        return env
+
+    def _source_setup_sh(self, root):
+        return textwrap.dedent('''
+            if [ -f {ros_setup} ]; then
```
kalikiana: What's the reason for the `-f` here? Can't you assume the setup.sh is already in place?

integration_tests/snaps/ros2/snap/snapcraft.yaml
```diff
@@ -0,0 +1,21 @@
...
+confinement: strict
+
+apps:
+  launch-project:
+    command: launch $(ros2 pkg prefix listener_py)/share/listener_py/launch/talk_and_listen.py
+    plugs: [network]
+    environment:
+      # Required to get interleaved stdout, see
+      # https://discourse.ros.org/t/ros2-launch-not-interleaving-stdout
+      PYTHONUNBUFFERED: 1
```
kalikiana: How about setting this in the plugin's env so it doesn't have to be repeated in every snap?

kyrofa: Because I currently believe it to be a hackaround as a result of using Python, whose support is new. It was not requires in ROS1, and I expect ROS2's launch will eventually take care of it as well. If not then we can revisit.

snapcraft/plugins/ament.py
kyrofa: Not necessarily. Remember that `env()` is used both to generate the wrappers in prime _as well_ as when you call `self.run()` to build. While setup.sh will definitely be there by the end, it has to _get_ there first.

kyrofa:
> Should the tests be series-specific? Just wondering since that's what we ended up needing for the catkin ones.

That's because each release of ROS1 supports a specific subset of Ubuntu releases. ROS2 isn't released yet. That said, right now the plugin is hard-coded to xenial, as I'm not sure if the ROS repository has other releases filled out with what we require for stage-packages. I'll check on that now. Until then, running the xenial:amd64 autopkgtest.

snapcraft/plugins/ament.py
kalikiana: Very true. I had a feeling I was missing something. Thanks for clarifying.

kalikiana:
Looks like zesty indeed fails with key errors: `apt_pkg.Error: W:GPG error: http://packages.ros.org/ros/ubuntu xenial InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 5523BAEEB01FA116, E:The repository 'http://packages.ros.org/ros/ubuntu xenial InRelease' is not signed.`

integration_tests/snapd/test_ament_snap.py
```diff
@@ -0,0 +1,42 @@
...
+    def test_ament_support(self):
+        with fixture_setup.WithoutSnapInstalled('ros2-example'):
+            self.run_snapcraft(project_dir='ros2')
+            self.install_snap()
+
+            self.assertThat(
+                subprocess.check_output(
+                    ['ros2-example.launch-project'],
+                    universal_newlines=True, stderr=subprocess.STDOUT),
+                Contains("I heard: Hello world"))
```
elopio: Hey, double quotes!!! :o

kyrofa: What... I'm horrified at myself!

kyrofa: Fixed.

kyrofa:
@nuclearsandwich, @tfoote, @wjwwood here is the final piece of the puzzle regarding ROS2 support in Snapcraft. You already reviewed the most important part, that is, the bootstrapping of ROS2, but if you have some time I'd love for you to check this out as well.

The rubber meets the road (i.e. the ROS2 bootstrapper is actually used) in the [Ament plugin](https://github.com/snapcore/snapcraft/pull/1583/files#diff-99df5e939aff00176d964f7c424fc768R42). Some quick background information:

- Plugins build their code by adhering to the plugin API. The important bits here are:
    - The `pull` function, where the source itself is pulled as well as the ROS2 underlay 
    - The `build` function, where the ROS2 underlay is built, and then used to build the source. Note that everything we want to end up in the snap needs to be in `self.installdir`, but the ROS2 underlay is built elsewhere so we don't need to build it again upon failure. As a result, you'll see some workarounds there when it comes to prefixes. I'm curious if there are better ways to deal with this.

If you want to actually poke at this, check out this [simple ROS2 snap guide](https://github.com/canonical-docs/snappy-docs/pull/140), and use the snapcraft specific to this PR:

    sudo snap install snapcraft --channel=edge/pr-1583 --classic

sergiusens:
```
ERROR: test_ament_support (test_ament_snap.AmentTestCase)
test_ament_snap.AmentTestCase.test_ament_support
```

kyrofa:
> ```
> ERROR: test_ament_support (test_ament_snap.AmentTestCase)
> test_ament_snap.AmentTestCase.test_ament_support
> ```

Looks like this was caused by #1578, this was passing before the rebase. I need to [quote the path](https://github.com/snapcore/snapcraft/pull/1583/commits/ea5eec3ba6c0fb47764c1dc4f910bc434f149936#diff-4b94577b9e6f6f57ab87fc7267adf7a1R12), now.

PR #1587, opened 2017-10-04
lifecycle: clean after deleting container

kalikiana:

-----
See [the forum discussion](https://forum.snapcraft.io/t/snapcraft-clean-doesnt-clean-with-snapcraft-container-builds-set/2291/2).


snapcraft/cli/lifecycle.py
```diff
@@ -151,10 +151,12 @@ def clean(parts, step, **kwargs):
...
         snapcraft clean my-part --step build
     """
+    if not step:
+        step = 'pull'
+
     project_options = get_project_options(**kwargs)
     container_config = env.get_container_config()
-    if container_config.use_container:
-        step = step or 'pull'
+    if container_config.use_container and step == 'pull' and not parts:
```
elopio: I don't get this @kalikiana. I might be doing something wrong while testing it, but it seems to me that this still just deletes the container.
The execute method in lxd checks if we are cleaning, and if so, it just deletes the container.
In this pr I see nothing to modify that behaviour.

kalikiana: Sorry about that. It looks like the second commit was lost... I re-did the missing piece. So: clean should be called with the container to delete it, in all other cases "normal" lifecycle clean is used so that it works even if the container is broken or already deleted.

kalikiana:
Note: All but the snap test have passed

snapcraft/cli/lifecycle.py
kyrofa: I'd like to rectify the poor separation of concerns, here. We're going to great lengths to ensure we only ever call `lifecycle.containerbuild('clean')` if `step == 'pull' and not parts`, with no apparent reason. We then proceed to pass the step as well as the parts (that we just ensured was empty) to the very generic-sounding `lifecycle.containerbuild()` method. That method then calls the very generic-sounding `lxd.Project().execute()` method. Finally, that method checks to see if we're cleaning the pull step with no parts, in which case it blows the container away.

I'd like to see `lxd.Project().execute()` be written more generically. If it was asked to clean with no parts, fine, blow the container away. If it was asked to clean _with_ parts, do nothing. If something else, call `super()`. Then we don't have to let the conditions where the container needs to be removed leak to this line, where we're checking those conditions again. We can simply call `lifecycle.containerbuild('clean')` just like we call `lifecycle.clean()` and expect it to do the right thing.

kalikiana:
@kyrofa Somehow I can reply... something's off with GitHub since today I think. In any case, I changed it as per your suggestion and the logic is only in lxd.py now.

snapcraft/internal/lxd.py
```diff
@@ -395,9 +395,10 @@ def _finish(self):
             process.terminate()
 
     def execute(self, step='snap', args=None):
-        # clean with no parts deletes the container
-        if step == 'clean' and args == ['--step', 'pull']:
-            if self._get_container_status():
+        if step == 'clean':
+            # we don't execute clean here but rely on CLI to do it
+            # clean with no parts deletes the container
+            if self._get_container_status() and args == ['--step', 'pull']:
```
kyrofa: I think re-purposing this function to clean is both confusing (`clean` is not a lifecycle step, and then one sees `--step pull` and gets _more_ confused) and is causing the diff of this PR to be larger than it needs to be (changing the step from `None` to `pull` everywhere). How difficult would it be to simply add a `clean` function here that accepts a step and args?

kyrofa: Eh, I changed my mind on defaulting step to pull at a higher level, otherwise you'll need to treat `None` and `'pull'` as equivalent here. However, I would still like to see a `clean` function here instead of `execute`.

snapcraft/internal/lifecycle.py
```diff
@@ -550,12 +550,7 @@ def _cleanup_internal_snapcraft_dir():
         shutil.rmtree(_SNAPCRAFT_INTERNAL_DIR)
 
 
-def clean(project_options, parts, step=None):
-    # step defaults to None because that's how it comes from docopt when it's
-    # not set.
-    if not step:
-        step = 'pull'
-
+def clean(project_options, parts, step):
```
kyrofa: Why not just default `step='pull'` so it continues working the way it was?

snapcraft/tests/test_lifecycle.py
```diff
@@ -653,7 +653,7 @@ def test_clean_removes_global_state(self):
     plugin: nil
 """)
         lifecycle.execute('pull', self.project_options)
-        lifecycle.clean(self.project_options, parts=None)
+        lifecycle.clean(self.project_options, parts=None, step='pull')
```
kyrofa: If you use `pull` as the default, this is unnecessary.

snapcraft/internal/lxd.py
kalikiana: I reverted the default change now. It's not essential to the PR, it just seemed like a simplification.

And I replaced the subclass' execute logic with a delete method - now lifecycle remains in charge and reading lxd.py doesn't require any knowledge of what lifecycle does with it.

kalikiana:
Tests are green again!

sergiusens:
the business logic seems to have spread a bit too much into the cli components, hasn't it?

kyrofa:
Yes indeed.

kalikiana:
@sergiusens I moved all logic back into the container, in the `clean()` method. Lifecycle knows nothing again like it did before.

kalikiana:
I updated the branch against master (no change in code, but it now lives in `_project.py`).

snapcraft/cli/lifecycle.py
```diff
@@ -175,11 +177,12 @@ def clean(parts, step, **kwargs):
     """
     project_options = get_project_options(**kwargs)
     container_config = env.get_container_config()
+    step = step or 'pull'
```
kyrofa: You can fix LP: [#1727445](https://bugs.launchpad.net/snapcraft/+bug/1727445) right here by not doing this.

PR #1591, opened 2017-10-04
snapd integration tests: print stdout/stderr

kyrofa:

-----

Some snapd integration tests take longer than 10 minutes to run. Since they print no output, Travis thinks they've hung and aborts. Start printing (and capturing) stdout/stderr, both to enable debugging as well as ensure Travis doesn't think a running test has hung.

### DIFF
--- a/integration_tests/__init__.py
+++ b/integration_tests/__init__.py
@@ -124,20 +124,28 @@ class TestCase(testtools.TestCase):
         snapcraft_command = [self.snapcraft_command]
         if debug:
             snapcraft_command.append('-d')
+
+        def _handle_some_stdout(snippet):
+            print(snippet, end='', flush=True)
+
+        def _handle_some_stderr(snippet):
+            print(snippet, end='', flush=True, file=sys.stderr)
+
+        process = subprocess_utils.PopenAsyncOutput(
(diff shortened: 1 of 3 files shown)


kyrofa:
@elopio hitting the maximum time limit here, I guess printing slows things down. Thoughts?

kyrofa:
That particular test actually seems unrelated, it doubled in time, so I suspect it was a network issue. Rerunning. However, the snapd tests (where we actually print stuff) went from 8 to 17 minutes, so printing really does slow things down. Is there no way to simply show a spinner?

kyrofa:
Ah ha, [brilliance](https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received). This PR may not be necessary unless people want it for other reasons.

kyrofa:
Alright, all green here. Thoughts on this? Does it seem useful?

elopio:
I don't like the slow down. But I think we need the output. Wait if we just save it and add it full to the details on error?

kyrofa:
I'm inclined to agree. Let's wait for #1583 and make sure that actually works.

kyrofa:
This was not immediately useful, so I'll close it. Please let me now if it's desired.

PR #1593, opened 2017-10-06
catkin tools plugin: add catkin tools support.

cratliff:
Adding suppot for catkin tools.  Initially the only difference is
the build is being performed by catkin tools instead of catkin_make,
while all other function inherits from catkin_make.

LP: #1693380

-----


snapcraft/plugins/catkin_tools.py
```diff
@@ -0,0 +1,124 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2015-2016 Canonical Ltd
```
elopio: please update the copyright year here.

snapcraft/plugins/catkin_tools.py
```diff
@@ -0,0 +1,124 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class CatkinToolsPlugin(snapcraft.plugins.catkin.CatkinPlugin):
+    def __init__(self, name, options, project):
+        super().__init__(name, options, project)
+        self.build_packages.extend(['python-catkin-tools'])
+
+    def _prepare_build(self):
```
elopio: I feel that this method is too long. Maybe it can be split into `_clean_catkin`, `_add_catkin_profile` and `_install_catkin_packages`

cratliff:
I edited the copyright and broke up the _prepare_build function to be to be a shorter.

That refactor makes sense.  I would be fine to do it now or later.

@kyrofa and I talked about that at the rally.  He had some thoughts about why this should be another plugin.  One I know of is that catkin tools has a number of options that may be added later at the plugin is expanded that wouldn't be compatible with catkin.

snapcraft/plugins/catkin_tools.py
```diff
@@ -0,0 +1,124 @@
...
+    def _clean_catkin(self):
+        # Clean to prevent conflicts with externally built packages.
+        catkincmd = ['catkin', 'clean', '-y']
+
+        self._run_in_bash(catkincmd)
+
+    def _add_catkin_profile(self):
+        # Overwrite the default catkin profile to ensure builds
+        # aren't affected by profile changes.
+        catkincmd = ['catkin', 'profile', 'add', '-f', 'default']
```
kyrofa: Is there any downside to potentially overwriting this profile? Is it possible for the developer to create one with settings that are necessary for the build to succeed, and our overwriting it here causes it to fail? Likewise, I'm sure it's possible for their changes to break the build in the snap... perhaps that's more likely. Just want to get your thoughts on this.

snapcraft/plugins/catkin.py
```diff
@@ -401,7 +401,7 @@ def pull(self):
 
         # Pull our own compilers so we use ones that match up with the version
         # of ROS we're using.
-        compilers = _Compilers(
+        compilers = Compilers(
```
kyrofa: It's on my TODO to split this out into a common area, FYI.

snapcraft/plugins/catkin_tools.py
```diff
@@ -0,0 +1,124 @@
...
+
+        self._run_in_bash(catkincmd)
+
+    def _configure_catkin_profile(self):
+        # Use catkin config to set all configurations before running build.
+        catkincmd = ['catkin', 'config']
+
+        # Use the newly created default profile.
+        catkincmd.extend(['--profile', 'default'])
```
kyrofa: Why not set this profile to be active in `_add_catkin_profile()`? It would save your needing to explicitly use it in every subsequent command, no?

snapcraft/plugins/catkin_tools.py
```diff
@@ -0,0 +1,124 @@
...
+    def _build_catkin_packages(self):
+        # Nothing to do if no packages were specified
+        if not self.catkin_packages:
+            return
+
+        # Call catkin build.
+        catkincmd = ['catkin', 'build']
+
+        # Prevent notification that the build is complete.
+        catkincmd.append('--no-notify')
```
kyrofa: Does this explode on headless systems? Might be kinda useful, really.

snapcraft/plugins/catkin_tools.py
```diff
@@ -0,0 +1,124 @@
...
+        # Use the newly created default profile.
+        catkincmd.extend(['--profile', 'default'])
+
+        catkincmd.extend(self.catkin_packages)
+
+        compilers = Compilers(
+            self._compilers_path, self.PLUGIN_STAGE_SOURCES, self.project)
+
+        # This command must run in bash due to a bug in Catkin that causes it
+        # to explode if there are spaces in the cmake args (which there are).
```
kyrofa: And this bug still bites catkin tools? What about [this bug](https://github.com/catkin/catkin_tools/issues/74) from of old?

snapcraft/tests/commands/test_list_plugins.py
```diff
@@ -28,12 +28,12 @@ class ListPluginsCommandTestCase(CommandBaseTestCase):
...
     # plugin list when wrapper at MAX_CHARACTERS_WRAP
     default_plugin_output = (
-        'ant        cmake  go      gulp     kbuild  maven  '
-        'nodejs             python2  ruby   tar-content\n'
-        'autotools  copy   godeps  jdk      kernel  meson  '
+        'ant        catkin-tools  dump    gradle  jhbuild  '
+        'make   nil                python   qmake  scons      \n'
+        'autotools  cmake         go      gulp    kbuild   '
+        'maven  nodejs             python2  ruby   tar-content\n'
+        'catkin     copy          godeps  jdk     kernel   meson  '
```
kyrofa: Haha, this is always so much fun to edit.

kyrofa:
> However, I'm not entirely convinced that this should be a separate plugin, instead of a single catkin plugin with an attribute like build-tool. But I will just run away and leave that discussion to the two of you :)
>
>> @kyrofa and I talked about that at the rally. He had some thoughts about why this should be another plugin. One I know of is that catkin tools has a number of options that may be added later at the plugin is expanded that wouldn't be compatible with catkin.

I don't have super strong opinions here, but let me lay out all my thoughts.

- The Catkin plugin is stable, and it uses the build tool recommended by ROS.
- Catkin Tools is still in beta, and as such could break API.
- Now in ROS2 they're moving to ament, so I'm not sure how long work will continue to go into Catkin Tools.
- As @cratliff mentions, Catkin Tools is a much more capability-rich tool than Catkin. I feel that combining them in such a way that both Catkin and Catkin Tools are supported would make for a very complex plugin over time as we try to expose various capabilities. And man... it's already complex.
- Having a Catkin Tools plugin means we can label it as a beta plugin.

Long-term, if development continues on Catkin Tools and we don't see breakage between Catkin and Catkin Tools, I envision eventually replacing the Catkin plugin with this. We just need to ensure this plugin continues to be a superset of capabilities from the Catkin plugin, and they don't diverge.

snapcraft/plugins/catkin_tools.py
```diff
@@ -0,0 +1,124 @@
...
+
+import snapcraft
+
+logger = logging.getLogger(__name__)
+
+
+class CatkinToolsPlugin(snapcraft.plugins.catkin.CatkinPlugin):
+    def __init__(self, name, options, project):
+        super().__init__(name, options, project)
+        self.build_packages.extend(['python-catkin-tools'])
```
kyrofa: Could you please add a note that this is a beta plugin? See the [ruby plugin](https://github.com/snapcore/snapcraft/blob/master/snapcraft/plugins/ruby.py#L83) for reference.

snapcraft/plugins/catkin_tools.py
cratliff: As a user I would probably prefer to be able to use in my own profile.  It would mean the catkin tools plugin would be less maintenance as well, since almost anything you would want to add can be done directly through a profile.  My thought was that a user's settings could break the build in the snap if they were utilizing some features, be using source not in the parts directory, or not install properly.  Would a field in the schema be appropriate to allow the user to use their schema understanding it could break things?  Even if we allow the user's profile we would still want to change the source, build, and install space.

snapcraft/plugins/catkin_tools.py
cratliff: Because there was only one subsequent command, I'm good to change that.

snapcraft/plugins/catkin_tools.py
cratliff: Not that I'm aware of.  My thought was that it could be confusing to receive a notification that your build is complete when you still need to go through the stage, prime, and snap steps of the build.

cratliff:
> This is looking pretty good. I'm not a catkin tools pro, so I have a few questions in here. Also, have you  tried it with content sharing at all? Finally, we will definitely need a snapd integration test for this.

No, by content sharing do you mean an overlay or underlay of the workspaces?

> Long-term, if development continues on Catkin Tools and we don't see breakage between Catkin and Catkin Tools, I envision eventually replacing the Catkin plugin with this. We just need to ensure this plugin continues to be a superset of capabilities from the Catkin plugin, and they don't diverge.

My initial transition to catkin tools required cleaning the build from catkin_make then using catkin tools.  Transitioning should be simple enough.

kyrofa:
> No, by content sharing do you mean an overlay or underlay of the workspaces?

Right, take a look at `demos/shared-ros/`. I'd like to see if that continues to work with catkin-tools.

snapcraft/plugins/catkin_tools.py
kyrofa: No need, just a thought.

snapcraft/plugins/catkin_tools.py
kyrofa: Yeah that's a good point.

snapcraft/plugins/catkin_tools.py
kyrofa: Ping here.

snapcraft/plugins/catkin_tools.py
kyrofa: Ping here.

kyrofa:
I'm mostly +1 on this. I'm curious about the bash bug and I'd like it to be marked a beta plugin, but other than that you just have a conflict to fix.

cratliff:
I tested the bash bug and am seeing it.  I also tested content sharing, it did not work, but didn't get a chance to look into why.  Since we had been talking about them I also tested the profile overwriting and cleaning a more.  They might not be necessary and just me being overcautious.  I'll make sure to add the beta plugin.

snapcraft/plugins/catkin_tools.py
kyrofa: Your logic seems sound, here. Indeed, the schema seems a good place for this, but I think what we have here is a good and safe first step. Anything more could come as a follow-up.

kyrofa:
> I tested the bash bug and am seeing it.

So it's _not_ actually fixed in catkin tools? Bah.

>  I also tested content sharing, it did not work, but didn't get a chance to look into why.

That's okay, we can figure it out after this lands.

cratliff:
Added the beta warning and fixed the conflict.

snapcraft/plugins/catkin_tools.py
```diff
@@ -0,0 +1,129 @@
...
+
+        # Beta Warning
+        # Remove this comment and warning once ruby plugin is stable.
```
kyrofa: Check the comment here-- still says "ruby" :stuck_out_tongue: .

kyrofa:
Sorry for the delay @cratliff, our testing infra fell over and we're trying to get a release out the door. As soon as that happens stuff can start landing again.

PR #1596, opened 2017-10-07
tests: move ruby demo test to snapd integration suite

elopio:

-----
This also fixes the command in the ruby-hello snap.

### DIFF
--- a/integration_tests/snapd/test_ruby_snap.py
+++ b/integration_tests/snapd/test_ruby_snap.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright (c) 2016 James Beedy <jamesbeedy@gmail.com>
+# Copyright (C) 2017 Canonical Ltd
 #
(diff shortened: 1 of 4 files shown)


integration_tests/snapd/test_ruby_snap.py
```diff
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright (c) 2016 James Beedy <jamesbeedy@gmail.com>
+# Copyright (C) 2017 Canonical Ltd
```
sergiusens: is it correct to do this?

elopio: I think it is. The test that James wrote is very different from the one I'm proposing here. I didn't move the code, I deleted it and added a new file. I have no idea why github considers it to be the same file in the diff. That's why I'm not using 2016-2017, this is new code, started a couple of weeks ago.

His copyright is preserved in the history of the project, and his legacy with this plugin will be huge :D I think we all have that clear.

PR #1597, opened 2017-10-07
pluginhandler: warn about migrated system libraries

sergiusens:
System libraries from the host where a snap build takes
place are brought into the priming area without
feedback to the user.

Since snapcraft is allowed to build on many systems,
even those where a proper base filesystem does
not exist, we need to warn users when such an
event occurs.

-----

### DIFF
--- a/snapcraft/internal/pluginhandler/__init__.py
+++ b/snapcraft/internal/pluginhandler/__init__.py
@@ -459,16 +459,26 @@ class PluginHandler:
 
         dependency_paths = part_dependency_paths | staged_dependency_paths
 
-        if not self._build_attributes.no_system_libraries():
+        if not self._build_attributes.no_system_libraries() and system:
             system_dependency_paths = {os.path.dirname(d) for d in system}
             dependency_paths.update(system_dependency_paths)
 
-            if system:
-                # Lots of dependencies are linked with a symlink, so we need to
-                # make sure we follow those symlinks when we migrate the
-                # dependencies.
-                _migrate_files(system, system_dependency_paths, '/',
-                               self.primedir, follow_symlinks=True)
+            logger.warning(
+                'The following files from the build host will migrate to the '
+                'snap to\nstatisfy library dependencies:\n\n{}\n'
+                'This behavior is problematic, these dependent libraries '
+                'should be provided\nthrough stage-packages entries or other '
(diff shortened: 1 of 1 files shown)


sergiusens:
I am not a fan of the `build-attributes` exposure as eventually everyone will just have it in their yaml though :-/

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -459,16 +459,26 @@ def prime(self, force=False):
...
-                _migrate_files(system, system_dependency_paths, '/',
-                               self.primedir, follow_symlinks=True)
+            logger.warning(
+                'The following files from the build host will migrate to the '
+                'snap to\nstatisfy library dependencies:\n\n{}\n'
+                'This behavior is problematic, these dependent libraries '
+                'should be provided\nthrough stage-packages entries or other '
+                'parts.\n'
+                'You can disable this behavior in the part by setting:'
+                '\n    <part-name>:'
```
kyrofa: We have access to the actual part name here, no? Would be slick to use that instead of this placeholder.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -459,16 +459,26 @@ def prime(self, force=False):
...
-            if system:
-                # Lots of dependencies are linked with a symlink, so we need to
-                # make sure we follow those symlinks when we migrate the
-                # dependencies.
-                _migrate_files(system, system_dependency_paths, '/',
-                               self.primedir, follow_symlinks=True)
+            logger.warning(
+                'The following files from the build host will migrate to the '
+                'snap to\nstatisfy library dependencies:\n\n{}\n'
+                'This behavior is problematic, these dependent libraries '
```
kyrofa: This behavior is only problematic because it's unexpected, right? If there are other reasons, it might be worth mentioning here.

Also, should this actually be a deprecation notice?

snapcraft/internal/pluginhandler/__init__.py
sergiusens: good point

snapcraft/internal/pluginhandler/__init__.py
sergiusens: another good point

PR #1600, opened 2017-10-10
options: fix core-dynamic-linker on ppc64el/s390x

cjwatson:
ppc64el and s390x had leading slashes on their settings for
`core-dynamic-linker`.  In some cases this caused snapcraft to
incorrectly believe that the core snap was not installed, as the target
of the symlink on the host system couldn't be resolved within the core
snap.

LP: #1722521


snapcraft/tests/project_loader/test_config.py
```diff
@@ -1626,6 +1626,7 @@ def test_config_stage_environment_confinement_classic(self):
             'snapcraft._options.ProjectOptions.get_core_dynamic_linker')
         mock_core_dynamic_linker = patcher.start()
         mock_core_dynamic_linker.return_value = dynamic_linker
+        self.addCleanup(patcher.stop)
```
elopio: good catch there, thank you.

PR #1601, opened 2017-10-10
snapcraft.yaml: don't re-use build dir

kyrofa:

-----

Creating the snapcraft snap works fine as long as one is using the snap or the deb to build. However, for reasons unknown to me, when running from source, the `make startup` of the `apt` part creates a directory named `build` that clashes with the `build` scriptlet that also tries to create a directory named `build`. I'm still not sure why this only happens when building from source, but an easy fix is to simply create a different directory in the `build` scriptlet.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -68,12 +68,12 @@ parts:
     prepare: |
         make startup
     build: |
-        mkdir build
-        cd build
+        mkdir apt-build
+        cd apt-build
         ../configure
         make
     install: |
-        cd build
+        cd apt-build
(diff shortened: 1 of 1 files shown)


kyrofa:
> no idea where it came from.

Haha, I STILL have no idea what's going on :stuck_out_tongue: .

sergiusens:
Can we try and rename `build` to `apt-build` first a first cut without the `-p` and see how it goes?

kyrofa:
Some further investigation:

`make startup` creates the `build` dir. But apparently only if running from source, which I still can't explain. However, @sergiusens yes, using `apt-build` works and is likely a better idea indeed. Done.

PR #1602, opened 2017-10-10
tests: add the slow tag for ros snapd integration test

elopio:

-----
Requires #1603


sergiusens:
can someone re-trigger the failing test jobs?

kalikiana:
@sergiusens This won't pass on `zesty` without #1614.

`ERROR: test_catkin_shared_ros.CatkinSharedRosTestCase.test_shared_ros_builds_without_catkin_in_underlay
[...]
apt.cache.FetchFailedException: W:GPG error: http://packages.ros.org/ros/ubuntu xenial InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 5523BAEEB01FA116, E:The repository 'http://packages.ros.org/ros/ubuntu xenial InRelease' is not signed.
[...]`

kyrofa:
We're skipping instead, as in #1621, which has now landed. I just updated the branch, and will re-run the tests.

TESTING.md
```diff
@@ -33,10 +33,16 @@ These tests are in the `snapcraft/tests` directory.
...
-This suite was split in three: plugins, store and other integration tests. This split is artificial, we made it just because the full suite takes more time than what Travis allows for a single job.
+This suite was split in four: plugins, store, snapd and other integration tests. This split is artificial, we made it just because the full suite takes more time than what Travis allows for a single job.
 
 These tests are in the `integration_tests` directory, with the `snapcraft.yamls` and other source files for the tests snaps in `integration_tests/snaps`.
 
+### Slow tests
+
+Some tests take too long. This affects the pull requests because we have to wait for a long time, and they will make Travis CI timeout because we have only 50 minutes per suite in there. The solution is to tag these tests as slow, and don't run them in all pull requests. These tests will only be run in autopkgtests.
+
+To mark a test as slow, set `slow_test = True` before the test `setUp`.
```
sergiusens: all implementations in here set it as a class attribute instead of in `setUp`. While working it is confusing and potentially error prone.

elopio: I think it makes it less error prone to put it as a class attribute, because that assignment will always run before setUp.

From your comment, it's not clear what's your preference. I could document that the slow_test = True must be set in a class attribute, which is not totally correct, but could make it easier to understand.

PR #1607, opened 2017-10-11
python plugin: use extracted pip

kyrofa:

-----

Simplify the Python plugin and its tests by using the independently tested pip that was extracted in previous commits. All functionality should remain the same. Please help me test this PR thoroughly!

### DIFF
--- a/snapcraft/plugins/_python/_pip.py
+++ b/snapcraft/plugins/_python/_pip.py
@@ -27,6 +27,7 @@ import tempfile
 
 import snapcraft
 from snapcraft import file_utils
+from snapcraft.internal import mangling
 from ._python_finder import (
     get_python_command,
     get_python_headers,
@@ -65,11 +66,18 @@ def _replicate_owner_mode(path):
         return
 
     file_mode = os.stat(path).st_mode
-    new_mode = file_mode & stat.S_IWUSR
+
+    # We at least need to write to it to fix shebangs later
+    new_mode = file_mode | stat.S_IWUSR
+
+    # If the owner can execute it, so should everyone else.
     if file_mode & stat.S_IXUSR:
         new_mode |= stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH
+
+    # If the owner can read it, so should everyone else
     if file_mode & stat.S_IRUSR:
         new_mode |= stat.S_IRUSR | stat.S_IRGRP | stat.S_IROTH
+
     os.chmod(path, new_mode)
 
 
@@ -129,10 +137,23 @@ class Pip:
         Check to see if pip has already been installed. If not, fetch pip,
         setuptools, and wheel, and install them so they can be used.
         """
-        # Check to see if we have our own pip, yet. If not, we need to use the
-        # pip on the host (installed via build-packages) to grab our own.
-        if not self.is_setup():
-            logger.info('Fetching pip, setuptools, and wheel...')
+
+        self._ensure_pip_installed()
+        self._ensure_wheel_installed()
+        self._ensure_setuptools_installed()
+
+    def is_setup(self):
+        """Return true if this class has already been setup."""
+
+        return (
+            self._is_pip_installed() and self._is_wheel_installed() and
+            self._is_setuptools_installed())
+
+    def _ensure_pip_installed(self):
+        # Check to see if we have our own pip. If not, we need to use the pip
+        # on the host (installed via build-packages) to grab our own.
+        if not self._is_pip_installed():
+            logger.info('Fetching and installing pip...')
 
             real_python_home = self._python_home
 
@@ -142,33 +163,49 @@ class Pip:
             try:
                 self._python_home = os.path.join(os.path.sep, 'usr')
 
-                # Using the host's pip, install our own pip and other tools we
-                # need.
-                self.download({'pip', 'setuptools', 'wheel'})
-                self.install(
-                    {'pip', 'setuptools', 'wheel'}, ignore_installed=True)
+                # Using the host's pip, install our own pip
+                self.download({'pip'})
+                self.install({'pip'}, ignore_installed=True)
             finally:
                 # Now that we have our own pip, reset the python home
                 self._python_home = real_python_home
 
-    def is_setup(self):
-        """Return true if this class has already been setup."""
+    def _ensure_wheel_installed(self):
+        if not self._is_wheel_installed():
+            logger.info('Fetching and installing wheel...')
+            self.download({'wheel'})
+            self.install({'wheel'}, ignore_installed=True)
 
+    def _ensure_setuptools_installed(self):
+        if not self._is_setuptools_installed():
+            logger.info('Fetching and installing setuptools...')
+            self.download({'setuptools'})
+            self.install({'setuptools'}, ignore_installed=True)
+
+    def _is_pip_installed(self):
         try:
             # We're expecting an error here at least once complaining about
             # pip not being installed. In order to verify that the error is the
             # one we think it is, we need to process the stderr. So we'll
             # redirect it to stdout. If it's not the error we expect, something
             # is wrong, so re-raise it.
-            self._run([], stderr=subprocess.STDOUT)
+            #
+            # Using _run_output here so stdout doesn't get printed to the
+            # terminal.
+            self._run_output([], stderr=subprocess.STDOUT)
         except subprocess.CalledProcessError as e:
             output = e.output.decode(sys.getfilesystemencoding()).strip()
             if 'no module named pip' in output.lower():
                 return False
             else:
                 raise e
-
         return True
+
+    def _is_wheel_installed(self):
+        return 'wheel' in self.list()
+
+    def _is_setuptools_installed(self):
+        return 'setuptools' in self.list()
 
     def download(self, packages, *, setup_py_dir=None, constraints=None,
                  requirements=None, process_dependency_links=False):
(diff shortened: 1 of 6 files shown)


snapcraft/plugins/_python/_pip.py
```diff
@@ -65,11 +66,18 @@ def _replicate_owner_mode(path):
...
+    new_mode = file_mode | stat.S_IWUSR
+
+    # If the owner can execute it, so should everyone else.
     if file_mode & stat.S_IXUSR:
         new_mode |= stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH
+
+    # If the owner can read it, so should everyone else
     if file_mode & stat.S_IRUSR:
         new_mode |= stat.S_IRUSR | stat.S_IRGRP | stat.S_IROTH
+
```
elopio: Thank you for the explanations!

snapcraft/plugins/_python/_pip.py
```diff
@@ -129,10 +137,23 @@ def setup(self):
...
-        # Check to see if we have our own pip, yet. If not, we need to use the
-        # pip on the host (installed via build-packages) to grab our own.
-        if not self.is_setup():
-            logger.info('Fetching pip, setuptools, and wheel...')
+
+        self._ensure_pip_installed()
+        self._ensure_wheel_installed()
+        self._ensure_setuptools_installed()
+
+    def is_setup(self):
```
elopio: Would this be better: `was_setup` or `is_already_setup`?
is_setup sounds to me like checking if this is of type `setup` or not. Not if `setup` happened.

snapcraft/plugins/_python/_pip.py
```diff
@@ -129,10 +137,23 @@ def setup(self):
...
+
+        return (
+            self._is_pip_installed() and self._is_wheel_installed() and
+            self._is_setuptools_installed())
+
+    def _ensure_pip_installed(self):
+        # Check to see if we have our own pip. If not, we need to use the pip
+        # on the host (installed via build-packages) to grab our own.
+        if not self._is_pip_installed():
+            logger.info('Fetching pip...')
```
elopio: A nit, but maybe this log entry would be more accurate as `Fetching and installing pip`. Or have two log entries, one for download and one for install.

snapcraft/plugins/_python/_pip.py
```diff
@@ -142,34 +163,50 @@ def setup(self):
...
+                self.install({'pip'}, ignore_installed=True)
             finally:
                 # Now that we have our own pip, reset the python home
                 self._python_home = real_python_home
 
-    def is_setup(self):
-        """Return true if this class has already been setup."""
+    def _ensure_wheel_installed(self):
+        if not self._is_wheel_installed():
+            logger.info('Fetching wheel...')
```
elopio: same comment as the one for the pip log. This fetches and installs.

snapcraft/plugins/_python/_pip.py
```diff
@@ -142,34 +163,50 @@ def setup(self):
...
+            self.download({'wheel'})
+            self.install({'wheel'}, ignore_installed=True)
+
+    def _ensure_setuptools_installed(self):
+        if not self._is_setuptools_installed():
+            logger.info('Fetching setuptools...')
```
elopio: same comment as the one for the pip log. This fetches and installs.

snapcraft/plugins/python.py
```diff
@@ -53,40 +53,25 @@
...
+import contextlib
 import os
 import re
-import shutil
 import stat
-import subprocess
-import tempfile
-from contextlib import contextmanager, suppress
-from glob import glob
+from contextlib import contextmanager
```
sergiusens: you already `import contextlib` above

snapcraft/plugins/python.py
```diff
@@ -365,14 +213,100 @@ def build(self):
...
+
+        # we want to avoid installing what is already provided in
+        # stage-packages
+        need_install = [k for k in wheel_names if k not in installed]
+        self._pip.install(
+            need_install + self.options.python_packages, upgrade=True,
+            install_deps=False,
+            process_dependency_links=self.options.process_dependency_links)
+
+    def _download_from_setup_py(self):
```
sergiusens: I am confused by this method name, should it not be something closer to `download_project`?

snapcraft/plugins/_python/_pip.py
kyrofa: That feels a little overly verbose to me, but I have no strong feelings about it. However, this API was already in place and is used elsewhere, so in the interest of not further inflating this diff I'd rather not make that change here.

snapcraft/plugins/python.py
kyrofa: Haha, I obviously got into the weeds with that. `{download,install}_project` is obviously a far better choice, thank you. Done.

snapcraft/plugins/python.py
kyrofa: Well what do you know! Fixed, thank you.

snapcraft/plugins/_python/_pip.py
kyrofa: Yeah good call, I went with "fetching and installing" instead of multiple prints. Done.

snapcraft/plugins/_python/_pip.py
kyrofa: Fixed.

snapcraft/plugins/_python/_pip.py
kyrofa: Fixed.

snapcraft/plugins/_python/_pip.py
kyrofa: Yeah I always have to look those things up.

kyrofa:
> We haven't yet agreed fully on the tests, but your refactor here is a big step forward.

Thanks @elopio. The dust settles after this PR, so we can dig back into testing.

kyrofa:
The plugins integration suite is consistently running over here. We need a solution to that before this can land.

sergiusens:
@kyrofa why do you think it is running over for this PR and not others? This worries me a bit, since if it is because the python stuff is taking longer, then maybe the new implementation has lost some caching advantages or the execution logic grew without us knowing

kyrofa:
> @kyrofa why do you think it is running over for this PR and not others?

It actually _is_ running over on several other PRs as well. @kalikiana has been battling the plugins suite all week.

snapcraft/plugins/_python/_pip.py
```diff
@@ -374,9 +416,18 @@ def env(self):
 
         return env
 
-    def _run(self, args, **kwargs):
+    def _run(self, args, runner=None, **kwargs):
         env = self.env()
 
-        return snapcraft.internal.common.run_output(
+        # Using None as the default value instead of common.run so we can mock
```
kalikiana: How about importing `run_output` by name so it can be mocked "normally"? If that's possible I would slightly prefer it, since adding options for the sole purpose of testing seems less desirable.

kyrofa: The option here wasn't added for testing, it was added so we didn't have two run functions doing essentially the same thing but with different runners, which would introduce an opportunity to diverge. Normally this function declaration would look like `def _run(self, args, runner=snapcraft.internal.common.run, **kwargs)`, but default arguments are evaluated at import time, not run time, thus it would still call the actual `snapcraft.internal.common.run` even if it was mocked. So I used `None` instead and wrote that comment.

kalikiana: Ah. That makes sense. Thanks for clarifying.

sergiusens:
Seems a random IO error killed the tests, will update branch to retrigger.

sergiusens:
@kyrofa can you add some notes on these failures, they don't seem related to the test suite themselves, do they work locally when run in a container?

kyrofa:
@sergiusens huh, I have never seen those before, they look temporary-- I retriggered. Mostly what we've been battling here is LP: [#1717921](https://bugs.launchpad.net/snapcraft/+bug/1717921), whose cause is unknown.

kyrofa:
Huh.

> error: cannot open: "snaps-cache/snapcraft-pr1607.snap"

Could this be due to #1625 landing?

sergiusens:
On vie, oct 27, 2017 at 12:25 PM, Kyle Fazzari 
> Huh.
> 
>> error: cannot open: "snaps-cache/snapcraft-pr1607.snap"
>> 
> Could this be due to #1625 landing?
> 

Yes, and it might require the follow up branch from Leo; now this all 
makes sense a nicer warning on the #1625 would have been nice if this 
was the case.

kyrofa:
Trying to switch out the test runner here-- we're still using unittest, whereas testtools has a different runner. Let's see what happens.

sergiusens:
Can you add a summary or the diff of code that made this problem go away?

kyrofa:
@sergiusens right, of course. Sorry, I tore this branch apart while investigating and the history was bothering me.

The issue came down to [this line](https://github.com/snapcore/snapcraft/pull/1607/files#diff-fda09d10795764cb4492ad3872109644R205), where I changed `packages` to `python-packages` but didn't update the integration test.

The tests have now passed twice in a row after not passing for the past few weeks, so I'm pretty confident in them at this point.

snapcraft/plugins/python.py
```diff
@@ -159,219 +143,173 @@ def plugin_stage_packages(self):
...
+
     def pull(self):
         super().pull()
 
-        setup = 'setup.py'
-        if os.listdir(self.sourcedir):
-            setup = os.path.join(self.sourcedir, 'setup.py')
+        self._pip.setup()
+
         with simple_env_bzr(os.path.join(self.installdir, 'bin')):
```
sergiusens: this might not (certainly since the start of our use of sitecustomize) is not needed anymore.

PR #1614, opened 2017-10-12
apt repo: allow insecure repos

kyrofa:

-----

In Xenial this was the default behavior, which explains why Snapcraft works. However, in Zesty this behavior changed, causing breakage when using external repositories (e.g. ROS). Instead of relying on defaults, ensure the behavior is consistent across releases.

This is not a long-term fix. We should add a way to ack repository keys.

### DIFF
--- a/snapcraft/internal/repo/_deb.py
+++ b/snapcraft/internal/repo/_deb.py
@@ -66,6 +66,13 @@ class _AptCache:
         # Do not install recommends
         apt.apt_pkg.config.set('Apt::Install-Recommends', 'False')
 
+        # Don't verify repository keys. FIXME: We should have a way to simply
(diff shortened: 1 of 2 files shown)


kyrofa:
> Maybe log bug report to follow up on how to add keys in the future?

Yeah we need to have a discussion about it. Here's a [forum post](https://forum.snapcraft.io/t/insecure-apt-repos/2464).

kyrofa:
@sergiusens this is causing autopkgtests to fail on zesty and artful today. We need to design a key-handling solution to use the same default. How would you like to deal with the autopkgtests?

kyrofa:
Alright we're going to skip instead.

PR #1617, opened 2017-10-13
Add options to configure applications socket activation

albertodonato:
This adds schema and rendering for the `sockets` entries in `apps`, to configure socket activations for deamons. See https://github.com/snapcore/snapd/pull/3916 for the snapd implementation

### DIFF
--- a/schema/snapcraft.yaml
+++ b/schema/snapcraft.yaml
@@ -249,6 +249,30 @@ properties:
             description: What kind of wrapper to generate for the given command
             enum:
               - none
+          sockets:
+            type: object
+            additionalProperties: false
+            validation-failure:
+              "{!r} is not a valid socket name. Socket names consist of
+              lower-case alphanumeric characters and hyphens."
+            patternProperties:
+              "^[a-z][a-z0-9_-]*$":
+                type: object
+                required:
+                  - listen-stream
+                description: Sockets for automatic service activation
+                additionalProperties: false
+                properties:
+                  listen-stream:
+                    anyOf:
+                      - type: integer
+                        usage: "port number, an integer between 1 and 65535"
+                        minimum: 1
+                        maximum: 65535
+                      - type: string
+                        usage: "socket path, a string"
+                  socket-mode:
+                    type: integer
   hooks:
     type: object
     additionalProperties: false
--- a/snapcraft/internal/meta/_snap_packaging.py
+++ b/snapcraft/internal/meta/_snap_packaging.py
@@ -73,6 +73,17 @@ _OPTIONAL_PACKAGE_KEYS = [
 ]
 
 
+class OctInt(int):
+    """An int represented in octal form."""
+
+
+def oct_int_representer(dumper, data):
+    return yaml.ScalarNode('tag:yaml.org,2002:int', '{:04o}'.format(data))
+
+
+yaml.add_representer(OctInt, oct_int_representer)
+
+
 def create_snap_packaging(
         config_data: Dict[str, Any],
         parts_config: project_loader.PartsConfig,
@@ -400,6 +411,7 @@ class _SnapPackaging:
         if 'apps' in self._config_data:
             _verify_app_paths(basedir='prime', apps=self._config_data['apps'])
             snap_yaml['apps'] = self._wrap_apps(self._config_data['apps'])
+            snap_yaml['apps'] = self._render_socket_modes(snap_yaml['apps'])
 
         return snap_yaml
 
@@ -515,6 +527,15 @@ class _SnapPackaging:
                 snap_name=self._config_data['name'], prime_dir=self._prime_dir)
             desktop_file.parse_and_reformat()
             desktop_file.write(gui_dir=os.path.join(self.meta_dir, 'gui'))
+
+    def _render_socket_modes(self, apps):
+        for app in apps.values():
+            sockets = app.get('sockets', {})
+            for socket in sockets.values():
+                mode = socket.get('socket-mode')
+                if mode is not None:
+                    socket['socket-mode'] = OctInt(mode)
+        return apps
 
 
 def _find_bin(binary, basedir):
--- a/snapcraft/tests/unit/test_meta.py
+++ b/snapcraft/tests/unit/test_meta.py
@@ -282,6 +282,22 @@ class CreateTestCase(CreateBaseTestCase):
         self.assertFalse('icon' in y,
                          'icon found in snap.yaml {}'.format(y))
 
+    def test_create_meta_with_sockets(self):
+        os.mkdir(self.prime_dir)
+        _create_file(os.path.join(self.prime_dir, 'app.sh'))
+        sockets = {
+            'sock1': {
+                'listen-stream': 8080,
+            },
+            'sock2': {
+                'listen-stream': '$SNAP_COMMON/sock2',
+                'socket-mode': 0o640}}
+        self.config_data['apps'] = {
+            'app': {'command': 'app.sh',
+                    'sockets':  sockets}}
+        y = self.generate_meta_yaml()
+        self.assertThat(y['apps']['app']['sockets'], Equals(sockets))
+
     def test_version_script(self):
         self.config_data['version-script'] = 'echo 10.1-devel'
 


schema/snapcraft.yaml
```diff
@@ -199,6 +199,25 @@ properties:
...
+                  - listen-stream
+                description: Sockets for automatic service activation
+                additionalProperties: false
+                properties:
+                  listen-stream:
+                    anyOf:
+                      - type: integer
+                        minimum: 1
+                        maximum: 65535
+                      - type: string
```
kyrofa: As it is, let's say I use a value that doesn't match this schema, say, `listen-stream: true`. That will result in the following dreadfully unhelpful error:

```
Issues while validating snapcraft.yaml: The 'apps/foo/sockets/bar/listen-stream' property does not match the required schema: True is not valid under any of the given schemas
```

This is of course not your fault, but it's something we're trying to improve. For example, if you change it to e.g.

```
properties:
  listen-stream:
    anyOf:
      - type: integer
        usage: "port number, an integer between 1 and 65535"
        minimum: 1
        maximum: 65535
      - type: string
        usage: "socket path, a string"
```

There error message is immediately more helpful:

```
Issues while validating snapcraft.yaml: The 'apps/foo/sockets/bar/listen-stream' property does not match the required schema: True is not valid under any of the given schemas (must be one of 'port number, an integer between 1 and 65535' or 'socket path, a string')
```

snapcraft/internal/meta.py
```diff
@@ -408,6 +420,15 @@ def _generate_desktop_file(self, name, app):
             desktop_file.parse_and_reformat()
             desktop_file.write(gui_dir=os.path.join(self.meta_dir, 'gui'))
 
+    def _render_socket_modes(self, apps):
+        for app in apps.values():
+            sockets = app.get('sockets', {})
+            for socket in sockets.values():
+                mode = socket.get('socket-mode')
+                if mode is not None:
+                    socket['socket-mode'] = OctInt(mode)
```
kyrofa: Nice job keeping the octal representation, here.

kyrofa:
There are some socket options documented here: https://github.com/canonical-docs/snappy-docs/blob/master/build-snaps/syntax.md . It looks like that stuff needs to be updated with this, if you're up for it.

schema/snapcraft.yaml
albertodonato: Nice, I didn't know about this. changed as suggested

schema/snapcraft.yaml
```diff
@@ -199,6 +199,25 @@ properties:
             description: What kind of wrapper to generate for the given command
             enum:
               - none
+          sockets:
+            type: object
+            additionalProperties: false
+            patternProperties:
+              "^[a-zA-Z0-9](?:-?[a-zA-Z0-9])*$":
```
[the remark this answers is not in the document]
albertodonato: Fixed to use the same regexp as the one for the snap name (as the snapd PR does)

albertodonato:
@kyrofa thanks for the review. All comments should be addressed, I will make a PR for the documentation update once it (and the snapd counterpart) lands

albertodonato:
@kyrofa WRT a snap using the `sockets` options with a snapd that doesn't support it, I think the option will be silently ignored

schema/snapcraft.yaml
```diff
@@ -202,6 +202,27 @@ properties:
             description: What kind of wrapper to generate for the given command
             enum:
               - none
+          sockets:
+            type: object
+            additionalProperties: false
+            patternProperties:
+              "^[a-z0-9][a-z0-9+-]*$":
```
kyrofa: This pattern is supposed to correspond to [this](https://github.com/snapcore/snapd/blob/61ee25526802e6866c7645f53812b43eedc521dc/snap/validate.go#L32), no? The pattern I see here would allow _just_ numbers, for example.

schema/snapcraft.yaml
```diff
@@ -202,6 +202,27 @@ properties:
             description: What kind of wrapper to generate for the given command
             enum:
               - none
+          sockets:
+            type: object
+            additionalProperties: false
```
kyrofa: Now that #1615 has landed, I have another suggestion. When the schema is:

1. An object
2. With no additional properties allowed
3. That uses `patternProperties` rather than hard-coded properties

... and is handed an object with properties that do not match the pattern, rather than barfing on the pattern, jsonschema unhelpfully barfs on the `additionalProperties: false` instead. So a YAML like this:

```
apps:
  foo:
    command: echo "hello"
    sockets:
      BAD_SOCKET:
        listen-stream: bar
```
Results in this:
```
Issues while validating snapcraft.yaml: The 'apps/foo/sockets' property does not match the required schema: Additional properties are not allowed ('BAD_SOCKET' was unexpected)
```
That can be improved if you add a `validation-failure` message like this:
```
sockets:
  type: object
  additionalProperties: false
  validation-failure:
    "{!r} is not a valid socket name. Socket names consist of
    lower-case alphanumeric characters and hyphens."
  patternProperties:
    "^[a-z0-9][a-z0-9+-]*$":
      <snip>
```
Then the error message becomes:
```
Issues while validating snapcraft.yaml: The 'apps/foo/sockets' property does not match the required schema: 'BAD_SOCKET' is not a valid socket name. Socket names consist of lower-case alphanumeric characters and hyphens.
```

schema/snapcraft.yaml
albertodonato: Yes, it should correspondo to that. I copied this from the snap `name`, attribute; doesn't that mean that a snap name with just numbers would also be allowed?

schema/snapcraft.yaml
albertodonato: updated schema as suggested

schema/snapcraft.yaml
kyrofa: Yes indeed, it appears snapcraft and snapd have diverged on this point. I suggest using the correct pattern here, I'll fix the snap name in another PR.

sergiusens:
On mar, oct 17, 2017 at 11:45 AM, Kyle Fazzari 
> @kyrofa approved this pull request.
> 
> GitHub is suffering from a queue backup this morning, but once tests 
> pass here, +1 from me. Thanks @albertodonato 😃 .
> 

Thanks, while the fix is correct, given that we are not adding an 
`assumes` entry when using this keyword, I am not confident of landing 
this before it is generally available in all distros.
>

kyrofa:
> Thanks, while the fix is correct, given that we are not adding an `assumes` entry when using this keyword, I am not confident of landing this before it is generally available in all distros.

Fine by me, although I feel like that's the developer's job, not ours. We don't add assumes for hooks, either, but we support creating snaps that use them.

sergiusens:
That is a maintenance burden and we can and should do better. We have not so far done this kind of thing where we released a version of snapcraft before the store or snapd side of the feature was available.

albertodonato:
@sergiusens @kyrofa can this be merged now that the snapd change has landed?

sergiusens:
On mar, nov 14, 2017 at 4:36 PM, Alberto Donato 
> @sergiusens @kyrofa can this be merged now that the snapd change has 
> landed?\
> 
in stable?
>

albertodonato:
no, it was merged to master

kyrofa:
@albertodonato as @sergiusens mentioned above, we don't want to support this until the feature is available in a stable snapd release.

pvanloo:
The snapd feature should now be stable.

https://forum.snapcraft.io/t/the-snapd-roadmap/1973

Would be great if this can be merged!

kalikiana:
For the record, the unit tests failed with this error:

    ERROR: test_create_meta_with_sockets (snapcraft.tests.unit.test_meta.CreateTestCase)
    snapcraft.tests.unit.test_meta.CreateTestCase.test_create_meta_with_sockets
    [...]
    snapcraft.internal.project_loader.errors.YamlValidationError: Issues while validating snapcraft.yaml: The 'apps/app/sockets/sock1' property does not match the required schema: 'listen-stream' is a required property

sergiusens:
Before this lands I would like to see a PR with documentation for https://docs.snapcraft.io/

albertodonato:
@sergiusens That's merged already: https://github.com/canonical-docs/snappy-docs/commit/eb1ce5e62527bc2b8d8aa3162081594ce934b788

sergiusens:
Great!

kyrofa:
@sergiusens this is green, are you good with it?

PR #1624, opened 2017-10-17
kernel plugin: use latest stable core snap

alfonsosanchezbeato:
The plugin was downloading ubuntu-core instead of core, which is
nowadays a very old version of the later.


kyrofa:
Thanks @alfonsosanchezbeato! This will need the eyes of @sergiusens. Changing to `core` seems obvious, but I don't remember the argument for using `edge` over `stable`.

sergiusens:
Yeah, the only reason this wasn't done faster was that I was hoping for the initrd split to happen sooner.

sergiusens:
If you want to level-up on the task, maybe also update to use the correct path in the os (core, base) snap -> [LP: #1567564](https://bugs.launchpad.net/snapcraft/+bug/1567564)

sergiusens:
On mar, oct 17, 2017 at 12:28 PM, Oliver Grawert 
> @ogra1 approved this pull request.
> 
> we might perhaps want to keep the same channel here ... to not make 
> the turnaround time for script fixes longer than they have to be ...
> 
> beyond that ... pretty please merge this... thats a gross oversight 
> ...
> 

Well, choose, merge it now or change the channel and then merge? ;-)
>

kyrofa:
> we might perhaps want to keep the same channel here ... to not make the turnaround time for script fixes longer than they have to be ...

That's a two-way street-- breakages come faster, too. I'd personally like to see `stable` used, here.

alfonsosanchezbeato:
I am more on the side of using stable too... it is not fun to have a build suddenly breaking because core's daily build is broken.

alfonsosanchezbeato:
Updated to try to fix the integration tests.

ogra1:
well, being democratically overruled, lets merge it with "stable" ... someone with permission needs to do that though ...

PR #1625, opened 2017-10-17
tests: use the snapcraft snap for integration tests

elopio:

-----

### DIFF
--- a/tools/travis/build_snapcraft_snap.sh
+++ b/tools/travis/build_snapcraft_snap.sh
@@ -31,6 +31,8 @@ $lxc exec snap-builder -- sh -c "snap install core" || echo "ignored error"
 $lxc file push --recursive $project_path snap-builder/root/
 # TODO use the stable snap once it's published.
 $lxc exec snap-builder -- sh -c "snap install snapcraft --candidate --classic"
-$lxc exec snap-builder -- sh -c "cd snapcraft && /snap/bin/snapcraft"
-
+$lxc exec snap-builder -- sh -c "cd snapcraft && /snap/bin/snapcraft snap --output snapcraft-pr$TRAVIS_PULL_REQUEST.snap"
+# Pull the snap from the container to save it into the cache.
+mkdir -p "$TRAVIS_BUILD_DIR/snaps-cache"
+$lxc file pull "snap-builder/root/snapcraft/snapcraft-pr$TRAVIS_PULL_REQUEST.snap" "$TRAVIS_BUILD_DIR/snaps-cache/"
 $lxc stop snap-builder
--- a/tools/travis/run_tests.sh
+++ b/tools/travis/run_tests.sh
@@ -34,7 +34,10 @@ if [ "$test" = "static" ]; then
 elif [ "$test" = "unit" ]; then
     dependencies="apt install -y git bzr subversion mercurial libnacl-dev libsodium-dev libffi-dev libapt-pkg-dev libarchive-dev python3-pip squashfs-tools xdelta3 && python3 -m pip install -r requirements-devel.txt -r requirements.txt codecov && apt install -y python3-coverage"
 elif [ "$test" = "integration" ] || [ "$test" = "plugins" ] || [ "$test" = "store" ] || [ "$test" = "containers" ] || [ "$test" = "snapd" ]; then
-    dependencies="apt install -y bzr curl git libnacl-dev libsodium-dev libffi-dev libapt-pkg-dev libarchive-dev mercurial python3-pip subversion squashfs-tools sudo snapd xdelta3 && python3 -m pip install -r requirements-devel.txt -r requirements.txt"
+    # snap install core exits with this error message:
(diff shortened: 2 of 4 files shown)


elopio:
This is looking good! Please give it a good review, and think about the implications because things might break if some of my assumptions are not good.

I think that the integration tests should run against the installed snap, because that's closer to what our users will get. And this opens the doors to my next experimental PR, coming soon.

tools/travis/build_snapcraft_snap.sh
```diff
@@ -28,9 +28,14 @@ lxc="/snap/bin/lxc"
 # Workaround for
 # - Setup snap "core" (2462) security profiles (cannot reload udev rules: exit status 2
 $lxc exec snap-builder -- sh -c "snap install core" || echo "ignored error"
+# Simplify the version to use it in the following stages.
+sed -i "s/version: git/version: test/g" "$project_path/snap/snapcraft.yaml"
```
sergiusens: If you are doing this to have a consistent filename, let me suggest using `--output` instead.

sergiusens:
The proposal looks good and adds one more level of security for the snap, so +1 from we on the approach.

tools/travis/build_snapcraft_snap.sh
elopio: good call.

kalikiana:
For the record, second unit test stage failed with `runtime: address space conflict: map(0xc820000000) = 0x7f095173e000
fatal error: runtime: address space conflict
runtime stack:
runtime.throw(0xc8200fd3e0, 0x1f)
	/usr/lib/go-1.6/src/runtime/panic.go:547 +0x92 fp=0x7fff70cde4d0 sp=0x7fff70cde4b8
[...]
	/usr/lib/go-1.6/src/runtime/asm_amd64.s:138 +0x134 fp=0x7fff70cde9e8 sp=0x7fff70cde9e0`

tools/travis/run_tests.sh
```diff
@@ -34,7 +34,9 @@ if [ "$test" = "static" ]; then
 elif [ "$test" = "unit" ]; then
     dependencies="apt install -y git bzr subversion mercurial libnacl-dev libsodium-dev libffi-dev libapt-pkg-dev libarchive-dev python3-pip squashfs-tools xdelta3 && python3 -m pip install -r requirements-devel.txt -r requirements.txt codecov && apt install -y python3-coverage"
 elif [ "$test" = "integration" ] || [ "$test" = "plugins" ] || [ "$test" = "store" ] || [ "$test" = "containers" ] || [ "$test" = "snapd" ]; then
-    dependencies="apt install -y bzr curl git libnacl-dev libsodium-dev libffi-dev libapt-pkg-dev libarchive-dev mercurial python3-pip subversion squashfs-tools sudo snapd xdelta3 && python3 -m pip install -r requirements-devel.txt -r requirements.txt"
+    # snap install core fails with:
+    # - Setup snap "core" (2462) security profiles (cannot reload udev rules: exit status 2
```
kalikiana: Is the core installed despite the error? The comment says it fails... but then why would we not drop the install command? I'd appreciate if you can clarify in the comment.

elopio: I have added a few more words.

kalikiana: Thanks! This is much clearer!

tools/travis/run_tests.sh
```diff
@@ -34,7 +34,10 @@ if [ "$test" = "static" ]; then
...
 elif [ "$test" = "integration" ] || [ "$test" = "plugins" ] || [ "$test" = "store" ] || [ "$test" = "containers" ] || [ "$test" = "snapd" ]; then
-    dependencies="apt install -y bzr curl git libnacl-dev libsodium-dev libffi-dev libapt-pkg-dev libarchive-dev mercurial python3-pip subversion squashfs-tools sudo snapd xdelta3 && python3 -m pip install -r requirements-devel.txt -r requirements.txt"
+    # snap install core exits with this error message:
+    # - Setup snap "core" (2462) security profiles (cannot reload udev rules: exit status 2
+    # but the installation succeeds, so we just ingore it.
+    dependencies="apt install -y bzr curl git libnacl-dev libsodium-dev libffi-dev libapt-pkg-dev libarchive-dev mercurial python3-pip subversion squashfs-tools sudo snapd xdelta3 && python3 -m pip install -r requirements-devel.txt -r requirements.txt && (snap install core || echo 'ignored error') && sudo snap install snaps-cache/snapcraft-pr$TRAVIS_PULL_REQUEST.snap --dangerous --classic"
```
kalikiana: Might be a good idea to break this down a bit since it becomes hard to read. Though it doesn't have to be done in this branch.

elopio: It always puzzles me how to break lines in bash. Is it only to put a \ at the end of the line inside the quotes? And then continue on the first column? agh, I hate bash.

kalikiana: I hear yeah. Feel free to leave it as-is for now.

PR #1632, opened 2017-10-18
libraries: exclude the full set of libc6

sergiusens:

-----

### DIFF
--- a/snapcraft/internal/libraries.py
+++ b/snapcraft/internal/libraries.py
@@ -20,7 +20,7 @@ import logging
 import os
 import subprocess
 
-from snapcraft.internal import common
+from snapcraft.internal import common, repo
 
 
 logger = logging.getLogger(__name__)
@@ -73,9 +73,10 @@ def _get_system_libs():
         lib_path = None
 
     if not lib_path or not os.path.exists(lib_path):
-        logger.debug('No libraries to exclude from this release')
-        # Always exclude libc.so.6
-        return frozenset(['libc.so.6'])
+        logger.debug('Only excluding libc libraries from the release')
+        libc6_libs = [os.path.basename(l)
+                      for l in repo.Repo.get_package_libraries('libc6')]
+        return frozenset(libc6_libs)
 
     with open(lib_path) as fn:
         _libraries = frozenset(fn.read().split())
--- a/snapcraft/tests/test_libraries.py
+++ b/snapcraft/tests/test_libraries.py
@@ -169,6 +169,8 @@ class TestSystemLibsOnReleasesWithNoVersionId(tests.TestCase):
                                         "https://bugs.gentoo.org/"}
         self.addCleanup(patcher.stop)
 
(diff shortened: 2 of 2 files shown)


kyrofa:
I like this, but it seems like a bit of a half measure. In an ideal scenario, wouldn't we want to do this for every package included in the core snap? A list that, of course, we do not have.

snapcraft/internal/libraries.py
```diff
@@ -73,9 +73,10 @@ def _get_system_libs():
         lib_path = None
 
     if not lib_path or not os.path.exists(lib_path):
-        logger.debug('No libraries to exclude from this release')
-        # Always exclude libc.so.6
-        return frozenset(['libc.so.6'])
+        logger.debug('Only excluding libc libraries from the release')
+        libc6_libs = [os.path.basename(l)
+                      for l in repo.Repo.get_package_libraries('libc6')]
```
kyrofa: `get_package_libraries` definitely includes more than just libs. Will that cause issues, here?

sergiusens:
On mié, oct 18, 2017 at 5:51 PM, Kyle Fazzari 
> I like this, but it seems like a bit of a half measure. In an ideal 
> scenario, wouldn't we want to do this for every package included in 
> the core snap? A list that, of course, we do not have.
> 

This is one step in the correct direction, we never want half of libc6 
consumed into the snap. The correct measure is to disable this 
functionality completely and never bring libraries in from the host.
>

sergiusens:
On mié, oct 18, 2017 at 5:54 PM, Kyle Fazzari 
> @kyrofa commented on this pull request.
> 
> In snapcraft/internal/libraries.py:
> 
> > @@ -73,9 +73,10 @@ def _get_system_libs():
>          lib_path = None
> 
>      if not lib_path or not os.path.exists(lib_path):
> -        logger.debug('No libraries to exclude from this release')
> -        # Always exclude libc.so.6
> -        return frozenset(['libc.so.6'])
> +        logger.debug('Only excluding libc libraries from the 
> release')
> +        libc6_libs = [os.path.basename(l)
> +                      for l in 
> repo.Repo.get_package_libraries('libc6')]
> get_package_libraries definitely includes more than just libs. Will 
> that cause issues, here?
> 
It should really not, but what did it bring? You use this logic in dump 
and copy already for similar reasons I believe.
>

kalikiana:
For the record, hitting the timeout again here `test_rust_plugin.RustPluginTestCase.test_cross_compiling ... 
The job exceeded the maximum time limit for jobs, and has been terminated.`

kyrofa:
Ah ha, once I removed the `LD_LIBRARY_PATH` from `bin/snapcraft-classic` I was actually able to hit this error, and this PR fixes it.

kalikiana:
@kyrofa What do you mean when you say "this error" here? Would it be worth linking a bug in a comment here? Or just a comment pointing out the problem?

sergiusens:
#1635 was the other problem which caused initial suspicions on this PR.

PR #1633, opened 2017-10-19
 recording: record information from the image in container builds 

### LINKED ISSUE #1628: record lxc image used
[design](https://forum.snapcraft.io/t/record-container-image-information/2527)

elopio:

-----
Closes #1628

### DIFF
--- a/snapcraft/internal/lxd/_containerbuild.py
+++ b/snapcraft/internal/lxd/_containerbuild.py
@@ -15,6 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import collections
 import json
 import logging
 import os
@@ -138,6 +139,33 @@ class Containerbuild:
         subprocess.check_call([
             'lxc', 'config', 'set', self._container_name,
             'environment.LC_ALL', 'C.UTF-8'])
+        self._set_image_info_env_var()
+
+    def _set_image_info_env_var(self):
+        FAILURE_WARNING_FORMAT = (
+            'Failed to get container image info: {}\n'
+            'It will not be recorded in manifest.')
+        try:
+            image_info = json.loads(subprocess.check_output(
+                ['lxc', 'image', 'list',
+                 '--format=json', self._image]).decode())
+        except subprocess.CalledProcessError as e:
+            message = '{}, output: {}'.format(str(e), e.output)
+            logger.warning(FAILURE_WARNING_FORMAT.format(message))
+            return
+        except json.decoder.JSONDecodeError as e:
+            logger.warning(FAILURE_WARNING_FORMAT.format('Not in JSON format'))
+            return
+        edited_image_info = collections.OrderedDict()
+        for field in ('fingerprint', 'architecture', 'created_at'):
+            if field in image_info[0]:
+                edited_image_info[field] = image_info[0][field]
+        # Pass the image info to the container so it can be used when recording
+        # information about the build environment.
+        subprocess.check_call([
+            'lxc', 'config', 'set', self._container_name,
+            'environment.SNAPCRAFT_IMAGE_INFO',
+            json.dumps(edited_image_info)])
 
     def execute(self, step='snap', args=None):
         with self._container_running():
--- a/snapcraft/internal/meta.py
+++ b/snapcraft/internal/meta.py
@@ -158,7 +158,9 @@ class _SnapPackaging:
             annotated_snapcraft = self._annotate_snapcraft(
                 copy.deepcopy(self._config_data))
             with open(manifest_file_path, 'w') as manifest_file:
-                yaml.dump(annotated_snapcraft, manifest_file)
+                yaml.dump(
+                    annotated_snapcraft, manifest_file,
+                    default_flow_style=False)
 
     def _annotate_snapcraft(self, data):
         image_info = os.environ.get('SNAPCRAFT_IMAGE_INFO')
--- a/snapcraft/internal/meta.py
+++ b/snapcraft/internal/meta.py
@@ -18,6 +18,7 @@ import copy
 import collections
 import contextlib
 import configparser
+import json
 import logging
 import os
 import re
@@ -160,6 +161,14 @@ class _SnapPackaging:
                 yaml.dump(annotated_snapcraft, manifest_file)
 
     def _annotate_snapcraft(self, data):
+        image_info = os.environ.get('SNAPCRAFT_IMAGE_INFO')
+        if image_info:
+            try:
+                image_info_dict = json.loads(image_info)
+            except json.decoder.JSONDecodeError as exception:
+                raise errors.InvalidContainerImageInfoError(
+                    image_info) from exception
+            data['image-info'] = image_info_dict
         for field in ('build-packages', 'build-snaps'):
             data[field] = get_global_state().assets.get(field, [])
         for part in data['parts']:
--- a/snapcraft/tests/fixture_setup.py
+++ b/snapcraft/tests/fixture_setup.py
@@ -535,40 +535,46 @@ class FakeLXD(fixtures.Fixture):
         self.architecture_mock.return_value = ('64bit', 'ELF')
         self.addCleanup(patcher.stop)
 
+    def call_effect(self, *args, **kwargs):
+        if args[0] == ['lxc', 'remote', 'get-default']:
+            return 'local'.encode('utf-8')
+        elif args[0][:2] == ['lxc', 'info']:
+            return '''
+                environment:
+                  kernel_architecture: {}
+                '''.format(self.kernel_arch).encode('utf-8')
+        elif args[0][:3] == ['lxc', 'list', '--format=json']:
+            if self.status and args[0][3] == self.name:
+                return string.Template('''
+                    [{"name": "$NAME",
+                      "status": "$STATUS",
+                      "devices": $DEVICES}]
+                    ''').substitute({
+                        # Container name without remote prefix
+                        'NAME': self.name.split(':')[-1],
+                        'STATUS': self.status,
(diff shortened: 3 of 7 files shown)


snapcraft/internal/meta.py
```diff
@@ -160,6 +161,12 @@ def _record_manifest_and_source_snapcraft_yaml(self):
                 yaml.dump(annotated_snapcraft, manifest_file)
 
     def _annotate_snapcraft(self, data):
+        image_info = os.environ.get('SNAPCRAFT_IMAGE_INFO')
+        if image_info:
+            image_info_dict = json.loads(image_info)
```
kalikiana: I'm thinking maybe this shouldn't be picking what fields to include, since the invoking snapcraft may be newer than the one it's calling in the container. So if in the future we want to add or drop fields, it'd be good if the invoking snapcraft was in control.
Also, some error handling would probably be good? Log a warning like `Couldn't parse JSON image info`. Again, thinking to rather be conservative because of the invoking/ container snapcraft relationship.

snapcraft/tests/fixture_setup.py
```diff
@@ -534,40 +534,46 @@ def _setUp(self):
...
-            elif args[0][:2] == ['lxc', 'exec']:
-                return self._lxc_exec(args)
-            elif args[0][0] == 'sha384sum':
-                return 'deadbeef {}'.format(args[0][1]).encode('utf-8')
-            elif '/usr/lib/sftp-server' in args[0]:
-                return self._popen(args[0])
-            else:
-                return ''.encode('utf-8')
-        return call_effect
+        return self.call_effect
```
kalikiana: Why are you moving this around? Can you clarify?

snapcraft/internal/lxd.py
```diff
@@ -130,6 +131,18 @@ def _ensure_container(self):
...
+
+    def _set_image_info_env_var(self):
+        image_info = json.loads(check_output(
+            ['lxc', 'image', 'list', '--format=json', self._image]).decode())
+        image_info_edited = collections.OrderedDict()
+        for field in ('fingerprint', 'architecture', 'created_at'):
+            image_info_edited[field] = image_info[0][field]
+        check_call([
+            'lxc', 'config', 'set', self._container_name,
+            'environment.SNAPCRAFT_IMAGE_INFO',
```
kalikiana: Maybe add a comment here like `# Set image info to be used in _annotate_snapcraft`.

elopio: Good call. I added a comment, please check if it's clear.

snapcraft/internal/meta.py
elopio: Ok, I'm raising an error if the parsing doesn't work. And I'm checking if the keys are present on the dict before recording them in the manifest.
I also removed the filtering on the host.

snapcraft/tests/fixture_setup.py
elopio: It was failing the complexity check. I tried hard to split the function but failed miserably.
This needs some rethinking, because it's a little hard to read and maintain, and now the next condition we add will fail again.

snapcraft/internal/meta.py
kalikiana: Actually I meant to suggest the reverse. I'll try to explain more clearly. The `lxd.py` code path is the invoking snapcraft, which can be newer than the snapcraft being called inside the container. So if you choose the keys in `lxd.py` we're more flexible in the future.

snapcraft/tests/fixture_setup.py
kalikiana: Hrm, leave it for now I guess, any split in there is arbitrary.

snapcraft/internal/lxd.py
kalikiana: Very nice. Thank you for adding that!

snapcraft/internal/meta.py
elopio: I get it now, and I think you are right. Leave the logic to the host, and on the guest just throw whatever is in the variable into the file.

kalikiana: Yeah. Thanks for changing it!

kalikiana:
@elopio It turns out `lxc image list --format=json` is not understood by all LXD versions. On the other hand `lxc image info`, parsed as YAML, fulfills the same purpose - I went ahead and changed it. Two points worth noting:
1. The timestamp format is different
2. The format of `SNAPCRAFT_IMAGE_INFO` hasn't changed

snapcraft/tests/test_lifecycle.py
```diff
@@ -1211,6 +1211,67 @@ def test_prime_with_plugin_manifest(self, fake_plugin_manifest):
...
+                build-packages: []
+                installed-packages: []
+                installed-snaps: []
+                plugin: nil
+                prime: []
+                stage: []
+                stage-packages: []
+                uname: Linux test uname 4.10 x86_64
+            architectures: [{}]
+            image-info: {}
```
sergiusens: For `image-info` it would be nice to see this actually expanded here (manually), instead of using `format` to get a real feel of how it would look like.

This should also look like yaml

kalikiana: It's YAML - curly braces are the one-line syntax for dictionaries, analoguous to square braces for lists.
But I agree after putting it into the test as per your suggestion that it feels inconsistent. I'm trying to find out how to force `yaml.dump` to use multi-line syntax here.

sergiusens:
It feels like `lxc image info` is not the best solution here as it is not guaranteed to be yaml; it would be better to just do `--format=json` and if the version of lxd used does not support it, raise a warning about it.

kyrofa:
> It feels like `lxc image info` is not the best solution here as it is not guaranteed to be yaml; it would be better to just do `--format=json` and if the version of lxd used does not support it, raise a warning about it.

I completely agree. Parsing something that we know isn't YAML as YAML is asking for trouble.

elopio:
@kalikiana I'm going to revert your two commits. One to just don't record the image info if lxc image fails, and the other because I tried  `default_flow_style=False` and it makes the entire yaml uglier, just to make this entry prettier. I couldn't find an easy solution to format the yaml like I would do, but I prefer how it looks with the default style.

kalikiana:
@elopio Uglier in what regard? The only other change I saw was architectures using `-` syntax.

Seems like you also reverted putting the image info directly in the expected YAML? Whichever way we go, I found it useful to see what it looked like.

snapcraft/tests/test_lifecycle.py
sergiusens: I would like to see the yaml here please, the revert moved that all away.

sergiusens:
@elopio please use `git revert <hash>` and push new commits without squashing as it is hard to keep track of a review like this.

sergiusens:
`default_flow=False` is what we use for generating `snap.yaml`, we should use that here too.

snapcraft/tests/test_lifecycle.py
elopio: Sorry, putting it back...

elopio:
I don't like the alignment of default_flow_style=False and that it doesn't use [] for small lists. But at least I could get rid of a few NOQA, not all is bad :)

snapcraft/internal/lxd/_containerbuild.py
```diff
@@ -138,6 +139,33 @@ def _configure_container(self):
...
+        try:
+            image_info = json.loads(subprocess.check_output(
+                ['lxc', 'image', 'list',
+                 '--format=json', self._image]).decode())
+        except subprocess.CalledProcessError as e:
+            message = '{}, output: {}'.format(str(e), e.output)
+            logger.warning(FAILURE_WARNING_FORMAT.format(message))
+            return
+        except json.decoder.JSONDecodeError as e:
+            logger.warning(FAILURE_WARNING_FORMAT.format('Not in JSON format'))
```
kalikiana: Would be nice to include the "what" here ie. the JSON that failed.

snapcraft/internal/lxd/_containerbuild.py
```diff
@@ -138,6 +139,33 @@ def _configure_container(self):
...
+        self._set_image_info_env_var()
+
+    def _set_image_info_env_var(self):
+        FAILURE_WARNING_FORMAT = (
+            'Failed to get container image info: {}\n'
+            'It will not be recorded in manifest.')
+        try:
+            image_info = json.loads(subprocess.check_output(
+                ['lxc', 'image', 'list',
+                 '--format=json', self._image]).decode())
```
kalikiana: This looks a bit ugly:

    $ SNAPCRAFT_BUILD_INFO=1 snapcraft cleanbuild  
    Creating snapcraft-scintillantly-gobony-leena  
    Starting snapcraft-scintillantly-gobony-leena
    error: flag provided but not defined: --format
    Failed to get container image info: Command '['lxc', 'image', 'list', '--format=json', 'ubuntu:xenial/amd64']' returned non-zero exit status 2., output: b''
    It will not be recorded in manifest.
    Waiting for a network connection...
    Network connection established

snapcraft/internal/lxd/_containerbuild.py
```diff
@@ -138,6 +139,33 @@ def _configure_container(self):
         subprocess.check_call([
             'lxc', 'config', 'set', self._container_name,
             'environment.LC_ALL', 'C.UTF-8'])
+        self._set_image_info_env_var()
+
+    def _set_image_info_env_var(self):
+        FAILURE_WARNING_FORMAT = (
+            'Failed to get container image info: {}\n'
+            'It will not be recorded in manifest.')
```
sergiusens: in *the* manifest
or _it will not be recorded as part of the manifest_

elopio: one of the recommendations from mtp was to not use unnecessary articles.

PR #1634, opened 2017-10-19
Push metadata to the Store.

facundobatista:
This is to fulfill first part of design discussed in [this forum post](https://forum.snapcraft.io/t/mechanisms-for-converging-store-and-snap-metadata/2200/7).


snapcraft/storeapi/__init__.py
```diff
@@ -364,6 +364,18 @@ def get_assertion(self, snap_id, endpoint):
     def sign_developer_agreement(self, latest_tos_accepted=False):
         return self.sca.sign_developer_agreement(latest_tos_accepted)
 
+    def update_metadata(self, snap_name, metadata, arch, force):
```
kalikiana: What's the purpose of `arch` here? It only seems to be used for the error.

snapcraft/storeapi/errors.py
```diff
@@ -334,6 +334,42 @@ def __init__(self, snap_name, response):
...
+            self.fmt = self.__FMT_NOT_REGISTERED
+        elif response.status_code == 409:
+            conflicts = [(error['extra']['name'], error)
+                         for error in response_json['error_list']
+                         if error['code'] == 'conflict']
+            parts = ["Metadata not updated!"]
+            for field_name, error in sorted(conflicts):
+                sent = metadata.get(field_name)
+                parts.extend((
+                    "Conflict in {!r} field:".format(field_name),
```
kalikiana: The error should probably say something like `You can use --force-metadata to update anyway`?

snapcraft/storeapi/__init__.py
facundobatista: Exactly! The idea is to maintain the detail level of the rest of SnapNotFoundError messages.

snapcraft/storeapi/errors.py
facundobatista: And repeat this message in ALL conflict lines, you say? 

Maybe it's better to add a line (in case of conflicts) after all conflicts, saying something like "You can repeat the command with --force-metadata to force the local values into the Store"?

What do you think? Thanks!

facundobatista:
Oh, Travis failed because a CLA check: `The command "./tools/travis/run_cla_check.sh" exited with 1.`, but I already signed the CLA.

Looks like failed the CLA check on @jocave  (joc@jocave.me.uk), but I don't know why he is related to my branch :(

snapcraft/storeapi/errors.py
kalikiana: Yeah, that sounds great!

snapcraft/storeapi/__init__.py
kalikiana: I see. It just strikes me as odd that the method would use a value in the error that isn't needed to do the actual work. As in, the error is providing more detail than is relevant to the failure. I might be missing something, though.

facundobatista:
@kalikiana there I did the changes we talked about, thanks!

snapcraft/_store.py
```diff
@@ -389,8 +409,18 @@ def push(snap_filename, release_channels=None):
 
     If release_channels is defined it also releases it to those channels if the
     store deems the uploaded snap as ready to release.
+
+    If only_metadata is True it will just send the metadata; otherwise the
+    whole process is done (including sending the metadata after sending the
+    snap binary).
```
elopio: Can you please document the force_metadata arg?

snapcraft/cli/store.py
```diff
@@ -101,12 +106,21 @@ def push(snap_file, release):
...
     operation takes place.
 
+    If --only-metadata is given, only the snap's metadata information will be
+    pushed to the server (not the whole .snap file). Because of this, --release
+    is not a compatible option to use with it.
+
     \b
     Examples:
         snapcraft push my-snap_0.1_amd64.snap
+        snapcraft push my-snap_0.1_amd64.snap --only-metadata
```
sergiusens: hey @matthewpaulthomas or @evandandrea  do you think `--only-metadata` is correct here, it feels strange to me even though it is technically correct. Should this be `--metadata` or something else?

snapcraft/cli/store.py
```diff
@@ -101,12 +106,21 @@ def push(snap_file, release):
...
         snapcraft push my-snap_0.2_amd64.snap --release edge
         snapcraft push my-snap_0.3_amd64.snap --release candidate,beta
     """
+    if release and only_metadata:
+        raise click.UsageError(
+            "--only-metadata and --release are mutually exclusive")
```
sergiusens: Can't this be done with click itself?

snapcraft/cli/store.py
```diff
@@ -115,7 +129,7 @@ def push(snap_file, release):
             'After pushing, an attempt to release to {} '
             'will be made'.format(channel_list))
 
-    snapcraft.push(snap_file, channel_list)
+    snapcraft.push(snap_file, channel_list, only_metadata, force_metadata)
```
sergiusens: instead of passing the logic over to push, can we handle the logic to call `update_metadata` from here? The printing with logger in that method should also be brought over here (it is a bit hard with some of the current logic we have but  most of this stuff should be easily movable.

snapcraft/cli/store.py
evandandrea: I don't think `--only-metadata` is self-explanatory. The option should reflect the developer's intention, not what happens internally (the metadata is sent, the snap is not).

I would be more explicit and use `--overwrite-metadata`.

facundobatista: Note however that "overwrite metadata" is not what it does, unless you also pass "force metadata", because pushing metadata can result in conflicts if you modified those fields through the web UI.

Also, these names are stablished in the proper [forum post](https://forum.snapcraft.io/t/mechanisms-for-converging-store-and-snap-metadata/2200), should discuss these there?

snapcraft/cli/store.py
facundobatista: I searched for this, it look it [can't be done and no plans for click to gain this capability](https://github.com/pallets/click/issues/257).

snapcraft/cli/store.py
facundobatista: It can be moved, but conceptually "it's a push" (complete, or only metadata). 

Another drawback of moving it to here is the need to call the private '_get_data_from_snap_file' from the private '_store".

I could start to move stuff around and/or make it public if you insist (but probably it would be better to do that in a separate branch, which I can commit to do if we decide this path)

snapcraft/_store.py
facundobatista: Done, thanks!

facundobatista:
Well, Travis now failed again with a weird message: "The job exceeded the maximum time limit for jobs, and has been terminated." 

How can I trigger a new Travis check?

kalikiana:
I re-triggered it for you.

facundobatista:
Thanks! It worked ok now...

snapcraft/storeapi/errors.py
```diff
@@ -379,6 +379,45 @@ def __fmt_error_unknown(self, response):
         return fmt
 
 
+class StoreMetadataError(StoreError):
+
+    __FMT_NOT_REGISTERED = (
+        'Sorry, try `snapcraft register {snap_name}` before trying to '
+        'update metadata.')
```
elopio: I'm getting this error when I try to push --only-metadata without having pushed the snap first.
We should catch when the snapcraft is registered but not pushed, and recommend to call without `--only-metadata`. Is that possible?

elopio: Following @matthewpaulthomas's suggestion, the error messages should say what happened, why, and how to fix it.
What do you think of something like: "Failed to update metadata: snap not registered. Run first `snapcraft register {snap_name} and then `snapcraft push {snap_file}". ?

snapcraft/storeapi/errors.py
```diff
@@ -379,6 +379,45 @@ def __fmt_error_unknown(self, response):
...
+            parts = ["Metadata not updated!"]
+            for field_name, error in sorted(conflicts):
+                sent = metadata.get(field_name)
+                parts.extend((
+                    "Conflict in {!r} field:".format(field_name),
+                    "    In snapcraft.yaml: {!r}".format(sent),
+                    "    In the Store:      {!r}".format(error['message']),
+                ))
+            parts.extend((
+                "You can repeat the command with --force-metadata",
```
elopio: This comma introduces a weird new line in the middle of the message. We generally don't wrap the messages, and if we do, we probably should do it in a single place. So, I think it would be better to add a single string.

sergiusens:
On jue, nov 9, 2017 at 1:40 AM, Leo Arias <notifications@github.com> 
wrote:
> Following @matthewpaulthomas's suggestion, the error messages should 
> say what happened, why, and how to fix it.
> What do you think of something like: "Failed to update metadata: snap 
> not registered. Run first snapcraft register {snap_name} and 
> thensnapcraft push {snap_file}". ?

Updating the metadata has failed: the snap is not registered. First run 
`snapcraft register <snap-name>` and the `snapcraft push <snap-file>`.

We also need a new word for metadata as discussed on the forum topic 
for information extracion. Maybe "Updating the information on the store 
has failed"

snapcraft/storeapi/errors.py
matiasb: Also related to this message, "You can repeat the command with --force-metadata" seems confusing when you get this error while pushing a new revision. Note that in that case the revision is correctly uploaded, so re-uploading it won't work (so you would need to do a --only-metadata + --force-metadata instead).

snapcraft/storeapi/errors.py
facundobatista: Will change, taking also Sergio's recommendation below

snapcraft/storeapi/errors.py
facundobatista: Regarding the comma, I was doing that *on purpose*, I'll fix it to have a single string.

And will improve the message.

facundobatista:
@elopio thanks for the comments, there I answered them. Also I'll add an integration test, and describe in the manual file how to test the situation of the conflict (as you said, this can not be done automatically)

@sergiusens I'll like your suggestion for the message. Also will check that we don't say "metadata" in any human text

facundobatista:
@elopio @sergiusens did the changes as discussed/reviewed, thanks!

snapcraft/storeapi/__init__.py
```diff
@@ -364,6 +364,18 @@ def get_assertion(self, snap_id, endpoint):
     def sign_developer_agreement(self, latest_tos_accepted=False):
         return self.sca.sign_developer_agreement(latest_tos_accepted)
 
+    def update_metadata(self, snap_name, metadata, force):
```
sergiusens: From conversations on IRC, can you call this `push_metadata`?

snapcraft/storeapi/__init__.py
```diff
@@ -609,6 +621,21 @@ def snap_push_metadata(self, snap_name, updown_data,
 
         return StatusTracker(response.json()['status_details_url'])
 
+    def update_metadata(self, snap_id, snap_name, metadata, force):
```
sergiusens: Please rename to `push_metadata`

snapcraft/_store.py
```diff
@@ -395,14 +402,14 @@ def update_metadata(snap_yaml, force):
     # hit the server
     store = storeapi.StoreClient()
     with _requires_login():
-        store.update_metadata(snap_name, metadata, force)
+        store.push_precheck(snap_name)
+        store.push_metadata(snap_name, metadata, force)
 
-    logger.info("The metadata has been updated")
+    logger.info("The metadata has been pushed")
```
sergiusens: Just one minor nit, can you remove this from here and put it under
```
snapcraft.push_metadata(snap_file, force)
```
in `snapcraft.cli.store`?

snapcraft/_store.py
```diff
@@ -381,9 +381,16 @@ def sign_build(snap_filename, key_name=None, local=False):
...
 
-def update_metadata(snap_yaml, force):
-    """Update metadata in the server."""
+def push_metadata(snap_filename, force):
+    """Push only the metadata to the server.
+
+    If force=True it will force the local metadata into the Store,
+    ignoring any possible conflict.
+    """
+    logger.info("Pushing metadata to the Store (force=%s)", force)
```
sergiusens: Mind removing this line or making it `.debug` or we will get a double print on regular runs?

facundobatista:
Pushed requested change

PR #1636, opened 2017-10-20
internal: more gracefully determine host OS

kyrofa:

-----

Currently snapcraft just reads the `/etc/os-release` file into a dict and accesses keys everywhere without caring whether or not they exist. This leads to difficult-to-diagnose errors on distros where e.g. the `VERSION_CODENAME` isn't set. That particular example includes Trusty, where we can use the VERSION_ID as a fallback to determine the codename.

Create a class around parsing `/etc/os-release` so we can gracefully fallback as appropriate. This allows the snapcraft snap to be used on Trusty, and also makes the errors when running on an unsupported distro more friendly.


snapcraft/internal/repo/_platform.py
```diff
@@ -30,12 +30,12 @@
...
 def _is_deb_based(distro=None):
     if not distro:
-        distro = get_os_release_info()['ID']
+        distro = OsRelease().id()
     return distro in _DEB_BASED_PLATFORM
 
 
 def _get_repo_for_platform():
-    distro = get_os_release_info()['ID']
+    distro = OsRelease().id()
```
kalikiana: How about a `.like()` method that looks at `ID` as well as `ID_LIKE` so we can support derivatives more easily? I essentially suggested the same in #1637 but with the new class we could simplify the calling site by baking it into the API. Thoughts?

integration_tests/snapd/test_catkin_snap.py
```diff
@@ -21,12 +21,12 @@
...
 import integration_tests
 from snapcraft.tests import fixture_setup
-from snapcraft.internal.common import get_os_release_info
+from snapcraft.internal.os_release import OsRelease
 
 
 class CatkinTestCase(integration_tests.SnapdIntegrationTestCase):
 
-    @skipUnless(get_os_release_info().get('VERSION_CODENAME') == 'xenial',
+    @skipUnless(OsRelease().version_codename() == 'xenial',
```
kalikiana: This will fail now with an error if there is no `VERSION_CODENAME` or `VERSION_ID`, won't it? The test should however be skipped if those optional fields are not present.
Maybe the method should just return `None` in that case and not raise an exception?

kyrofa:
> Looks really good, I am inclined to ask to move this one level up (out of internal) and even make it a package with its own set of errors that snapcraft would wrap around in its own errors but this might be a larger PR giving the double wrapping of errors.

I think that's a great idea, although I'd like to get a little mileage out of the API before we make it public if that's okay? So I'd like to make that a follow-up PR.

integration_tests/snapd/test_catkin_snap.py
kyrofa: I agree that the test should be skipped instead of toss an exception, but I don't think that's the responsibility of `OsRelease`, but rather something on the test base class. I'll add a helper method there.

snapcraft/internal/os_release.py
```diff
@@ -0,0 +1,89 @@
...
+        tries to use the VERSION_ID to figure out the codename on its own.
+
+        :raises OsReleaseCodenameError: If no version codename can be
+                                        determined.
+        """
+        with contextlib.suppress(KeyError):
+            return self._os_release['VERSION_CODENAME']
+
+        with contextlib.suppress(KeyError):
+            return _ID_TO_UBUNTU_CODENAME[self._os_release['VERSION_ID']]
```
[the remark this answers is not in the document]
kyrofa: We chatted about this but I'll mentioned it here for completeness: we can't use the ID everywhere, as we build [sources lists](https://github.com/snapcore/snapcraft/pull/1636/files#diff-e405f18ae8c707ccb3d5ea4e828718c9R152), which requires the codename.

snapcraft/internal/repo/_platform.py
kyrofa: I think that's an interesting idea for #1637, but doing it here would inflate the API with no users.

integration_tests/snapd/test_catkin_snap.py
kyrofa: Done.

snapcraft/tests/test_libraries.py
```diff
@@ -155,18 +168,27 @@ def setUp(self):
...
+            f.write(dedent("""\
+                NAME="Ubuntu"
+                VERSION="16.04.3 LTS (Xenial Xerus)"
+                ID=ubuntu
+                ID_LIKE=debian
+                PRETTY_NAME="Ubuntu 16.04.3 LTS"
+                HOME_URL="http://www.ubuntu.com/"
+                SUPPORT_URL="http://help.ubuntu.com/"
+                BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
+                UBUNTU_CODENAME=xenial
```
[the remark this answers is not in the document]
kyrofa: Haha, that seems arbitrarily pedantic, but I went ahead and make the change since it was easy.

snapcraft/tests/test_libraries.py
```diff
@@ -118,19 +122,28 @@ class TestSystemLibsOnNewRelease(tests.TestCase):
...
-                                    'BUG_REPORT_URL':
-                                        'http://bugs.launchpad.net/ubuntu/',
-                                    'VERSION_ID': '16.04',
-                                    'UBUNTU_CODENAME': 'xenial',
-                                    'ID': 'ubuntu', 'NAME': 'Ubuntu',
-                                    'ID_LIKE': 'debian',
-                                    'PRETTY_NAME': 'Ubuntu 16.04.3 LTS',
-                                    'VERSION': '16.04.3 LTS (Xenial Xerus)',
-                                    'SUPPORT_URL': 'http://help.ubuntu.com/'}
+        with open('os-release', 'w') as f:
```
sergiusens: yay for no mocking

snapcraft/internal/os_release.py
```diff
@@ -0,0 +1,89 @@
...
+
+class OsRelease:
+    """A class to intelligently determine the OS on which we're running"""
+
+    def __init__(self, *, os_release_file='/etc/os-release'):
+        """Create a new OsRelease instance.
+
+        :param str os_release_file: Path to os-release file to be parsed.
+        """
+        with open(os_release_file) as os_release_file:
```
elopio: Just a nit about reusing variables here. I would prefer to name one os_release_file_path or os_release_path, and the other os_release_file.

snapcraft/tests/test_libraries.py
```diff
@@ -118,19 +122,28 @@ class TestSystemLibsOnNewRelease(tests.TestCase):
...
+                UBUNTU_CODENAME=xenial
+            """))
+        release = os_release.OsRelease(os_release_file='os-release')
+
+        def _create_os_release(*args, **kwargs):
+            return release
+
+        patcher = mock.patch(
+            'snapcraft.internal.os_release.OsRelease',
+            wraps=_create_os_release)
```
elopio: fancy!

snapcraft/internal/os_release.py
kyrofa: Oh, yeah I agree, not sure what I was thinking there!

PR #1638, opened 2017-10-23
 tests: reorganize unit and integration suites to make them easier to split for travis

elopio:

-----
requires #1630


.travis.yml
```diff
@@ -13,20 +13,22 @@ jobs:
       script: sudo ./tools/travis/run_tests.sh static
     - stage: unit
       if: type != cron
-      script: sudo ./tools/travis/run_tests.sh unit
+      script: sudo ./tools/travis/run_tests.sh snapcraft/tests/unit
```
kyrofa: Is everything nested within `snapcraft/tests/`? If so, would it be worth making it so that `snapcraft/tests/` doesn't need to be specified each time?

elopio: @kyrofa I could preppend snapcraft/tests if the argument doesn't start with that.
But it worries me that we might fall in the same trap as before, where the script is some dark magic. It seemed to me that using the full path would make it clearer if you want to filter and run only one subdirectory, which before had to do with a weird thing like `*python*`.

No strong opinion though, I would say it's a trivial change, but bash is never trivial. Let's say "easy".

kyrofa: Haha, no that's fine.

elopio:
@sergiusens how would that work for the tests in setup.py though? Shouldn't they be inside the package?

sergiusens:
on `xenial-armhf` this is the only failure:
```
ERROR: test_shared_ros_builds_without_catkin_in_underlay (test_catkin.CatkinTestCase)
test_catkin.CatkinTestCase.test_shared_ros_builds_without_catkin_in_underlay
```

due to
```
ERROR: Not all sources were able to be updated.
[[[
ERROR: unable to process source [https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/python.yaml]:
	<urlopen error [Errno -3] Temporary failure in name resolution> (https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/python.yaml)
ERROR: unable to process source [https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/ruby.yaml]:
	<urlopen error [Errno -3] Temporary failure in name resolution> (https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/ruby.yaml)
]]]
}}}
```

PR #1639, opened 2017-10-25
grammar: to statement

kalikiana:

-----

This is the implementation of the "to" statement [as discussed in the forum](https://forum.snapcraft.io/t/specifying-architecture-on-cross-compilations/1641/20). This PR implements the **bold items**.
Items 7 and 8 are implemented in the follow-up PR #1800 to allow for quicker review (although I can also add them here if that's preferred, whichever is clearer).

1. The existing “on <arch>:” syntax will be accepted under build-packages as well
2. The default architecture for packages under “on <arch>” is the one from the build machine
3. Packages under “on <arch>:” are installed only when the build machine has that architecture
4. **A new “to <arch>:” statement will be supported in build-packages and stage-packages**
5. **The default architecture for packages under “to <arch>:” is the target architecture**
6. **Packages under “to <arch>:” are installed only when the target architecture is the one listed, which includes both a cross-compilation towards that architecture and a native build on that architecture**
7. The two statements may be used together as “on <arch> to <arch>:” to precisely specify the case of a particular cross-compilation
8. The default architecture for packages under “on <arch> to <arch>:” is the target architecture


snapcraft/internal/project_loader/grammar/_to.py
```diff
@@ -0,0 +1,149 @@
...
+        return self.selectors == other.selectors
+
+    def __repr__(self):
+        return "'to {}'".format(','.join(sorted(self.selectors)))
+
+
+def _extract_to_clause_selectors(to):
+    """Extract the list of selectors within a to clause.
+
+    :param str to: The 'to <selector' part of the 'to' clause.
```
kyrofa: Missing a bracket here: `'to <selector>'`

snapcraft/tests/project_loader/grammar/test_to_statement.py
```diff
@@ -0,0 +1,264 @@
...
+            'body': ['foo'],
+            'else_bodies': [
+                [
+                    {'to armhf': ['bar']},
+                    {'else': ['baz']},
+                ],
+            ],
+            'target_arch': 'i386',
+            'expected_packages': {'baz'}
+        }),
```
kyrofa: You're missing a test for ensuring one can still have packages like `name:arch` without the grammar rewriting the requested arch. Please add a test with this in both the body and an else.

kalikiana: I just realized I didn't actually respond here: I added two new cases, with "arch specified" in the tag, for that. Good call!

kalikiana:
For the record, CI failed with `FAIL: test_failed_registration_already_registered (snapcraft.tests.integration.store.test_store_register.RegisterTestCase)
snapcraft.tests.integration.store.test_store_register.RegisterTestCase.test_failed_registration_already_registered
testtools.testresult.real._StringException: Traceback (most recent call last):
  File "/root/snapcraft/snapcraft/tests/integration/store/test_store_register.py", line 41, in test_failed_registration_already_registered
    self.register, 'test-snap-name-already-registered')
  File "/usr/local/lib/python3.5/dist-packages/testtools/testcase.py", line 485, in assertRaises
    self.assertThat(our_callable, matcher)
  File "/usr/local/lib/python3.5/dist-packages/testtools/testcase.py", line 498, in assertThat
    raise mismatch_error
testtools.matchers._impl.MismatchError: <bound method StoreTestCase.register of <snapcraft.tests.integration.store.test_store_register.RegisterTestCase.test_failed_registration_already_registered id=0x7f44d08842e8>> returned None`

snapcraft/tests/integration/general/test_build_package_grammar.py
```diff
@@ -105,3 +108,59 @@ def test_global_build_package_on_other_arch_else(self):
         self.run_snapcraft(['pull'], 'build-package-grammar-global')
 
         self.assertTrue(self._hello_is_installed())
+
+    @contextlib.contextmanager
+    def modified_yaml(self,
```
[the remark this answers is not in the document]
elopio: I don't like either.
I would like a function that makes a yaml, and it takes a string with the contents. This string is optional, if not present it will create the most simple and small valid snapcraft.yaml. This function also takes keyword arguments, for each of the keys of the yaml, so we can modify the basic default yaml.
We have some tests that do something like this, but nothing is consistent. Until we have that, I would prefer to either write the full yaml, or save it to snapcraft/tests/snaps. I don't like this modifier very much.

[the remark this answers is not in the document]
kalikiana: Thanks for your feedback! It's true we have little consistency, which is why I proposed this.

In an effort to alleviate discussion I went ahead and added a function to construct the YAML from scratch, accepting keyword arguments for toplevel keys with defaults, based on @elopio's suggestion above. Although I wasn't entirely sure about the exact details regarding strings.
We should have a hangout to flesh this out.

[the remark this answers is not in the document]
kalikiana: For the record: @elopio and discussed this and decided to consolidate on extending the fixture to take arguments for parts and build packages like the function proposed here. We don't want a contextmanager, and rather dictionaries than strings. And sharing the fixture across the test suite including unit tests will be easier than making a function available everywhere.

kalikiana:
For the record:

        FAIL: test_grammar 
    (snapcraft.tests.integration.general.test_build_snap_grammar.BuildSnapGrammarTestCase) snapcraft.tests.integration.general.test_build_snap_grammar.BuildSnapGrammarTestCase.test_grammar (build-snap-grammar-on)
    testtools.testresult.real._StringException: Traceback (most recent call last):
      File "/root/snapcraft/snapcraft/tests/integration/general/test_build_snap_grammar.py", line 66, in 
    setUp
        self.useFixture(fixture_setup.WithoutSnapInstalled('hello'))
      File "/usr/local/lib/python3.5/dist-packages/testtools/testcase.py", line 756, in useFixture
        reraise(*exc_info)
      File "/usr/local/lib/python3.5/dist-packages/testtools/_compat3x.py", line 16, in reraise
        raise exc_obj.with_traceback(exc_tb)
      File "/usr/local/lib/python3.5/dist-packages/testtools/testcase.py", line 731, in useFixture
        fixture.setUp()
      File "/root/snapcraft/snapcraft/tests/fixture_setup.py", line 934, in setUp
        "'sudo snap remove {snap}'.".format(snap=self.snap_name))
    AssertionError: This test cannot run if you already have the 'hello' snap installed. Please uninstall it by running 'sudo snap remove hello'.

Not sure how this is happening. Locally it's fine and I didn't modify the test setup...

sergiusens:
For your error the only thing I can think of is that there is some sort of dictionary collision/squatting somewhere along the way

snapcraft/tests/unit/project_loader/grammar/test_to_statement.py
```diff
@@ -0,0 +1,280 @@
...
+            'expected_packages': {'baz'}
+        }),
+    ]
+
+    @patch('platform.architecture')
+    @patch('platform.machine')
+    def test_to_statement_grammar(self, platform_machine_mock,
+                                  platform_architecture_mock):
+        platform_machine_mock.return_value = 'x86_64'
+        platform_architecture_mock.return_value = ('64bit', 'ELF')
```
kyrofa: Why are we setting this, here? `to` doesn't care about the build host, right? Just the target arch, which we're setting here.

snapcraft/tests/integration/__init__.py
```diff
@@ -196,6 +197,25 @@ def copy_project_to_cwd(self, project_dir: str) -> None:
             os.path.join(self.snaps_dir, project_dir), self.path,
             preserve_symlinks=True)
 
+    def construct_yaml(self, name='test', version='0.1',
```
kyrofa: This is better than it was before, but I feel like this is introducing yet another inconsistency in how we're doing this. I'd still rather see snaps in integration/snaps/ that we can test standalone as well, like the other grammar tests. However, I've already made that argument, so I won't block on it.

PR #1640, opened 2017-10-26
cli: add what, why, and how to fix to the common errors

elopio:

-----
First iteration over the suggestions by mpt.

### DIFF
--- a/CODE_STYLE.md
+++ b/CODE_STYLE.md
@@ -42,9 +42,9 @@ fix it.
     self.assertThat(actual, Equals(expected))
     ```
 
-* When writing unit tests that raise error conditions, the tests should only
-  check the class of the exception raised and it's attributes, not the format
-  of the error message. The formatting of the exception as a string should be
+* When writing unit tests that raise errors, the tests should only check the
+  class of the exception raised and it's attributes, not the format of the
+  error message. The formatting of the exception as a string should be
   tested only once, in the module snapcraft/tests/unit/test_errors.py
 
 [1]: TESTING.md
--- a/CODE_STYLE.md
+++ b/CODE_STYLE.md
@@ -28,6 +28,11 @@ For multiline strings, we prefer to use `textwrap.dedent`:
 
 (from https://docs.python.org/3/library/textwrap.html#textwrap.dedent)
 
+## Errors
+
+Error messages must say what happened, why it happened and what you can do to
+fix it.
+
 ## Tests
 
 * When asserting for equality, we prefer to use the `Equals` matcher from
--- a/CODE_STYLE.md
+++ b/CODE_STYLE.md
@@ -42,5 +42,10 @@ fix it.
     self.assertThat(actual, Equals(expected))
     ```
 
+* When writing unit tests that raise error conditions, the tests should only
+  check the class of the exception raised and it's attributes, not the format
+  of the error message. The formatting of the exception as a string should be
+  tested only once, in the module snapcraft/tests/unit/test_errors.py
+
 [1]: TESTING.md
 [2]: https://www.python.org/dev/peps/pep-0008
--- a/snapcraft/internal/errors.py
+++ b/snapcraft/internal/errors.py
@@ -101,6 +101,10 @@ class StepOutdatedError(SnapcraftError):
 
 
 class SnapcraftEnvironmentError(SnapcraftError):
+    # FIXME This exception is too generic.
+    # https://bugs.launchpad.net/snapcraft/+bug/1734231
+    # --elopio - 20171123
+
     fmt = '{message}'
 
     def __init__(self, message):
@@ -128,7 +132,9 @@ class ContainerRunError(SnapcraftError):
     )
 
     def __init__(self, *, command, exit_code):
-        super().__init__(command=' '.join(command), exit_code=exit_code)
+        if isinstance(command, list):
+            command = ' '.join(command)
+        super().__init__(command=command, exit_code=exit_code)
 
 
 class ContainerSnapcraftCmdError(ContainerRunError):
@@ -139,7 +145,9 @@ class ContainerSnapcraftCmdError(ContainerRunError):
     )
 
     def __init__(self, *, command, exit_code):
-        super().__init__(command=' '.join(command), exit_code=exit_code)
+        if isinstance(command, list):
+            command = ' '.join(command)
+        super().__init__(command=command, exit_code=exit_code)
 
 
 class SnapdError(SnapcraftError):
@@ -191,7 +199,7 @@ class InvalidDesktopFileError(SnapcraftError):
 
     fmt = (
         'Failed to generate desktop file: '
-        'Invalid desktop file {filename!r}: {message}. '
+        'Invalid desktop file {filename!r}: {message}.'
         # FIXME include how to fix each of the possible desktop file errors.
         # https://bugs.launchpad.net/snapcraft/+bug/1727435
         # --elopio - 2017-10-25
--- a/snapcraft/internal/errors.py
+++ b/snapcraft/internal/errors.py
@@ -40,7 +40,7 @@ class SnapcraftError(Exception):
 class MissingStateCleanError(SnapcraftError):
     fmt = (
         "Failed to clean step: "
-        "State missing for {step!r}. "
+        "Missing state for {step!r}. "
         "This won't work until a complete clean has occurred."
     )
 
@@ -51,10 +51,10 @@ class MissingStateCleanError(SnapcraftError):
 class StepOutdatedError(SnapcraftError):
 
     fmt = (
-        'Failed to re-use files from previous build: '
+        'Failed to reuse files from previous build: '
         'The {step!r} step of {part!r} is out of date:\n'
         '{report}'
-        'In order to continue, please clean that part\'s {step!r} step '
+        'To continue, please clean that part\'s {step!r} step '
         'by running:\n'
         'snapcraft clean {parts_names} -s {step}\n'
     )
--- a/snapcraft/internal/errors.py
+++ b/snapcraft/internal/errors.py
@@ -119,6 +119,10 @@ class ContainerError(SnapcraftError):
 
 
 class ContainerConnectionError(ContainerError):
+    # FIXME this exception is too generic.
+    # https://bugs.launchpad.net/snapcraft/+bug/1734233
+    # --elopio - 20171123
+
     fmt = ('{message}\n'
            'Refer to the documentation at '
            'https://linuxcontainers.org/lxd/getting-started-cli.')
@@ -151,6 +155,10 @@ class ContainerSnapcraftCmdError(ContainerRunError):
 
 
 class SnapdError(SnapcraftError):
+    # FIXME this exception is too generic.
+    # https://bugs.launchpad.net/snapcraft/+bug/1734235
+    # --elopio - 20171123
+
     fmt = '{message}'
 
     def __init__(self, message):
--- a/snapcraft/internal/errors.py
+++ b/snapcraft/internal/errors.py
@@ -39,8 +39,8 @@ class SnapcraftError(Exception):
 
 class MissingStateCleanError(SnapcraftError):
     fmt = (
-        "Failed to clean step {step!r}: "
-        "Missing necessary state. "
+        "Failed to clean step: "
+        "State missing for {step!r}. "
         "This won't work until a complete clean has occurred."
     )
 
@@ -51,7 +51,7 @@ class MissingStateCleanError(SnapcraftError):
 class StepOutdatedError(SnapcraftError):
 
     fmt = (
-        'Failed to resume a snapcraft execution: '
+        'Failed to re-use files from previous build: '
         'The {step!r} step of {part!r} is out of date:\n'
         '{report}'
         'In order to continue, please clean that part\'s {step!r} step '
@@ -218,7 +218,7 @@ class PartNotInCacheError(SnapcraftError):
     fmt = (
         'Failed to get remote part information: '
         'Cannot find the part name {part_name!r} in the cache. '
-        'If it is an existent remote part, run `snapcraft update` and try '
+        'If it is an existing remote part, run `snapcraft update` and try '
         'again. If it has not been defined, consider going to '
         'https://wiki.ubuntu.com/snapcraft/parts to add it.'
     )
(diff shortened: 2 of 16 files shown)


snapcraft/tests/test_lifecycle.py
```diff
@@ -659,12 +661,12 @@ def test_pull_is_dirty_if_target_arch_changes(
...
 
+        self.assertThat(raised.step, Equals('pull'))
+        self.assertThat(raised.part, Equals('part1'))
         self.assertThat(
-            str(raised), Equals(
-                "The 'pull' step of 'part1' is out of date:\n"
-                "The 'deb_arch' project option appears to have changed.\n"
-                "In order to continue, please clean that part's 'pull' step "
-                "by running:\nsnapcraft clean part1 -s pull\n"))
+            raised.report,
```
sergiusens: maybe do some line cleanup on these such that we get
```yaml
self.assertThat(raised.report, Equals(
     "The 'deb_arch' project option appears to have changed.\n"))
```

to make it easier on the eyes wrt indentation

snapcraft/internal/errors.py
```diff
@@ -168,39 +182,44 @@ def __init__(self, filename, message):
...
     fmt = (
-        'Cannot find the definition for part {part_name!r}.\n'
-        'It may be a remote part, run `snapcraft update` '
-        'to refresh the remote parts cache.'
+        'Failed to get part information: '
+        'Cannot find the definition for part {part_name!r}. '
+        'If it is a remote part, run `snapcraft update` '
+        'to refresh the remote parts cache. '
+        'If it is a local part, make sure that it is defined in the '
+        '`snapcraft.yaml`.'
```
kyrofa: Which case here do we feel is more likely: a remote part being used with an out-of-date cache, or the user misspelling a part (or forgetting to put it in, etc.)? Whichever one is more likely, we should probably say first. I kinda feel like the latter is more likely, so perhaps the order here should be switched?

snapcraft/internal/errors.py
```diff
@@ -168,39 +182,44 @@ def __init__(self, filename, message):
...
 class PartNotInCacheError(SnapcraftError):
 
     fmt = (
-        'Cannot find the part name {part_name!r} in the cache. Please '
-        'run `snapcraft update` and try again.\nIf it is indeed missing, '
-        'consider going to https://wiki.ubuntu.com/snapcraft/parts '
-        'to add it.'
+        'Failed to get remote part information: '
+        'Cannot find the part name {part_name!r} in the cache. '
+        'If it is an existent remote part, run `snapcraft update` and try '
```
kyrofa: s/existent/existing/

snapcraft/internal/errors.py
```diff
@@ -168,39 +182,44 @@ def __init__(self, filename, message):
...
-           "missing for the {part_name} part.")
-
-
 class SnapcraftPartConflictError(SnapcraftError):
 
     fmt = (
+        'Failed to run the stage step: '
         'Parts {other_part_name!r} and {part_name!r} have the following file '
         'paths in common which have different contents:\n'
         '{file_paths}\n\n'
```
kyrofa: Do you feel like this is more readable?

```
'Failed to stage {part_name!r}: '
'It and {other_part_name!r} both have the following files, but with different contents:[...]'
```

snapcraft/internal/errors.py
```diff
@@ -224,6 +243,7 @@ def __init__(self, *, part_name, other_part_name, conflict_files):
 class MissingCommandError(SnapcraftError):
 
     fmt = (
+        'Failed to run command: '
         'One or more required commands are missing, please install:'
```
kyrofa: This doesn't read particularly well: "Failed to run command: One or more required commands[...]"

What about "Failed to run command: One or more requires packages[...]" or "utilities" or something similar?

snapcraft/internal/errors.py
```diff
@@ -243,7 +269,10 @@ def __init__(self, error=None):
 class MissingGadgetError(SnapcraftError):
 
     fmt = (
-        'When creating gadget snaps you are required to provide a gadget.yaml file\n'  # noqa
+        'Failed to generate snap metadata: '
```
kyrofa: We actually have three exceptions that all have the same "what." This suggests a common base class to me: `MetadataGenerationError` or some such thing.

snapcraft/internal/errors.py
```diff
@@ -274,14 +303,21 @@ class RequiredPathDoesNotExist(SnapcraftError):
 
 class SnapcraftPathEntryError(SnapcraftError):
 
-    fmt = 'The path {value!r} set for {key!r} in {app!r} does not exist.'
+    fmt = (
+        'Failed to generate snap metadata: '
+        'The path {value!r} set for {key!r} in {app!r} does not exist. '
+        'Make sure taht the files are in the `prime` directory.'
```
kyrofa: s/taht/that/

snapcraft/internal/errors.py
```diff
@@ -291,8 +327,11 @@ def __init__(self, plugin_name, properties):
 class InvalidBuildPropertiesError(SnapcraftError):
 
     fmt = (
+        'Failed to load plugin: '
```
kyrofa: This suggests another common base class.

snapcraft/internal/errors.py
```diff
@@ -301,7 +340,10 @@ def __init__(self, plugin_name, properties):
 
 class StagePackageDownloadError(SnapcraftError):
 
-    fmt = 'Error downloading stage packages for part {part_name!r}: {message}'
+    fmt = (
+        'Failed to fetch stage packages: '
+        'Error downloading stage packages for part {part_name!r}: {message}.'
```
kyrofa: This addition doesn't add meaning, it seems redundant. Perhaps this refactor is good enough?

```
'Failed to fetch stage packages for {part_name!r}: '
'{message}.'
```

snapcraft/internal/errors.py
kalikiana: I would recommend the most important go last actually. As a general rule of thumb that's what stands out more when you read a text.

snapcraft/internal/errors.py
kalikiana: Definitely sounds clearer to me and more like something I'd say

snapcraft/internal/errors.py
kalikiana: I think the problem here is that technically speaking this does refer to the command, and not the package. If we wanted to be more helpful, we'd have to move check_for_command into the Repo class and have it lookup the corresponding package.

snapcraft/internal/errors.py
```diff
@@ -39,8 +39,9 @@ def get_exit_code(self):
 
 class MissingStateCleanError(SnapcraftError):
     fmt = (
-        "Failed to clean step {step!r}: Missing necessary state. This won't "
-        "work until a complete clean has occurred."
+        "Failed to clean step {step!r}: "
+        "Missing necessary state. "
+        "This won't work until a complete clean has occurred."
```
kalikiana: The sentence sounds like a factual statement to me whereas it should be advice on how to proceed. How about  "Run 'snapcraft clean' to start off fresh.".

elopio: That's for the next PR, where I will change the "how to fix" part for snapcraft commands.

snapcraft/internal/errors.py
```diff
@@ -50,6 +51,7 @@ def __init__(self, step):
 class StepOutdatedError(SnapcraftError):
 
     fmt = (
+        'Failed to resume a snapcraft execution: '
```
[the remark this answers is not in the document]
elopio: What about: "Failed to re-use files from previous build" ?

snapcraft/internal/errors.py
elopio: done, thanks.

snapcraft/internal/errors.py
elopio: You are adding too many unnecessary words, it's hard for me to strip them :)
Also, I was kind of trying to keep the first sentence generic, without parameters. Then the second includes parameters to to specify why the problem happened. I'm not sure if that's the best idea, but found that a comfortable consistency. Please review the new version I'm pushing.

snapcraft/internal/errors.py
elopio: I'm leaving it as "packages" for now, but this message seems to require more work

snapcraft/internal/errors.py
elopio: I'm trying to stick with what+why here.
What: failed to fetch stage packages. Why: they couldn't be downloaded.
Another reason to fail to fetch could be that the cache storage is broken, or something maybe?

snapcraft/internal/errors.py
matthewpaulthomas: More detail is better, but “Missing state” might be better than “State missing” because it avoids being misinterpreted as a command to the user.

snapcraft/internal/errors.py
```diff
@@ -51,7 +51,7 @@ def __init__(self, step):
 class StepOutdatedError(SnapcraftError):
 
     fmt = (
-        'Failed to resume a snapcraft execution: '
+        'Failed to re-use files from previous build: '
```
matthewpaulthomas: “Reuse” is seldom hyphenated.

snapcraft/internal/errors.py
```diff
@@ -50,6 +51,7 @@ def __init__(self, step):
 class StepOutdatedError(SnapcraftError):
 
     fmt = (
+        'Failed to re-use files from previous build: '
```
matthewpaulthomas: The verb “reuse” is seldom hyphenated.

snapcraft/internal/errors.py
```diff
@@ -50,6 +51,7 @@ def __init__(self, step):
...
         'The {step!r} step of {part!r} is out of date:\n'
         '{report}'
         'In order to continue, please clean that part\'s {step!r} step '
```
matthewpaulthomas: “In order to” can nearly always be shortened to just “To”.

snapcraft/internal/errors.py
```diff
@@ -169,8 +178,9 @@ def __init__(self, command, app):
 class InvalidContainerRemoteError(SnapcraftError):
 
     fmt = (
-        '{remote!r} is not a valid LXD remote name.\n'
-        "Colons, spaces and slashes can't be used.\n"
+        "Failed to use LXD remote: "
+        "{remote!r} is not a valid name.\n"
+        "Use a LXD remote without colons, spaces and slashes in the name.\n"
```
matthewpaulthomas: This is good, but “a LXD” should be “an LXD”. (“An” precedes words that _sound like_ they start with a vowel: a European, an HTML document.)

snapcraft/internal/errors.py
```diff
@@ -190,41 +204,46 @@ def __init__(self, filename, message):
...
-    fmt = ("Issues while validating snapcraft.yaml: the 'plugin' keyword is "
-           "missing for the {part_name} part.")
-
-
 class SnapcraftPartConflictError(SnapcraftError):
 
     fmt = (
-        'Parts {other_part_name!r} and {part_name!r} have the following file '
-        'paths in common which have different contents:\n'
+        'Failed to run stage step: '
```
matthewpaulthomas: I’m not sure that adding “Failed to run stage step:” is an improvement. Should I know what a “stage step” is?

matthewpaulthomas: I’m not sure that adding “Failed to run stage step:” is an improvement. Should I know what a “stage step” is?

snapcraft/internal/errors.py
kalikiana: It's pronounced "lex-dee" as a matter of fact, so "a LXD remote" is correct.

kalikiana: (And it's worth keeping in mind HTML is a tricky example since some English speakers pronounce the "h" as "haitch")

snapcraft/internal/errors.py
kalikiana: Stage is one of the lifecycle steps. Pull would be another example. It's something for the user of Snapcraft to learn.
Maybe we should point at some documentation here, or suggest one of the "help" commands?

snapcraft/internal/errors.py
elopio: @matthewpaulthomas is the `a` or `an` actually necessary?
What about `Use LXD remote name without colons, spaces ans slashes`

snapcraft/internal/errors.py
elopio: fixed.

snapcraft/internal/errors.py
elopio: fixed.

snapcraft/internal/errors.py
matthewpaulthomas: Sorry, I had no idea about “lex-dee”. In my defence, it’s [no longer in the readme](https://github.com/lxc/lxd/commit/7bcba143993f8956ef47c01e655501d9391fb806#diff-04c6e90faac2675aa89e2176d2eec7d8), and [barely mentioned elsewhere](https://www.google.com/search?q=pronounced+OR+pronunciation+site%3Alinuxcontainers.org). And yes, an article is needed here: the “Failed to ______” lines might be using headline grammar (“Local resident fails to use LXD remote”), but the suggestions at least should use normal grammar (like “slashes in _the_ name”). So, disregard this.

snapcraft/internal/errors.py
matthewpaulthomas: Ok, so how about “Failed to stage:”?

If there’s a directly relevant help topic, something like ‘For more explanation, “snapcraft help error-paths”’ at the end would be nice. The command-line equivalent to progressive disclosure.

elopio: That would be a nice addition, a help topic for each of the stages. What do think @sergiusens, should I report a bug to add it later?
I've update it to "Failed to stage"

CODE_STYLE.md
```diff
@@ -28,6 +28,11 @@ For multiline strings, we prefer to use `textwrap.dedent`:
 
 (from https://docs.python.org/3/library/textwrap.html#textwrap.dedent)
 
+## Errors
+
+Error messages must say what happened, why it happened and what you can do to
+fix it.
```
kyrofa: We're changing how we actually test the errors as well, right? That would probably be worth noting here.

elopio: Good point, added.

elopio:
Tests failed with `websocket: close 1006 (abnormal closure): unexpected EOF`. Rerunning...

PR #1644, opened 2017-10-27
lxd: fix the push in container builds

elopio:

-----

### DIFF
--- a/snapcraft/internal/lxd/_containerbuild.py
+++ b/snapcraft/internal/lxd/_containerbuild.py
@@ -256,6 +256,10 @@ class Containerbuild:
                            is_dangerous=rev.startswith('x'),
                            is_classic=is_classic)
 
+    def _push_file(self, src, dst):
+        subprocess.check_call(['lxc', 'file', 'push',
+                              src, '{}{}'.format(self._container_name, dst)])
+
     def _install_snap(self, name, channel=None,
                       is_dangerous=False,
                       is_classic=False):
--- a/integration_tests/containers/test_container_builds.py
+++ b/integration_tests/containers/test_container_builds.py
@@ -1,0 +1,50 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2017 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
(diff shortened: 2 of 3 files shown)


elopio:
This actually requires #1625 to be really tested in travis.

kalikiana:
@elopio Good catch! Seeing this, I was wondering how it's possible this wasn't caught in our unit tests already which test that code path with `Cleanbuilder` as well as `Project` since the `test_*` methods are inherited by from the `LXDTestCase`... eventually I discovered it's because the tests use `make_cleanbuilder()` and `make_project()` respectively so the subclass would only re-test exactly the same code. So I went ahead and renamed both methods to `make_project` and moved the two Cleanbuilder-specific test cases to a separate class.

snapcraft/internal/lxd/_containerbuild.py
```diff
@@ -256,6 +256,10 @@ def _inject_snap(self, name):
                            is_dangerous=rev.startswith('x'),
                            is_classic=is_classic)
 
+    def _push_file(self, src, dst):
+        subprocess.check_call(['lxc', 'file', 'push',
+                              src, '{}{}'.format(self._container_name, dst)])
+
```
kyrofa: Does it seem odd to anyone else that, while the parent class now has the ability to push files, only the subclass has the ability to pull them? It sorta feels like both should be promoted.

kalikiana: This is a case of need versus ability - I tried to put methods where they're actually needed in the split.  Pull is only used by Cleanbuilder to get the resulting snap.

sergiusens: from a design PoV I would agree with @kyrofa though, but this is also a private method.

kyrofa: @kalikiana I don't mean to debate why they were in cleanbuilder at all, but not having them together (regardless of where they are) is pulling at my OCD. I say we keep them together. @sergiusens I would call them "protected" methods, where users of the class shouldn't be calling them, but subclasses can. Welcome to C++!

sergiusens:
@kalikiana if you are the author of a PR and doing a refactor, don't rely on only unit tests!!!

PR #1646, opened 2017-10-30
sources: enforce default language in subversion info

clobrano:
subversion `_get_source_details` expects the output of `svn info` to
contain an english prefix and this will fail in a non-english environment.

Setting the `env` subprocess.check_output argument to have the `LANG`
key set to the default language `C` to ensure english language in
the command output.

Fixes [LP 1724674](https://bugs.launchpad.net/snapcraft/+bug/1724674)

-----

### DIFF
--- a/snapcraft/internal/sources/_subversion.py
+++ b/snapcraft/internal/sources/_subversion.py
@@ -81,8 +81,11 @@ class Subversion(Base):
         commit = self.source_commit
 
         if not commit:  # retrieve the commit id
-            lines = subprocess.check_output(['svn', 'info', self.source_dir]
-                                            ).decode('utf-8').split('\n')
+            env = os.environ.copy()
+            env['LANG'] = 'C'
+            lines = subprocess.check_output(
+                ['svn', 'info', self.source_dir],
+                env=env).decode('utf-8').split('\n')
             prefix = 'Last Changed Rev: '
             for line in lines:
                 if line.startswith(prefix):


snapcraft/internal/sources/_subversion.py
```diff
@@ -81,8 +81,11 @@ def _get_source_details(self):
         commit = self.source_commit
 
         if not commit:  # retrieve the commit id
-            lines = subprocess.check_output(['svn', 'info', self.source_dir]
-                                            ).decode('utf-8').split('\n')
+            env = dict(os.environ)
```
kalikiana: Please use `os.environ.copy()` here instead.

clobrano: Sure, I'll do

clobrano: Oh sorry, I haven't seen your request about the test. I'll add the test in the next commit

snapcraft/internal/sources/_subversion.py
```diff
@@ -81,8 +81,11 @@ def _get_source_details(self):
...
 
         if not commit:  # retrieve the commit id
-            lines = subprocess.check_output(['svn', 'info', self.source_dir]
-                                            ).decode('utf-8').split('\n')
+            env = os.environ.copy()
+            env['LANG'] = 'C'
+            lines = subprocess.check_output(
+                ['svn', 'info', self.source_dir],
+                env=env).decode('utf-8').split('\n')
             prefix = 'Last Changed Rev: '
```
elopio: ugh, this is so flaky :(
I searched for a better way to do this and found https://stackoverflow.com/a/44632181/2665322
`svn info --show-item revision --no-newline` seems to do the same. @sergiusens @kyrofa any memories of why we are doing this weird thing here?

clobrano: @elopio interesting!
Shouldn't it be `svn info --show-item last-changed-revision --no-newline`, instead?

elopio: makes sense to me.

clobrano: alright, so this PR doesn't have sense anymore, I think.
If you all agree, I'd propose a new one that uses the above command in place of parsing `svn info` output. This way we'll remove any dependency from the language and fix the bug as well.

sergiusens:
#1648 replaces this

PR #1647, opened 2017-10-31
lxd: better surfacing of errors

sergiusens:
We do not want to print the same error twice, so the caller should
produce a silent error with a proper exit code as the message
related to the error would be displayed by the callee.

-----
I do wish ContainerRunError would be more generalized and that these Container* exceptions live inside the lxd package, but oh well, this fixes the problem today.

### DIFF
--- a/snapcraft/internal/lxd/_containerbuild.py
+++ b/snapcraft/internal/lxd/_containerbuild.py
@@ -33,6 +33,8 @@ from urllib import parse
 from snapcraft.internal import common
 from snapcraft.internal.errors import (
         ContainerConnectionError,
+        ContainerRunError,
+        ContainerSnapcraftCmdError,
         SnapdError,
 )
 from snapcraft._options import _get_deb_arch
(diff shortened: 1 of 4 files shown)


snapcraft/internal/lxd/_containerbuild.py
```diff
@@ -151,6 +153,7 @@ def execute(self, step='snap', args=None):
 
     def _container_run(self, cmd, cwd=None, **kwargs):
         sh = ''
+        original_cmd = cmd
```
elopio: maybe safer to copy?

sergiusens:
- `xenial-amd64` is :green_apple: 
- `xenial-armhf` is :apple:, has a:
    - `integration_tests` `test_shared_ros_builds_without_catkin_in_underlay`
       ```
       ERROR: cannot download default sources list from:
       https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/sources.list.d/20-default.list
       ```
    - `integration_tests` `test_asset_recording.ManifestRecordingBuildPackagesTestCase.test_prime_with_build_packages`
       ```
       Failed to establish a new connection: [Errno -3] Temporary failure in name resolution
       ```
    - `demos` `demos_tests.test_qt4_text_editor.Qt4TextEditorTestCase.test_text_editor`
      ```
      Could not resolve proxy: squid.internal
      ```
- `xenial-arm64` is :apple:, has a:
    - `integration_tests` timed out

PR #1648, opened 2017-11-01
sources: get svn revision with --show-item flag

clobrano:
Currently, svn last changed revision is extracted from `svn info`
text output. The same information can be retrieved directly from
the command

`svn info --show-item last-changed-revision`

without the need to parse text.

Adding also the flag `--no-newline` to avoid the need to strip
newline character from the output.

Since there is not need to parse text anymore, this also remove
the dependency from the current locale and then
fixes https://bugs.launchpad.net/snapcraft/+bug/1724674

-----

### DIFF
--- a/snapcraft/internal/sources/_subversion.py
+++ b/snapcraft/internal/sources/_subversion.py
@@ -80,14 +80,12 @@ class Subversion(Base):
         source = self.source
         commit = self.source_commit
 
-        if not commit:  # retrieve the commit id
-            lines = subprocess.check_output(['svn', 'info', self.source_dir]
-                                            ).decode('utf-8').split('\n')
-            prefix = 'Last Changed Rev: '
-            for line in lines:
-                if line.startswith(prefix):
-                    commit = line.replace(prefix, '')
-                    break
+        if not commit:
+            commit = subprocess.check_output(
+                ['svn', 'info',
+                 '--show-item', 'last-changed-revision',
+                 '--no-newline',
+                 self.source_dir]).decode('utf-8').strip()
 
         return {
             'source-commit': commit,


clobrano:
:smile:

snapcraft/internal/sources/_subversion.py
```diff
@@ -80,14 +80,12 @@ def _get_source_details(self):
...
-            prefix = 'Last Changed Rev: '
-            for line in lines:
-                if line.startswith(prefix):
-                    commit = line.replace(prefix, '')
-                    break
+        if not commit:
+            commit = subprocess.check_output(
+                ['svn', 'info',
+                 '--show-item', 'last-changed-revision',
+                 '--no-newline',
```
facundobatista: Why is the `--no-newline` parameter needed? 

Maybe it's better (as simpler) without it?

sergiusens:
On vie, nov 3, 2017 at 7:06 AM, Christian Dywan 
> @kalikiana commented on this pull request.
> 
> Nice simplification!
> 
> Can we by any chance unit test this? The fact that the test pass 
> unchanged concerns me slightly...
> 
> 
> 
The unit tests for all things vcs are not mocked commands
>

snapcraft/internal/sources/_subversion.py
clobrano: @facundobatista since we are already using  `str.strip`, `--no-newline` should not be necessary, however I decided to keep it to make it more explicit.

clobrano:
@elopio found the solution, I just implemented it, but thank you :)

sergiusens:
@kalikiana can you make sure this works manually as well?

kalikiana:
Verifying the tests pass:
    LANGUAGE=es SNAPCRAFT_TEST_MOCK_MACHINE=armv7l SUDO_UID=5678 python3 -m unittest run snapcraft.tests.sources.test_subversion.SubversionDetailsTestCase.test_svn_details_commit -f

I couldn't actually find a real world case that is affected. I thought at first ` LANGUAGE=en SNAPCRAFT_BUILD_INFO=1 snapcraft` in a project using svn would have hit the problem but it works perfectly... @clobrano Did you hit the issue building an actual snap as well?

clobrano:
Not with a real snap, but running the test in a VM with es_AR  locale

PR #1654, opened 2017-11-02
autotools: cross-compile using --host instead of env

kyrofa:

-----

Take a look at the `configure` output for our autotools cross-compile integration test (I'm doing this by hand, running `snapcraft --target-arch=armhf`):

```
[...]
./configure --prefix= --host=armhf
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for armhf-strip... no
checking for strip... strip
[...]
```

Note that it's using the build system's `strip`, not the cross-toolchain's `strip`. What is `armhf-strip` anyway? Nothing, I tell you! We're setting --host to the deb arch instead of the triplet.

Currently the autotools plugin uses the CC and CXX environment variables to force autotools to cross-compile. However, autotools will do this on its own if one sets the --host parameter correctly, and will actually do a better job (e.g. use the cross-toolchain's strip as well).

Remove the environment variables, and set the --host to the target triplet if cross-compiling. More info on how to cross-compile using autotools (including details on --host) can be found [here](https://www.gnu.org/software/automake/manual/html_node/Cross_002dCompilation.html).

### DIFF
--- a/integration_tests/plugins/test_autotools_plugin.py
+++ b/integration_tests/plugins/test_autotools_plugin.py
@@ -29,8 +29,8 @@ class AutotoolsPluginTestCase(integration_tests.TestCase):
         self.run_snapcraft('stage', 'autotools-hello')
 
         binary_output = self.get_output_ignoring_non_zero_exit(
-            os.path.join(self.stage_dir, 'bin', 'test'))
-        self.assertThat(binary_output, Equals('Hello world\n'))
+            os.path.join(self.stage_dir, 'bin', 'hello'))
+        self.assertThat(binary_output, Equals('Hello, world!\n'))
 
     def test_cross_compiling(self):
         if snapcraft.ProjectOptions().deb_arch != 'amd64':
@@ -39,5 +39,5 @@ class AutotoolsPluginTestCase(integration_tests.TestCase):
         self.run_snapcraft(['build', '--target-arch=arm64'],
                            'autotools-hello')
         binary = os.path.join(self.parts_dir, 'make-project', 'install', 'bin',
-                              'test')
+                              'hello')
         self.assertThat(binary, HasArchitecture('aarch64'))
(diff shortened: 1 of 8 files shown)


kyrofa:
@lool could you take a look at this when you're able?

lool:
I've tested cross-compilations of a hello world snap for armhf and arm64 which failed before and now pass (mainly thanks to the change from --host=armhf to --host=arm-linux-gnueabihf). Test suite also passes (unit and static).

Other changes look good. Removal of the env vars definitions could be a regression if the recipe was working before, but it's better to do this now and see if it's useful to set it or not.

kalikiana:
Some thoughts:
* Shouldn't we test that the correct `strip` is being used now?
* Should we be testing that CC and CXX are set through --host?
* @lool What's "a hello world snap"? The autotools test snap in the snapcraft repo doesn't require this fix at all. It'd be awesome to have something we can test this with.

kalikiana:
Replying to my own points raised above: using gnu-hello as the basis for the integration test does test the bug fix and should be quite sufficient.

kalikiana:
`Logging into Launchpad...
Amount of emails to check the CLA for 3
Checking the CLA for 'malept@users.noreply.github.com'
The contributor does not have a Launchpad account.`

kalikiana:
For the record, failed with:

Collecting docopt==0.6.2 (from -r requirements.txt (line 3))
  Downloading docopt-0.6.2.tar.gz
Collecting file-magic==0.3.0 (from -r requirements.txt (line 4))
Exception:

  File "/usr/lib/python3.5/ssl.py", line 575, in read
    v = self._sslobj.read(len, buffer)
ConnectionResetError: [Errno 104] Connection reset by peer

TypeError: unsupported operand type(s) for -=: 'Retry' and 'int'

PR #1718, opened 2017-11-06
lxd: let lxd choose the architecture

mwhudson:
lxd implements exactly the same logic as snapcraft does to choose an
architecture when one is not specified, so remove snapcraft's copy.

It is still necessary to record the architecture of the created
container to detect whether snaps can just be copied across, but this
is now done after the container is started, which will still be correct
when/if support for building on a non-default image is merged.

### DIFF
--- a/snapcraft/internal/lxd/_containerbuild.py
+++ b/snapcraft/internal/lxd/_containerbuild.py
@@ -25,7 +25,6 @@ import requests_unixsocket
 import shutil
 import sys
 import tempfile
-import yaml
 from contextlib import contextmanager
 import subprocess
 import time
@@ -34,6 +33,7 @@ from urllib import parse
 from snapcraft.internal import common
 from snapcraft.internal.errors import (
         ContainerConnectionError,
+        ContainerError,
         ContainerRunError,
         ContainerSnapcraftCmdError,
         SnapdError,
@@ -68,27 +68,12 @@ class Containerbuild:
             remote = _get_default_remote()
         _verify_remote(remote)
         self._container_name = '{}:snapcraft-{}'.format(remote, container_name)
-        server_environment = self._get_remote_info()['environment']
-        # Use the server architecture to avoid emulation overhead
-        try:
-            kernel = server_environment['kernel_architecture']
-        except KeyError:
-            kernel = server_environment['kernelarchitecture']
-        self._server_arch = _get_deb_arch(kernel)
-        if not self._server_arch:
-            raise ContainerConnectionError(
-                'Unrecognized server architecture {}'.format(kernel))
-        self._image = 'ubuntu:xenial/{}'.format(self._server_arch)
+        self._image = 'ubuntu:xenial'
         # Use a temporary folder the 'lxd' snap can access
         lxd_common_dir = os.path.expanduser(
             os.path.join('~', 'snap', 'lxd', 'common'))
         os.makedirs(lxd_common_dir, exist_ok=True)
         self.tmp_dir = tempfile.mkdtemp(prefix='snapcraft', dir=lxd_common_dir)
-
-    def _get_remote_info(self):
-        remote = self._container_name.split(':')[0]
-        return yaml.load(subprocess.check_output([
-            'lxc', 'info', '{}:'.format(remote)]).decode())
 
     @contextmanager
     def _container_running(self):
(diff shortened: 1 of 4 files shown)


kalikiana:
Did you realize that `_get_remote_info` is implemented as `yaml.load` on `lxc info`? You seem to be re-implementing the same, but splitting lines by hand.

Perhaps you can elaborate on what you're trying to do?

mwhudson:
> Did you realize that _get_remote_info is implemented as yaml.load on lxc info? 

I did not.

> You seem to be re-implementing the same, but splitting lines by hand.

lxc info $container does not produce YAML, sadly.

> Perhaps you can elaborate on what you're trying to do?

I want to allow arbitrary images to be used for cleanbuild as in my other PR.

The logic around whether to append /$deb_arch or not to the image name in that branch is a bit confusing, and moreover totally unnecessary given that lxd applies the same default architecture on the inside anyway.

My first thought was to delete all the arch detection, but _server_arch is also used to detect if snaps can be directly copied to the container so I readded (badly, it turns out) code to read it from the image. Something like this is needed to allow `snapcraft cleanbuild --image ubuntu:xenial/i386` to work on an amd64 host, although I really should rename the variable to _container_arch for that.

mwhudson:
I don't think the travis failure is my fault this time?

kalikiana:
You seem to be re-implementing the same, but splitting lines by hand.
    lxc info $container does not produce YAML, sadly.

Wow! I didn't realize the output format is not YAML in this case... I feel sticking to the server version might more maintainable even if we don't use it in the image?
@elopio @kyrofa Any opinion on this?

kyrofa:
> lxc info $container does not produce YAML, sadly.

Indeed, you can see the different cases [here](https://github.com/lxc/lxd/blob/693df2ceddd3287ff1d5bcefea848259899f8151/lxc/info.go): it only seems to be YAML if the container name is not given.

> I feel sticking to the server version might more maintainable even if we don't use it in the image?

I'm not sure what you mean when you say "server version."

I like this change, overall.

kalikiana:
@kyrofa LXD server info as opposed to container info ie. the code we had before, which was parsable as YAML as opposed to the new code which requires splitting strings by hand.

mwhudson:
> @kyrofa LXD server info as opposed to container info ie. the code we had before, which was parsable as YAML as opposed to the new code which requires splitting strings by hand.

That would be nice but it not answering the right question, really it's the container arch that matters. With a newer lxd than in Xenial you can run query /1.0/containers/$container and parse that as json (or yaml I guess) but this is the best we can do for xenial, at least @stgraber didn't seem to have any better ideas when I was talking to him about it...

mwhudson:
> That would be nice but it not answering the right question, really it's the container arch that matters.

To be clear those are going to be the same today, but I want to land something like https://github.com/mwhudson/snapcraft/commit/c095678a4e624043a269cd5a1fc30f4720964d11 next, after which they might not be the same.

kyrofa:
> @kyrofa LXD server info as opposed to container info ie. the code we had before, which was parsable as YAML as opposed to the new code which requires splitting strings by hand.

Eh? Are you talking about [this](https://github.com/snapcore/snapcraft/pull/1718/files#diff-5b011603826eb296b0f1a98486fe3a3dL90)? That's not YAML, and we shouldn't parse it as such.

kyrofa:
> @kyrofa LXD server info as opposed to container info ie. the code we had before, which was parsable as YAML as opposed to the new code which requires splitting strings by hand.

Eh? Are you talking about [this](https://github.com/snapcore/snapcraft/pull/1718/files#diff-5b011603826eb296b0f1a98486fe3a3dL90)? That's not YAML, and we shouldn't parse it as such.

kalikiana:
For the record, one of the jobs failed with:

test_python_plugin.PythonPluginTestCase.test_build_doesnt_get_bad_install_directory_lp1586546

File "/usr/lib/python3.5/ssl.py", line 575, in read
    v = self._sslobj.read(len, buffer)
ConnectionResetError: [Errno 104] Connection reset by peer

File "/usr/share/python-wheels/urllib3-1.13.1-py2.py3-none-any.whl/urllib3/util/retry.py", line 228, in increment
    total -= 1
TypeError: unsupported operand type(s) for -=: 'Retry' and 'int'

Downloading 'PyYAML-3.11.tar.gz'  29%  

BlockingIOError: [Errno 11] write could not complete without blocking

mwhudson:
Thanks for the approvals. Am I right to think that this can't be merged until 2.35 is out?

kyrofa:
> Thanks for the approvals. Am I right to think that this can't be merged until 2.35 is out?

Indeed, we're focusing on only landing things necessary for autopkgtests to pass until we can ship 2.35, then we'll start landing stuff for 2.36. Thanks for your patience!

mwhudson:
Can this be merged now?

sergiusens:
@mwhudson yeah, it has two approvals, let me update the branch to get a green run and if the test deities allow for it, we would get a green result and merge it into master

PR #1645, opened 2017-10-29
ruby plugin: new stable release 2.4.2

nathanhaines:
- [x] Have you followed the [guidelines for contributing](https://github.com/snapcore/snapcraft/blob/master/CONTRIBUTING.md)?
- [x] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [ ] If this is a bugfix. Have you checked that there is a bug report open for the issue you are trying to fix on [bug reports](https://bugs.launchpad.net/snapcraft)?
- [ ] If this is a new feature. Have you discussed the design on the [forum](https://forum.snapcraft.io)?
- [ ] Have you successfully run `./runtests.sh static`?
- [ ] Have you successfully run `./runtests.sh unit`?

-----

PR #1721, opened 2017-11-08
integration tests: use ruby v2.4.2

kyrofa:

-----

Anything earlier fails to build with GCC7 on Artful.


elopio:
Thanks @kyrofa. The travis test failed to download lxd, maybe because of the canonical maintenance. I've restarted the jobs.

The default `ruby-version` on the plugin is `2.4.0`. Wouldn't it be better to remove it from all the integration tests, and update only the default?

sergiusens:
How about just piggybacking on #1645 and making it just work through that?

sergiusens:
Actually, that should be the path.

kyrofa:
Yeah that's fine as well. I'll propose removing the version from these instead, and we can land both.

PR #1722, opened 2017-11-09
unit tests: reset log level after test

kyrofa:

-----

Several tests (the command tests, parser tests, etc.) call the main executables with `--debug`, which changes the Snapcraft log level. However, when running the entire suite together, that level tends to stick, and can make later tests (that depend on debug mode NOT being enabled) fail. I hit this pretty much every time trying to run adt locally.

Reset the log level to the default after each test.

### DIFF
--- a/snapcraft/tests/__init__.py
+++ b/snapcraft/tests/__init__.py
@@ -27,7 +27,10 @@ import testscenarios
 import testtools
 
 import snapcraft
-from snapcraft.internal import common
+from snapcraft.internal import (
+    common,
+    log,
+)
 from snapcraft.tests import fake_servers, fixture_setup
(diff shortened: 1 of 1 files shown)


elopio:
I'm not sure about this. Does it fix it for you?

The log.configure function is only called for cli commands. There are many things I hate about how we use the log, but I think that each test that depends on the log should configure it, instead of relying on the set up.

Instead of calling the CLI configure even for tests that won't use the CLI, I think that the fake logger output should be reset. I thought that was the case, maybe it's this nuke option that we are not passing? https://github.com/testing-cabal/fixtures/blob/master/fixtures/_fixtures/logger.py#L85

kyrofa:
Hopefully also solved by #1724, closing for now anyway.

PR #1725, opened 2017-11-09
tests: share the cache in ros tests

elopio:

-----

### DIFF
--- a/snapcraft/tests/fixture_setup.py
+++ b/snapcraft/tests/fixture_setup.py
@@ -1024,3 +1024,23 @@ class FakeSnapd(fixtures.Fixture):
         self.server.shutdown()
         self.server.socket.close()
         thread.join()
+
+
+class SharedCache(fixtures.Fixture):
+
+    def __init__(self, name):
+        super().__init__()
+        self.name = name
+
+    def setUp(self):
+        super().setUp()
+        shared_cache_dir = os.path.join(
+            tempfile.gettempdir(), 'snapcraft_test_cache_{}'.format(self.name))
+        os.makedirs(shared_cache_dir, exist_ok=True)
+        self.useFixture(fixtures.EnvironmentVariable(
+            'XDG_CACHE_HOME', shared_cache_dir))
+        patcher = mock.patch(
+            'xdg.BaseDirectory.xdg_cache_home',
+            new=os.path.join(shared_cache_dir))
+        patcher.start()
+        self.addCleanup(patcher.stop)
--- a/snapcraft/tests/integration/general/test_catkin.py
+++ b/snapcraft/tests/integration/general/test_catkin.py
@@ -19,12 +19,18 @@ import subprocess
 import tempfile
 
 from snapcraft.tests import (
+    fixture_setup,
     integration,
     skip
 )
 
 
 class CatkinTestCase(integration.TestCase):
+
+    def setUp(self):
+        super().setUp()
+        # share the cache in all tests.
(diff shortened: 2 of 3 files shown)


kyrofa:
What about the tests that aren't in the snapd suite? They don't all strictly need to run (although it wouldn't hurt, it would slow things down further).

elopio:
@kyrofa we can turn this into a ros shared cache fixture. It could be a base class, but then we end up with multiple inheritance which makes me nervous.

sergiusens:
Yes, I would say no on multiple inheritance as well

snapcraft/tests/fixture_setup.py
```diff
@@ -1013,3 +1013,23 @@ def _stop_fake_server(self, thread):
...
+class SharedCache(fixtures.Fixture):
+
+    def __init__(self, name):
+        super().__init__()
+        self.name = name
+
+    def setUp(self):
+        super().setUp()
+        shared_cache_dir = os.path.join(
+            tempfile.gettempdir(), 'snapcraft_test_cache_{}'.format(self.name))
```
sergiusens: This makes it so that there is no share cache

sergiusens: Oh, my mistake, just saw what `tempfile.gettempdir` is doing. Why not set it the the actual `xdg_cache_home`?

elopio: The xdg_cache_home is patched in all the tests. I started complicating things trying to make an exception to that rule only for ros on the other PR. So, I jumped back a little and thought that we don't really want to pollute home in any test. This sounded like a clean and better approach to me.

sergiusens:
On vie, nov 10, 2017 at 12:58 PM, Leo Arias <notifications@github.com> 
wrote:
> @elopio commented on this pull request.
> 
> In snapcraft/tests/fixture_setup.py:
> 
> > @@ -1013,3 +1013,23 @@ def _stop_fake_server(self, thread):
>          self.server.shutdown()
>          self.server.socket.close()
>          thread.join()
> +
> +
> +class SharedCache(fixtures.Fixture):
> +
> +    def __init__(self, name):
> +        super().__init__()
> +        self.name = name
> +
> +    def setUp(self):
> +        super().setUp()
> +        shared_cache_dir = os.path.join(
> +            tempfile.gettempdir(), 
> 'snapcraft_test_cache_{}'.format(self.name))
> The xdg_cache_home is patched in all the tests. I started 
> complicating things trying to make an exception to that rule only for 
> ros on the other PR. So, I jumped back a little and thought that we 
> don't really want to pollute home in any test. This sounded like a 
> clean and better approach to me.
> 

But you are using tmp! Maybe better use /var/tmp
>

elopio:
@sergiusens I'm using whatever tempdir returns. Instead of hardcoding /var/tmp, I prefer to pass it in an env var:
```
>>> import os
>>> os.environ['TMPDIR'] = '/var/tmp'
>>> import tempfile
>>> tempfile.gettempdir()
'/var/tmp'
```

That way the caller of the test can control the tmp to use.

kyrofa:
@elopio do you plan on adding that variable definition into the autopkgtest, then?

elopio:
@kyrofa it's already in the integration tests: https://github.com/snapcore/snapcraft/blob/master/debian/tests/integrationtests#L19
I'm not sure if we should add it to the snaps tests too, because we want to remove the ros tests from there soon.

kyrofa:
Nah, works for me.

PR #1726, opened 2017-11-10
schema: sources should not have defaults

sergiusens:
Having a default for the source entry leads to hidden and unexpected behavior.

There is no documentation where we omit source so most users should not be in a broken state by this change.

-----
Some of the changes are caused by autopep8, they seem sane so I left them.
This work brought up a thought that we are not recording the defaults to use later, so maybe after this there is more work to make sure the default values get loaded into the project yaml that is later going to be recorded so these can be tracked.

### DIFF
--- a/snapcraft/internal/project_loader/_config.py
+++ b/snapcraft/internal/project_loader/_config.py
@@ -96,11 +96,16 @@ class Config:
 
         self.snapcraft_yaml_path = get_snapcraft_yaml()
         snapcraft_yaml = _snapcraft_yaml_load(self.snapcraft_yaml_path)
+        snapcraft_yaml = _mangle_yaml(snapcraft_yaml)
 
         self._validator = Validator(snapcraft_yaml)
         self._validator.validate()
 
         snapcraft_yaml = self._process_remote_parts(snapcraft_yaml)
+
+        # Now that remote parts are loaded we can finally validate sources.
+        self._validator.validate(snapcraft_yaml, validate_sources=True)
+
         snapcraft_yaml = self._expand_filesets(snapcraft_yaml)
 
         # both confinement type and build quality are optionals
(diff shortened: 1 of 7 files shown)


snapcraft/internal/project_loader/_config.py
```diff
@@ -308,3 +313,15 @@ def _expand_filesets_for(step, properties):
             new_step_set.append(item)
 
     return new_step_set
+
+
+def _mangle_yaml(snapcraft_yaml):
```
elopio: I can't think of a good name here, but maybe we can improve mangle.
_add_sources_for_sourceless_plugins?

snapcraft/internal/project_loader/_config.py
```diff
@@ -96,11 +96,16 @@ def __init__(self, project_options=None):
...
         snapcraft_yaml = _snapcraft_yaml_load(self.snapcraft_yaml_path)
+        snapcraft_yaml = _mangle_yaml(snapcraft_yaml)
 
         self._validator = Validator(snapcraft_yaml)
         self._validator.validate()
 
         snapcraft_yaml = self._process_remote_parts(snapcraft_yaml)
+
+        # Now that remote parts are loaded we can finally validate sources.
+        self._validator.validate(snapcraft_yaml, validate_sources=True)
```
elopio: So, are we validating twice now?

elopio:
This looks good to me. I know a few projects where we are not defining a source, but it's mostly hacking around other things, so I'm not worried if we break them. I think we won't, anyway.
There is one case where I have found I need something like source: nil: https://github.com/elopio/raiden/blob/082acd1cb76a3a71dffead08685b8ee1516b7721/snap/snapcraft.yaml#L40
Is that something we will ever support?

kyrofa:
(several test failures here).

sergiusens:
On lun, nov 13, 2017 at 7:33 PM, Kyle Fazzari 
> (several test failures here).
> 

Yes, I am punting this into the future as my mind is now divided on how 
to handle this.
>

PR #1734, opened 2017-11-14
recording: don't log the command when image info fails

elopio:

-----
In here I'm removing the part that we don't control from the message, and at the same time making it look a little nicer for the issue that Christian pointed out in the original branch.

### DIFF
--- a/snapcraft/tests/test_lxd.py
+++ b/snapcraft/tests/test_lxd.py
@@ -635,8 +635,9 @@ class FailedImageInfoTestCase(LXDBaseTestCase):
             exception=CalledProcessError,
             kwargs=dict(cmd='testcmd', returncode=1, output='test output'),
             expected_warn=(
-                "Failed to get container image info: Command 'testcmd' "
-                "returned non-zero exit status 1, output: test output\n"
+                "Failed to get container image info: "
+                "`lxc image list --format=json ubuntu:xenial/amd64` "
+                "returned with exit code 1, output: test output\n"
                 "It will not be recorded in manifest.\n"))),
         ('JSONDecodeError', dict(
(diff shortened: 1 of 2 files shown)


snapcraft/tests/test_lxd.py
```diff
@@ -635,8 +635,9 @@ class FailedImageInfoTestCase(LXDBaseTestCase):
             exception=CalledProcessError,
             kwargs=dict(cmd='testcmd', returncode=1, output='test output'),
             expected_warn=(
-                "Failed to get container image info: Command 'testcmd' "
-                "returned non-zero exit status 1, output: test output\n"
+                "Failed to get container image info: "
+                "`lxc image list --format=json ubuntu:xenial/amd64` "
+                "returned with exit code 1, output: test output\n"
```
kyrofa: Are we wanting to get rid of checking the string completely?

elopio: I am not sure, because there was a bug here found by @kalikiana that we were not testing. One could argue that once it's fixed and tested manually, it's not likely to regress, so I might be ok removing this test, or doing a regex check as we discussed earlier. But now we control all the pieces of information shown, so it felt weird and I reverted the regex assertion.

What about this? I can extract a method called `_get_image_info`. That method can raise a two exceptions that will take care of the formatting, and we can test them like in #1640. Then the caller catches those exceptions and puts them into the warn log, so we just test that in case of exception, the text from the exception is logged.

elopio: No. xenial is hardcoded, and the arch is faked in https://github.com/snapcore/snapcraft/pull/1734/files#diff-4212d20bfff850afa80d654279d34beeL630

Issue #1666, opened 2017-11-06
Refactor to support elf patching


sergiusens:
Estimate: 2 days

PR #1744, opened 2017-11-17
elf: conversion from libraries

sergiusens:
Converting the libraries module into elf and making it more focused on handling elf
files as a whole.

The logic also has type hinting annotations with the associated fixes from running
with mypy.

Fixes: #1666

-----


kyrofa:
I'm all for experimenting with type checking, but if we go this route, I think it's probably a good idea to bake type checking into our static tests. Otherwise this can quickly get stale if we're not all doing it, which is worse than not doing it at all. Thoughts?

sergiusens:
Yeah, I have it locally, but it breaks with the old flake8 version we have.
Also, starting 2.37 all code should be mypy friendly.

snapcraft/internal/elf.py
```diff
@@ -91,27 +94,60 @@ def _get_system_libs():
...
+
+    fs_encoding = sys.getfilesystemencoding()
+
+    for part_file in file_list:
+        # Filter out object (*.o) files-- we only care about binaries.
+        if part_file.endswith('.o'):
+            continue
+
+        # No need to crawl links-- the original should be here, too.
+        path: str = os.path.join(root, part_file)
```
kyrofa: According to [this](https://mypy.readthedocs.io/en/latest/cheat_sheet_py3.html), variable annotations (PEP 526) were added in Python 3.6. Given that Xenial is only 3.5.2, that will probably be an issue.

snapcraft/internal/elf.py
```diff
@@ -91,27 +94,60 @@ def _get_system_libs():
...
     # Classic confinement won't do what you want with library crawling
     # and has a nasty side effect described in LP: #1670100
-    libs = [l for l in libs if not l.startswith('/snap/')]
+    libs = {l for l in libs if not l.startswith('/snap/')}
 
     return libs
+
+
+def get_elf_files(root: str, file_list: Sequence[str]) -> FrozenSet[str]:
+    """Return a frozenset of elf files from file_list prepended with root."""
```
kyrofa: Appreciate the addition of the docstring, but again would like to see more here, covering the parameters.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -444,7 +441,10 @@ def prime(self, force=False):
         snap_files, snap_dirs = self.migratable_fileset_for('prime')
         _migrate_files(snap_files, snap_dirs, self.stagedir, self.primedir)
 
-        dependencies = _find_dependencies(self.primedir, snap_files)
+        elf_files = elf.get_elf_files(self.primedir, snap_files)
+        dependencies = set()
+        for elf_file in elf_files:
+            dependencies = dependencies.union(elf.get_dependencies(elf_file))
```
kyrofa: This creates multiple copies of the dependencies. Would probably be more efficient to update it in place, either with `dependencies |= elf.get_dependencies(elf_file)` or `dependencies.update(elf.get_dependencies(elf_file))`.

snapcraft/internal/elf.py
kyrofa: Yeah, Travis is barfing.

sergiusens:
On Nov 17 2017, at 6:09 pm, Kyle Fazzari <notifications@github.com> wrote:
>
> @kyrofa requested changes on this pull request.
>
>
> Content looks good, although I have a few suggestions. Most importantly though, this syntax needs a revamp to be compatible with Python 3.5.
>
>
> In snapcraft/internal/elf.py (https://github.com/snapcore/snapcraft/pull/1744?recipient=reply%2B00117fe580fc4a41652bbee5dd5b53a696ef2f50ff1aba1c92cf000000011627120692a169ce10636356%40reply.github.com#discussion_r151786332):
>
> > try: ldd_out = common.run_output(['ldd', elf]).split('\n') except subprocess.CalledProcessError: logger.warning( 'Unable to determine library dependencies for {!r}'.format(elf)) - return [] - ldd_out = [l.split() for l in ldd_out] - ldd_out = [l[2] for l in ldd_out if len(l) > 2 and os.path.exists(l[2])] + return set() + ldd_out_split = [l.split() for l in ldd_out] + ldd_out_list = [l[2] for l in ldd_out_split + if len(l) > 2 and os.path.exists(l[2])]
>
> Why not continue using ldd_out here? Seems a bit more efficient, and you re-use libs below for similar progressive filtering.
>
>
>
>
>
>

mypy and types, such magic :-) List strings and list of lists (the former) are not the same type :-)

>
> In snapcraft/internal/elf.py (https://github.com/snapcore/snapcraft/pull/1744?recipient=reply%2B00117fe580fc4a41652bbee5dd5b53a696ef2f50ff1aba1c92cf000000011627120692a169ce10636356%40reply.github.com#discussion_r151788008):
>
> > @@ -91,27 +94,60 @@ def _get_system_libs(): return _libraries -def get_dependencies(elf): - """Return a list of libraries that are needed to satisfy elf's runtime. +def get_dependencies(elf: str) -> Set[str]: + """Return a set of libraries that are needed to satisfy elf's runtime.
>
> Would be nice to see Sphinx-style docs including the parameters.
>
>

OK, I am divided now, this syntax is pretty much complete on that aspect, sort of making sphinx obsolete. But yeah, we can still use it until we have a strategy.

>
>
>
>
>
> In snapcraft/internal/elf.py (https://github.com/snapcore/snapcraft/pull/1744?recipient=reply%2B00117fe580fc4a41652bbee5dd5b53a696ef2f50ff1aba1c92cf000000011627120692a169ce10636356%40reply.github.com#discussion_r151788377):
>
> > return libs + + +def get_elf_files(root: str, file_list: Sequence[str]) -> FrozenSet[str]: + """Return a frozenset of elf files from file_list prepended with root."""
>
> Appreciate the addition of the docstring, but again would like to see more here, covering the parameters.
>
>

Yeah, as mentioned, I was happy enough with the types in the params :-)

>
> In snapcraft/internal/pluginhandler/__init__.py (https://github.com/snapcore/snapcraft/pull/1744?recipient=reply%2B00117fe580fc4a41652bbee5dd5b53a696ef2f50ff1aba1c92cf000000011627120692a169ce10636356%40reply.github.com#discussion_r151789370):
>
> > @@ -444,7 +441,10 @@ def prime(self, force=False): snap_files, snap_dirs = self.migratable_fileset_for('prime') _migrate_files(snap_files, snap_dirs, self.stagedir, self.primedir) - dependencies = _find_dependencies(self.primedir, snap_files) + elf_files = elf.get_elf_files(self.primedir, snap_files) + dependencies = set() + for elf_file in elf_files: + dependencies = dependencies.union(elf.get_dependencies(elf_file))
>
> This creates multiple copies of the dependencies. Would probably be more efficient to update it in place, either with dependencies |= elf.get_dependencies(elf_file) or dependencies.update(elf.get_dependencies(elf_file)) .
>
>

I think I went with union as it felt more readable, using `|=` might have caused issues (oh, the perils of PRs written so long ago). I'll try `update` first and see if mypy refrains from complaining.

sergiusens:
On Nov 17 2017, at 6:29 pm, Kyle Fazzari <notifications@github.com> wrote:
>
> @kyrofa commented on this pull request.
>
>
> In snapcraft/internal/elf.py (https://github.com/snapcore/snapcraft/pull/1744?recipient=reply%2B00117fe5ca892882feec174855e5be2e46570ff5e8414d7792cf00000001162716a892a169ce10636356%40reply.github.com#discussion_r151793791):
>
> > + """Return a frozenset of elf files from file_list prepended with root.""" + ms = magic.open(magic.NONE) + if ms.load() != 0: + raise RuntimeError('Cannot load magic header detection') + + elf_files = set() + + fs_encoding = sys.getfilesystemencoding() + + for part_file in file_list: + # Filter out object (*.o) files-- we only care about binaries. + if part_file.endswith('.o'): + continue + + # No need to crawl links-- the original should be here, too. + path: str = os.path.join(root, part_file)
>
> Yeah, Travis is barfing.
>
>

The flake8 update PR I have up should take care of the barf #1745

kyrofa:
You realize I have to respond to you out of thread now, right? :stuck_out_tongue: 

> mypy and types, such magic :-) List strings and list of lists (the former) are not the same type :-)

Might be worth using Any for that type, then.

> OK, I am divided now, this syntax is pretty much complete on that aspect, sort of making sphinx obsolete. But yeah, we can still use it until we have a strategy.

I disagree. Sphinx docs include type information yes, but they also include what this parameter _is_, and how it's used, which super useful. If you want to drop type specifiers in sphinx and use this instead that's fine, but I think documenting the parameters is still important.

sergiusens:
On Nov 17 2017, at 7:17 pm, Kyle Fazzari <notifications@github.com> wrote:
>
> You realize I have to respond to you out of thread now, right? 😛�
>
> >
> > mypy and types, such magic :-) List strings and list of lists (the former) are not the same type :-)
> >
>
>
> Might be worth using Any for that type, then.
>
>

I actually prefer the change fwiw, it conveys to the reader that something is changing in a quick way without needing to look too deep into the list comprehension.
> >
> > OK, I am divided now, this syntax is pretty much complete on that aspect, sort of making sphinx obsolete. But yeah, we can still use it until we have a strategy.
> >
>
>
> I disagree. Sphinx docs include type information yes, but they also include what this parameter is, and how it's used, which super useful. If you want to drop type specifiers in sphinx and use this instead that's fine, but I think documenting the parameters is still important.
>
>

Oh, I certainly agree on documenting parameters in a human descriptive way, I just no longer thing sphinx is the right tool :-)

snapcraft/internal/elf.py
```diff
@@ -91,27 +94,60 @@ def _get_system_libs():
     return _libraries
 
 
-def get_dependencies(elf):
-    """Return a list of libraries that are needed to satisfy elf's runtime.
+def get_dependencies(elf: str) -> Set[str]:
+    """Return a set of libraries that are needed to satisfy elf's runtime.
```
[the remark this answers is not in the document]
sergiusens: done

PR #1746, opened 2017-11-20
cli: add version command

gsilvapt:
* As discussed in [LP #1590349](https://bugs.launchpad.net/snapcraft/+bug/1590349), snapcraft should return the version
number when called like snapcraft version and snapcraft --version.

-----


snapcraft/tests/unit/commands/test_version.py
```diff
@@ -19,7 +19,11 @@
...
 
 class VersionCommandTestCase(CommandBaseTestCase):
-
     def test_has_version(self):
         result = self.run_command(['--version'])
         self.assertThat(result.exit_code, Equals(0))
+
+    def test_has_version_without_hyphens(self):
+        result = self.run_command(['version'])
+        print(result.output)
```
kalikiana: In general, tests shouldn't print anything - instead I would recommend you do something like `self.assertThat(result.output, Equals(snapcraft.__version__))` which will only output the value if it fails.

snapcraft/tests/unit/commands/test_version.py
```diff
@@ -19,7 +19,10 @@
 
 
 class VersionCommandTestCase(CommandBaseTestCase):
-
     def test_has_version(self):
         result = self.run_command(['--version'])
         self.assertThat(result.exit_code, Equals(0))
+
+    def test_has_version_without_hyphens(self):
```
kyrofa: I'd like to see one more test ensuring that the output from `--version` == the output from `version`. Because I can tell you right now, it's not :wink: .

gsilvapt:
Thank you for your review. Indeed it is important to add that.

But I do not want to hard code the text string before the method returns the version number. Maybe I am missing something from the click framework to also include in the method I wrote? This would actually allow the results to be tested.

kyrofa:
To quote the `click.version_option` section in the [click docs](http://click.pocoo.org/5/api/):

> **message** – custom message to show instead of the default (`'%(prog)s, version %(version)s'`)

I don't have a problem hard-coding that string here. However, if your concern is that Click might change out from under us, you can also add the `message` parameter where we use `version_option` to ensure they always stay in sync.

gsilvapt:
Both methods are now printing the same output and the same text. A test was added to guarantee this.

kalikiana:
@gsilvapt I see a failure from flake8:
`snapcraft/cli/__init__.py:102: error: Unexpected keyword argument "message" for "version_option"
/usr/local/lib/mypy/typeshed/third_party/2and3/click/decorators.pyi:174: note: "version_option" defined here`

I *think* you should probably add `  # type: ignore` here, it makes mypy pass - @sergiusens Can you confirm if this the right solution?

gsilvapt:
I can give that a try, for sure. It's awkward that it passed all tests locally :sweat_smile:

snapcraft/tests/unit/commands/test_version.py
```diff
@@ -19,7 +19,15 @@
...
+        result = self.run_command(['version'])
+        self.assertThat(result.exit_code, Equals(0))
+
+    def test_method_return_same_value(self):
+        result1 = self.run_command(['version'])
+        result2 = self.run_command(['--version'])
+        self.assertEqual(result1.output, result2.output)
```
kyrofa: Perfect, thank you!

kyrofa:
It's almost like it's validating against its own bundled (and apparently old) version of Click. @sergiusens could use some guidance here.

gsilvapt:
Oh, nice! Thank you for reviewing this PR all the time, @kyrofa.
Keep me posted about this issue so I know how to fix if something happen next time.

snapcraft/cli/version.py
```diff
@@ -0,0 +1,36 @@
...
+
+@versioncli.command('version')
+def version():
+    """Obtain snapcraft's version number.
+
+    Examples:
+        snapcraft version
+        snapcraft --version
+    """
+    click.echo('snapcraft, version {}'.format(snapcraft.__version__))
```
sergiusens: add a constant here with the version message, make it look like the current on in click, from the source:
```
'%(prog)s, version %(version)s'
```

snapcraft/cli/__init__.py
```diff
@@ -100,7 +99,8 @@ def list_commands(self, ctx):
 
 
 @click.group(cls=SnapcraftGroup, invoke_without_command=True)
-@click.version_option(version=snapcraft.__version__)
+@click.version_option(message='snapcraft, version %(version)s',
```
sergiusens: use the const defined in version

gsilvapt: this command changed to `( = MESSAGE_SNAPCRAFT_VERSION,version = snapcraft.__version__)

snapcraft/cli/version.py
gsilvapt: I had to hardcode "snapcraft, version %(version)" otherwise it would produce issues.

Thank you for your review. Hopefully I did not mess up commit after a long time (had a couple of busy weeks). If you prefer, I can start over. to avoid multiple commits and such.

gsilvapt:
There is unnecessary white space causing the static tests to fail. If I fix this, some conflicts arise in other files. I need guidance here, if you please. I can push a single commit just to fix the white spaces. I wanted to squash this into the previous commit but it is a no-brainer for me.

gsilvapt:
@kyrofa no, I don't mind at all!, although I was waiting for some feedback on how to solve this. I'll have a look at what you don to see if I can understand the fix. Sorry for the trouble and much, much thanks for your help!

PR #1747, opened 2017-11-21
static tests: enable type checking by use of mypy

sergiusens:
Make the initial set of changes for mypy to pass and run it as
part of ./runtests.sh static.

As part of this, `compat` from the storeapi was removed as it was
dead code had mypy issues on almost every line of its implementation.

-----


snapcraft/tests/unit/plugins/python/_basesuite.py
```diff
@@ -13,13 +13,17 @@
...
 
-try:  # pragma: no cover
-    from builtins import open  # noqa
-    from urllib.parse import quote_plus, urljoin
-except ImportError:  # pragma: no cover
-    from __builtin__ import open  # noqa
-    from urllib import quote_plus  # noqa
-    from urlparse import urljoin  # noqa
+
+# LP: #1733584
```
kyrofa: Thank you for this, you saved yourself a question :wink: .

snapcraft/tests/unit/plugins/test_jhbuild.py
```diff
@@ -18,7 +18,8 @@
 import snapcraft.plugins.jhbuild
 
 
-class JHBuildPluginTestCase(snapcraft.tests.unit.TestCase):
+# LP: #1733584
+class JHBuildPluginTestCase(snapcraft.tests.unit.TestCase):  # type: ignore
```
kyrofa: How is this ignore (and all the other test case ones below) related to that bug?

sergiusens: There are a bunch of issues in `mypy` related to using `__init__.py`, one of them is that for this to work you would need to `import snapcraft.tests.unit.__init__` or something like that since it sort of works at the module level and not really at a package level. This is somewhat not urgent as it is discouraged to implement things in `__init__.py` in the python world.

kyrofa: Ah, so the issue here is the fact that `TestCase` is in `snapcraft.tests.unit.__init__.py`? Okay, that makes more sense.

sergiusens:
@kyrofa I already discussed with @elopio 
Given the `__init__` for the `snapcraft.tests.unit` package is a pain, I have just went on and made it a TODO for @elopio :wink:

runtests.sh
```diff
@@ -46,6 +46,7 @@ python3 -m coverage 1>/dev/null 2>&1 && coverage="true"
 run_static_tests(){
     SRC_PATHS="bin snapcraft snaps_tests external_snaps_tests setup.py"
     python3 -m flake8 --max-complexity=10 $SRC_PATHS
+    mypy --ignore-missing-imports --follow-imports=silent -p snapcraft
```
kyrofa: Last question: we go to lengths to check more than just `snapcraft` in the flake8 tests. Is there a reason we're not doing the same for mypy?

sergiusens: I gave it thought and there is minimal to no gain in checking setup.py while the others are in there path to going away so not worth fixing if they have problems.

PR #1754, opened 2017-11-22
nodejs plugin: pass proxy configuration to yarn

cjwatson:
yarn apparently doesn't pick this up from the environment, so we need to
tell it explicitly.

LP: #1733718
Signed-off-by: Colin Watson <cjwatson@ubuntu.com>


elopio:
I retriggered the tests because they failed with a connection problem.

cjwatson:
I think that we're actually running into https://github.com/yarnpkg/yarn/issues/4885, which was closed a couple of days ago, but perhaps isn't fixed in the latest published build yet.

snapcraft/tests/unit/plugins/test_nodejs.py
```diff
@@ -16,8 +16,9 @@
 
 import collections
 import os
-
 from unittest import mock
+
```
sergiusens: thanks for reordering this correctly :-)

PR #1755, opened 2017-11-23
tests: collect the autopkgtest results and save them in git 

elopio:

-----

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,8 +41,10 @@ jobs:
       script: ./tools/travis/run_cla_check.sh
     # Trigger nightly tests, only in cron.
     - stage: nightly
+    # Run spread tests for the snapcraft snap in edge.
       if: type = cron
       script: ./runtests.sh spread
+    # Run autopkgtests in LXD amd64 for the snapcraft-daily PPA
     - if: type = cron
       script: sudo ./tools/travis/run_autopkgtests.sh ubuntu:xenial integrationtests-general
     - if: type = cron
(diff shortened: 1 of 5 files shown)


tools/collect_ppa_autopkgtests_results.py
```diff
@@ -0,0 +1,78 @@
...
+
+    subprocess.check_call([
+        'git', '-C', repo_dir, 'add' , '--all'])
+    subprocess.check_call([
+        'git', '-C', repo_dir, 'commit' , '--message',
+        'Add the results for {}'.format(day)])
+
+    subprocess.check_call([
+        'git', '-C', repo_dir, 'push',
+        'https://{GH_TOKEN}@github.com/elopio/autopkgtest-results.git'.format(
```
kyrofa: Huh, I didn't realize the token was also used as the username.

tools/travis/run_ppa_autopkgtests.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright (C) 2015-2017 Canonical Ltd
+# Copyright (C) 2017 Canonical Ltd
```
kyrofa: Is this change on purpose?

tools/collect_ppa_autopkgtests_results.py
kalikiana: Official docs say it can be used in place of the password. Presumably you can just set it to foo:bar to use it the way it's done here. But perhaps we should document somewhere how we use it?

tools/travis/run_ppa_autopkgtests.sh
elopio: Yes, the 2015 was a copy and paste mistake. This is a file we added this year.

tools/collect_ppa_autopkgtests_results.py
elopio: I'm not sure what are you suggesting to document. This is the common usage pattern for the token in scripts.

.travis.yml
```diff
@@ -63,5 +65,9 @@ jobs:
       script: sudo ./tools/travis/run_autopkgtests.sh ubuntu-daily:bionic integrationtests-plugins-catkin
     - if: type = cron
       script: sudo ./tools/travis/run_autopkgtests.sh ubuntu-daily:bionic integrationtests-snapd
+    # Trigger autopkgtests in all the supported distros and architectures for the snapcraft-daily PPA
     - if: type = cron
-      script: ./tools/travis/run_ppa_autopkgtests.sh
+      script: sudo ./tools/travis/run_ppa_autopkgtests.sh
+    # Collect the results for the PPA autopkgtests that were triggered the day before.
+    - if: type = cron
+      script: sudo ./tools/travis/collect_ppa_autopkgtests_results.sh $(date -d "yesterday" "+%Y%m%d")
```
sergiusens: sometimes tests take more than a day to run on other architectures, we should be a bit more smart here.

PR #1759, opened 2017-11-23
many: ensure classic confined binaries use the correct interpreter

sergiusens:
During the prime step, the pluginhandler collects all elf files,
as a postprocessing step, the elf files that contain an .interp
section in the elf headers are set to use the linker loader
from core.

Fixes: #1662

-----
I have built a snap (but still sets the LD_FLAG) which can be obtained by `snap install snapcraft --classic --channel edge/patchelf`

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -52,6 +52,12 @@ parts:
         LIBSODIUM=$(readlink -n $TRIPLET_PATH/libsodium.so.18)
         ln -s $LIBSODIUM $TRIPLET_PATH/libsodium.so
     after: [python, apt]
+  patchelf:
+    source: https://launchpad.net/ubuntu/+archive/primary/+files/patchelf_0.9.orig.tar.gz
+    source-checksum: sha256/f2aa40a6148cb3b0ca807a1bf836b081793e55ec9e5540a5356d800132be7e0a
+    plugin: autotools
+    stage:
+        - bin/patchelf
   python:
     source: https://www.python.org/ftp/python/3.6.0/Python-3.6.0.tar.xz
     plugin: autotools
--- a/snapcraft/internal/elf.py
+++ b/snapcraft/internal/elf.py
@@ -20,7 +20,7 @@ import logging
 import os
 import subprocess
 import sys
-from typing import List, Set, Sequence, FrozenSet
+from typing import Dict, List, Set, Sequence, FrozenSet
 
 import magic
 
@@ -135,7 +135,8 @@ def get_dependencies(elf: str) -> Set[str]:
     return libs
 
 
-def get_elf_files(root: str, file_list: Sequence[str]) -> FrozenSet[str]:
+def get_elf_files(root: str,
+                  file_list: Sequence[str]) -> Dict[str, Dict[str, str]]:
     """Return a frozenset of elf files from file_list prepended with root.
 
     :param str root: the root directory from where the file_list is generated.
@@ -147,7 +148,7 @@ def get_elf_files(root: str, file_list: Sequence[str]) -> FrozenSet[str]:
     if ms.load() != 0:
         raise RuntimeError('Cannot load magic header detection')
 
-    elf_files = set()
+    elf_files = dict()  # type: Dict[str, Dict[str, str]]
 
     fs_encoding = sys.getfilesystemencoding()
 
@@ -169,6 +170,58 @@ def get_elf_files(root: str, file_list: Sequence[str]) -> FrozenSet[str]:
         # for an ldd call.
         file_m = ms.file(path_b)
         if file_m.startswith('ELF') and 'dynamically linked' in file_m:
-            elf_files.add(path)
+            is_executable = 'interpreter' in file_m
+            elf_files[path] = dict(executable=is_executable)
 
-    return frozenset(elf_files)
+    return elf_files
+
+
+class Patcher:
+    """Patcher holds the necessary logic to patch elf files."""
+
+    def __init__(self, *, dynamic_linker: str) -> None:
+        """Create a Patcher instance.
+
+        :param str dynamic_linker: the path to the dynamic linker to set the
+                                   elf file to.
+        """
+        self._dynamic_linker = dynamic_linker
+
+        # If we are running from the snap we want to use the patchelf
+        # bundled there as it would have the capabilty of working
+        # anywhere given the fixed ld it would have.
+        # If not found, resort to whatever is on the system brought
+        # in by packaging dependencies.
+        if common.is_snap():
+            snap_dir = os.getenv('SNAP')
+            self._patchelf = os.path.join(snap_dir, 'bin', 'patchelf')
+        else:
+            self._patchelf = 'patchelf'
+
+    def patch(self, *, elf_file: str, elf_properties: Dict[str, str]) -> None:
+        """Patch elf_file with the Patcher instance configuration.
+
+        patch will to the right thing with regards to the different type
+        of elf files.
+
+        :param str elf_file: path to the elf_file to patch.
+        :param dict elf_properties:
+            properties of the elf file. The property list is as follows:
+                * executable: boolean stating if elf_file is executable.
+        :raises snapcraft.internal.errors.PatcherError:
+            raised when the elf_file cannot be patched.
+        """
+        # When setting rpath for libraries is implemented, we will do more
+        # here.
+        if not elf_properties['executable']:
+            return
+        try:
+            subprocess.check_call([self._patchelf,
+                                   '--set-interpreter',  self._dynamic_linker,
+                                   elf_file])
+        # There is no need to catch FileNotFoundError as patchelf should be
+        # bundled with snapcraft which means its lack of existence is a
+        # "packager" error.
+        except subprocess.CalledProcessError as call_error:
+            raise errors.PatcherError(elf_file=elf_file,
+                                      message=str(call_error))
--- a/snapcraft/internal/elf.py
+++ b/snapcraft/internal/elf.py
@@ -75,7 +75,7 @@ class Patcher:
     def patch(self, *, elf_file: ElfFile) -> None:
         """Patch elf_file with the Patcher instance configuration.
 
-        patch will to the right thing with regards to the different type
+        patch will to the right thing with regards to the different types
         of elf files.
 
         :param str elf_file: path to the elf_file to patch.
--- a/snapcraft/internal/elf.py
+++ b/snapcraft/internal/elf.py
@@ -20,7 +20,7 @@ import logging
 import os
 import subprocess
 import sys
-from typing import Dict, List, Set, Sequence, FrozenSet
+from typing import FrozenSet, List, Set, Sequence
 
 import magic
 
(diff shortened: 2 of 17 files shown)


sergiusens:
Everything passed except for that one hidden test.

snap/snapcraft.yaml
```diff
@@ -52,6 +52,12 @@ parts:
         LIBSODIUM=$(readlink -n $TRIPLET_PATH/libsodium.so.18)
         ln -s $LIBSODIUM $TRIPLET_PATH/libsodium.so
     after: [python, apt]
+  patchelf:
+    source: https://launchpad.net/ubuntu/+archive/primary/+files/patchelf_0.9.orig.tar.gz
```
elopio: Wouldn't it be better to build the snap with the version that is on xenial, until we release bionic?
Like, keep our snap in sync with the latest LTS, as much as possible.

sergiusens: No, because this patchelf needs to be `patchelf`ed itself, additionally, we need the features of this version everywhere for the future PRs related to rpath and no-default-libs

snapcraft/internal/elf.py
```diff
@@ -169,6 +170,58 @@ def get_elf_files(root: str, file_list: Sequence[str]) -> FrozenSet[str]:
...
+        if common.is_snap():
+            snap_dir = os.getenv('SNAP')
+            self._patchelf = os.path.join(snap_dir, 'bin', 'patchelf')
+        else:
+            self._patchelf = 'patchelf'
+
+    def patch(self, *, elf_file: str, elf_properties: Dict[str, str]) -> None:
+        """Patch elf_file with the Patcher instance configuration.
+
+        patch will to the right thing with regards to the different type
```
elopio: typo: to/do

elopio: typo: type/types

snapcraft/internal/elf.py
```diff
@@ -169,6 +170,58 @@ def get_elf_files(root: str, file_list: Sequence[str]) -> FrozenSet[str]:
...
+        self._dynamic_linker = dynamic_linker
+
+        # If we are running from the snap we want to use the patchelf
+        # bundled there as it would have the capabilty of working
+        # anywhere given the fixed ld it would have.
+        # If not found, resort to whatever is on the system brought
+        # in by packaging dependencies.
+        if common.is_snap():
+            snap_dir = os.getenv('SNAP')
+            self._patchelf = os.path.join(snap_dir, 'bin', 'patchelf')
```
elopio: a nit, this would be clearer as _patchelf_command.

snapcraft/internal/elf.py
```diff
@@ -169,6 +170,58 @@ def get_elf_files(root: str, file_list: Sequence[str]) -> FrozenSet[str]:
...
+        # here.
+        if not elf_properties['executable']:
+            return
+        try:
+            subprocess.check_call([self._patchelf,
+                                   '--set-interpreter',  self._dynamic_linker,
+                                   elf_file])
+        # There is no need to catch FileNotFoundError as patchelf should be
+        # bundled with snapcraft which means its lack of existence is a
+        # "packager" error.
```
elopio: Thanks for the note.

snapcraft/internal/elf.py
sergiusens: ok, `_patchelf_cmd`; we can use the full word command when we move away from the 79 char limit :wink:

elopio:
This looks very clear to me. From what I've been learning about elf, it also looks correct. And the nice integration tests give me a lot of confidence, nice work with that.

But, I will leave the +1 to Kyle. I'm not near to be an expert in elf.

snapcraft/internal/elf.py
elopio: a decent compromise :)

sergiusens:
that failing test is killing me, it works fine locally.

snapcraft/internal/elf.py
```diff
@@ -147,7 +148,7 @@ def get_elf_files(root: str, file_list: Sequence[str]) -> FrozenSet[str]:
     if ms.load() != 0:
         raise RuntimeError('Cannot load magic header detection')
 
-    elf_files = set()
+    elf_files = dict()  # type: Dict[str, Dict[str, str]]
```
kyrofa: As this functionality grows, I wonder if it would make more sense to make the `Dict[str, str]` part of this a class instead.

snapcraft/internal/elf.py
```diff
@@ -169,6 +170,58 @@ def get_elf_files(root: str, file_list: Sequence[str]) -> FrozenSet[str]:
...
+    def patch(self, *, elf_file: str, elf_properties: Dict[str, str]) -> None:
+        """Patch elf_file with the Patcher instance configuration.
+
+        patch will to the right thing with regards to the different type
+        of elf files.
+
+        :param str elf_file: path to the elf_file to patch.
+        :param dict elf_properties:
+            properties of the elf file. The property list is as follows:
+                * executable: boolean stating if elf_file is executable.
```
kyrofa: Yeah if we're making it part of the API, I doubly think this should be a class. It'll make the API much less error-prone.

snapcraft/internal/elf.py
```diff
@@ -169,6 +170,58 @@ def get_elf_files(root: str, file_list: Sequence[str]) -> FrozenSet[str]:
...
+        :raises snapcraft.internal.errors.PatcherError:
+            raised when the elf_file cannot be patched.
+        """
+        # When setting rpath for libraries is implemented, we will do more
+        # here.
```
kyrofa: Thanks for the comment!

snapcraft/internal/elf.py
kyrofa: Although that ignores the possibility of running from source and not having it installed.

snapcraft/internal/errors.py
```diff
@@ -355,3 +355,13 @@ class InvalidContainerImageInfoError(SnapcraftError):
 
     def __init__(self, image_info):
         super().__init__(image_info=image_info)
+
+
+class PatcherError(SnapcraftError):
+
+    fmt = (
+        '{elf_file!r} cannot be patched to function properly as a classic '
```
kyrofa: s/as/in/ (the file is not the snap, it's just part of one). However, this error can actually be multiple errors, and has no "how to fix" side of it. Do you have an idea of how patchelf might fail, and how one might fix the issue? Perhaps we can break this error class up a little.

snapcraft/internal/lifecycle/_runner.py
```diff
@@ -209,7 +209,9 @@ def _run_step(self, step, part, part_names):
 
         part = _replace_in_part(part)
 
-        getattr(part, step)()
+        is_classic = self.config.data['confinement'] == 'classic'  # type: bool
+
+        getattr(part, step)(is_classic=is_classic)
```
kyrofa: For some reason adding kwargs to all steps just to handle `is_classic` in `prime()` is bothering me. It's not something about the snap that will change, it's a fundamental property of it. It seems like something that would just be generally known, like its name. Would it be weird to put this in the `ProjectOptions` instead, or otherwise make it generally available? Maybe even the `PluginHandler`'s `__init__`.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -470,6 +469,16 @@ def prime(self, force=False):
                 _migrate_files(system, system_dependency_paths, '/',
                                self.primedir, follow_symlinks=True)
 
+        # If we are classic we need to patch the elf files for which
+        # prime will be called with instructions to do so
```
kyrofa: I don't quite understand this sentence. "elf files for which prime will be called" ? We're _in_ prime.

snap/snapcraft.yaml
kyrofa: Isn't the whole point of this PR that patchelf would still be patchelf'd itself, even if it was a stage-package? More importantly, though, the idea that we'll be relying on v0.9 features when the deb can only ever use v0.8 seems somewhat concerning unless we're quickly moving toward not having a deb at all.

snapcraft/internal/elf.py
sergiusens: the last `str` of that is probably going to morph into an `Any`. I don't mind making it a class, and given your comment below I might as well.

snapcraft/internal/elf.py
sergiusens: I would want to discourage people who wouldn't read `HACKING.md` to run from source :wink:

sergiusens: they will get an exception any developer would know how to resolve.

snapcraft/internal/errors.py
sergiusens: If an error occurs here, I would have little to no idea on how this would be fixed.

sergiusens: Also, I am pro for every error message to be reformatted as part of the exception class, but I do not think we need one exception per type of error, unless it is really useful.

snapcraft/internal/lifecycle/_runner.py
sergiusens: ok, I'll go down that path (`project_options`)

snapcraft/internal/pluginhandler/__init__.py
sergiusens: meaning, this `prime` was called with instructions to patch; no matters anymore as I'll be moving the parameter to the class/object.

snap/snapcraft.yaml
sergiusens: not in the first release where it is built from source with the linker flag that it is removed. This is a bootstrapping problem.

We also *DO* need features from 0.9 and I'll be moving to using the snap on builders almost effectively making th deb useless (it already runs behind by a lot given our iterations).

That said, I am making patchelf available as a snap the deb would need to install for everything to work.

sergiusens: Alternatively, we could vendor patchelf into snapcraft for the deb.

snapcraft/internal/lifecycle/_runner.py
sergiusens: I ended up adding it to the PluginHandler

snap/snapcraft.yaml
kyrofa: Yeah your call, just something I wanted to point out.

snapcraft/internal/elf.py
```diff
@@ -35,6 +35,72 @@
...
+        if common.is_snap():
+            snap_dir = os.getenv('SNAP')
+            self._patchelf_cmd = os.path.join(snap_dir, 'bin', 'patchelf')
+        else:
+            self._patchelf_cmd = 'patchelf'
+
+    def patch(self, *, elf_file: ElfFile) -> None:
+        """Patch elf_file with the Patcher instance configuration.
+
+        patch will to the right thing with regards to the different types
```
kyrofa: s/to/do/, although I must say this isn't the most helpful comment :stuck_out_tongue: . Perhaps something like:

"If the ELF is executable, patch it to use the configured linker."

snapcraft/internal/elf.py
kyrofa: Loving the new class, though this bit of the docstring needs to be updated to reflect.

snapcraft/internal/errors.py
kyrofa: It's only useful if the fix for the issue changes depending on the error. However, it sounds like we genuinely don't expect an error here and have no idea how it would be fixed, so this seems fine.

kyrofa: (although the typo remains)

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -470,6 +470,12 @@ def prime(self, force=False):
                 _migrate_files(system, system_dependency_paths, '/',
                                self.primedir, follow_symlinks=True)
 
+        if self._confinement == 'classic':
```
kyrofa: Yeah I like this better, thank you!

snapcraft/tests/integration/plugins/test_rust_plugin.py
```diff
@@ -129,8 +129,7 @@ def test_prime(self):
 
         self.run_snapcraft('prime')
 
-        bin_path = os.path.join(self.parts_dir, 'rust-hello', 'install',
-                                'bin', 'rust-hello')
+        bin_path = os.path.join('prime', 'bin', 'rust-hello')
```
kyrofa: Hahahahaha, we all missed it!

PR #1760, opened 2017-11-24
tests: move the catkin integration tests to a separate suite

elopio:

-----
Applying the results of the discussion with @kyrofa earlier this week. This is not a nice solution, but none of the alternatives were nice either. We have some grand ideas to make the suite dynamic, but that's very complex.

I think that we will start hitting the timeouts often in the remaining plugins suite, so we probably will have to do the same for the python or ruby tests.


.travis.yml
```diff
@@ -28,6 +28,8 @@ jobs:
       script: sudo ./tools/travis/run_tests.sh snapcraft/tests/integration/general
     - if: type != cron
       script: sudo ./tools/travis/run_tests.sh snapcraft/tests/integration/plugins
+    - if: type != cron
+      script: sudo ./tools/travis/run_tests.sh snapcraft/tests/integration/plugins/catkin
```
kyrofa: I'm afraid specifying `-t` to discover means this must be importable, which of course it is not (see the Travis error).

Using -t seems unnecessary, things work fine without it, and then our plan actually works. I'm going to push that change up and see how things run.

debian/tests/integrationtests
```diff
@@ -16,5 +16,6 @@ fi
 
 mkdir --parents /home/ubuntu/autopkgtest_tmp --mode 777
 for suite in $suites; do
+    echo $suite
```
kyrofa: So nice! Thank you for this.

PR #1764, opened 2017-11-24
snapcraft.yaml: use gcc to determine tuple

kyrofa:

-----

Currently, when building the Snapcraft snap, we try to create the libsodium.so symlink by manually constructing the arch tuple using `uname`. However, this does not work correctly in all cases, for example i686.

This PR fixes LP: [#1733922](https://bugs.launchpad.net/snapcraft/+bug/1733922) by using `gcc -print-multiarch` instead. It's exactly what we need, and it's already a `build-package`.

Please test this PR by simply smoke-testing this snap on i386 (even an i386 LXD container would work):

    $ sudo snap install snapcraft --channel=stable/pr-1764 --classic

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -46,7 +46,7 @@ parts:
         - '*'
         - '**/*.pyc'
     install: |
-        TRIPLET_PATH=$(ls -d $SNAPCRAFT_PART_INSTALL/usr/lib/$(uname -p)-*)
+        TRIPLET_PATH="$SNAPCRAFT_PART_INSTALL/usr/lib/$(gcc -print-multiarch)"
         LIBARCHIVE=$(readlink -n $TRIPLET_PATH/libarchive.so.13)
         ln -s $LIBARCHIVE $TRIPLET_PATH/libarchive.so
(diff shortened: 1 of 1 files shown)


sergiusens:
this looks fine, but I wonder if you want to take this one step further and just make sure `SNAPCRAFT_ARCH_TRIPLET` is available in the environment where this runs and optionally maybe remove the static map of archtriplets under `snapcraft._options` and use gcc. This latter part depends on adding a strict build-packages incantation in the snapcraft code base.

You would be fixing @elopio's task this way.

sergiusens:
That would be #1752

kyrofa:
Sure, I'll take a look. As we discussed on IRC, instead of adding `gcc` as a build-package everywhere, I'll continue using the map we already have.

kyrofa:
Yeah your addition looks good. I apparently can't approve my own PR, but +1. Let's make sure to merge with the correct commit title etc. (i.e. this no longer involves gcc).

kyrofa:
Are you in a position to update the snap in that temporary branch? Just to double-check that this works on i386.

sergiusens:
```
                 edge/pr-1764    0+git.da47019             854         2017-12-28T21:03:47.731989
The 'edge/pr-1764' channel is now open.
```

PR #1766, opened 2017-11-28
Enhanced fake store to detect that metadata can't be pushed prior to pushing the binary, and fixed the test

facundobatista:
This is because the real store needs for the binary of snap to be first pushed before really the metadata can be sent.

This was already understood by snapcraft (the error message when that doesn't happen express the problem correctly) but one integration test was incorrectly built (see [this failure](https://travis-ci.org/snapcore/snapcraft/jobs/308248189)).

So I fixed the fake store to make that bad test fail, and then fixed the test.

### DIFF
--- a/snapcraft/tests/integration/store/test_store_push_metadata.py
+++ b/snapcraft/tests/integration/store/test_store_push_metadata.py
@@ -31,7 +31,7 @@ class PushMetadataTestCase(integration.StoreTestCase):
 
         error = self.assertRaises(
             subprocess.CalledProcessError,
-            self.run_snapcraft, ['push', snap_file_path])
+            self.run_snapcraft, ['push-metadata', snap_file_path])
         self.assertIn('No valid credentials found. Have you run "snapcraft '
                       'login"?', str(error.output))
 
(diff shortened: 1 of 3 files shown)


snapcraft/tests/integration/store/test_store_push_metadata.py
```diff
@@ -54,6 +54,9 @@ def test_with_login(self):
         # Push the snap
         snap_file_path = '{}_{}_{}.snap'.format(name, version, 'all')
         self.assertThat(os.path.join(snap_file_path), FileExists())
+        output = self.run_snapcraft(['push', snap_file_path])
```
sergiusens: Thanks for looking into this! Can we add a test which looks like the original but fails as no snap has been pushed before?

facundobatista: Yes!

facundobatista:
Added a test for the case of push-metadata without push, and a little refactor so it's easier to build the snaps for the different tests.

sergiusens:
the unit tests are now failing

PR #1769, opened 2017-11-29
lxd: add an --image argument to cleanbuild

mwhudson:
A replacement for #1557

### DIFF
--- a/snapcraft/cli/env.py
+++ b/snapcraft/cli/env.py
@@ -15,7 +15,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 import os
 from distutils import util
-from snapcraft.internal import errors, lxd
+from snapcraft.internal import common, errors, lxd
 
 
 class ContainerConfig:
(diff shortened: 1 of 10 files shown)


snapcraft/cli/env.py
```diff
@@ -50,6 +50,11 @@ def use_container(self):
     def remote(self):
         return self._remote
 
+    @property
+    def image(self):
+        # TODO: allow this to be overridden
```
kalikiana: It slightly bugs me that now there's two places where the default image is defined... how about putting the default image in snapcraft/internal/common.py?

kalikiana:
I proposed  #1787 to fix the refresh test.

snapcraft/cli/env.py
mwhudson: Done

mwhudson:
I updated the refresh command now that the test for it fails :)

mwhudson:
Any update? I opened #1557 nearly four months ago now :-)

sergiusens:
Moving forward, we have decided to move away from containers as the mechanism to execute builds https://forum.snapcraft.io/t/draft-consistent-build-environments-build-vms/5374/3

Sorry for the late reply and the wasted effort.

PR #1771, opened 2017-11-29
lxd: suppress traceback when lxc launch / init fails

mwhudson:
lxc itself prints a useful enough error message when this happens.

### DIFF
--- a/snapcraft/internal/lxd/_cleanbuilder.py
+++ b/snapcraft/internal/lxd/_cleanbuilder.py
@@ -21,7 +21,7 @@ import petname
 import subprocess
 
 from ._containerbuild import Containerbuild
-from snapcraft.internal.errors import ContainerError
+from snapcraft.internal.errors import ContainerConnectionError
 
 logger = logging.getLogger(__name__)
 
@@ -40,7 +40,7 @@ class Cleanbuilder(Containerbuild):
             subprocess.check_call([
                 'lxc', 'launch', '-e', self._image, self._container_name])
         except subprocess.CalledProcessError as e:
-            raise ContainerError(e)
+            raise ContainerConnectionError('Failed to setup container')
         self._configure_container()
         self._wait_for_network()
         self._container_run(['apt-get', 'update'])
--- a/snapcraft/internal/lxd/_cleanbuilder.py
+++ b/snapcraft/internal/lxd/_cleanbuilder.py
@@ -21,6 +21,7 @@ import petname
 import subprocess
 
 from ._containerbuild import Containerbuild
+from snapcraft.internal.errors import ContainerError
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


snapcraft/internal/lxd/_cleanbuilder.py
```diff
@@ -35,8 +36,11 @@ def __init__(self, *, output=None, source, project_options,
                          container_name=container_name, remote=remote)
 
     def _ensure_container(self):
-        subprocess.check_call([
-            'lxc', 'launch', '-e', self._image, self._container_name])
+        try:
+            subprocess.check_call([
+                'lxc', 'launch', '-e', self._image, self._container_name])
+        except subprocess.CalledProcessError as e:
+            raise ContainerError(e)
```
sergiusens: you want this to be `ContainerSnapcraftCmdError` which is more correct and silences the exception printing of `cmd`

snapcraft/internal/lxd/_project.py
```diff
@@ -46,8 +47,11 @@ def __init__(self, *, output, source, project_options,
     def _ensure_container(self):
         new_container = not self._get_container_status()
         if new_container:
-            subprocess.check_call([
-                'lxc', 'init', self._image, self._container_name])
+            try:
+                subprocess.check_call([
+                    'lxc', 'init', self._image, self._container_name])
+            except subprocess.CalledProcessError as e:
+                raise ContainerError(e)
```
sergiusens: ditto

snapcraft/internal/lxd/_cleanbuilder.py
mwhudson: That can't be the right exception, it's message is "Snapcraft command failed in the container:". Do you mean ContainerRunError? (I find over-elaborate exception hierarchies a bit of an anti-pattern myself...)

kalikiana: In my similar PR #1761 I used `ContainerConnectionError('Failed to setup container')` because it will refer the user to online docs for LXD, which I think is helpful when you get something like "No storage pool found. Please create a new storage pool." and don't know what that means. So personally I think we want that message here.

What @sergiusens is suggesting is to use `ContainerSnapcraftCmdError` because it won't add anything on top of what the `lxc` command may have printed. I also think that's the wrong error because it's specifically referring to running Snapcraft inside the container - that's not the case here.
If we wanted to go that route I think it's time we moved `should_print_error` into the `SnapcraftError` class to be used with a new exception here.

mwhudson: I don't want to be a jerk but at this stage given how simple the change is, maybe it would be easier for one of you to JFDI this?

kalikiana: Done

PR #1774, opened 2017-11-29
storeapi: support for pushing binary metadata

matiasb:
Update push-metadata command to also push binary metadata (so far, icon only).


snapcraft/_store.py
```diff
@@ -64,6 +64,27 @@ def _get_data_from_snap_file(snap_path):
...
 
+@contextmanager
+def _get_icon_from_snap_file(snap_path):
+    icon_file = None
+    with tempfile.TemporaryDirectory() as temp_dir:
+        output = subprocess.check_output(
+            ['unsquashfs', '-d',
+             os.path.join(temp_dir, 'squashfs-root'),
+             snap_path, '-e', 'meta/gui'])
+        logger.debug(output)
```
kyrofa: I suspect this debug output will not be very useful on its own. Would it be worth adding some context to it?

snapcraft/_store.py
```diff
@@ -404,6 +425,9 @@ def push_metadata(snap_filename, force):
     with _requires_login():
         store.push_precheck(snap_name)
         store.push_metadata(snap_name, metadata, force)
+        with _get_icon_from_snap_file(snap_filename) as icon:
+            metadata = {'icon': icon}
+            store.push_binary_metadata(snap_name, metadata, force)
```
kyrofa: It's possible for the `icon` to be `None`, here. I can't tell from the below: will that remove the icon from the store? That situation does not seem covered by tests.

snapcraft/_store.py
matiasb: Yeah, makes sense, let me fix that.

snapcraft/_store.py
matiasb: Yes, `icon` could be None, and that will remove the icon from the store.
I'll review tests and add it if missing. Thanks!

kyrofa:
Thanks for the additional test, @matiasb. I've also tested this against the staging store with success. However, is there no way to selectively update metadata without updating it all in one go? What if I only wanted to update the icon, but not the other metadata?

matiasb:
The implementation is the result of the discussion in https://forum.snapcraft.io/t/mechanisms-for-converging-store-and-snap-metadata/2200 (that summarizes the discussions during the rally too).
It would be simple to split text and binary metadata in different commands.
Note that, in any case, the metadata being pushed is extracted from a given revision, so not sure if you would want to push an icon from a revision and the description from another, for example (ie. you cannot push a specific value for a particular field, independent of a revision, as things were discussed at least -the store API would allow it, though-).

manual-tests.md
```diff
@@ -231,3 +231,19 @@
 
     * Check that it should end ok
     * Check in the Web that the description is now what the YAML says
+
+
+# Test push binary metadata with conflicts
+
+1. 'snapcraft build' a simple snap
```
sergiusens: this should be `snapcraft snap`, `build` is an intermediate command

manual-tests.md
```diff
@@ -231,3 +231,19 @@
...
+2. Do a simple 'snapcraft push SNAP'
+3. Go to the Web and change snap's icon
+4. Change the snap's icon in the YAML file to something different than you put in the Web
+5. Try to update snap's metadata using `snapcraft push-metadata SNAP`
```
sergiusens: Does pushing the snap also raise the conflict?

sergiusens: Can we add a test for that?

snapcraft/_store.py
```diff
@@ -64,6 +64,27 @@ def _get_data_from_snap_file(snap_path):
...
+        logger.debug("Output extracting icon from snap: %s", output)
+        for extension in ('png', 'svg'):
+            icon_name = 'icon.{}'.format(extension)
+            icon_path = os.path.join(
+                temp_dir, 'squashfs-root', 'meta/gui', icon_name)
+            if os.path.exists(icon_path):
+                icon_file = open(icon_path, 'rb')
+                break
+        yield icon_file
+        if icon_file is not None:
```
sergiusens: should the yield above be wrapped in between a `try` and `finally` to always check and close the icon handle?

snapcraft/storeapi/__init__.py
```diff
@@ -638,6 +658,64 @@ def push_metadata(self, snap_id, snap_name, metadata, force):
...
+        binary_metadata = response.json()
+        # current icons and screenshots
+        icons = [media for media in binary_metadata
+                 if media.get('type') == 'icon']
+        screenshots = [media for media in binary_metadata
+                       if media.get('type') == 'screenshot']
+
+        files = {}
+        # only icon support atm
+        icon = metadata.get('icon')
```
sergiusens: this is going to end up being a rather big function, how about a private module inside `storeapi` to handle this with some OO? Nothing complex

what do you think?

manual-tests.md
matiasb: Right, fixing here and the previous one.

manual-tests.md
matiasb: Push won't raise the conflict, and it won't update the metadata either. Metadata from a pushed revision will only be used if there are no previous changes made in the web UI.
Should we think about adding some warning/message while processing a push about conflicting metadata?

snapcraft/_store.py
matiasb: Yes, updating.

snapcraft/storeapi/__init__.py
matiasb: Let me take a look, will try to refactor it.

manual-tests.md
sergiusens: We shouldn't block on this for the PR, but an error code saying the metadata is not being updated due to divergences for which we can on the client instruct the user would be good. Being silent by default is fine.

sergiusens:
works great
we do need to polish all that red in those errors.

PR #1778, opened 2017-11-30
Add sentences to error when registering name

daniellim2000:
This commit will add 2 more sentences to error when trying to register
a reserved name.

Addresses bug at [LP: 1602791](https://pad.lv/1602791).

### DIFF
--- a/snapcraft/cli/store.py
+++ b/snapcraft/cli/store.py
@@ -32,8 +32,8 @@ _MESSAGE_REGISTER_PRIVATE = dedent("""\
     then we suggest you prefix the name with your developer identity,
     As ‘nessita-yoyodyne-www-site-content’.""")
 _MESSAGE_REGISTER_CONFIRM = dedent("""
-    We always want to ensure that users get the software they expect for a
-    particular name.
+    We always want to ensure that users get the software they expect
+     for a particular name.
 
     If needed, we will rename snaps to ensure that a particular name
     reflects the software most widely expected by our community.
--- a/snapcraft/cli/store.py
+++ b/snapcraft/cli/store.py
@@ -33,7 +33,7 @@ _MESSAGE_REGISTER_PRIVATE = dedent("""\
     As ‘nessita-yoyodyne-www-site-content’.""")
 _MESSAGE_REGISTER_CONFIRM = dedent("""
     We always want to ensure that users get the software they expect
-     for a particular name.
+    for a particular name.
 
     If needed, we will rename snaps to ensure that a particular name
     reflects the software most widely expected by our community.
--- a/snapcraft/cli/store.py
+++ b/snapcraft/cli/store.py
@@ -32,8 +32,8 @@ _MESSAGE_REGISTER_PRIVATE = dedent("""\
     then we suggest you prefix the name with your developer identity,
     As ‘nessita-yoyodyne-www-site-content’.""")
 _MESSAGE_REGISTER_CONFIRM = dedent("""
-    We always want to ensure that users get the software they expect
-    for a particular name.
+    On behalf of the snapcraft store, we always want to ensure that
+    users get the software they expect for a particular name.
 
     If needed, we will rename snaps to ensure that a particular name
     reflects the software most widely expected by our community.
--- a/snapcraft/cli/store.py
+++ b/snapcraft/cli/store.py
@@ -32,8 +32,8 @@ _MESSAGE_REGISTER_PRIVATE = dedent("""\
     then we suggest you prefix the name with your developer identity,
     As ‘nessita-yoyodyne-www-site-content’.""")
 _MESSAGE_REGISTER_CONFIRM = dedent("""
-    On behalf of the snapcraft store, we always want to ensure that
-    users get the software they expect for a particular name.
+    We always want to ensure that users get the software they expect for a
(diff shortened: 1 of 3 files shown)


snapcraft/storeapi/errors.py
```diff
@@ -177,7 +177,10 @@ class StoreRegistrationError(StoreError):
     __FMT_RESERVED = (
         'The name {snap_name!r} is reserved.\n\n'
         'If you are the publisher most users expect for '
-        '{snap_name!r} then please claim the name at {register_name_url!r}')
+        '{snap_name!r} then please claim the name at {register_name_url!r}\n\n'
+        'Please register a name that people will associate with you. '
+        'We can rename your snap later if needed.\n\n'
+        'You can push your own versions of firefox, as firefox@nessita.')
```
sergiusens: I am interested in where the guidance to write the error this way `'You can push your own versions of firefox, as firefox@nessita.'` came from as I am pretty sure `<snap-name>@<user>` is not a supported form of pushing and a name would need to be registered.

I would use the message from `snapcraft register` as inspiration for this one.

daniellim2000: It's in the comments by elopio it seems. I was confused as well.

daniellim2000: Should I remove it or modify it?

matthewpaulthomas: I would remove it, because it’s needlessly repeating something the user just read before they typed the name (“They would also expect to be able to get other snaps of Thunderbird as ‘thunderbird-nessita’).

If it was retained, two things would need fixing:
- Finding out which punctuation is correct, “-” or “@”
- Changing “nessita” to your actual store account ID.

matthewpaulthomas:
Apart from the last line, this change accurately follows the “Snapcraft UX” document.

Unfortunately I think the text given in that document is misguided, for two reasons:
- “A name that people will associate with you” makes sense only for software that is already well-known before it ever becomes a snap — not for software that is released as a snap from the beginning.
- “We” is needlessly mysterious.

I reported these last month as [bug 1723726](https://launchpad.net/bugs/1723726). I’d prefer those to be fixed rather than proliferated.

daniellim2000:
Hi @matthewpaulthomas, any suggestions on what to replace those text with?

sergiusens:
@matthewpaulthomas what text do you suggest?

daniellim2000:
```
error: Failed container creation:
 - https://cloud-images.ubuntu.com/releases: websocket: close 1006 (abnormal closure): unexpected EOF
```

This was from Travis CI, weird should not be my fault right?

sergiusens:
The travis error was not your fault, don't worry about that.

matthewpaulthomas:
@daniellim2000 @sergiusens I suggest “Otherwise, choose a different name.” (That would also make it clear that it’s an _alternative_ to claiming the name, not something to do in addition.)

daniellim2000:
@matthewpaulthomas 
Are you referring to the last sentence, `Please register a name that people will associate with you. We can rename your snap later if needed.\n\n`?

Also, isn't it rather clear that 'we' refers to the snapcraft store, and 'our community' refers to the snapcraft users? What do you think about replacing them with 'we'-->'the snapcraft store' and 'our community'-->'the snapcraft community'.

snapcraft/storeapi/errors.py
```diff
@@ -177,7 +177,9 @@ class StoreRegistrationError(StoreError):
     __FMT_RESERVED = (
         'The name {snap_name!r} is reserved.\n\n'
         'If you are the publisher most users expect for '
-        '{snap_name!r} then please claim the name at {register_name_url!r}')
+        '{snap_name!r} then please claim the name at {register_name_url!r}\n\n'
+        'Otherwise, please register a name that people will associate with '
```
sergiusens: To clear out all ambigous language let's just remove the bits that cause contention.

So changing 
```python
'Otherwise, please register a name that people will associate with '
'you. We can rename your snap later if needed.')
```
To just
```python
'Otherwise, please register another name.')
```

should do the trick.

snapcraft/cli/store.py
```diff
@@ -32,8 +32,8 @@
     then we suggest you prefix the name with your developer identity,
     As ‘nessita-yoyodyne-www-site-content’.""")
 _MESSAGE_REGISTER_CONFIRM = dedent("""
-    We always want to ensure that users get the software they expect
-    for a particular name.
+    On behalf of the snapcraft store, we always want to ensure that
```
sergiusens: can you please revert this? You will have to update many more things and it is leaving the scope of the bug.

While I appreciate the effort, this change will just make this PR take longer to land :wink:

daniellim2000: Okay sure

daniellim2000: Regarding, "You will have to update many more things", may I ask why?

daniellim2000:
Sorry about the mistake.

sergiusens:
I am skipping the CLA check error as the contribution is not sizeable to warrant one, but I would encourage you to sign one to make contributions more straightforward.

daniellim2000:
I did sign before I made the push though. Perhaps it takes some time to update?

PR #1779, opened 2017-11-30
catkin-tools plugin: use stage-packages

kyrofa:

-----

Currently the catkin-tools plugin uses `build-packages` for its build dependency instead of `stage-packages`, which fails on clean systems.

Use `stage-packages` instead. Also, since this would have been caught by an integration test, add one of those as well (to the snapd suite, specifically).

### DIFF
--- a/snapcraft/plugins/catkin_tools.py
+++ b/snapcraft/plugins/catkin_tools.py
@@ -38,7 +38,7 @@ logger = logging.getLogger(__name__)
 class CatkinToolsPlugin(snapcraft.plugins.catkin.CatkinPlugin):
     def __init__(self, name, options, project):
         super().__init__(name, options, project)
-        self.build_packages.extend(['python-catkin-tools'])
+        self.stage_packages.append('python-catkin-tools')
 
         # Beta Warning
         # Remove this comment and warning once catkin tools plugin is stable.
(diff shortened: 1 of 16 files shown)


kyrofa:
(running autopkgtest because that's the only way the slow integration tests will run.)

snapcraft/plugins/catkin_tools.py
```diff
@@ -38,7 +38,7 @@
 class CatkinToolsPlugin(snapcraft.plugins.catkin.CatkinPlugin):
     def __init__(self, name, options, project):
         super().__init__(name, options, project)
-        self.build_packages.extend(['python-catkin-tools'])
+        self.stage_packages.append('python-catkin-tools')
```
cratliff: I'm a bit confused by the need for this.

It works and runs on my system, but conceptually it doesn't seem like stage packages is the correct place for python-catkin-tools to be added.  python-catkin-tools is added as a build package so if the tool is not installed when snapcraft is run it will install for the plugin to be able to build the part.  By adding it to stage_packages won't any catkin-tools part ship the catkin tools package in the snap?  Shipping the build tool seems unnecessary.

I know there was a build issue without this, but I'm wondering if there might be a deeper issue if you're unable to build when the package is specified as a build packages.

cratliff: That makes sense, I figured I was missing a detail.

kyrofa:
> I think all the changes look good.

Thank you for taking a look!

Issue #1663, opened 2017-11-06
patchelf with common rpath (not runpath)


sergiusens:
Estimate: 2 days

PR #1781, opened 2017-12-01
many: set rpath for elf files for classic

sergiusens:
When working with classic confinement `LD_LIBRARY_PATH`
cannot be set without consequences on when other assets on the host are
called from the snap.

This sets an rpath for every elf file when confinement is set to classic
and allows for further extending this mechanism to non classic.

When setting rpaths it also makes use of `$ORIGIN` to setup relative paths.

Fixes: #1663

-----

### DIFF
--- a/snapcraft/internal/elf.py
+++ b/snapcraft/internal/elf.py
@@ -105,7 +105,7 @@ class Patcher:
         self._base_rpaths = base_rpaths
 
         # If we are running from the snap we want to use the patchelf
-        # bundled there as it would have the capabilty of working
+        # bundled there as it would have the capability of working
         # anywhere given the fixed ld it would have.
         # If not found, resort to whatever is on the system brought
         # in by packaging dependencies.
@@ -136,6 +136,7 @@ class Patcher:
 
     def _patch_rpath(self, elf_file: ElfFile) -> None:
         rpath = self._get_rpath(elf_file)
+        print(elf_file.path, rpath)
         # Parameters:
         # --force-rpath: use RPATH instead of RUNPATH.
         # --shrink-rpath: will remove unneeded entries, with the
@@ -165,7 +166,10 @@ class Patcher:
         origin_paths = ':'.join((r for r in set(rpaths) if r))
         base_rpaths = ':'.join(self._base_rpaths)
 
-        return '{}:{}'.format(origin_paths, base_rpaths)
+        if origin_paths:
+            return '{}:{}'.format(origin_paths, base_rpaths)
+        else:
+            return base_rpaths
 
 
 def determine_ld_library_path(root: str) -> List[str]:
@@ -177,7 +181,7 @@ def determine_ld_library_path(root: str) -> List[str]:
 
     :param root str: the root directory to search for specific ld.so.conf
                      entries.
-    :returns: a list of strings of library paths where releavant libraries
+    :returns: a list of strings of library paths where relevant libraries
               can be found within root.
     """
     # If more ld.so.conf files need to be supported, add them here.
--- a/snapcraft/internal/elf.py
+++ b/snapcraft/internal/elf.py
@@ -38,16 +38,14 @@ logger = logging.getLogger(__name__)
 class ElfFile:
     """ElfFile represents and elf file on a path and its attributes."""
 
-    def __init__(self, *, path: str, is_executable: bool=True) -> None:
+    def __init__(self, *, path: str, magic: str) -> None:
         """Initialize an ElfFile instance.
 
         :param str path: path to an elf_file within a snapcraft project.
-        :param bool is_executable: True if the elf_file is an executable,
-                                   meaning that it has in .interp entry
-                                   in its headers.
+        :param str magic: the magic string for path.
         """
         self.path = path
-        self.is_executable = is_executable
+        self.is_executable = 'interpreter' in magic
         self.dependencies = set()  # type: Set[str]
 
     def load_dependencies(self) -> Set[str]:
@@ -119,6 +117,7 @@ class Patcher:
         """Patch elf_file with the Patcher instance configuration.
 
         If the ELF is executable, patch it to use the configured linker.
+        If the ELF has dependencies, set an rpath to them.
 
         :param ElfFile elf: a data object representing an elf file and its
                             relevant attributes.
@@ -136,7 +135,6 @@ class Patcher:
 
     def _patch_rpath(self, elf_file: ElfFile) -> None:
         rpath = self._get_rpath(elf_file)
-        print(elf_file.path, rpath)
         # Parameters:
         # --force-rpath: use RPATH instead of RUNPATH.
         # --shrink-rpath: will remove unneeded entries, with the
@@ -276,7 +274,6 @@ def get_elf_files(root: str,
         # for an ldd call.
         file_m = ms.file(path_b)
         if file_m.startswith('ELF') and 'dynamically linked' in file_m:
-            is_executable = 'interpreter' in file_m
-            elf_files.add(ElfFile(path=path, is_executable=is_executable))
+            elf_files.add(ElfFile(path=path, magic=file_m))
 
     return frozenset(elf_files)
--- a/snapcraft/internal/elf.py
+++ b/snapcraft/internal/elf.py
@@ -139,7 +139,7 @@ class Patcher:
         # Parameters:
         # --force-rpath: use RPATH instead of RUNPATH.
         # --shrink-rpath: will remove unneeded entries, with the
-        #                 side effect of prefering host libraries
+        #                 side effect of preferring host libraries
         #                 so we simply do not use it.
         # --set-rpath: set the RPATH to the colon separated argument.
         self._run_patchelf(['--force-rpath',
--- a/snapcraft/internal/elf.py
+++ b/snapcraft/internal/elf.py
@@ -136,8 +136,14 @@ class Patcher:
 
     def _patch_rpath(self, elf_file: ElfFile) -> None:
         rpath = self._get_rpath(elf_file)
-        print(elf_file.path, 'rpath: ', rpath)
-        self._run_patchelf(['--force-rpath', '--set-rpath', rpath],
+        # Parameters:
+        # --force-rpath: use RPATH instead of RUNPATH.
+        # --shrink-rpath: will remove unneeded entries, with the
+        #                 side effect of prefering host libraries
+        #                 so we simply do not use it.
+        # --set-rpath: set the RPATH to the colon separated argument.
+        self._run_patchelf(['--force-rpath',
+                            '--set-rpath', rpath],
                            elf_file.path)
 
     def _run_patchelf(self, args: List[str], elf_file_path: str) -> None:
--- a/snapcraft/internal/elf.py
+++ b/snapcraft/internal/elf.py
@@ -38,7 +38,7 @@ logger = logging.getLogger(__name__)
 class ElfFile:
     """ElfFile represents and elf file on a path and its attributes."""
 
-    def __init__(self, *, path: str, is_executable: bool) -> None:
+    def __init__(self, *, path: str, is_executable: bool=True) -> None:
         """Initialize an ElfFile instance.
 
         :param str path: path to an elf_file within a snapcraft project.
@@ -48,18 +48,61 @@ class ElfFile:
         """
         self.path = path
         self.is_executable = is_executable
+        self.dependencies = set()  # type: Set[str]
+
+    def load_dependencies(self) -> Set[str]:
+        """Load the set of libraries that are needed to satisfy elf's runtime.
+
+        This may include libraries contained within the project.
+        The object's .dependencies attribute is set after loading.
+
+        :returns: a set of string with paths to the library dependencies of
+                  elf.
+        """
+        logger.debug('Getting dependencies for {!r}'.format(self.path))
+        ldd_out = []  # type: List[str]
+        try:
+            ldd_out = common.run_output(['ldd', self.path]).split('\n')
+        except subprocess.CalledProcessError:
+            logger.warning(
+                'Unable to determine library dependencies for '
+                '{!r}'.format(self.path))
+            return set()
+        ldd_out_split = [l.split() for l in ldd_out]
+        ldd_out_list = [l[2] for l in ldd_out_split
+                        if len(l) > 2 and os.path.exists(l[2])]
+
+        # Now lets filter out what would be on the system
+        system_libs = _get_system_libs()
+        libs = {l for l in ldd_out_list
+                if not os.path.basename(l) in system_libs}
+        # Classic confinement won't do what you want with library crawling
+        # and has a nasty side effect described in LP: #1670100
+        libs = {l for l in libs if not l.startswith('/snap/')}
+
+        self.dependencies = libs
+        return libs
 
 
 class Patcher:
     """Patcher holds the necessary logic to patch elf files."""
 
-    def __init__(self, *, dynamic_linker: str) -> None:
+    def __init__(self, *, dynamic_linker: str,
+                 library_path_func=lambda x: x,
+                 base_rpaths: List[str]= None) -> None:
         """Create a Patcher instance.
 
         :param str dynamic_linker: the path to the dynamic linker to set the
                                    elf file to.
+        :param library_path_func: function to determine the library path to a
+                                  given dependency of ElfFile.
+        :param list base_rpaths: library paths for the used base snap.
         """
         self._dynamic_linker = dynamic_linker
+        self._library_path_func = library_path_func
+        if not base_rpaths:
+            base_rpaths = []
+        self._base_rpaths = base_rpaths
 
         # If we are running from the snap we want to use the patchelf
         # bundled there as it would have the capabilty of working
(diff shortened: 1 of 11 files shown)


sergiusens:
Great, only the CLA check failed due to merges.

sergiusens:
going to rebase to get this CLA error fixed, everything else should be good

snapcraft/internal/elf.py
```diff
@@ -82,21 +125,42 @@ def patch(self, *, elf_file: ElfFile) -> None:
...
+            self._patch_rpath(elf_file)
+
+    def _patch_interpreter(self, elf_file: ElfFile) -> None:
+        self._run_patchelf(['--set-interpreter',  self._dynamic_linker],
+                           elf_file.path)
+
+    def _patch_rpath(self, elf_file: ElfFile) -> None:
+        rpath = self._get_rpath(elf_file)
+        print(elf_file.path, 'rpath: ', rpath)
+        self._run_patchelf(['--force-rpath', '--set-rpath', rpath],
```
kyrofa: Is it asking for problems to add `--shrink-rpath` here? Note that it would be nice to see a blurb about the options you're using here and why.

snapcraft/internal/elf.py
```diff
@@ -82,21 +125,42 @@ def patch(self, *, elf_file: ElfFile) -> None:
...
+        if elf_file.dependencies:
+            self._patch_rpath(elf_file)
+
+    def _patch_interpreter(self, elf_file: ElfFile) -> None:
+        self._run_patchelf(['--set-interpreter',  self._dynamic_linker],
+                           elf_file.path)
+
+    def _patch_rpath(self, elf_file: ElfFile) -> None:
+        rpath = self._get_rpath(elf_file)
+        print(elf_file.path, 'rpath: ', rpath)
```
kyrofa: This looks like something that should be removed.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -470,9 +470,36 @@ def prime(self, force=False) -> None:
...
+            # we can move to a smarter library capturing mechanism.
+            library_path = library_path.replace(self.installdir, self.primedir)
+            if library_path.startswith(core_path):
+                return library_path
+            elif (library_path.startswith(self.primedir) and
+                  os.path.exists(self.primedir)):
+                rel_library_path = os.path.relpath(library_path, elf_file_path)
+                rel_library_path_dir = os.path.dirname(rel_library_path)
+                # return the dirname, with the first .. replace with $ORIGIN
+                return rel_library_path_dir.replace('..', '$ORIGIN', 1)
```
kyrofa: This doesn't sound quite right. Doesn't `$ORIGIN` mean "this file"? It sounds like you want this path prepended with it, not using it to replace `..`.

sergiusens: I thought so to, originally, but no, after testing, this is what I want; if you play with relpath you will see what I mean.

snapcraft/internal/elf.py
sergiusens: sure, let's add that and document the params here

snapcraft/internal/elf.py
sergiusens: indeed

sergiusens:
@kyrofa thanks for taking a look; this is the best I could do without breaking any separation of concerns too much and without having to rewrite snapcraft completely :smile:

snapcraft/internal/pluginhandler/__init__.py
sergiusens: Here's an example of the current behaviour:
```
(snapcraft) ubuntu@snapcraft-dev:~/git$ ldd /home/ubuntu/git/prime/libexec/git-core/git-imap-send|grep curl
	libcurl.so.4 => /home/ubuntu/git/prime/libexec/git-core/../../usr/lib/x86_64-linux-gnu/libcurl.so.4 (0x00007f4dab7bb000)
(snapcraft) ubuntu@snapcraft-dev:~/git$ ls /home/ubuntu/git/prime/libexec/git-core/../../usr/lib/x86_64-linux-gnu/libcurl.so.4
/home/ubuntu/git/prime/libexec/git-core/../../usr/lib/x86_64-linux-gnu/libcurl.so.4
```

snapcraft/internal/elf.py
sergiusens: `--shrink-path` is not something we want, it will discard `$ORIGIN` paths in favour of the standard paths, e.g.;
```
$ ldd /home/ubuntu/git/prime/libexec/git-core/git-imap-send|grep curl
	libcurl.so.4 => /usr/lib/x86_64-linux-gnu/libcurl.so.4 (0x00007f02f7b7f000)
```

versus not using `--shrink-path`:
```
$ ldd /home/ubuntu/git/prime/libexec/git-core/git-imap-send|grep curl
	libcurl.so.4 => /home/ubuntu/git/prime/libexec/git-core/../../usr/lib/x86_64-linux-gnu/libcurl.so.4 (0x00007f4dab7bb000)
```

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -470,9 +470,36 @@ def prime(self, force=False) -> None:
...
+        def library_pather(library_path: str, elf_file_path: str) -> str:
+            # If the path is is in the core snap, use the absolute path,
+            # if the path is primed, use $ORIGIN, and last if the dependency
+            # is not anywhere return an empty string.
+            #
+            # Once we move away from the system library grabbing logic
+            # we can move to a smarter library capturing mechanism.
+            library_path = library_path.replace(self.installdir, self.primedir)
+            if library_path.startswith(core_path):
+                return library_path
```
sergiusens: this should be os.path.dirname

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -470,9 +470,36 @@ def prime(self, force=False) -> None:
...
+            elif (library_path.startswith(self.primedir) and
+                  os.path.exists(self.primedir)):
```
sergiusens: this should be library_path

elopio:
Tests failed with: `ConnectionResetError: [Errno 104] Connection reset by peer`
Rerunning...

snapcraft/internal/elf.py
```diff
@@ -48,18 +48,61 @@ def __init__(self, *, path: str, is_executable: bool) -> None:
         """
         self.path = path
         self.is_executable = is_executable
+        self.dependencies = set()  # type: Set[str]
+
+    def load_dependencies(self) -> Set[str]:
+        """Load the set of libraries that are needed to satisfy elf's runtime.
+
+        This may include libraries contained within the project.
+        The object's .dependencies attribute is set after loading.
```
elopio: What about making this a lazy property instead of a side-effect?
Make _load_dependencies internal, and when ElfFile.dependencies is called the first time, it will load the dependencies. The following times, it will just return the loaded value.

snapcraft/internal/elf.py
```diff
@@ -38,7 +38,7 @@
 class ElfFile:
     """ElfFile represents and elf file on a path and its attributes."""
 
-    def __init__(self, *, path: str, is_executable: bool) -> None:
+    def __init__(self, *, path: str, is_executable: bool=True) -> None:
```
elopio: This is probably a nit that you can ignore, but I'm not very happy with this `is_executable` attribute. Wouldn't it be better to remove the attribute, and add an `is_executable()` method that checks the headers?
The downside is that you would have to use magic here again. Or maybe, we move all the checks inside this constructor. If it's not an elf, init raises an exception. And we could have also a method for `is_dynamically_linked()`.

snapcraft/internal/elf.py
```diff
@@ -82,21 +125,48 @@ def patch(self, *, elf_file: ElfFile) -> None:
         :raises snapcraft.internal.errors.PatcherError:
             raised when the elf_file cannot be patched.
         """
-        # When setting rpath for libraries is implemented, we will do more
-        # here.
-        if not elf_file.is_executable:
-            return
+        if elf_file.is_executable:
+            self._patch_interpreter(elf_file)
+        if elf_file.dependencies:
```
elopio: The docstring of the function should be updated. It currently just mentions what will happen when it's executable, not when it has dependencies.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -470,9 +470,36 @@ def prime(self, force=False) -> None:
...
+            # we can move to a smarter library capturing mechanism.
+            library_path = library_path.replace(self.installdir, self.primedir)
+            if library_path.startswith(core_path):
+                return os.path.dirname(library_path)
+            elif (library_path.startswith(self.primedir) and
+                  os.path.exists(library_path)):
+                rel_library_path = os.path.relpath(library_path, elf_file_path)
+                rel_library_path_dir = os.path.dirname(rel_library_path)
+                # return the dirname, with the first .. replace with $ORIGIN
+                return rel_library_path_dir.replace('..', '$ORIGIN', 1)
```
elopio: Why is there no test that checks for this $ORIGIN condition?

sergiusens: Because I missed it I guess?

snapcraft/internal/elf.py
sergiusens: I'll let @kyrofa weigh in. I am not a fan of either.

sergiusens: as in, either your or my proposal :wink:

kyrofa:
Ah no, ignore me.

kyrofa:
I'm poking at the $ORIGIN bit, and have a question.

My first test, I'm using the opencv demo with classic confinement. I run `snapcraft build`, and then see this:

```
$ ldd parts/example/install/bin/example 
	linux-vdso.so.1 =>  (0x00007ffc69c36000)
	libopencv_core.so.2.4 => /usr/lib/x86_64-linux-gnu/libopencv_core.so.2.4 (0x00007f22db4f5000)
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f22db173000)
        <snip>
```

So far so good. Then I run `snapcraft prime`, and see this:

```
$ ldd prime/bin/example 
	linux-vdso.so.1 =>  (0x00007ffe78799000)
	libopencv_core.so.2.4 => /usr/lib/x86_64-linux-gnu/libopencv_core.so.2.4 (0x00007f9f26c4a000)
	libstdc++.so.6 => /snap/core/current/usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f9f268c8000)
        <snip>
```

Note that libstdc++ is looking in the core snap, as it should, but libopencv_core is still looking in /usr/lib, when I expected it to be looking in prime/usr/lib. What happened?

Second test, same demo but now with classic confinement as well as `stage-packages: [libopencv-core2.4v5]` on the part, after `snapcraft prime` I see this:

```
$ ldd prime/bin/example 
	linux-vdso.so.1 =>  (0x00007ffc46dfd000)
	libopencv_core.so.2.4 => /home/kyrofa/src/snapcraft/demos/opencv/prime/bin/../usr/lib/x86_64-linux-gnu/libopencv_core.so.2.4 (0x00007f69e6077000)
	libstdc++.so.6 => /snap/core/current/usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f69e5cf5000)
        <snip>
```

That looks good. So something is off with the crawling relationship.

sergiusens:
The `libopencv_core.so.2.4` from the system is expected current behaviour. For this we need to disable loading libraries from anything different than rpath and that is why we need the newer patchelf.

This is why system library crawling did not work for classic snaps to begin with.

Also, consider this, looks like inodes change (depending on the FS).
```
(snapcraft) ubuntu@snapcraft-dev:~/source/snapcraft/demos/opencv$ ls -i parts/example/install/bin/example
527045 parts/example/install/bin/example
(snapcraft) ubuntu@snapcraft-dev:~/source/snapcraft/demos/opencv$ ls -i /home/ubuntu/source/snapcraft/demos/opencv/prime/bin/example
527083 /home/ubuntu/source/snapcraft/demos/opencv/prime/bin/example
(snapcraft) ubuntu@snapcraft-dev:~/source/snapcraft/demos/opencv$ ldd parts/example/install/bin/example 
	linux-vdso.so.1 =>  (0x00007fff2e1f1000)
	libopencv_core.so.2.4 => /usr/lib/x86_64-linux-gnu/libopencv_core.so.2.4 (0x00007fb9557c6000)
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007fb955444000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fb95522e000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fb954e64000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007fb954c4a000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fb954a2d000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007fb954825000)
	libGL.so.1 => /usr/lib/x86_64-linux-gnu/mesa/libGL.so.1 (0x00007fb9545b3000)
	libtbb.so.2 => /usr/lib/x86_64-linux-gnu/libtbb.so.2 (0x00007fb954376000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fb95406d000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fb955bf0000)
	libexpat.so.1 => /lib/x86_64-linux-gnu/libexpat.so.1 (0x00007fb953e44000)
	libxcb-dri3.so.0 => /usr/lib/x86_64-linux-gnu/libxcb-dri3.so.0 (0x00007fb953c41000)
	libxcb-present.so.0 => /usr/lib/x86_64-linux-gnu/libxcb-present.so.0 (0x00007fb953a3e000)
	libxcb-sync.so.1 => /usr/lib/x86_64-linux-gnu/libxcb-sync.so.1 (0x00007fb953837000)
	libxshmfence.so.1 => /usr/lib/x86_64-linux-gnu/libxshmfence.so.1 (0x00007fb953634000)
	libglapi.so.0 => /usr/lib/x86_64-linux-gnu/libglapi.so.0 (0x00007fb953405000)
	libXext.so.6 => /usr/lib/x86_64-linux-gnu/libXext.so.6 (0x00007fb9531f3000)
	libXdamage.so.1 => /usr/lib/x86_64-linux-gnu/libXdamage.so.1 (0x00007fb952ff0000)
	libXfixes.so.3 => /usr/lib/x86_64-linux-gnu/libXfixes.so.3 (0x00007fb952dea000)
	libX11-xcb.so.1 => /usr/lib/x86_64-linux-gnu/libX11-xcb.so.1 (0x00007fb952be8000)
	libX11.so.6 => /usr/lib/x86_64-linux-gnu/libX11.so.6 (0x00007fb9528ae000)
	libxcb-glx.so.0 => /usr/lib/x86_64-linux-gnu/libxcb-glx.so.0 (0x00007fb952695000)
	libxcb-dri2.so.0 => /usr/lib/x86_64-linux-gnu/libxcb-dri2.so.0 (0x00007fb952490000)
	libxcb.so.1 => /usr/lib/x86_64-linux-gnu/libxcb.so.1 (0x00007fb95226e000)
	libXxf86vm.so.1 => /usr/lib/x86_64-linux-gnu/libXxf86vm.so.1 (0x00007fb952068000)
	libdrm.so.2 => /usr/lib/x86_64-linux-gnu/libdrm.so.2 (0x00007fb951e57000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fb951c53000)
	libXau.so.6 => /usr/lib/x86_64-linux-gnu/libXau.so.6 (0x00007fb951a4f000)
	libXdmcp.so.6 => /usr/lib/x86_64-linux-gnu/libXdmcp.so.6 (0x00007fb951849000)
```

kyrofa:
> For this we need to disable loading libraries from anything different than rpath and that is why we need the newer patchelf.

Ah! So if I ran from the snap you think this would work?

sergiusens:
@kyrofa the case where you haven't staged libopencv_core would fail just as bad.

sergiusens:
@kyrofa here's `master`:
```
(snapcraft) ubuntu@snapcraft-dev:~/source/snapcraft/demos/opencv$ ldd parts/example/install/bin/example |grep opencv
	libopencv_core.so.2.4 => /usr/lib/x86_64-linux-gnu/libopencv_core.so.2.4 (0x00007f9943e6c000)
(snapcraft) ubuntu@snapcraft-dev:~/source/snapcraft/demos/opencv$ git branch ^C
(snapcraft) ubuntu@snapcraft-dev:~/source/snapcraft/demos/opencv$ git symbolic-ref --short HEAD
master
```

The take away is that the opencv lib *needs* to be staged for things to work. Now that we are using `patchelf` instead, we can do things in `prime` which weren't possible before, such as running `patchelf` to patch libraries with `--no-default-lib`

kyrofa:
@sergiusens talked about this offline. Turns out that libraries aren't actually supposed to be crawled at all on classic, so the behavior I witnessed, while unexpected, is not the fault of this PR.

snapcraft/internal/elf.py
kyrofa: Yeah I agree with @elopio, it'd be nice to see this encapsulated a little more, although I'd like to see it either lazy or in `__init__()`. I suspect you wouldn't want to load up magic every time though... I suppose you could pass it in.

sergiusens: heh, this was previous PR material ;)

snapcraft/internal/elf.py
sergiusens: this func/method will die in the following PRs, I'd rather not do it.

sergiusens: We also, in its current form would want to reload (that can be solved with a new method)

snapcraft/internal/elf.py
```diff
@@ -38,31 +38,72 @@
 class ElfFile:
     """ElfFile represents and elf file on a path and its attributes."""
 
-    def __init__(self, *, path: str, is_executable: bool) -> None:
+    def __init__(self, *, path: str, magic: str) -> None:
```
elopio: I like this better, thanks.

snaps_tests/demos_tests/test_opencv.py
```diff
@@ -13,18 +13,49 @@
...
+        expected_interpreter = r'^/snap/core/current/.*'
+        self.assertThat(interpreter, MatchesRegex(expected_interpreter))
+
+        arch_triplet = snapcraft.ProjectOptions().arch_triplet
+
+        # test $ORIGIN in action
+        rpath = subprocess.check_output([
+            'patchelf', '--print-rpath', bin_path]).decode()
+        expected_rpath = '$ORIGIN/../usr/lib/{}:'.format(arch_triplet)
+        self.assertThat(rpath, Contains(expected_rpath))
```
elopio: This is good to me for now. Maybe we can raise the priority to split this test, so we get it done before the year is over.

PR #1782, opened 2017-12-01
project_loader: support grammar on architectures

### LINKED ISSUE #1681: Design/document support for architecture selection during builds
Estimate: 1 day

Snapd’s snap.yaml supports an architectures keyword that defines the architectures on which that particular snap runs. Snapcraft currently supports the same keyword, and it acts as a pass-through, copying it into the snap.yaml. Now that build.snapcraft.io is a thing, we’d like to expand the architectures keyword to also serve the purpose of telling build.snapcraft.io which architectures to use for building, and on which architectures the resulting snap runs.

kyrofa:

-----

Currently Snapcraft supports a flat list of architecture overrides in the snapcraft.yaml that look something like this:

    architectures: [amd64, i386]

If this property is specified, Snapcraft passes it on verbatim to the snap.yaml, which then assumes the same generated snap supports both of those architectures (called a "fat" snap).

This PR resolves #1681 by expanding Snapcraft's `architectures` property to support the advanced grammar, which allows the property to be dual-purposed into not only run-time architecture support, but build-time (i.e. specify the architecture on which it should be built, as well as the architectures on which the resulting snap can be run). That looks something like this:

    architectures:
      - on i386: [amd64, i386]
      - on armhf: [arm64, armhf]

The final snap.yaml will still only contain a flat list of architectures.

### DIFF
--- a/snapcraft/_options.py
+++ b/snapcraft/_options.py
@@ -20,7 +20,7 @@ import os
 import platform
 import sys
 
-# These are used my a mypy 'type:' comment below, and nowhere else. As a
+# These are used by a mypy 'type:' comment below, and nowhere else. As a
 # result, flake8 doesn't think they're used. Thus noqa.
 from typing import Dict, List, Union  # noqa
 
--- a/snapcraft/_options.py
+++ b/snapcraft/_options.py
@@ -20,6 +20,10 @@ import os
 import platform
 import sys
 
+# These are used my a mypy 'type:' comment below, and nowhere else. As a
+# result, flake8 doesn't think they're used. Thus noqa.
+from typing import Dict, List, Union  # noqa
+
 from snapcraft.internal import common
 from snapcraft.internal.deprecations import handle_deprecation_notice
 from snapcraft.internal.errors import SnapcraftEnvironmentError
@@ -91,7 +95,7 @@ _ARCH_TRANSLATIONS = {
         'triplet': 's390x-linux-gnu',
         'core-dynamic-linker': 'lib/ld64.so.1',
     }
-}
+}  # type: Dict[str, Dict[str, Union[str, List[str]]]]
 
 
 _32BIT_USERSPACE_ARCHITECTURE = {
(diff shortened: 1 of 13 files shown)


snapcraft/_options.py
```diff
@@ -20,6 +20,10 @@
 import platform
 import sys
 
+# These are used my a mypy 'type:' comment below, and nowhere else. As a
```
sergiusens: used _by_ a?

kyrofa: Whoops, yes thank you.

sergiusens:
Seems like we are making it or breaking it with this one.

So here are some questions:
- do we want to support backwards compatibility with this entry? During our initial conversations we said we would.
- If we are, we are missing a deprecation notice.

sergiusens:
Also for any reader, this PR is not accounting for the cross compilation aspects that would be introduced later.

kyrofa:
> do we want to support backwards compatibility with this entry? During our initial conversations we said we would.

Oh most certainly! It's enshrined in numerous tests, including [this one](https://github.com/snapcore/snapcraft/pull/1782/files#diff-8be6197695f8be133d0bc895f04ce724R35).

kyrofa:
> [...] this PR is not accounting for the cross compilation aspects that would be introduced later.

Eh? What cross-compilation aspects? Perhaps my memory is failing me, but I'm under the impression that this is for consumption by build.snapcraft.io, not for snapcraft to cross-compile as necessary.

elopio:
Tests failed with a weird `error: websocket: close 1006 (abnormal closure): unexpected EOF`. I've rerun them.

sergiusens:
Can we add the the deprecation notice for when it is a plain list?

kyrofa:
> Can we add the the deprecation notice for when it is a plain list?

Sure, although I didn't realize we were deprecating that. Why?

sergiusens:
@kyrofa because it is backwards behavior and build.snapcraft.io will eventually not understand it

sergiusens:
closing for further discussion

PR #1789, opened 2017-12-04
cli: improve the help command

sergiusens:
Give it parity with --help but also extend it so topics and plugin help is
not completely hidden.

LP: #1698116

-----

### DIFF
--- a/snapcraft/cli/help.py
+++ b/snapcraft/cli/help.py
@@ -20,6 +20,7 @@ from textwrap import dedent
 import click
 
 import snapcraft
+from . import echo
 from snapcraft.internal import sources
 
 
@@ -61,7 +62,7 @@ def help_command(ctx, topic, devel):
         click.echo(ctx.parent.get_help())
         click.echo(dedent("""\
 
-            To get additional help, run:
+            For more help, use:
                 snapcraft help topics
                 snapcraft help <topic>
                 snapcraft help <plugin-name>
@@ -75,16 +76,20 @@ def help_command(ctx, topic, devel):
         try:
             _module_help(topic, devel)
         except ImportError:
-            click.echo(dedent("""\
-    The argument to help is not a plugin name nor topic, to see the
-    available topics run:
+            # 10 is the limit which determines ellipsis is needed
+            if len(topic) > 10:
+                topic = '{}...'.format(topic[:10])
+            echo.wrapped(dedent("""\
+    There is no help topic or plugin {!r}. Try:
+
+    For topics:
 
         snapcraft help topics
 
-    to see the available plugins run:
+    For valid plugins:
 
         snapcraft list-plugins
-    """))
+    """).format(topic))
             sys.exit(1)
 
 
--- a/snapcraft/cli/help.py
+++ b/snapcraft/cli/help.py
@@ -61,7 +61,7 @@ def help_command(ctx, topic, devel):
         click.echo(ctx.parent.get_help())
         click.echo(dedent("""\
 
-            You can obtain more help by running
+            To get additional help, run:
                 snapcraft help topics
                 snapcraft help <topic>
                 snapcraft help <plugin-name>
--- a/snapcraft/cli/help.py
+++ b/snapcraft/cli/help.py
@@ -15,12 +15,12 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 import importlib
 import sys
+from textwrap import dedent
 
 import click
 
 import snapcraft
 from snapcraft.internal import sources
-from . import echo
 
 
 _TOPICS = {
@@ -36,10 +36,11 @@ def helpcli():
 
 
 @helpcli.command('help')
-@click.argument('topic', metavar='<topic>')
+@click.argument('topic', metavar='<topic>', required=False)
 @click.option('--devel', is_flag=True,
               help='Show more details for snapcraft developers')
-def help_command(topic, devel):
+@click.pass_context
+def help_command(ctx, topic, devel):
     """Obtain help for a certain plugin or topic.
 
     The <topic> can either be a plugin name or one of:
@@ -56,7 +57,16 @@ def help_command(topic, devel):
         snapcraft help sources
         snapcraft help go
     """
-    if topic == 'topics':
+    if not topic:
+        click.echo(ctx.parent.get_help())
+        click.echo(dedent("""\
+
+            You can obtain more help by running
+                snapcraft help topics
+                snapcraft help <topic>
+                snapcraft help <plugin-name>
+        """))
+    elif topic == 'topics':
         for key in _TOPICS:
             click.echo(key)
     elif topic in _TOPICS:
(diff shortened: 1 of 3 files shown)


snapcraft/cli/help.py
```diff
@@ -56,7 +57,16 @@ def help_command(topic, devel):
         snapcraft help sources
         snapcraft help go
     """
-    if topic == 'topics':
+    if not topic:
+        click.echo(ctx.parent.get_help())
+        click.echo(dedent("""\
+
+            You can obtain more help by running
```
elopio: On my pr about commands on the error message, I tried to make them consistent using a wording like this:
`To obtain more help, run`

snapcraft/cli/help.py
```diff
@@ -65,8 +75,16 @@ def help_command(topic, devel):
         try:
             _module_help(topic, devel)
         except ImportError:
-            echo.error('The plugin does not exist. Run `snapcraft '
-                       'list-plugins` to see the available plugins.')
+            click.echo(dedent("""\
```
elopio: can _module_help raise and exception that takes care of this message and formatting, so it can be in `snapcraft/internal/errors` like the rest?

snapcraft/cli/help.py
```diff
@@ -65,8 +75,16 @@ def help_command(topic, devel):
...
+    The argument to help is not a plugin name nor topic, to see the
+    available topics run:
+
+        snapcraft help topics
+
+    to see the available plugins run:
+
+        snapcraft list-plugins
+    """))
```
elopio: To make this message consistent with the others on the errors pr branch, I would change it to something like:
```
Failed to display help: '{topic}' is not a plugin name nor a topic. To see the available topics, run `snapcraft help topics`. To see the available plugins, run `snapcraft list-plugins`.
```

snapcraft/cli/help.py
sergiusens: I want to keep things implemented in cli independent from `internal.errors`. That would also print the error in red which is just horrible for so much text.

snapcraft/cli/help.py
sergiusens: That has no wrapping, doesn't seem elegant to me.

snapcraft/cli/help.py
sergiusens: sure thing

sergiusens: oh wait, this is not an error message

elopio: I know this one is not an error message, but I would like us to use consistent wording when asking or suggesting the user to run a command.

snapcraft/cli/help.py
elopio: In my proposal, I left the messages as short as possible, with no unnecessary white spaces. And left the wrapping to the terminal. This was not suggested by mpt, just my suggestion. I don't have a strong opinion on the format, just that it should be consistent, so I can adjust my pr if you prefer a different style.

sergiusens: The wrapping on the terminal will cut words in half, the only place where we shouldn't cut is for links so they can still open.

If you look at the in-master errors, some have `\n` for exactly this reason.

sergiusens: Also, consider this the help for running a command incorrectly, not necessarily an error in "snapcraft" operation.

sergiusens:
Here's the PR in action:

```
(snapcraft) ubuntu@snapcraft-dev:~/source/snapcraft$ snapcraft help
Usage: snapcraft [OPTIONS] COMMAND [ARGS]...

  Snapcraft is a delightful packaging tool.

Options:
  --version    Show the version and exit.
  -d, --debug
  --help       Show this message and exit.

Commands:
  build            Build artifacts defined for a part.
  clean            Remove content - cleans downloads, builds or...
  cleanbuild       Create a snap using a clean environment...
  close            Close <channel> for <snap-name>.
  create-key       Create a key to sign assertions.
  define           Shows the definition for the cloud part.
  enable-ci        Enable continuous-integration systems to...
  gated            Get the list of snaps and revisions gating a...
  help             Obtain help for a certain plugin or topic.
  init             Initialize a snapcraft project.
  list-keys        List the keys available to sign assertions.
  list-plugins     List the available plugins that handle...
  list-registered  List snap names registered or shared with...
  list-revisions   Get the history on the store for <snap-name>.
  login            Login with your Ubuntu One e-mail address and...
  logout           Clear session credentials.
  pack             Create a snap from a directory holding a...
  prime            Final copy and preparation for the snap.
  pull             Download or retrieve artifacts defined for a...
  push             Push <snap-file> to the store.
  push-metadata    Push metadata from <snap-file> to the store.
  register         Register <snap-name> with the store.
  register-key     Register a key with the store to sign...
  release          Release <snap-name> on <revision> to the...
  search           Searches the remote parts cache for matching...
  sign-build       Sign a built snap file and assert it using...
  snap             Create a snap.
  stage            Stage the part's built artifacts into the...
  status           Get the status on the store for <snap-name>.
  update           Updates the parts listing from the cloud.
  validate         Validate a gated snap.
  whoami           Returns your login information relevant to...

To get additional help, run:
    snapcraft help topics
    snapcraft help <topic>
    snapcraft help <plugin-name>

(snapcraft) ubuntu@snapcraft-dev:~/source/snapcraft$ snapcraft help topics
plugins
sources
(snapcraft) ubuntu@snapcraft-dev:~/source/snapcraft$ snapcraft help non-topic
The argument to help is not a plugin name nor topic, to see the
available topics run:

    snapcraft help topics

to see the available plugins run:

    snapcraft list-plugins

```

snapcraft/cli/help.py
```diff
@@ -56,7 +57,16 @@ def help_command(topic, devel):
         snapcraft help sources
         snapcraft help go
     """
-    if topic == 'topics':
+    if not topic:
+        click.echo(ctx.parent.get_help())
+        click.echo(dedent("""\
+
+            To get additional help, run:
```
matthewpaulthomas: This is quite correct, but it’s odd to think of help as something being “run”, rather than looked up. Instead, maybe “For more help, use:”

The same applies to all the following uses of “run:”

snapcraft/cli/help.py
matthewpaulthomas: I agree with @elopio that it’s best to repeat the unknown command, since that would help you realize if you’d mistyped it. (But ellipsize it at, say, 64 characters, in case you accidentally pasted a whole StackOverflow page into the terminal.)

However, in this case I don’t actually think it’s a good idea to follow the “what went wrong, why it happened, what you can do” format, for the same reason as I don’t think it’s a good idea to use the word “run”: looking up help doesn’t feel like a command. Enough just to begin ‘There is no help topic or plugin “fqwhgads”. Try:’

> The wrapping on the terminal will cut words in half, the only place where we shouldn't cut is for links so they can still open.

When output is mostly words (like it is for help, and for error messages), should it perhaps go through a formatter that word-wraps it to the current terminal width? A sprinkling of `\n` will sometimes be too much and sometimes not enough.

sergiusens: I've added a formatter, the same one we use for help already.

PR #1795, opened 2017-12-05
lifecycle: do not include source tarballs from previous runs

ted-gould:
Playing with the version number made me realize my source tarballs were getting bigger and bigger... fixing.

### DIFF
--- a/snapcraft/internal/lifecycle/_containers.py
+++ b/snapcraft/internal/lifecycle/_containers.py
@@ -32,6 +32,8 @@ def _create_tar_filter(tar_filename):
             return None
         elif fn.endswith('.snap'):
             return None
+        elif fn.endswith('_source.tar.bz2'):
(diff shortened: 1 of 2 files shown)


sergiusens:
thanks Ted!

snapcraft/internal/lifecycle/_containers.py
```diff
@@ -32,6 +32,8 @@ def _tar_filter(tarinfo):
             return None
         elif fn.endswith('.snap'):
             return None
+        elif fn.endswith('_source.tar.bz2'):
+            return None
```
kalikiana: Very nice! Thank you!

kyrofa:
This PR seems sensible, but I'm curious: why do we keep the tarball that we make lying around anyway? Shouldn't we clean it up?

PR #1798, opened 2017-12-07
elf: strip the .note.go.buildid to make room for patching elf

sergiusens:
patchelf does not handle the existence of the .note.go.buildid section
so we strip it from the resulting binary as from the perspective of the
snap it is not needed.

A newer patchelf is also required which handles the math required for
the way the headers are layed out when using golang's go.

LP: #1736861

Signed-off-by: Ubuntu <sergio.schvezov@canonical.com>

-----

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -8,7 +8,6 @@ Build-Depends: bash-completion,
                dh-python,
                git,
                mercurial,
-               patchelf,
                pkg-config,
                python3 (>= 3.4),
                python3-apt,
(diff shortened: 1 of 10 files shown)


elopio:
I'm confused about what will happen in the deb in xenial. Will we just warn, and the behaviour will be the same as if we had build the snap with snapcraft 2.35?

elopio:
With this change, gotty builds successfully.

sergiusens:
@elopio an integration test is in the works, I was just sharing this with you for you to verify. For the deb we will just add `patchelf` as a `build-snap`, it is already in the store on the `edge` channel.

sergiusens:
binutils is a base requirement on every distro, more so in Ubuntu, so I'd say no.
The tests did pass already, I just added that last commit for more versatility.

I do wish to release it to stable as it would solve many vendors problems. We should at least move it to candidate.

PR #1802, opened 2017-12-12
metadata: extract metadata from appstream

### LINKED ISSUE #1693: Develop metadata handler system
Estimate: 1 week

Currently Snapcraft requires all its metadata (snap name, version, etc.) to be specified in the snapcraft.yaml. However, lots of parts already contain such information, whether from a setup.py, XML, AppStream, etc.

I’m considering this task to only be for existing metadata (name, version, summary, etc.). Adding support for an AppStream ID etc. is a separate task

elopio:

-----
See the design discussion in the [forum](https://forum.snapcraft.io/t/extracting-existing-data-from-projects-to-feed-into-snap-yaml/2285/8). Closes #1693

### DIFF
--- a/schema/snapcraft.yaml
+++ b/schema/snapcraft.yaml
@@ -34,6 +34,9 @@ properties:
   build-packages:
     $ref: "#/definitions/grammar-array"
     description: top level build packages.
+  adopt-info:
+    type: string
+    description: name of the part that provides source files that will be parsed to extract snap metadata information
   name:
     type: string
     description: name of the snap package
@@ -361,6 +364,8 @@ properties:
           prepare:
             type: string
             default: ''
+          parse-info:
+            type: array
   plugs:
     type: object
   slots:
@@ -368,9 +373,11 @@ properties:
 required:
   - name
   - version
-  - summary
-  - description
   - parts
+# required by snap.yaml, but that can be filled later by other metadata
+# sources, like appstream.
+#  - summary
+#  - description
 dependencies:
   license-agreement: ["license"]
   license-version: ["license"]
--- a/snapcraft/internal/meta/_appstream.py
+++ b/snapcraft/internal/meta/_appstream.py
@@ -24,10 +24,10 @@ from xml.etree.ElementTree import (
 
 class AppstreamInfoParser:
 
-    def __init__(self, path):
+    def __init__(self, path: str) -> None:
         self._path = path
 
-    def extract(self, keys):
+    def extract(self, keys: list) -> dict:
         info = {}
         try:
             tree = ElementTree().parse(self._path)
--- a/snapcraft/internal/meta/_appstream.py
+++ b/snapcraft/internal/meta/_appstream.py
@@ -1,0 +1,40 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2017 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from snapcraft.internal.meta import _errors
+
+from xml.etree.ElementTree import (
+    ElementTree,
+    ParseError
+)
+
+
+class AppstreamInfoParser:
+
+    def __init__(self, path):
+        self._path = path
+
+    def extract(self, keys):
+        info = {}
+        try:
+            tree = ElementTree().parse(self._path)
+        except ParseError:
+            raise _errors.AppstreamFileParseError(self._path)
+        for key in keys:
+            child = tree.find(key)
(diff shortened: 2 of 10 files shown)


sergiusens:
I was thinking more that this should have been handled during prime by the pluginhandler storing the _values_ into the state tracker for the part, pushing them onto a stack; then during meta creation, this would be retrievable from this stack.

I say stack it instead of replace to get an idea of what is going on when weird things happen. Doing this in the pluginhandler will also ease the implementation of the helpers such as `set-description` and `set-name`

sergiusens:
This is missing icon extraction and possibly desktop

kyrofa:
I assume this is referencing the design [here](https://forum.snapcraft.io/t/extracting-existing-data-from-projects-to-feed-into-snap-yaml/2285/8)?

schema/snapcraft.yaml
```diff
@@ -361,16 +364,20 @@ properties:
...
   slots:
     type: object
 required:
   - name
   - version
-  - summary
-  - description
   - parts
+# required by snap.yaml, but that can be filled later by other metadata
+# sources, like appstream.
```
kyrofa: Excellent comment, thank you!

schema/snapcraft.yaml
```diff
@@ -361,16 +364,20 @@ properties:
           prepare:
             type: string
             default: ''
+          parse-info:
+            type: array
```
kyrofa: The array items are always going to be strings, right? We should probably flesh this schema definition out to enforce that.

kyrofa:
Also, I thought the underlying metadata extraction was going to be its own PR, separate from the appstream implementation of it?

snapcraft/internal/meta/_appstream.py
```diff
@@ -0,0 +1,40 @@
...
+    ParseError
+)
+
+
+class AppstreamInfoParser:
+
+    def __init__(self, path: str) -> None:
+        self._path = path
+
+    def extract(self, keys: list) -> dict:
```
kyrofa: I didn't realize this was an option, haha! You might consider using `Dict` though, it gives you the ability to specify key/value types as well.

snapcraft/internal/meta/_errors.py
```diff
@@ -21,3 +21,72 @@ class CommandError(errors.SnapcraftError):
 
     def __init__(self, message: str) -> None:
         self.fmt = message
+
+
+class SnapMetaGenerationError(errors.SnapcraftError):
+    pass
+
+
+class SnapcraftYamlMissingRequiredValue(SnapMetaGenerationError):
```
kyrofa: Just looking at the errors, I can't figure out the difference between this error class and `MissingSnapcraftYamlKeys`.

snapcraft/internal/meta/_from_sources.py
```diff
@@ -0,0 +1,74 @@
...
+
+import os
+
+from snapcraft.internal.meta import (
+    _appstream,
+    _errors
+)
+
+
+def get_info(keys: list, config_data: dict, parts_dir: str) -> dict:
```
kyrofa: Consider using something more specific than `dict` here as well.

snapcraft/internal/meta/_errors.py
kyrofa: After reading where it's used, it sounds like this should be part-specific, no?

snapcraft/internal/meta/_from_sources.py
```diff
@@ -0,0 +1,74 @@
...
+            if 'parse-info' in config_data['parts'][part]:
+                part_dir = os.path.join(parts_dir, part)
+                info_paths = config_data['parts'][part]['parse-info']
+                return _extract_info_from_part(
+                    keys,
+                    config_data['parts'][part]['plugin'],
+                    part_dir,
+                    info_paths)
+            else:
+                raise _errors.SnapcraftYamlMissingRequiredValue('parse-info')
```
kyrofa: Shouldn't this include the part name?

snapcraft/internal/meta/_errors.py
```diff
@@ -21,3 +21,72 @@ class CommandError(errors.SnapcraftError):
...
+    fmt = (
+        "Failed to generate snap metadata: "
+        "Missing {key!r} from the 'snapcraft.yaml' file."
+    )
+
+    def __init__(self, key: str) -> None:
+        super().__init__(key=key)
+
+
+class WrongAdoptInfo(SnapMetaGenerationError):
```
kyrofa: How about `AdoptedPartMissingError`?

snapcraft/internal/meta/_from_sources.py
```diff
@@ -0,0 +1,74 @@
...
+        raise _errors.SnapcraftYamlMissingRequiredValue('adopt-info')
+
+
+def _extract_info_from_part(keys, plugin, part_dir, info_paths):
+    info = {}
+    for path in info_paths:
+        if keys:
+            source_full_paths = [
+                os.path.join(part_dir, step_dir, path)
+                for step_dir in ('install', 'build')]
```
kyrofa: What about parts that build out-of-source (e.g. the cmake plugin)? We should probably also search the `src/` dir, no?

sergiusens: We should revisit that, maybe it is fine to use src and not just for cmake; in most cases appstream.xml files in sources are found as `.in` templates to be finalized during build. Let's add it, it doesn't hurt and it a fallback mechanism anyways.

sergiusens:
@kyrofa you are right, this needs to be split to solve #1693

sergiusens:
@kyrofa and  #1694

PR #1807, opened 2017-12-14
tests: run test_cleanbuild in LXD on Travis

kalikiana:

-----
Attempt to sort out test_cleanbuild being skipped in #1803 when run as a snap
Fixes: [bug 1738210](https://bugs.launchpad.net/snapcraft/+bug/1738210)

### DIFF
--- a/snapcraft/internal/project_loader/_config.py
+++ b/snapcraft/internal/project_loader/_config.py
@@ -25,6 +25,7 @@ import yaml
 
 import snapcraft
 from snapcraft.internal import (
+    common,
     deprecations,
     remote_parts,
     states,
@@ -116,6 +117,11 @@ class Config:
 
         self.build_tools = grammar_processor.get_build_packages()
         self.build_tools |= set(project_options.additional_build_packages)
+
+        # This is required for patching to work correctly
+        if _requires_patchelf_snap(self.data['confinement']):
+            # TODO use stable once possible
+            self.build_snaps.add('patchelf/edge')
 
         self.parts = PartsConfig(parts=self.data,
                                  project_options=self._project_options,
@@ -305,3 +311,11 @@ def _expand_filesets_for(step, properties):
             new_step_set.append(item)
 
     return new_step_set
+
+
+def _requires_patchelf_snap(confinement: str) -> bool:
+    is_snap = common.is_snap()
+    is_docker_instance = common.is_docker_instance()
+    is_classic = confinement == 'classic'
+
+    return (is_classic and not is_snap and not is_docker_instance)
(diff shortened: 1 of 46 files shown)


sergiusens:
The problem, shown in the previous PR too,
```
error: open /var/lib/snapd/hostfs/root/snap/lxd/common/snapcrafteh8w2lfl/core_3604.assert: no such file or directory
```

kalikiana:
The test is passing now that I updated the script to also use the LXD snap in the containers created by the test cases, which 1) is good for conistency 2) is necessary to avoid an issue (regression?) with snapd confinement (the hostfs path is not available outside of confinement, I'm investigating this separately).

snapcraft/internal/project_loader/_config.py
```diff
@@ -121,7 +121,7 @@ def __init__(self, project_options=None):
         # This is required for patching to work correctly
         if _requires_patchelf_snap(self.data['confinement']):
             # TODO use stable once possible
-            self.build_snaps.add('patchelf/edge')
+            self.build_snaps.add('patchelf/latest/edge')
```
sergiusens: Why this?

kalikiana: Otherwise the test failed like this:

    "/tmp/autopkgtest.6TVtmK/build.46t/snapcraft/snapcraft/internal/repo/snaps.py", line 119, in is_classic
        return store_channels[self.channel]['confinement'] == 'classic'
    KeyError: 'edge'

Which from my understanding is because of "tracking":

    $ snap info patchelf | grep tracking
    tracking:    latest/edge

The same is true for me locally, not just on Travis.

kalikiana: Also, this is the result of `_get_store_channels()`:

    {'latest/edge': {'epoch': '0', 'channel': 'edge', 'version': '0.9-36-g29c085f', 'revision': '6', 'confinement': 'classic', 'size': 327680}}

kalikiana:
test_container_builds failed with this now:

    error: Failed to setup device: mkdir /var/snap/lxd/common/lxd/devices/snapcraft-basic/disk./root/build_basic.root-build_basic: no such file or directory
    [...]
    subprocess.CalledProcessError: Command '['lxc', 'config', 'device', 'add', 'local:snapcraft-basic', '/root/build_basic', 'disk', 'source=/tmp/tmpuv_odnx8', 'path=/root/build_basic']' returned non-zero exit status 1.

Need to investigate this more.

sergiusens:
Just revert that last commit and let's keep it for cleanbuild only.

PR #1810, opened 2017-12-16
[WIP] i18n support

m4sk1n:
Signed-off-by: Marcin Mikołajczak <me@m4sk.in>

TODO:

[ ] resolve conflicts
[ ] make it working (by now I only made strings translatable and added option to generate translation files)


sergiusens:
Thanks, I am not sure bind2domain will work when run as a snap so that would need some research (or maybe require choosing a different tool to deal with translations all together).

Feel free to re-propose once you have time to look into this more.

PR #1815, opened 2017-12-18
repo: handle invalid snaps

sergiusens:
Handle invalid snaps listed under `build-snaps`, provide a nice error. Also add retries to the information fetching logic to be able to have easier debugging information when the network goes wrong.

-----

### DIFF
--- a/snapcraft/internal/repo/errors.py
+++ b/snapcraft/internal/repo/errors.py
@@ -85,7 +85,8 @@ class SnapUnavailableError(RepoError):
 
     fmt = ('Failed to install or refresh a snap: {snap_name!r} does not exist '
            'or is not available on the desired channel {snap_channel!r}. '
-           'Verify that the snap is available.')
+           'Use `snap info {snap_name}` to get a list of channels the '
+           'snap is available on.')
 
     def __init__(self, *, snap_name: str, snap_channel: str) -> None:
         super().__init__(snap_name=snap_name, snap_channel=snap_channel)
--- a/snapcraft/internal/repo/errors.py
+++ b/snapcraft/internal/repo/errors.py
@@ -81,6 +81,16 @@ class UnpackError(RepoError):
         super().__init__(package=package)
 
 
+class SnapUnavailableError(RepoError):
+
+    fmt = ('Failed to install or refresh a snap: {snap_name!r} does not exist '
+           'or is not available on the desired channel {snap_channel!r}. '
+           'Verify that the snap is available.')
+
+    def __init__(self, *, snap_name: str, snap_channel: str) -> None:
+        super().__init__(snap_name=snap_name, snap_channel=snap_channel)
+
+
 class SnapInstallError(RepoError):
 
     fmt = ('Error while installing snap {snap_name!r} from channel '
--- a/snapcraft/internal/repo/snaps.py
+++ b/snapcraft/internal/repo/snaps.py
@@ -138,6 +138,9 @@ class SnapPackage:
         try:
             return store_channels[self.channel]['confinement'] == 'classic'
         except KeyError:
+            # We have seen some KeyError issues when running tests that are
+            # hard to debug as they only occur there, logging in debug mode
+            # will help uncover the root cause if it happens again.
             logger.debug('Current store channels are {!r} and the store'
                          'payload is {!r}'.format(store_channels,
                                                   self._store_snap_info))
--- a/snapcraft/internal/repo/snaps.py
+++ b/snapcraft/internal/repo/snaps.py
@@ -76,7 +76,10 @@ class SnapPackage:
     @property
     def in_store(self):
         if self._is_in_store is None:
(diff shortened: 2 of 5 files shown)


sergiusens:
This is mostly a non invasive change; I will think about using exceptions more broadly next year and how to fit that into the grammar.

sergiusens:
I don't understand how this failed as I ran static before committing :-/

kyrofa:
> I don't understand how this failed as I ran static before committing :-/

Is it possible you have a newer version of mypy installed than the one in our dependencies? We should update that...

snapcraft/internal/repo/errors.py
```diff
@@ -81,6 +81,16 @@ def __init__(self, package):
         super().__init__(package=package)
 
 
+class SnapUnavailableError(RepoError):
+
+    fmt = ('Failed to install or refresh a snap: {snap_name!r} does not exist '
+           'or is not available on the desired channel {snap_channel!r}. '
+           'Verify that the snap is available.')
```
kyrofa: Perhaps "use `snap info {snap_name}` to list available channels" or something that similarly directs them to how to verify that it's available?

sergiusens:
@kyrofa no, it was me pressing the space bar on the editor by mistake while committing :confused:

snapcraft/internal/repo/snaps.py
```diff
@@ -89,12 +92,25 @@ def get_local_snap_info(self):
...
     def get_store_snap_info(self):
-        """Returns a store payload for the snap.
-
-        Validity of the results are determined by checking self.remote."""
+        """Returns a store payload for the snap."""
         if self._is_in_store is None:
-            with contextlib.suppress(exceptions.HTTPError):
-                self._store_snap_info = _get_store_snap_info(self.name)
+            retry_count = 5
+            while retry_count > 0:
```
kyrofa: Why the retry loop, are we seeing a lot of timeouts? Can you add a comment in the code as well?

snapcraft/internal/repo/snaps.py
```diff
@@ -116,7 +132,13 @@ def get_current_channel(self):
 
     def is_classic(self):
         store_channels = self._get_store_channels()
-        return store_channels[self.channel]['confinement'] == 'classic'
+        try:
+            return store_channels[self.channel]['confinement'] == 'classic'
+        except KeyError:
+            logger.debug('Current store channels are {!r} and the store'
```
kyrofa: Does this ever happen when it isn't a legit error? Might be helpful to actually print this like an error (but formatted a little better), which would be nice for determining the channels that _are_ available in case they mistyped one or the snap isn't released into that channel.

snapcraft/tests/integration/general/test_build_snaps.py
```diff
@@ -45,3 +47,26 @@ def test_build_snap(self):
         self.run_snapcraft('build')
         self.assertTrue(
             snapcraft.repo.snaps.SnapPackage.is_snap_installed(self.snap))
+
+
+class BuildSnapsErrorsTestCase(integration.TestCase):
+
+    def test_inexistent_build_snap(self):
```
kyrofa: I'd also like to see a test for a snap that exists, but is unavailable in the given channel.

snapcraft/internal/repo/snaps.py
sergiusens: on armhf adt tests none else ;-)

snapcraft/internal/repo/snaps.py
sergiusens: we don't know why this happens, we see it in the armhf tests, this PR just started as printing this log then I discovered something was wrong with the logic that verifies we were doing the right thing (as was mentioned in part of the docstring).

snapcraft/tests/integration/general/test_build_snaps.py
sergiusens: oh, that is a much harder one :-)

sergiusens:
@kyrofa I believe to have addressed all the issues, please take another look

snapcraft/tests/integration/general/test_build_snaps.py
```diff
@@ -45,3 +47,53 @@ def test_build_snap(self):
...
+        # If the snap tested here does not exist, then BuildSnapsTestCase
+        # will fail.
+        if os.environ.get('ADT_TEST') and self.deb_arch == 'armhf':
+            self.expectFailure('The autopkgtest armhf runners cannot '
+                               'install snaps')
+        snapcraft_yaml = fixture_setup.SnapcraftYaml(self.path)
+        snapcraft_yaml.update_part(
+            'test-part-with-build-snap', {
+                'plugin': 'nil',
+                'build-snaps': ['u1test-snap-with-tracks/not-exists/candidate']
```
kyrofa: Hey, there you go.

Issue #1686, opened 2017-11-06
Support strings in grammar


sergiusens:
Estimate: 1 week

Snapcraft currently has a grammar that allows for behavior to change based on, for example, the host architecture. The use of this grammar is slowly making its way across the snapcraft.yaml. Currently the only items supported are lists (e.g. stage-packages and build-packages). Adding support for strings is required to support further expansion (e.g. making the source key grammar-powered).

sergiusens:
This was not needed as the change was simple enough to support sources

om26er:
> Currently the only items supported are lists (e.g. stage-packages and build-packages)

@sergiusens Does that mean, its already possible to have different stage-packages based on the architecture ? -- Can you please point to any examples for that ?

my usecase: I want to stage some 32bit packages on my 64bit snap. Currently those packages are also being shipped with 32bit snap, which doesn't need them.

sergiusens:
On Dec 21 2017, at 6:04 pm, Omer Akram <notifications@github.com> wrote:
> >
> > Currently the only items supported are lists (e.g. stage-packages and build-packages)
> >
>
>
> @sergiusens (https://github.com/sergiusens) Does that mean, its already possible to have different stage-packages based on the architecture ? -- Can you please point to any examples for that ?
>
>
> my usecase: I want to stage some 32bit packages on my 64bit snap. Currently those packages are also being shipped with 32bit snap, which doesn't need them.
>
>

It is all documented here https://docs.snapcraft.io/build-snaps/syntax

Issue #1687, opened 2017-11-06
Make source grammar-powered


sergiusens:
Estimate: 1 day

Snapcraft currently has a grammar that allows for behavior to change based on, for example, the host architecture. The use of this grammar is slowly making its way across the snapcraft.yaml. Currently the only items supported are lists (e.g. stage-packages and build-packages). Adding support for strings is required to support further expansion (e.g. making the source key grammar-powered).

PR #1817, opened 2017-12-19
grammar: support strings

kalikiana:

-----
This adds generic support for processing grammar of string values. `source` is the first property to support it, although adding more should be just a matter of updating the schema.

Closes: #1686

### DIFF
--- a/snapcraft/internal/pluginhandler/__init__.py
+++ b/snapcraft/internal/pluginhandler/__init__.py
@@ -66,6 +66,7 @@ class PluginHandler:
         self.plugin = plugin
         self._part_properties = _expand_part_properties(
             part_properties, part_schema)
+        self._part_property_grammar = part_properties
         self.stage_packages = []
         self._stage_packages_repo = stage_packages_repo
         self._grammar_processor = grammar_processor
@@ -96,10 +97,14 @@ class PluginHandler:
         # the default '.'
         source_handler = None
         if properties['source']:
+            if 'source' in self._part_property_grammar:
+                source = self._grammar_processor.get_source()
+            else:
+                source = properties['source']
             handler_class = sources.get_source_handler(
-                properties['source'], source_type=properties['source-type'])
+                source, source_type=properties['source-type'])
             source_handler = handler_class(
-                properties['source'],
+                source,
                 self.plugin.sourcedir,
                 source_checksum=properties['source-checksum'],
                 source_branch=properties['source-branch'],
--- a/snapcraft/internal/pluginhandler/__init__.py
+++ b/snapcraft/internal/pluginhandler/__init__.py
@@ -66,11 +66,13 @@ class PluginHandler:
         self.plugin = plugin
         self._part_properties = _expand_part_properties(
             part_properties, part_schema)
-        self._part_property_grammar = part_properties
         self.stage_packages = []
         self._stage_packages_repo = stage_packages_repo
         self._grammar_processor = grammar_processor
         self._confinement = confinement
+        self._source = grammar_processor.get_source()
+        if not self._source:
+            self._source = part_schema['source'].get('default')
 
         self._project_options = project_options
         self.deps = []
@@ -96,15 +98,11 @@ class PluginHandler:
         # TODO: we cannot pop source as it is used by plugins. We also make
         # the default '.'
         source_handler = None
-        if properties['source']:
-            if 'source' in self._part_property_grammar:
-                source = self._grammar_processor.get_source()
-            else:
-                source = properties['source']
+        if self._source:
             handler_class = sources.get_source_handler(
-                source, source_type=properties['source-type'])
+                self._source, source_type=properties['source-type'])
             source_handler = handler_class(
-                source,
+                self._source,
                 self.plugin.sourcedir,
                 source_checksum=properties['source-checksum'],
                 source_branch=properties['source-branch'],
--- a/snapcraft/internal/project_loader/_parts_config.py
+++ b/snapcraft/internal/project_loader/_parts_config.py
@@ -187,7 +187,7 @@ class PartsConfig:
 
         part = pluginhandler.PluginHandler(
             plugin=plugin,
-            part_properties=grammar_processor.get_properties(),
+            part_properties=part_properties,
             project_options=self._project_options,
             part_schema=self._validator.part_schema,
             definitions_schema=self._validator.definitions_schema,
--- a/snapcraft/internal/project_loader/_parts_config.py
+++ b/snapcraft/internal/project_loader/_parts_config.py
@@ -187,7 +187,7 @@ class PartsConfig:
 
         part = pluginhandler.PluginHandler(
             plugin=plugin,
-            part_properties=part_properties,
+            part_properties=grammar_processor.get_properties(),
             project_options=self._project_options,
             part_schema=self._validator.part_schema,
             definitions_schema=self._validator.definitions_schema,
--- a/snapcraft/internal/project_loader/grammar/_processor.py
+++ b/snapcraft/internal/project_loader/grammar/_processor.py
@@ -73,7 +73,11 @@ def _parse_dict(section, statement, statements, project_options,
     from ._try import TryStatement
 
     for key, value in section.items():
-        value = value if isinstance(value, list) else {value}
+        # Grammar is always written as a list of selectors but the value can
+        # be a list or a string, in the latter case we wrap it so no special
+        # care needs to be taken when fetching the result from the primitive
+        if not isinstance(value, list):
+            value = {value}
 
         if _ON_CLAUSE_PATTERN.match(key):
             # We've come across the beginning of an 'on' statement.
--- a/snapcraft/internal/project_loader/grammar/_processor.py
+++ b/snapcraft/internal/project_loader/grammar/_processor.py
@@ -73,6 +73,8 @@ def _parse_dict(section, statement, statements, project_options,
     from ._try import TryStatement
 
     for key, value in section.items():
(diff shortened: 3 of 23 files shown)


snapcraft/internal/project_loader/grammar_processing/_part_grammar_processor.py
```diff
@@ -58,6 +58,12 @@ def __init__(self, *, plugin, properties, project_options, repo):
         self._project_options = project_options
         self._repo = repo
 
+        for key, value in properties.items():
+            if key in ['source']:
```
sergiusens: is this just to support future expansion? I wouldn't complicate it now unless we know there is a very close next candidate to add.

snapcraft/tests/integration/general/test_parts_grammar.py
```diff
@@ -0,0 +1,62 @@
...
+from snapcraft.tests import integration, fixture_setup
+from testtools.matchers import Contains
+
+
+class PartsGrammarTestCase(integration.TestCase):
+
+    def setUp(self):
+        super().setUp()
+
+    def test_source(self):
```
sergiusens: a better name might be `test_plain_source_string`

snapcraft/internal/project_loader/grammar_processing/_part_grammar_processor.py
kalikiana: The next candidate is `source-checksum` which I think is paramount to have for different sources depending on the architecture.

snapcraft/internal/project_loader/grammar/_processor.py
```diff
@@ -73,6 +73,8 @@ def _parse_dict(section, statement, statements, project_options,
     from ._try import TryStatement
 
     for key, value in section.items():
+        value = value if isinstance(value, list) else {value}
```
kyrofa: I was thinking we were going to need to rewrite some of the grammar to support strings, but treating them as iterables is clever. However, I'm not a huge fan of the inline if statements-- would you mind rewriting this to be a bit more verbose? Adding a comment about why this is required would also be useful in the future.

snapcraft/internal/project_loader/grammar_processing/_part_grammar_processor.py
```diff
@@ -58,6 +58,12 @@ def __init__(self, *, plugin, properties, project_options, repo):
         self._project_options = project_options
         self._repo = repo
 
+        for key, value in properties.items():
```
kyrofa: Actually modifying the properties here instead of providing accessor methods to get the processed grammar doesn't fit into the pattern established in this class and may have further ramifications. Can you explain why you chose this path?

sergiusens:
I am good but please take into account @kyrofa's comments and review.

snapcraft/internal/project_loader/grammar/_processor.py
kalikiana: Done.

snapcraft/internal/project_loader/grammar_processing/_part_grammar_processor.py
kalikiana: I was under the impression we'd discussed this and you said wasn't necessary to wrap the properties, but I think I got something wrong. So I went ahead and added an explicit get_properties() method now. Sorry for the mix-up.

snapcraft/internal/project_loader/grammar_processing/_part_grammar_processor.py
```diff
@@ -58,11 +58,8 @@ def __init__(self, *, plugin, properties, project_options, repo):
         self._project_options = project_options
         self._repo = repo
 
-        for key, value in properties.items():
-            if key in ['source']:
-                properties[key] = next(iter(grammar.process_grammar(
-                    value if isinstance(value, list) else {value},
-                    self._project_options, True)))
+        self._property_grammar = properties
```
sergiusens: this should probably be a copy or the result is almost the same.

snapcraft/internal/project_loader/grammar_processing/_part_grammar_processor.py
```diff
@@ -73,6 +70,16 @@ def __init__(self, *, plugin, properties, project_options, repo):
...
 
+    def get_properties(self):
+        if not self.__properties:
+            self.__properties = self._property_grammar
+            for key, value in self._property_grammar.items():
+                if key in ['source']:
+                    self.__properties[key] = next(iter(grammar.process_grammar(
+                        value if isinstance(value, list) else {value},
+                        self._project_options, True)))
+        return self.__properties
```
sergiusens: ok, a copy might not be necessary; but you are not adding the original ones that don't need parsing to self.__properties; is that desired?

snapcraft/internal/project_loader/grammar_processing/_part_grammar_processor.py
```diff
@@ -52,6 +52,23 @@ class PartGrammarProcessor:
...
+    >>> plugin = mock.Mock()
+    >>> plugin.properties = {'plugin': 'dump',
+    ...                      'source': [{'on amd64': '.'}, 'else fail']}
+    >>> processor = PartGrammarProcessor(
+    ...    plugin=plugin,
+    ...    properties=plugin.properties,
+    ...    project_options=snapcraft.ProjectOptions(),
+    ...    repo=repo)
+    >>> processor.get_properties()['source']
+    '.'
```
sergiusens: can you also add `processor.get_properties()['plugin']` to verify it hasn't been modified? (add a code comment mentioning that)

snapcraft/tests/unit/project_loader/grammar_processing/test_part_grammar_processor.py
```diff
@@ -14,13 +14,58 @@
...
+            'target_arch': 'amd64',
+            'expected': {'source': 'foo'}
+        }),
+        ('on i386', {
+            'properties': {'source': [{'on i386': 'foo'}, {'else': 'bar'}]},
+            'target_arch': 'amd64',
+            'expected': {'source': 'bar'}
+        }),
+        ('on i386, target_arch=i386', {
+            'properties': {'source': [{'on i386': 'foo'}, {'else': 'bar'}]},
```
sergiusens: can you add more values than just the ones that need modification to verify the remain intact?

snapcraft/internal/project_loader/grammar_processing/_part_grammar_processor.py
kalikiana: The original values are assigned initially before the loop. However, you were right suspecting we need to copy, so I added `.copy()` now and a test case to see that `plugin.properties` retains the original properties including grammar (and crucially the test itself needs to `.copy()` so it can verify). Thanks for catching that!

snapcraft/internal/project_loader/_parts_config.py
```diff
@@ -187,7 +187,7 @@ def load_part(self, part_name, plugin_name, part_properties):
 
         part = pluginhandler.PluginHandler(
             plugin=plugin,
-            part_properties=part_properties,
+            part_properties=grammar_processor.get_properties(),
```
kyrofa: This is still going to mess up state tracking. The properties here need to be the raw ones. Note how we pass the grammar processor to the PluginHandler anyway-- it should just use the processor to grab the source like it does for e.g. stage-packages.

snapcraft/internal/project_loader/grammar_processing/_part_grammar_processor.py
kyrofa: Combined with my comment above, I think each property we want grammar-enabled should have its own getter, like stage-packages etc. here. Not the entire set of properties.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -96,10 +97,14 @@ def _get_source_handler(self, properties):
         # the default '.'
         source_handler = None
         if properties['source']:
+            if 'source' in self._part_property_grammar:
+                source = self._grammar_processor.get_source()
```
kyrofa: Rather than saving off the grammar before defaults are applied, could you not just call this function to see if it returns a source? If not, use the default.

sergiusens: can you expand or maybe just fix it? :slightly_smiling_face:

sergiusens:
This actually solves #1687

snapcraft/internal/project_loader/grammar_processing/_part_grammar_processor.py
```diff
@@ -67,6 +87,14 @@ def __init__(self, *, plugin, properties, project_options, repo):
         self._stage_package_grammar = getattr(plugin, 'stage_packages', [])
         self.__stage_packages = set()
 
+    def get_source(self):
+        if not self.__source:
+            value = self._property_grammar['source']
+            self.__source = next(iter(grammar.process_grammar(
+                value if isinstance(value, list) else {value},
+                self._project_options, True)))
```
kyrofa: There are no tests using `try`. They would fail. This needs to be a callable, as the docs say.

snapcraft/internal/pluginhandler/__init__.py
kyrofa: Yep, on it.

kyrofa: Fixed.

snapcraft/internal/project_loader/grammar_processing/_part_grammar_processor.py
kyrofa: Fixed.

PR #1822, opened 2017-12-20
tests: skip classic confined tests on armhf

sergiusens:
We recently added a requirement to the snapcraft deb to install patchelf
from the snapstore which disabled a couple of tests.

-----

### DIFF
--- a/snaps_tests/__init__.py
+++ b/snaps_tests/__init__.py
@@ -93,8 +93,6 @@ def _get_latest_ssh_private_key():
 
 class SnapsTestCase(testtools.TestCase):
 
-    snap_content_dir = ''
-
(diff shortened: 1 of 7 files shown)


snaps_tests/__init__.py
```diff
@@ -93,7 +93,7 @@ def _get_latest_ssh_private_key():
 
 class SnapsTestCase(testtools.TestCase):
 
-    snap_content_dir = None
+    snap_content_dir = ''
```
kalikiana: Why are you changing this here? Isn't it `None` so that tests which don't set it fail?

kalikiana: Ah. Actually removing it is even better. Nice.

Issue #1664, opened 2017-11-06
Search prime for DT_NEEDED and add rpath


sergiusens:

sergiusens:
We have determined that we don't want to do all this on one snatch.

PR #1823, opened 2017-12-21
elf: search for DT_NEEDED in prime and the base.

sergiusens:

-----
Closes: #1664

### DIFF
--- a/snapcraft/internal/elf.py
+++ b/snapcraft/internal/elf.py
@@ -14,8 +14,10 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 import contextlib
+import glob
 import logging
 import os
+import re
 import subprocess
 import sys
 from functools import lru_cache, wraps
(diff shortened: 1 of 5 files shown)


snapcraft/tests/unit/test_elf.py
```diff
@@ -379,7 +401,8 @@ def test_patch_fails_raises_patcherror_exception(self, check_call_mock):
                       'dynamically linked, interpreter '
                       '/lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32')
         elf_file = elf.ElfFile(path='/fake-elf', magic=stub_magic)
-        elf_patcher = elf.Patcher(dynamic_linker='/lib/fake-ld')
+        elf_patcher = elf.Patcher(dynamic_linker='/lib/fake-ld',
+                                  base_path='/random')
```
kalikiana: Is this intentionally not `self.prime_dir`? If so, can you add a comment to the test to explain why?

snapcraft/internal/elf.py
```diff
@@ -272,6 +287,22 @@ def _get_system_libs() -> FrozenSet[str]:
...
+        raise RuntimeError('Cannot load magic header detection')
+
+    fs_encoding = sys.getfilesystemencoding()
+    path_b = path.encode(
+        fs_encoding, errors='surrogateescape')  # type: bytes
+    return ms.file(path_b)
+
+
+def _is_elf(file_m: str) -> bool:
+    return file_m.startswith('ELF') and 'dynamically linked' in file_m
```
kalikiana: This strikes me as wrong. I read the function name as "this is an elf file" but it also checks that it's dynamically linked... and it's a bit unexpected to pass a string rather than a filename.
In both cases where it's used right now we want to know if a file is a dynamic elf library. If that's all we want here, I'd find a single function like `_file_is_dynamically_linked_elf` that takes the path a lot clearer. Thoughts?

sergiusens: this means we would need to read the magic hdr twice as we also need the magic header.

snapcraft/tests/unit/test_elf.py
sergiusens: easy fix, but it is irrelevant as there are no files in there.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -443,8 +443,13 @@ def prime(self, force=False) -> None:
 
         elf_files = elf.get_elf_files(self.primedir, snap_files)
         all_dependencies = set()
+        # TODO: base snap support
+        core_path = common.get_core_path()
```
kalikiana: Do you have a plan for this ie. how base snaps will be handled? Will the core snap be replaced? self.primedir is not what I'd expect to pass as a "base" or at least ambiguous. So maybe something like base_folder_path=self.primedir, base_snap_path=core_path - ideally another term than base but this would be more explicit.

snapcraft/internal/elf.py
kalikiana: Fair point. I guess I could live with `_is_dynamically_linked_elf` taking the magic...maybe `..._header` as a suffix? Even better would be having a class wrapping it, but that seems like overkill here (unless you feel like adding that).

snapcraft/internal/pluginhandler/__init__.py
sergiusens: I am leaving the complex core_base_path as it is technically correct. multi base I will discuss with Gustavo live in January.

I have however renamed `base_path` to `root_path`

snapcraft/internal/elf.py
```diff
@@ -36,6 +36,49 @@
...
+        else:
+            self.system_lib = False
+
+        if path.startswith(core_base_path):
+            self.in_base_snap = True
+        else:
+            self.in_base_snap = False
+
+
+@lru_cache()
```
kyrofa: This is new to me, pretty sweet! This doesn't really seem to fit into the "most recent calls are the best predictors of upcoming calls" mold, but I doubt this hurts. Would be interesting to see a benchmark.

sergiusens: they probably aren't but given the size of the default cache (128) I thought it was good enough (better than constant calls).

PR #1825, opened 2017-12-21
metadata: add infrastructure for extracting metadata from parts

### LINKED ISSUE #1693: Develop metadata handler system

kyrofa:

-----

This PR resolves #1693 by adding the underlying infrastructure necessary to extract metadata from parts. No actual metadata extraction is done here (e.g. Appstream). That will come in follow-up PRs, and will be as simple as adding a new module to `snapcraft.extractors` with an `extract` function.

### DIFF
--- a/schema/snapcraft.yaml
+++ b/schema/snapcraft.yaml
@@ -63,6 +63,9 @@ properties:
   build-packages:
     $ref: "#/definitions/grammar-array"
     description: top level build packages.
+  adopt-info:
+    type: string
+    description: name of the part that provides source files that will be parsed to extract snap metadata information
   name:
     type: string
     description: name of the snap package
@@ -390,6 +393,13 @@ properties:
           prepare:
             type: string
             default: ''
+          parse-info:
+            type: array
+            minitems: 1
+            uniqueItems: true
+            items:
+              type: string
+            default: []
   plugs:
     type: object
   slots:
@@ -397,9 +407,16 @@ properties:
 required:
   - name
   - version
-  - summary
-  - description
   - parts
+
+# Either summary/description is required, or adopt-info is required to specify
+# the part from which this metadata will be retrieved.
+anyOf:
+  - required:
+    - summary
+    - description
+  - required:
+    - adopt-info
 dependencies:
   license-agreement: ["license"]
   license-version: ["license"]
--- a/snapcraft/extractors/__init__.py
+++ b/snapcraft/extractors/__init__.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright (C) 2016 Canonical Ltd
+# Copyright (C) 2017 Canonical Ltd
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License version 3 as
@@ -14,6 +14,5 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-from snapcraft.internal import cache             # noqa
-from snapcraft.internal import deltas            # noqa
-from snapcraft.internal import states            # noqa
+from ._metadata import ExtractedMetadata  # noqa
+from ._errors import UnhandledFileError  # noqa
--- a/snapcraft/extractors/_metadata.py
+++ b/snapcraft/extractors/_metadata.py
@@ -1,0 +1,78 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2017 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import yaml
+from typing import Any, Dict
+
+
+class ExtractedMetadata(yaml.YAMLObject):
+    """Collection of metadata extracted from a part."""
+
+    yaml_tag = u'!ExtractedMetadata'
+
+    def __init__(self, *, summary='', description='') -> None:
+        """Create a new ExtractedMetadata instance.
+
+        :param str summary: Extracted summary
+        :param str description: Extracted description
+        """
+
+        self._data = {}  # type: Dict[str, str]
+
+        if summary:
+            self._data['summary'] = summary
+        if description:
+            self._data['description'] = description
+
+    def update(self, other: 'ExtractedMetadata') -> None:
+        """Update this metadata with other metadata.
+
+        Note that the other metadata will take precedence, and may overwrite
+        data contained here.
+
+        :param ExtractedMetadata other: Metadata from which to update
+        """
+        self._data.update(other.to_dict())
+
+    def get_summary(self) -> str:
+        """Return extracted summary.
+
+        :returns: Extracted summary
+        :rtype: str
+        """
+        return self._data.get('summary')
+
+    def get_description(self) -> str:
+        """Return extracted description.
+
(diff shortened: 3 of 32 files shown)


schema/snapcraft.yaml
```diff
@@ -361,16 +364,30 @@ properties:
...
   - parts
+
+# Either summary/description is required, or adopt-info is required to specify
+# the part from which this metadata will be retrieved.
+anyOf:
+  - required:
+    - summary
+    - description
+  - required:
+    - adopt-info
```
sergiusens: nice

snapcraft/extractors/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from ._metadata import ExtractedMetadata  # noqa
+from ._errors import UnhandledFileError  # noqa
```
sergiusens: I wouldn't import this, only if it is public (which I wouldn't be against)

sergiusens: err, abort!

snapcraft/extractors/_metadata.py
```diff
@@ -0,0 +1,78 @@
...
+        :param str summary: Extracted summary
+        :param str description: Extracted description
+        """
+
+        self._data = {}  # type: Dict[str, str]
+
+        if summary:
+            self._data['summary'] = summary
+        if description:
+            self._data['description'] = description
```
sergiusens: what is the story with blobs like icon and desktop file?

snapcraft/internal/lifecycle/_runner.py
```diff
@@ -58,7 +59,7 @@ def execute(step, project_options, part_names=None):
                           over.
     :returns: A dict with the snap name, version, type and architectures.
     """
-    config = snapcraft.internal.load_config(project_options)
+    config = project_loader.load_config(project_options)
```
sergiusens: might as well consider unimporting `snapcraft`?

sergiusens: and also `snapcraft.internal`

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -90,6 +97,26 @@ def __init__(self, *, plugin, part_properties, project_options,
 
         self._migrate_state_file()
 
+    def pull_state(self) -> states.PullState:
```
sergiusens: these should have a `get` in front

sergiusens: maybe this really merits a mixin as well? Thoughts?

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -334,9 +372,40 @@ def mark_build_done(self):
         plugin_manifest = self.plugin.get_manifest()
         machine_manifest = self._get_machine_manifest()
 
+        # Extract any requested metadata available in the build directory,
+        # followed by the install directory (which takes precedence)
```
sergiusens: \o/

snapcraft/internal/pluginhandler/_metadata_extraction.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def extract_metadata(part_name: str,
+                     file_path: str) -> extractors.ExtractedMetadata:
+    if not os.path.exists(file_path):
+        raise MissingMetadataFileError(part_name, file_path)
+
+    # Iterate through each extractor module, calling the 'extract' function
+    # from it. If it raises an 'UnhandledFileError' move onto the next.
+    for _, module_name, _ in pkgutil.iter_modules(
```
sergiusens: this is clever, I am just afraid we might have some false positives
it would be nice to be able to infer some preference given the file name as well

snapcraft/tests/unit/project_loader/test_config.py
```diff
@@ -1916,9 +1893,11 @@ def setUp(self):
             'version': '1.0-snapcraft1~ppa1',
             'summary': 'my summary less that 79 chars',
             'description': 'description which can be pretty long',
+            'adopt-info': 'part1',
```
sergiusens: why is this necessary?

snapcraft/tests/unit/project_loader/test_config.py
```diff
@@ -2038,6 +2017,20 @@ def test_both_snap_and_prime_specified(self):
...
+    def test_missing_required_property_and_missing_adopt_info(self):
+        del self.data['summary']
+        del self.data['adopt-info']
+
+        raised = self.assertRaises(
+            errors.YamlValidationError,
+            project_loader.Validator(self.data).validate)
+
+        expected_message = (
+            "'adopt-info' is a required property or 'summary' is a required "
```
sergiusens: now that I see the message here, it feels rather off.

snapcraft/tests/unit/test_errors.py
```diff
@@ -286,7 +287,51 @@ class ErrorFormattingTestCase(unit.TestCase):
             'exception': errors.InvalidContainerImageInfoError,
             'kwargs': {'image_info': 'test-image-info'},
             'expected_message': (
-                'Error parsing the container image info: test-image-info')})
+                'Error parsing the container image info: test-image-info')}),
+        # meta errors.
```
sergiusens: nice but are we bundling ALL errors here? This is just rhetoric ;-)

snapcraft/tests/unit/project_loader/test_config.py
kyrofa: Mannnn, this took me hours. Would you prefer what jsonschema gives us?

> Issues while validating snapcraft.yaml: OrderedDict([('name', 'my-snap-name'), ('version', '0.1'), ('description', "This is my-snap's description. You have a paragraph or two to tell the\nmost important story about your snap. Keep it under 100 words though,\nwe live in tweetspace and your description wants to look good in the snap\nstore.\n"), ('grade', 'devel'), ('confinement', 'devmode'), ('parts', OrderedDict([('my-part', OrderedDict([('plugin', 'nil'), ('parse-info', ['foo'])]))]))]) is not valid under any of the given schemas

Just kidding. I'd love to improve it, but couldn't figure out how to do so in a generic manner. Should I hard-code handling for this specific case?

snapcraft/tests/unit/project_loader/test_config.py
kyrofa: Just so I don't have to explicitly add it in the test cases. Harmless for the other ones.

snapcraft/internal/pluginhandler/_metadata_extraction.py
kyrofa: Yeah I was thinking about preference as well, although I couldn't think of any extractors that overlap-- do we ever anticipate that multiple extractors could handle the same file? If not, no preference is needed.

snapcraft/extractors/_metadata.py
kyrofa: As we briefly discussed, we can base64 them, or save file paths here and grab them when creating the meta/ dir.

snapcraft/internal/pluginhandler/__init__.py
kyrofa: Oh right-- I'll fix the lack of "get". Regarding the mixin, I feel it's only worth it if we want the functionality mixed into another class as well. I don't think we do... do we?

snapcraft/internal/lifecycle/_runner.py
kyrofa: We use `snapcraft`, but you're right, `snapcraft.internal` does not seem necessary. Wonder why flake8 didn't grab that. Fixed.

snapcraft/internal/pluginhandler/__init__.py
kyrofa: Fixed.

PR #1831, opened 2017-12-22
cli: add expiration option to export-login

kyrofa:

-----

With this PR one can e.g. call `snapcraft export-login --expires=2018-01-01T00:00:00` and export a login that is only valid until the first of the year.

### DIFF
--- a/manual-tests.md
+++ b/manual-tests.md
@@ -247,3 +247,14 @@
 
     * Check that it should end ok
     * Check in the Web that the icon is now what the YAML says
+
+
+# Test creating a macaroon with a specific expiration
+
+1. Take note of the current date. Pick a date two days later, and convert it
+   into ISO 8601. For example, midnight on January 1st, 2019 is
+   '2019-01-01T00:00:00'.
+2. Run 'snapcraft export-login --expires="2019-01-01T00:00:00" exported', using
+   the expiration date you calculated.
+3. Snapcraft will print the capabilities of the exported login. Verify that the
+   expiration date you requested is the one that it prints.
--- a/manual-tests.md
+++ b/manual-tests.md
@@ -252,9 +252,9 @@
 # Test creating a macaroon with a specific expiration
 
 1. Take note of the current date. Pick a date two days later, and convert it
-   into ISO 8601. For example, midnight on January 1st, 2019 is
-   '2019-01-01T00:00:00'.
-2. Run 'snapcraft export-login --expires="2019-01-01T00:00:00" exported', using
+   into ISO 8601. For example, midnight on February 5th, 2019 is
+   '2019-02-05T00:00:00'.
+2. Run 'snapcraft export-login --expires="2019-02-05T00:00:00" exported', using
    the expiration date you calculated.
 3. Snapcraft will print the capabilities of the exported login. Verify that the
    expiration date you requested is the one that it prints.
--- a/snapcraft/cli/store.py
+++ b/snapcraft/cli/store.py
@@ -13,6 +13,7 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+from dateutil import parser
 import os
 import sys
 from textwrap import dedent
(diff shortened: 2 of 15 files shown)


snapcraft/cli/store.py
```diff
@@ -76,10 +77,14 @@ def _human_readable_acls(store: storeapi.StoreClient) -> str:
         if not acl[key]:
             human_readable_acl[key] = 'No restriction'
 
+    expiration_date = parser.parse(str(acl['expires']))
```
sergiusens: can't this be solve with `datetime.strptime` like in `snapcraft.cli.assertions`?

kyrofa: strptime [doesn't handle 8601 timezones](https://stackoverflow.com/questions/969285/how-do-i-translate-a-iso-8601-datetime-string-into-a-python-datetime-object).

kyrofa:
> I would rather have a strict format on the expiration time for a macaroon.

Fair enough, although I find ISO 8601 not to be the most human-readable/friendly format. Especially when you start talking timezones. Take a look at `snapcraft status <snap>` or `snapcraft list-revisions <snap>`; I have trouble figuring out what datetimes those actually are. It also seems that you must always include the time, even if you just want zeros, and if you get it wrong you get an "invalid login" response, so we'll probably need to validate it. Is that the format you want to use?

sergiusens:
Let's keep it ISO 8601, this is a very corner case-y feature for the security minded folks

kyrofa:
> Let's keep it ISO 8601, this is a very corner case-y feature for the security minded folks

Very well, I've updated the PR to reflect this desire. **dateutil** is no longer a dependency. Note that, since we're potentially passing invalid expiration times to the store, I needed to refactor some store error handling in order to actually surface the store's validation errors to the user so invalid expiration dates didn't look like incorrect passwords. It's a little closer to how the rest of Snapcraft works, now, which is nice.

snapcraft/cli/store.py
```diff
@@ -295,11 +301,15 @@ def export_login(login_file: TextIO, snaps: str, channels: str, acls: str):
...
 
-        snapcraft export-login --channels=edge
+        snapcraft export-login --channels=edge exported
 
     Or to limit access to only the edge channel of a single snap:
 
-        snapcraft export-login --snaps=my-snap --channels=edge
+        snapcraft export-login --snaps=my-snap --channels=edge exported
+
+    To limit access to a single snap, but only until the first of the year:
```
elopio: first *day*.

I think it would be clearer to use the last day of the year in the example, but that's just a nit.

kyrofa:
> What do you think about adding a manual test that sets the expiration date a few minutes after now, and then checks that the macaroon no longer works?

Good idea, although I think that's testing the store's behavior, not Snapcraft's. As long as the expiration date we get back from the store is the one we requested, I think Snapcraft did its job. So just verifying that the correct expiration date it printed is good enough, I think. Thoughts?

snapcraft/cli/store.py
kyrofa: Well, this is midnight of the first day of the new year, which makes sense, right? I did reword to say "only until 2019" and then updated it to use midnight of the first day of 2019.

kyrofa:
(I added a manual test).

manual-tests.md
```diff
@@ -247,3 +247,14 @@
 
     * Check that it should end ok
     * Check in the Web that the icon is now what the YAML says
+
+
+# Test creating a macaroon with a specific expiration
+
+1. Take note of the current date. Pick a date two days later, and convert it
+   into ISO 8601. For example, midnight on January 1st, 2019 is
+   '2019-01-01T00:00:00'.
```
kalikiana: How about `midnight on February 5th, 2019 is 2019-02-05T00:00:00` to also demonstrate that the month comes before the day? The order is reversed in many traditional formats.

snapcraft/storeapi/errors.py
```diff
@@ -84,9 +91,22 @@ def __init__(self, path, expected_sha):
...
+    fmt = 'Authentication error: {message}.'
+
+    def __init__(self, message, response=None):
+        # Unfortunately the store doesn't give us a consistent error response,
+        # so we'll check the ones of which we're aware.
+        if response:
+            response_json = response.json()
+            extra_error_message = ''
+            if 'error_message' in response_json:
+                extra_error_message = response_json['error_message']
```
kalikiana: Can we make the fake server behave this way so we can unit test it?

manual-tests.md
kyrofa: Indeed, that's probably more helpful. Done.

elopio:
The test is good to me, thanks for adding it.

elopio:
There are two store integration tests failing.

Issue #1667, opened 2017-11-06
Walk the prime directory for elf files and flag the glibc required version


sergiusens:
Estimate: 2 days

sergiusens:
Options:
- `readelf -V`
- `readelf --dyn-syms`
- `objdump --dynamic-syms`

kyrofa:
Another interesting one: `readelf -d prime/bin/example | grep NEEDED`

PR #1843, opened 2018-01-03
elf: warn if primed files will not work with the base's linker

sergiusens:
Parse all the primed files and make sure it will work with
the version of the linker provided by the supporting base.

Closes: #1667

-----

### DIFF
--- a/snapcraft/internal/elf.py
+++ b/snapcraft/internal/elf.py
@@ -24,6 +24,7 @@ from functools import lru_cache, wraps
 from typing import FrozenSet, List, Set, Sequence
 
 import magic
+from pkg_resources import parse_version
 
 from snapcraft.internal import (
     common,
@@ -34,6 +35,18 @@ from snapcraft.internal import (
 
 
 logger = logging.getLogger(__name__)
+
+
+class Symbol:
+    """Represents a symbol within an ELF file."""
+
+    def __init__(self, *, name: str, version: str, section: str) -> None:
+        self.name = name
+        self.version = version
+        self.section = section
+
+    def is_undefined(self) -> bool:
+        return self.section == 'UND'
 
 
 class Library:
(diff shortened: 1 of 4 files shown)


snapcraft/internal/elf.py
```diff
@@ -91,6 +104,65 @@ def __init__(self, *, path: str, magic: str) -> None:
...
+        symbols_section = subprocess.check_output([
+            'readelf', '--wide', '--dyn-syms', path]).decode()
+        # Regex inspired by the regex in lintian to match entries similar to
+        # number '1' from the following sample output
+        #
+        # Symbol table '.dynsym' contains 2281 entries:
+        #   Num:    Value          Size Type    Bind   Vis      Ndx Name
+        #     0: 0000000000000000     0 NOTYPE  LOCAL  DEFAULT  UND
+        #     1: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND endgrent@GLIBC_2.2.5 (2)  # noqa
+        symbol_match = (r'\s*\d+:\s*[0-9a-f]+\s*\d+\s*\S+\s*\S+\s*\S+\s*'
```
kyrofa: I suspect a few of the `\s*`s could be `\s+`s instead, which I think would be a better match. Otherwise it's possible for most of this regex to match what you didn't have in mind:

```
>>> re.match(r'[0-9a-f]+\s*\d+\s*\S+\s*\S+\s*\S+\s*', '0000000000000000')
<_sre.SRE_Match object; span=(0, 16), match='0000000000000000'>
```

snapcraft/internal/elf.py
```diff
@@ -91,6 +104,65 @@ def __init__(self, *, path: str, magic: str) -> None:
...
+                        r'(?P<section>\S+)\s*(?P<symbol>\S+).*\Z')
+        regex = re.compile(symbol_match)
+        for line in symbols_section.split('\n'):
+            m = regex.search(line)
+            if m:
+                section = m.group('section')
+                try:
+                    name, version = m.group('symbol').split('@')
+                except ValueError:
+                    name = m.group('symbol')
```
kyrofa: You're calling `m.group('symbol')` twice, potentially. It would be more efficient to just call it once before trying to split.

snapcraft/internal/elf.py
```diff
@@ -91,6 +104,65 @@ def __init__(self, *, path: str, magic: str) -> None:
...
+
+        The linker passed needs to be of the format <root>/ld-<X>.<Y>.so.
+        """
+        version_required = self.get_required_glibc()
+        m = re.search(r'ld-(?P<linker_version>[\d.]+).so$',
+                      os.path.basename(linker))
+        if not m:
+            # This is a programmatic error, we don't want to be friendly
+            # about this.
+            raise EnvironmentError('{!r} is incorrect.')
```
kyrofa: If you really didn't want to be friendly, you could just use `m` and let a "None has no attribute 'group'" error be thrown by Python itself. Since you're trying to be MORE friendly than that, I suggest a slightly more helpful exception message, perhaps '{!r} is not a valid linker path'? Also, you're probably missing a `.format()` call.

snapcraft/internal/elf.py
```diff
@@ -91,6 +104,65 @@ def __init__(self, *, path: str, magic: str) -> None:
...
+        r = parse_version(version_required) <= parse_version(linker_version)
+        logger.debug('Checking if linker {!r} will work with '
+                     'GLIBC_{}: {!r}'.format(linker, version_required, r))
+        return r
+
+    def get_required_glibc(self) -> str:
+        """Returns the required glibc version for this ELF file."""
+        version_required = ''
+        for symbol in self.symbols:
+            if symbol.section != 'UND':
```
kyrofa: Shouldn't this be using `symbol.is_undefined()` instead?

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -536,6 +536,20 @@ def prime(self, force=False) -> None:
...
+        # with that of the base.
+        # TODO the linker version depends on the chosen base, but that
+        # base may not be installed so we cannot depend on
+        # get_core_dynamic_linker to resolve the final path for which
+        # we resort to our only working base 16, ld-2.23.so.
+        linker_compatible = (e.is_linker_compatible(linker='ld-2.23.so')
+                             for e in elf_files)
+        if not all((x for x in linker_compatible)):
+            logger.warning('The primed files will not work with the current '
+                           'base given the GLIBC missmatch of the primed '
```
kyrofa: s/missmatch/mismatch/
  
Also, it would be useful to print the actual mismatched versions if possible.

snapcraft/internal/elf.py
sergiusens: doh, that is why I implemented that :-P

snapcraft/internal/elf.py
sergiusens: nice catch. I want it to be developer friendly, not user friendly :smile:

snapcraft/internal/pluginhandler/__init__.py
sergiusens: sure, that is another list comprehension to figure it out though

sergiusens:
thanks for the review, I will work on the changes.

snapcraft/internal/elf.py
sergiusens: done

snapcraft/internal/elf.py
kyrofa: Haha, I wondered!

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -545,9 +545,14 @@ def prime(self, force=False) -> None:
...
+            files = ('- {} -> GLIBC {}'.format(e.path, e.get_required_glibc())
+                     for e in elf_files if e.get_required_glibc())
             logger.warning('The primed files will not work with the current '
-                           'base given the GLIBC missmatch of the primed '
-                           'files and the linker version used in the base.')
+                           'base given the GLIBC mismatch of the primed '
+                           'files and the linker version (2.23) used in the '
+                           'base. These are the GLIBC versions required by '
+                           'the primed files that do not match:\n {}\n'.format(
+                               '\n'.join(files)))
```
kyrofa: Yes, very nice!

snapcraft/internal/elf.py
```diff
@@ -153,15 +156,20 @@ def is_linker_compatible(self, *, linker: str) -> bool:
 
     def get_required_glibc(self) -> str:
         """Returns the required glibc version for this ELF file."""
+        with contextlib.suppress(AttributeError):
+            return self._required_glibc  # type: ignore
```
kyrofa: I would prefer to see the more familiar pattern of setting this to `None` in `__init__` and setting it if so before returning here (like we do elsewhere), but I don't have strong feelings about it.

sergiusens: I am divided, now that `None` and `''` means you are changing types :confused:

kyrofa: Nah, `self._required_glibc = None # type: str`. You could use `self._required_glibc = ''` and get pretty much the same thing, too.

sergiusens:
That means that for every library where there is no glibc required version we'd parse over and over again (if using '' instead of None or anything different).

PR #1846, opened 2018-01-04
go plugin: remove confusing importpath message

Nissaar:

-----


go.py
```diff
@@ -0,0 +1,223 @@
+bugbu# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
```
kyrofa: This probably shouldn't be here :wink: .

kyrofa:
Also, your pull request and title (and commit message, from which the former are derived) should probably discuss why the change was needed and what bug it's fixing. The current title doesn't mean much.

go.py
kyrofa: Please delete this file.

kyrofa:
Alright, now this is ready for a good review! Thanks for the changes so far, @Nissaar. I have a question. I know this is fixing LP: [#1613980](https://bugs.launchpad.net/snapcraft/+bug/1613980), however, that bug contains the following paragraph:

> The "This is not needed" and printing systematically a warning is confusing. As for most of projects, we don't need go-importpath, I would suggest removing the warning.

I read that as suggesting that we remove the "please consider" warning, not the "this is not needed" documentation. Did you read it the other way? I'm afraid I don't remember why we added the warning in the first place, so I'd like to get @elopio (who marked the bug as confirmed) and @sergiusens's input on this as well.

sergiusens:
The "This is not needed" should probably be better read as "in some cases this is not required", the reason it is not required is to not break backwards compatibility and ideally should be set. Maybe this should become a deprecation message in the sense of it being deprecated to craft a snapcraft.yaml with the go plugin without this entry.

kyrofa:
So @sergiusens is this PR good as-is? Or would you like to see it also upgrade the warning to a deprecation?

snapcraft/plugins/go.py
```diff
@@ -30,13 +30,12 @@
       not be useful either.
       If the package is a part of the go-importpath the local package
       corresponding to those sources will be used.
-
-    - go-importpath:
+      
+      - go-importpath:
```
kyrofa: Please fix the extra spaces and indentation here.

kyrofa:
Closing due to inactivity. Please re-propose once you've addressed the issues.

PR #1847, opened 2018-01-04
lxd: Use user in container when mounting remote via sshfs

kalikiana:

-----
Based on the proof of concept #1484 I'm proposing setting up a user in the LXD container specifically when sshfs is used to mount the project folder into a remote instance of LXD to [ensure correct file ownership](https://bugs.launchpad.net/snapcraft/+bug/1741082).

Fixes: [LP: #1741082](https://bugs.launchpad.net/snapcraft/+bug/1741082 )

### DIFF
--- a/snapcraft/internal/lxd/_containerbuild.py
+++ b/snapcraft/internal/lxd/_containerbuild.py
@@ -179,9 +179,10 @@ class Containerbuild:
                 command += ['--target-arch', self._project_options.target_arch]
             if args:
                 command += args
-            self._container_run(command, cwd=self._project_folder, user=True)
+            self._container_run(command, cwd=self._project_folder,
+                                user=self._user)
 
-    def _container_run(self, cmd: List[str], cwd=None, user=False, **kwargs):
+    def _container_run(self, cmd: List[str], cwd=None, user='root', **kwargs):
         sh = ''
         original_cmd = cmd.copy()
         # Automatically wait on lock files before running commands
@@ -196,8 +197,8 @@ class Containerbuild:
         if sh:
             cmd = ['sh', '-c', '{}{}'.format(sh,
                    ' '.join(pipes.quote(arg) for arg in cmd))]
-        if user:
-            cmd = ['sudo', '-H', '-E', '-u', self._user] + cmd
+        if user != 'root':
+            cmd = ['sudo', '-H', '-E', '-u', user] + cmd
         try:
             subprocess.check_call([
                 'lxc', 'exec', self._container_name, '--'] + cmd,
--- a/snapcraft/internal/lxd/_containerbuild.py
+++ b/snapcraft/internal/lxd/_containerbuild.py
@@ -197,7 +197,7 @@ class Containerbuild:
             cmd = ['sh', '-c', '{}{}'.format(sh,
                    ' '.join(pipes.quote(arg) for arg in cmd))]
         if user:
-            cmd = ['sudo', '-H', '-u', self._user] + cmd
+            cmd = ['sudo', '-H', '-E', '-u', self._user] + cmd
         try:
             subprocess.check_call([
                 'lxc', 'exec', self._container_name, '--'] + cmd,
--- a/snapcraft/internal/lxd/_containerbuild.py
+++ b/snapcraft/internal/lxd/_containerbuild.py
@@ -63,7 +63,15 @@ class Containerbuild:
         self._source = os.path.realpath(source)
         self._project_options = project_options
         self._metadata = metadata
-        self._project_folder = '/root/build_{}'.format(metadata['name'])
+        if os.geteuid() > 0:
+            self._user = os.environ['USER']
+            self._home = '/home/{}'.format(self._user)
+        else:
+            self._user = 'root'
+            self._home = '/{}'.format(self._user)
+        # os.sep needs to be `/` and on Windows it will be set to `\`
+        self._project_folder = '{}/build_{}'.format(self._home,
+                                                    metadata['name'])
 
         if not remote:
             remote = _get_default_remote()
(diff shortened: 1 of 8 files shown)


snapcraft/internal/lxd/_containerbuild.py
```diff
@@ -195,6 +196,9 @@ def _container_run(self, cmd, cwd=None, **kwargs):
         if sh:
             cmd = ['sh', '-c', '{}{}'.format(sh,
                    ' '.join(pipes.quote(arg) for arg in cmd))]
+        # Run commands where file ownership matters as user
+        if os.geteuid() > 0 and (cwd or cmd[0] == 'mkdir'):
```
kyrofa: Wait... what's happening here? Turning this if statement into English I get: "If I'm not root, and I either have a cwd or I'm running mkdir, run as \<configured user\>". I'll admit I don't quite understand the problem behind the bug, but I don't see how these three conditionals are related in any way :stuck_out_tongue: . Can you explain that? The comment here isn't helpful.

snapcraft/internal/lxd/_containerbuild.py
```diff
@@ -289,6 +293,7 @@ def _inject_snap(self, name: str, tmp_dir: str):
             ], tmp_dir)
 
         container_filename = os.path.join(os.sep, 'run', filename)
+        os.chmod(filepath, 0o777)
```
kyrofa: Is this actually what you want? Or do you simply want to copy the owner's permissions over to everything else? I'd rather be smart about this instead of just setting everything to 777. If you agree, perhaps we should extract [_replicate_owner_mode](https://github.com/snapcore/snapcraft/blob/master/snapcraft/plugins/_python/_pip.py#L61) to be more generally useful.

snapcraft/internal/lxd/_project.py
```diff
@@ -151,6 +155,30 @@ def _remote_mount(self, destination, source):
...
+                'useradd', self._user, '--create-home'])
+        except ContainerRunError as e:
+            # username already in use
+            if e.exit_code != 9:
+                raise e
+        self._container_run([
+            'usermod', self._user, '-o',
+            '-u', str(os.getuid()), '-G', 'sudo'])
+        self._container_run([
+            'chown', '{}:0'.format(os.getuid(), 0),
```
kyrofa: You're calling `.format()` with two args, but only one replacement. I suspect you don't want the `0`, here. I'm also curious why we're using `0` for the group, here. Can you add a comment about that?

snapcraft/internal/lxd/_project.py
```diff
@@ -151,6 +155,30 @@ def _remote_mount(self, destination, source):
             'You can run the following command to enable it:\n'
             'echo Y | sudo tee /sys/module/fuse/parameters/userns_mounts')
 
+    def _setup_user(self):
```
kyrofa: This will setup a user even if snapcraft happens to be run as sudo. Is that desired?

snapcraft/internal/lxd/_project.py
```diff
@@ -151,6 +155,30 @@ def _remote_mount(self, destination, source):
...
+        self._user = os.environ['USER']
+        self._project_folder = '/home/{}/build_{}'.format(
+            self._user, self._metadata['name'])
+        logger.debug('Setting up user {!r} in container'.format(self._user))
+        try:
+            self._container_run([
+                'useradd', self._user, '--create-home'])
+        except ContainerRunError as e:
+            # username already in use
+            if e.exit_code != 9:
```
elopio: so, exit_code == 9 means username already in use and we ignore it?
The comment could be clearer, saying this, that the error will be ignored.

snapcraft/internal/lxd/_containerbuild.py
kalikiana: Your translation is correct. If Snapcraft is being run as a user commands inside the container that need to create folders or touch the project folder need to run as the user in the container so that file permissions will match. A counter example is `apt-get` which needs to be run with superuser permissions.

This was the least intrusive option I could think of at the time without modifying callers all over the place. Now I'm only using it for persistent remotes in the cloud, so I'll make it a boolean `user=True`. I find it slightly awkward but maybe it's clearer that way.

kalikiana: Btw thanks for questioning this. The code should make sense even without reading the bug and I sometimes don't see the perspective from an unbiased pair of eyes. Much appreciated!

snapcraft/internal/lxd/_project.py
kalikiana: The check is at the calling site. I'll move it around.

kalikiana:
@elopio I 'm not sure how you'd check that. Except that you can build a snap which you couldn't build before. It's basically a bug fix... although I guess using `-d` and watching for the message `Setting up user ... in container` would tell you that a user is being setup.

snapcraft/internal/lxd/_containerbuild.py
kalikiana: I did some more testing on this. It looks like the `chmod` is not necessary afterall, I'm guessing something else was wrong before and I fixed it in the meanwhile.

elopio:
Checking the message works for me. Or we could explore fancier weirder solutions, like making a script called snapcraft that fails if run as root, and calls the real snapcraft otherwise. Or something like that. I'm not saying we should do it, just throwing ideas :)

kalikiana:
@kyrofa @elopio All review comments should be addressed now. FYI I also modified the `sudo` call to use `-E` to preverse the environment because otherwise variables like `SNAPCRAFT_SETUP_CORE` won't work as expected (plus a new test case for it).

snapcraft/internal/lxd/_containerbuild.py
```diff
@@ -195,6 +196,8 @@ def _container_run(self, cmd, cwd=None, **kwargs):
         if sh:
             cmd = ['sh', '-c', '{}{}'.format(sh,
                    ' '.join(pipes.quote(arg) for arg in cmd))]
+        if user:
+            cmd = ['sudo', '-H', '-E', '-u', self._user] + cmd
```
kyrofa: I'm afraid I still find this confusing. We use the phrase "set up a user" and "run as user" and now we have "user is true or false" and yet by default self._user is 'root'. :stuck_out_tongue:

snapcraft/internal/lxd/_project.py
```diff
@@ -81,7 +82,8 @@ def _configure_container(self):
             subprocess.check_call([
                 'lxc', 'config', 'set', self._container_name,
                 'raw.idmap',
-                'both {} 0'.format(os.getenv('SUDO_UID', os.getuid()))])
+                'both {} {}'.format(os.getenv('SUDO_UID', os.getuid()),
+                                    os.getuid())])
```
kyrofa: Will this be thrown off if we're running with sudo (given the use of uid as opposed to euid)?

snapcraft/internal/lxd/_project.py
kyrofa: A comment here would be helpful.

snapcraft/internal/lxd/_containerbuild.py
kalikiana: I changed it to a `user='foo'` argument. I'm actually liking this best now.

snapcraft/internal/lxd/_project.py
kalikiana: Afair it only affects mounting folders as LXD devices. And `_user` would be set to `'root'` in the sudo case.

PR #1849, opened 2018-01-04
tests: add snap not found tests

daniellimws:

-----

### DIFF
--- a/snapcraft/tests/unit/store/test_store_client.py
+++ b/snapcraft/tests/unit/store/test_store_client.py
@@ -23,6 +23,7 @@ from unittest import mock
 
 import fixtures
 import pymacaroons
+import testscenarios
 from testtools.matchers import Contains, Equals
 
 from snapcraft import (
@@ -1433,16 +1434,6 @@ class PushMetadataTestCase(StoreTestCase):
         result = self.client.push_metadata('basic', metadata, True)
         self.assertIsNone(result)
 
-    def test_snap_not_found(self):
-        self._setup_snap()
-        metadata = {'field_ok': 'foo'}
-        raised = self.assertRaises(
-            errors.SnapNotFoundError,
-            self.client.push_metadata, 'notbasic', metadata, False)
-        self.assertThat(str(raised),
-                        Equals("Snap 'notbasic' was not found in '{}' series."
-                               .format(constants.DEFAULT_SERIES)))
-
 
 class PushBinaryMetadataTestCase(StoreTestCase):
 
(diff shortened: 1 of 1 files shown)


snapcraft/tests/unit/store/test_store_client.py
```diff
@@ -1506,3 +1517,13 @@ def test_conflicting_force(self):
             # force the update, even on conflicts!
             result = self.client.push_binary_metadata('basic', metadata, True)
         self.assertIsNone(result)
+
+    def test_snap_not_found(self):
```
kyrofa: It seems you have the same test twice, here. Is that true, or am I missing something?

kyrofa: Ah, binary metadata, very good!

daniellimws: Hi sorry I made a mistake it should be `push_binary_metadata` called inside this function, to cover this line too. [https://github.com/snapcore/snapcraft/blob/master/snapcraft/storeapi/_store_client.py#L302](https://github.com/snapcore/snapcraft/blob/master/snapcraft/storeapi/_store_client.py#L302)

kyrofa: Yeah I saw that, no problem :) .

daniellimws:
The integration tests keep failing because it's running too long.

snapcraft/tests/unit/store/test_store_client.py
elopio: Please rename this test.

snapcraft/tests/unit/store/test_store_client.py
```diff
@@ -1432,6 +1433,16 @@ def test_conflicting_force(self):
         result = self.client.push_metadata('basic', metadata, True)
         self.assertIsNone(result)
 
+    def test_snap_not_found(self):
+        self._setup_snap()
+        metadata = {'field_ok': 'foo'}
```
elopio: I dislike the practice of using 'foo'. In tests, when data is needed but the specific value is not important, it's common to use 'dummy'. If the value is used during the test, I use something like 'test-value'.

snapcraft/tests/unit/store/test_store_client.py
```diff
@@ -1432,6 +1433,16 @@ def test_conflicting_force(self):
...
+        raised = self.assertRaises(
+            errors.SnapNotFoundError,
+            self.client.push_metadata, 'notbasic', metadata, False)
```
elopio: Instead of 'notbasic', it would be clearer to use something like: 'test-unexisting-snap'.

snapcraft/tests/unit/store/test_store_client.py
daniellimws: @elopio What should I change it to?

snapcraft/tests/unit/store/test_store_client.py
daniellimws: @elopio what about the rest of the functions in the same class? I see them using foo, should we do anything about them?

elopio: not on this PR, but it would be nice to update them too in a separate one.
Not a big deal, and not everybody will agree with my style, but I find it nice to do at least in the new code.

snapcraft/tests/unit/store/test_store_client.py
elopio: Maybe call the first one test_snap_not_found_in_push_metadata, and this one test_snap_not_found_in_binary_metadata.

An alternative would be to use testscenarios. Your two tests are the same, so you can move them to their own test case, and pass the function to call in a scenario.
https://pypi.python.org/pypi/testscenarios

daniellimws: Thanks for letting me know about testscenarios, this looks better because I also hate repeated code.

Issue #1668, opened 2017-11-06
Bring in newer ld-linux and glibc libraries


sergiusens:
Estimate: 1 day

Issue #1669, opened 2017-11-06
patchelf with ld-linux from the future


sergiusens:
Estimate: 1 day

PR #1850, opened 2018-01-04
pluginhandler: patch and handle elf files on glibc mismatch

sergiusens:
If the required glibc is greater than the one provided by the given
base, fetch libc6 from the host, prime it and use the dynamic loader
from there instead.

Closes: #1668 #1669

-----


snapcraft/internal/elf.py
```diff
@@ -77,7 +77,6 @@ def __init__(self, *, soname: str, path: str, root_path: str,
             self.in_base_snap = False
 
 
-@lru_cache()
```
elopio: I'm curious, why did you remove it?

snapcraft/internal/pluginhandler/_mangling.py
```diff
@@ -0,0 +1,83 @@
...
+def _get_dynamic_linker(library_list: List[str]) -> str:
+    """Return the dynamic linker from library_list."""
+    regex = re.compile(r'(?P<dynamic_linker>ld-[\d.]+.so)$')
+
+    for library in library_list:
+        m = regex.search(os.path.basename(library))
+        if m:
+            return m.group('dynamic_linker')
+
+    raise NotImplementedError(
```
elopio: I don't think this should be a NotImplementedError.
A more snapcraft-specific exception would be better.

kyrofa: Agreed, although I think this is the one @sergiusens specifically did not want to make friendly. A RuntimeError would be better-suited, if needed.

snapcraft/internal/pluginhandler/_mangling.py
```diff
@@ -0,0 +1,83 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
```
kyrofa: We already have a `snapcraft.internal.mangling` module. Should this go in there?

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -575,7 +550,20 @@ def prime(self, force=False) -> None:
...
+        # get_core_dynamic_linker to resolve the final path for which
+        # we resort to our only working base 16, ld-2.23.so.
+        linker_compatible = (e.is_linker_compatible(linker='ld-2.23.so')
+                             for e in elf_files)
+        if not all((x for x in linker_compatible)):
+            handle_glibc_mismatch(elf_files=elf_files,
+                                  root_path=self.primedir,
+                                  snap_base_path=self._snap_base_path,
+                                  core_base_path=core_path)
+        elif self._confinement == 'classic':
```
kyrofa: Why are we patching the mismatched versions only on non-classic snaps?

snapcraft/internal/pluginhandler/_mangling.py
```diff
@@ -0,0 +1,83 @@
...
+        if required_glibc:
+            formatted_list.append('- {} -> GLIBC {}'.format(
+                elf_file.path, required_glibc))
+            patch_elf_files.append(elf_file)
+    logger.warning('The primed files will not work with the current '
+                   'base given the GLIBC mismatch of the primed '
+                   'files and the linker version (2.23) used in the '
+                   'base. These are the GLIBC versions required by '
+                   'the primed files that do not match and will be '
+                   'patched:\n {}\n'.format(
```
kyrofa: This message implies (to me) that we're patching the mismatching files to use the older version of libc, which isn't the case. We're migrating the newer libc into the snap and patching them to use that instead of the one in the core snap, right? Perhaps this warning could say something like:

"The primed files will not work[...] These are the GLIBC versions required by the primed files that don't match[...]. To work around this, the newer libc will be migrated into the snap, and these files will be patched to use it."

snapcraft/internal/elf.py
sergiusens: simple pimple; on running `load_dependencies` to get the new ones, we'd get the old ones. There is no API to wipe the cache so I just decided to remove it

snapcraft/internal/pluginhandler/_mangling.py
sergiusens: `RuntimeError` it shall be, this is not anything a user can fix and a horrible thing if it happens so the error should look just as grotesque for us to fix fast.

snapcraft/internal/mangling.py
```diff
@@ -46,3 +52,89 @@ def rewrite_python_shebangs(root_dir):
...
+
+    logger.warning('The primed files will not work with the current '
+                   'base given the GLIBC mismatch of the primed '
+                   'files and the linker version (2.23) used in the '
+                   'base. These are the GLIBC versions required by '
+                   'the primed files that do not match and will be '
+                   'patched:\n {}\n'
+                   'To work around this, the newer libc will be '
+                   'migrated into the snap, and these files will be '
+                   'patched to use it.'.format('\n'.join(formatted_list)))
```
kyrofa: This was tremendously helpful when testing, by the way.

snapcraft/tests/bin/readelf/readelf
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python3
```
kyrofa: Nice idea.

snapcraft/tests/unit/test_mangling.py
```diff
@@ -106,3 +108,116 @@ def test_following_docstring_no_rewrite(self):
             This is a test
             =======================
         """)))
+
+
+class FakeElf(fixtures.Fixture):
```
kyrofa: We have snapcraft/tests/fixture_setup.py-- should this go in there?

sergiusens: the only reason I didn't move it into fixtures_setup is because it is so big and I believe it has the wrong name.

I'll propose a followup with a refactor to `fixtures_setup`

PR #1851, opened 2018-01-05
storeapi: add docstrings to _client

daniellimws:
Add reST docstrings to _client

I'm not sure what to do for kwargs, so I've left them out for now.

-----

### DIFF
--- a/snapcraft/storeapi/_client.py
+++ b/snapcraft/storeapi/_client.py
@@ -16,6 +16,12 @@ class Client():
     """
 
     def __init__(self, conf, root_url):
+        """Set up Client object
+
+        :param config conf: Configuration details for the client
+        :param str root_url: Root url for all requests to base on.
+        :type config: snapcraft.config
+        """
         self.conf = conf
         self.root_url = root_url
         self.session = requests.Session()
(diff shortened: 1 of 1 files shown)


snapcraft/storeapi/_client.py
```diff
@@ -16,6 +16,12 @@ class Client():
     """
 
     def __init__(self, conf, root_url):
+        """Set up Client object
```
elopio: Instead of Set up, which is a concept for the test cases, I think it would be better to use the word `Initialize` here. If you search the codebase, we are already using this word for other docstrings.

snapcraft/storeapi/_client.py
```diff
@@ -16,6 +16,12 @@ class Client():
...
+
+        :param config conf: Configuration details for the client
+        :param str root_url: Root url for all requests to base on.
```
elopio: I would remove `to base on`.

snapcraft/storeapi/_client.py
```diff
@@ -16,6 +16,12 @@ class Client():
...
+        :type config: snapcraft.config
```
elopio: This is actually snapcraft.config.Config.

snapcraft/storeapi/_client.py
```diff
@@ -31,7 +37,16 @@ def __init__(self, conf, root_url):
         }
 
     def request(self, method, url, params=None, headers=None, **kwargs):
-        """Overriding base class to handle the root url."""
+        """Send a request to url relative to the root url.
+
+        :param str method: Method used for the request.
+        :param str url: url to send the request to, which will be appended with
+                        the root url first.
```
elopio: Just say: `appended to the root url`.

daniellimws:
@elopio thanks for the comments. May I know if I need to do anything about kwargs?

elopio:
I don't think so @daniellimws, your docstrings are readable and nice already.
Just note that we are migrating to use type hinting, and a more human prose for the docstrings, instead of the sphinx syntax. So, we will probably update this again in the future, but for now, it's a good change in my opinion.

daniellimws:
Thanks @elopio!

PR #1853, opened 2018-01-08
extractors: support appstream icon and desktop

### LINKED ISSUE #1828: Support desktop and icon in appstream handler

elopio:

-----
Closes #1828.


snapcraft/extractors/_metadata.py
```diff
@@ -63,7 +73,23 @@ def get_description(self) -> str:
...
-    def to_dict(self) -> Dict[str, str]:
+    def get_icon(self) -> str:
+        """Return extracted icon.
+
+        :returns: Extracted icon
+        :rtype: str
+        """
+        return self._data.get('icon')
+
+    def get_desktop(self) -> List[str]:
```
kalikiana: I'm finding the singular confusing here since there can be several files. Can we make that `get_desktops` or `get_desktop_files`?

snapcraft/extractors/_metadata.py
```diff
@@ -63,7 +73,23 @@ def get_description(self) -> str:
...
+        """Return extracted desktop files.
+
+        :returns: Extracted desktop files ids
```
kalikiana: One comment says "files", the other says "files ids". I think we should use only one term to avoid confusion. Since we rely on the fact that these are filenames, I'll vote for "desktop files".

snapcraft/tests/integration/general/test_metadata_appstream.py
```diff
@@ -76,3 +82,49 @@ def test_specified_metadata_not_overwritten(self):
...
+                'plugin': 'dump',
+                'parse-info': ['test.metainfo.xml']})
+        self.useFixture(snapcraft_yaml)
+
+        self.run_snapcraft('prime')
+
+        self.assertThat(
+            os.path.join('prime', 'meta', 'gui', 'icon.png'), FileExists())
+
+    def test_desktop_extracted_from_appstream(self):
```
kalikiana: Can you please add another test for two desktop files?

snapcraft/extractors/_metadata.py
kyrofa: Yeah the term "ID" confuses me as well. I suggest we keep the language in this class as generic as possible given that multiple extractors will be using it, not just appstream.

snapcraft/extractors/_metadata.py
kyrofa: I much prefer the "desktop files" language.

snapcraft/extractors/appstream.py
```diff
@@ -35,11 +35,25 @@ def extract(path: str) -> ExtractedMetadata:
...
-        description = node.text
-
-    return ExtractedMetadata(summary=summary, description=description)
+        description = node.text.strip()
+
+    icon = None
+    node = tree.find('icon')
+    if (node is not None and 'type' in node.attrib and
+            node.attrib['type'] == 'local'):
+        icon = node.text.strip()
```
kyrofa: Can you explain this a little (perhaps with a comment)? Is it possible to have a URL here? Would it be dumb for us to download it?

snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -133,6 +133,97 @@ def _update_yaml_with_extracted_metadata(
...
+        extracted_metadata: _metadata.ExtractedMetadata):
+    metadata_dict = extracted_metadata.to_dict()
+    for key, value in filter(
+            lambda item: item != 'desktop',
+            metadata_dict.items()):
+        if key not in config_data:
+            if key == 'icon':
+                if _icon_file_exists() or not os.path.exists(value):
+                    # Do not overwrite the icon file.
+                    continue
```
kyrofa: My understanding of this is: "If the icon has been specified by its conventional location, or the icon specified in the extracted metadata does not exist, skip over it."

Assuming my understanding is correct, I of course agree with giving the locally-supplied icon priority, but if the icon specified in the extracted metadata doesn't exist, I kinda feel like that should be an error. Is there a reason it's not?

snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -133,6 +133,97 @@ def _update_yaml_with_extracted_metadata(
         raise meta_errors.MissingSnapcraftYamlKeysError(keys=missing_keys)
 
 
+def _adopt_info(
+        config_data: Dict[str, Any],
+        extracted_metadata: _metadata.ExtractedMetadata):
+    metadata_dict = extracted_metadata.to_dict()
+    for key, value in filter(
+            lambda item: item != 'desktop',
```
kyrofa: Personally, I'd rather just see an `if` statement in the loop instead of this fancy thing :stuck_out_tongue: . It causes a double-take for something that is otherwise a simple one-liner.

snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -133,6 +133,97 @@ def _update_yaml_with_extracted_metadata(
...
+                ['setup', 'snap'],
+                ['icon.png', 'icon.svg'])):
+        if os.path.exists(icon_path):
+            return True
+    else:
+        return False
+
+
+def _get_app_name_from_desktop_id(
+        config_data: Dict[str, Any], desktop_id: str) -> str:
```
kyrofa: Hmm... yeah I don't feel like this scales well. The reason we have to do this is obvious: apps aren't tied to parts. However, I have two thoughts:

1) Why use the file name? Why not actually take a look at the `Exec` line to see what app it's running? We don't want to ship desktop files that don't run an app in the snap, anyway.

2) App-specific `desktop` entries are only _one_ way to get .desktop files into the snap. The other way is the conventional location: the `meta/gui` directory. What if we just installed all the adopted desktop files there, no YAML updates needed? Then we could take a look at the `Exec` lines and filter based on our list of app names. Thoughts?

snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -133,6 +133,97 @@ def _update_yaml_with_extracted_metadata(
...
+        if key not in config_data:
+            if key == 'icon':
+                if _icon_file_exists() or not os.path.exists(value):
+                    # Do not overwrite the icon file.
+                    continue
+            config_data[key] = value
+    if 'desktop' in metadata_dict:
+        for desktop_id in metadata_dict['desktop']:
+            app_name = _get_app_name_from_desktop_id(config_data, desktop_id)
+            if app_name and not _desktop_file_exists(app_name):
```
kyrofa: If not app_name, would it be helpful to display a warning?

snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -133,6 +133,97 @@ def _update_yaml_with_extracted_metadata(
...
+                for desktop_file_path in (
+                        os.path.join(
+                            xdg_data_dir, 'applications',
+                            desktop_id.replace('-', '/')) for
+                        xdg_data_dir in ('usr/local/share', 'usr/share')):
```
kyrofa: Another double-take here (triple, actually). If a list comprehension goes to multiple lines I get lost quickly, and here it's nested inside a `for`! Mind extracting it and making it a little easier to read?

snapcraft/extractors/_metadata.py
```diff
@@ -15,27 +15,37 @@
...
 class ExtractedMetadata(yaml.YAMLObject):
     """Collection of metadata extracted from a part."""
 
     yaml_tag = u'!ExtractedMetadata'
 
-    def __init__(self, *, summary='', description='') -> None:
+    def __init__(
+            self, *, summary: str='',
+            description: str='', icon: str='',
+            desktop: List[str]=None) -> None:
```
kyrofa: Please name this in a somehow-obviously-plural way.

snapcraft/internal/meta/_snap_packaging.py
sergiusens: I sort of like functional steps so I don't mind either way :-)

If we didn't have the 80 char limit I would be rooting for this :100: though :wink:. Alas I will let @kyrofa has his way :smile:

snapcraft/extractors/_metadata.py
elopio: The thing is that these are not desktop_files. They are desktop_file_ids, as defined in https://standards.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#desktop-file-id
I kind of wanted not to specify too much in the name, because other sources of metadata might user file paths instead of ids, or file names, or something else. But well, this is not clear, I agree with you.

I will rename it get_destkop_file_ids. And later, if there's another type, we can add an attribute and a get_desktop_file_paths, or something like that. It won't be nice and consistent, but that seems unavoidable given this weird desktop_file_id thing.

snapcraft/extractors/appstream.py
elopio: Yes, it ccould be remote instead of local.
I think it will not be dumb, but I think it must be handled somewhere else. In here we just copy the value to the icon key in config_data. So we would need to support remote URLs in config_data first. I think we should do it, as a future branch.
I will add a comment, and a bug.

snapcraft/internal/meta/_snap_packaging.py
elopio: I might have gone a little too far with the itertools ^_^

snapcraft/internal/meta/_snap_packaging.py
elopio: I was thinking about cases when the icon is moved during a step in the build phase, or the packaging phase. I think that's common. So, I was imagining a case where we find the appstream file, but the icon is not in the location specified there because that's a packaging build task that doesn't run on the normal compilation.
Maybe I was thinking too much? Should we wait for people to complain about us forcing them to move the icon around?

snapcraft/internal/meta/_snap_packaging.py
elopio: Maybe. This seemed to me like a best-effort extractor, trying to match as much as possible but not complaining too much, so I didn't put any warning. But we certainly can.
However, because the appstream file can have multiple desktop files, there will be cases in which the snap only defines a few of the apps that have a launchable in appstream. Do you think we should warn in such cases?

snapcraft/internal/meta/_snap_packaging.py
elopio: sure thing. This was totally unnecessary.

snapcraft/internal/meta/_snap_packaging.py
elopio: The Exec would work, I think. But would you expect the app name or the command in the Exec?
I don't like very much because we rewrite the Exec line during prime, so it feels more like a throwaway value to me. Also, snaps sometimes force us to have an app name that's not the same as the executed command. 

I'm not sure which is best, but I would prefer to keep it simple mostly because this is temporary. Once we have the appstream-id there is no more reason to guess.

About your second point, I'm not sure I get it. The other way to define them is to put them in `snap/gui`, and they are matched with the corresponding app by file name. They have to be called test-app.desktop. And during prime they are moved to `meta/gui`. So do you want to replace matching them by name to match them by exec?

snapcraft/tests/integration/general/test_metadata_appstream.py
elopio: Yes!

snapcraft/extractors/_metadata.py
```diff
@@ -15,27 +15,37 @@
...
         :param str summary: Extracted summary
         :param str description: Extracted description
-        """
+        :param str icon: Extracted icon
+        :param dict desktop: Extracted desktop ids, as defined in
+            https://standards.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#desktop-file-id
+        """  # noqa
 
-        self._data = {}  # type: Dict[str, str]
+        self._data = {}  # type: Dict[str, Any]
```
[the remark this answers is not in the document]
elopio: I tried with Union, but then it required me to use as return type for all the other attributes Union[str, List[str]] instead of just str. That felt wrong, because then if we add a new type we would have to modify all the return types. Apparently, it is not smart enough. Or maybe I was doing something wrong, have you tried it?

snapcraft/extractors/_metadata.py
kalikiana: Aha! Thanks for elaborating on that. In that case `get_desktop_file_ids` sounds perfect.

snapcraft/tests/integration/general/test_metadata_appstream.py
kalikiana: Hmm did you add it? I don't see it. Seems like you squashed or amended the commit?

kalikiana: Aha, nevermind, it's in `test_appstream_with_multiple_launchables`. Thanks a lot for adding that!

snapcraft/extractors/_metadata.py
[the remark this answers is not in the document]
kyrofa: Well, that's correct, is it not? Any value that comes out of this dict could be either a string or a list of strings.

snapcraft/internal/meta/_snap_packaging.py
kyrofa: No actually, I think you're right. Since we check several different root directories for this file, it may very well not exist.

snapcraft/internal/meta/_snap_packaging.py
kyrofa: Ah yes. No, I think what you've done is correct.

snapcraft/internal/meta/_snap_packaging.py
kyrofa: > I don't like very much because we rewrite the Exec line during prime, so it feels more like a throwaway value to me.

Good call, I forgot about that.

> About your second point, I'm not sure I get it. The other way to define them is to put them in snap/gui, and they are matched with the corresponding app by file name.

I don't see them being matched anywhere, so perhaps we're misunderstanding each other. They're just [migrated straight over](https://github.com/snapcore/snapcraft/blob/master/snapcraft/internal/meta/_snap_packaging.py#L229), and we can put stuff there ourselves as well. I don't think we need to rename them at all, but they do need to have a valid Exec line, which is perhaps too much to ask for.

kyrofa: > I'm not sure which is best, but I would prefer to keep it simple mostly because this is temporary. Once we have the appstream-id there is no more reason to guess.

But that only stands for appstream metadata, right? What about desktop files coming from something else?

snapcraft/extractors/_metadata.py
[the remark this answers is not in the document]
elopio: I think I got it with your hint about casting. Please take a look at the last commit.

snapcraft/extractors/_metadata.py
```diff
@@ -53,17 +63,33 @@ def get_summary(self) -> str:
         :returns: Extracted summary
         :rtype: str
         """
-        return self._data.get('summary')
+        return str(self._data.get('summary'))
```
kyrofa: Yep, that's what I was talking about.

snapcraft/internal/meta/_snap_packaging.py
elopio: Sorry, I was rebasing and apparently did it wrong. I'm happy to revisit this, we can still change the behaviour without much fuss even after this lands. Specially, because what you suggested preserves the behaviour for the most common cases.

So I'm looking at the code that generates the desktop files again, and it's very confusing. I've just noticed that when the desktop keyword is specified in the snapcraft.yaml, the exec line is rewritten. But when the file is saved in snap/gui, it is left untouched. That's inconsistent, and I think it's wrong. I think that when the file is in snap/gui, we should move it to prime/meta/gui, and check if there's an app with the same name as the desktop file. If so, rewrite the exec line to make it consistent with the other way of specifiying them.

This is what I thought that was happening, but it's only for half of the story.

elopio: About the other question, that's one of the reasons why I wanted to leave all the metadata handling to the prime step only. That way we would have a single place that requires to know the specifics of the metadata source, to extract and convert at the same time.

Once we add appstream id support, we can turn this desktop file ids into a dictionary that maps appstream id to desktop file id. As you point out, this is appstream-specific, so the field in the metadata object should be called appstream_launchable, or something. I see no way around this with the current implementation. I'm happy to meet to talk more about this too.

kyrofa:
Got some test failures, here.

elopio:
All green now. Some unexpected side effects of the str cast, I had to add a few ifs.

snapcraft/extractors/_metadata.py
```diff
@@ -63,31 +63,35 @@ def get_summary(self) -> str:
         :returns: Extracted summary
         :rtype: str
         """
-        return str(self._data.get('summary'))
+        summary = self._data.get('summary')
```
sergiusens: why not

```python
summary = self._data.get('summary', '')?

kyrofa: I kinda like that the API returns `None` if the value wasn't extracted, but I don't think we have a use-case for that right now.

sergiusens: fair enough

PR #1857, opened 2018-01-09
grammar: make on statement work with host arch

kalikiana:

-----
This is changing the "on" statement [as discussed in the forum](https://forum.snapcraft.io/t/specifying-architecture-on-cross-compilations/1641/20) to work with the host architecture. This PR implements the **bold items**.

1. The existing “on <arch>:” syntax will be accepted under build-packages as well
2. The default architecture for packages under “on <arch>” is the one from the build machine
3. **Packages under “on <arch>:” are installed only when the build machine has that architecture**
4. A new “to <arch>:” statement will be supported in build-packages and stage-packages
5. The default architecture for packages under “to <arch>:” is the target architecture
6. Packages under “to <arch>:” are installed only when the target architecture is the one listed, which includes both a cross-compilation towards that architecture and a native build on that architecture
7. The two statements may be used together as “on <arch> to <arch>:” to precisely specify the case of a particular cross-compilation
8. The default architecture for packages under “on <arch> to <arch>:” is the target architecture
 
Test cases are updated to reflect the new semantics.


snapcraft/tests/unit/project_loader/grammar/test_on_statement.py
```diff
@@ -151,7 +152,12 @@ class OnStatementGrammarTestCase(GrammarTestCase):
...
     ]
 
-    def test_on_statement_grammar(self):
+    @patch('platform.architecture')
+    @patch('platform.machine')
+    def test_on_statement_grammar(self, platform_machine_mock,
+                                  platform_architecture_mock):
+        platform_machine_mock.return_value = 'x86_64'
+        platform_architecture_mock.return_value = ('64bit', 'ELF')
         options = snapcraft.ProjectOptions(target_deb_arch=self.target_arch)
```
kyrofa: This test case was written with the idea that it could switch what it was comparing its selectors against. I don't think it's testing the same cases, now. Is there a way to preserve that now that it's no longer based on the target arch?

kalikiana: By switch you mean pass on different architectures? Generally it's more reliable to use scenarios for that rather than relying on the architecture of the test machine.
I changed it now to run all tests with both amd64 and i386 target architectures. Let me know what you think.

kalikiana: Perhaps to expand a little, the tests will no longer pass on different host architectures since that's the bit we're testing now. For example `on amd64` won't trigger on anything but amd64 hosts, unlike before where only the target was relevant.

kalikiana: Arg, I just realized somehow I didn't actually finish the commit... that explains the test failure.  Pushed it now...

kyrofa: > For example on amd64 won't trigger on anything but amd64 hosts, unlike before where only the target was relevant.

Indeed, perhaps I wasn't clear. The tests were previously setting target arch. I'd like to see the tests unchanged except that they're now setting the host arch instead of the target. That way they're still testing the same things.

kalikiana: I went back to how they were and swapped the target arch for the host arch (plus adjusting them so they pass). Also dropped the target_arch from a few tests that didn't use it at all while I was at it. But It's rather tedious to go through these so I hope we are on the same page now :-D

snapcraft/internal/project_loader/grammar/_on.py
```diff
@@ -36,7 +38,7 @@ class OnStatement:
     ...     return True
     >>> with tempfile.TemporaryDirectory() as cache_dir:
     ...     options = ProjectOptions(target_deb_arch='i386')
-    ...     clause = OnStatement(on='on amd64', body=['foo'],
+    ...     clause = OnStatement(on='on i386', body=['foo'],
```
kyrofa: This test will fail on i386, now.

snapcraft/tests/unit/project_loader/grammar/test_processor.py
```diff
@@ -57,55 +58,55 @@ class BasicGrammarTestCase(GrammarTestCase):
...
-            'target_arch': 'amd64',
+            'host_arch': 'x86_64',
             'expected_packages': {'foo'}
         }),
         ('on i386', {
             'grammar': [
-                {'on amd64': ['foo']},
-                {'on i386': ['bar']},
+                {'on i386': ['foo']},
+                {'on amd64': ['bar']},
```
kyrofa: I really don't understand why you felt it necessary to switch these all around :frowning_face: .

PR #1860, opened 2018-01-09
setup: simplify bin/snapcraft and correct tests

sergiusens:
bin/snapcraft has been drastically simplified and a comment added
that it is a temprorary solution until cx_Freeze supports
console_scripts.

The integration and snaps tests do not fallback to running bin/snapcraft
anymore but expect a proper venv setup instead to guarantee consistent
results between developers.

-----


bin/snapcraft
```diff
@@ -14,23 +14,4 @@
...
-    sys.path = [topdir] + sys.path
-
-
-if __name__ == '__main__':
-    import snapcraft.cli
-    import snapcraft.internal.dirs
-    snapcraft.internal.dirs.setup_dirs()
-    snapcraft.cli.run(prog_name='snapcraft',
-                      obj=dict(project=snapcraft.ProjectOptions()))
+__import__('snapcraft.cli.__main__', fromlist=['__name__'], level=0)
```
elopio: Why are you using this instead of a normal import?

Also, the __import__ docs say: "Direct use of __import__() is also discouraged in favor of importlib.import_module()"

snapcraft/tests/integration/__init__.py
```diff
@@ -57,11 +57,17 @@ def setUp(self):
...
+                os.getenv('VIRTUAL_ENV'), 'bin', 'snapcraft')
             self.snapcraft_parser_command = os.path.join(
-                os.getcwd(), 'bin', 'snapcraft-parser')
+                os.getenv('VIRTUAL_ENV'), 'bin', 'snapcraft-parser')
+        else:
+            raise EnvironmentError(
+                'snapcraft is not setup correctly for testing. Either set '
+                'SNAPCRAFT_FROM_SNAP or SNAPCRAFT_FROM_DEB to run from either '
+                'the snap or deb, or make sure your venv is properly setup '
+                'as described in HACKING.md.')
```
elopio: This comment, thanks!

elopio:
This looks correct, and works for the things I've tried. But I left a question. And maybe I misunderstood, but I see no comment about cx_Freeze.

bin/snapcraft
sergiusens: I replicated `pkg_resources`

sergiusens:
added the comment, must of got wiped when I simplified to the one liner it is now

bin/snapcraft
sergiusens: it is now a comment in the code as well

PR #1865, opened 2018-01-11
lxd: always (re-)injects snaps

kalikiana:

-----
Fixes: [LP: #1741752](https://bugs.launchpad.net/snapcraft/+bug/1741752)

### DIFF
--- a/snapcraft/internal/lxd/_containerbuild.py
+++ b/snapcraft/internal/lxd/_containerbuild.py
@@ -226,13 +226,13 @@ class Containerbuild:
                     raise e
         logger.info('Network connection established')
 
-    def _inject_snapcraft(self):
+    def _inject_snapcraft(self, *, new_container: bool):
         if common.is_snap():
             with tempfile.TemporaryDirectory(
                     prefix='snapcraft', dir=self._lxd_common_dir) as tmp_dir:
                 self._inject_snap('core', tmp_dir)
                 self._inject_snap('snapcraft', tmp_dir)
-        else:
+        elif new_container:
             self._container_run(['apt-get', 'install', 'snapcraft', '-y'])
 
     def _inject_snap(self, name: str, tmp_dir: str):
--- a/snapcraft/internal/lxd/_containerbuild.py
+++ b/snapcraft/internal/lxd/_containerbuild.py
@@ -126,15 +126,6 @@ class Containerbuild:
             'lxc', 'config', 'set', self._container_name,
             'environment.LC_ALL', 'C.UTF-8'])
         self._set_image_info_env_var()
-        info = subprocess.check_output([
-            'lxc', 'info', self._container_name]).decode('utf-8')
-        for line in info.splitlines():
-            if line.startswith("Architecture:"):
-                self._container_arch = _get_deb_arch(
-                    line.split(None, 1)[1].strip())
-                break
-        else:
-            raise ContainerError("Could not find architecture for container")
 
     def _set_image_info_env_var(self):
         FAILURE_WARNING_FORMAT = (
@@ -255,8 +246,8 @@ class Containerbuild:
         is_classic = json['result']['confinement'] == 'classic'
 
         # If the server has a different arch we can't inject local snaps
-        if (self._project_options.target_arch
-                and self._project_options.target_arch != self._container_arch):
+        target_arch = self._project_options.target_arch
+        if (target_arch and target_arch != self._get_container_arch()):
             channel = json['result']['channel']
             return self._install_snap(name, channel, is_classic=is_classic)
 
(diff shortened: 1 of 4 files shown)


snapcraft/tests/unit/test_lxd.py
```diff
@@ -537,6 +537,49 @@ def make_containerbuild(self):
...
+            {'name': 'snapcraft',
+             'confinement': 'classic',
+             'id': '3lljuRvshPlM4gpJnH5xExo0DJBOvImu',
+             'channel': 'edge',
+             'revision': '345'},
+        ]
+
+        self.make_containerbuild().execute()
+
+        if hasattr(self, 'cross') and self.cross:
```
kyrofa: What is this doing, exactly? What is this attribute?

snapcraft/tests/unit/test_lxd.py
```diff
@@ -537,6 +537,49 @@ def make_containerbuild(self):
                            project_options=self.project_options,
                            remote=self.remote)
 
+    @patch('snapcraft.internal.lxd.Containerbuild._container_run')
+    @patch('snapcraft.internal.common.is_snap')
+    def test_inject_snap_existing_container(self,
```
kyrofa: What makes this container already-existing?

snapcraft/internal/lxd/_containerbuild.py
```diff
@@ -226,13 +226,13 @@ def _wait_for_network(self):
...
 
-    def _inject_snapcraft(self):
+    def _inject_snapcraft(self, *, new_container: bool):
         if common.is_snap():
             with tempfile.TemporaryDirectory(
                     prefix='snapcraft', dir=self._lxd_common_dir) as tmp_dir:
                 self._inject_snap('core', tmp_dir)
                 self._inject_snap('snapcraft', tmp_dir)
-        else:
+        elif new_container:
```
kyrofa: I feel like I'm missing something, here. How does this re-inject the snap?

snapcraft/tests/unit/test_lxd.py
kalikiana: Apparently I didn't commit that bit... fixed

snapcraft/tests/unit/test_lxd.py
kalikiana: This is using the same approach as the other test cases in this file. The snaps are installed from the store and not from a local file in the case of a remote with another architecture.

kalikiana: Oh, I didn't realize they're not used in the ProjectTestCase. Adding.

kalikiana:
@sergiusens Note: this branch is about injecting snaps into existing containers. The `_inject_snap` code is the same except for replacing the `_container_arch` variable with a method. So tweaks of that logic should potentially go into a separate branch.
The file movements come first because 1) x-revisioned snaps aren't user-accessible 2) we compare the sha384sum since we can't universally rely on revisions. We would need to have a separate code path for store-owned snaps (would have to check if snapd talks HTTP, or use `snap info`).

I don't understand your point about confusing results. Assertions are not relevant to side-loaded snaps, and only pushed before actual installation of the snap.

snapcraft/internal/lxd/_containerbuild.py
```diff
@@ -298,6 +289,14 @@ def _inject_snap(self, name: str, tmp_dir: str):
                            is_dangerous=rev.startswith('x'),
                            is_classic=is_classic)
 
+    def _get_container_arch(self):
```
kyrofa: Good call, moving this to be checked every time.

snapcraft/tests/unit/test_lxd.py
```diff
@@ -537,6 +537,52 @@ def make_containerbuild(self):
...
+                                            mock_is_snap,
+                                            mock_container_run):
+        mock_is_snap.side_effect = lambda: True
```
kyrofa: Why not just `mock_is_snap.return_value = True`?

snapcraft/tests/unit/test_lxd.py
```diff
@@ -537,6 +537,52 @@ def make_containerbuild(self):
...
+        mock_container_run.side_effect = lambda cmd, **kwargs: cmd
```
kyrofa: The real function doesn't actually return anything... is this really what you want?

snapcraft/tests/unit/test_lxd.py
```diff
@@ -537,6 +537,52 @@ def make_containerbuild(self):
...
+        if hasattr(self, 'cross') and self.cross:
+            mock_container_run.assert_has_calls([
+                call(['snap', 'install', 'core', '--channel', 'stable']),
+                call(['snap', 'refresh', 'core', '--channel', 'stable']),
+                call(['snap', 'install', 'snapcraft', '--channel', 'edge',
+                      '--classic']),
+                call(['snap', 'refresh', 'snapcraft', '--channel', 'edge',
+                      '--classic']),
+            ])
+            return
```
kyrofa: I'd rather see an `else` here instead of an early exit.

PR #1868, opened 2018-01-12
tests: call file directly for the HasArchitecture checker

sergiusens:

-----

### DIFF
--- a/snapcraft/tests/matchers.py
+++ b/snapcraft/tests/matchers.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright (C) 2017 Canonical Ltd
+# Copyright (C) 2017-2018 Canonical Ltd
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License version 3 as
@@ -13,8 +13,8 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+import subprocess
 
-import magic
 import testtools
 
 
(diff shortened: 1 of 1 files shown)


snapcraft/tests/matchers.py
```diff
@@ -23,20 +23,29 @@ class HasArchitecture:
...
+                'Failed to get architecture info: {}', cmd_err) from cmd_err
+
+        try:
+            # output is of the form
+            # ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), \
+            # dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, \
+            # for GNU/Linux 3.7.0, \
+            # BuildID[sha1]=58b8495bef7a6c9e5f403879e29dc63ad4ed2d11, \
+            # not stripped
+            arch = output.split(',')[1]
```
kyrofa: On amd64 I get:

`<file name>: ELF 64-bit LSB executable, x86-64, version 1 (GNU/Linux), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32, BuildID[sha1]=a4b02c379f4482e7cb3148e0d3fc9459a9126cfd, not stripped`

`output.split(',')[1]` is ` x86-64` (notice the space). Might want to strip this.

snapcraft/tests/matchers.py
```diff
@@ -23,20 +23,29 @@ class HasArchitecture:
...
         return 'HasArchitecture()'
 
     def match(self, file_path):
-        magic = self._ms.file(file_path)
-        # Catch exceptions on splitting the string to provide context
-        # This includes "cannot open `...' (No such file or directory)"
+        cmd = ['file', file_path]
         try:
-            arch = magic.split(',')[1]
+            output = subprocess.check_output(cmd).decode()
```
kyrofa: Should you be specifying an encoding, here?

sergiusens: The default is utf-8 and from what I've read is sane enough (and common practice)

sergiusens: which we required btw due to use of click

snapcraft/tests/matchers.py
sergiusens: the original implementation is exactly the same, except that we call `file` directly.

This is why we do `expected_arch in arch`. Notice that on arm64 `output.split(',')[1]` is `ARM aarch64`

kyrofa: Ah yes!

PR #1870, opened 2018-01-12
elf: remove dependency on magic

sergiusens:
Migrate to using readelf which we already use and do a general cleanup.

LP: #1742504

-----


snapcraft/internal/elf.py
```diff
@@ -86,52 +84,94 @@ def _crawl_for_path(*, soname: str, root_path: str,
...
+    def _extract_readelf(self, path) -> Tuple[str, List[Symbol]]:
+        file_type = str()
         symbols = list()  # type: List[Symbol]
-        symbols_section = subprocess.check_output([
-            'readelf', '--wide', '--dyn-syms', path]).decode()
+        output = subprocess.check_output([
+            'readelf', '--wide', '--file-header', '--dyn-syms', path])
+        readelf_lines = output.decode().split('\n')
         # Regex inspired by the regex in lintian to match entries similar to
         # number '1' from the following sample output
```
kyrofa: This comment may be out of place, now. Or perhaps needs to be rewritten. The '1' to which you're referring is unclear.

snapcraft/internal/elf.py
```diff
@@ -86,52 +84,94 @@ def _crawl_for_path(*, soname: str, root_path: str,
...
-                    name, version = symbol.split('@')
-                except ValueError:
-                    name = symbol
-                    version = ''
-                symbols.append(Symbol(name=name, version=version,
-                                      section=section))
-        return symbols
+        symbol_regex = re.compile(symbol_match)
+        in_symbols_section = False
+        for line in readelf_lines:
```
kyrofa: I feel like we're pushing the readelf stdout parsing pretty far, here. Have we considered using pyelftools and reading the ELF files natively? The API there looks pretty decent, and I feel like native python solutions will work better in the long run than parsing stdout. Thoughts?

snapcraft/internal/elf.py
```diff
@@ -86,52 +84,94 @@ def _crawl_for_path(*, soname: str, root_path: str,
...
+                    symbol = m.group('symbol')
+                    try:
+                        name, version = symbol.split('@')
+                    except ValueError:
+                        name = symbol
+                        version = ''
+                    symbols.append(Symbol(name=name, version=version,
+                                          section=section))
+            else:
+                m = type_regex.match(line)
```
kyrofa: Would it simply the loop if you just used a multiline regex for this?

snapcraft/tests/bin/readelf/ldd
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env python3
```
kyrofa: Is there a reason this is nested under a 'readelf' directory instead of just being in bin?

snapcraft/internal/elf.py
kyrofa: Or using pyelftools you can just use `elffile.header['e_type']`.

snapcraft/internal/elf.py
```diff
@@ -86,52 +84,94 @@ def _crawl_for_path(*, soname: str, root_path: str,
...
-                except ValueError:
-                    name = symbol
-                    version = ''
-                symbols.append(Symbol(name=name, version=version,
-                                      section=section))
-        return symbols
+        symbol_regex = re.compile(symbol_match)
+        in_symbols_section = False
+        for line in readelf_lines:
+            if line.startswith("Symbol table '.dynsym'"):
```
kyrofa: Imagine this:

```python
section = elffile.get_section_by_name(b'.dynsym')
for symbol in section.iter_symbols():
   # ...
```

kyrofa:
I've successfully tested this PR on artful and it works the same way as master regarding the newer libc, but I would like to get your thoughts on a more native solution to this.

snap/snapcraft.yaml
```diff
@@ -12,7 +12,7 @@ confinement: classic
 
 apps:
   snapcraft:
-    command: bin/snapcraft-classic
+    command: $SNAP/usr/bin/python3 $SNAP/bin/snapcraft
```
kyrofa: Travis does like this:

> FileNotFoundError: [Errno 2] No such file or directory: '/root/snapcraft/prime/$SNAP/usr/bin/python3'

snapcraft/internal/elf.py
sergiusens: Prior to thi I did look at pyelftools, try ir on aarch64 or armv7 and come back to (you can shortcut the process by reading their docs :wink:)
Lintian has done this for years, so I am pretty confident in it, `readelf` is part of binutils.

snapcraft/tests/bin/readelf/ldd
sergiusens: I should call this topdir elf

snapcraft/internal/elf.py
sergiusens: I will when it works on anything but x86-64

snap/snapcraft.yaml
sergiusens: meh, mvo told me it would work, but from the looks of it, we are the ones messing up

sergiusens:
@kyrofa if you can confirm it works on all architectures, I am fine, I already took a look at this 3 weeks ago and then saw this https://github.com/eliben/pyelftools/wiki/User's-guide#elf
"The current focus of the library is on Intel's x86 and x64 architectures, with some ARM support added recently."

If you can confirm that is not the case and works on all the architectures in ubuntu then lets go ahead and migrate. But so far lintian has had a good time using `readelf` which means it is tried and tested.

kyrofa:
Note that bug occurred on i386 and armhf. If I had the option, I'd run adt on armhf. Since we can't, let's make sure we test the release well, there.

sergiusens:
@kyrofa why are you awake! :-)

WRT pyelfutils, did it work for you on armlv8/aaarch64/arm64 ?

Just like `patchelf` I am afraid we will hit corner cases where segments or the data within the segments are ordered differently and fall into the need of becoming maintainers of `pyelfutils`. I am not sure we have the bandwidth for that.

@elopio mind testing building on armhf or i386?

FWIW, this now enables me to do a `patchelf` improvement everyone will be happy with (like @flexiondotorg and @popey)

sergiusens:
@elopio should be from the snap though and should be on a clean environment (no venv or deps previously installed).

popey:
Happy to also test this on 368 and armhf today if wanted. Can you push a snap for each arch I can test?

popey:
Happy to also test this on 368 and armhf today if wanted. Can you push a snap for each arch I can test?

PR #1873, opened 2018-01-16
elf: cleaner patchelf experience

sergiusens:
Use the patchelf on the system by default if not running from the
snap and if it fails instruct the user to use patchelf as a part
or as a build-snaps entry.

Given that when using as a part the user can choose to not use
filters to remove the file from prime, we also make a copy of
the file to patch (breaking any potential existing hard link
with the other steps) to avoid the problem of trying to patch
the patchelf primed file with the same patchelf binary file.

-----


elopio:
@sergiusens your unit tests are failing.

snapcraft/internal/elf.py
```diff
@@ -301,14 +306,17 @@ def __init__(self, *, dynamic_linker: str, root_path: str) -> None:
         # in by packaging dependencies.
         # The docker conditional will work if the docker image has the
         # snaps unpacked in the corresponding locations.
-        if common.is_snap():
+        which_patchelf = shell_utils.which('patchelf')
+        if which_patchelf.endswith('stage/patchelf'):
```
elopio: Isn't this flaky? It could point to somewhere else completely different. I would say that this needs to check the full stage path.

snapcraft/internal/elf.py
```diff
@@ -301,14 +306,17 @@ def __init__(self, *, dynamic_linker: str, root_path: str) -> None:
...
+            self._patchelf_cmd = which_patchelf
+        elif common.is_snap():
             snap_dir = os.getenv('SNAP')
             self._patchelf_cmd = os.path.join(snap_dir, 'bin', 'patchelf')
         elif (common.is_docker_instance() and
               os.path.exists('/snap/snapcraft/current/bin/patchelf')):
             self._patchelf_cmd = '/snap/snapcraft/current/bin/patchelf'
         else:
-            self._patchelf_cmd = 'patchelf'
+            self._patchelf_cmd = which_patchelf
```
elopio: This is the same check as the first if, so you can rewrite this statement to something like.
```
self._patchelf_cmd = shell_utils.which('patchelf')
if not patchelf in stage:
    if is_snap():
        [...]
    elif docker:
        [...]
```
Looks more readable to me.

snapcraft/internal/elf.py
```diff
@@ -321,39 +329,57 @@ def patch(self, *, elf_file: ElfFile) -> None:
...
-        # bundled with snapcraft which means its lack of existence is a
-        # "packager" error.
-        except subprocess.CalledProcessError as call_error:
-            raise errors.PatcherError(elf_file=elf_file_path,
-                                      message=str(call_error))
+    def _run_patchelf(self, *, patchelf_args: List[str],
+                      elf_file_path: str) -> None:
+
+        with tempfile.NamedTemporaryFile() as temp_file:
+            shutil.copy2(elf_file_path, temp_file.name)
```
elopio: It would be nice to explain here why you copied it.

snapcraft/tests/bin/elf/strip
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+exit 0
```
elopio: nit, missing empty line at the end.

elopio:
I have no more suggestions about the code, so +1. waiting for all the tests to pass and @kyrofa's review to land.

Saviq:
This does seem to help when building [multipass](https://github.com/CanonicalLtd/multipass/) over `edge`, +1 here!

snapcraft/internal/elf.py
```diff
@@ -269,6 +273,8 @@ def wrapper(*args, **kwargs):
             # should eventually be removed once patchelf catches up.
             try:
                 elf_file_path = kwargs['elf_file_path']
+                logger.info('Failed to update {!r}. Retrying with '
+                            'different parameters.'.format(elf_file_path))
```
kyrofa: Would it be helpful to actually say what we're doing, here (removing the `.note.go.buildid` section)?

snapcraft/internal/errors.py
```diff
@@ -415,15 +415,43 @@ def __init__(self, image_info):
 
 
 class PatcherError(SnapcraftError):
+    pass
+
+
+class PatcherGenericError(PatcherError):
 
     fmt = (
         '{elf_file!r} cannot be patched to function properly as a classic '
```
kyrofa: This should say "[...] properly in a classic [...]"

snapcraft/internal/errors.py
```diff
@@ -415,15 +415,43 @@ def __init__(self, image_info):
...
+        message = '{} failed with exit code {}'.format(
+                ' '.join(process_exception.cmd),
+                process_exception.returncode)
         super().__init__(elf_file=elf_file, message=message)
 
 
+class PatcherNewerPatchelfError(PatcherError):
+
+    fmt = (
+        '{elf_file!r} cannot be patched to function properly as a classic '
```
kyrofa: You probably want "in" here as well.

This message overall doesn't really explicitly say why this failed, although it's alluded to at the end-- it goes straight from what happened to how to fix it without saying why what happened is a problem. Could we make it say something about "patchelf version {patchelf_version} is too old, Snapcraft requires at least version {good_patchelf_version}" ?

snapcraft/internal/elf.py
```diff
@@ -283,25 +289,37 @@ def wrapper(*args, **kwargs):
...
         # in by packaging dependencies.
         # The docker conditional will work if the docker image has the
         # snaps unpacked in the corresponding locations.
-        if common.is_snap():
+        if preferred_patchelf_path:
+            self._patchelf_cmd = preferred_patchelf_path
+        # We use the full path here as the path may not be set on
+        # build systems where the path is recently created and added
+        # to the environment
+        elif os.path.exists('/snap/bin/pathelf'):
```
kyrofa: I doubt you'll ever find the `pathelf` command :wink: .

Speaking of that, I could imagine it being helpful  in bug reports to, once we've decided which patchelf we're using, provide a debug print saying "using patchelf: {patchelf_path}".

sergiusens:
thanks @kyrofa ... this happens to me every time I do a `git clean ...` as it wipes out my config :angry:

PR #1878, opened 2018-01-17
repo: use debian.arfile instead of dpkg-deb

kalikiana:

-----

    $ cd snapcraft/tests/integration/snaps/python-hello
    $ SNAPCRAFT_CONTAINER_BUILDS=<remote> snapcraft
    tar: ./usr/bin/pdb3.5: Cannot utime: No such file or directory
    tar: ./usr/share/doc/python3.5/changelog.Debian.gz: Cannot utime: No such file or directory
    tar: Exiting with failure status due to previous errors
    dpkg-deb: error: subprocess tar returned error exit status 2
    Error while provisioning '/home/cris/build_python-hello/parts/python-
    part/ubuntu/download/python3.5_3.5.2-2ubuntu0~16.04.4_amd64.deb'

From my research it seems to be a bug concerning tar, which dpkg-deb is using, in combination with sshfs and the wrong libc. [See the references here](https://github.com/ValveSoftware/steam-for-linux/issues/2358).
We're already using debian.arfile in `snapcraft/internal/sources/_deb.py` which is doing pretty much the same thing, except it's not hitting the bug, so I'm porting repo to the same approach.

### DIFF
--- a/snapcraft/internal/repo/_deb.py
+++ b/snapcraft/internal/repo/_deb.py
@@ -24,11 +24,13 @@ import stat
 import string
 import subprocess
 import sys
+import tarfile
 import urllib
 import urllib.request
 from typing import Dict, Set, List  # noqa
 
 import apt
+import debian.arfile
 from xml.etree import ElementTree
 
 import snapcraft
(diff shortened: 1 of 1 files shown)


elopio:
It makes sense to me to use the same strategy everywhere. But what about extracting the duplicated code somewhere?

snapcraft/internal/repo/_deb.py
```diff
@@ -371,11 +373,15 @@ def _get(self, apt_cache):
...
         for pkg in pkgs_abs_path:
-            # TODO needs elegance and error control
+            # Importing DebFile causes LP: #1731478 when snapcraft is a snap
+            ar = debian.arfile.ArFile(pkg)
             try:
-                subprocess.check_call(
-                    ['dpkg-deb', '--extract', pkg, unpackdir])
-            except subprocess.CalledProcessError:
+                data_member_name = [
+                    i for i in ar.getnames() if i.startswith('data.tar')][0]
```
elopio: On the other version, IndexError is caught. Why is that ignored here?

kyrofa:
> what about extracting the duplicated code somewhere?

+1 to that, it feels like we should have a function to extract a deb somewhere.

Other than that, nice job here. Native is so much better than shelling out.

PR #1879, opened 2018-01-18
extractors: replace desktop file ids with paths

elopio:

-----
@kyrofa and I discussed about moving the task of translating from desktop file id to desktop file path to the appstream extractor. This opens the door to save in the metadata the contents of the file, instead of the path. But we haven't yet decided if that next step would be good.

### DIFF
--- a/snapcraft/extractors/appstream.py
+++ b/snapcraft/extractors/appstream.py
@@ -14,6 +14,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import os
+
 from ._metadata import ExtractedMetadata
 from snapcraft.extractors import _errors
 
@@ -53,13 +55,29 @@ def extract(path: str) -> ExtractedMetadata:
         # --elopio -20180109
         icon = node.text.strip()
 
-    desktop_file_ids = []
+    desktop_file_paths = []
     nodes = tree.findall('launchable')
     for node in nodes:
         if ('type' in node.attrib and
                 node.attrib['type'] == 'desktop-id'):
-            desktop_file_ids.append(node.text.strip())
+            desktop_file_id = node.text.strip()
+            desktop_file_path = _desktop_file_id_to_path(desktop_file_id)
+            if desktop_file_path:
+                desktop_file_paths.append(desktop_file_path)
 
     return ExtractedMetadata(
         summary=summary, description=description, icon=icon,
-        desktop_file_ids=desktop_file_ids)
+        desktop_file_paths=desktop_file_paths)
+
+
+def _desktop_file_id_to_path(desktop_file_id: str) -> str:
+    # For details about desktop file ids and their corresponding paths, see
+    # https://standards.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#desktop-file-id  # noqa
+    for xdg_data_dir in ('usr/local/share', 'usr/share'):
+        desktop_file_path = os.path.join(
+            xdg_data_dir, 'applications',
+            desktop_file_id.replace('-', '/'))
+        if os.path.exists(desktop_file_path):
+            return desktop_file_path
+    else:
+        return None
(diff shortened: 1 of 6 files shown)


snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -180,22 +176,23 @@ def _icon_file_exists() -> bool:
...
     --elopio - 20180108
 
-    :params desktop_file_id: The identifier of the desktop file.
+    :params desktop_file_path: The path to the desktop file.
     :returns: The name of the snap app that corresponds to the desktop file.
 
-    """  # noqa
+    """
+    desktop_file_id = desktop_file_path[
+        desktop_file_path.find('applications') +
```
kyrofa: If this happens to not be in the string, you'll get a -1. I don't feel like that will produce desired results, will it?

sergiusens:
How will this look when/if I do
```yaml
apps:
    my-app:
        common-id: my-app.desktop
```

snapcraft/extractors/appstream.py
```diff
@@ -53,13 +55,29 @@ def extract(path: str) -> ExtractedMetadata:
...
     return ExtractedMetadata(
         summary=summary, description=description, icon=icon,
-        desktop_file_ids=desktop_file_ids)
+        desktop_file_paths=desktop_file_paths)
+
+
+def _desktop_file_id_to_path(desktop_file_id: str) -> str:
+    # For details about desktop file ids and their corresponding paths, see
+    # https://standards.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#desktop-file-id  # noqa
+    for xdg_data_dir in ('usr/local/share', 'usr/share'):
```
kalikiana: Should this also take `share`? Snaps built from source may not use "usr" and that's where they end up by default when using eg. the CMake plugin.

snapcraft/internal/meta/_snap_packaging.py
kalikiana: `_desktop_file_id_to_path` would seem to force that. But I was wondering, do we want to ignore other desktop files entirely? Currently this isn't required to just use them as launchers.

kyrofa: Only if they're coming from appstream. We're making this change to make this easier to use from other metadata extraction methods (file paths instead of IDs), but this section seems to be pretty specific to appstream.

elopio: But remember this guessing will go away when we add the appstream id. If `applications` is not in the path, I would prefer just to skip instead of try another guess. But, no strong opinion here, I can go with your option.

elopio:
@sergiusens with common-id, this would mutate to be a dictionary instead of a list. And the part where we guess the app id from the desktop path will be removed.

snapcraft/extractors/appstream.py
elopio: I would say no. If the desktop file id doesn't follow the xdg rules, it is wrong. If we go further and add our own heuristics to find the files, it would be confusing.
Another option that we discuss was to try to find the files in all the directories. I would prefer that over adding more paths not in the standard.

kalikiana: Was there a reason not to go through all directories? Like .desktop files not being used for launchers? My main concern is that the layout in the snap doesn't necessarily look the same as in a Debian package and it may not be obvious to the user of Snapcraft where files will be found.

snapcraft/internal/meta/_snap_packaging.py
kyrofa: Alright, that's fine then. However, can you update the code to bail if applications isn't found instead of continuing the search with a -1? That will make it more clear that we don't intend for that to work (along with a comment, perhaps).

sergiusens:
You have conflicts @elopio ;-)

snapcraft/extractors/appstream.py
elopio: You can always specify the path on the snapcraft.yaml, using the `desktop` attribute. My thought here was to first do just the common and standard cases, not more. Then we can add complexity, doing a walk through all the files in pull, build or prime, if for some reason we require it.

PR #1880, opened 2018-01-19
Provide stub for when /etc/os-release doesn't exist

kalikiana:

-----
It seems there's a regression with running Snapcraft on MacOS (via PyPI). If no file `/etc/os-release` exists on the system, no commands will work and even the "no native builds" error will break.

    $ snapcraft
    [...]
    FileNotFoundError: [Errno 2] No such file or directory: '/etc/os-release'

This branch presents a possible approach for falling back to defaults. We could as a next step try and fill in versions from other systems here. Call sites won't need to be changed this way.

There's no new tests yet - early feedback would be great.

### DIFF
--- a/snapcraft/internal/os_release.py
+++ b/snapcraft/internal/os_release.py
@@ -46,8 +46,7 @@ class OsRelease:
                     if len(entry) == 2:
                         self._os_release[entry[0]] = entry[1].strip('"')
         except FileNotFoundError:
-            self._os_release['ID'] = 'unknown'
-            self._os_release['NAME'] = 'unknown'
+            pass
 
     def id(self) -> str:
         """Return the OS ID
--- a/snapcraft/internal/os_release.py
+++ b/snapcraft/internal/os_release.py
@@ -38,15 +38,13 @@ class OsRelease:
 
         :param str os_release_file: Path to os-release file to be parsed.
         """
-        self._os_release = {}  # type: Dict[str, str]
-        try:
+        with contextlib.suppress(FileNotFoundError):
+            self._os_release = {}  # type: Dict[str, str]
             with open(os_release_file) as f:
                 for line in f:
                     entry = line.rstrip().split('=')
                     if len(entry) == 2:
                         self._os_release[entry[0]] = entry[1].strip('"')
-        except FileNotFoundError:
-            pass
 
     def id(self) -> str:
         """Return the OS ID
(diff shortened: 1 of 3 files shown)


snapcraft/internal/os_release.py
```diff
@@ -38,12 +38,16 @@ def __init__(self, *, os_release_file: str = '/etc/os-release') -> None:
...
+        self._os_release = {}  # type: Dict[str, str]
+        try:
+            with open(os_release_file) as f:
+                for line in f:
+                    entry = line.rstrip().split('=')
+                    if len(entry) == 2:
+                        self._os_release[entry[0]] = entry[1].strip('"')
+        except FileNotFoundError:
+            self._os_release['ID'] = 'unknown'
+            self._os_release['NAME'] = 'unknown'
```
kyrofa: The accessor methods are designed to raise errors if the ID or NAME aren't specified, so it's probably not necessary to set these to "unknown". Just don't set anything at all if the file doesn't exist and the rest _should_ work out.

kalikiana: My goal was to have a fallback for the case where 1) `/etc/os-release` doesn't exist 2) `_is_deb_based` called from `repo/__init__.py` should return `False` 3) `_get_repo_for_platform` should fallback to the `DummyRepo`. The idea being, we could extend this in the future to populate the values from other sources.

That is, assuming we want `OsRelease` be our main entry point for OS delection. Otherwise we'd have to wrap this in a new API that's not based on a text file.

kyrofa: Hmm. I think I'd rather see `_is_deb_based` and `_get_repo_for_platform` catch the `OsReleaseIdError` rather than make `OsRelease` work when there _is_ no `/etc/os-release`. It seems to dilute its purpose.

snapcraft/internal/os_release.py
```diff
@@ -38,12 +38,15 @@ def __init__(self, *, os_release_file: str = '/etc/os-release') -> None:
...
-                    self._os_release[entry[0]] = entry[1].strip('"')
+        self._os_release = {}  # type: Dict[str, str]
+        try:
+            with open(os_release_file) as f:
+                for line in f:
+                    entry = line.rstrip().split('=')
+                    if len(entry) == 2:
+                        self._os_release[entry[0]] = entry[1].strip('"')
+        except FileNotFoundError:
+            pass
```
kyrofa: I kinda prefer `contextlib.suppress` for the type of operation that we know may toss an exception that we know we want to ignore.

PR #1881, opened 2018-01-23
elf: better handling for newer libc6

sergiusens:
Require that the snapcraft.yaml explicitly state it needs libc6 so that it is recorded and an explicit action from the developer.

-----

### DIFF
--- a/snapcraft/internal/errors.py
+++ b/snapcraft/internal/errors.py
@@ -459,7 +459,7 @@ class StagePackageMissingError(SnapcraftError):
     fmt = (
         '{package!r} is required as a `stage-packages` entry for this part to '
         'work properly.\n'
-        'Add {package!r} as a stage-packages entry for this part.'
+        'Add {package!r} as a `stage-packages` entry for this part.'
     )
 
     def __init__(self, *, package):
--- a/snapcraft/internal/errors.py
+++ b/snapcraft/internal/errors.py
@@ -454,6 +454,18 @@ class PatcherNewerPatchelfError(PatcherError):
                          patchelf_version=patchelf_version)
 
 
+class StagePackageMissingError(SnapcraftError):
+
+    fmt = (
+        '{package!r} is required as a `stage-packages` entry for this part to '
+        'work properly.\n'
+        'Add {package!r} as a stage-packages entry for this part.'
+    )
+
+    def __init__(self, *, package):
+        super().__init__(package=package)
+
+
 class MetadataExtractionError(SnapcraftError):
     pass
 
--- a/snapcraft/internal/errors.py
+++ b/snapcraft/internal/errors.py
@@ -457,9 +457,9 @@ class PatcherNewerPatchelfError(PatcherError):
 class StagePackageMissingError(SnapcraftError):
 
     fmt = (
-        '{package!r} is required as a `stage-packages` entry for this part to '
-        'work properly.\n'
-        'Add {package!r} as a `stage-packages` entry for this part.'
+        '{package!r} is required inside the snap for this part to work '
+        'properly.\n'
+        'Add it as a `stage-packages` entry for this part.'
     )
 
     def __init__(self, *, package):
--- a/snapcraft/internal/mangling.py
+++ b/snapcraft/internal/mangling.py
@@ -129,7 +129,7 @@ def handle_glibc_mismatch(*, elf_files: FrozenSet[elf.ElfFile],
 
     # Before doing anything else, verify there's a dynamic linker we can use.
     dynamic_linker_path = os.path.join(
-        snap_base_path, dynamic_linker[1:])
+        snap_base_path, dynamic_linker[len(root_path)+1:])
 
     elf_patcher = elf.Patcher(dynamic_linker=dynamic_linker_path,
                               root_path=root_path,
--- a/snapcraft/internal/mangling.py
+++ b/snapcraft/internal/mangling.py
@@ -76,6 +76,8 @@ def handle_glibc_mismatch(*, elf_files: FrozenSet[elf.ElfFile],
     """Copy over libc6 libraries from the host and patch necessary elf files.
 
     If no newer glibc version is detected in elf_files, this function returns.
+    The dynamic linker and related libraries to libc6 are expected to be found
+    in root_path.
 
     :param snapcraft.internal.elf.ElfFile elf_files:
         set of candidate elf files to patch if a newer libc6 is required.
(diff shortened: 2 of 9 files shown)


snapcraft/internal/errors.py
```diff
@@ -454,6 +454,18 @@ def __init__(self, *, elf_file, process_exception, patchelf_version):
...
 
+class StagePackageMissingError(SnapcraftError):
+
+    fmt = (
+        '{package!r} is required as a `stage-packages` entry for this part to '
+        'work properly.\n'
+        'Add {package!r} as a stage-packages entry for this part.'
+    )
+
+    def __init__(self, *, package):
```
kyrofa: This seems nice and forward-thinking when it comes to removing the automatic crawling (I assume you intend to re-use this exception), however, for that use-case we won't have package names, but library paths. As a result, I only ever see this being used for libc. Perhaps this exception doesn't need to be so generic?

snapcraft/internal/mangling.py
```diff
@@ -113,25 +113,25 @@ def handle_glibc_mismatch(*, elf_files: FrozenSet[elf.ElfFile],
                    'base. These are the GLIBC versions required by '
                    'the primed files that do not match and will be '
                    'patched:\n {}\n'
-                   'To work around this, the newer libc will be '
-                   'migrated into the snap, and these files will be '
-                   'patched to use it.'.format('\n'.join(formatted_list)))
+                   'For this to work, `libc6` will need to be explicitly '
+                   'added to `stage-packages` for this part. These files '
+                   'will be patched to use it.'.format(
+                       '\n'.join(formatted_list)))
```
kyrofa: Hmm... this is a warning, but it's phrased like an error. Should it be an error instead? If not, and libc is actually staged, is there still a reason to show this warning?

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -564,6 +564,9 @@ def prime(self, force=False) -> None:
         linker_compatible = (e.is_linker_compatible(linker='ld-2.23.so')
                              for e in elf_files)
         if not all((x for x in linker_compatible)):
+            if 'libc6' not in self.stage_packages:
+                raise errors.StagePackageMissingError(package='libc6')
```
kyrofa: Okay so we have an error here, but the patching text above is just a warning.... there's a disconnect here.

snapcraft/internal/errors.py
sergiusens: the other case that may come in the future is when the new openssl comes in

snapcraft/internal/mangling.py
sergiusens: this is left over, the caller of this method resolves that doubt. I will add appropirate docstrings

snapcraft/internal/pluginhandler/__init__.py
sergiusens: When you say above, I hope you mean above in this review screen. Then yes, I will take care of that.

sergiusens:
thanks for the review @kyrofa; the logic you shared is true if the package is not explicitly called out (which is what we are requiring).

sergiusens:
I will address your comments during my morning

snapcraft/internal/mangling.py
```diff
@@ -112,26 +114,22 @@ def handle_glibc_mismatch(*, elf_files: FrozenSet[elf.ElfFile],
...
 
-    dest_dir = os.path.join(root_path, libc6_path)
-    os.makedirs(dest_dir, exist_ok=True)
+    dynamic_linker = _get_dynamic_linker(libc6_libraries_paths)
 
-    for src in libc6_libraries:
-        dst = os.path.join(dest_dir, os.path.basename(src))
-        # follow_symlinks is set to True for elf crawling to work.
-        file_utils.link_or_copy(src, dst, follow_symlinks=True)
+    # Before doing anything else, verify there's a dynamic linker we can use.
```
elopio: This comment is misleading, because we are not verifying anything here, we are just joining a path.

snapcraft/internal/errors.py
```diff
@@ -454,6 +454,18 @@ def __init__(self, *, elf_file, process_exception, patchelf_version):
                          patchelf_version=patchelf_version)
 
 
+class StagePackageMissingError(SnapcraftError):
+
+    fmt = (
+        '{package!r} is required as a `stage-packages` entry for this part to '
+        'work properly.\n'
+        'Add {package!r} as a `stage-packages` entry for this part.'
```
kyrofa: This doesn't flow super well, it feels like it's being shoved into our new error structure instead of massaging it :stuck_out_tongue: .

How about:

```
'{package!r} is required inside the snap for this part to work properly.\n'
'Add it as a `stage-packages` entry for this part.'
```

sergiusens: lol, ok, I just wished you raised this earlier as the PR is finally green ;-)

sergiusens: is this the last tidbit on your side?

snapcraft/internal/mangling.py
```diff
@@ -112,26 +114,25 @@ def handle_glibc_mismatch(*, elf_files: FrozenSet[elf.ElfFile],
                    'files and the linker version (2.23) used in the '
                    'base. These are the GLIBC versions required by '
                    'the primed files that do not match and will be '
-                   'patched:\n {}\n'
-                   'To work around this, the newer libc will be '
-                   'migrated into the snap, and these files will be '
-                   'patched to use it.'.format('\n'.join(formatted_list)))
+                   'patched:\n{}\n'.format('\n'.join(formatted_list)))
```
kyrofa: Once folks hit the "Hey! Stage libc!" error, does this warning add any meaning? Would it be better to show (essentially) this information only if libc isn't staged, and just trudge along and patch if they've staged it?

sergiusens: I think you have the process backwards; first we check for libc6 and only then do we list what needs to be patched.

sergiusens: I do think it is valuable, it is essentially modifying binaries that would not be modified otherwise

kyrofa: The order doesn't matter, I just question the value-added of this warning once they're already said "oh right right, stage libc". You're right though, this is modifying binaries, we should probably continue to show it.

sergiusens:
Got a :+1: from @elopio (_will keep on testing afterwards_) and @kyrofa (error message change was the last thing).

elopio:
I tested building a python devmode snap in bionic with your branch. Without libc6 in stage-packages, it failed during prime telling me to add it. I added the stage packages, and it finished building.

elopio:
@sergiusens just to confirm that this behaviour is the expected:

on bionic, with your branch and libc6 as stage-package, when I run the app it segfaults. With master, without libc6 as stage-package, it runs.

sergiusens:
@elopio the segfaulting part really depends on what you are snapping up and the potential need of using `adapter: none` under `apps`

PR #1890, opened 2018-01-29
Fixed typo

chrisglass:
This is a one character typo fix in a help string, I'm not sure how relevant that checklist is :)
Being a Canonical employee covers the CLA question.


sergiusens:
Thanks, these fixes are not copyrightable either so no need to worry about a CLA regardless.

PR #1892, opened 2018-01-30
meta: parse float version as string

kalikiana:

-----
[LP: #1669291](https://bugs.launchpad.net/snapd/+bug/1669291)

### DIFF
--- a/snapcraft/internal/meta/_snap_packaging.py
+++ b/snapcraft/internal/meta/_snap_packaging.py
@@ -402,9 +402,6 @@ class _SnapPackaging:
         snap_yaml['version'] = _version.get_version(
             self._config_data['version'],
             self._config_data.get('version-script'))
-        if not isinstance(snap_yaml['version'], str):
-            logger.warn('"version" property is not a string: {!r}'.format(
-                        snap_yaml['version']))
 
         for key_name in _OPTIONAL_PACKAGE_KEYS:
             if key_name in self._config_data:
--- a/snapcraft/internal/meta/_snap_packaging.py
+++ b/snapcraft/internal/meta/_snap_packaging.py
@@ -402,6 +402,9 @@ class _SnapPackaging:
         snap_yaml['version'] = _version.get_version(
             self._config_data['version'],
             self._config_data.get('version-script'))
+        if not isinstance(snap_yaml['version'], str):
+            logger.warn('"version" property is not a string: {!r}'.format(
+                        snap_yaml['version']))
 
         for key_name in _OPTIONAL_PACKAGE_KEYS:
             if key_name in self._config_data:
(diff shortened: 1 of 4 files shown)


sergiusens:
I am not so sure about just a warning, as those have been proven to be ignored by most. That said it is still better that doing nothing (although I strongly believe we should enforce it to be a string).

Aside from that my only comment would be that this should probably better be served by a unit test instead of an integration one.

snapcraft/tests/integration/snaps/version/snap/snapcraft.yaml
```diff
@@ -0,0 +1,9 @@
+name: version-test
+version: 2.10
+summary: test version string
+description: |
+  The version should be parsed as a string.
```
kyrofa: There doesn't actually appear to be a test using that, is that true?

I agree with @sergiusens though, there are plenty of unit tests covering this stuff, might as well add one more case there.

snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -402,6 +402,9 @@ def _compose_snap_yaml(self):
         snap_yaml['version'] = _version.get_version(
             self._config_data['version'],
             self._config_data.get('version-script'))
+        if not isinstance(snap_yaml['version'], str):
+            logger.warn('"version" property is not a string: {!r}'.format(
```
kyrofa: Why do I as a random user care if it's a string or not?

kyrofa:
> although I strongly believe we should enforce it to be a string

That's an easy schema change, but of course, not backward compatible. Would it be worth deprecating this instead?

snapcraft/internal/meta/_snap_packaging.py
kalikiana: I changed my approach to parsing float as string instead of a warning.

snapcraft/internal/project_loader/_config.py
```diff
@@ -260,7 +268,11 @@ def _snapcraft_yaml_load(yaml_file):
 
     try:
         with open(yaml_file, encoding=encoding) as fp:
-            return yaml.load(fp)
+            # Override the loader here so we can parse float as str
+            # which matters when eg. having "version: 2.10"
+            Loader.add_constructor(
+                u'tag:yaml.org,2002:float', float_constructor)
```
kyrofa: Doesn't this override the loading of _every_ float to be a string instead? Are we sure that's what we want? Remember that this will affect custom/local plugins as well. If this is the direction we want to go, perhaps this change should be more localized.

sergiusens: If this happens, it should be only for `version`; I would rather have a proper schema check, just like we change the on in `version` anyways (allowed/disallowed chars)

sergiusens:
I am closing this for now until we have a good conversation around it.

PR #1894, opened 2018-01-30
tests: allow to overwrite the snapcraft install command

elopio:

-----
The snap store team wants to run only one stage. They can define their stage and the script to run through the api. However, in here we have a dependency on a previous stage, so it wouldn't be possible to call `sudo ./tools/travis/run_tests.sh snapcraft/tests/integration/store` independently.
Their tests don't need to build the snapcraft snap from source, so we could just install snapcraft from edge.
I am adding here an optional env var SNAPCRAFT_INSTALL_COMMAND, that the store team can send on their API request set to `sudo snap install snapcraft --edge --classic`.


elopio:
Reviewing @cprov...

elopio:
@cprov I triggered your script, adding the new env var, and look at it:

https://travis-ci.org/snapcore/snapcraft/builds/335899780#L1580

PR #1900, opened 2018-01-31
many: handle copying of symlink permissions gracefully

kalikiana:

-----
Fixes: [LP: #1744921](https://bugs.launchpad.net/snapcraft/+bug/1744921)

Manual test steps:
    > # Install Snapcraft snap from this branch
    > export SNAPCRAFT_CONTAINER_BUILDS=<remote>
    > cd snapcraft/tests/integration/snaps/symlinks/
    > snapcraft

My proposed fix is to replace `shutil.copytree` with our own version to achieve two things:

- Don't swallow errors, which makes both handling and debugging them difficult
- Suppress the "not found" error that may occur with copying of symlink permissions

Note: Testing this is a bit involved because it only occurs when Snapcraft is run inside a container inside an sshfs mount. We don't currently have any integration tests for this setup.


kyrofa:
Rewriting copytree feels like overkill. What is the actual _problem_ here? And why does it only happen for sshfs? All I see on the bug are errors, no triaging.

PR #1901, opened 2018-01-31
elf: check if file exists before checking if ELF

sergiusens:
LP: #1746397

-----
I personally do not see this issue, nor do the tests which build the example for the dotnet plugin which exposes this for some folks.

That said, this check does not hurt.


elopio:
I can't reproduce this in a clean machine either. It's weird that it happens every team on my dev box and on Rakesh's machine all the time. Maybe it's caused by installing, refreshing and reinstalling many times snapcraft.
I would say this change doesn't hurt, so +1. But it's also not clear what it is fixing. I would also be ok with closing it and waiting for the bug to be clearly reproducible.

PR #1905, opened 2018-02-01
Remove dangling symlink from JDK plugin (LP Bug #1617296)

ARostovsky:
Using the JDK plugin results a dangling symlink error

package contains external symlinks:
usr/lib/jvm/java-8-openjdk-amd64/jre/lib/security/cacerts

This pull request adds 'ca-certificates-java' package to the build stage. So at the build part it will just take needed file from symlink and that's it.
LP Bug #1617296

Previous bull requests about this bug:
https://github.com/snapcore/snapcraft/pull/761 (cc @gnouy)


sergiusens:
Thanks, we however should not need this anymore after the refactor for bases

PR #1906, opened 2018-02-01
remote_parts: handle connection errors

kyrofa:

-----

Currently if Snapcraft can't reach the remote parts URL for any reason, it prints a traceback from requests. This PR fixes LP: [#1746784](https://bugs.launchpad.net/snapcraft/+bug/1746784) by handling this in a nicer manner, and suggesting a simple retry.


snapcraft/internal/errors.py
```diff
@@ -239,6 +243,17 @@ class PartNotInCacheError(SnapcraftError):
     )
 
 
+class RemotePartsUpdateConnectionError(RemotePartsError):
+
+    fmt = (
+        'Failed to update cache of remote parts: {message}\n'
+        'Please try again.'
```
elopio: Shouldn't this say the command to run?

kyrofa: Unfortunately we have no clue what command was run! This could be caused by `snapcraft update`, or just `snapcraft`, or `snapcraft pull`...

kalikiana: How about using [click.get_current_context](http://click.pocoo.org/5/api/#click.get_current_context) here?

kyrofa: Meh, probably, but it still seems a little silly to give the user the exact command they just ran instead of just saying "try again." Do either of you have strong feelings about it?

kalikiana: I regularly make the mistake of thinking "this is obvious" in such cases when for a typical user it isn't. So I'd err on the side of being more specific to make sure they know how to proceed.

kyrofa: I dug into this, and it quickly got silly. We have no context in tests, which means we need to have a special case for when we have no context here, and fall back to a default message. There also seems to be no way to reconstruct what is essentially `sys.argv` from the Click context or command, but using `sys.argv` is also out of the question since that changes depending on tests running as well.

kyrofa: I have pushed a change using argv for you to mull over.

kyrofa: After talking with @elopio, we agreed to revert that commit and log LP: [#1749010](https://bugs.launchpad.net/snapcraft/+bug/1749010) instead.

PR #1908, opened 2018-02-02
tests: update tests to work in adt

sergiusens:
add patchelf_command attribute in SnapsTestCase and add the patchelf
and add the build-snaps dependency to gotty.

-----


snapcraft/tests/unit/project_loader/grammar_processing/test_part_grammar_processor.py
```diff
@@ -29,7 +29,8 @@
 
 
 def load_tests(loader, tests, ignore):
-    tests.addTests(doctest.DocTestSuite(processor))
+    if snapcraft.ProjectOptions().deb_arch == 'amd64':
+        tests.addTests(doctest.DocTestSuite(processor))
```
kalikiana: Nice idea actually.

An alternative I was contemplating is to subclass `DocTestSuite` and add a mock there.

kyrofa: The `on` statement [does this differently](https://github.com/snapcore/snapcraft/blob/master/snapcraft/internal/project_loader/grammar/_on.py#L46). That example in the processor should probably be doing something similar.

kalikiana: I apparently only voiced this in the chat, and not here, so I'll re-iterate my thoughts for completeness:

To my mind mocking the architecture is counter-intuitive in example code like this. You wouldn't use the mock in production but instead want it to behave differently. That's why I think handling this in `load_tests` is more sensible.

snapcraft/tests/integration/general/test_build_properties.py
```diff
@@ -51,6 +51,8 @@ def assert_expected_build_state(self, project_dir):
         self.assertThat(state.properties['stage-packages'], Equals(['curl']))
 
     def test_build_with_arch(self):
+        if self.deb_arch == 'armhf':
+            self.skipTest('For now, we just support crosscompile from amd64')
```
kalikiana: Perhaps I'm nit-picking here, but this seems a bit wrong. The condition will let the test run on anything that's not armhf.

So either it should be
-  `if self.deb_arch == 'armhf': self.skipTest('For now we're not testing cross-compiling from armhf')`
or
 - `if self.deb_arch == 'amd64': self.skipTest('For now we're only testing cross-compiling from amd64')`

Note that I'd avoid the term "support" here because this is only about testing.

elopio:
armhf failures: 
* test_stage (test_autotools_plugin.AutotoolsPluginTestCase): Max retries exceeded with url: http://ftp.gnu.org/gnu/hello/hello-2.10.tar.gz
* test_shared_ros_builds_without_catkin_in_underlay (snapcraft.tests.integration.plugins.catkin.test_catkin.CatkinTestCase): RuntimeError: Error updating rosdep database:
reading in sources list data from /home/ubuntu/autopkgtest_tmp/tmp3tqyus1p/parts/underlay/rosdep/sources.list.d
* demos_tests.test_qt4_text_editor.Qt4TextEditorTestCase.test_text_editor: Temporary failure resolving ftpmaster.internal
* demos_tests.test_shared_ros.SharedROSTestCase.test_shared_ros: Temporary failure in name resolution> https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/ruby.yaml

The only weird one is test_shared_ros_builds_without_catkin_in_underlay. @kyrofa have you seen that error updating database before?

kyrofa:
> The only weird one is test_shared_ros_builds_without_catkin_in_underlay. @kyrofa have you seen that error updating database before?

Yeah looks like the connection flaked is all:

> Updating rosdep database...
> ERROR: error loading sources list:
>	<urlopen error <urlopen error [Errno -3] Temporary failure in name resolution> (https://raw.githubusercontent.com/ros/rosdistro/master/jade/distribution.yaml)>

elopio:
amd64 timed out, only in the ROS tests: integrationtests-snapd-slow FAIL timed out
I've told Steve to increase the timeout for that one on the archive tests.

elopio:
arm64 also timed out on integrationtests-snapd-slow. Hopefully, the bump on the timeout will be enough to make them pass.

elopio:
ship it?

snapcraft/tests/integration/general/test_build_properties.py
sergiusens: Is it only about testing though?

PR #1909, opened 2018-02-02
dotnet plugin: add dotnet command to path and enable developer scenario for  snap

rakeshsinghranchi:
Add dotnet command to path and enable developer scenario for  snap ild via application yaml

-----

### DIFF
--- a/snapcraft/plugins/dotnet.py
+++ b/snapcraft/plugins/dotnet.py
@@ -140,3 +140,9 @@ class DotNetPlugin(snapcraft.BasePlugin):
             if fnmatch.fnmatch(file, '*.??proj'):
                 return os.path.splitext(file)[0]
                 break
+
+    def env(self, root):
+        if root == self.installdir:
+            return ['PATH={}:$PATH'.format(self._dotnet_sdk_dir)]
+        else:
+            return []
--- a/snapcraft/plugins/dotnet.py
+++ b/snapcraft/plugins/dotnet.py
@@ -142,6 +142,7 @@ class DotNetPlugin(snapcraft.BasePlugin):
                 break
 
     def env(self, root):
+        # Update the PATH only during the Build and Install step
         if root == self.installdir:
             return ['PATH={}:$PATH'.format(self._dotnet_sdk_dir)]
         else:
--- a/tests/unit/plugins/test_dotnet.py
+++ b/tests/unit/plugins/test_dotnet.py
@@ -95,6 +95,12 @@ class DotNetProjectBaseTestCase(unit.TestCase):
 
 class DotNetProjectTestCase(DotNetProjectBaseTestCase):
 
+    def test_sdk_in_path(self):
+        plugin = dotnet.DotNetPlugin(
+            'test-part', self.options, self.project)
+        self.assertThat(plugin.env(plugin.installdir), Contains(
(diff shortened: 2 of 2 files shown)


snapcraft/plugins/dotnet.py
```diff
@@ -140,3 +140,9 @@ def _get_appname(self):
             if fnmatch.fnmatch(file, '*.??proj'):
                 return os.path.splitext(file)[0]
                 break
+
+    def env(self, root):
+        if root == self.installdir:
```
kyrofa: This logic is _slightly_ unusual. Would you mind adding a comment specifying why you only want this done for the installdir?

rakeshsinghranchi:
Hi @elopio  ,  can you please take a look at the failure CI failure? Looks like unrelated to my change. How do I retry the test run ?

sergiusens:
@rakeshsinghranchi I am rerunning with updates that fix that.
Cheers and thanks for following up.

snapcraft/plugins/dotnet.py
```diff
@@ -140,3 +140,10 @@ def _get_appname(self):
             if fnmatch.fnmatch(file, '*.??proj'):
                 return os.path.splitext(file)[0]
                 break
+
+    def env(self, root):
+        # Update the PATH only during the Build and Install step
```
kyrofa: Thanks for adding the comment, but it still leaves me asking: why? Why not during the pull step as well?

sergiusens: this should be ok (although overriding pull is not possible).

kyrofa: I'm still confused why this special case is needed. It doesn't matter if the dir doesn't exist:

```
$ ls -l /tmp/foo
ls: cannot access '/tmp/foo': No such file or directory
$ export PATH=/tmp/foo:$PATH
$ which footest
$ mkdir /tmp/foo
$ touch /tmp/foo/footest
$ chmod a+x /tmp/foo/footest 
$ which footest
/tmp/foo/footest
```

tests/unit/plugins/test_dotnet.py
```diff
@@ -95,6 +95,12 @@ def side_effect(cmd, *args, **kwargs):
 
 class DotNetProjectTestCase(DotNetProjectBaseTestCase):
 
+    def test_sdk_in_path(self):
+        plugin = dotnet.DotNetPlugin(
+            'test-part', self.options, self.project)
+        self.assertThat(plugin.env(plugin.installdir), Contains(
+            'PATH={}:$PATH'.format(plugin._dotnet_sdk_dir)))
```
kyrofa: If we keep the "only set path if root is installdir" then we should probably test that scenario as well.

kyrofa: If we keep the "only set path if root is installdir" then we should probably test that scenario as well.

snapcraft/plugins/dotnet.py
kyrofa: Ah, I am indeed missing something-- this is to avoid having this PATH definition make it into the final snap. We need to support this better, but this is a good workaround for now, so you can ignore me.

rakeshsinghranchi:
Thanks @kalikiana and @kyrofa for getting this in.

PR #1910, opened 2018-02-02
tests: expect in-snap unsquashfs when using docker

kyrofa:

---

This test isn't correct, but it's possible for it to pass if conditions are right. Fix the expectations to make it consistent (hat tip to @rakeshsinghranchi).


kyrofa:
Argh test failure, I'll fix it in the morning.

sergiusens:
But if we go with my approach, let's just put it in the techdebt backlog.

kyrofa:
Yeah this is a pretty wildly complex test, but it mirrors some of the other ones we have for the same type of thing. I wouldn't mind restructuring that at all, but you're right, that makes it a technical debt item. Created #1932.

PR #1911, opened 2018-02-02
Add support enable configurable runtime version for .NET Core applications

rakeshsinghranchi:

-----

### DIFF
--- a/snapcraft/plugins/dotnet.py
+++ b/snapcraft/plugins/dotnet.py
@@ -100,9 +100,11 @@ class DotNetPlugin(snapcraft.BasePlugin):
     def _get_sdk(self):
         sdk_arch = self.project.deb_arch
         if sdk_arch not in _SDK_ARCH:
+            # TODO: Add an exception class for this error
             raise NotImplementedError(
                 'This plugin does not support architecture '
-                '{}'.format(sdk_arch))
+                '{!r}'.format(sdk_arch))
+        # TODO: Make this a class that takes care of retrieving the infos
         sdk_info = self._get_sdk_info(self.options.dotnet_runtime_version)
 
         sdk_url = sdk_info['package_url']
@@ -152,7 +154,7 @@ class DotNetPlugin(snapcraft.BasePlugin):
 
     def _get_version_metadata(self, version):
         jsonData = self._get_dotnet_release_metadata()
-        package_data = list(filter(lambda x: x['version-runtime']
+        package_data = list(filter(lambda x: x.get('version-runtime')
                             == version, jsonData))
 
         if not package_data or len(package_data) < 1:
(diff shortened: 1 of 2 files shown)


rakeshsinghranchi:
@elopio  , can you please help looking at the integration test failures ? I have performed static test validation?
cc @sergiusens

elopio:
Hey @rakeshsinghranchi, the failures are on the unit tests. You can see the results here: https://travis-ci.org/snapcore/snapcraft/jobs/336410787
And you can reproduce this errors locally, take a look at the TESTING.md file.

rakeshsinghranchi:
@kalikiana , Thanks so much for helping out here! Much appreciated.

snapcraft/plugins/dotnet.py
```diff
@@ -58,11 +52,28 @@ class DotNetPlugin(snapcraft.BasePlugin):
...
+    def get_pull_properties(cls):
+        # Inform Snapcraft of the properties associated with pulling. If these
+        # change in the YAML Snapcraft will consider the build step dirty.
+        return ['dotnet-runtime-version']
+
+    @classmethod
+    def get_build_properties(cls):
+        # Inform Snapcraft of the properties associated with building. If these
+        # change in the YAML Snapcraft will consider the build step dirty.
+        return ['dotnet-runtime-version']
```
kyrofa: There's no need to specify both of these-- if they effect both, just put them in `get_pull_properties`. Marking the pull step dirty marks the build step dirty as well.

snapcraft/plugins/dotnet.py
```diff
@@ -140,3 +151,64 @@ def _get_appname(self):
             if fnmatch.fnmatch(file, '*.??proj'):
                 return os.path.splitext(file)[0]
                 break
+
+    def _get_version_metadata(self, version):
+        jsonData = self._get_dotnet_release_metadata()
+        package_data = list(filter(lambda x: x.get('version-runtime')
+                            == version, jsonData))
+
+        if not package_data or len(package_data) < 1:
```
kyrofa: This is redundant: just use `if not package_data`. Python's [truth value testing](https://docs.python.org/3/library/stdtypes.html#truth) takes the length into account.

snapcraft/plugins/dotnet.py
```diff
@@ -140,3 +151,64 @@ def _get_appname(self):
...
+            print('error occurred while fetching  the version details'
+                  'or the version specified is incorrect')
```
kyrofa: This will just carry on its merry way-- should an error be raised instead to halt things? An error that preferably inherits from `SnapcraftError`. Otherwise we'll just get an out-of-bounds error [here](https://github.com/snapcore/snapcraft/pull/1911/files#diff-5d24e70cc05a6003012e19fc045d6592R179).

snapcraft/plugins/dotnet.py
```diff
@@ -140,3 +151,64 @@ def _get_appname(self):
...
+                hash = text[1]
+
+            if len(text) == 2 and filename in text[1]:
+                checksum = text[0]
+                break
+
+        return '{}/{}'.format(hash.lower(), checksum)
+
+    def env(self, root):
+        if root == self.installdir:
```
kyrofa: This will conflict with the other .net PR, and I took issue with this part, but I won't cover it again here.

snapcraft/plugins/dotnet.py
kalikiana: Indeed. I didn't want to blow up this PR but I went ahead and added classes for the errors.

snapcraft/plugins/dotnet.py
kalikiana: I completely overlooked this being in here. Thanks for catching it!

PR #1912, opened 2018-02-02
lxd: unset SNAP to work-around LXD deb thinking it's a snap

kalikiana:

-----
Work-around for [LP: #1746612](https://bugs.launchpad.net/snapcraft/+bug/1746612)

Manual test steps:
    1. Have LXD v2.21 deb installeed
    2. `$ snapcraft cleanbuild`
    3. Observe that the build succeeds

### DIFF
--- a/snapcraft/internal/lxd/_containerbuild.py
+++ b/snapcraft/internal/lxd/_containerbuild.py
@@ -75,6 +75,8 @@ class Containerbuild:
         self._lxd_common_dir = os.path.expanduser(
             os.path.join('~', 'snap', 'lxd', 'common'))
         os.makedirs(self._lxd_common_dir, exist_ok=True)
+        # Work-around for https://github.com/lxc/lxd/issues/4183
+        os.environ['SNAP'] = ''
 
     @contextmanager
     def _container_running(self):


snapcraft/internal/lxd/_containerbuild.py
```diff
@@ -75,6 +75,8 @@ def __init__(self, *, output, source, project_options,
         self._lxd_common_dir = os.path.expanduser(
             os.path.join('~', 'snap', 'lxd', 'common'))
         os.makedirs(self._lxd_common_dir, exist_ok=True)
+        # Work-around for https://github.com/lxc/lxd/issues/4183
+        os.environ['SNAP'] = ''
```
kyrofa: This feels a little heavy-handed, would it not be better to do this temporarily just whenever we shell out to lxc?

sergiusens:
Let's not do a workaround. This will probably take more to release than a fix in `LXD`. That said, I don't mind if we have a common run handler that takes care of unsetting these environment variables (we should design this first and not go straight to PR).

basak:
> This will probably take more to release than a fix in LXD.

I'm affected by this since I'm using the lxd deb (from xenial-updates) and the snapcraft snap. Debs are harder to get landed than snaps.

In the lxd issue, Stéphane says it'll take a couple of weeks. But that was just for xenial-backports, which isn't installed by default. How long is it acceptable for users to remain broken on Xenial by default?

basak:
This is still broken. Please land the workaround. Right now, `snapcraft cleanbuild` still fails *by default* on Xenial (since the lxd deb is default on Xenial).

> Let's not do a workaround. This will probably take more to release than a fix in LXD.

This is quite clearly untrue at this point.

PR #1913, opened 2018-02-02
elf: pyelftools to parse ELF files rather than readelf

jhenstridge:
Update the `snapcraft.internal.elf.ElfFile` class to use pyelftools instead of shelling out to readelf.  This is faster, and doesn't rely on parsing the output of a utility that is intended for humans.


kyrofa:
We'll probably need to add pyelftools to both `debian/control` as well as `snap/snapcraft.yaml`

sergiusens:
I like this so far!

snapcraft/internal/elf.py
```diff
@@ -107,77 +118,49 @@ def __init__(self, *, path: str) -> None:
         """
         self.path = path
         self.dependencies = set()  # type: Set[Library]
-        self.interp, self.symbols = self._extract_readelf(path)
+        self.interp, self.soname, self.needed = self._extract(path)
 
-    def _extract_readelf(self, path) -> Tuple[str, List[Symbol]]:
+    def _extract(self, path) -> Tuple[str, str, Dict[str, NeededLibrary]]:  # noqa: C901,E501
```
kyrofa: Well this seems a bit like cheating. Especially C901.

kyrofa:
I can't pretend to understand all this, but I can say I've had no failures so far, everything seems to be working the same as before. I'll try a benchmark now to see what kind of savings we get.

kyrofa:
Hmm, I'm getting an error priming the ros-talker-listener integration test:

```
Priming ros-project 
  File "/home/kyrofa/venv/snapcraft/bin/snapcraft", line 9, in <module>
    load_entry_point('snapcraft', 'console_scripts', 'snapcraft')()
  File "/home/kyrofa/venv/snapcraft/lib/python3.5/site-packages/pkg_resources/__init__.py", line 542, in load_entry_point
  File "/home/kyrofa/venv/snapcraft/lib/python3.5/site-packages/pkg_resources/__init__.py", line 2569, in load_entry_point
    return ep.load()
  File "/home/kyrofa/venv/snapcraft/lib/python3.5/site-packages/pkg_resources/__init__.py", line 2229, in load
    return self.resolve()
  File "/home/kyrofa/venv/snapcraft/lib/python3.5/site-packages/pkg_resources/__init__.py", line 2235, in resolve
  File "/home/kyrofa/src/snapcraft/snapcraft/cli/__main__.py", line 43, in <module>
    run(prog_name='snapcraft')
  File "/home/kyrofa/venv/snapcraft/lib/python3.5/site-packages/click/core.py", line 722, in __call__
  File "/home/kyrofa/venv/snapcraft/lib/python3.5/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/home/kyrofa/venv/snapcraft/lib/python3.5/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/kyrofa/venv/snapcraft/lib/python3.5/site-packages/click/core.py", line 895, in invoke
  File "/home/kyrofa/venv/snapcraft/lib/python3.5/site-packages/click/core.py", line 535, in invoke
  File "/home/kyrofa/src/snapcraft/snapcraft/cli/lifecycle.py", line 111, in prime
    _execute('prime', parts, **kwargs)
  File "/home/kyrofa/src/snapcraft/snapcraft/cli/lifecycle.py", line 33, in _execute
    lifecycle.execute(command, project_options, parts)
  File "/home/kyrofa/src/snapcraft/snapcraft/internal/lifecycle/_runner.py", line 79, in execute
    _Executor(config, project_options).run(step, part_names)
  File "/home/kyrofa/src/snapcraft/snapcraft/internal/lifecycle/_runner.py", line 185, in run
    self._run_step(step, part, part_names)
  File "/home/kyrofa/src/snapcraft/snapcraft/internal/lifecycle/_runner.py", line 222, in _run_step
    getattr(part, step)()
  File "/home/kyrofa/src/snapcraft/snapcraft/internal/pluginhandler/__init__.py", line 510, in prime
    elf_files = elf.get_elf_files(self.primedir, snap_files)
  File "/home/kyrofa/src/snapcraft/snapcraft/internal/elf.py", line 522, in get_elf_files
    elf_file = ElfFile(path=path)
  File "/home/kyrofa/src/snapcraft/snapcraft/internal/elf.py", line 121, in __init__
    self.interp, self.soname, self.needed = self._extract(path)
  File "/home/kyrofa/src/snapcraft/snapcraft/internal/elf.py", line 157, in _extract
    lib = libs[library_name]
KeyError: 'libpthread.so.0'
```

snapcraft/internal/elf.py
jhenstridge: I initially tried wrapping the line, and pyflakes ignored the annotation: it seems that it must appear on the same line as the beginning of the method definition.

As for the method itself, it was simpler before adding the str/bytes checks to support xenial's old pyelftools.  I'll see what I can do.

kyrofa:
> So there's basically two ways we could go here:
>
> 1. ignore versions for libraries not in DT_NEEDED.
> 2. add the missing NeededLibrary when we encounter the version.
>
> I'm not sure which is better. The runtime linker doesn't consider it a (direct) dependency, but if the symbol is used, it will need to be provided with the given version.

Given my (admittedly limited) understanding of this, if it was actually needed it would be a strong ref, no? Given that it's weak, it must have some sort of default implementation. I say we go with (1).

elopio:
Time comparison with the autopgktest execution of #1908, for xemial amd64:
https://paste.ubuntu.com/26544718/

This is only one run of each, so it's not something we can say with statistical accuracy, but it seems that this branch reduces the time to less than the half, on every test.

sergiusens:
@elopio thanks for those numbers, to be extra sure of where we are, mind going ahead and grabbing the snap represented here and build the known projects we have using `classic` and verify them?

jhenstridge:
While I expect this will have a modest effect, I think the bigger wins can be built on top of this PR.  In particular:

1. Run patchelf on fewer files, as described in [bug 1747097](https://bugs.launchpad.net/snapcraft/+bug/1747097).
2. Stop running ldd.  We can borrow ctypes' `ldconfig -p` trick to list system libraries, and then process everything in a single pass.

We'll probably need to pull a bit more information from the ELF files to do all this, but it shouldn't be too hard:
1. the class and machine values from the header in order to distinguish different versions of the library on multi-arch systems.  We need both to e.g. distinguish amd64 binaries from x32: both list x86-64 as the machine type, but one is 64-bit and the other 32-bit.
2. the exported versions from `.gnu.version_d`.  This would provide a more reliable way of telling if executables/libraries inside the snap can use the libraries provided by the core/base snap, rather than just special casing libc.

jhenstridge:
That's why I included one "smoke test" that ensures the code is run at least once unmocked.  I agree that it doesn't cover all possibilities though, so the integration tests are still important.

PR #1914, opened 2018-02-03
docker: add Dockerfiles for all risk levels

kyrofa:

-----

See the README for more details.


docker/README.md
```diff
@@ -0,0 +1,29 @@
+# Creating docker containers for snapcraft
+
+## Setup
+
+    sudo snap install docker
+
+Ensure networking works, in some cases `dockerd` might need to have something
+like `--dns=8.8.8.8` added to the command.
```
elopio: Can you put the command to ensure that networking works?
Something like docker run ... ping google.com ?

docker/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Creating containers
+
+Note that in the following commands, `--network host` is not strictly needed
+if a proper docker bridge is setup.
```
elopio: I would move this right after the command that uses --network. Or maybe before, but definitely closer.

docker/README.md
```diff
@@ -0,0 +1,29 @@
...
+their respective risk level:
+
+- **edge.Dockerfile**: Using the snap from edge
+- **beta.Dockerfile**: Using the snap from beta
+- **candidate.Dockerfile**: Using the snap from candidate
+- **stable.Dockerfile**: Using the snap from stable
+
+Build the docker image with:
+
+    docker build --no-cache -f <risk>.Dockerfile --label snapcore/snapcraft --tag snapcore/snapcraft:<risk> --network host .
```
elopio: It would be cool to put this command inside a for loop with edge, beta, candidate and stable, to generate the four with a single pasted command. What do you think?

docker/README.md
kyrofa: This bit already existed, I just renamed the file for consistency. I added some bits at the bottom relating to this PR, but I'd rather not alter what is already there. Especially because I don't quite understand this blurb in the first place (why would it have DNS issues?).

docker/README.md
kyrofa: Well, judging from how I've done it so far, the only time I run this command is when I'm poking at one specifically and want to rebuild just that one, not all. Otherwise I'd write a script to build them all, but since dockerhub does this automatically I don't see it being used. Thoughts?

docker/README.md
kyrofa: Ah, good call.

docker/README.md
elopio: ok

docker/README.md
elopio: That makes sense. No loop then :)

PR #1915, opened 2018-02-03
snap: patch ctypes for the snap

sergiusens:

-----


ctypes_override/__init__.py
```diff
@@ -0,0 +1,552 @@
+"""create and manipulate C data types in Python"""
```
kyrofa: Now that you're no longer patching, it would be useful to see a blurb about why this is necessary and what was changed in this override here.

sergiusens: where would the blurb live? `snap/snapcraft.yaml` or inside this same file; I might go to patching again since I figured I was doing it wrong the other day

kyrofa: Yeah I would say inside this same file. But going back to patching is also a good option.

ctypes_override/ctypes_init.diff
```diff
@@ -0,0 +1,20 @@
...
+>     'i386': 'i386-linux-gnu',
+>     'ppc64el': 'powerpc64le-linux-gnu',
+>     'powerpc': 'powerpc-linux-gnu',
+>     'amd64': 'x86_64-linux-gnu',
+>     's390x': 's390x-linux-gnu',
+> }
+> 
+> 
+421a433,439
+>         if name == 'libsodium.so' and _os.getenv('SNAP_NAME', '') == 'snapcraft':
```
kyrofa: Why limit this to libsodium? Won't any lib that hits this code path suffer from the same issues?

PR #1924, opened 2018-02-13
schema: update version regex

kyrofa:

-----

In response to https://forum.snapcraft.io/t/3974, update Snapcraft to use the same pattern for version validation as snapd and the review tools.

### DIFF
--- a/snapcraft/tests/unit/project_loader/test_config.py
+++ b/snapcraft/tests/unit/project_loader/test_config.py
@@ -864,7 +864,7 @@ class ValidVersionTestCase(YamlTestCase):
         version in ['buttered-popcorn', '1.2.3', 'v12.4:1:2~', 'HeLlo']
     ]
 
-    def test_valid_yaml_invalid_version(self):
+    def test_valid_version(self):
         fake_logger = fixtures.FakeLogger(level=logging.ERROR)
         self.useFixture(fake_logger)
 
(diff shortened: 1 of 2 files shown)


snapcraft/tests/unit/project_loader/test_config.py
```diff
@@ -909,6 +858,67 @@ def load_effect(*args, **kwargs):
         self.mock_load_part.assert_has_calls([call1, call2], any_order=True)
 
 
+class ValidVersionTestCase(YamlTestCase):
+    scenarios = [
+        (version, dict(version=version)) for
+        version in ['buttered-popcorn', '1.2.3', 'v12.4:1:2~', 'HeLlo']
+    ]
+
+    def test_valid_yaml_invalid_version(self):
```
kalikiana: This should probably be called `test_valid_yaml_valid_version`? Or even `test_valid_version`? I find these double valid/invalid rather confusing personally so I like the shorter name more.

snapcraft/tests/unit/project_loader/test_config.py
```diff
@@ -909,6 +858,67 @@ def load_effect(*args, **kwargs):
...
+              part1:
+                plugin: nil
+        """).format(self.version))
+        _config.Config()
+
+
+class InvalidVersionTestCase(YamlTestCase):
+    scenarios = [
+        (version, dict(version=version)) for
+        version in ['*', '', '-bad', 'still-bad-', ':foo']
```
kalikiana: Mind adding instances of the remaining special cases as well? That would be `.foo`, `+foo`, `~foo`, `_foo`, `bar.`, `bar:`, `bar-`, `bar_`.

snapcraft/tests/unit/project_loader/test_config.py
kyrofa: Yes these are silly indeed, thank you. Fixed.

snapcraft/tests/unit/project_loader/test_config.py
kyrofa: Sure, done.

PR #1925, opened 2018-02-14
elf: cache crawled files

sergiusens:
LP: #1746607

-----

### DIFF
--- a/snapcraft/internal/elf.py
+++ b/snapcraft/internal/elf.py
@@ -50,6 +50,27 @@ else:
     _INTERP = b'.interp'
 
 
+class SonameCache:
+    """A cache for sonames."""
+    def __getitem__(self, key):
+        return self._soname_paths[key]
+
+    def __setitem__(self, key, item):
+        self._soname_paths[key] = item
+
+    def __contains__(self, key):
+        return key in self._soname_paths
+
+    def __init__(self):
+        """Initialize a cache for sonames"""
+        self._soname_paths = dict()  # type: Dict[str, str]
+
+    def reset(self):
+        """Reset the cache values that are empty."""
+        self._soname_paths = {k: v for (k, v) in self._soname_paths
+                              if v is not None}
+
+
 class NeededLibrary:
     """Represents an ELF library version."""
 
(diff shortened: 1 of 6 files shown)


snapcraft/internal/elf.py
```diff
@@ -513,6 +534,10 @@ def get_elf_files(root: str,
     """
     elf_files = set()  # type: Set[ElfFile]
 
+    # We need to reset the empty soname cache entries in case they are provided
+    # by a new set of root files
+    reset_soname_paths(only_none=True)
```
kalikiana: How about calling the argument `empty_entries_only`? That's the wording used in the comment and much more descriptive to my mind than referring to how it's implemented.

kalikiana: How about calling the argument `empty_entries_only`? That's the wording used in the comment and much more descriptive to my mind than referring to how it's implemented.

snapcraft/internal/elf.py
```diff
@@ -89,8 +89,25 @@ def __init__(self, *, soname: str, path: str, root_path: str,
             self.in_base_snap = False
 
 
+_soname_paths = dict()  # type: Dict[str, str]
```
kyrofa: Global state tends to bite us. Would this be better as a class instead?

snapcraft/internal/elf.py
kyrofa: Or consider flipping the logic. Most calls only want to remove the `None` entries. The only one that wants a complete reset is a test. Perhaps that call should be the one to include a "special" option.

kyrofa:
Time to prime ros-talker-listener looks much better now indeed:

real	1m51.336s
user	1m23.128s
sys	0m18.288s

snapcraft/internal/elf.py
sergiusens: sounds good.

snapcraft/internal/elf.py
sergiusens: I can work something out

sergiusens:
this must be a special day, adt passed on the first run, but travis did not :-P

sergiusens:
ooh, lxd failure, showing up in all the tests on travis

kyrofa:
> this must be a special day, adt passed on the first run, but travis did not :-P

Haha, I poked Travis like four times, too, because the error was so odd.

PR #1929, opened 2018-02-16
sources: proper errors for invalid handlers

kalikiana:

-----
Add proper error classes for unhandled sources derived from SnapcraftError.

Fixes: [LP: #1741952](https://bugs.launchpad.net/snapcraft/+bug/1741952)


snapcraft/internal/sources/errors.py
```diff
@@ -21,6 +21,22 @@ class VCSError(errors.SnapcraftError):
     fmt = '{message}'
 
 
+class UnhandledSourceError(errors.SnapcraftError):
```
sergiusens: can you use consistent names? In the form of `SnapcraftSource<something>`?

snapcraft/internal/sources/errors.py
```diff
@@ -21,6 +21,22 @@ class VCSError(errors.SnapcraftError):
...
+
+    fmt = ('no handler to manage source ({source})')
```
sergiusens: Can you use the format we agreed on which @elopio has worked on a lot?

kalikiana: It took me a while to think of proper wording here... hope you like what I put in there now.

snapcraft/tests/unit/pluginhandler/test_pluginhandler.py
```diff
@@ -98,14 +102,16 @@ def test_build_without_subdir_copies_sourcedir(self):
...
     def test_local_non_dir_source_path_must_raise_exception(self, mock_isdir):
         raised = self.assertRaises(
-            ValueError,
+            SnapcraftSourceNotADirectoryError,
             self.load_part,
             'test-part', part_properties={'source': 'file'})
 
         mock_isdir.assert_any_call('file')
 
         self.assertThat(str(raised), Equals(
```
sergiusens: these should just catch the exception type

snapcraft/tests/unit/pluginhandler/test_pluginhandler.py
```diff
@@ -2347,12 +2353,14 @@ def test_source_with_unrecognized_source_must_raise_exception(self):
...
 
         raised = self.assertRaises(
-            ValueError,
+            SnapcraftSourceUnhandledError,
             self.load_part, 'test-part', part_properties=properties)
 
-        self.assertThat(raised.__str__(),
-                        Equals('no handler to manage source '
-                               '(unrecognized://test_source)'))
+        self.assertThat(raised.__str__(), Equals(
```
sergiusens: these should just catch the exception type

snapcraft/internal/sources/errors.py
```diff
@@ -21,6 +21,26 @@ class VCSError(errors.SnapcraftError):
     fmt = '{message}'
 
 
+class SnapcraftSourceUnhandledError(errors.SnapcraftError):
+
+    fmt = ('Failed to pick an appropriate source handler:\n'
```
sergiusens: These messages aren't ideal, they don't feel right

snapcraft/internal/sources/errors.py
```diff
@@ -21,6 +21,26 @@ class VCSError(errors.SnapcraftError):
...
+           '{source} does not look like a known format.\n'
+           'Check that the URL is correct and consider adding "source-type".')
+
+    def __init__(self, source):
+        super().__init__(source=source)
+
+
+class SnapcraftSourceNotADirectoryError(errors.SnapcraftError):
+
+    fmt = ('Failed to pick an appropriate source handler:\n'
```
sergiusens: this message is more confusing than helpful

snapcraft/internal/sources/__init__.py
```diff
@@ -186,8 +187,8 @@ def _get_source_type_from_uri(source, ignore_errors=False):  # noqa: C901
     elif source.endswith('7z'):
         source_type = '7z'
     elif common.isurl(source) and not ignore_errors:
-        raise ValueError('no handler to manage source ({})'.format(source))
+        raise errors.SnapcraftSourceUnhandledError(source)
     elif not os.path.isdir(source) and not ignore_errors:
-        raise ValueError('local source ({}) is not a directory'.format(source))
+        raise errors.SnapcraftSourceNotADirectoryError(source)
```
elopio: this is a nit, but I would prefer SnapcraftSourceNotDictionaryError. Without the `a`.

snapcraft/tests/unit/sources/test_errors.py
```diff
@@ -0,0 +1,47 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+from testtools.matchers import Equals
+
+from snapcraft.internal.sources import errors
+from snapcraft.tests import unit
+
+
+class ErrorFormattingTestCase(unit.TestCase):
```
elopio: My idea was to keep all of these on the same file, `snapcraft/tests/unit/test_errors.py`, because that would make it very easy to compare the style, find inconsistencies and make improvements over all the messages.

The downside is that the file will be big. What do you think?

snapcraft/internal/sources/errors.py
elopio: Let's see. This is still hard for me, but what about:

Failed to get source: unknown source-type for {source}.
Run `snapcraft help sources` to read the documentation.

?

snapcraft/internal/sources/errors.py
elopio: Maybe:
```
Failed to get source: {source!} is not a directory.
Run `snapcraft help sources` to read the documentation
```

kalikiana: Making both say the same generic suggestion seems less helpful to me... though I can totally be convinced if you guys like it. Would go for a single error class `SnapcraftSourceUnhandledError` for both cases then.

snapcraft/tests/unit/sources/test_errors.py
kalikiana: No strong opinion personally - I did this as suggested in this PR earlier.

snapcraft/internal/sources/errors.py
kalikiana: Correction. The second half of the first part is different as I've just come to realize... I feel like I really need a line break in there.

elopio: The style suggested by mpt was:
```
generic error: specific reason \n
how to fix
```
We still need to make the other errors consistent with this. And because in here so many things could have gone wrong, like a typo, missing source-type, a type that we don't support yet, etc. I think the appropriate action to fix this would be for the user to read the docs and figure it out by themselves. But as I said, just a suggestion. 

And the error messages is something we'll have to iterate over and over again, making small improvements. During the sprint we talked to mpt about how to do it, so I think it's good to just give it our best and then get corrections from him now and then.

snapcraft/tests/unit/sources/test_errors.py
sergiusens: I would really rather separate them and have them under a common module name `test_errors`

snapcraft/internal/sources/errors.py
```diff
@@ -21,6 +21,26 @@ class VCSError(errors.SnapcraftError):
     fmt = '{message}'
 
 
+class SnapcraftSourceUnhandledError(errors.SnapcraftError):
+
+    fmt = ('Failed to pick an appropriate source handler:\n'
+           '{source} does not look like a known format.\n'
+           'Check that the URL is correct and consider adding "source-type".')
```
sergiusens: "or" consider?

snapcraft/internal/sources/__init__.py
kyrofa: Given that python actually has a `NotADirectoryError` I personally think this one makes sense.

kyrofa:
Might I suggest we also handle an invalid `source-type` by catching a `KeyError` [here](https://github.com/snapcore/snapcraft/pull/1929/files#diff-8c6bb564ac3a3ca26deb09b5c0b8b2e2R164)?

snapcraft/internal/sources/errors.py
kyrofa: Oh I want to join in!

I think Leo is on the right track here, but the issue isn't that the `source-type` is unknown (that would bite us [here](https://github.com/snapcore/snapcraft/pull/1929/files#diff-8c6bb564ac3a3ca26deb09b5c0b8b2e2R164) instead), it's that we're unable to _determine_ the source type given the source.

Remembering that our guidance is:

```
# <what>: What went wrong? Not "what technical detail broke in snapcraft" like is here now, but "what, that the user cares about, went wrong?"
# <why>: Why did that terrible thing happen?
# <fix>: How might the user carry on?
<what>: <why>
<fix>
```

(note that there is no line break between `<what>` and `<why>`. Something new we learned at the summit. @elopio double check me?)

If we can manage to get the part name into this, I suggest a combination of both Leo's suggestion and what's already there:

```
Failed to fetch source for part {part_name!r}: unable to determine source type of {source!r}.\n
Consider specifying `source-type`. See `snapcraft help sources` for more information.
```

snapcraft/internal/sources/errors.py
kyrofa: I like Leo's suggestion here, although I'd personally change `get` to `pull`.

snapcraft/tests/unit/sources/test_errors.py
kyrofa: I think keeping tests in roughly a layout that mirrors what they're testing makes sense. Testing `internal/x.py`? Then there should probably be a `snapcraft/tests/unit/test_x.py` Testing `internal/sources/errors.py`? Then a `snapcraft/tests/sources/test_errors.py` makes sense.

The one downside to this is that having one large file of error messages gives mpt something relatively easy to review.

snapcraft/internal/sources/errors.py
kalikiana: Thank you for these suggestions. I went with the last one, retaining the consideration for a typo in the URL.
Getting the part_name here is slightly painful since the source handlers are used in several different places... but I'm trying to figure it out, since actually it's really useful to have there.

snapcraft/internal/sources/errors.py
kalikiana: I went with a modified one based on the input on the other error, including "pulling the source for part {part_name!r}".

kalikiana:
> Might I suggest we also handle an invalid source-type by catching a KeyError here?

@kyrofa Do you mean to have an error when the source-type doesn't exist? That case will already by caught by the schema. So you won't see an exception. Or did you mean something else?

snapcraft/internal/sources/errors.py
```diff
@@ -21,6 +21,28 @@ class VCSError(errors.SnapcraftError):
...
+           'See `snapcraft help sources` for more information.')
+
+    def __init__(self, source, part_name):
+        super().__init__(source=source, part_name=part_name)
+
+
+class SnapcraftSourceNotADirectoryError(errors.SnapcraftError):
+
+    fmt = ('Failed to pull source for part {part_name!r}: '
+           '{source!r} is not a directory.\n'
```
sergiusens: The exception name has been misleading the message, which I guess comes from the check (`isdir(source)`). The implications of that is that we found no handler to handle a file source (`source: mytar.tgz`) which is not a directory nor is this source a directory, maybe something like this would be better:
```
'Failed to pull source for part {part_name!r}: '
'the `source-type` for {source!r} could not be detected.\n'
```

Or change the logic a bit,
```python
     elif _tar_type_regex.match(source):
         source_type = 'tar'
     elif  os.path.is_dir(source):
        source_type = 'dir'
     else:
        raise errors.SnapcraftSourceUnhandledError(source, part_name)
```

and in the `_source_handler` add `'dir': Local`.

This rids us of that last exception.

sergiusens:
thanks for taking care of this @elopio

PR #1930, opened 2018-02-16
lxd: friendly errror with suggestions if network is broken

kalikiana:

-----
As evidenced by [LP: #1749679](https://bugs.launchpad.net/snapcraft/+bug/1749679) the verbatim network probing failures aren't very helpful. We can try to reduce noise and give some hints as to what's wrong. `ContainerConnectionError` will ensure there's a link to LXD docs and the URL we check is mentioned in the message.
`urllib.error.URLError` is handled explicitly to get a niceish error message in the case of connection errors. All other exceptions will generically be caught as well since we don't want a stack trace in any event and **urllib** can leak other exceptions such as `UnicodeError`.

Fixes: [LP: #1719952](https://bugs.launchpad.net/snapcraft/+bug/1719952)

The following tests were updated:

 - tests.unit.commands.test_snap
 - tests.unit.test_lxd

I locally ran the tests:

Manual test steps:
 - cd tests/integration/snaps/basic
 - snapcraft cleanbuild
 - lxc profile set default environment.http_proxy localhost:1234
 - Observe `Failed to open 'http://start.ubuntu.com/connectivity-check.html': [Errno 111] Connection refused`
 - lxc profile set default environment.http_proxy invalid
 - Observe `Failed to open 'http://start.ubuntu.com/connectivity-check.html': [Errno -2] Name or service not known`
  - lxc profile set default environment.http_proxy http://unknown
  - Observe `Failed to open 'http://start.ubuntu.com/connectivity-check.html': [Errno -2] Name or service not known`
 - lxc profile set default environment.http_proxy .invalid
 - Observe `Failed to open 'http://start.ubuntu.com/connectivity-check.html': encoding with 'idna' codec failed (UnicodeError: label empty or too long)`

Note: Use `lxc profile set default environment.http_proxy ''` to unset the proxy in the default LXD profile after testing, or set it to the value you were using previously.


snapcraft/internal/lxd/_containerbuild.py
```diff
@@ -45,6 +45,8 @@
 logger = logging.getLogger(__name__)
 
 _NETWORK_PROBE_COMMAND = \
+    'import sys; sys.excepthook = lambda t,e,b : ' \
+    'print("{{}}: {{}}".format(t.__name__, e));' \
```
kyrofa: excepthook feels like overkill here. Why not just a try/except?

kalikiana: The lack of line breaks is deliberate here because this is being run with `python -c`. As such we can't use try/except because that requires a line break.
If we wanted this to be more beautiful we'd have to have a script that we push into the container after it's created and run instead of the string. To my mind that's not worth it. Unless others feel strongly about pretty code.

kyrofa: You've already had to modify this code. I think it's important to make it as maintainable as possible, for the same reason we want other code to be maintainable. In addition, you simply can't accomplish everything in a single line. Try running this:

```
import subprocess
import textwrap

_CODE = "import textwrap; for i in range(2): print('hello')"

subprocess.check_call(['python3', '-c', _CODE])
```

It doesn't work. However, this does, is just as easy, and looks way nicer:

```
import subprocess
import textwrap

_CODE = textwrap.dedent("""\
        import textwrap

        for i in range(2):
                print('hello')
""")

subprocess.check_call(['python3', '-c', _CODE])
```

Note that this works because the line breaks are preserved, but it's still just a string that you're handing to `-c` like normal.

kalikiana:
@kyrofa Please give it another look. I changed the snippet as you requested.

PR #1934, opened 2018-02-16
catkin plugin: support recursive rosinstall files

### LINKED ISSUE #1714: catkin plugin: recursively merge rosinstall files
Estimate: 4 days

The Catkin plugin currently supports parsing rosinstall files to pull the various repositories that make up the workspace that it will build. A request from Shadow Robotics is to add the ability to then look inside the pulled workspace for more rosinstall files, and fetch their repositories as well. This is not supported by rosinstall alone. This will require us to have a concept of being finished. How do we know that we’ve finished pulling everything necessary if we’re doing so in a recursive manner?

kyrofa:

-----

Currently, the Catkin plugin supports rosinstall files in that, if you hand it a list of rosinstall files in the source, it'll merge them into the workspace and fetch the repositories listed therein. However, what happens if _those_ repositories have rosinstall files of their own? This is not a situation wstool handles, and is something vendors script around.

This PR resolves #1714 by adding explicit support in Snapcraft for this situation by way of a new boolean option: `recursive-rosinstall`. This option defaults to false (since that's typically how wstool behaves), but if it's true, it'll continue merging/updating the workspace with newly-fetched rosinstall files until no more are found.


kyrofa:
This PR can be tested using the `stable/pr-1934-recursive-rosinstall` channel:

    $ sudo snap install snapcraft --classic --channel=stable/pr-1934-recursive-rosinstall

Or, if you already have it installed:

    $ sudo snap refresh snapcraft --channel=stable/pr-1934-recursive-rosinstall

snapcraft/plugins/catkin.py
```diff
@@ -38,6 +38,11 @@
       (list of strings)
       List of rosinstall files to merge while pulling. Paths are relative to
       the source.
+    - recursive-rosinstall:
```
sergiusens: we need a consolidated strategy for these, wouldn't an enum be better long term to cater for all the flags or are boolean properties the way to go?

Asking for a friend ;-)

snapcraft/plugins/catkin.py
```diff
@@ -372,8 +383,21 @@ def pull(self):
             if self.options.source_subdir:
                 source_path = os.path.join(self.sourcedir,
                                            self.options.source_subdir)
-            _handle_rosinstall_files(
-                wstool,  source_path, self.options.rosinstall_files)
+
+            # Create a set of paths to rosinstall files
```
sergiusens: this seems to only be used in your `else` clause bellow, why not just define it there? I also do not understand why you just don't pass `self.options.rosinstall_files` to `_handle_rosinstall_files`

sergiusens: A `source_path`

sergiusens: ``python
else:
    rosinstall_files_full_paths = {os.path.join(source_path, ri) for ri in self.options.rosinstall_files} 
    _handler_rosinstall_files(wstool, rosinstall_files_full_paths)
```

maybe?

kyrofa:
>> _This is not a situation wstool handles, and is something vendors script around._
>
> Care to clarify?

wstool doesn't support recursive rosinstall files, so vendors using this tend to write custom shell scripts to pull their workspaces down.

snapcraft/plugins/catkin.py
kyrofa: Good catch, it should be under the `else` indeed. As to why we don't just pass the option itself, it's because in this PR the `_handle_rosinstall_files` function was updated to no longer accept the parent directory path, and just take absolute paths to rosinstall files instead. This was done to make the recursive case easier and more efficient. Otherwise we'd need to process each rosinstall path recursively discovered, figure out the common parent directory, then pass that and relative paths onto `_handle_rosinstall_files`.

PR #1941, opened 2018-02-19
project_loader: handle invalid unicode chars

kalikiana:

-----
Fixes: [LP: #1737571](https://bugs.launchpad.net/snapcraft/+bug/1737571)

Note that the exception is currently raised with valid and invalid unicode chbaracters due to the upstream bug [pyyaml#25](https://github.com/yaml/pyyaml/issues/25). But we'll want to handle the error cleanly even if the upstream issue is fixed.

Steps to reproduce:

    $ snap install snapcraft --classic --edge
    $ snapcraft init
    $ # Modify "summary" or "version" to include the 💩 emoji
    $ snapcraft pull
    $ # Build fails to validate snapcraft.yaml

This branch adds a patch to the PyYAML used by the Snapcraft snap to handle the unicode code points erroneously flagged as invalid, such as the 💩 emoji. A Snapcraft snap built from this PR will successfully validate a summary or description making use of the 💩 emoji.

New test cases:
 - tests.unit.project_loader.test_config.test_invalid_yaml_invalid_unicode_chars
 - tests.integration.general.test_global_properties

Note: The integration level test for the PyYAML work-around is skipped unless testing with a snap or Debian package (ie. `SNAPCRAFT_FROM_SNAP=1` is set or `SNAPCRAFT_FROM_DEB=1` is set), both of which are patched. I verified with a snap from the branch, and running the tests in a virtual environment respectively.


chipaca:
Why is that invalid in a description? What is a valid description?

(I don't think descriptions should be ascii-only)

kalikiana:
@chipaca See my note in the description. There's a bug in pyyaml's handling of the unicode spec.

chipaca:
yes, that's what's brought me here: you're upgrading a "the yaml parser we chose to use does not handle this" to "this is not allowed". it's fine to say "not supported (use a \u escape sequence instead)", or even do the conversion ourselves ahead of time, but to say "not allowed" is a lie (you can work around it by using a \u escape), and not what we want either (these fields are for human-readable text, ie unicode.

There _is_ validation to be done on these fields: I don't think they should include any character from the private use area nor any control character. But that's separate.

chipaca:
FWIW,
```
>>> sys.stdout.buffer.write("💩\n".encode("ascii", errors='backslashreplace'))
\U0001f4a9
```

kalikiana:
Seems like the Debian package was patched, so the version in pip behaves differently, causing the failure in CI. I changed the original test case now to use a code point that's actually invalid vs. parsed as invalid upstream, and also added a work-around so that emoji like in the bug report are considered valid regardless of version.

snap/snapcraft.yaml
```diff
@@ -47,6 +48,7 @@ parts:
         TRIPLET_PATH="$SNAPCRAFT_PART_INSTALL/usr/lib/$(gcc -print-multiarch)"
         LIBSODIUM=$(readlink -n $TRIPLET_PATH/libsodium.so.18)
         ln -s $LIBSODIUM $TRIPLET_PATH/libsodium.so
+        patch -d $SNAPCRAFT_PART_INSTALL/lib/python3.6/site-packages -p1 < $SNAPCRAFT_STAGE/pyyaml-support-high-codepoints.diff
     after: [python, apt]
```
kyrofa: You probably want to add `patches` to after here.

sergiusens: the after for python includes this, so the chain is somewhat complete; I will sort of ignore this as we are going back to `stage-packages` as soon as our SRUed deb gets in with the fixes for classic

sergiusens: Good catch otherwise ;-)

sergiusens:
tested the snap, also works.

PR #1947, opened 2018-02-21
errors: add ability to send stack traces to sentry

sergiusens:
Adds the ability to send reports, asking for confirmation before doing
so. This feature is only enabled for the deb as raven is not provided in
`xenial` and most of the user base will be using the snap.

Closes: #1918

-----

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,6 +17,7 @@ pymacaroons==0.9.2; sys_platform != 'win32'
 pymacaroons==0.10.0; sys_platform == 'win32'
 pymacaroons-pynacl==0.9.3
 pysha3==1.0b1; python_version < '3.6'
+raven==6.5.0
 simplejson==3.8.2
 tabulate==0.7.5
 python-debian==0.1.28
--- a/schema/snapcraft.yaml
+++ b/schema/snapcraft.yaml
@@ -73,13 +73,31 @@ properties:
     type: string
     description: name of the part that provides source files that will be parsed to extract snap metadata information
   name:
-    type: string
     description: name of the snap package
-    validation-failure:
-      "{.instance!r} is not a valid snap name. Snap names consist of lower-case
-      alphanumeric characters and hyphens. They cannot be all numbers. They
-      also cannot start or end with a hyphen."
-    pattern: "^(?:[a-z0-9]+-?)*[a-z](?:-?[a-z0-9])*$"
+    allOf:
+      - type: string
+        # this failure message avoids printing repr of the thing, as it could be huge
+        validation-failure: "snap names need to be strings."
+        # string, but too long, is caught by this
+        maxLength: 40
+      - pattern: "^[a-z0-9-]*[a-z][a-z0-9-]*$"
+        validation-failure:
+          "{.instance!r} is not a valid snap name. Snap names can only use ASCII
+          lowercase letters, numbers, and hyphens, and must have at least one
+          letter."
+      - pattern: "^[^-]"
+        validation-failure:
+          "{.instance!r} is not a valid snap name. Snap names cannot start with
+          a hyphen."
+      - pattern: "[^-]$"
+        validation-failure:
+          "{.instance!r} is not a valid snap name. Snap names cannot end with a
+          hyphen."
+      - not:
+          pattern: "--"
+        validation-failure:
+          "{.instance!r} is not a valid snap name. Snap names cannot have two
+          hyphens in a row."
   architectures:
     type: array
     description: architectures to override with
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -35,7 +35,7 @@ parts:
         - libsodium-dev
         - liblzma-dev
     stage-packages:
-        - binutils
+        - execstack
         - libffi6
         - libsodium18
         - squashfs-tools
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -13,6 +13,7 @@ confinement: classic
 apps:
   snapcraft:
     command: bin/snapcraft
+    completer: snapcraft-completion
 
 parts:
   ctypes:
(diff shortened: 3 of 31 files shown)


snapcraft/cli/_errors.py
```diff
@@ -20,6 +20,19 @@
...
+# raven is not available on 16.04
+try:
+    from raven import Client as RavenClient
+    from raven.transport import RequestsHTTPTransport
+except ImportError:
+    RavenClient = None
+
+_SEND_TO_SENTRY = (
+    'Sorry, Snapcraft had an internal error.\n'
+    'To help us improve, would you like to send error data?'
```
kalikiana: This message raises a couple of questions in my head:

- What data is this going to send?
- Will this create a bug report a la apport or ubuntu-bug or something else?

That is to say, perhaps the message can be more helpful in telling the user what is going to happen with regard to privacy concerns and what Snapcraft is going to do.

snapcraft/cli/_errors.py
```diff
@@ -46,6 +59,9 @@ def exception_handler(exception_type, exception, exception_traceback, *,
         traceback.print_exception(
             exception_type, exception, exception_traceback)
 
+    if not is_snapcraft_error and RavenClient is not None:
+        _submit_trace(exception)
```
kalikiana: Can we have a test case for this? It seems like there's three cases here at least that we'll care about.

1. Trace prompt will be triggered
2. Non-interactive build skips past the prompt (rather than timing out)
3. No trace available

snapcraft/cli/_errors.py
```diff
@@ -55,3 +71,15 @@ def exception_handler(exception_type, exception, exception_traceback, *,
             echo.error(str(exception))
 
     sys.exit(exit_code)
+
+
+def _submit_trace(exception):
+    client = RavenClient(
+        'https://b0fef3e0ced2443c92143ae0d038b0a4:'
+        'b7c67d7fa4ee46caae12b29a80594c54@sentry.io/277754',
```
elopio: Curious, how can we prevent to be spammed with evil messages to our sentry?

sergiusens: I don't think we can.

requirements.txt
```diff
@@ -17,6 +17,7 @@ pymacaroons==0.9.2; sys_platform != 'win32'
 pymacaroons==0.10.0; sys_platform == 'win32'
 pymacaroons-pynacl==0.9.3
 pysha3==1.0b1; python_version < '3.6'
+raven==6.5.0
```
elopio: I think it would be good to explain somewhere, maybe in the description of this pr, why raven is not a requirement for the deb.

schema/snapcraft.yaml
```diff
@@ -73,13 +73,31 @@ properties:
...
+          a hyphen."
+      - pattern: "[^-]$"
+        validation-failure:
+          "{.instance!r} is not a valid snap name. Snap names cannot end with a
+          hyphen."
+      - not:
+          pattern: "--"
+        validation-failure:
+          "{.instance!r} is not a valid snap name. Snap names cannot have two
+          hyphens in a row."
```
elopio: This diff shouldn't be here. What happened?

snap/snapcraft.yaml
```diff
@@ -13,6 +13,7 @@ confinement: classic
 apps:
   snapcraft:
     command: bin/snapcraft
+    completer: snapcraft-completion
```
elopio: This is also not related to this pr.

snapcraft/cli/_errors.py
```diff
@@ -42,10 +54,15 @@ def exception_handler(exception_type, exception, exception_traceback, *,
...
 
-    if debug or not is_snapcraft_error:
+    if debug:
         traceback.print_exception(
             exception_type, exception, exception_traceback)
 
+    if not is_snapcraft_error:
+        echo.error('Sorry, Snapcraft had an internal error.')
+        if RavenClient is not None and _is_environment_ok():
+            _submit_trace(exception)
```
elopio: When raven is not available, the message about sorry feels very empty. Shouldn't we give a link to report the issue, at least? Or point people to the forum to talk about it.

snapcraft/cli/_errors.py
```diff
@@ -55,3 +72,20 @@ def exception_handler(exception_type, exception, exception_traceback, *,
             echo.error(str(exception))
 
     sys.exit(exit_code)
+
+
+def _is_environment_ok():
```
elopio: I don't like the name of this function.
What about: `_send_error_data()`?

snapcraft/cli/_errors.py
```diff
@@ -14,12 +14,24 @@
...
+import click
+# raven is not available on 16.04
+try:
+    from raven import Client as RavenClient
+    from raven.transport import RequestsHTTPTransport
+except ImportError:
+    RavenClient = None
+
+_MSG_INTERNAL_ERROR = 'Sorry, Snapcraft had an internal error.'
+_MSG_SEND_TO_SENTRY = 'To help us improve, would you like to send error data?'
```
elopio: I would ask Mathew to improve the wording her.
It should mention that it's anonymous, that only the trace relevant to snapcraft and the snap being build will be collected, and that the info will never be shared outside of the snapcraft team. "error data" sounds scary to me.

snapcraft/cli/_errors.py
sergiusens: But it doesn't encompass everything

snap/snapcraft.yaml
sergiusens: It is because I rebased, but this PR is not against master, and even though the target branch is now updated, github missed out on that fact.

snapcraft/cli/_errors.py
sergiusens: This will only happen with the deb. This is according to the spec we already designed.

requirements.txt
sergiusens: In the PR now

PR #1950, opened 2018-02-22
elf: better debug messages

sergiusens:
Some debug messages needed polishing, like the linker check.
Others, like the check for exclusions of libraries, needed
polish to not repeat itself over and over again when
running off of a non supported base. This comes with a
side effect of reducing the amount of calls necessary
for the check where the debug message is displayed.

-----

### DIFF
--- a/snapcraft/internal/elf.py
+++ b/snapcraft/internal/elf.py
@@ -217,7 +217,8 @@ class ElfFile:
         linker_version = m.group('linker_version')
         r = parse_version(version_required) <= parse_version(linker_version)
         logger.debug('Checking if linker {!r} will work with '
-                     'GLIBC_{}: {!r}'.format(linker, version_required, r))
+                     'GLIBC_{} required by {!r}: {!r}'.format(
+                         linker, version_required, self.path, r))
         return r
 
     def get_required_glibc(self) -> str:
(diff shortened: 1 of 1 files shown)


snapcraft/internal/elf.py
```diff
@@ -531,10 +532,10 @@ def _get_system_libs() -> FrozenSet[str]:
...
         libc6_libs = [os.path.basename(l)
                       for l in repo.Repo.get_package_libraries('libc6')]
-        return frozenset(libc6_libs)
-
-    with open(lib_path) as fn:
-        _libraries = frozenset(fn.read().split())
+        _libraries = frozenset(libc6_libs)
+    else:
+        with open(lib_path) as fn:
+            _libraries = frozenset(fn.read().split())
```
elopio: I fail to see how this reduces the number of calls. The change seems equivalent to me, so I might be missing something.
I mean, +1 because it looks correct, but can you explain it to me in case I'm doing a wrong review, missing something obvious because I'm being dense?

sergiusens: it is at the beginning of the function:
```python
global _libraries
     if _libraries:  # type: ignore
return _libraries # type: ignore
```

elopio: Ahh, I get it now. Thanks.

PR #1959, opened 2018-02-26
elf: only patch elf files that aren't referenced by DT_NEEDED

jhenstridge:

-----
This probably needs a bit more testing, but I thought I'd push the PR now to get some CI feedback.

This is based on an idea I mentioned at the Snapcraft Summit to minimise the number of files we run patchelf on, as described in [bug 1747097](https://bugs.launchpad.net/snapcraft/+bug/1747097).

In brief, when the dynamic linker is searching for DT_NEEDED dependencies of a library, it can use the rpath of the executable or library that pulled in the library.  So for classic snaps, it should only be necessary to patch the rpath of executables or shared library plugins.  As a simple heuristic, any elf file whose soname is listed in the DT_NEEDED of some other elf file doesn't need patching.

I also updated ElfFile to extract architecture information from elf files.  This way we don't e.g. treat an ARM library as satisfying a dependency of an AMD64 executable.

LP: #1747097


snapcraft/internal/elf.py
```diff
@@ -175,6 +176,14 @@ def _extract(self, path):  # noqa: C901
         with open(path, 'rb') as fp:
             elf = elftools.elf.elffile.ELFFile(fp)
 
+            # A set of fields to identify the architecture of the ELF file:
+            #  e_machine: instruction set (e.g. x86-64 vs. arm64)
+            #  EI_CLASS: 32/64 bit (e.g. amd64 vs. x32)
+            #  EI_DATA: byte orer (e.g. ppc64 vs. ppc64le)
```
sergiusens: order

snapcraft/internal/elf.py
```diff
@@ -175,6 +176,14 @@ def _extract(self, path):  # noqa: C901
         with open(path, 'rb') as fp:
             elf = elftools.elf.elffile.ELFFile(fp)
 
+            # A set of fields to identify the architecture of the ELF file:
```
sergiusens: Thanks for adding a comment, can you order these as in the tuple and also add the example value of the constant for x86-64 in e_machine and so on and so forth.

snapcraft/internal/elf.py
```diff
@@ -175,6 +176,17 @@ def _extract(self, path):  # noqa: C901
...
+            #  EI_CLASS: 32/64 bit (e.g. amd64 vs. x32)
+            #  EI_DATA: byte orer (e.g. ppc64 vs. ppc64le)
+            #  e_machine: instruction set (e.g. x86-64 vs. arm64)
+            #
+            # For amd64 binaries, this will evaluate to:
+            #   ('ELFCLASS64', 'ELFDATA2LSB', 'EM_X86_64')
```
kyrofa: Do we need all this information? Or can we just use `get_machine_arch()`?

snapcraft/internal/elf.py
```diff
@@ -589,3 +601,17 @@ def get_elf_files(root: str,
                 elf_files.add(ElfFile(path=path))
 
     return frozenset(elf_files)
+
+
+def get_elf_files_to_patch(elf_files):
```
kyrofa: This function plays an important role here, and would benefit from a good docstring.

kyrofa:
I went ahead and resolved the conflicts and brought this bad boy up to date.

snapcraft/internal/elf.py
jhenstridge: The point of this code is to handle the case where a snap contains ELF files for multiple architectures.  Some examples where this happen include:

1. the core snap for amd64/arm64, which contains libraries to support the matching 32-bit architecture:

    /snap/core/current/lib/i386-linux-gnu/libc-2.23.so
    /snap/core/current/lib/x86_64-linux-gnu/libc-2.23.so

2. architecture all snaps that contain executables for multiple architectures.

The aim here is to e.g. make sure a 32-bit executable depending on libfoo.so is matched to the 32-bit version of that library, and the equivalent for 64-bit.  The `platform.get_machine_arch` function doesn't help with this.

jhenstridge: bah.  I was looking at the wrong function.  So, `ELFFile.get_machine_arch()` is not sufficient because it only references `e_machine`:

https://github.com/eliben/pyelftools/blob/master/elftools/elf/elffile.py#L212

So it can't distinguish amd64 vs. x32 binaries (both share the same e_machine value, differ in word size), or ppc64 vs. ppc64le (both share same e_machine and word size, but have different byte order).  Also, get_machine_arch() folds everything but 5 values to "<unknown>".

sergiusens:
thank you very much @jhenstridge

PR #1961, opened 2018-02-27
Sentry

sergiusens:
Adds the ability to send reports, asking for confirmation before doing
so. This feature is only enabled for the deb as raven is not provided in
xenial and most of the user base will be using the snap.

Closes: #1918

- [x] Hae you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?

-----

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,6 +17,7 @@ pymacaroons==0.9.2; sys_platform != 'win32'
 pymacaroons==0.10.0; sys_platform == 'win32'
 pymacaroons-pynacl==0.9.3
 pysha3==1.0b1; python_version < '3.6'
+raven==6.5.0
 simplejson==3.8.2
 tabulate==0.7.5
 python-debian==0.1.28
--- a/snapcraft/cli/_errors.py
+++ b/snapcraft/cli/_errors.py
@@ -28,10 +28,8 @@ try:
 except ImportError:
     RavenClient = None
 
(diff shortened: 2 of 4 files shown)


kyrofa:
> This feature is only enabled for the deb as raven is not provided in
xenial and most of the user base will be using the snap.

I assume you mean that it's only enabled for the snap?

kyrofa:
Took this for a test run, and have a few thoughts:

- Running `snapcraft` and getting a non-snapcraft error works as expected. Saying yes or no also works as expected
- Running `snapcraft` and hitting ctrl+c when prompted for a yes or no results in a pretty brutal traceback. Would be nice to interpret that as "no".
- Running snapcraft in debug mode shows the traceback as expected, although it also shows this which isn't tremendously helpful: `Configuring Raven for host: <raven.conf.remote.RemoteConfig object at 0x7f47edea1c88>`. Could that be more helpful, or should it be removed?

sergiusens:
@kyrofa for each,
- feel free to suggest that in the spec
- this we can work on
- this is a major problem we have all across snapcraft and its imported modules, we need to scope the debugging for the logger to the `snapcraft` name and have every `getLogger` get the scoped one.

Thanks for the review!

sergiusens:
and yes, only enabled for the snap :-)

kyrofa:
What spec? There's nothing here or in the issue that I see anyway.

PR #1968, opened 2018-02-28
demos: avoid use of the wrapper for java-hello-world

sergiusens:
When building on a host different that the target base, there is no
wrapper created. Make use of `adapter: none` to make this demo a bit
more generic and manually setup the environment.

-----
Part of #1943

### DIFF
--- a/demos/java-hello-world/snap/snapcraft.yaml
+++ b/demos/java-hello-world/snap/snapcraft.yaml
@@ -7,7 +7,12 @@ grade: stable
 
 apps:
  hello:
-   command: usr/lib/jvm/default-java/bin/java oata.HelloWorld
+    command: usr/lib/jvm/default-java/bin/java oata.HelloWorld
+    adapter: none
+    environment:
+       JAVA_HOME: $SNAP/usr/lib/jvm/default-java
+       CLASSPATH: $SNAP/jar/HelloWorld.jar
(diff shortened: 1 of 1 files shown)


demos/java-hello-world/snap/snapcraft.yaml
```diff
@@ -7,7 +7,12 @@ grade: stable
 
 apps:
  hello:
-   command: usr/lib/jvm/default-java/bin/java oata.HelloWorld
+    command: usr/lib/jvm/default-java/bin/java oata.HelloWorld
```
kyrofa: Most of our YAMLs use two spaces for indentation, this is now using three.

kyrofa:
> When building on a host different that the target base, there is no wrapper created.

I don't quite understand why this means we need to change our demo. Mind explaining?

demos/java-hello-world/snap/snapcraft.yaml
sergiusens: interesting, on my editor I somehow counted 4!

sergiusens:
@kyrofa if no wrapper is provided, the environment variables usually used won't be set, so we set them ourselves. The reason they are not set is due to the patching of `libc6`, this is like a hybrid `classic` confined snap working on top of an already existing core which basically might not have ABI matches anywhere.

PR #1969, opened 2018-03-01
Add a "--profile" parameter to cleanbuild

chrisglass:
This allows the user to select another LXD profile when running
cleanbuild, in case some profile-specific options are required. This
allows users not to have to change their default profile.

This partially fixes LP:1686766

The change follows the existing pattern (passing an explicit option instead of an "extra options" free-form parameter), as with the existing "--remote" option.

-----


chrisglass:
I think that failed check is not related to my code.

sergiusens:
Moving forward, we have decided to move away from containers as the mechanism to execute builds https://forum.snapcraft.io/t/draft-consistent-build-environments-build-vms/5374/3

Sorry for the late reply and the wasted effort.

PR #1975, opened 2018-03-01
python plugin: find setup.py when source-subdir is used

sergiusens:
LP: #1752481

-----

### DIFF
--- a/snapcraft/plugins/python.py
+++ b/snapcraft/plugins/python.py
@@ -220,8 +220,13 @@ class PythonPlugin(snapcraft.BasePlugin):
     def _get_setup_py_dir(self):
         setup_py_dir = None
         setup_py = 'setup.py'
-        if os.listdir(self.sourcedir):
-            setup_py = os.path.join(self.sourcedir, 'setup.py')
+        if self.options.source_subdir:
+            sourcedir = os.path.join(self.sourcedir,
+                                     self.options.source_subdir)
+        else:
+            sourcedir = self.sourcedir
+        if os.listdir(sourcedir):
+            setup_py = os.path.join(sourcedir, 'setup.py')
         if os.path.exists(setup_py):
             setup_py_dir = os.path.dirname(setup_py)
 
--- a/snapcraft/plugins/python.py
+++ b/snapcraft/plugins/python.py
@@ -220,9 +220,13 @@ class PythonPlugin(snapcraft.BasePlugin):
     def _find_file(self, *, filename: str) -> str:
         # source-subdir defaults to ''
         for basepath in [self.builddir, self.sourcedir]:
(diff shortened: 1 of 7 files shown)


kyrofa:
I think this is obviously correct as far as taking the source_subdir into account to find the setup.py, so I'll leave a +1 on that aspect.

However, I think there's a larger issue here. Building should not be reaching back into the sourcedir, it should be using the builddir. The build step calls `_install_project` which then calls `_get_setup_py_dir`, `_get_constraints`, and `_get_requirements` all of which use the sourcedir. These functions are also used in pulling, so they need to understand the context in which they're called to return the appropriate path.

kyrofa:
> Ideally, this would happen only on the base plugin.

Agreed, like we do for builddir.

snapcraft/plugins/python.py
```diff
@@ -238,8 +242,11 @@ def _get_constraints(self):
             if isurl(self.options.constraints):
                 constraints = {self.options.constraints}
             else:
-                constraints = {os.path.join(
-                    self.sourcedir, self.options.constraints)}
+                constraints = {self._find_file(
+                    filename=self.options.constraints)}
+                if not constraints:
+                    # TODO snapcraft owned message
+                    raise FileNotFoundError(self.options.constraints)
```
kyrofa: This is new behavior, is it not? Why not define the error class here? Also would be nice to see tests covering these cases.

sergiusens: we don't have any place to add them for plugins, this is urgent and needs to come out today (that is why we have a TODO)

snapcraft/plugins/python.py
```diff
@@ -217,18 +217,22 @@ def build(self):
             self._python_major_version, stage_dir=self.project.stage_dir,
             install_dir=self.installdir)
 
+    def _find_file(self, *, filename: str) -> str:
+        # source-subdir defaults to ''
+        for basepath in [self.builddir, self.sourcedir]:
```
kyrofa: Builddir includes the subdir, FYI.

snapcraft/plugins/python.py
kyrofa: That [hasn't stopped us in the past](https://github.com/snapcore/snapcraft/blob/master/snapcraft/plugins/catkin.py#L103). Your call, of course. No tests either, then?

snapcraft/plugins/python.py
sergiusens: yeah, that's in the last push :-/

kyrofa: Would using `self.build_basedir` not be cleaner than special-casing the sourcedir?

sergiusens:
@kyrofa fwiw, this will not land in master but will be made available to outside consumers

PR #1980, opened 2018-03-02
elf: remove all un-primed libs from soname cache

kyrofa:

-----

Snapcraft currently resets the cache for each part, but the cache's concept of "resetting" is only to remove libs that weren't found. Caching library lookups between parts causes issues in snaps where multiple parts rely on the same lib, but they come from different places (e.g. git-ubuntu).

This PR makes more progress on LP: [#1752481](https://bugs.launchpad.net/ubuntu/+source/snapcraft/+bug/1752481) by emptying the cache of all libs not already primed. This may slow things down a little, but will further isolate each part, preventing these kinds of issues.

### DIFF
--- a/snapcraft/internal/elf.py
+++ b/snapcraft/internal/elf.py
@@ -65,10 +65,14 @@ class SonameCache:
         """Initialize a cache for sonames"""
         self._soname_paths = dict()  # type: Dict[str, str]
 
-    def reset(self):
-        """Reset the cache values that are empty."""
-        self._soname_paths = {k: v for (k, v) in self._soname_paths.items()
-                              if v is not None}
+    def reset_except_root(self, root):
+        """Reset the cache values that aren't contained within root."""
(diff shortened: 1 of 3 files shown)


snapcraft/internal/elf.py
```diff
@@ -65,10 +65,14 @@ def __init__(self):
...
 
-    def reset(self):
-        """Reset the cache values that are empty."""
-        self._soname_paths = {k: v for (k, v) in self._soname_paths.items()
-                              if v is not None}
+    def reset_except_root(self, root):
+        """Reset the cache values that aren't contained within root."""
+        new_soname_paths = {}  # type: Dict[str, str]
+        for key, value in self._soname_paths.items():
+            if value is not None and value.startswith(root):
```
sergiusens: I might be stretching it, but `os.path.exists` might be interesting (although I do not know how it would happen, it maybe can).

PR #1988, opened 2018-03-07
elf: only set rpaths to libs of the same arch

kyrofa:

-----

This PR is a spike RFC. Tests will fail. This has been tested, and works, but performance will take a hit. I'll be working on the tests (and rebasing onto #1959) tomorrow, but want to get it up tonight so you guys can verify the approach.

Currently Snapcraft looks for libs by soname, and gets tripped up if it finds multiple sonames that match (i.e. it takes the first one). This happens in the case of libc coming from the core snap, which includes both amd64's and i386's libc, and it takes i386's libc by default.

Fix this issue by matching architecture in addition to soname.

### DIFF
--- a/snapcraft/internal/elf.py
+++ b/snapcraft/internal/elf.py
@@ -38,6 +38,22 @@ from snapcraft.internal import (
 logger = logging.getLogger(__name__)
 
 
+class NeededLibrary:
+    """Represents an ELF library version."""
+
+    def __init__(self, *, name: str) -> None:
+        self.name = name
+        self.versions = set()  # type: Set[str]
+
+    def add_version(self, version: str) -> None:
+        self.versions.add(version)
+
+
+ElfArchitectureTuple = Tuple[str, str, str]
+ElfDataTuple = Tuple[ElfArchitectureTuple, str, str, Dict[str, NeededLibrary], bool]  # noqa: E501
+SonameCacheDict = Dict[Tuple[ElfArchitectureTuple, str], str]
+
+
 # Old pyelftools uses byte strings for section names.  Some data is
 # also returned as bytes, which is handled below.
 if parse_version(elftools.__version__) >= parse_version('0.24'):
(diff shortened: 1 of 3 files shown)


snapcraft/internal/elf.py
```diff
@@ -137,8 +138,12 @@ def _crawl_for_path(*, soname: str, root_path: str,
                 if file_name == soname:
                     file_path = os.path.join(root, file_name)
                     if ElfFile.is_elf(file_path):
-                        soname_cache[soname] = file_path
-                        return file_path
+                        # We found a match by name, anyway. Let's verify that
+                        # the architecture is the one we want.
+                        elf_file = ElfFile(path=file_path)
```
sergiusens: list comprehension of the elf files we already have loaded as a first option might be faster here.

snapcraft/internal/elf.py
```diff
@@ -137,8 +138,12 @@ def _crawl_for_path(*, soname: str, root_path: str,
...
+                        if elf_file.arch == arch:
+                            soname_cache[soname] = file_path
```
jhenstridge: One concern here is that if the same SonameCache instance gets reused to crawl dependencies of two ELF files of different architectures, the second will match entries added by the first.

One way to handle this would be to make the keys in the SonameCache be `(arch, soname)`.  This way it doesn't matter if cache entries are added for other architectures because you'll never look up those keys.

kyrofa:
Yeah, go team!

snapcraft/internal/elf.py
kyrofa: Oh great catch, that would have been one nasty bug.

snapcraft/internal/elf.py
```diff
@@ -56,41 +72,41 @@ def __getitem__(self, key):
         return self._soname_paths[key]
 
     def __setitem__(self, key, item):
```
kyrofa: We've probably moved beyond using `__setitem__` now, we could make our API more explicit.

PR #1993, opened 2018-03-08
core: initial support for bases

sergiusens:
Closes: #1659

-----

### DIFF
--- a/snapcraft/_options.py
+++ b/snapcraft/_options.py
@@ -224,22 +224,24 @@ class ProjectOptions:
         self._set_machine(target_deb_arch)
         self.__debug = debug
 
-    @property
-    def is_host_compatible_with_base(self):
+    def is_host_compatible_with_base(self, base):
         codename = None
         with suppress(errors.OsReleaseCodenameError):
             codename = OsRelease().version_codename()
             logger.debug('Running on {!r}'.format(codename))
 
-        # TODO support more bases
-        return codename == 'xenial' or codename == 'trusty'
+        # TODO do this more generically
+        if base == 'core18':
+            return codename in ('trusty', 'xenial', 'bionic')
+        else:
+            return codename in ('trusty', 'xenial')
 
-    def get_core_dynamic_linker(self):
+    def get_core_dynamic_linker(self, base):
         """Returns the dynamic linker used for the targeted core.
         If not found realpath for `/lib/ld-linux.so.2` is returned.
         However if core is not installed None will be returned.
         """
-        core_path = common.get_core_path()
+        core_path = common.get_core_path(base)
         dynamic_linker_path = os.path.join(
             core_path,
             self.__machine_info.get('core-dynamic-linker',
--- a/snapcraft/_options.py
+++ b/snapcraft/_options.py
@@ -20,7 +20,7 @@ import os
 import platform
 import sys
 from contextlib import suppress
-from typing import Set  # noqa: F401
+from typing import List, Set  # noqa: F401
 
 from snapcraft.internal import common, errors
 from snapcraft.internal.deprecations import handle_deprecation_notice
@@ -106,6 +106,16 @@ _32BIT_USERSPACE_ARCHITECTURE = {
 
 _WINDOWS_TRANSLATIONS = {
     'AMD64': 'x86_64'
+}
+
+
+_HOST_CODENAME_FOR_BASE = {
+    'core18': 'bionic',
+    'core': 'xenial',
+}
+_HOST_COMPATIBILITY = {
+    'xenial': ['trusty', 'xenial'],
+    'bionic': ['trusty', 'xenial', 'bionic'],
 }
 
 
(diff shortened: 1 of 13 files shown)


elopio:
This is failing because we fake armhf on the second unit test suite. Then, the linker is /snap/core/current/lib/ld-linux-armhf.so.3, which is not found on the travis runners of course.

To pass, I think the best option would be to patch the machine translations dictionary to change this path to one that the test can write, and write and empty file on that patched path.

tools/travis/run_lxd_container.sh
```diff
@@ -47,13 +47,15 @@ $lxc config set "$name" environment.TEST_SNAP_WITH_TRACKS "$TEST_SNAP_WITH_TRACK
...
 $lxc config set "$name" environment.GITHUB_TEST_PASSWORD "$GITHUB_TEST_PASSWORD"
 $lxc config set "$name" environment.SNAPCRAFT_AUTOPKGTEST_SECRET "$SNAPCRAFT_AUTOPKGTEST_SECRET"
-$lxc config set "$name" environment.SNAPCRAFT_TEST_MOCK_MACHINE "$SNAPCRAFT_TEST_MOCK_MACHINE"
 $lxc config set "$name" environment.GH_TOKEN "$GH_TOKEN"
 $lxc config set "$name" environment.CODECOV_TOKEN "$CODECOV_TOKEN"
 $lxc config set "$name" environment.SNAPCRAFT_AUTOPKGTEST_COOKIE "$SNAPCRAFT_AUTOPKGTEST_COOKIE"
 $lxc config set "$name" environment.GH_TOKEN_PPA_AUTOPKGTEST_RESULTS "$GH_TOKEN_PPA_AUTOPKGTEST_RESULTS"
 $lxc config set "$name" environment.LC_ALL "C.UTF-8"
 $lxc config set "$name" environment.SNAPCRAFT_FROM_SNAP "1"
+if [ -n "$SNAPCRAFT_TEST_MOCK_MACHINE" ]; then
```
kyrofa: I'm assuming this change is to ensure the `machine` is actually None [here](https://github.com/snapcore/snapcraft/pull/1993/files#diff-6ad0160b1cd2b6182140f009f57c56aeR149) as opposed to an empty string?

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -569,7 +579,7 @@ def prime(self, force=False) -> None:  # noqa: C901
                 root_path=self.primedir,
                 snap_base_path=self._snap_base_path)
         elif is_classic:
-            dynamic_linker = self._project_options.get_core_dynamic_linker()
+            dynamic_linker = linker
```
kyrofa: At this point, this kind of needs to be the pull path to the linker, but because of [this](https://github.com/snapcore/snapcraft/pull/1993/files#diff-660a35e421f136410b0d3e5168c7a62cR557) it's only the basename. We then use it to patch the interpreter :stuck_out_tongue: .

snapcraft/internal/common.py
```diff
@@ -81,9 +81,9 @@ def run_output(cmd, **kwargs):
             return output.decode('latin-1', 'surrogateescape').strip()
 
 
-def get_core_path():
-    """Returns the path to the core snap."""
-    return os.path.join(os.path.sep, 'snap', 'core', 'current')
+def get_core_path(base):
+    """Returns the path to the core base snap."""
```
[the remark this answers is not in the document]
sergiusens: the keyword is base, but it is a core type of snap, it is complicated due to $reasons

tools/travis/run_lxd_container.sh
sergiusens: yes, for the first time in my life I do a `is None` and I get this :confused:

snapcraft/internal/pluginhandler/__init__.py
sergiusens: I've fixed this in the final way actually; a way that is independent of the base using the FHS for each arch

snapcraft/_options.py
```diff
@@ -224,22 +224,24 @@ def __init__(self, use_geoip=False, parallel_builds=True,
...
-    def is_host_compatible_with_base(self):
+    def is_host_compatible_with_base(self, base):
         codename = None
         with suppress(errors.OsReleaseCodenameError):
             codename = OsRelease().version_codename()
             logger.debug('Running on {!r}'.format(codename))
 
-        # TODO support more bases
-        return codename == 'xenial' or codename == 'trusty'
+        # TODO do this more generically
```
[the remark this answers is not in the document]
sergiusens: I'll get this going

sergiusens:
@kyrofa thanks for the review, I've fixed the interpreter patching to work.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -579,7 +579,8 @@ def prime(self, force=False) -> None:  # noqa: C901
                 root_path=self.primedir,
                 snap_base_path=self._snap_base_path)
         elif is_classic:
-            dynamic_linker = linker
+            dynamic_linker = self._project_options.get_core_dynamic_linker(
```
kyrofa: Yeah there we go.

snapcraft/_options.py
```diff
@@ -224,33 +235,58 @@ def __init__(self, use_geoip=False, parallel_builds=True,
...
+        return codename in compatible_hosts
 
-    def get_core_dynamic_linker(self, base):
+    def get_core_dynamic_linker(self, base: str, expand: bool=True) -> str:
         """Returns the dynamic linker used for the targeted core.
-        If not found realpath for `/lib/ld-linux.so.2` is returned.
-        However if core is not installed None will be returned.
+
+        :param str base: the base core snap to search for linker.
+        :param bool expand: expand the linker to the actual linker if True,
```
kyrofa: We have sort of similar symlink following logic elsewhere, and the parameters we tend to use (and the ones Python itself tends to use) tend to say something about the symlinks themselves. For example, `follow_symlinks`. Would it make sense to name this parameter `follow_symlinks` to indicate what exactly is happening?

sergiusens: about `follow_symlinks`, the semantics are a bit different, as is the logic, not sure I prefer it. Feel free to counter back :slightly_smiling_face:

PR #1995, opened 2018-03-08
options: introduce Project and ProjectInfo

kalikiana:

-----
This PR introduces a new **Project** class which supersedes but also subclasses **ProjectOptions** so that it's backwards-compatible while.exposing additional API through accessors.

The first new accessor `Project.info` yields a **ProjectInfo** object which exposes details of the snap being built, namely `name`, `version`, `summary`, `description` and `confinement` verbatim, as set or not set in the YAML. This is gathered from the `snapcraft.yaml` file which previously wasn't exposed anywhere. Notably plugins will have access to `project.info` via the instance passed to the constructor.

Note that a call to the `ProjectOptions()` instructor won't have its `info` set. Code that requires the new API will have to use the instance provided **or** set `Project.info` to a valid `ProjectInfo` instance. `ProjectLoader.Config` sets this after the fact because it needs to re-use an existing `ProjectOptions` instance from CLI.

More API (`ProjectEnvironment`, `ProjectPartInfo`, `ProjectAppInfo`) is going to be added in follow-up branches.

The following new tests are being added:

 - tests.unit.project_loader.test_config.ProjectInfoTestCase
 - To verify that the `ProjectInfo` API works.
 - tests.unit.project_loader.test_config.ProjectTestCase
 - To verify `Project` exposes `info` from the YAML
 - ...and returns the right values
 - ...and is compatible with `ProjectOptions`
 - ...and doesn't fail when there's no summary
 - ...and doesn't fail when there's no description

I locally ran the tests:

Manual test steps:
 - git clone https://git.launchpad.net/subiquity
 - cd subiquity
 - snapcraft
 - git clone git@github.com:LaughingLove/branch-diff-snap.git
 - cd branch-diff-snap
 - snapcraft


snapcraft/_options.py
```diff
@@ -124,7 +124,68 @@ def _get_platform_architecture():
...
+    def __init__(self) -> None:
+        self.__info = None
+
+    @property
+    def info(self):
+        """Returns the project info.
+        """
+        return self.__info
+
+    @info.setter
```
kyrofa: What is the goal, here? What do we gain beyond simply setting `self.info = None` within `__init__`? I'd rather see good-old functions, personally. Gives us an API that scales a bit better.

snapcraft/_options.py
```diff
@@ -124,7 +124,68 @@ def _get_platform_architecture():
...
+        self.__version = version
+        self.__summary = summary
+        self.__description = description
+        self.__confinement = confinement
+
+    @property
+    def name(self):
+        """Returns the name of the snap.
+        """
+        return self.__name
```
kyrofa: I think these should be actual functions. Otherwise they look like attributes until someone tries to set them. Functions provide a more obvious and easy-to-use API.

snapcraft/internal/project_loader/_config.py
```diff
@@ -108,6 +108,13 @@ def __init__(self, project_options=None):
         _ensure_grade_default(snapcraft_yaml, self._validator.schema)
 
         self.data = self._expand_env(snapcraft_yaml)
+        self._project_options.info = snapcraft.ProjectInfo(
+            name=self.data['name'],
+            version=self.data['version'],
+            summary=self.data.get('summary', ''),
+            description=self.data.get('description', ''),
+            confinement=self.data['confinement'],
```
kyrofa: Why leave extracting the relevant information to the caller? `ProjectInfo` could just be handed the YAML and extract stuff on its own.

snapcraft/_options.py
```diff
@@ -124,7 +124,68 @@ def _get_platform_architecture():
...
+        """Returns the project info.
+        """
+        return self.__info
+
+    @info.setter
+    def info(self, info):
+        self.__info = info
+
+
+class ProjectOptions(Project):
```
kyrofa: How much work would it be to flip this relationship around and replace all the `ProjectOptions` with `Project`s? We're gonna have to do it at some point.

snapcraft/internal/project_loader/_config.py
```diff
@@ -108,6 +108,13 @@ def __init__(self, project_options=None):
         _ensure_grade_default(snapcraft_yaml, self._validator.schema)
 
         self.data = self._expand_env(snapcraft_yaml)
+        self._project_options.info = snapcraft.ProjectInfo(
```
kyrofa: Quick note based on the PR description: eventually the CLI should be passing in a `ProjectEnvironment` instance and the `Project` can be assembled here, removing the need to set attributes.

snapcraft/_options.py
kalikiana: How is this different to how we use properties elsewhere?

snapcraft/_options.py
kalikiana: As said in the PR description, info needs to be set after creating the instance. It's a setter because otherwise the property would be read-only.
But I'm happy to make it a method for clarity.

snapcraft/internal/project_loader/_config.py
kalikiana: I thought that's the style we prefer, but no strong opinion. I'm fine with handling extraction in the constructor.

snapcraft/_options.py
kalikiana: That would mean we can never drop `ProjectOptions` or its properties from the API. Right now you have the option of creating a `Project` instance without it.

snapcraft/internal/project_loader/_config.py
kalikiana: I'll add a comment.

snapcraft/_options.py
```diff
@@ -280,6 +281,30 @@ def _set_machine(self, target_deb_arch):
...
+        self.description = data.get('description')
+        self.confinement = data['confinement']
+
+
+class Project(ProjectOptions):
+    """All details around building a project concerning the build environment
+    and the snap being built."""
+
+    # Mirror ProjectOptions constructor here to guarantee order of arguments
+    # which would not be preserved with **kwargs
```
kyrofa: But the only one we care about is [this](https://github.com/snapcore/snapcraft/pull/1995/files#diff-1b5470a9f2e177fc1606479628098324R68) one, right? Take the time to design this `__init__` the way we want it, and change the callers as necessary. At least make these keyword-only args, which should continue to work fine with [this call](https://github.com/snapcore/snapcraft/pull/1995/files#diff-1b5470a9f2e177fc1606479628098324R68).

snapcraft/internal/project_loader/_config.py
```diff
@@ -88,7 +88,7 @@ def _remote_parts(self):
 
     def __init__(self, project_options=None):
         if project_options is None:
-            project_options = snapcraft.ProjectOptions()
+            project_options = snapcraft.Project()
```
kyrofa: This should probably be called `project` now, not `project_options`.

snapcraft/_options.py
kalikiana: For the record: clarified in an offline conversation that this would be confusing without changing all the classes used by `Config` and porting all the consumers would make this PR too big so I'm going to do that in a follow-up branch.

snapcraft/internal/project_loader/_config.py
kalikiana: For the record: clarified in an offline conversation that this would be confusing without changing all the classes used by `Config` and porting all the consumers would make this PR too big so I'm going to do that in a follow-up branch.

snapcraft/_options.py
kalikiana: Ack. I changed the constructor and the caller in CLI.

snapcraft/__init__.py
```diff
@@ -382,6 +382,8 @@ def _get_version():
 
 
 from snapcraft._baseplugin import BasePlugin        # noqa
+from snapcraft._options import Project              # noqa
+from snapcraft._options import ProjectInfo          # noqa
```
sergiusens: is there a reason these need to be in `__init__`? Let's please avoid adding things to here.

sergiusens:
please create a package called `project`, create separate modules for each (I suspect this will get big if it starts to be our central point of communication). Also in _options, rename it to _project and rename ProjectOptions (the class) to just Project

snapcraft/_options.py
```diff
@@ -280,6 +281,28 @@ def _set_machine(self, target_deb_arch):
...
+
+    def __init__(self, data: Dict[str, Any]) -> None:
+        self.name = data['name']
+        self.version = data['version']
+        self.summary = data.get('summary')
+        self.description = data.get('description')
+        self.confinement = data['confinement']
+
+
+class Project(ProjectOptions):
```
sergiusens: or call this ProjectCLIOptions

snapcraft/internal/project_loader/_config.py
```diff
@@ -86,12 +89,12 @@ def _remote_parts(self):
...
-    def __init__(self, project_options=None):
+    def __init__(self, project_options: project.Project=None) -> None:
         if project_options is None:
-            project_options = snapcraft.ProjectOptions()
+            project_options = project.Project()
 
-        self.build_snaps = set()
-        self.build_tools = []
+        self.build_snaps = set()  # type: Set[str]
+        self.build_tools = set()  # type: Set[str]
```
kyrofa: Hold up, this is a larger change-- are you sure you want to change build_tools from a list to a set?

snapcraft/project/_project_options.py
```diff
@@ -17,122 +17,24 @@
...
 from snapcraft.internal.os_release import OsRelease
 
 
-logger = logging.getLogger(__name__)
+from ._arch_mapping import (
+    _ARCH_TRANSLATIONS,
+    _HOST_CODENAME_FOR_BASE,
+    _HOST_COMPATIBILITY,
+    _get_platform_architecture,
+    _find_machine
```
kyrofa: What is the benefit of extracting these if they're aren't encapsulated at all?

snapcraft/internal/lxd/_containerbuild.py
```diff
@@ -39,7 +39,7 @@
         ContainerSnapcraftCmdError,
         SnapdError,
 )
-from snapcraft._options import _get_deb_arch
+from snapcraft.project._arch_mapping import _get_deb_arch
```
kyrofa: You're importing a private method from a private module into a completely different part of the project. It implies that perhaps this functionality should live elsewhere.

snapcraft/project/_arch_mapping.py
```diff
@@ -0,0 +1,142 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+import platform
+import sys
+
+from snapcraft.internal import errors
+
+
+_ARCH_TRANSLATIONS = {
```
kyrofa: At least according to how I interpret the underscores (happy to be corrected), an underscore here indicates that this variable is intended to be private to this module. However, we directly import it into other modules. That means it's not actually private. Consider either renaming this, or reworking things to properly encapsulate it.

snapcraft/internal/project_loader/_config.py
kalikiana: As clarified offline, it already needs to be a set, ie. we use `|= set()` further down in the same constructor. But we can do even better, by not initializing it here since we're overwriting it anyway.

snapcraft/project/_project_options.py
kalikiana: I'm moving them back for now, as discussed offline, and we tackle do this properly in another PR.

snapcraft/project/_project.py
```diff
@@ -0,0 +1,29 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from ._project_options import ProjectOptions           # noqa
+from ._project_info import ProjectInfo                 # noqa
```
kyrofa: Hmm. What is this?

snapcraft/project/_project.py
```diff
@@ -0,0 +1,29 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from ._project_options import ProjectOptions           # noqa
```
kyrofa: You inherit from this, why noqa?

snapcraft/project/_project_info.py
```diff
@@ -0,0 +1,28 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from typing import Any, Dict  # noqa: F401
```
kyrofa: Why noqa?

snapcraft/project/_project.py
kalikiana: Good call, this needs to be `F401`.

snapcraft/project/_project.py
kalikiana: Indeed, doesn't need it here.

snapcraft/project/_project_info.py
kalikiana: I think I mistakenly copied it over. Not needed indeed.

snapcraft/project/_project.py
kyrofa: I'm sorry, I should have made my question more clear, as it still stands: why is this imported at all? It's not used here.

snapcraft/project/_project.py
```diff
@@ -0,0 +1,29 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from ._project_options import ProjectOptions
+from ._project_info import ProjectInfo                 # noqa: F401
```
kyrofa: Ah, that thread disappeared, I'll ask it here again:

I'm sorry, I should have made my question more clear, as it still stands: why is this imported at all? It's not used here.

kalikiana: It is used as a type annotation in a comment here https://github.com/kalikiana/snapcraft/blob/c30a72598b6f4132d99523bc3eaf8a654a95587e/snapcraft/project/_project.py#L27

kyrofa: Oh man... I'm blind, thank you for clarifying.

PR #1998, opened 2018-03-08
pluginhandler: add {pre,post}-build scriptlets

### LINKED ISSUE #1671: Rename prepare/build/install to pre-build/build/post-build, and deprecate prepare/build/install
Estimate: 1 day

kyrofa:

-----

Snapcraft currently supports three scriptlets:

  - `prepare`: run before build
  - `build`: replaces plugin's build
  - `install`: run after build

The end goal is to have scriptlets surrounding each lifecycle step. To that end, this PR resolves #1671 by adding some generic scriptlet scaffolding to make future additions easier, and by renaming `prepare` and `install` to `pre-build` and `post-build`, respectively. The old names are deprecated (https://github.com/canonical-docs/snappy-docs/pull/371).


schema/snapcraft.yaml
```diff
@@ -324,13 +324,27 @@ properties:
...
+          - not:
+              type: object
+              required: [prepare, pre-build]
+              validation-failure:
+                  "{.instance} cannot contain both 'prepare' and 'pre-build'
+                  keywords. Use 'pre-build'."
+          # Make sure install/post-build are mutually exclusive
+          - not:
+              type: object
+              required: [install, post-build]
```
sergiusens: we should make sure that `pre-build` and `install` do not show up together either

sergiusens: and viceversa

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -355,15 +360,9 @@ def ignore(directory, files):
...
-        script_runner = ScriptRunner(builddir=self.plugin.build_basedir)
-
-        script_runner.run(scriptlet=self._part_properties.get('prepare'))
-        build_scriptlet = self._part_properties.get('build')
-        if build_scriptlet:
-            script_runner.run(scriptlet=build_scriptlet)
-        else:
-            self.plugin.build()
-        script_runner.run(scriptlet=self._part_properties.get('install'))
+        self._runner.pre_build()
```
sergiusens: I am still on the edge about this, but we should probably make `pre/post-build` actual lifecycle stages that do things (what `prepare-` does today`) and have real things take place there (which could be called or setup by the script), in way:

```
pre-build: |
    snapcraft_install_stage_packages
    patch -p1 file < ...
```

and so on and so forth. Let's talk about this on Monday!

kyrofa: > we should probably make pre/post-build actual lifecycle stages that do things [...]

That's exactly what I'm shooting for with [this logic](https://github.com/snapcore/snapcraft/pull/1998/files#diff-313e8dc0aa8d02f56e6ec918c25ddd93R29) to have default actions for pre/post-build, it's just keeping that logic in the runner instead of the pluginhandler, so the pluginhandler can just set it up in `__init__` and not worry about it anymore (it's pretty bloated). Does that align with the goals, here?

kyrofa: From your example, perhaps you're simply wanting to rewrite our current `prepare_pull` and `prepare_build` as default `pre-pull` and `pre-build` scriptlets that calls the `snapcraft_{fetch,unpack}_stage_packages` function we define to fetch/install stage-packages. That's an interesting idea.

sergiusens: let's sync up after the weekly call today

kyrofa:
After our conversation I'm gonna re-shuffle this and land things in a different order.

PR #2002, opened 2018-03-13
many: add snapcraftctl command for scriptlets

### LINKED ISSUE #1675: Base logic for metadata setters set-
Estimate: 4 days

kyrofa:

-----

There is the desire for scriptlets to be able to communicate with Snapcraft, e.g. to request that it use a specific name, or version, or do a specific action, such as fetch stage packages, or run a lifecycle step.

This PR resolves #1675 by adding the basic framework for this communication, adding a new CLI
utility called `snapcraftctl` that is used to communicate with the parent Snapcraft process. It proves this out by adding a new subcommand to this utility called `build` which, if called, will run the plugin's build step. It also adds the `overide-build` scriptlet, which can use `snapcraftctl build` like so:

    # ...
    override-build: |
      snapcraftctl build
    # ...

Note that, as a side-effect of this work, LP: [#1751749](https://bugs.launchpad.net/snapcraft/+bug/1751749) is fixed as well.


snapcraft/internal/errors.py
```diff
@@ -517,3 +517,17 @@ class InvalidExtractorValueError(MetadataExtractionError):
 
     def __init__(self, path: str, extractor_name: str) -> None:
         super().__init__(path=path, extractor_name=extractor_name)
+
+
+class ScriptletError(SnapcraftError):
+    pass
```
sergiusens: I've noticed that instead of pass we could also add a simple docstring to this explaining that it is a base class for others (and even if it will bring in inconcistency, we should add Base to the name, wdyt?)

snapcraft/internal/pluginhandler/_runner.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+class Runner:
+    """The Runner class is responsible for orchestrating scriptlets."""
+
+    # FIXME: Need to quote builtin_functions typing because of
+    # https://github.com/python/typing/issues/259 which is fixed in Python
+    # 3.5.3.
+    def __init__(self, *, part_properties: Dict[str, Any], builddir: str,
+                 builtin_functions: 'Dict[str, Callable[..., None]]') -> None:
```
sergiusens: try creating a type from `'Dict[str, Callable[..., None]]`, look at elf.py for inspiration

kyrofa: I can't actually get it to work with the quoting, keeps telling me it's an invalid type.

snapcraft/internal/errors.py
kyrofa: Agreed on both points, done.

snapcraft/internal/pluginhandler/_runner.py
```diff
@@ -0,0 +1,193 @@
...
+                ['/bin/sh', '-e', '-c', script], cwd=self._builddir)
+
+            status = None
+            try:
+                while status is None:
+                    function_call = call_fifo.read()
+                    if function_call:
+                        self._handle_builtin_function(
+                            scriptlet_name, function_call.strip())
+                        feedback_fifo.write('success\n')
```
sergiusens: why do we write `success\n`?

snapcraft/internal/pluginhandler/_runner.py
```diff
@@ -0,0 +1,193 @@
...
+                    status = process.poll()
+            finally:
+                call_fifo.close()
+                feedback_fifo.close()
+
+            if status:
+                raise errors.ScriptletRunError(
+                    scriptlet_name=scriptlet_name, code=status)
```
sergiusens: this I think solves a bug with scriptlet errors you might want to take attributes for

snapcraft/internal/pluginhandler/_runner.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def _handle_builtin_function(self, scriptlet_name, function_call):
+        function_call_components = shlex.split(function_call)
+        if len(function_call_components) == 0:
+            # This means a snapcraft developer screwed up. Should never be
```
sergiusens: let's try and avoid the word "screwed", it feels a bit confrontational :-)

snapcraft/internal/pluginhandler/_runner.py
kyrofa: All that's actually required is a '\n' to get the [read](https://github.com/snapcore/snapcraft/pull/2002/files#diff-8b247ab4b649deb9ee488e09837eba39R33) to stop waiting for the call to be handled, but that felt a little too magical for my taste :stuck_out_tongue: . This could eventually grow into actually passing information back to the scriptlet from the builtins if we want.

snapcraft/internal/pluginhandler/_runner.py
kyrofa: This in combination with the fact that we're no longer using a temporary script should indeed fix LP: [#1751749](https://bugs.launchpad.net/snapcraft/+bug/1751749), thank you for reminding me.

snapcraft/internal/pluginhandler/_runner.py
kyrofa: Of course, good call. Fixed.

snapcraft/internal/pluginhandler/_runner.py
```diff
@@ -0,0 +1,193 @@
...
+
+            status = None
+            try:
+                while status is None:
+                    function_call = call_fifo.read()
+                    if function_call:
+                        self._handle_builtin_function(
+                            scriptlet_name, function_call.strip())
+                        feedback_fifo.write('success\n')
+                    status = process.poll()
```
chipaca: I'd add a time.sleep(.1) here (or .01 if you're impatient :-)

snapcraft/internal/pluginhandler/_runner.py
```diff
@@ -0,0 +1,193 @@
...
+                '{!r} scriptlet called an undefined builtin function: '
+                '{}'.format(scriptlet_name, function_call)) from e
+
+
+class _NonBlockingFifo:
+
+    def __init__(self, *, path: str, permissions) -> None:
+        os.mkfifo(path)
+        self.path = path
+        self._permissions = permissions
```
chipaca: I'd get rid of 'permissions' entirely, and always open the fifo `O_RDWR`. This'd allow you to drop the `_fifo_open` method entirely and just open in in `__init__`.

kyrofa: Ah! A genius you are! That was the part of this that was really bothering me.

chipaca: merely a reader of manpages :-D

snapcraft/internal/pluginhandler/_runner.py
kyrofa: Yeah good call, thank you :) .

snapcraft/cli/__main__.py
```diff
@@ -21,6 +21,7 @@
 import subprocess
 
 from ._runner import run
+from .snapcraftctl._runner import run as run_snapcraftctl  # noqa
```
sergiusens: why is this needed here?

sergiusens: oh, never mind, console_scripts

snapcraft/cli/snapcraftctl/_runner.py
```diff
@@ -0,0 +1,80 @@
...
+
+import click
+
+from snapcraft.internal.errors import SnapcraftEnvironmentError
+from snapcraft.cli._errors import exception_handler
+from snapcraft.internal import log
+
+
+@click.group()
+@click.option('--debug', '-d', is_flag=True)
```
sergiusens: it would be nice if this could be picked up from the environment as well `SNAPCRAFT_DEBUG` (click supports this) so then it is hooked up with snapcraft's flag too

snapcraft/cli/__main__.py
kyrofa: Yeah, didn't want to duplicate the whole "set UTF-8" thing.

snapcraft/cli/snapcraftctl/_runner.py
kyrofa: Oh good call, fixed in the most recent commit.

bin/snapcraftctl
```diff
@@ -0,0 +1,47 @@
...
+snapcraft.cli.__main__.run_snapcraftctl(prog_name=\"snapcraftctl\")
+' ""$@"
+
+
+# We don't actually want a 100% clean environment. Pass on the SNAP variables,
+# locale settings, and environment variables required by snapcraftctl itself.
+/usr/bin/env -i -- sh -<<END
+# Required for snapcraftctl to actually find snapcraft when snapped via
+# sitecustomize
+export SNAP="$SNAP"
```
kyrofa: Only set if defined.

PR #2004, opened 2018-03-14
errros: add a specific error when running commands from plugins

sergiusens:
When a plugin fails a horrible traceback is printed because there is no
SnapcraftError class handling the situation.

LP: #1753965

-----

### DIFF
--- a/snapcraft/_baseplugin.py
+++ b/snapcraft/_baseplugin.py
@@ -205,7 +205,7 @@ class BasePlugin:
         except CalledProcessError as process_error:
             raise errors.SnapcraftPluginCommandError(
                 command=cmd, part_name=self.name,
-                exit_code=process_error.returncode)
+                exit_code=process_error.returncode) from process_error
 
     def run_output(self, cmd, cwd=None, **kwargs):
         if not cwd:
@@ -216,4 +216,4 @@ class BasePlugin:
         except CalledProcessError as process_error:
             raise errors.SnapcraftPluginCommandError(
                 command=cmd, part_name=self.name,
-                exit_code=process_error.returncode)
+                exit_code=process_error.returncode) from process_error
--- a/snapcraft/_baseplugin.py
+++ b/snapcraft/_baseplugin.py
@@ -202,12 +202,18 @@ class BasePlugin:
         os.makedirs(cwd, exist_ok=True)
         try:
             return common.run(cmd, cwd=cwd, **kwargs)
-        except CalledProcessError:
+        except CalledProcessError as process_error:
             raise errors.SnapcraftPluginCommandError(
-                command=cmd, part_name=self.name)
+                command=cmd, part_name=self.name,
+                exit_code=process_error.returncode)
 
     def run_output(self, cmd, cwd=None, **kwargs):
         if not cwd:
             cwd = self.builddir
         os.makedirs(cwd, exist_ok=True)
-        return common.run_output(cmd, cwd=cwd, **kwargs)
+        try:
+            return common.run_output(cmd, cwd=cwd, **kwargs)
+        except CalledProcessError as process_error:
+            raise errors.SnapcraftPluginCommandError(
+                command=cmd, part_name=self.name,
+                exit_code=process_error.returncode)
--- a/snapcraft/_baseplugin.py
+++ b/snapcraft/_baseplugin.py
@@ -17,8 +17,9 @@
 import contextlib
 import logging
 import os
+from subprocess import CalledProcessError
 
-from snapcraft.internal import common
+from snapcraft.internal import common, errors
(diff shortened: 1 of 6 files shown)


snapcraft/_baseplugin.py
```diff
@@ -199,7 +200,11 @@ def run(self, cmd, cwd=None, **kwargs):
...
         print(' '.join(cmd))
         os.makedirs(cwd, exist_ok=True)
-        return common.run(cmd, cwd=cwd, **kwargs)
+        try:
+            return common.run(cmd, cwd=cwd, **kwargs)
+        except CalledProcessError:
+            raise errors.SnapcraftPluginCommandError(
+                command=cmd, part_name=self.name)
 
     def run_output(self, cmd, cwd=None, **kwargs):
```
kyrofa: How about similar logic here?

snapcraft/internal/errors.py
```diff
@@ -531,3 +531,17 @@ class InvalidExtractorValueError(MetadataExtractionError):
...
     def __init__(self, path: str, extractor_name: str) -> None:
         super().__init__(path=path, extractor_name=extractor_name)
+
+
+class SnapcraftPluginCommandError(SnapcraftError):
+    """Command executed by a plugin fails."""
+
+    fmt = (
+        'The command {command!r} run for the {part_name!r} part has failed.\n'
+        'Verify that the part is using the correct parameters and try again.'
```
kyrofa: I think the exit code would be helpful as well. Perhaps something like:

```
'Failed to run {command!r} for {part_name!r}: '
'Exited with code {exit_code}.\n'
'Verify that the part is using the correct parameters and try again.'
```

snapcraft/_baseplugin.py
sergiusens: done

snapcraft/internal/errors.py
sergiusens: done

snapcraft/_baseplugin.py
```diff
@@ -199,10 +200,20 @@ def run(self, cmd, cwd=None, **kwargs):
             cwd = self.builddir
         print(' '.join(cmd))
         os.makedirs(cwd, exist_ok=True)
-        return common.run(cmd, cwd=cwd, **kwargs)
+        try:
+            return common.run(cmd, cwd=cwd, **kwargs)
+        except CalledProcessError as process_error:
+            raise errors.SnapcraftPluginCommandError(
+                command=cmd, part_name=self.name,
+                exit_code=process_error.returncode)
```
kyrofa: Agh sorry, missed this before, but you probably want a `from process_error` on both of these.

snapcraft/internal/errors.py
```diff
@@ -531,3 +532,20 @@ class InvalidExtractorValueError(MetadataExtractionError):
...
         super().__init__(path=path, extractor_name=extractor_name)
+
+
+class SnapcraftPluginCommandError(SnapcraftError):
+    """Command executed by a plugin fails."""
+
+    fmt = (
+        'The command {command!r} run for the {part_name!r} part has failed.\n'
+        'Exited with exit code {exit_code}.\n'
+        'Verify that the part is using the correct parameters and try again.'
```
kyrofa: Channeling my inner Leo, this does not meet the agreed-upon error message format which is:

```
<what>: <why>\n
<how to fix>
```

It seems to be

```
<what>\n
<why>\n
<how to fix>
```

snapcraft/plugins/_python/_pip.py
```diff
@@ -196,7 +197,8 @@ def _is_pip_installed(self):
             # Using _run_output here so stdout doesn't get printed to the
             # terminal.
             self._run_output([], stderr=subprocess.STDOUT)
-        except subprocess.CalledProcessError as e:
+        except (subprocess.CalledProcessError,
+                SnapcraftPluginCommandError) as e:
```
kyrofa: Note that `self._run{_output}` here actually goes through `common.run{_output}` as it's not part of a plugin. You shouldn't need to be handling `SnapcraftPluginCommandErrors` here.

snapcraft/plugins/python.py
```diff
@@ -327,7 +328,8 @@ def _install_project(self):
                 # There is also a chance that this setup.py is distutils based
                 # in which case we will rely on the `pip install .` ran before
                 #  this.
-                with contextlib.suppress(subprocess.CalledProcessError):
+                with contextlib.suppress(subprocess.CalledProcessError,
```
kyrofa: Oh good catch, we would have had issues! However, given that this uses `self.run`, there's no longer a need to catch `CalledProcessError` here, is there?

PR #2005, opened 2018-03-16
pluginhandler: special case go patchelf failures for classic confinement

sergiusens:
Classic confinement for go with patchelf is still a little problematic
and has never really work from a confinement perspective. So we warn if
patchelf fails to run through.

Clean up the logic after the introduction of bases and the requirement
to explicitly add libc6 to stage-packages in case of non host
compatibilities

-----

### DIFF
--- a/snapcraft/internal/pluginhandler/__init__.py
+++ b/snapcraft/internal/pluginhandler/__init__.py
@@ -532,11 +532,9 @@ class PluginHandler:
 
         if self._build_attributes.no_patchelf():
             logger.warning(
-                'The primed files for part {!r} files are not going to be '
-                'verified for correctness nor patched to work correctly '
-                'in the environment if required as '
-                '`build-attributes: [no-patchelf]` is set for the '
-                'part.'.format(self.name))
+                'The primed files for part {!r} will not be verified for '
+                'correctness or patched: build-attributes: [no-patchelf] '
+                'is set.'.format(self.name))
         else:
             part_patcher = PartPatcher(
                 elf_files=elf_files,
--- a/snapcraft/internal/pluginhandler/__init__.py
+++ b/snapcraft/internal/pluginhandler/__init__.py
@@ -22,7 +22,7 @@ import os
 import shutil
 import sys
 from glob import glob, iglob
-from typing import Dict, Set  # noqa
+from typing import Dict, Set  # noqa: F401
 
 import yaml
 
@@ -35,6 +35,7 @@ from ._build_attributes import BuildAttributes
 from ._metadata_extraction import extract_metadata
 from ._plugin_loader import load_plugin  # noqa
 from ._scriptlets import ScriptRunner
+from ._patchelf import PartPatcher
 
 logger = logging.getLogger(__name__)
 
@@ -505,7 +506,7 @@ class PluginHandler:
 
         self.mark_cleaned('stage')
 
-    def prime(self, force=False) -> None:  # noqa: C901
+    def prime(self, force=False) -> None:
         self.makedirs()
         self.notify_part_progress('Priming')
         snap_files, snap_dirs = self.migratable_fileset_for('prime')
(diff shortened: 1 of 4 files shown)


snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -22,7 +22,7 @@
 import shutil
 import sys
 from glob import glob, iglob
-from typing import Dict, Set  # noqa
+from typing import Dict, Set  # noqa: F401
```
kyrofa: Oh good idea, I'll do this from now on as well.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -529,79 +530,26 @@ def prime(self, force=False) -> None:  # noqa: C901
...
-            logger.warning(
-                'The following files are not going to be patched to work '
-                'correctly in the environment as '
+                'The primed files for part {!r} files are not going to be '
+                'verified for correctness nor patched to work correctly '
+                'in the environment if required as '
                 '`build-attributes: [no-patchelf]` is set for the '
-                'part:\n{}'.format(
-                    ''.join(['- {}\n'.format(e.path) for e in elf_files])))
+                'part.'.format(self.name))
```
kyrofa: Yeah, I like this better. Although I think it could be worded a little shorter. Maybe:

"The primed files for part {!r} will not be verified for correctness or patched: `build-attributes: [no-patchelf]` is set."

Reasoning: removing a few redundant references to the part, and also assuming that folks know WHY we typically patch, so no need to make the message more verbose than necessary. Thoughts?

snapcraft/internal/pluginhandler/_patchelf.py
```diff
@@ -0,0 +1,141 @@
...
+
+        if self._is_libc6_staged:
+            dynamic_linker = elf.find_linker(
+                root_path=self._primedir,
+                snap_base_path=self._snap_base_path)
+        elif self._is_classic:
+            dynamic_linker = self._project.get_core_dynamic_linker(
+                self._core_base, expand=False)
+        else:
+            # This would be bad
```
kyrofa: Worth raising?

snapcraft/internal/pluginhandler/__init__.py
sergiusens: agreed, how about the symmetric message for the other one?

kyrofa:
When building a classic go snap on Xenial, I get this:

```
patchelf: patchelf.cc:697: void ElfFile<Elf_Ehdr, Elf_Phdr, Elf_Shdr, Elf_Addr, Elf_Off, Elf_Dyn, Elf_Sym>::rewriteSectionsExecutable() [with Elf_Ehdr = Elf64_Ehdr; Elf_Phdr = Elf64_Phdr; Elf_Shdr = Elf64_Shdr; Elf_Addr = long unsigned int; Elf_Off = long unsigned int; Elf_Dyn = Elf64_Dyn; Elf_Sym = Elf64_Sym]: Assertion `(off_t) rdi(hdr->e_shoff) >= startOffset' failed.
Failed to update '/tmp/foo/prime/bin/mdns-publisher'. Retrying after stripping the .note.go.buildid from the elf file.
cannot find section .rela.dyn
An attempt to patch {!r} so that it would work correctly in diverse environments was made and failed. To disable this behavior set `build-attributes: [no-patchelf]` for the part.
```

This seems reasonable, although that crash from patchelf is ugly, and it's made worse by the fact that apport catches it and wants me to report a problem. Is there a way to handle that better?

snapcraft/internal/pluginhandler/_patchelf.py
```diff
@@ -0,0 +1,149 @@
...
+        files_to_patch = elf.get_elf_files_to_patch(self._elf_files)
+        for elf_file in files_to_patch:
+            try:
+                elf_patcher.patch(elf_file=elf_file)
+            except errors.PatcherError as patch_error:
+                logger.warning(
+                    'An attempt to patch {!r} so that it would work '
+                    'correctly in diverse environments was made and failed. '
+                    'To disable this behavior set '
+                    '`build-attributes: [no-patchelf]` for the part.')
```
kyrofa: Missing a `.format` here.

snapcraft/internal/pluginhandler/__init__.py
kyrofa: I'm not sure I understand, which other one? I assumed you were talking about [this one](https://github.com/snapcore/snapcraft/pull/2005/files#diff-660a35e421f136410b0d3e5168c7a62cL590) which you're getting rid of here.

sergiusens: in 8b7c78582c68fa0f03a730af0c76db48e82e47b7
```
The primed files for part 'gotty' will not be verified for correctness or patched: build-attributes: [no-patchelf] is set.
```

sergiusens:
@kyrofa the fact that apport wants to handle a crash for a binary that came from a deb is a good thing. I would indeed not touch that. While I agree it is ugly, it is the output of the tool we chose to use and should not hide it (patchelf is pretty much silent when things go well).

sergiusens:
@kyrofa last but not least, you most likely will not see that error if you had patchelf 0.9-1 from xenial-proposed

sergiusens:
@kyrofa I am assuming, given our conversations elsewhere that you just forgot to approve.

kyrofa:
Indeed, I just wanted those questioned answered before approving, so consider this a retroactive +1 ;) .

PR #2007, opened 2018-03-17
catkin plugin: replace python calls in all profile.d scripts

kyrofa:

-----

Currently, Snapcraft replaces a Python call in `etc/catkin/profile.d/10.ros.sh` that uses an absolute path (that doesn't exist in the core snap) to use just the `python` command instead, thereby ensuring the one from the snap is used. However, a recent ROS Kinetic update switched up the profile.d files, using absolute paths to Python elsewhere, breaking all newly-built ROS Kinetic snaps on Ubuntu Core.

Fix this by taking broader approach, checking _every_ file in `etc/catkin/profile.d/` for absolute Python paths and replacing each if found. Note that this should fix the Catkin Tools plugin as well.

Refer to https://forum.snapcraft.io/t/4518 for more information.

Please test this PR using the `edge/pr-2007-catkin-kinetic-fix` channel:

    sudo snap install snapcraft --channel=edge/pr-2007-catkin-kinetic-fix

Or if you already have it installed:

    sudo snap refresh snapcraft --channel=edge/pr-2007-catkin-kinetic-fix

### DIFF
--- a/snapcraft/plugins/catkin.py
+++ b/snapcraft/plugins/catkin.py
@@ -701,16 +701,15 @@ class CatkinPlugin(snapcraft.BasePlugin):
         # Fix all shebangs to use the in-snap python.
         mangling.rewrite_python_shebangs(self.installdir)
 
-        # Also replace the python usage in 10.ros.sh to use the in-snap python.
-        ros10_file = os.path.join(self.rosdir,
-                                  'etc/catkin/profile.d/10.ros.sh')
-        if os.path.isfile(ros10_file):
-            with open(ros10_file, 'r+') as f:
-                pattern = re.compile(r'/usr/bin/python')
-                replaced = pattern.sub(r'python', f.read())
-                f.seek(0)
-                f.truncate()
-                f.write(replaced)
+        # Also replace all the /usr/bin/python calls in etc/catkin/profile.d/
+        # files with the in-snap python
+        profile_d_path = os.path.join(
+            self.rosdir, 'etc', 'catkin', 'profile.d')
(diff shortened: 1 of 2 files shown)


snapcraft/plugins/catkin.py
```diff
@@ -701,16 +701,15 @@ def _use_in_snap_python(self):
...
-                                  'etc/catkin/profile.d/10.ros.sh')
-        if os.path.isfile(ros10_file):
-            with open(ros10_file, 'r+') as f:
-                pattern = re.compile(r'/usr/bin/python')
-                replaced = pattern.sub(r'python', f.read())
-                f.seek(0)
-                f.truncate()
-                f.write(replaced)
+        # Also replace all the /usr/bin/python calls in etc/catkin/profile.d/
+        # files with the in-snap python
```
kalikiana: If I were to nit-pick, it would be more accurate to say the path is stripped so the in-snap Python is used.
Should this be using the absolute path of the snap, though?

sergiusens: we cannot know it before hand and it ties the build to the name of the snap which we only do for classic (in very specific cases) as it is unavoidable.

sergiusens: if this is not a shebang, it could make use of `SNAP` in the env though

kyrofa: I'm not completely clear on whether these are always shell scripts. I suspect they just need to be executable. As a result, I'm more comfortable using `python` from the $PATH (as we've been doing) instead of trying to figure out the language and use environment variables appropriately.

PR #2035, opened 2018-03-26
ci: cache core and lxd to avoid redownloading

sergiusens:
Download core and lxd once and cache them in snaps-cache for reuse.

-----

### DIFF
--- a/tools/travis/setup_lxd.sh
+++ b/tools/travis/setup_lxd.sh
@@ -28,14 +28,21 @@ done
 apt-get install --yes snapd
 apt-get remove --yes lxd lxd-client
 
-# Use edge because the feature to copy links to the container has not yet been
-# released to stable:
-# https://github.com/lxc/lxd/commit/004e7c361e1d914795d3ba7582654622e32ff193
+if [ ! -f $TRAVIS_BUILD_DIR/snaps-cache/downloaded ]; then
+    mkdir -p "$TRAVIS_BUILD_DIR/snaps-cache"
+    pushd "$TRAVIS_BUILD_DIR/snaps-cache"
+    snap download core
+    snap download lxd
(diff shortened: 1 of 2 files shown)


tools/travis/setup_lxd.sh
```diff
@@ -28,14 +28,21 @@ done
...
 
-# Use edge because the feature to copy links to the container has not yet been
-# released to stable:
-# https://github.com/lxc/lxd/commit/004e7c361e1d914795d3ba7582654622e32ff193
+if [ ! -f $TRAVIS_BUILD_DIR/snaps-cache/downloaded ]; then
+    mkdir -p "$TRAVIS_BUILD_DIR/snaps-cache"
+    pushd "$TRAVIS_BUILD_DIR/snaps-cache"
+    snap download core
+    snap download lxd
+    touch downloaded
```
kyrofa: I don't know about this. I've been [reading up](https://docs.travis-ci.com/user/caching) on the caching, and it sounds like it lives forever unless we manually clear it (gets applied across branches and PRs). That works okay for the snapcraft snap built since we name it according to PR, but not so much here. Perhaps we should also name the core the lxd snap according to PR so they actually get updated? It also sounds like it may not be tremendously fast, so the time savings may not help us much.

sergiusens: Let's take a look at the numbers, I do think the CDN is sometimes too slow

Also, what do you make of this statement: "There is one cache per branch and language version / compiler version / JDK version / Gemfile location, etc."

kyrofa: I think that sentence only applies to the built-in caches (e.g. bundler, cocoapods, etc.)

sergiusens:
no gains with this branch

PR #2041, opened 2018-03-28
kernel plugin: add kmod as build-package

kyrofa:

-----

Currently, in some cases, the kernel plugin shells out to modprobe. However, it doesn't ensure modprobe is actually present, which it isn't in the Launchpad builders.

This PR fixes LP: [#1759592](https://bugs.launchpad.net/snapcraft/+bug/1759592) by adding kmod as a build-package to ensure modprobe is always available. `/sbin` is also added to the PATH, in case it's not already there.


tests/unit/plugins/test_kernel.py
```diff
@@ -742,11 +742,22 @@ def fake_output(*args, **kwargs):
...
+                self.test.assertThat(paths, Contains('/sbin'))
+                return True
+
         self.assertThat(self.run_output_mock.call_count, Equals(1))
         self.run_output_mock.assert_has_calls([
             mock.call([
                 'modprobe', '-n', '--show-depends', '-d',
-                plugin.installdir, '-S', '4.4.2', 'my-fake-module'])])
+                plugin.installdir, '-S', '4.4.2', 'my-fake-module'],
+                env=_check_env(self))])
```
kalikiana: I find passing an object here that calls assertions rather confusing, compared to the `os.environ.copy(); env['PATH']=...; env=env`.

More importantly, there should probably be two separate tests with a fake environment that has and hasn't got sbin in the PATH respectively, so that the outcome doesn't depend on the host?

kyrofa: How do you suggest verifying that `/sbin` would put on the PATH? This has always been my approach, do you know a better one? Regarding depending on the host, that's a good call, but it doesn't require another test, I can just set the PATH for this one.

kyrofa: > I can just set the PATH for this one.

Done in the most recent commit.

kalikiana: Thanks for setting the PATH!

Leave the check as-is then.

PR #2042, opened 2018-03-29
many: implement vendoring support

kalikiana:

-----
This PR introduces a new **vendoring** keyword to `snapcraft.yaml` which accepts a list of host names [as outlined in this proposal](https://forum.snapcraft.io/t/proposal-vendoring-hosts-in-snapcraft-yaml/4739). This is exposed in the API via `ProjectInfo.vendoring`.

Note: This PR requires #1995.

The following new tests are affected:

 - tests.unit.project_loader.test_config.ProjectInfoTestCase
 - To verify that `vendoring` is accessible via the API.
 - tests.unit.project_loader.test_config.ProjectTestCase
 - To verify `ProjectInfo` exposes `vendoring` from the YAML
 - tests.unit.project_loader.test_config.InvalidVendoringTestCase
 - To verify that only valid hosts are accepted as specified in the schema.
 - tests.unit.test_lxd.VendoringTestCase
 - To verify that a squid proxy is installed and configured in a container.

I locally ran the tests:

Manual test steps **with the snap built from this branch**:
 - git clone https://git.launchpad.net/subiquity
 - cd subiquity
 - snapcraft cleanbuild
 - Observe that the build succeeds normally.
 - Update `snapcraft.yaml` by adding this snippet:
       vendoring: [archive.ubuntu.com, security.ubuntu.com]
 - snapcraft cleanbuild
 - Observe that the build fails with a proxy error.
 - Update `snapcraft.yaml` to this:
       vendoring: [api.snapcraft.io, archive.ubuntu.com, security.ubuntu.com, 068ed04f23.site.internapcdn.net, github.com, pypi.python.org]
 - Observe that the build succeeds.

### DIFF
--- a/schema/snapcraft.yaml
+++ b/schema/snapcraft.yaml
@@ -77,6 +77,11 @@ properties:
       - type: string
         validation-failure:
           "{.instance!r} is not a valid hostname."
+        # This format accepts hosts of the form foo or example.com valid.
+        # We don't allow wildcards like .foo or *.example.com because that
+        # would include unwanted resources as for example in the case of
+        # *.launchpad.net which includes ppa.launchpad.net as opposed to being
+        # explicit with [launchpad.net, git.launchpad.net] instead.
         format: hostname
   build-packages:
     $ref: "#/definitions/grammar-array"
--- a/schema/snapcraft.yaml
+++ b/schema/snapcraft.yaml
@@ -66,6 +66,18 @@ definitions:
 title: snapcraft schema
 type: object
 properties:
+  vendoring:
+    type: array
+    validation-failure:
+      "a list of valid host names needs to be provided."
+    description: external sources available to resolve dependencies
+    minItems: 1
+    uniqueItems: true
+    items:
+      - type: string
+        validation-failure:
+          "{.instance!r} is not a valid hostname."
+        format: hostname
   build-packages:
     $ref: "#/definitions/grammar-array"
     description: top level build packages.
--- a/snapcraft/internal/lifecycle/_runner.py
+++ b/snapcraft/internal/lifecycle/_runner.py
@@ -59,6 +59,8 @@ def execute(step, project_options, part_names=None):
     :returns: A dict with the snap name, version, type and architectures.
     """
     config = project_loader.load_config(project_options)
+    if project_options.info.vendoring:
+        logger.warning('Vendoring should be used with a LXD container')
     installed_packages = repo.Repo.install_build_packages(
         config.build_tools)
     if installed_packages is None:
--- a/snapcraft/internal/lifecycle/_runner.py
+++ b/snapcraft/internal/lifecycle/_runner.py
@@ -60,7 +60,12 @@ def execute(step, project_options, part_names=None):
     """
     config = project_loader.load_config(project_options)
     if project_options.info.vendoring:
-        logger.warning('Vendoring should be used with a LXD container')
+        logger.warning(
+            'Vendoring is set to {} but it will not be enforced. '
+            'Snapcraft must build inside a LXD container to setup a proxy '
+            'that limits access to external resources to the specified '
+            'hosts. The build will proceed without restrictions.'.format(
+                ', '.join(project_options.info.vendoring)))
     installed_packages = repo.Repo.install_build_packages(
         config.build_tools)
     if installed_packages is None:
--- a/snapcraft/internal/lifecycle/_runner.py
+++ b/snapcraft/internal/lifecycle/_runner.py
@@ -59,7 +59,7 @@ def execute(step, project_options, part_names=None):
     :returns: A dict with the snap name, version, type and architectures.
     """
     config = project_loader.load_config(project_options)
-    if project_options.info.vendoring:
+    if project_options.info.vendoring and not common.is_lxd_container():
         logger.warning(
             'This snap uses vendoring but it cannot be enforced without '
             'using cleanbuild. Proceeding without restriction.'.format(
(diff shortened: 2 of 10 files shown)


kyrofa:
(this is ready for a rebase)

snapcraft/internal/lifecycle/_runner.py
```diff
@@ -59,6 +59,8 @@ def execute(step, project_options, part_names=None):
     :returns: A dict with the snap name, version, type and architectures.
     """
     config = project_loader.load_config(project_options)
+    if project_options.info.vendoring:
+        logger.warning('Vendoring should be used with a LXD container')
```
kyrofa: We could probably make this a bit more helpful. WHY do we recommend a LXD container to be used? What side effects will NOT using a LXD container introduce?

snapcraft/internal/lxd/_containerbuild.py
```diff
@@ -234,6 +236,44 @@ def _wait_for_network(self):
...
+            logger.info('Vendoring snap to {}'.format(', '.join(vendoring)))
+            # Configure a proxy that blocks access to any hosts which are not
+            # specified by vendoring in snapcraft.yaml.
+            # The rules below mean:
+            # Specify the port of the proxy server.
+            # Allow access to a whitelist on ports for http, https and git.
+            # Deny everything else.
+            self._container_run(['apt-get', 'install', 'squid', '-y'])
+            port = 3129
+            rules = dedent('''
```
kyrofa: Please use triple double quotes, here.

snapcraft/internal/lifecycle/_runner.py
```diff
@@ -59,6 +59,13 @@ def execute(step, project_options, part_names=None):
     :returns: A dict with the snap name, version, type and architectures.
     """
     config = project_loader.load_config(project_options)
+    if project_options.info.vendoring:
+        logger.warning(
+            'Vendoring is set to {} but it will not be enforced. '
+            'Snapcraft must build inside a LXD container to setup a proxy '
+            'that limits access to external resources to the specified '
+            'hosts. The build will proceed without restrictions.'.format(
```
kyrofa: This is better, but there's too much technical detail, and the actual hostnames don't matter, either. How about "This snap uses vendoring, but vendoring cannot be enforced without using cleanbuild. Proceeding without restriction."

snapcraft/internal/lxd/_containerbuild.py
```diff
@@ -234,6 +236,47 @@ def _wait_for_network(self):
...
         logger.info('Network connection established')
+        self._setup_vendoring()
+
+    def _setup_vendoring(self):
+        vendoring = self._project_options.info.vendoring
+        if vendoring:
+            logger.info(
+                'Vendoring is set to {}. Access to external resources not '
+                'specified here will be denied during the build.'
+                .format(', '.join(vendoring)))
```
kyrofa: This message gets nasty quickly if there are several items in the vendoring list. It is, however, helpful to show them here, so I suggest a reformat. Perhaps:

"This snap uses vendoring. In order to enforce it, network access to resources not on the vendor list will be denied. The vendor list is:\n{}" and print each domain on its own line. Something like that, anyway. I also wonder if this should be a warning given that denials are so very opaque.

schema/snapcraft.yaml
```diff
@@ -66,6 +66,18 @@ definitions:
...
+    validation-failure:
+      "a list of valid host names needs to be provided."
+    description: external sources available to resolve dependencies
+    minItems: 1
+    uniqueItems: true
+    items:
+      - type: string
+        validation-failure:
+          "{.instance!r} is not a valid hostname."
+        format: hostname
```
kyrofa: This prevents the use of any sort of wildcard domain, e.g. `.ubuntu.com` to cover both `archive.ubuntu.com` and `security.ubuntu.com`. Squid allows such things, should we?

kyrofa:
I've said this a few times, and my feelings haven't changed: I'm not a big fan of this. Proxy failures are so opaque and unfriendly, I'm afraid this will bite people. However, I know this is what was agreed upon, so I will try to keep my whining to a minimum :wink: .

I'm testing this by running from source, which tries to bring in the deb in the container. However, it tries to do that after it sets up the proxy, which means the container can't bootstrap itself. We should probably consider re-ordering things so that the container is fully functional before we lock down the proxy.

kyrofa:
FYI, I pushed a snap built from this to the `edge/pr-2042` channel (for amd64).

snapcraft/internal/lifecycle/_runner.py
```diff
@@ -59,6 +59,11 @@ def execute(step, project_options, part_names=None):
     :returns: A dict with the snap name, version, type and architectures.
     """
     config = project_loader.load_config(project_options)
+    if project_options.info.vendoring:
+        logger.warning(
+            'This snap uses vendoring but it cannot be enforced without '
+            'using cleanbuild. Proceeding without restriction.'.format(
+                ', '.join(project_options.info.vendoring)))
```
kyrofa: This makes sense when running `snapcraft`, but when I run `snapcraft cleanbuild`, I still get it from the snapcraft instance running inside the container:

```
This snap uses vendoring. In order to enforce it, network access to resources not on the following list will be denied:

foo
[...]
This snap uses vendoring but it cannot be enforced without using cleanbuild. Proceeding without restriction.
```

We should probably clean that up.

snapcraft/internal/lxd/_project.py
```diff
@@ -73,7 +73,11 @@ def _ensure_container(self):
             # Because of https://bugs.launchpad.net/snappy/+bug/1628289
             # Needed to run snapcraft as a snap and build-snaps
             self._container_run(['apt-get', 'install', 'squashfuse', '-y'])
+        if self._remote != 'local':
+            # For remote mounts we will need sshfs.
+            self._container_run(['apt-get', 'install', 'sshfs', '-y'])
         self._inject_snapcraft(new_container=new_container)
+        self._setup_vendoring()
```
kyrofa: At least the last few lines here should probably be refactored into something callable from both here and _cleanbuilder, or we have a good opportunity for diversion.

PR #2043, opened 2018-03-30
cli: support exporting login to stdout

kyrofa:

-----
And instead of simply printing the ini file, base64-encode it so it can be more easily handled. This is intended to fit the use-case of sticking the exported login straight into the environment of a CI system instead of needing to encrypt a file.

### DIFF
--- a/snapcraft/cli/store.py
+++ b/snapcraft/cli/store.py
@@ -298,7 +298,10 @@ def export_login(login_file: str, snaps: str, channels: str, acls: str,
     """Save login configuration for a store account in FILE.
 
     This file can then be used to log in to the given account with the
-    specified permissions.
+    specified permissions. One can also request the login to be exported to
+    stdout instead of a file:
+
+        snapcraft export-login -
 
     For example, to limit access to the edge channel of any snap the account
     can access:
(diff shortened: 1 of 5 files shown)


snapcraft/cli/store.py
```diff
@@ -341,19 +341,34 @@ def export_login(login_file: str, snaps: str, channels: str, acls: str,
...
-    # mypy doesn't have the opener arg in its stub. Ignore its warning
-    with open(login_file, 'w', opener=private_open) as f:  # type: ignore
-        store.conf.save(config_fd=f)
+    # Support a login_file of '-', which indicates a desire to print to stdout
+    if login_file.strip() == '-':
+        echo.info("\nExported login starts on next line:")
+        store.conf.save(config_fd=sys.stdout, encode=True)
+        print()
+
+        login_location = 'and printed above'
```
kalikiana: I'd find it clearer to echo "Login successfully exported and printed above." here than to split up the string like that. Or alternatively set `login_location` to the whole string and use it below.

snapcraft/cli/store.py
```diff
@@ -341,19 +341,34 @@ def export_login(login_file: str, snaps: str, channels: str, acls: str,
     # This is sensitive-- it should only be accessible by the owner
     private_open = functools.partial(os.open, mode=0o600)
 
-    # mypy doesn't have the opener arg in its stub. Ignore its warning
-    with open(login_file, 'w', opener=private_open) as f:  # type: ignore
-        store.conf.save(config_fd=f)
+    # Support a login_file of '-', which indicates a desire to print to stdout
```
kalikiana: Mind adding the `-` to the login command documentation?

snapcraft/cli/store.py
kyrofa: Sure, done.

snapcraft/cli/store.py
kyrofa: Done.

PR #2049, opened 2018-04-04
many: add override-stage scriptlet

### LINKED ISSUE #1673: Add pre-stage/stage/post-stage
Estimate: 1 day

kyrofa:

-----

Currently Snapcraft supports override scriptlets for both the pull and build steps. This PR resolves #1673 by adding one for the stage step called `override-stage` which, if specified, overrides the part's stage step. It also adds a `stage` subcommand to `snapcraftctl` to ensure developers can still complete the default stage action.


kyrofa:
> Note the test failure

Yeah those are the FakeLXD mocking issues I mentioned in chat. I'm fixing them this morning.

tests/fixture_setup.py
```diff
@@ -584,9 +584,23 @@ def _setUp(self):
...
+        def _fake_popen(*args, **kwargs):
+            if '/usr/lib/sftp-server' in args[0]:
+                return self._popen(args[0])
+            elif (args[0][:2] == ['lxc', 'exec'] and self.status and
+                    args[0][2] == self.name and args[0][8] == 'sshfs'):
+                self.files = ['foo', 'bar']
+                return self._popen(args[0])
+
+            # Fall back to the real deal
+            return self._real_popen(*args, **kwargs)
```
kalikiana: Nice solution. This should be standard behavior alas it's not.

PR #2053, opened 2018-04-06
meta: implement pass-through of properties to snap.yaml

kalikiana:

-----
This PR introduces a new **pass-through** keyword to `snapcraft.yaml` which comes in two forms, a global and an app-local property [as proposed in the forum](https://forum.snapcraft.io/t/support-for-passthrough-into-snap-yaml/4698). As proposed, the key is just defined as a dictionary (object in the schema) and not validated, but instead applied in `snap_packaging.py` where its use will be logged and ambiguous keys will raise an error. The original snapcraft.yaml is used to validate what values have been specified as opposed to default values filled in by Snapcraft.
Note that, as discussed offline, properties like `name` or `command` are not special-cased and not supported by pass-through due to their being strings processed by Snapcraft.

The following tests are added/ modified:

 - tests.unit.test_errors
 - To verify AmbiguousPassThroughError
 - tests.unit.test_meta.PassThroughTestCase
 - To verify that a) a new key not present in the schema can be added b) an existing key can be overridden c) a key both present in the schema and pass-through raising an error d) defaults can be overridden rather than error out because of the check for amiguity e) a warning being logged for using pass-through. Each of these is tested with the global and the apps form.

I locally ran the tests:

Manual test steps:
 - cd tests/integration/snaps/basic
 - Add some pass-through properties:
   pass-through:
     foo: bar
     confinement: next-generation
   apps:
     foo:
       command: echo
       daemon: simple
       pass-through:
         foo: bar
         daemon: complex
 - snapcraft
 - Observe the error "Failed to pass-through keys to snap metadata: 'daemon' is specified as both a regular key and in pass-through. Remove one of the keys.".
 - Fix the error by removing "daemon: simple" from the "foo" app.
 - snapcraft
 - Observe the error "Failed to pass-through keys to snap metadata: 'confinement' is specified as both a regular key and in pass-through. Remove one of the keys.".
 - Fix the error by removing "confinement: classic" from the top.
 - snapcraft
 - Observe that that the build succeeds with two warnings:
   Pass-through is being used to add experimental properties to 'foo' that have not been validated. The snap cannot be released to the store.
   Pass-through is being used to add experimental properties to 'snapcraft.yaml' that have not been validated. The snap cannot be released to the store.
 - grep -C 3 foo prime/meta/snap.yaml
   confinement: next-generation
   grade: stable
   apps:
     foo:
       command: command-foo.wrapper
       daemon: complex
       foo: bar
   foo: bar
 - Observe keys being propagated to snap.yaml without validation.

### DIFF
--- a/snapcraft/internal/meta/_errors.py
+++ b/snapcraft/internal/meta/_errors.py
@@ -60,3 +60,15 @@ class AdoptedPartNotParsingInfo(SnapMetaGenerationError):
 
     def __init__(self, part: str) -> None:
         super().__init__(part=part)
+
+
+class AmbiguousPassThroughKeyError(SnapMetaGenerationError):
+
+    fmt = (
+        "Failed to pass-through keys to snap metadata: "
+        "{key!r} is specified as both a regular key and in pass-through. "
+        "Remove one of the keys."
+    )
+
+    def __init__(self, key: str) -> None:
+        super().__init__(key=key)
--- a/snapcraft/internal/meta/_errors.py
+++ b/snapcraft/internal/meta/_errors.py
@@ -16,6 +16,7 @@
 
 from snapcraft import formatting_utils
 from snapcraft.internal import errors
+from typing import Set
 
 
 class CommandError(errors.SnapcraftError):
@@ -66,9 +67,11 @@ class AmbiguousPassthroughKeyError(SnapMetaGenerationError):
 
     fmt = (
         "Failed to propagate keys to snap metadata: "
-        "{key!r} is specified as both a regular key and in pass-through. "
-        "Remove one of the keys."
+        "The following keys are specified as both regular keys and in "
+        "passthrough:\n"
+        '{keys}\n\n'
+        "Remove duplicate keys."
     )
 
-    def __init__(self, key: str) -> None:
-        super().__init__(key=key)
+    def __init__(self, keys: Set[str]) -> None:
+        super().__init__(keys=', '.join(keys))
--- a/snapcraft/internal/meta/_errors.py
+++ b/snapcraft/internal/meta/_errors.py
@@ -62,10 +62,10 @@ class AdoptedPartNotParsingInfo(SnapMetaGenerationError):
         super().__init__(part=part)
 
 
-class AmbiguousPassThroughKeyError(SnapMetaGenerationError):
+class AmbiguousPassthroughKeyError(SnapMetaGenerationError):
 
     fmt = (
-        "Failed to pass-through keys to snap metadata: "
+        "Failed to propagate keys to snap metadata: "
         "{key!r} is specified as both a regular key and in pass-through. "
         "Remove one of the keys."
     )
--- a/snapcraft/internal/meta/_errors.py
+++ b/snapcraft/internal/meta/_errors.py
@@ -16,7 +16,7 @@
 
 from snapcraft import formatting_utils
 from snapcraft.internal import errors
-from typing import Set
+from typing import List
 
 
 class CommandError(errors.SnapcraftError):
@@ -66,12 +66,11 @@ class AdoptedPartNotParsingInfo(SnapMetaGenerationError):
 class AmbiguousPassthroughKeyError(SnapMetaGenerationError):
 
     fmt = (
-        "Failed to propagate keys to snap metadata: "
-        "The following keys are specified as both regular keys and in "
-        "passthrough:\n"
-        '{keys}\n\n'
+        "Failed to generate snap metadata: "
+        "The following keys are specified in their regular location "
+        "as well as in passthrough: {keys}. "
         "Remove duplicate keys."
     )
 
-    def __init__(self, keys: Set[str]) -> None:
-        super().__init__(keys=', '.join(keys))
+    def __init__(self, keys: List[str]) -> None:
+        super().__init__(keys=formatting_utils.humanize_list(keys, 'and'))
--- a/snapcraft/internal/meta/_snap_packaging.py
+++ b/snapcraft/internal/meta/_snap_packaging.py
@@ -399,18 +399,27 @@ class _SnapPackaging:
             if key_name in self._config_data:
                 snap_yaml[key_name] = self._config_data[key_name]
 
+        passthrough_applied = False
+
         if 'apps' in self._config_data:
             _verify_app_paths(basedir='prime', apps=self._config_data['apps'])
             snap_yaml['apps'] = self._wrap_apps(self._config_data['apps'])
             self._render_socket_modes(snap_yaml['apps'])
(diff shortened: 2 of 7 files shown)


snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -513,14 +525,29 @@ def _generate_desktop_file(self, name, app):
...
+        # not be specified in pass-through at the same time.
+        for key in pass_through:
+            if key in original:
+                raise meta_errors.AmbiguousPassThroughKeyError(key)
+            this[key] = pass_through[key]
+        if pass_through:
+            logger.warn('Pass-through is being used to add experimental '
+                        'properties to {!r} that have not been '
+                        'validated. The snap cannot be released to the store.'
+                        .format(location))
```
kyrofa: This prints twice if one uses this feature in both the root and in apps. I think they'll probably know where they're using it, so I suggest only printing once if we see this keyword (no matter how many times we see it) and not specifying where it was.

snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -513,14 +525,29 @@ def _generate_desktop_file(self, name, app):
...
+    def _render_socket_modes(self, apps: Dict[str, Any]) -> None:
         for app in apps.values():
             sockets = app.get('sockets', {})
             for socket in sockets.values():
                 mode = socket.get('socket-mode')
                 if mode is not None:
                     socket['socket-mode'] = OctInt(mode)
-        return apps
+
+    def _apply_pass_through(self, location: str, this: Dict[str, Any],
```
kyrofa: `this` is not the most descriptive name. How about `section`?

snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -513,14 +525,29 @@ def _generate_desktop_file(self, name, app):
...
+                            pass_through: Dict[str, Any],
+                            original: Dict[str, Any]) -> None:
+        # Look for properties defined in pass-through and add them to the
+        # given dictionary. Any value already in the original dictionary must
+        # not be specified in pass-through at the same time.
+        for key in pass_through:
+            if key in original:
+                raise meta_errors.AmbiguousPassThroughKeyError(key)
```
kyrofa: If one has multiple keys overlapping, this error will result in a bit of a whack-a-mole experience. May I suggest instead turning each collection of keys into a set and intersecting them, and raising an error with _all_ conflicting keys?

snapcraft/internal/meta/_snap_packaging.py
kalikiana: The point of the message is to very clearly point out that the feature is being used, right? Whether or not they remember using it, they should be reminded that this is something that will need to be removed in the future. So I think we don't want it to be less obtrusive.

kyrofa: I think the real goal is to let them know that, because this keyword is being used, the store will flag it. Why do we need to say that multiple times?

snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -513,14 +525,28 @@ def _generate_desktop_file(self, name, app):
...
+                           passthrough: Dict[str, Any],
+                           original: Dict[str, Any]) -> None:
+        # Any value already in the original dictionary must
+        # not be specified in passthrough at the same time.
+        duplicates = set(original.keys() & passthrough.keys())
+        if duplicates:
+            raise meta_errors.AmbiguousPassthroughKeyError(duplicates)
+        section.update(passthrough)
+        if passthrough:
+            logger.warn('Passthrough is being used to add experimental '
```
kyrofa: Our thread is considered "outdated" by github, so I'm gonna copy/paste and start a new one :stuck_out_tongue: .

I think the real goal is to let them know that, because this keyword is being used, the store will flag it. Why do we need to say that multiple times? If you want to say where it was used, that's fine, but let's not print this more than once either way.

snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -513,14 +534,24 @@ def _generate_desktop_file(self, name, app):
...
                 if mode is not None:
                     socket['socket-mode'] = OctInt(mode)
-        return apps
+
+    def _apply_passthrough(self, location: str, section: Dict[str, Any],
+                           passthrough: Dict[str, Any],
+                           original: Dict[str, Any]) -> bool:
+        # Any value already in the original dictionary must
+        # not be specified in passthrough at the same time.
+        duplicates = set(original.keys() & passthrough.keys())
```
kyrofa: The result of this operation is already a set, no need for the cast.

snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -395,10 +399,27 @@ def _compose_snap_yaml(self):
...
+                        self._original_snapcraft_yaml['apps'][app_name]):
+                    passthrough_applied = True
+
+        if self._apply_passthrough('snapcraft.yaml', snap_yaml,
+                                   self._config_data.get('passthrough', {}),
+                                   self._original_snapcraft_yaml):
+            passthrough_applied = True
+
+        if passthrough_applied:
+            logger.warn('Passthrough is being used to propagate experimental '
```
kyrofa: Instead of Passthrough, I suggest using: `The 'passthrough' property is being used[...]`. That format matches up with some of our other warnings.

snapcraft/internal/meta/_errors.py
```diff
@@ -60,3 +61,17 @@ class AdoptedPartNotParsingInfo(SnapMetaGenerationError):
...
+
+
+class AmbiguousPassthroughKeyError(SnapMetaGenerationError):
+
+    fmt = (
+        "Failed to propagate keys to snap metadata: "
+        "The following keys are specified as both regular keys and in "
+        "passthrough:\n"
+        '{keys}\n\n'
+        "Remove duplicate keys."
```
kyrofa: This formatting isn't quite right:

"Failed to propagate keys to snap metadata: The following keys are specified as both regular keys and in passthrough:
grade, confinement

Remove duplicate keys."

How about using `formatting_utils.humanize_list`? Something like

```
fmt = (
    "Failed to generate snap metadata: "
    "The following keys are specified in their regular location as well as in passthrough: {}. "
    "Remove duplicate keys."
)

def __init__(self, keys: Set[str]) -> None:
        super().__init__(keys=formatting_utils.humanize_list(keys, 'and')
```

kyrofa:
Got some test failures, here.

niemeyer:
We should probably follow @kyrofa's advice regarding hooks. The structure is quite analogous to apps, and people will probably expect it to behave the same unconsciously.

Plus and slots don't seem like a problem, though, because their attributes are already open ended.

PR #2055, opened 2018-04-06
python: bring back support for older versions of pip

kalikiana:

-----
This PR brings back support for older versions of pip, most notably 8.1.1 which is in Xenial, that don't support `pip list --format=json`. Ease of parsing is why we dropped that support previously. And we support pulling in pip from stage-packages so that should include Xenial and not require resorting to PyPI, which is what Snapcraft does with pip, setuptools and wheel when they're not available.

Note that no effort is made to hide the error message as per what we do in other places.

The following new test is being added:

 - tests.unit.plugins.python.test_pip.PipListTestCase.test_no_json_format
 - To verify that we fallback to the legacy format if pip returns an error and list the packages from that successfully.

I locally ran the tests:

Manual test steps:
 - cd tests/integration/snaps/python-hello
 - Update `snapcraft.yaml` by adding this snippet to "python-part":
   stage-packages:
    - python3-pip
    - python3-wheel
    - python3-setuptools
 - snapcraft
 - Observe that the build proceeds without an error and notice the error message "no such option: --format=json" being logged.

### DIFF
--- a/snapcraft/plugins/_python/_pip.py
+++ b/snapcraft/plugins/_python/_pip.py
@@ -378,26 +378,27 @@ class Pip:
             output = self._run_output(command)
             json_output = json.loads(
                     output, object_pairs_hook=collections.OrderedDict)
-            for package in json_output:
-                if 'name' not in package:
-                    raise errors.PipListMissingFieldError('name', output)
-                if 'version' not in package:
-                    raise errors.PipListMissingFieldError('version', output)
-                packages[package['name']] = package['version']
-        except json.decoder.JSONDecodeError as e:
-            raise errors.PipListInvalidJsonError(output) from e
         except subprocess.CalledProcessError:
+            # --format requires a newer pip, so fall back to legacy output
             command = ['list']
             if user:
                 command.append('--user')
             output = self._run_output(command)
-            # --format requires a newer pip, so fall back to legacy output
+            json_output = []  # type: List[Dict[str, str]]
             version_regex = re.compile('\((.+)\)')
             for line in output.splitlines():
                 line = line.split()
                 m = version_regex.search(line[1])
-                packages[line[0]] = m.group(1)
+                json_output.append({'name': line[0], 'version': m.group(1)})
+        except json.decoder.JSONDecodeError as e:
+            raise errors.PipListInvalidJsonError(output) from e
 
+        for package in json_output:
+            if 'name' not in package:
+                raise errors.PipListMissingFieldError('name', output)
+            if 'version' not in package:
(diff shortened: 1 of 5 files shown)


snapcraft/plugins/_python/_pip.py
```diff
@@ -372,20 +373,31 @@ def list(self, *, user=False):
...
+            command = ['list']
+            if user:
+                command.append('--user')
+            output = self._run_output(command)
+            # --format requires a newer pip, so fall back to legacy output
+            version_regex = re.compile('\((.+)\)')
+            for line in output.splitlines():
+                line = line.split()
+                m = version_regex.search(line[1])
+                packages[line[0]] = m.group(1)
```
sergiusens: can you recontruct `json_output` here instead and leave the logic below untouched?

snapcraft/plugins/_python/_pip.py
```diff
@@ -372,11 +373,23 @@ def list(self, *, user=False):
...
         packages = collections.OrderedDict()
         try:
+            output = self._run_output(command)
             json_output = json.loads(
                     output, object_pairs_hook=collections.OrderedDict)
+        except subprocess.CalledProcessError:
+            # --format requires a newer pip, so fall back to legacy output
+            command = ['list']
+            if user:
+                command.append('--user')
```
kyrofa: We're asking for trouble by duplicating this logic. Perhaps the [base command](https://github.com/snapcore/snapcraft/pull/2055/files#diff-c4fa5148d03d9fab1aa1f2284d873aa6R372) should be `[list]` with `--format=json` thrown on at the last minute in the `try`, so we can just re-use the original `command` (and its `--user` logic).

tests/unit/plugins/python/test_pip.py
```diff
@@ -736,6 +736,18 @@ def test_invalid_json(self):
         self.assertThat(
             str(raised), Contains("Pip packages output isn't valid json"))
 
+    def test_no_json_format(self):
+        self.mock_run.side_effect = [
+            subprocess.CalledProcessError(2, 'pip', 'no such option'),
+            'foo (1.0)',
+        ]
+        self.mock_run.return_value = '[{"name": "foo", "version": "1.0"}]'
```
kyrofa: This seems to serve no purpose given the two side_effects. Am I missing something?

kalikiana: Indeed it seems I missed a left-over. Thanks for catching it.

snapcraft/plugins/_python/_pip.py
kalikiana: Done.

PR #2056, opened 2018-04-06
Fix formatting of some store errors

cjwatson:
`SnapcraftError.fmt` is run through `str.format` by
`SnapcraftError.__str__`, so it shouldn't itself be the output of a
`str.format` operation.  If it is, and the output contains `{}` (as for
example if it contains raw JSON output), then `__str__` will raise an
exception.

LP: [#1761488](https://bugs.launchpad.net/bugs/1761488)


kyrofa:
Thanks @cjwatson! I've merged #2050 if you don't mind rebasing. Really appreciate the help here :smiley: .

cjwatson:
OK, I've rebased and added a few more tests (the remaining bugs were in `StoreReviewError` and `StoreMetadataError`).

Issue #2028, opened 2018-03-26
When asking to release to a branch that's too long, a traceback is printed that gives no hints as to the source of the error


sergiusens:
- [When asking to release to a branch that's too long, a traceback is printed that gives no hints as to the source of the error](https://bugs.launchpad.net/snapcraft/+bug/1750177)

kyrofa:
Unfortunately, the store response is a 500 in this case, containing HTML and no clue as to what happened. We can stop tracing back, but the best we can do is say "Uh oh. Sorry." Check #2059 for details.

PR #2059, opened 2018-04-09
storeapi: handle 500 error response when releasing snap

kyrofa:

-----

This PR fixes LP: [#1750177](https://bugs.launchpad.net/snapstore/+bug/1750177) and #2028 by properly handling the store returning a 500 error formatted as HTML instead of a typical error response. It also starts printing store error responses when in debug mode in order to streamline such fixes in the future.

### DIFF
--- a/snapcraft/storeapi/errors.py
+++ b/snapcraft/storeapi/errors.py
@@ -15,11 +15,15 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import contextlib
+import logging
 from simplejson.scanner import JSONDecodeError
 from typing import List  # noqa
 
 from snapcraft.internal.errors import SnapcraftError
 from snapcraft import formatting_utils
+
+
+logger = logging.getLogger(__name__)
 
 
 class StoreError(SnapcraftError):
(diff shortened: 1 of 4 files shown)


snapcraft/storeapi/errors.py
```diff
@@ -396,6 +410,16 @@ def __fmt_error_401_or_403(self, response):
     def __fmt_error_404(self, response):
         return self.__FMT_NOT_REGISTERED
 
+    def __fmt_error_500(self, response):
+        response_json = self.__to_json(response)
+        message = ('The store encountered an internal error. Please report a '
+                   'bug if this continues.')
```
sergiusens: where?

sergiusens: also reporting a bug without much information or the fact that it is repeatable would be a really hard to fix thing. This should probably be something done with sentry.
I would state that if the error persists, to log a bug or reach out for support, but not on the first try (we could also tell them to look at the status page)

kyrofa: Alright, done.

kyrofa:
> Also, if we are passing in response, I would remove the need to pass thing such as response.status_code.

As far as I can tell, none of the errors require this (although some format strings include them). Can you please clarify?

PR #2061, opened 2018-04-10
go: only use Go build package if not using the snap

kalikiana:

-----
This PR makes the Go plugin prefer a Go build snap over Go installed from the archive. A staged Go has preference over both like before. Build packages will still be installed as package installation happens after the plugin initialization, so the right command will be picked afterwards.
A debug message will inform which Go was picked up.

Fixes: [LP: #1616985](https://bugs.launchpad.net/snapcraft/+bug/1616985)

The following new test is being added:

 - tests.unit.plugins.go.test_go.GoPluginCommandTest
 - To verify that a snap/ build package/ stage go is used respectively, by mocking the existence of the respective files, and that the debug message points at the right path. Also a test to verify that stage packages are preferred.
 - Existing tests get mocked to not look for the actual Go that happens to be installed during the test run.

I locally ran the tests:

Manual test steps:
 - cd tests/integration/snaps/go-hello
 - Update `snapcraft.yaml` by adding this snippet to "go-hello":
   build-snaps:
    - go
 - snapcraft -d
 - Observe that the build proceeds and a message is logged saying that `/snap/bin/go` is used.
 - Undo the YAML changes by removing build-snaps and `sudo snap remove go`.
 - snapcraft -d
 - Observe that the build succeeds and a message is logged saying "go" is used.

### DIFF
--- a/snapcraft/plugins/go.py
+++ b/snapcraft/plugins/go.py
@@ -101,11 +101,29 @@ class GoPlugin(snapcraft.BasePlugin):
 
     def __init__(self, name, options, project):
         super().__init__(name, options, project)
+        self._go_command = None  # type: str
+        # The build package needs to be installed here. We will only decide
+        # which Go to use during pull however after dependencies have been
+        # staged and build snaps were installed.
         self.build_packages.append('golang-go')
         self._gopath = os.path.join(self.partdir, 'go')
         self._gopath_src = os.path.join(self._gopath, 'src')
         self._gopath_bin = os.path.join(self._gopath, 'bin')
         self._gopath_pkg = os.path.join(self._gopath, 'pkg')
+
+    def _get_go_command(self) -> str:
+        if not self._go_command:
+            # Look for Go in stage or a build snap
+            for path in [self.project.stage_dir, '/snap']:
+                command = os.path.join(path, 'bin', 'go')
+                if os.path.exists(command):
+                    self._go_command = command
+                    break
+            # Fallback to the command installed via build packages
+            if not self._go_command:
+                self._go_command = 'go'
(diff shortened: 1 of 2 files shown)


snapcraft/plugins/go.py
```diff
@@ -101,12 +101,30 @@ def get_pull_properties(cls):
 
     def __init__(self, name, options, project):
         super().__init__(name, options, project)
+        self._go_command = None  # type: str
+        # The build package needs to be installed here. We will only decide
+        # which Go to use during pull however after dependencies have been
+        # staged and build snaps were installed.
         self.build_packages.append('golang-go')
```
sergiusens: I was hoping this would take another approach. Similar to what we do in the `python` plugin in `stage_packages` as a property but with `build_packages` here. As in, if we see a `build_snap` entry we do not add `golang-go` to `build_packages` (or by finding it in `stage` using what @kyrofa wrote to find common commands).

snapcraft/plugins/go.py
```diff
@@ -101,12 +101,30 @@ def get_pull_properties(cls):
...
         self._gopath = os.path.join(self.partdir, 'go')
         self._gopath_src = os.path.join(self._gopath, 'src')
         self._gopath_bin = os.path.join(self._gopath, 'bin')
         self._gopath_pkg = os.path.join(self._gopath, 'pkg')
 
+    def _get_go_command(self) -> str:
+        if not self._go_command:
```
sergiusens: can we improve this logic to be a bit more functional please?

kalikiana:
As discussed offline, this needs more design so I'm closing this for now.

PR #2063, opened 2018-04-10
many: add snapcraftctl set-version

### LINKED ISSUE #1676: Support for set-version

kyrofa:

-----

This PR resolves #1676 by adding the ability to set the snap version from a scriptlet using `snapcraftctl set-version <version>`. It also lays down generic infrastructure to make adding future such "setters" easy.

### DIFF
--- a/snapcraft/internal/meta/_snap_packaging.py
+++ b/snapcraft/internal/meta/_snap_packaging.py
@@ -25,7 +25,7 @@ import shlex
 import shutil
 import stat
 import subprocess
-from typing import Any, Dict, List  # noqa
+from typing import Any, Dict, List, Set  # noqa
 
 import yaml
 
@@ -165,6 +165,21 @@ def _update_yaml_with_extracted_metadata(
 def _adopt_info(
         config_data: Dict[str, Any],
         extracted_metadata: _metadata.ExtractedMetadata):
+    ignored_keys = _adopt_keys(config_data, extracted_metadata)
+    if ignored_keys:
+        logger.warning(
+            'The {keys} {plural_property} {plural_is} specified in adopted '
+            'info as well as the YAML: taking the {plural_property} from the '
+            'YAML'.format(
+                keys=formatting_utils.humanize_list(list(ignored_keys), 'and'),
+                plural_property=formatting_utils.pluralize(
+                    ignored_keys, 'property', 'properties'),
+                plural_is=formatting_utils.pluralize(
+                    ignored_keys, 'is', 'are')))
+
+
+def _adopt_keys(config_data: Dict[str, Any],
+                extracted_metadata: _metadata.ExtractedMetadata) -> Set[str]:
     ignored_keys = set()
     metadata_dict = extracted_metadata.to_dict()
     for key, value in metadata_dict.items():
@@ -193,16 +208,7 @@ def _adopt_info(
                         desktop_file_path)
                     break
 
-    if ignored_keys:
-        logger.warning(
-            'The {keys} {plural_property} {plural_is} specified in adopted '
-            'info as well as the YAML: taking the {plural_property} from the '
-            'YAML'.format(
-                keys=formatting_utils.humanize_list(ignored_keys, 'and'),
-                plural_property=formatting_utils.pluralize(
-                    ignored_keys, 'property', 'properties'),
-                plural_is=formatting_utils.pluralize(
-                    ignored_keys, 'is', 'are')))
+    return ignored_keys
 
 
 def _icon_file_exists() -> bool:
--- a/snapcraft/internal/meta/_snap_packaging.py
+++ b/snapcraft/internal/meta/_snap_packaging.py
@@ -120,17 +120,35 @@ def _update_yaml_with_extracted_metadata(
         if not part:
             raise meta_errors.AdoptedPartMissingError(part_name)
 
-        # This would be caught since metadata would be missing, but we want
-        # to be clear about the issue here. This really should be caught by the
-        # schema, but it doesn't seem to support such dynamic behavior.
-        if 'parse-info' not in config_data['parts'][part_name]:
-            raise meta_errors.AdoptedPartNotParsingInfo(part_name)
+        pull_state = part.get_pull_state()
+        build_state = part.get_build_state()
+        stage_state = part.get_stage_state()
+        prime_state = part.get_prime_state()
 
-        # Get the metadata from the pull step first, then update it using the
-        # metadata from the build step (i.e. the data from the build step takes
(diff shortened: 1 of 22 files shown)


tests/unit/pluginhandler/test_pluginhandler.py
```diff
@@ -824,6 +829,100 @@ def test_prime(self):
         self.assertThat(self.handler.next_step(), Equals(None))
 
 
+class ScriptletSetVersionTestCase(unit.TestCase):
```
sergiusens: you could just create `tests/unit/pluginhandler/test_scriptlets.py` and put this in there, not a blocker though

tests/unit/pluginhandler/test_pluginhandler.py
```diff
@@ -824,6 +829,100 @@ def test_prime(self):
...
+        with contextlib.suppress(AttributeError):
+            part_properties['override-pull'] = self.override_pull
+        with contextlib.suppress(AttributeError):
+            part_properties['override-build'] = self.override_build
+        with contextlib.suppress(AttributeError):
+            part_properties['override-stage'] = self.override_stage
+        with contextlib.suppress(AttributeError):
+            part_properties['override-prime'] = self.override_prime
+
+        # A few of these test cases result in only one of these scriptlets
```
sergiusens: This is why I don't like scenarios as much any more :smile:

snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -120,17 +120,34 @@ def _update_yaml_with_extracted_metadata(
...
-        # schema, but it doesn't seem to support such dynamic behavior.
-        if 'parse-info' not in config_data['parts'][part_name]:
-            raise meta_errors.AdoptedPartNotParsingInfo(part_name)
-
-        # Get the metadata from the pull step first, then update it using the
-        # metadata from the build step (i.e. the data from the build step takes
-        # precedence over the pull step)
-        metadata = part.get_pull_state().extracted_metadata['metadata']
-        metadata.update(part.get_build_state().extracted_metadata['metadata'])
+        pull_state = part.get_pull_state()
```
sergiusens: this should probably be improved upon in the future with a `part.get_state()` call

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -186,6 +224,16 @@ def last_step(self):
 
         return None
 
+    def next_step(self):
```
sergiusens: this should probably be queried by the lifecycle runner eventually

tests/unit/pluginhandler/test_pluginhandler.py
kyrofa: I know, I really feel like there should be a good way to do that :thinking: .

snapcraft/internal/meta/_snap_packaging.py
kyrofa: I'm not sure I understand, what would `part.get_state()` do?

snapcraft/internal/pluginhandler/__init__.py
kyrofa: Ah, yeah that's a good idea, let's keep that in mind for a future refactor.

tests/integration/general/test_snapcraftctl_set_version.py
```diff
@@ -0,0 +1,76 @@
...
+                  plugin: nil
+                  override-pull: snapcraftctl set-version override-version
+                """))
+
+        self.run_snapcraft('prime')
+
+        with open(os.path.join('prime', 'meta', 'snap.yaml')) as f:
+            y = yaml.load(f)
+
+        self.assertThat(y['version'], Equals('test-version'))
```
sergiusens: this should actually be the opposite

tests/unit/pluginhandler/test_pluginhandler.py
```diff
@@ -781,6 +786,143 @@ def test_makedirs_with_existing_dirs(self):
...
+        self.assertThat(self.handler.last_step(), Equals('stage'))
+        self.assertThat(self.handler.next_step(), Equals('prime'))
+
+    def test_prime(self):
+        self.assertThat(self.handler.last_step(), Equals(None))
+        self.assertThat(self.handler.next_step(), Equals('pull'))
+
+        self.handler.prime()
+
+        self.assertThat(self.handler.last_step(), Equals('prime'))
```
sergiusens: seeing these together I think we should call this `previous_step`

tests/integration/general/test_snapcraftctl_set_version.py
kyrofa: This is testing the case when a version is specified in the snapcraft.yaml, which takes ultimate precedence unless it's missing. The metadata extractors work that way, so I duplicated that for scriptlets. Otherwise it's possible to be using a part that gives you e.g. an icon, but a version you don't want.

tests/unit/pluginhandler/test_pluginhandler.py
kyrofa: Yeah fine by me, although this is pretty old and used in a number of places. Maybe a followup?

tests/integration/general/test_snapcraftctl_set_version.py
sergiusens: this is the other way around to what we already have with `version-script` where you can do

```bash
if [ $SNAPCRAFT_PROJECT_GRADE = 'stable' ]; then
    echo $SNAPCRAFT_PROJECT_VERSION
else
    echo $SNAPCRAFT_PROJECT_VERSION-unstable
fi
```

sergiusens: If you feel strongly about this, we should warn about all the properties we are ignoring because they are already set in YAML.

tests/unit/pluginhandler/test_pluginhandler.py
kyrofa: Good call, done.

snapcraft/internal/meta/_snap_packaging.py
sergiusens: Bring in a `state` object for the part for which we could do `state.pull_state.extracted_metadata['metadata']`

sergiusens: Or maybe encapsulate the knowledge of that dict too

tests/unit/pluginhandler/test_pluginhandler.py
sergiusens: follow up is fine, it is all internal

PR #2083, opened 2018-04-18
python plugin: properly handle distutils on bionic

sergiusens:
distutils has been split out into its own package for python3 but
not for python2.

-----

### DIFF
--- a/snapcraft/plugins/python.py
+++ b/snapcraft/plugins/python.py
@@ -159,9 +159,9 @@ class PythonPlugin(snapcraft.BasePlugin):
             return
 
         stage_packages = [python_base]
-        if release_codename == 'bionic':
-            # In bionic, pip started requiring python-distutils to be
-            # installed.
+        # In bionic, python3's pip started requiring python-distutils
+        # to be installed.
+        if python_base == 'python3' and release_codename == 'bionic':
             stage_packages.append('{}-distutils'.format(python_base))
         return stage_packages
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,8 @@ jobs:
     - if: type != cron
       script: sudo ./tools/travis/run_tests.sh tests.integration.plugins_python
     - if: type != cron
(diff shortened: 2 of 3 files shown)


sergiusens:
If you run `rmadison <package>` you will notice why it needs removing. Unfortunate this slipped through (only hit by a full bionic test run).

snapcraft/plugins/python.py
```diff
@@ -159,9 +159,9 @@ def plugin_stage_packages(self):
             return
 
         stage_packages = [python_base]
-        if release_codename == 'bionic':
-            # In bionic, pip started requiring python-distutils to be
-            # installed.
+        # In bionic, python3's pip started requiring python-distutils
+        # to be installed.
+        if python_base == 'python3' and release_codename == 'bionic':
```
kyrofa: I dunno, I'm kinda worried they'll start doing this for python2 now, too. I see no downside to leaving this the way it was, do you?

sergiusens: Which is why I made my comment about checking rmadison :-)
`python-distutils` does not exist.

sergiusens:
@kyrofa the package only exists on bionic, so we cannot.

sergiusens:
as in, it will go into debian/control when we release to bionic, it needs to be in the plugin for stage-packages to work when running from the snap.

snapcraft/plugins/python.py
```diff
@@ -159,9 +159,9 @@ def plugin_stage_packages(self):
             return
 
         stage_packages = [python_base]
-        if release_codename == 'bionic':
-            # In bionic, pip started requiring python-distutils to be
-            # installed.
+        # In bionic, python3's pip started requiring python-distutils
```
bjornt: To be correct, it should be python3-distutils

sergiusens: ah, indeed.

sergiusens:
thanks for the review @bjornt

PR #2087, opened 2018-04-18
ci: setup AppVeyor

kalikiana:

-----
This PR adds a configuration file for CI in AppVeyor. To support integration tests on non-unix platforms I'm splitting out the fixtures that require unix sockets and moving pexpect away from unix-specifc API.

Requires: #2096

No tests were added or removed. However the store login method used for testing uses `SNAPCRAFT_TEST_INPUT` now to force regular input for passwords to avoid depending on a `tty` which is platform-specific.

I locally ran the tests:


sergiusens:
to make the change easier just create the package and add all the classes in `__init__.py` using the logic described for `FakeSnapd` or make use of `from _fixtures import *`

tests/fixture_setup/__init__.py
```diff
@@ -0,0 +1,45 @@
+import socket
+
+if hasattr(socket, "AF_UNIX"):
+    from ._unix import FakeSnapd  # noqa
+from ._fixtures import (  # noqa
```
sergiusens: add the specific code for the ignored error

tests/fixture_setup/__init__.py
```diff
@@ -0,0 +1,45 @@
+import socket
+
+if hasattr(socket, "AF_UNIX"):
+    from ._unix import FakeSnapd  # noqa
```
sergiusens: add the specific code for the ignored error

tests/integration/__init__.py
```diff
@@ -39,6 +41,7 @@
 from testtools.matchers import MatchesRegex
 
 from snapcraft.internal.os_release import OsRelease
+from snapcraft.internal import errors
```
sergiusens: this is wrong, we should not import snapcraft into the integration tests.

kalikiana: I fixed this in #2096 and re-based this one on top of it.

appveyor.yml
```diff
@@ -0,0 +1,32 @@
...
+build: off
+before_test:
+- cmd: |
+    %PYTHON%\\python.exe -m venv venv
+    venv\\Scripts\\activate.bat
+    ECHO %VIRTUAL_ENV%
+    python -c "import sys; print(sys.executable)"
+    python -m pip install -r requirements.txt
+    python -m pip install -r requirements-devel.txt
+    python setup.py install
```
sergiusens: this might be better as `python -m pip install .` or does that not work on windows?

kalikiana: It didn't seem to work when I first tried it. Apparently with `--prefix` it does. Unfortunately we're not getting an `.exe` this way so I need to execute Snapcraft in the tests via `python -m` now.

PR #2089, opened 2018-04-18
many: remove support for remote lxd per project containers

sergiusens:
Remove the support as it has not proved to be the best experience, cleanbuild should be used
instead.

Also future feature proof the environment detection.

-----

### DIFF
--- a/snapcraft/cli/_command_group.py
+++ b/snapcraft/cli/_command_group.py
@@ -62,6 +62,6 @@ class SnapcraftGroup(click.Group):
         # from the store.
         commands.pop(commands.index('edit-collaborators'))
         build_environment = env.BuilderEnvironmentConfig()
-        if build_environment.provider == 'host':
+        if build_environment.is_host:
             commands.pop(commands.index('refresh'))
         return commands
--- a/snapcraft/cli/_command_group.py
+++ b/snapcraft/cli/_command_group.py
@@ -61,7 +61,7 @@ class SnapcraftGroup(click.Group):
         # Let's keep edit-collaborators hidden until we get the green light
         # from the store.
         commands.pop(commands.index('edit-collaborators'))
-        container_config = env.get_container_config()
-        if not container_config.use_container:
+        build_environment = env.BuilderEnvironmentConfig()
+        if build_environment.provider == 'host':
             commands.pop(commands.index('refresh'))
         return commands
--- a/snapcraft/cli/env.py
+++ b/snapcraft/cli/env.py
@@ -34,8 +34,8 @@ class BuilderEnvironmentConfig:
                 use_lxd = util.strtobool(container_builds)
             except ValueError:
                 raise errors.SnapcraftEnvironmentError(
-                    'The experimental feature of using lxd remotes other '
-                    'that cleanbuild has been dropped.')
+                    'The experimental feature of using non-local LXD remotes '
+                    'with SNAPCRAFT_CONTAINER_BUILDS has been dropped.')
 
         build_provider = os.environ.get('SNAPCRAFT_BUILD_ENVIRONMENT')
         if use_lxd:
--- a/snapcraft/cli/env.py
+++ b/snapcraft/cli/env.py
@@ -44,7 +44,7 @@ class BuilderEnvironmentConfig:
             echo.warning('Using the host as the build environment.')
             build_provider = 'host'
         # TODO add multipass
-        elif build_provider not in ['host', 'multipass', 'lxd']:
+        elif build_provider not in ['host', 'lxd']:
             raise errors.SnapcraftEnvironmentError(
                 'SNAPCRAFT_BUILD_ENVIRONMENT must be one of: host or lxd.')
 
--- a/snapcraft/cli/env.py
+++ b/snapcraft/cli/env.py
@@ -15,41 +15,38 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 import os
 from distutils import util
-from snapcraft.internal import errors, lxd
+
+from . import echo
+from snapcraft.internal import errors
 
 
-class ContainerConfig:
+class BuilderEnvironmentConfig:
+    """Handle the chosen build provider."""
 
-    def __init__(self):
-        """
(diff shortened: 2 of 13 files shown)


snapcraft/cli/env.py
```diff
@@ -15,41 +15,38 @@
...
+        if container_builds:
+            echo.warning(
+                'The flag SNAPCRAFT_CONTAINER_BUILDS has been deprecated. '
+                'Use SNAPCRAFT_BUILD_ENVIRONMENT=lxd instead.')
+            try:
+                use_lxd = util.strtobool(container_builds)
+            except ValueError:
+                raise errors.SnapcraftEnvironmentError(
+                    'The experimental feature of using lxd remotes other '
+                    'that cleanbuild has been dropped.')
```
kalikiana: That phrasing makes little sense to me. Cleanbuild is a command. It supports LXD remotes via `--remote=`. What is actually being dropped here is support with non-local LXD remotes for persistent containers.
How about `The experimental feature of using non-local LXD remotes with SNAPCRAFT_CONTAINER_BUILDS has been dropped.`?

snapcraft/cli/env.py
```diff
@@ -15,41 +15,38 @@
...
+
+        build_provider = os.environ.get('SNAPCRAFT_BUILD_ENVIRONMENT')
+        if use_lxd:
```
kalikiana: Shouldn't `SNAPCRAFT_BUILD_ENVIRONMENT` have precedence here?

snapcraft/cli/env.py
```diff
@@ -15,41 +15,38 @@
...
+            build_provider = 'lxd'
+        elif not build_provider:
+            echo.warning('Using the host as the build environment.')
+            build_provider = 'host'
+        # TODO add multipass
+        elif build_provider not in ['host', 'multipass', 'lxd']:
```
kalikiana: This seems wrong. You're checking for a value that's not implemented.

tests/unit/test_lxd.py
```diff
@@ -536,216 +530,7 @@ def call_effect(*args, **kwargs):
...
-
-    def make_containerbuild(self):
-        return lxd.Project(output='snap.snap', source='project.tar',
-                           metadata={'name': 'project'},
-                           project_options=self.project_options,
-                           remote=self.remote)
-
-    @patch('snapcraft.internal.lxd.Containerbuild._container_run')
-    @patch('snapcraft.internal.common.is_snap')
-    def test_inject_snap_existing_container(self,
```
kalikiana: Why are you dropping this test?

tests/unit/test_lxd.py
```diff
@@ -536,216 +530,7 @@ def call_effect(*args, **kwargs):
...
-                call(['lxc', 'exec', self.fake_lxd.name, '--',
-                      'tee', '-a', '/etc/sudoers'],
-                     input='{} ALL=(ALL) NOPASSWD: ALL\n'.format(
-                         self.user).encode()),
-                call(['lxc', 'exec', self.fake_lxd.name, '--',
-                      'sudo', '-H', '-u', 'me', 'ls',
-                      '{}/build_project'.format(self.home)]),
-            ])
-
-    def test_init_failed(self):
```
kalikiana: Why are you dropping this test?

snapcraft/cli/env.py
```diff
@@ -15,41 +15,38 @@
...
-
-
-def get_container_config():
-    return ContainerConfig()
+from . import echo
+from snapcraft.internal import errors
+
+
+class BuilderEnvironmentConfig:
+    """Handle the chosen build provider."""
```
kalikiana: Can we keep the docs of how the variables work?

snapcraft/cli/env.py
sergiusens: that is debatable, but I will warn that both are in use and which one will take precedence.

snapcraft/cli/env.py
sergiusens: oh, I forgot to delete that part when I added the `TODO` part

tests/unit/test_lxd.py
sergiusens: because the `scenarios` all used remote remotes

snapcraft/cli/_command_group.py
```diff
@@ -61,7 +61,7 @@ def list_commands(self, ctx):
         # Let's keep edit-collaborators hidden until we get the green light
         # from the store.
         commands.pop(commands.index('edit-collaborators'))
-        container_config = env.get_container_config()
-        if not container_config.use_container:
+        build_environment = env.BuilderEnvironmentConfig()
+        if build_environment.provider == 'host':
```
kyrofa: Did you mean to use `build_environment.is_host` here?

snapcraft/cli/env.py
```diff
@@ -15,41 +15,60 @@
...
+                    'with SNAPCRAFT_CONTAINER_BUILDS has been dropped.')
 
-def get_container_config():
-    return ContainerConfig()
+        build_provider = os.environ.get('SNAPCRAFT_BUILD_ENVIRONMENT')
+        if build_provider and use_lxd:
+            raise errors.SnapcraftEnvironmentError(
+                'SNAPCRAFT_BUILD_ENVIRONMENT and SNAPCRAFT_CONTAINER_BUILDS '
+                'cannot be used together.\n'
+                'Unset one of them from the environment and try again.')
```
kyrofa: SNAPCRAFT_CONTAINER_BUILDS is deprecated though, perhaps we should point them toward SNAPCRAFT_BUILD_ENVIRONMENT here instead of saying "just pick one." Otherwise they may just be smacked with a deprecation notice if they pick the wrong one :stuck_out_tongue: .

sergiusens: I thought about this. The flow is nice though. If I use both I get:
- the deprecation message
- the error about using both

and this only shows up if using both, so there is no picking the wrong one.

snapcraft/cli/_command_group.py
sergiusens: yeah, bummer this retriggers the full suite again :-/

PR #2091, opened 2018-04-19
meta: soften warning about using passthrough

kyrofa:

-----

Currently Snapcraft warns that a snap using passthrough cannot be released to the store. What it means is that it may be flagged for manual review upon push, but even that isn't necessarily the case, as the store may gain support for the option being passed through. Also, some users are interpreting this warning as a fatal error, i.e. that Snapcraft won't allow release.

Stop mentioning the store in the passthrough warning.

### DIFF
--- a/snapcraft/internal/meta/_snap_packaging.py
+++ b/snapcraft/internal/meta/_snap_packaging.py
@@ -619,8 +619,7 @@ class _SnapPackaging:
         if passthrough_applied:
             logger.warn("The 'passthrough' property is being used to "
                         "propagate experimental properties to snap.yaml "
-                        "that have not been validated. The snap cannot be "
-                        "released to the store.")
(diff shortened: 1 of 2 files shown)


snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -619,8 +619,7 @@ def _process_passthrough_properties(
         if passthrough_applied:
             logger.warn("The 'passthrough' property is being used to "
                         "propagate experimental properties to snap.yaml "
-                        "that have not been validated. The snap cannot be "
-                        "released to the store.")
+                        "that have not been validated")
```
sparkiegeek: Needs a `.`.

kyrofa: Sure, done.

Issue #1705, opened 2017-11-06
Remove the current tests that assert exact error messages and only assert on the error class


sergiusens:
Estimate: 2 days

PR #2100, opened 2018-04-23
build_providers: new build provider using multipass

sergiusens:
build_providers is a new package to handle different providers snapcraft
can use to create a snap.

In addition to the scaffolding, an implementation to handle
multipass is provided. This is the MVP for multipass integration.

-----

### DIFF
--- a/snapcraft/cli/env.py
+++ b/snapcraft/cli/env.py
@@ -42,9 +42,12 @@ class BuilderEnvironmentConfig:
     results in an error.
     """
 
-    def __init__(self, *, additional_providers: List[str]=None) -> None:
+    def __init__(self, *, default='host',
+                 additional_providers: List[str]=None) -> None:
         """Instantiate a BuildEnvironmentConfig.
 
+        :param str default: the default provider to use among the list of valid
+                            ones.
         :param str additional_providers: Additional providers allowed in the
                                          environment.
         """
@@ -76,7 +79,7 @@ class BuilderEnvironmentConfig:
         if use_lxd:
             build_provider = 'lxd'
         elif not build_provider:
-            build_provider = 'host'
+            build_provider = default
         elif build_provider not in valid_providers:
             raise errors.SnapcraftEnvironmentError(
                 'SNAPCRAFT_BUILD_ENVIRONMENT must be one of: {}.'.format(
--- a/snapcraft/cli/env.py
+++ b/snapcraft/cli/env.py
@@ -15,9 +15,11 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 import os
 from distutils import util
+from typing import List
 
 from . import echo
 from snapcraft.internal import errors
+from snapcraft.formatting_utils import humanize_list
 
 
 class BuilderEnvironmentConfig:
@@ -31,6 +33,7 @@ class BuilderEnvironmentConfig:
 
     - host: the host will drive the build.
     - lxd: the host will setup a container to drive the build.
+    - multipass: a vm driven by multipass will be created to drive the build.
 
     Use of the lxd value is equivalent to setting the now deprecated
     SNAPCRAFT_CONTAINER_BUILDS environment variable to a value that
@@ -39,7 +42,16 @@ class BuilderEnvironmentConfig:
     results in an error.
     """
 
-    def __init__(self) -> None:
+    def __init__(self, *, additional_providers: List[str]=None) -> None:
+        """Instantiate a BuildEnvironmentConfig.
+
+        :param str additional_providers: Additional providers allowed in the
+                                         environment.
+        """
+        valid_providers = ['host', 'lxd']
+        if additional_providers is not None:
+            valid_providers.extend(additional_providers)
+
         use_lxd = None
         container_builds = os.environ.get('SNAPCRAFT_CONTAINER_BUILDS')
         if container_builds:
@@ -64,12 +76,12 @@ class BuilderEnvironmentConfig:
         if use_lxd:
             build_provider = 'lxd'
         elif not build_provider:
-            echo.warning('Using the host as the build environment.')
             build_provider = 'host'
-        # TODO add multipass
-        elif build_provider not in ['host', 'lxd']:
+        elif build_provider not in valid_providers:
             raise errors.SnapcraftEnvironmentError(
-                'SNAPCRAFT_BUILD_ENVIRONMENT must be one of: host or lxd.')
+                'SNAPCRAFT_BUILD_ENVIRONMENT must be one of: {}.'.format(
+                    humanize_list(items=valid_providers, conjunction='or')))
 
         self.provider = build_provider
         self.is_host = build_provider == 'host'
+        self.is_lxd = build_provider == 'lxd'
--- a/snapcraft/internal/build_providers/_base_provider.py
+++ b/snapcraft/internal/build_providers/_base_provider.py
@@ -65,7 +65,11 @@ class Provider():
 
     @abc.abstractmethod
     def destroy(self):
-        """Provider steps needed to ensure the instance is destroyed."""
+        """Provider steps needed to ensure the instance is destroyed.
+        
+        This method should be safe to call multiple times and do nothing
+        if the instance to destroy is already destroyed.
+        """
 
     @abc.abstractmethod
     def provision_project(self, tarball: str) -> None:
--- a/snapcraft/internal/build_providers/_base_provider.py
+++ b/snapcraft/internal/build_providers/_base_provider.py
@@ -1,0 +1,48 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2018 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import shlex
+
+import petname
+
+
+# This class is a mix of a parent and a mixin.
+class BaseProvider:
+
+    def __init__(self, *, project, executor, echoer) -> None:
+        self.project = project
+        self.echoer = echoer
+        self.instance_name = petname.Generate(2, '-')
+        self.project_dir = shlex.quote(project.info.name)
+        self._exec = executor
+
+        if project.info.version:
+            self.snap_filename = '{}_{}_{}.snap'.format(
+                project.info.name, project.info.version, project.deb_arch)
+        else:
+            self.snap_filename = '{}_{}.snap'.format(
+                project.info.name, project.deb_arch)
+
+    def launch_instance(self, method, **kwargs):
+        self.echoer.info('Creating a build environment named {!r}'.format(
+            self.instance_name))
+        method(instance_name=self.instance_name, **kwargs)
+
+    def setup_snapcraft(self):
+        self.echoer.info('Setting up snapcraft in {!r}'.format(
+            self.instance_name))
+        install_cmd = ['sudo', 'snap', 'install', 'snapcraft', '--classic']
+        self._exec(command=install_cmd, instance_name=self.instance_name)
--- a/snapcraft/internal/build_providers/_base_provider.py
+++ b/snapcraft/internal/build_providers/_base_provider.py
@@ -14,22 +14,22 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import abc
 import shlex
+from typing import Callable
 
 import petname
 
 
-# This class is a mix of a parent and a mixin.
-class BaseProvider:
+class Provider():
 
-    def __init__(self, *, project, executor, echoer) -> None:
+    def __init__(self, *, project, echoer) -> None:
         self.project = project
         self.echoer = echoer
         # Once https://github.com/CanonicalLtd/multipass/issues/220 is
         # closed we can prepend snapcraft- again.
         self.instance_name = petname.Generate(2, '-')
         self.project_dir = shlex.quote(project.info.name)
-        self._exec = executor
 
         if project.info.version:
             self.snap_filename = '{}_{}_{}.snap'.format(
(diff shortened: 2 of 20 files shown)


sergiusens:
To test out this drop
```
snap install snapcraft --channel edge/multipass
cd <project>
SNAPCRAFT_BUILD_ENVIRONMENT=multipass snapcraft cleanbuild
```

popey:
Observations. The naming scheme isn't consistent with lxd. On lxd I end up with snapcraft-foo-bar-baz. With multipass I get foo-bar.

Also, it failed. Took 10 minutes to get to this failure.

```
$ SNAPCRAFT_BUILD_ENVIRONMENT=multipass snapcraft cleanbuild                                                                                                    
Creating a build environment named 'unreelable-kyndall'
Starting unreelable-kyndall                                                  
failed to launch: Attempt to access value of a disengaged optional object       
An error occurred when trying to launch the instance with 'multipass'.  
```

popey:
Tried a second time with the same result.

snapcraft/cli/env.py
```diff
@@ -64,12 +79,12 @@ def __init__(self) -> None:
...
-            echo.warning('Using the host as the build environment.')
-            build_provider = 'host'
-        # TODO add multipass
-        elif build_provider not in ['host', 'lxd']:
+            build_provider = default
+        elif build_provider not in valid_providers:
             raise errors.SnapcraftEnvironmentError(
-                'SNAPCRAFT_BUILD_ENVIRONMENT must be one of: host or lxd.')
+                'SNAPCRAFT_BUILD_ENVIRONMENT must be one of: {}.'.format(
+                    humanize_list(items=valid_providers, conjunction='or')))
```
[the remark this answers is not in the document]
kalikiana: What about testing? Having #1705 in mind where we want to get rid of messages in tests... or are you suggesting to not test this case?

snapcraft/internal/build_providers/_multipass/_multipass_command.py
```diff
@@ -0,0 +1,147 @@
...
+import subprocess
+from typing import List
+
+from snapcraft.internal.build_providers import errors
+
+
+logger = logging.getLogger(__name__)
+
+
+def _run(command: List) -> None:
```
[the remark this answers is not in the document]
kalikiana: The lxd module would be an obvious candidate. Plugin commands could use it as well. But yeah, it can be in a follow-up.

snapcraft/internal/build_providers/_multipass/_instance_info.py
```diff
@@ -0,0 +1,81 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import json
+from typing import TypeVar, Type
+
+from snapcraft.internal.build_providers import errors
+
+T = TypeVar('T', bound='InstanceInfo')
```
kyrofa: Is there a need for this? You can just use `'InstanceInfo'` (the string) as the type; as long as it's valid after this file is processed, mypy will be happy with it.

snapcraft/internal/build_providers/_multipass/_instance_info.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+class InstanceInfo:
+
+    @classmethod
+    def new_instance_info_from_json(cls: Type[T], *, instance_name: str,
```
kyrofa: Callers will be using this as `InstanceInfo.new_instance_info_from_json(...)` which seems redundant. I find it reads very well to name these types of methods `from_X`, so `InstanceInfo.from_json(...)`. Thoughts?

snapcraft/internal/build_providers/_multipass/_instance_info.py
```diff
@@ -0,0 +1,81 @@
...
+                provider_name='multipass',
+                key_info=str(missing_key),
+                data=instance_info) from missing_key
+
+    def __init__(self, *, name: str, state: str, image_release: str) -> None:
+        """Initialize an InstanceInfo.
+
+        :param str name: the instance name.
+        :param str state: the state of the instance which can be any one of
+                          RUNNING, STOPPED, DELETED.
```
kyrofa: This seems a perfect use-case for enums. Thoughts?

snapcraft/internal/build_providers/_multipass/_multipass.py
```diff
@@ -0,0 +1,107 @@
...
+import shlex
+
+from .. import errors
+from .._base_provider import BaseProvider
+from ._instance_info import InstanceInfo
+from ._multipass_command import MultipassCommand
+
+
+class Multipass(BaseProvider):
+    """A multipass provider for snapcraft execute its lifecycle."""
```
kyrofa: To execute?

snapcraft/internal/build_providers/_multipass/_multipass.py
```diff
@@ -0,0 +1,107 @@
...
+        """Create the multipass instance and setup the build environment."""
+        self.launch_instance(self._multipass_cmd.launch, image='16.04')
+        self.setup_snapcraft()
+
+    def destroy(self):
+        """Destroy the instance, trying to stop it first."""
+        try:
+            instance_info = self._get_instance_info()
+        except errors.ProviderInfoError as info_error:
+            self.echoer.warning(
```
kyrofa: Is this not an error? We weren't able to do as requested, but to the caller, this will look like a success.

snapcraft/internal/build_providers/_multipass/_multipass_command.py
```diff
@@ -0,0 +1,147 @@
...
+
+        :raises errors.ProviderCommandNotFound:
+            if the multipass command is not found.
+        """
+        provider_cmd = 'multipass'
+        if not shutil.which(provider_cmd):
+            raise errors.ProviderCommandNotFound(command=provider_cmd)
+        self.provider_cmd = provider_cmd
+        # Workaround for https://github.com/CanonicalLtd/multipass/issues/221
+        signal.signal(signal.SIGUSR1, _ignore_signal)
```
kyrofa: What.

snapcraft/internal/build_providers/_multipass/_multipass_command.py
```diff
@@ -0,0 +1,147 @@
...
+        """
+        cmd = [self.provider_cmd, 'exec', instance_name, '--'] + command
+        try:
+            _run(cmd)
+        except subprocess.CalledProcessError as process_error:
+            raise errors.ProviderExecError(
+                provider_name=self.provider_name,
+                command=command) from process_error
+
+    def copy_files(self, *, source: str, destination: str):
```
kyrofa: Missing a `-> None` here

snapcraft/internal/build_providers/_multipass/_multipass_command.py
```diff
@@ -0,0 +1,147 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import logging
+import signal
+import shutil
+import subprocess
+from typing import List
```
Saviq: Uh oh, /me needs to school up on new python...

snapcraft/internal/build_providers/errors.py
```diff
@@ -0,0 +1,136 @@
...
+    )
+
+    def __init__(self, *, provider: str) -> None:
+        super().__init__(provider=provider)
+
+
+class ProviderCommandNotFound(_SnapcraftError):
+
+    fmt = (
+        'The {command!r} command is necessary to be able to build in this '
```
kyrofa: This jumps straight to why this is a problem instead of starting with the actual problem. Consider "{command!r} command not found: this command is necessary to build in this environment.\nInstall {command!r}...".

snapcraft/internal/build_providers/errors.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def __init__(self, *, command: str) -> None:
+        super().__init__(command=command)
+
+
+class ProviderLaunchError(_SnapcraftError):
+
+    fmt = (
+        'An error occurred when trying to launch the instance with '
+        '{provider_name!r}.'
```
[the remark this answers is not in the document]
sergiusens: Should be optional, these errors should be usable by things that talk over protocols as well (which do have error codes!)

snapcraft/internal/build_providers/errors.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def __init__(self, *, provider_name: str) -> None:
+        super().__init__(provider_name=provider_name)
+
+
+class ProviderInfoDataKeyError(_SnapcraftError):
+
+    fmt = (
+        'The data returned by {provider_name!r} was not expected. '
+        'It is missing some key data {key_info!r} in {data!r}.'
```
kyrofa: I don't quite parse this. "key data {key_info!r}" -> is key_info a key in data? What is key data? Like... key = important?

snapcraft/internal/build_providers/_multipass/_multipass.py
sergiusens: Let's live chat about this one

snapcraft/internal/build_providers/_multipass/_multipass_command.py
sergiusens: That was most of my Saturday!

snapcraft/internal/build_providers/_factory.py
```diff
@@ -0,0 +1,28 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from typing import Union, Type
+
+from . import errors
+from ._multipass import Multipass
+
+
+def get_provider_for(provider_name: str) -> Union[Type[Multipass]]:
```
kyrofa: This should probably return a BaseProvider, no?

snapcraft/internal/build_providers/_base_provider.py
```diff
@@ -0,0 +1,50 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import shlex
+
+import petname
+
+
+# This class is a mix of a parent and a mixin.
+class BaseProvider:
```
kyrofa: I'd personally rather see this called `Provider`. "Multipass is a provider" versus "multipass is a base provider".

snapcraft/internal/lifecycle/_containers.py
```diff
@@ -45,16 +45,33 @@ def containerbuild(step, project_options, output=None, args=[]):
...
-        config.data['name'], config.data['version'])
+    if build_environment.is_lxd:
+        _deprecated_cleanbuild(project, remote, config, tar_filename)
+        return
 
+    build_provider_class = build_providers.get_provider_for('multipass')
+    with build_provider_class(project=project, echoer=echoer) as instance:
+        instance.provision_project(tar_filename)
+        instance.build_project()
+        instance.retrieve_snap()
```
kyrofa: Given the factory nature of this, the interface used here should probably be implemented by the `BaseProvider`.

snapcraft/project/__init__.py
```diff
@@ -15,3 +15,4 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 from ._project import Project                          # noqa F401
+from ._project_info import ProjectInfo                 # noqa F401
```
kyrofa: Eventually we want to get to the point where the Project can fill out the ProjectInfo based on its __init__, which means this won't need to be imported here. Would it make sense to just import the private module in the test instead of making it publicly available now?

snapcraft/internal/build_providers/_multipass/_multipass_command.py
kyrofa: :hugs:

snapcraft/project/__init__.py
kalikiana: Fwiw `test_config` imports the private module by name today as "from snapcraft.project._project_info import ProjectInfo" for that reason.

snapcraft/internal/build_providers/_multipass/_multipass_command.py
Saviq: Fixed in `edge` ;)

snapcraft/internal/build_providers/_base_provider.py
```diff
@@ -0,0 +1,97 @@
...
+    def __enter__(self):
+        self.create()
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.destroy()
+
+    @property
+    @abc.abstractmethod
+    def run(self) -> Callable[[str], None]:
```
kyrofa: It seems odd to have this as a property, especially considering that it's actually a callable, so it LOOKS like a method when used. Why not just make it a method, and the implementer can just pass it on to their command instance?

snapcraft/internal/build_providers/_base_provider.py
```diff
@@ -0,0 +1,97 @@
...
+        """Return a callable to run commands on the the instance.
+
+        :returns: a callable which takes one argument, a list, with the
+                  command to run.
+        """
+
+    @property
+    @abc.abstractmethod
+    def launch(self) -> Callable[[], None]:
```
kyrofa: Same here, feels like it should be a method.

snapcraft/internal/build_providers/_multipass/_multipass.py
```diff
@@ -0,0 +1,110 @@
...
+
+        try:
+            if not self._instance_info.is_stopped():
+                self._multipass_cmd.stop(instance_name=self.instance_name)
+            self._multipass_cmd.delete(instance_name=self.instance_name)
+        except errors.ProviderStopError as stop_error:
+            self.echoer.warning('Could not stop {!r}: {}.'.format(
+                self.instance_name, stop_error))
+        except errors.ProviderDeleteError as stop_error:
+            self.echoer.warning('Could not stop {!r}: {}.'.format(
```
kyrofa: Could not delete?

snapcraft/internal/build_providers/_base_provider.py
```diff
@@ -0,0 +1,97 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import abc
+import shlex
+from typing import Callable
+
+import petname
+
+
+class Provider():
```
kyrofa: This is heading the right direction, but I'm losing track of what is and isn't public API ("public" defined to be "meant to be called by a user of `Provider`). For example, `run` and `launch` are both public, but I don't believe they're actually meant to be called by anything except `Provider`, right?

snapcraft/internal/build_providers/_base_provider.py
```diff
@@ -0,0 +1,97 @@
...
+        """Provider steps needed to copy project assests to the instance."""
+
+    @abc.abstractmethod
+    def build_project(self) -> None:
+        """Provider steps needed build the project on the instance."""
+
+    @abc.abstractmethod
+    def retrieve_snap(self) -> str:
+        """
+        Provider steps needed to retrieve the built snap from the instance.
```
kyrofa: Can you please document what this should return?

snapcraft/internal/build_providers/_base_provider.py
```diff
@@ -0,0 +1,97 @@
...
+                  command to run.
+        """
+
+    @property
+    @abc.abstractmethod
+    def launch(self) -> Callable[[], None]:
+        """Return a callable that can be used to launch an instance."""
+
+    @abc.abstractmethod
+    def create(self):
```
kyrofa: Missing `-> None` for both of these.

PR #2101, opened 2018-04-23
errors: implement an always option to send to sentry

sergiusens:

-----
Meant for a rebase and commit

### DIFF
--- a/snapcraft/cli/_errors.py
+++ b/snapcraft/cli/_errors.py
@@ -21,6 +21,7 @@ import traceback
 from textwrap import dedent
 
 from . import echo
+from snapcraft.config import CLIConfig as _CLIConfig
 from snapcraft.internal import errors
 
 import click
@@ -38,12 +39,16 @@ except ImportError:
 _MSG_TRACEBACK = dedent("""\
     Sorry, Snapcraft ran into an error when trying to running through its
     lifecycle that generated the following traceback:""")
-_MSG_SEND_TO_SENTRY_TRACEBACK_CONFIRM = dedent("""\
+_MSG_SEND_TO_SENTRY_TRACEBACK_PROMPT = dedent("""\
     You can anonymously report this issue to the snapcraft developers.
     No other data than this traceback and the version of snapcraft in use will
     be sent.
-    Would you like send this error data?""")
+    Would you like send this error data? (Yes/No/Always)""")
 _MSG_SEND_TO_SENTRY_THANKS = 'Thank you for sending the report.'
+
+_YES_VALUES = ['yes', 'y']
+_NO_VALUES = ['no', 'n']
+_ALWAYS_VALUES = ['always', 'a']
 
 
 def exception_handler(exception_type, exception, exception_traceback, *,
@@ -74,12 +79,12 @@ def exception_handler(exception_type, exception, exception_traceback, *,
         click.echo(_MSG_TRACEBACK)
         traceback.print_exception(
             exception_type, exception, exception_traceback)
-        msg = _MSG_SEND_TO_SENTRY_TRACEBACK_CONFIRM
         if not is_raven_setup:
             echo.warning(
                 'raven is not installed on this system, cannot send data '
                 'to sentry')
-        elif click.confirm(msg):
+        elif _is_send_to_sentry():
+            echo.info('Sending this error report.')
             _submit_trace(exception)
             click.echo(_MSG_SEND_TO_SENTRY_THANKS)
     elif not is_snapcraft_error:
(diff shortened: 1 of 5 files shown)


snapcraft/cli/_errors.py
```diff
@@ -110,6 +108,57 @@ def exception_handler(exception_type, exception, exception_traceback, *,
...
+    # prompt again.
+    config_errors = None
+    try:
+        cli_config = _CLIConfig()
+        always_value = cli_config.get_option('Sentry', 'always_send')
+        if always_value == 'true':
+            return True
+    except errors.SnapcraftInvalidCLIConfigError as config_error:
+        raise config_error
+        echo.warning('There was an issue while trying to retrieve '
```
kyrofa: This probably isn't what you want. Do we want to re-raise, or print a warning?

snapcraft/cli/_errors.py
```diff
@@ -110,6 +108,57 @@ def exception_handler(exception_type, exception, exception_traceback, *,
...
+                     'configuration data: {!s}'.format(config_error))
+        config_errors = config_error
+
+    # Either ALWAYS has not been selected in a previous run so resort to
```
kyrofa: s/Either//

snapcraft/cli/_errors.py
```diff
@@ -110,6 +108,57 @@ def exception_handler(exception_type, exception, exception_traceback, *,
...
+        echo.warning('Could not determine choice, assuming no.')
+        return False
+
+
+def _prompt_sentry():
+    msg = _MSG_SEND_TO_SENTRY_TRACEBACK_PROMPT
+    all_valid = _YES_VALUES + _NO_VALUES + _ALWAYS_VALUES
+
+    def validate(value):
+        if value.lower() in all_valid:
```
kyrofa: We check that it's valid in lower-case, but then don't use `.lower()` when decided on the proper action. We probably should.

snapcraft/config.py
```diff
@@ -25,16 +25,60 @@
...
-)
+from snapcraft.internal.errors import SnapcraftInvalidCLIConfigError
+from snapcraft.storeapi import constants, errors
 
 LOCAL_CONFIG_FILENAME = '.snapcraft/snapcraft.cfg'
 
 logger = logging.getLogger(__name__)
 
 
+class CLIConfig:
```
kyrofa: This class is really just a dict that saves keys and values to to a file. I'm not a huge fan, for the same reason I don't like accessing the well-defined snapcraft.yaml as a dict: one typo, and you're getting the wrong thing. Want to change the key name? You need to change it in every place you use it. We have the same problem with needing to remember that we need to use `'Sentry'` and `'always_send'`. Why not encapsulate that knowledge here and provide a specific API for setting/getting that property?

snapcraft/cli/_errors.py
sergiusens: ooh, I was playing with this, yes, just the warning, we cannot re raise as we are in the exception already

snapcraft/config.py
sergiusens: sure

sergiusens: I am plus one on this, just didn't want to move away from the status quo too much :-)

snapcraft/cli/_errors.py
sergiusens: interesting, they were there :-/

snapcraft/cli/_errors.py
sergiusens: I was actually missing the **or** :sunglasses:

snapcraft/cli/_errors.py
kyrofa: Bah, I'm blind, sorry :grin: .

snapcraft/cli/_errors.py
```diff
@@ -38,15 +39,17 @@
...
     lifecycle that generated the following traceback:""")
-_MSG_SEND_TO_SENTRY_TRACEBACK_CONFIRM = dedent("""\
+_MSG_SEND_TO_SENTRY_TRACEBACK_PROMPT = dedent("""\
     You can anonymously report this issue to the snapcraft developers.
     No other data than this traceback and the version of snapcraft in use will
     be sent.
-    Would you like send this error data?""")
-_MSG_SEND_TO_SENTRY_ENV = dedent("""\
-    Sending error data: SNAPCRAFT_SEND_ERROR_DATA is set to 'y'.""")
+    Would you like send this error data? (Yes/[No]/Always)""")
```
kyrofa: This actually turns into "Would you like send this error data? (Yes/[No]/Always) [no]"; indicating the default twice looks a tad odd.

snapcraft/cli/_errors.py
```diff
@@ -38,15 +39,17 @@
...
 _MSG_SEND_TO_SENTRY_THANKS = 'Thank you for sending the report.'
```
kyrofa: Once I've said "always", tracebacks get automatically submitted to Sentry, but it still just says "Thank you for sending the report." The spec says it should actually say that it's sending before thanking, and it might make sense to expand slightly and say specifically that it's sending BECAUSE always was specifically requested previously: "Always sending error reports as requested" or something.

snapcraft/cli/_errors.py
sergiusens: heh, the brackets are what define the default answer.

sergiusens: oh noes, err.

PR #2104, opened 2018-04-24
many: allow building in containers with no version in project

sergiusens:
Support building in containers now that version can be retrieved
with `adopt-info`.

-----

### DIFF
--- a/snapcraft/internal/lxd/_containerbuild.py
+++ b/snapcraft/internal/lxd/_containerbuild.py
@@ -66,10 +66,11 @@ _STORE_KEY = (
 
 class Containerbuild:
 
-    def __init__(self, *, output, source, project_options,
-                 metadata, container_name, remote=None):
-        if not output:
-            output = common.format_snap_name(metadata)
+    def __init__(self, *, source, project_options, metadata,
+                 container_name, output=None, remote=None):
+        if output is None:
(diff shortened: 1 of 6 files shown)


snapcraft/internal/lxd/_containerbuild.py
```diff
@@ -66,10 +66,11 @@
...
 
-    def __init__(self, *, output, source, project_options,
-                 metadata, container_name, remote=None):
-        if not output:
-            output = common.format_snap_name(metadata)
+    def __init__(self, *, source, project_options, metadata,
+                 container_name, output=None, remote=None):
+        if output is None:
+            output = common.format_snap_name(
+                metadata, allow_empty_version=True)
```
kalikiana: Can we check that adopt-info is actually present in the YAML? To avoid spawning the container only to fail at the build stage.

kyrofa: That's part of the schema validation, which happens before this point. No need to recheck.

sergiusens: I am enforcing this parameter to not break contract unknowingly. At this point in time, version **will** be empty even if adopt-info is in use and as @kyrofa mentioned, it is checked for.

PR #2115, opened 2018-05-01
storeapi: ensure snap ID is sane before using it

kyrofa:

-----

The store had a bug that ended up causing a snap to be registered to a user, but have no ID. The snapcraft CLI didn't handle this well, producing a traceback. This PR fixes [LP: #1763721](https://bugs.launchpad.net/snapcraft/+bug/1763721) by checking the snap ID before using it.

### DIFF
--- a/snapcraft/storeapi/_store_client.py
+++ b/snapcraft/storeapi/_store_client.py
@@ -172,6 +172,9 @@ class StoreClient():
         except KeyError:
             raise errors.SnapNotFoundError(snap_name, series=series, arch=arch)
 
+        if not snap_id:
+            raise errors.NoSnapIdError(snap_name)
+
         response = self._refresh_if_necessary(
             self.sca.snap_revisions, snap_id, series, arch)
 
@@ -189,6 +192,9 @@ class StoreClient():
             snap_id = account_info['snaps'][series][snap_name]['snap-id']
         except KeyError:
             raise errors.SnapNotFoundError(snap_name, series=series, arch=arch)
+
+        if not snap_id:
+            raise errors.NoSnapIdError(snap_name)
 
         response = self._refresh_if_necessary(
             self.sca.snap_status, snap_id, series, arch)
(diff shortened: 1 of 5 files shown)


snapcraft/storeapi/_store_client.py
```diff
@@ -172,6 +172,9 @@ def get_snap_revisions(self, snap_name, series=None, arch=None):
         except KeyError:
             raise errors.SnapNotFoundError(snap_name, series=series, arch=arch)
 
+        if not snap_id:
```
sergiusens: should we do `is None` to contemplate the possibility of something else? Maybe not worth it, but it may be a valid `snap_id`

sparkiegeek: I would like an explicit comparison to `None` too, we shouldn't make any assumptions about truthiness of valid `snap_id` s

snapcraft/storeapi/errors.py
```diff
@@ -105,6 +106,17 @@ def __init__(self, name, channel=None, arch=None, series=None):
         super().__init__(name=name, channel=channel, arch=arch, series=series)
 
 
+class NoSnapIdError(StoreError):
+
+    fmt = (
+        'Unable to get snap ID for snap {snap_name!r}: The store returned '
+        'data, but included no ID. This is an error in the store, please open '
+        'a new topic on the forum: {forum_url}')
```
sparkiegeek: A better error message suggestion:

"Missing snap ID for {snap_name!r}. Please open a new topic in the store category on the forum: {forum_url}"

Can we link to `/c/store` directly?

snapcraft/storeapi/_store_client.py
kyrofa: This is used [here](https://github.com/snapcore/snapcraft/blob/master/snapcraft/storeapi/_sca_client.py#L281). Surely there are no falsey values that would work in this case?

snapcraft/storeapi/errors.py
kyrofa: I reworked the error a little to link to the store category.

snapcraft/storeapi/_store_client.py
kyrofa: Alright, I've updated the PR to compare to None.

PR #2116, opened 2018-05-01
storeapi: handle 5xx error codes for all store endpoints

kyrofa:

-----

Currently, the snapcraft CLI handles a 500 response when releasing a snap. This PR fixes [LP: #1768370](https://bugs.launchpad.net/snapcraft/+bug/1768370) by making this more robust, and nicely handling 5xx codes for all endpoints.

### DIFF
--- a/snapcraft/storeapi/errors.py
+++ b/snapcraft/storeapi/errors.py
@@ -375,11 +375,12 @@ class StoreServerError(StoreError):
     fmt = '{what}: {error_text} (code {error_code}).\n{action}'
 
     def __init__(self, response):
-        what = 'The store encountered a server error'
-        error_text = responses[response.status_code].lower()
+        what = ('The Snap Store encountered an error while processing your '
+                'request')
         error_code = response.status_code
-        action = ('The status of the store and associated services can be '
-                  'checked at {}'.format(_STORE_STATUS_URL))
+        error_text = responses[error_code].lower()
+        action = ('The operational status of the Snap Store can be checked at '
+                  '{}'.format(_STORE_STATUS_URL))
 
         super().__init__(
             response=response, what=what, error_text=error_text,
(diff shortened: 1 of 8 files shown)


snapcraft/storeapi/errors.py
```diff
@@ -357,6 +357,28 @@ def __init__(self, result):
         super().__init__()
 
 
+class StoreInternalError(StoreError):
+
+    fmt = '{what}: {why}\n{action}'
+
+    def __init__(self, response):
+        what = 'The store encountered an internal error'
+        why = None
```
sergiusens: There will never be a case (and if there is, wee shouldn't show) any scenario where we have json for a 500. We should also leave this as a fixed message error, raised when we do not find json in a response.

kyrofa: Alright, I've updated the PR to no longer use the error list.

snapcraft/storeapi/errors.py
```diff
@@ -369,6 +370,22 @@ def __init__(self, result):
...
 
 
+class StoreServerError(StoreError):
+
+    fmt = '{what}: {error_text} (code {error_code}).\n{action}'
+
+    def __init__(self, response):
+        what = 'The store encountered a server error'
+        error_text = responses[response.status_code].lower()
+        error_code = response.status_code
```
sparkiegeek: if you switched these lines around, you could use `error_code` to lookup `responses`

snapcraft/storeapi/errors.py
```diff
@@ -369,6 +370,22 @@ def __init__(self, result):
         super().__init__()
 
 
+class StoreServerError(StoreError):
+
+    fmt = '{what}: {error_text} (code {error_code}).\n{action}'
+
+    def __init__(self, response):
+        what = 'The store encountered a server error'
```
evandandrea: “The Snap Store encountered an error while processing your request:”
…
“The operational status of the Snap Store can be checked at https://status.snapcraft.io”

Let’s be consistent. Snap Store is always capitalised.

kyrofa: Alright, done.

PR #2120, opened 2018-05-03
many: dedup environment entries

sergiusens:
Deduplicate the environment entries when generating it for a specific part.
Also wrap the override script into an actual file to reduce the possibility
of an argument list error occurring.

LP: #1767625

-----

### DIFF
--- a/snapcraft/internal/pluginhandler/_runner.py
+++ b/snapcraft/internal/pluginhandler/_runner.py
@@ -116,20 +116,6 @@ class Runner:
                 'override-prime', self._override_prime_scriptlet,
                 self._primedir)
 
-    def _get_env(self):
-        env = ''
-        if common.is_snap():
-            # Since the snap is classic, $SNAP/bin is not on the $PATH.
-            # Let's set an alias to make sure it's found (but only if it
-            # exists).
-            snapcraftctl_path = os.path.join(
-                os.getenv('SNAP'), 'bin', 'snapcraftctl')
-            if os.path.exists(snapcraftctl_path):
-                env += 'alias snapcraftctl="$SNAP/bin/snapcraftctl"\n'
-        env += common.assemble_env()
-
-        return env
-
     def _run_scriptlet(self, scriptlet_name: str, scriptlet: str,
                        workdir: str) -> None:
         with tempfile.TemporaryDirectory() as tempdir:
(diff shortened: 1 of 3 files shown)


snapcraft/internal/pluginhandler/_runner.py
```diff
@@ -124,32 +138,27 @@ def _run_scriptlet(self, scriptlet_name: str, scriptlet: str,
...
-
-            process = subprocess.Popen(
-                ['/bin/sh', '-e', '-c', script], cwd=workdir)
+                    feedback_fifo=feedback_fifo.path, scriptlet=scriptlet)
+
+            # We write to a file as the script may be to long and hit the
+            # argument list limit.
+            run_script = os.path.join(tempdir, 'run_script')
+            with open(run_script, 'w+') as script_file:
+                print('set -e', file=script_file)
```
kyrofa: You can just leave the `-e` in the call to `/bin/sh`, unless you prefer this.

snapcraft/internal/project_loader/_parts_config.py
```diff
@@ -254,4 +254,13 @@ def build_env_for_part(self, part, root_part=True):
             env += dep_part.env(stagedir)
             env += self.build_env_for_part(dep_part, root_part=False)
 
-        return env
+        # LP: #1767625
+        # Remove duplicates from using the same plugin in dependent parts.
+        seen = dict()
```
kyrofa: This looks like it should be a set, no?

sergiusens: yeah, what was I thinking

snapcraft/internal/pluginhandler/_runner.py
```diff
@@ -116,6 +116,20 @@ def prime(self) -> None:
                 'override-prime', self._override_prime_scriptlet,
                 self._primedir)
 
+    def _get_env(self):
```
kyrofa: I don't see a reason this should be a method of this class, I suggest we make it standalone.

snapcraft/internal/pluginhandler/_runner.py
```diff
@@ -124,32 +138,27 @@ def _run_scriptlet(self, scriptlet_name: str, scriptlet: str,
...
-            env += common.assemble_env()
-
             # snapcraftctl only works consistently if it's using the exact same
             # interpreter as that used by snapcraft itself, thus the definition
             # of SNAPCRAFT_INTERPRETER.
             script = textwrap.dedent("""\
                 export SNAPCRAFTCTL_CALL_FIFO={call_fifo}
                 export SNAPCRAFTCTL_FEEDBACK_FIFO={feedback_fifo}
                 export SNAPCRAFT_INTERPRETER={interpreter}
-                {env}
```
kyrofa: I definitely understand and appreciate the extraction of the env stuff above into a method, but I don't understand why it was extracted from this script only to be added later to the script below. I'd kinda prefer to see the script all together in one place, personally.

snapcraft/internal/pluginhandler/_runner.py
```diff
@@ -124,32 +138,27 @@ def _run_scriptlet(self, scriptlet_name: str, scriptlet: str,
...
+            # We write to a file as the script may be to long and hit the
+            # argument list limit.
+            run_script = os.path.join(tempdir, 'run_script')
+            with open(run_script, 'w+') as script_file:
+                print('set -e', file=script_file)
+                print(self._get_env(), file=script_file)
+                print(script, file=script_file)
+
+            process = subprocess.Popen(['/bin/sh', script_file.name],
+                                       cwd=workdir)
```
kyrofa: We knew this would hide errors, and it does. I really don't like it. I think this should also get around the max args limit, but it keeps the good errors (I tested it):

```python
run_script = os.path.join(tempdir, 'run_script')
with open(run_script, 'w+') as script_file:
    print('set -e', file=script_file)
    print(self._get_env(), file=script_file)
    print(script, file=script_file)
    script_file.flush()
    script_file.seek(0)

    process = subprocess.Popen(['/bin/sh'], stdin=script_file,
                               cwd=workdir)
```

What do you think? That can be further simplified by moving the `set -e` to the `-e` option and keeping `_get_env()` as part of the `script`.

snapcraft/internal/pluginhandler/_runner.py
sergiusens: Oh, I wanted to split up the "real" script with the modifications we've done to the script, it becomes easier to debug at that point

snapcraft/internal/pluginhandler/_runner.py
sergiusens: I guess we could

PR #2122, opened 2018-05-04
many: introduce variables for part src and build

kyrofa:

-----

This PR resolves [LP: #1763028](https://bugs.launchpad.net/snapcraft/+bug/1763028) by adding new environment variables (and replacements) for part src and build directories. It also refactors the `SNAPCRAFT_*` environment handling to happen in one place instead of multiple.

### DIFF
--- a/snapcraft/internal/project_loader/_env.py
+++ b/snapcraft/internal/project_loader/_env.py
@@ -121,9 +121,9 @@ def snapcraft_global_environment(project: project.Project) -> Dict[str, str]:
 def snapcraft_part_environment(
         part: pluginhandler.PluginHandler) -> Dict[str, str]:
     return {
-        'SNAPCRAFT_PART_SRC': part.sourcedir,
-        'SNAPCRAFT_PART_BUILD': part.builddir,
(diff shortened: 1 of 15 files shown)


snapcraft/internal/project_loader/_env.py
```diff
@@ -89,3 +89,49 @@ def build_env_for_stage(stagedir: str, snap_name: str,
...
+        'SNAPCRAFT_PRIME': project.prime_dir,
+    }
+
+
+def snapcraft_part_environment(
+        part: pluginhandler.PluginHandler) -> Dict[str, str]:
+    return {
+        'SNAPCRAFT_PART_SRC': part.sourcedir,
+        'SNAPCRAFT_PART_BUILD': part.builddir,
+        'SNAPCRAFT_PART_INSTALL': part.installdir,
```
sergiusens: if this is the only place `part.installdir` is used, I would rather go the other way around and removing the property definition instead of creating a new one. And if not make this use `part.plugin.<dir>` instead and later on remove the `installdir` property.

kyrofa: Done.

PR #2131, opened 2018-05-11
No user site for snapcraft

sergiusens:

-----

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -36,6 +36,7 @@ parts:
         - libsodium-dev
         - liblzma-dev
         - patch
+        - sed
     stage-packages:
         - binutils
         - execstack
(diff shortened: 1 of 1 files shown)


snap/snapcraft.yaml
```diff
@@ -36,6 +36,7 @@ parts:
         - libsodium-dev
         - liblzma-dev
         - patch
+        - sed
```
sparkiegeek: sed is an Essential package, for normal Debian(-based) systems this means that packages don't need to express a dependency on it, since it will *always* be installed. Do you anticipate needing to support build environments that don't respect this?

sergiusens: it is harmless and while low risk of removal, when work on the minimal docker images took place we were hit by it (I forget which package it was now).

kyrofa: Strip, which isn't essential, but yeah I think it makes sense to keep dependencies listed here, even if it's essentially a noop in this particular case.

Issue #2134, opened 2018-05-12
gradle plugin doesn't generate a war


bsutton:
I have a build.gradle that uses the 'war' gradle plugin to generate a war file.
If I run ./gradlew build I get the expected war file.

The gradle plug however essentially runs:

gradlew jar

The result is that the 'jar' task is executed and a jar is generated rather than the desired war file.
I'm no gradle expert but would it not make more sense to just run 'gradle build' ?

Alternatively add a gradle option to allow the desired task/tasks to be executed.
This would work for backward compatibility in that you could do:

gradle:
    task: [test, war]

If the task was not provided then it would default to 

gradle:
  tasks: [jar]

Currently the 'jar' task is just automatically appended to the 'gradle-option' list so this should work nicely.

kyrofa:
For future reference, please note that we use launchpad for bugs, not github: https://bugs.launchpad.net/snapcraft

PR #2135, opened 2018-05-12
Add gradle support for task other than just 'jar'

bsutton:
improved error when no jar found by displaying the path that was searched. Will make diagnosing build problems faster.

-----

### DIFF
--- a/snapcraft/plugins/gradle.py
+++ b/snapcraft/plugins/gradle.py
@@ -104,7 +104,7 @@ class GradlePlugin(snapcraft.plugins.jdk.JdkPlugin):
             basedir = 'war'
             jarfiles = warfiles
         else:
-            raise RuntimeError("Could not find any built jar files for part")
+            raise RuntimeError("Could not find any built jar files for part in: " + src)
 
         snapcraft.file_utils.link_or_copy_tree(
             src, os.path.join(self.installdir, basedir),
--- a/snapcraft/plugins/gradle.py
+++ b/snapcraft/plugins/gradle.py
@@ -118,7 +118,7 @@ class GradlePlugin(snapcraft.plugins.jdk.JdkPlugin):
             basedir = 'war'
             jarfiles = warfiles
         else:
-            raise RuntimeError("Could not find any built jar files in the part directory:" + src)
+            raise RuntimeError("Could not find any built jar/war files in the part directory:" + src)
 
         snapcraft.file_utils.link_or_copy_tree(
             src, os.path.join(self.installdir, basedir),
--- a/snapcraft/plugins/gradle.py
+++ b/snapcraft/plugins/gradle.py
@@ -118,7 +118,7 @@ class GradlePlugin(snapcraft.plugins.jdk.JdkPlugin):
             basedir = 'war'
             jarfiles = warfiles
         else:
-            raise RuntimeError("Could not find any built jar/war files in the part directory:" + src)
+            raise RuntimeError('Could not find any built jar/war files in the part directory: {!r}'.format(src))
 
         snapcraft.file_utils.link_or_copy_tree(
             src, os.path.join(self.installdir, basedir),
--- a/snapcraft/plugins/gradle.py
+++ b/snapcraft/plugins/gradle.py
@@ -72,7 +72,7 @@ class GradlePlugin(snapcraft.plugins.jdk.JdkPlugin):
             'items': {
                 'type': 'string',
             },
-            'default': [jar],
+            'default': ['jar'],
         }
         schema['properties']['gradle-output-dir'] = {
             'type': 'string',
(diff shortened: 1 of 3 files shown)


bsutton:
I've test this patch using snapcraft-pr 2135.

The test have included:
No gradle-options keyword - the jar task was run as expected

gradle-options: [-xtest] - the jar task was run as expected

gradle-options:[-xtest]
gradle-tasks:[jar]  - the jar task was run as expected

gradle-options: [-xtest]
gradle-tasks:[war] - the war task was run as expected, the jar task did not run as expected.

gradle-options: [-xtest]
gradle-tasks:[jar, war] - both the jar and war task ran as expected.

bsutton:
Fixes #2134

bsutton:
How do I get this pull request accepted?

There is a travis error but that appears to be a problem with travis not my patch:

Travis error:

    100% |████████████████████████████████| 61kB 2.3MB/s 
Collecting coverage==4.1 (from -r requirements-devel.txt (line 2))
Exception:

snapcraft/plugins/gradle.py
```diff
@@ -92,7 +105,8 @@ def build(self):
             gradle_cmd = ['gradle']
         self.run(gradle_cmd +
                  self._get_proxy_options() +
-                 self.options.gradle_options + ['jar'])
+                 self.options.gradle_options +
+				 self.options.gradle_tasks)
```
kyrofa: I'm all for this change in principle, as gradle isn't limited to the `jar` tasks, or even the `java` plugin (there's a `cpp` one, for example). However, now that we've opened up the ability to use gradle for other things, we still have the following logic that looks for .jar or .war files and bails out if it doesn't find them. If we open up this capability, we probably need to make that more generic as well. Thoughts?

snapcraft/plugins/gradle.py
```diff
@@ -104,7 +118,7 @@ def build(self):
             basedir = 'war'
             jarfiles = warfiles
         else:
-            raise RuntimeError("Could not find any built jar files for part")
+            raise RuntimeError("Could not find any built jar/war files in the part directory:" + src)
```
kyrofa: Please use `.format()` here instead of appending (preferably with `{!r}`). Also, please update this to use single quotes.

kyrofa:
Also, this is lacking tests.

snapcraft/plugins/gradle.py
bsutton: so I couple of issues. 1) I'm new to gradle 2) I'm not a python programmer.

I agree that we should extend this to support other outputs, however I've not sufficient knowledge of gradle/python to propose a patch that is likely to do the required job.

So two options:
1) we take the pull request as is as it adds useful functionality and does it in a way that would allow for a future update to support other output targets

2) we find a second contributor that can provide some advice on what's required to make this more generic to other output targets.

I'm inclined to go for 1) as it makes the new functionality available now and I think its sufficiently low risk that it will make future enhancements break backwards compatibility.

snapcraft/plugins/gradle.py
bsutton: I will push an update shortly.

snapcraft/plugins/gradle.py
bsutton: As to tests can you point me in the right direction as to how to create a test/where to place a test for this?

bsutton: OK, so I found the test cases, but in looking at the code I think its going to be well beyond my ability with python to build a coherent test case for this.
I'm using it in production so it does work, but that doesn't help for future maintainability.

bsutton: OK, so wasn't as hard I as i thought.
I've created some unit tests but I'm unable to run them the the dev setup guide is a bit vauge.

I'm getting the following error, what do I need to install to run the unit tests?

Google didn't turn up anything useful.


/runtests.sh tests/unit
tests.unit (unittest.loader._FailedTest) ... ERROR

ERROR: tests.unit (unittest.loader._FailedTest)
ImportError: Failed to import test module: tests.unit
  File "/usr/lib/python3.6/unittest/loader.py", line 462, in _find_test_path
    package = self._get_module_from_name(name)
  File "/usr/lib/python3.6/unittest/loader.py", line 369, in _get_module_from_name
    __import__(name)
  File "/home/bsutton/git/snapcore/snapcraft/tests/unit/__init__.py", line 23, in <module>
    import fixtures
ModuleNotFoundError: No module named 'fixtures'

sergiusens:
@kyrofa are you going to remove your review or verify you are satisfied?
@bsutton are you going to look into the test failures and merge conflicts?

Much has changed since this PR was open. Might be worth considering closing it and starting from a clean slate?

kyrofa:
@sergiusens as far as I can tell, my review still stands. The tests never passed and the issue I outlined is still present. If you want to land it as-is, you're welcome to dismiss my review.

sergiusens:
I am closing this considering @kyrofa's comment that the review still stands, @bsutton feel free to reopen once the issues have been addressed.

PR #2146, opened 2018-05-24
project_loader: process parts in consistent order

kyrofa:

-----

Currently, if parts are not strictly ordered by using `after`, the snapcraft CLI will process them in a completely non-deterministic order that changes between runs. This can make it extremely difficult to track down bugs that only occur when building in a specific order. This PR resolves [LP: #1759875](https://bugs.launchpad.net/snapcraft/+bug/1759875) by processing parts consistently and deterministically by sorting them by name.

### DIFF
--- a/tests/unit/project_loader/test_parts.py
+++ b/tests/unit/project_loader/test_parts.py
@@ -21,7 +21,7 @@ import unittest.mock
 from textwrap import dedent
 
 import fixtures
-from testtools.matchers import Contains, Equals
+from testtools.matchers import Contains, Equals, HasLength
 
 import snapcraft
 from snapcraft.internal import (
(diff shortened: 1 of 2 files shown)


tests/unit/project_loader/test_parts.py
```diff
@@ -138,3 +138,71 @@ def test_load_with_dummy_repo(self, os_release_mock):
...
+                description: test
+                confinement: strict
+
+                parts:
+                  part2:
+                    plugin: nil
+                    after: [part3]
+                  part1:
+                    plugin: nil
+                    after: [part3]
```
sergiusens: can I see a test with only `part2` having `after: [part3]`?

kyrofa: Of course, done.

kyrofa:
> I think we need one more test, to make sure consistency is a given with the use of after.

Done.

> Also, a counter proposal, albeit harder to implement would be to just use the ordered load from the OrderedDict and when an unsatisfied prereq is detected call it out in the order that it is declared in on the after list.

The YAML spec does not guarantee order in this case; I feel this would be relying on implementation details. I also like that the current approach will always process using the same order, regardless of the order things are declared.

sergiusens:
@kyrofa I am aware of the spec not guaranteeing order, but we are providing one regardless.

kyrofa:
I mean we aren't guaranteed to get them in order from pyyaml, so I'd rather not rely on it.

PR #2152, opened 2018-05-31
many: automatically redo step for specified part

kyrofa:

-----

Currently, if a step for a given part has already run, the snapcraft CLI won't run it again without it first being cleaned. For example, if the `pull` step of part `A` had already run, running `snapcraft pull A` would say that it had already run, and skip it. This PR resolves [LP: #1774260](https://bugs.launchpad.net/snapcraft/+bug/1774260) by changing this behavior so that, if a part is specified, that step for that part is cleaned and re-run.

### DIFF
--- a/snapcraft/internal/errors.py
+++ b/snapcraft/internal/errors.py
@@ -54,15 +54,15 @@ class MissingStateCleanError(SnapcraftError):
 class StepOutdatedError(SnapcraftError):
 
     fmt = (
-        'Failed to reuse files from previous build: '
+        'Failed to reuse files from previous run: '
         'The {step.name!r} step of {part!r} is out of date:\n'
         '{report}'
-        'To continue, clean that part\'s {step.name!r} step, run '
+        'To continue, clean that part\'s {step.name!r} step by running '
         '`snapcraft clean {parts_names} -s {step.name}`.'
     )
 
-    def __init__(self, *, step, part,
-                 dirty_properties=None, dirty_project_options=None,
+    def __init__(self, *, step, part, dirty_properties=None,
+                 dirty_project_options=None, changed_dependencies=None,
                  dependents=None):
         messages = []
         if dirty_properties:
@@ -83,6 +83,14 @@ class StepOutdatedError(SnapcraftError):
             messages.append(
                 'The {} project {} to have changed.\n'.format(
                     humanized_options, pluralized_connection))
+        if changed_dependencies:
+            dependencies = [
+                d['name'] for d in changed_dependencies]
+            messages.append('{} changed: {}\n'.format(
+                formatting_utils.pluralize(
+                    dependencies, 'A dependency has',
+                    'Some dependencies have'),
+                formatting_utils.humanize_list(dependencies, 'and')))
         if dependents:
             humanized_dependents = formatting_utils.humanize_list(
                 dependents, 'and')
--- a/snapcraft/internal/lifecycle/_clean.py
+++ b/snapcraft/internal/lifecycle/_clean.py
@@ -36,55 +36,50 @@ def _reverse_dependency_tree(config, part_name):
     return dependents
 
 
-def _clean_part_and_all_dependents(part_name, step, config, staged_state,
-                                   primed_state):
-    # Obtain the reverse dependency tree for this part. Make sure all
-    # dependents are cleaned.
+def _clean_part(part_name, step, config, staged_state, primed_state):
+    if step.next_step() is None:
+        template = 'Cleaning {step} step for {part}'
+    else:
+        template = 'Cleaning {step} step (and all subsequent steps) for {part}'
+
+    logger.info(template.format(step=step.name, part=part_name))
+    config.parts.clean_part(part_name, staged_state, primed_state, step)
+
+    # If we just cleaned the root of a dependency tree, we need to mark all
+    # its dependents as dirty so they require cleaning
+    return mark_dependents_dirty(part_name, step, config)
+
+
+def mark_dependents_dirty(part_name, cleaned_step, config):
     dependents = _reverse_dependency_tree(config, part_name)
     dependent_parts = {p for p in config.all_parts
                        if p.name in dependents}
-    for dependent_part in dependent_parts:
-        dependent_part.clean(staged_state, primed_state, step)
 
-    # Finally, clean the part in question
-    config.parts.clean_part(part_name, staged_state, primed_state, step)
+    dirty_part_names = set()
+    for part in dependent_parts:
+        if part.mark_dependency_change(part_name, cleaned_step):
+            dirty_part_names.add(part.name)
 
-
-def _verify_dependents_will_be_cleaned(part_name, clean_part_names, step,
-                                       config):
-    # Get the name of the parts that depend upon this one
-    dependents = config.parts.get_dependents(part_name)
-    additional_dependents = []
-
-    # Verify that they're either already clean, or that they will be cleaned.
-    if not dependents.issubset(clean_part_names):
-        for part in config.all_parts:
-            if part.name in dependents and not part.is_clean(step):
-                humanized_parts = formatting_utils.humanize_list(
-                    dependents, 'and')
-                additional_dependents.append(part_name)
-
-                logger.warning(
-                    'Requested clean of {!r} which requires also cleaning '
-                    'the part{} {}'.format(part_name,
-                                           '' if len(dependents) == 1 else 's',
-                                           humanized_parts))
+    return dirty_part_names
 
 
 def _clean_parts(part_names, step, config, staged_state, primed_state):
     if not step:
-        step = 'pull'
+        step = steps.next_step(None)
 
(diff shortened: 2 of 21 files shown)


snapcraft/internal/lifecycle/_clean.py
```diff
@@ -176,6 +171,7 @@ def clean(project_options, parts, step=None):
 
     config = project_loader.load_config()
 
+    # FIXME: SHOULDN'T THIS DO THE SAME IF STEP < stage?
```
sergiusens: it would be easy to parse if we had an ordered enum (a list, oh we do), but that index thing is not as easy to read. Feel free to change given that we might add more steps.

snapcraft/internal/lifecycle/_runner.py
```diff
@@ -152,83 +159,171 @@ def __init__(self, config, project_options):
...
-            if step == 'stage':
-                # XXX check only for collisions on the parts that have already
-                # been built --elopio - 20170713
-                pluginhandler.check_for_collisions(self.config.all_parts)
-            for part in parts:
-                if step not in self._steps_run[part.name]:
-                    self._run_step(step, part, part_names)
-                    self._steps_run[part.name].add(step)
+        with config.CLIConfig() as cli_config:
+            for current_step in common.COMMAND_ORDER[0:step_index]:
```
sergiusens: This CLIConfig context feels to broad like this, but I cannot think of an easier way using a context which would not require opening and closing all the time. Is loading outside of the context a better alternative here, oh right, the context closes it, bah.
This would look nicer if we had `defer` or maybe if it decorated `run` somehow ;-)

snapcraft/internal/lifecycle/_clean.py
kyrofa: Yeah I agree, that index thing works nicely but it's rough to read. I'll see if I can wrap it in at least a function or enum to improve readability.

snapcraft/internal/lifecycle/_runner.py
kyrofa: Wow, you just echoed the exact thought process I had when writing this :laughing: .

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -280,8 +280,14 @@ def get_dirty_report(self, step):
             differing_options = state.diff_project_options_of_interest(
                 self._project_options)
 
-        if differing_properties or differing_options:
-            return DirtyReport(differing_properties, differing_options)
+        with contextlib.suppress(AttributeError):
```
sergiusens: can you add a note as to why this needs to be suppressed, as in the error scenario you are trying to avoid. I have my guesses, and I bet you know today, but I don't want to be scratching my head in the future :-)

snapcraft/internal/errors.py
```diff
@@ -53,35 +53,44 @@ def __init__(self, step):
...
         if dirty_properties:
-            humanized_properties = formatting_utils.humanize_list(
-                dirty_properties, 'and')
-            pluralized_connection = formatting_utils.pluralize(
-                dirty_properties, 'property appears',
-                'properties appear')
-            messages.append(
-                'The {} part {} to have changed.\n'.format(
-                    humanized_properties, pluralized_connection))
+                humanized_properties = formatting_utils.humanize_list(
```
sergiusens: why the over-indent here?

snapcraft/internal/errors.py
```diff
@@ -53,35 +53,44 @@ def __init__(self, step):
...
+                humanized_options = formatting_utils.humanize_list(
+                    dirty_project_options, 'and')
+                pluralized_connection = formatting_utils.pluralize(
+                    dirty_project_options, 'option appears',
+                    'options appear')
+                messages.append(
+                    'The {} project {} to have changed.\n'.format(
+                        humanized_options, pluralized_connection))
+        if changed_dependencies:
+                dependencies = [
```
sergiusens: ditto on indent

snapcraft/internal/lifecycle/_clean.py
```diff
@@ -36,56 +36,51 @@ def _reverse_dependency_tree(config, part_name):
...
-                                   primed_state):
-    # Obtain the reverse dependency tree for this part. Make sure all
-    # dependents are cleaned.
-    dependents = _reverse_dependency_tree(config, part_name)
-    dependent_parts = {p for p in config.all_parts
-                       if p.name in dependents}
-    for dependent_part in dependent_parts:
-        dependent_part.clean(staged_state, primed_state, step)
+def _clean_part(part_name, step, config, staged_state, primed_state):
+    if step == common.COMMAND_ORDER[-1]:
```
sergiusens: to avoid needing to look at COMMAND_ORDER over and over again, mind adding a quick note to the check?

snapcraft/internal/lifecycle/_clean.py
```diff
@@ -36,56 +36,51 @@ def _reverse_dependency_tree(config, part_name):
...
+        template = 'Cleaning {step} step for {part}'
+    else:
+        template = 'Cleaning {step} step (and all subsequent steps) for {part}'
 
-    # Finally, clean the part in question
+    logger.info(template.format(step=step, part=part_name))
     config.parts.clean_part(part_name, staged_state, primed_state, step)
 
+    # If we just cleaned the root of a dependency tree, we need to mark all
+    # its dependents as dirty so they require cleaning
```
sergiusens: this is overly pedantic :roll_eyes: , sorry, but sentences here should end with a period (like you did in other places)

snapcraft/internal/lifecycle/_clean.py
```diff
@@ -36,56 +36,51 @@ def _reverse_dependency_tree(config, part_name):
...
 
-    # Verify that they're either already clean, or that they will be cleaned.
-    if not dependents.issubset(clean_part_names):
-        for part in config.all_parts:
-            if part.name in dependents and not part.is_clean(step):
-                humanized_parts = formatting_utils.humanize_list(
-                    dependents, 'and')
-                additional_dependents.append(part_name)
+def mark_dependents_dirty(part_name, cleaned_step, config):
+    dependents = _reverse_dependency_tree(config, part_name)
```
sergiusens: hmm, this isn't new, but it is seems `_reverse_dependency_tree` belongs in config (as is structured today).

snapcraft/internal/lifecycle/_clean.py
```diff
@@ -36,56 +36,51 @@ def _reverse_dependency_tree(config, part_name):
...
+    dependent_parts = {p for p in config.all_parts
```
sergiusens: which makes this confusing given that we just got the reverse dependency tree for `part_name` and now we need to iterate again to get the _real_ dependents?

snapcraft/internal/lifecycle/_runner.py
```diff
@@ -152,83 +159,171 @@ def __init__(self, config, project_options):
...
+                        getattr(self, '_run_{}'.format(current_step))(part)
+                        self._steps_run[part.name].add(current_step)
+
+        self._create_meta(step, processed_part_names)
+
+    def _run_pull(self, part):
+        self._run_step(step='pull', part=part, progress='Pulling')
 
-        self._create_meta(step, part_names)
+    def _repull(self, part, hint=''):
```
sergiusens: seems you stopped using `hint` completely, I might be wrong, but I haven't seen any uses of it, why not just remove it?

snapcraft/internal/lifecycle/_clean.py
kyrofa: This is in #2159.

snapcraft/internal/pluginhandler/__init__.py
kyrofa: I've reworked this to check the state before trying to use it. This may be a case where clarity wins over pythonic-ness.

snapcraft/internal/errors.py
kyrofa: I didn't tell you I decided we should start using 8 spaces as our indentation?

snapcraft/internal/lifecycle/_clean.py
kyrofa: This has been completely reworked thanks to #2159.

snapcraft/internal/lifecycle/_clean.py
kyrofa: Fixed.

snapcraft/internal/lifecycle/_clean.py
kyrofa: Good call, I moved it over there.

snapcraft/internal/lifecycle/_clean.py
kyrofa: Yeah, that only gives us part names, and in order to clean we need the actual `PluginHandler` instances.

snapcraft/internal/lifecycle/_runner.py
kyrofa: It is still being used, actually.

PR #2155, opened 2018-06-04
build_providers: support for communicating with a qemu VM

sergiusens:
Add a qemu driver that can launch a qcow2 image, provide an
interface to execute commands over ssh and access over telnet
to the VMs control plane.

LP: #1774014

-----


snapcraft/internal/build_providers/_qemu/__init__.py
```diff
@@ -0,0 +1,15 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
```
kyrofa: No need to have a copyright notice here, I think.

snapcraft/internal/build_providers/_qemu/_qemu_driver.py
```diff
@@ -0,0 +1,175 @@
...
+from typing import List, Sequence
+
+import paramiko
+
+from snapcraft.internal.build_providers import errors
+
+
+logger = logging.getLogger(__name__)
+# Avoid getting paramiko logs which are overly verbose.
+logging.getLogger('paramiko').setLevel(logging.CRITICAL)
```
kyrofa: Even ERROR? Dang.

snapcraft/internal/build_providers/_qemu/_qemu_driver.py
```diff
@@ -0,0 +1,175 @@
...
+
+class QemuDriver:
+    """A driver to interact with qemu virtual machines."""
+
+    provider_name = 'qemu'
+    _PROJECT_DEV = 'project_dev'
+    _PROJECT_MOUNT = 'project_mount'
+
+    def __init__(self, *, ssh_username: str, ssh_key_file: str) -> None:
+        """Initialize a QemuCommand instance.
```
kyrofa: s/QemuCommand/QemuDriver/

snapcraft/internal/build_providers/_qemu/_qemu_driver.py
```diff
@@ -0,0 +1,175 @@
...
+            raise errors.ProviderCommandNotFound(command=provider_cmd)
+        self.provider_cmd = provider_cmd
+        # TODO detect collisions and make dynamic
+        self._telnet_port = 64444
+        # TODO detect collisions and make dynamic
+        self._ssh_port = 5555
+        self._ssh_username = ssh_username
+        self._ssh_key_file = ssh_key_file
+        self._qemu_proc = None  # type: subprocess.Popen
+        self._ssh_handle = paramiko.SSHClient()
```
kyrofa: Why "ssh_handle" instead of "ssh_client"?

snapcraft/internal/build_providers/_qemu/_qemu_driver.py
```diff
@@ -0,0 +1,175 @@
...
+        self._ssh_handle.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+
+    def launch(self, *, hda: str, qcow2_drives: Sequence,
+               project_9p_dev: str, ram: str=None,
+               enable_kvm: bool=True) -> None:
+        # TODO check for latest snapshot to launch for it instead of a cold
+        #      boot.
+        # TODO replace -nographic with a spinner and redirect to a log file.
+        cmd = [
+            'sudo', self.provider_cmd,
```
kyrofa: We should probably say what we're doing in case a password is required.

snapcraft/internal/build_providers/_qemu/_qemu_driver.py
```diff
@@ -0,0 +1,175 @@
...
+            '-device', 'virtio-9p-pci,fsdev={},mount_tag={}'.format(
+                self._PROJECT_DEV, self._PROJECT_MOUNT),
+            '-device', 'e1000,netdev=net0',
+            '-netdev', 'user,id=net0,hostfwd=tcp::{}-:22'.format(
+                self._ssh_port)]
+        for drive in qcow2_drives:
+            cmd.append('-drive')
+            cmd.append('file={},if=virtio,format=qcow2'.format(drive))
+        if enable_kvm:
+            cmd.append('-enable-kvm')
```
kyrofa: Is there a reason to make this optional? We could specify `accel=kvm:tcg` and they will be tried in order.

snapcraft/internal/build_providers/_qemu/_qemu_driver.py
```diff
@@ -0,0 +1,175 @@
...
+    def launch(self, *, hda: str, qcow2_drives: Sequence,
+               project_9p_dev: str, ram: str=None,
+               enable_kvm: bool=True) -> None:
+        # TODO check for latest snapshot to launch for it instead of a cold
+        #      boot.
+        # TODO replace -nographic with a spinner and redirect to a log file.
+        cmd = [
+            'sudo', self.provider_cmd,
+            '-m', ram, '-nographic',
+            '-monitor', 'telnet::{},server,nowait'.format(self._telnet_port),
```
kyrofa: I'm curious why we're using telnet here as opposed to stdio. Does it make things easier, even though we need to deal with the potential port conflicts?

snapcraft/internal/build_providers/_qemu/_qemu_driver.py
```diff
@@ -0,0 +1,175 @@
...
+            '-hda', hda,
+            '-fsdev', 'local,id={},path={},security_model=none'.format(
+                self._PROJECT_DEV, project_9p_dev),
+            '-device', 'virtio-9p-pci,fsdev={},mount_tag={}'.format(
+                self._PROJECT_DEV, self._PROJECT_MOUNT),
+            '-device', 'e1000,netdev=net0',
+            '-netdev', 'user,id=net0,hostfwd=tcp::{}-:22'.format(
+                self._ssh_port)]
```
kyrofa: Went doc-diving and found the `-nic` option, which I think covers both this and the above option in one go, while allowing you to drop some of the options (e.g. e1000 is the default):

>     -nic [tap|bridge|user|l2tpv3|vde|netmap|vhost-user|socket][,...][,mac=macaddr][,model=mn]
>
> This option is a shortcut for configuring both the on-board (default) guest NIC hardware and the host network backend in one go. The host backend options are the same as with the corresponding -netdev options below. The guest NIC model can be set with model=modelname. Use model=help to list the available device types. The hardware MAC address can be set with mac=macaddr.
>
> The following two example do exactly the same, to show how -nic can be used to shorten the command line length (note that the e1000 is the default on i386, so the model=e1000 parameter could even be omitted here, too):
>
>     qemu-system-i386 -netdev user,id=n1,ipv6=off -device e1000,netdev=n1,mac=52:54:98:76:54:32
>     qemu-system-i386 -nic user,ipv6=off,model=e1000,mac=52:54:98:76:54:32

snapcraft/internal/build_providers/_qemu/_qemu_driver.py
```diff
@@ -0,0 +1,175 @@
...
+    def _wait_for_ssh(self):
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        ssh_port_listening = False
+        while not ssh_port_listening:
+            time.sleep(1)
+            result = sock.connect_ex(('localhost', self._ssh_port))
+            logger.debug('Pinging for ssh availability: port check {}'.format(
+                result))
+            if result == 0:
+                sock.close()
```
kyrofa: Perhaps it would be best to use `with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as socket:` so we don't need to remember to do this.

snapcraft/internal/build_providers/_qemu/_qemu_driver.py
```diff
@@ -0,0 +1,175 @@
...
+                ssh_port_listening = True
+
+        ssh_service_ready = False
+        while not ssh_service_ready:
+            time.sleep(1)
+            try:
+                self._ssh_handle.connect('localhost',
+                                         port=self._ssh_port,
+                                         username=self._ssh_username,
+                                         key_filename=self._ssh_key_file)
```
kyrofa: Is the port check necessary with this step as well? I mean, this will obviously fail if the port isn't open, right?

snapcraft/internal/build_providers/_qemu/_qemu_driver.py
sergiusens: ERROR prints a horrible traceback.

snapcraft/internal/build_providers/_qemu/_qemu_driver.py
sergiusens: Yes, I actually feel like it would be a good idea to wrap every sudo command.

snapcraft/internal/build_providers/_qemu/_qemu_driver.py
sergiusens: just for when running for other architectures, like `qemu-system-aarch64`

snapcraft/internal/build_providers/_qemu/_qemu_driver.py
sergiusens: How do we talk to the backplane over stdio without mixing up common queries and output we do with snapcraft?

snapcraft/internal/build_providers/_qemu/_qemu_driver.py
sergiusens: I thought that was deprecated, I'll check again. What I have I think is what multipass is using as well. I don't see where you do the host forwarding here though.

snapcraft/internal/build_providers/_qemu/_qemu_driver.py
sergiusens: Ok, done at a higher level as this command is now redirected.

snapcraft/internal/build_providers/_qemu/_qemu_driver.py
sergiusens: I got confused with `-net` https://wiki.qemu.org/Documentation/Networking#The_legacy_-net_option
It seems `-nic` is not available on 18.04 yet though (sans the fact that it would miss the `hostfwd` part too.

kyrofa: > The host backend options are the same as with the corresponding -netdev options below

That means to me that `hostfwd` would work here as well. But yeah, it not being available is a deal-killer.

snapcraft/internal/build_providers/_qemu/_qemu_driver.py
kyrofa: Ah, okay.

PR #2158, opened 2018-06-07
rust plugin: fix cargo builds and run tests

tismith:
This patch changes it so that we run all instances of cargo using the build environment. Fixes cases where a snap cleanbuild will error out with rustc being unable to be found during the pull step.

This is the issue I asked for help with on the forum: [https://forum.snapcraft.io/t/problem-building-rust-based-snap/5776](https://forum.snapcraft.io/t/problem-building-rust-based-snap/5776)

Also adds 'cargo test' to the build step, to fail the build if the tests aren't passing.

Rust plugin unittests have been updated.

-----


sergiusens:
thanks for the contribution! At first glance this looks good. I have re-triggered travis to take into account the timeout.
I do have a question though, is the test target a conventional target (as it would be for makefiles) or built in to cargo itself? If the latter and the fallbacks are good when no tests are available, this should be good as is.

tismith:
> is the test target a conventional target (as it would be for makefiles) or built in to cargo itself? If the latter and the fallbacks are good when no tests are available, this should be good as is.

Yes, `cargo test` is a builtin part of cargo, and will pass if no tests are available. E.g. here's the output from a basically empty crate (i.e. just after running `cargo init --name testy` in an empty directory):

```
$ cargo test
   Compiling testy v0.1.0 (file:///Users/tobysmith/src/home/rust/test)
    Finished dev [unoptimized + debuginfo] target(s) in 3.4 secs
     Running target/debug/deps/testy-2006db7acae3b59a

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
$ echo $?
0
```

tismith:
Looks like 'cargo test' doesn't play nice with cross compiling. I'll rework the patch to not run the tests when cross compiling.

tismith:
Looks like someone else has run into this same problem, and worked around it like I did with a custom rust plugin.

[https://forum.snapcraft.io/t/snap-of-gutenberg/6080/3](https://forum.snapcraft.io/t/snap-of-gutenberg/6080/3)

snapcraft/plugins/rust.py
```diff
@@ -97,11 +97,24 @@ def __init__(self, name, options, project):
         self._rustup = os.path.join(self._rustpath, "rustup.sh")
         self._manifest = collections.OrderedDict()
 
+    def _test(self):
+        if self.project.is_cross_compiling:
```
sergiusens: it would be nice to issue a warning here about skipping the test execution since it is cross compiling.

sergiusens:
Travis seems to be confused
```
Cloning into 'snapcore/snapcraft'...
warning: Could not find remote branch rust-plugin-improvements to clone.
fatal: Remote branch rust-plugin-improvements not found in upstream origin
```

sergiusens:
I have re-created the PR on #2170

PR #2159, opened 2018-06-07
many: extract lifecycle ordering into own module

kyrofa:

-----

Currently the snapcraft CLI has step names and ordering hard-coded in a number of places, and those areas where it isn't hard-coded, a difficult-to-read index method is used. Simplify and extract this logic into a single module used by the others to determine lifecycle steps and ordering.

### DIFF
--- a/snapcraft/cli/lifecycle.py
+++ b/snapcraft/cli/lifecycle.py
@@ -17,7 +17,13 @@
 import click
 import os
 
-from snapcraft.internal import deprecations, lifecycle, lxd, project_loader
+from snapcraft.internal import (
+    deprecations,
+    lifecycle,
+    lxd,
+    project_loader,
+    steps,
+)
 from ._options import add_build_options, get_project_options
 from . import echo
 from . import env
@@ -62,7 +68,7 @@ def pull(ctx, parts, **kwargs):
         snapcraft pull my-part1 my-part2
 
     """
-    _execute('pull', parts, **kwargs)
+    _execute(steps.PULL, parts, **kwargs)
 
 
 @lifecyclecli.command()
@@ -77,7 +83,7 @@ def build(parts, **kwargs):
         snapcraft build my-part1 my-part2
 
     """
-    _execute('build', parts, **kwargs)
+    _execute(steps.BUILD, parts, **kwargs)
 
 
 @lifecyclecli.command()
@@ -92,7 +98,7 @@ def stage(parts, **kwargs):
         snapcraft stage my-part1 my-part2
 
     """
-    _execute('stage', parts, **kwargs)
+    _execute(steps.STAGE, parts, **kwargs)
 
 
 @lifecyclecli.command()
@@ -107,7 +113,7 @@ def prime(parts, **kwargs):
         snapcraft prime my-part1 my-part2
 
     """
-    _execute('prime', parts, **kwargs)
+    _execute(steps.PRIME, parts, **kwargs)
 
 
 @lifecyclecli.command()
@@ -175,12 +181,12 @@ def clean(parts, step, **kwargs):
     project_options = get_project_options(**kwargs)
     build_environment = env.BuilderEnvironmentConfig()
     if build_environment.is_host:
-        step = step or 'pull'
+        step = step or steps.next_step(step).name
         if step == 'strip':
             echo.warning('DEPRECATED: Use `prime` instead of `strip` '
                          'as the step to clean')
-            step = 'prime'
-        lifecycle.clean(project_options, parts, step)
+            step = steps.PRIME.name
+        lifecycle.clean(project_options, parts, steps.Step(step))
     else:
         config = project_loader.load_config(project_options)
         lxd.Project(project_options=project_options,
--- a/snapcraft/cli/lifecycle.py
+++ b/snapcraft/cli/lifecycle.py
@@ -180,13 +180,16 @@ def clean(parts, step, **kwargs):
     """
     project_options = get_project_options(**kwargs)
     build_environment = env.BuilderEnvironmentConfig()
-    if build_environment.is_host:
-        step = step or steps.next_step(step).name
+
+    if step:
         if step == 'strip':
             echo.warning('DEPRECATED: Use `prime` instead of `strip` '
                          'as the step to clean')
-            step = steps.PRIME.name
-        lifecycle.clean(project_options, parts, steps.Step(step))
+            step = 'prime'
+        step = steps.get_step_by_name(step)
+
+    if build_environment.is_host:
+        lifecycle.clean(project_options, parts, step)
     else:
         config = project_loader.load_config(project_options)
         lxd.Project(project_options=project_options,
--- a/snapcraft/cli/lifecycle.py
+++ b/snapcraft/cli/lifecycle.py
@@ -166,11 +166,11 @@ def pack(directory, output, **kwargs):
 @lifecyclecli.command()
 @add_build_options()
 @click.argument('parts', nargs=-1, metavar='<part>...', required=False)
-@click.option('--step', '-s',
+@click.option('--step', '-s', 'step_name',
               type=click.Choice(['pull', 'build', 'stage', 'prime', 'strip']),
               help='only clean the specified step and those that '
                    'depend on it.')
-def clean(parts, step, **kwargs):
+def clean(parts, step_name, **kwargs):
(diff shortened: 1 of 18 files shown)


snapcraft/internal/lifecycle/constants.py
```diff
@@ -15,5 +15,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 import os
 
+from snapcraft.internal import steps
+
 SNAPCRAFT_INTERNAL_DIR = os.path.join('snap', '.snapcraft')
-STEPS_TO_AUTOMATICALLY_CLEAN_IF_DIRTY = {'stage', 'prime'}
+STEPS_TO_AUTOMATICALLY_CLEAN_IF_DIRTY = {steps.STAGE, steps.PRIME}
```
sergiusens: would this look better if it were an attribute of the step?

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -29,7 +29,15 @@
...
-from snapcraft.internal import common, elf, errors, repo, sources, states
+from snapcraft.internal import (
+    common,
+    elf,
+    errors,
+    repo,
+    sources,
+    states,
+    steps,
+)
```
sergiusens: this import style really drives me nuts for some reason I cannot explain :stuck_out_tongue:

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -215,38 +225,33 @@ def _migrate_state_file(self):
...
     def next_step(self):
-        next_step = None
-        for step in reversed(common.COMMAND_ORDER):
-            if os.path.exists(
-                    states.get_step_state_file(self.plugin.statedir, step)):
-                break
-            next_step = step
-
-        return next_step
+        latest_step = None
```
sergiusens: if what we return is the next_step, should what we eval be the current_step instead of the latest_step? I am not sure why, but calling this latest_step confuses me, we had this conversation already as this is triggering things (I think we even opened a bug for this).

sergiusens: Ah, everything revolves around `latest_step` now it seems

snapcraft/internal/steps.py
```diff
@@ -0,0 +1,97 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from snapcraft.internal import errors
+
+
+class Step:
+    def __init__(self, name: str) -> None:
```
sergiusens: I am just going to throw this out, but what if instead of keeping a tight coupling with `STEPS` and those `index` calls to know where it lives we create the Step instance with knowledge of the next step and the previous step, kind of like a linked list and create a map of step names to the instances?

snapcraft/internal/lifecycle/constants.py
kyrofa: Ah, excellent idea!

snapcraft/internal/pluginhandler/__init__.py
kyrofa: Haha, this one was long enough I had to split it, sorry!

snapcraft/internal/pluginhandler/__init__.py
kyrofa: Nothing has changed here, it was just called "last step" which never felt quite right, so I changed it to "latest". Everything naturally revolves around which step was run most recently, since they build on each other. Do you prefer another name?

snapcraft/internal/steps.py
kyrofa: Yeah I thought about that, but ended up having two issues:

1. Order needs to be provided somehow, whether that's by providing a list giving the order or by hard-coding it into each `Step` instance. The former seemed simpler to understand and modify than the latter.
2. We need a way to know which step is first. The `STEPS` variable provides that knowledge (STEPS[0]).

It's an interesting idea, though. Let me try it real quick, I think it's a small change. Then we can see what we think.

kyrofa: Alright, just pushed it up. Let me know what you think.

kyrofa: We discussed offline, and I've moved it back to a list, but kept the same methods.

snapcraft/internal/lifecycle/constants.py
kyrofa: Done.

snapcraft/internal/lxd/_project.py
```diff
@@ -112,10 +112,10 @@ def clean(self, parts: List[str], step: str):
                     print('Deleting {}'.format(self._container_name))
                     subprocess.check_call([
                         'lxc', 'delete', '-f', self._container_name])
-            step = 'pull'
+            step = steps.PULL.name
```
sergiusens: I see two options here, but one thing first, we should rename the signature of this method to taking `step_name` instead of `step` or have it pass down an actual `Step`, where a possible value is `None` and move the DEPRECATED warning a level up when using `strip` passing it the appropriate step.

If using the former, I wouldn't get the name just to get the step again, but instead, only create the step for the case of `strip`

snapcraft/internal/pluginhandler/__init__.py
sergiusens: If you use autopep8 it will use the format I like :-P

snapcraft/internal/steps.py
```diff
@@ -0,0 +1,102 @@
...
+        if type(other) is type(self):
+            return self._order >= other._order
+
+        return NotImplemented
+
+    def __hash__(self):
+        return hash(self.name)
+
+    def __repr__(self):
+        return 'snapcraft.internal.steps.Step({!r})'.format(self.name)
```
sergiusens: It would of been nice to have some sort of `__modpath__` to not hardcode this, closest I know of is `__file__`, but it will look too hacky. If we ever move this module, this change will most likely go unnoticed, so why not remove the leading qualifiers?

snapcraft/internal/steps.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+# Step names and order are now maintained in a single place: right here. If
+# another step is needed, add it here, and insert it into STEPS in the proper
+# order.
+PULL = Step('pull', False)
+BUILD = Step('build', False)
+STAGE = Step('stage', True)
+PRIME = Step('prime', True)
```
sergiusens: maybe add `STRIP = PRIME` with a comment about it being there for backwards compat?

snapcraft/internal/steps.py
kyrofa: Yeah, you're right. I'm mostly going off of [python docs](https://docs.python.org/3/reference/datamodel.html#object.__repr__) here that say:

> [...] If at all possible, this should look like a valid Python expression that could be used to recreate an object with the same value

So I added the qualifiers. I agree that it's just asking to break though, and I question its usefulness anyway. This is really just to make test failures nice, so we can make it whatever we want. I'll remove the leading qualifiers.

snapcraft/internal/steps.py
kyrofa: That doesn't really help us since the step `.name` would still be `prime`. It's okay though, I think we can handle this a level above rather than pushing deprecation all the way down here.

snapcraft/internal/steps.py
kyrofa: Done.

snapcraft/internal/lxd/_project.py
kyrofa: Agreed, fixed by passing down a Step.

snapcraft/cli/lifecycle.py
```diff
@@ -180,13 +180,16 @@ def clean(parts, step, **kwargs):
     """
     project_options = get_project_options(**kwargs)
     build_environment = env.BuilderEnvironmentConfig()
-    if build_environment.is_host:
-        step = step or steps.next_step(step).name
+
+    if step:
```
sergiusens: it gets confusing a bit when `step` can be a `step_name` or an actual instance of `Step`.

kyrofa: Agreed, fixed by using an identifier decl for click to make it use a different parameter name.

sergiusens:
\o/

PR #2160, opened 2018-06-14
many: refactor snapcraft.yaml loading out of load_config

sergiusens:
In order to have separated logical bits and clear overlaying
communicated a logical separation is introduced between the
actual snapcraft loading in raw form and its processed/validated
form.

LP: #1775849

-----


snapcraft/project/_project_info.py
```diff
@@ -14,16 +14,86 @@
...
 
-    def __init__(self, data: Dict[str, Any]) -> None:
-        self.name = data['name']
-        self.version = data.get('version')
-        self.summary = data.get('summary')
-        self.description = data.get('description')
-        self.confinement = data.get('confinement')
-        self.grade = data.get('grade')
+    def __init__(self, *, snapcraft_yaml_file_path) -> None:
+        self.snapcraft_yaml_file_path = get_snapcraft_yaml()
```
kyrofa: This should probably be using `snapcraft_yaml_file_path`, no?

snapcraft/cli/_options.py
```diff
@@ -52,7 +52,13 @@ def _add_build_options(func):
     return _add_build_options
 
 
-def get_project_options(**kwargs):
+def get_project(**kwargs):
+    skip_snapcraft_yaml = kwargs.pop('skip_snapcraft_yaml', False)
```
kyrofa: It would be cleaner and more intuitive to include this as an actual kwarg so you don't need to pop it:

```python
def get_project(*, skip_snapcraft_yaml: bool=False, **kwargs):
    # ...
```

snapcraft/project/_project_info.py
```diff
@@ -14,16 +14,86 @@
...
+    except yaml.reader.ReaderError as e:
+        raise errors.YamlValidationError(
+            'Invalid character {!r} at position {} of {}: {}'.format(
+                chr(e.character), e.position + 1, yaml_file_path,
+                e.reason)) from e
+
+    return yaml_contents
+
+
+def get_snapcraft_yaml(base_dir=None):
```
kyrofa: I suggest putting this either in `snapcraft/project/__init__.py`, or in another module imported by `__init__.py`. Reasoning: it shouldn't be used here, but belongs in the `snapcraft/project` package somewhere for use elsewhere.

snapcraft/project/_project_info.py
sergiusens: err, yes!

snapcraft/project/_project_info.py
sergiusens: ok, which means the tests for it need to be moved as well

PR #2161, opened 2018-06-14
many: automatically detect dependency changes

kyrofa:

-----

Currently the only way parts know they're dirty from dependency changes is if the lifecycle tells them so. This is a very error-prone approach, since if the lifecycle forgets to tell the part, then it doesn't know it needs to be cleaned. Dependency changes should be automatically detected instead of being noted in code, which also simplifies the upcoming feature of detecting source changes.


kyrofa:
This needs another tweak. I'll re-propose.

PR #2165, opened 2018-06-20
many: automatically detect dependency changes

kyrofa:

-----

Currently the only way parts know they're dirty from dependency changes is if the lifecycle tells them so. This is a very error-prone approach, since if the lifecycle forgets to tell the part, then it doesn't know it needs to be cleaned. This PR fixes [LP: #1778081](https://bugs.launchpad.net/snapcraft/+bug/1778081) by automatically detecting changes instead of noting them in code, which also simplifies the upcoming feature of detecting source changes.

Note that this PR breaks backward state compatibility with #2152 (i.e. running on trees made dirty since then will break). This was judged to be acceptable based on the fact that it has not yet been contained in a release.

### DIFF
--- a/snapcraft/internal/lifecycle/__init__.py
+++ b/snapcraft/internal/lifecycle/__init__.py
@@ -19,4 +19,3 @@ from ._containers import containerbuild   # noqa
 from ._init import init                   # noqa
 from ._packer import pack                 # noqa
 from ._runner import execute              # noqa
-from ._status_cache import StatusCache    # noqa
--- a/snapcraft/internal/lifecycle/__init__.py
+++ b/snapcraft/internal/lifecycle/__init__.py
@@ -19,3 +19,4 @@ from ._containers import containerbuild   # noqa
 from ._init import init                   # noqa
 from ._packer import pack                 # noqa
 from ._runner import execute              # noqa
+from ._status_cache import StatusCache    # noqa
--- a/snapcraft/internal/lifecycle/_runner.py
+++ b/snapcraft/internal/lifecycle/_runner.py
@@ -189,7 +189,7 @@ class _Executor:
                     if dirty_report:
                         self._handle_dirty(
                             part, current_step, dirty_report, cli_config)
-                    elif self._cache.step_has_run(part, current_step):
+                    elif self._cache.has_step_run(part, current_step):
                         # By default, if a step has already run, don't run it
                         # again. However, automatically clean and re-run the
                         # step if all the following conditions apply:
@@ -209,7 +209,6 @@ class _Executor:
                     else:
                         getattr(self, '_run_{}'.format(
                             current_step.name))(part)
-                        self._cache.add_step_run(part, current_step)
 
         self._create_meta(step, processed_part_names)
 
@@ -247,20 +246,23 @@ class _Executor:
 
     def _run_step(self, *, step: steps.Step, part, progress, hint=''):
         common.reset_env()
-        prereq_parts = self.parts_config.get_dependencies(part.name)
+        all_dependencies = self.parts_config.get_dependencies(part.name)
 
+        # Filter dependencies down to only those that need to run the
+        # prerequisite step
         prerequisite_step = steps.get_dependency_prerequisite_step(step)
-        step_prereqs = {p for p in prereq_parts
-                        if self._cache.step_should_run(p, prerequisite_step)}
+        dependencies = {p for p in all_dependencies
+                        if self._cache.should_step_run(p, prerequisite_step)}
 
-        if step_prereqs:
-            prereq_names = {p.name for p in step_prereqs}
-            # prerequisites need to go all the way to the prerequisite step to
-            # be able to share the common assets that make them a dependency.
+        if dependencies:
+            dependency_names = {p.name for p in dependencies}
+            # Dependencies need to go all the way to the prerequisite step to
+            # be able to share the common assets that make them a dependency
             logger.info(
-                '{!r} has prerequisites that need to be {}d: {}'.format(
-                    part.name, prerequisite_step.name, ' '.join(prereq_names)))
-            self.run(prerequisite_step, prereq_names)
+                '{!r} has dependencies that need to be {}d: {}'.format(
+                    part.name, prerequisite_step.name, ' '.join(
+                        dependency_names)))
+            self.run(prerequisite_step, dependency_names)
 
         # Run the preparation function for this step (if implemented)
         preparation_function = getattr(
@@ -277,6 +279,9 @@ class _Executor:
 
         notify_part_progress(part, progress, hint)
         getattr(part, step.name)()
+
+        # We know we just ran this step, so rather than check, manually twiddle
+        # the cache
         self._cache.add_step_run(part, step)
         self.steps_were_run = True
 
@@ -309,7 +314,7 @@ class _Executor:
                     step=step, part=part.name, dirty_report=dirty_report)
 
         getattr(self, '_re{}'.format(step.name))(part, hint='({})'.format(
-            dirty_report.summary()))
+            dirty_report.get_summary()))
 
 
 def notify_part_progress(part, progress, hint='', debug=False):
--- a/snapcraft/internal/lifecycle/_runner.py
+++ b/snapcraft/internal/lifecycle/_runner.py
@@ -14,7 +14,6 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import collections
 import logging
 import os
 from subprocess import check_call
@@ -22,10 +21,9 @@ from tempfile import TemporaryDirectory
 from typing import Sequence
 
 import yaml
-from tabulate import tabulate
 
 import snapcraft
-from snapcraft import config, formatting_utils
+from snapcraft import config
 from snapcraft.internal import (
     common,
     errors,
@@ -38,7 +36,7 @@ from snapcraft.internal import (
 )
 from snapcraft.internal.cache import SnapCache
 from . import constants
-from ._clean import mark_dependents_dirty
+from ._status_cache import StatusCache
 
 
 logger = logging.getLogger(__name__)
@@ -58,8 +56,8 @@ def execute(step: steps.Step,
     and after is not in this set, an exception will be raised.
 
     :param str step: A valid step in the lifecycle: pull, build, prime or snap.
-    :param ProjectConfig: Fully loaded project (old logic moving either to
-                          Project or the PluginHandler).
+    :param project_config: Fully loaded project (old logic moving either to
+                           Project or the PluginHandler).
     :param list part_names: A list of parts to execute the lifecycle on.
     :raises RuntimeError: If a prerequesite of the part needs to be staged
                           and such part is not in the list of parts to iterate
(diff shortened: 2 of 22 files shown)


snapcraft/internal/lifecycle/_runner.py
```diff
@@ -299,36 +277,22 @@ def _run_step(self, *, step: steps.Step, part, progress, hint='',
 
         notify_part_progress(part, progress, hint)
         getattr(part, step.name)()
-        self._steps_run[part.name].add(step)
+        self._cache.add_step_run(part, step)
```
kyrofa: Add comments here

snapcraft/internal/pluginhandler/_dirty_report.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+class DirtyReport:
+    def __init__(self, *, dirty_properties=None, dirty_project_options=None,
+                 changed_dependencies=None):
+        self.dirty_properties = dirty_properties
+        self.dirty_project_options = dirty_project_options
+        self.changed_dependencies = changed_dependencies
+
+    def report(self):
```
kyrofa: Make these get_*

snapcraft/internal/lifecycle/_status_cache.py
```diff
@@ -0,0 +1,174 @@
...
+    """The StatusCache is a lazy caching interface for the status of parts."""
+
+    def __init__(self, config: _config.Config) -> None:
+        """Create a new StatusCache.
+
+        :param _config.Config config: Project config.
+        """
+        self.config = config
+        self._steps_run = dict()  # type: Dict[str, Set[steps.Step]]
+        self._dirty_reports = collections.defaultdict(dict)  # type: Dict[str, Dict[steps.Step, pluginhandler.DirtyReport]]  # noqa
```
kyrofa: Create type for this

snapcraft/internal/lifecycle/_status_cache.py
```diff
@@ -0,0 +1,174 @@
...
+                     step: steps.Step) -> None:
+        """Cache the fact that a given step has now run for the given part.
+
+        :param pluginhandler.PluginHandler part: Part in question.
+        :param steps.Step step: Step in question.
+        """
+        self._ensure_steps_run(part)
+        self._steps_run[part.name].add(step)
+
+    def step_has_run(self, part: pluginhandler.PluginHandler,
```
kyrofa: has_step_run

snapcraft/internal/lifecycle/_runner.py
```diff
@@ -260,29 +239,28 @@ def _restage(self, part, hint=''):
...
-            step=steps.PRIME, part=part, progress='Re-priming', hint=hint,
-            prerequisite_step=steps.PRIME)
+            step=steps.PRIME, part=part, progress='Re-priming', hint=hint)
 
-    def _run_step(self, *, step: steps.Step, part, progress, hint='',
-                  prerequisite_step: steps.Step=steps.STAGE):
+    def _run_step(self, *, step: steps.Step, part, progress, hint=''):
         common.reset_env()
-        prereqs = self.parts_config.get_prereqs(part.name)
+        prereq_parts = self.parts_config.get_dependencies(part.name)
```
sergiusens: so we had a nice conversation about `get_dependencies` being renamed, but the local var is still called `prereq_parts`?

snapcraft/internal/lifecycle/_runner.py
```diff
@@ -260,29 +239,28 @@ def _restage(self, part, hint=''):
...
 
-        step_prereqs = {p for p in prereqs
-                        if prerequisite_step not in self._steps_run[p]}
+        prerequisite_step = steps.get_dependency_prerequisite_step(step)
+        step_prereqs = {p for p in prereq_parts
```
sergiusens: ditto

snapcraft/internal/lifecycle/_runner.py
```diff
@@ -260,29 +239,28 @@ def _restage(self, part, hint=''):
...
+                        if self._cache.step_should_run(p, prerequisite_step)}
 
         if step_prereqs:
+            prereq_names = {p.name for p in step_prereqs}
```
sergiusens: ditto

snapcraft/internal/lifecycle/_status_cache.py
```diff
@@ -0,0 +1,174 @@
...
+        :param steps.Step step: Step in question.
+        :return: Whether or not step should run.
+        :rtype: bool
+
+        A given step should run if it:
+            1. Hasn't yet run
+            2. Is dirty
+            3. Either (1) or (2) apply to any earlier steps in its lifecycle
+        """
+        if (not self.step_has_run(part, step) or
```
sergiusens: in this use I can see why this may be nice to have in the pluginhandler itself.

snapcraft/internal/lifecycle/_status_cache.py
```diff
@@ -0,0 +1,174 @@
...
+                    for dependency in dependencies:
+                        # Make sure the prerequisite step of this dependency
+                        # has not run more recently than (or should run
+                        # _before_) this step.
+                        try:
+                            prerequisite_timestamp = dependency.step_timestamp(
+                                prerequisite_step)
+                        except errors.StepHasNotRunError:
+                            dependency_changed = True
+                        else:
```
sergiusens: I always have to lookup if `else` is for the `try` succeeding or no matching `except` :man_shrugging:

snapcraft/internal/lifecycle/_status_cache.py
```diff
@@ -0,0 +1,174 @@
...
+        if not self._dirty_reports[part.name]:
+            _del_key(self._dirty_reports, part.name)
+
+    def _ensure_steps_run(self, part: pluginhandler.PluginHandler) -> None:
+        if part.name not in self._steps_run:
+            self._steps_run[part.name] = _get_steps_run(part)
+
+    def _ensure_dirty_report(self, part: pluginhandler.PluginHandler,
+                             step: steps.Step) -> None:
+        if step not in self._dirty_reports[part.name]:
```
sergiusens: how about returning early if it is in and bumping everything below one level in?
It would make it obvious that there is no `else` case below

sergiusens: fwiw, this is the only scenario where I am mostly ok with multiple return scenarios, i.e.; prechecks.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -284,18 +277,23 @@ def get_dirty_report(self, step):
...
+                    dirty_properties=properties, dirty_project_options=options)
 
         return None
 
     def should_step_run(self, step, force=False):
         return force or self.is_clean(step)
 
+    def step_timestamp(self, step):
+        try:
+            return os.stat(
```
sergiusens: I can tell you like stairs :taunt: (emoji request :smile: )

snapcraft/internal/pluginhandler/_dirty_report.py
```diff
@@ -0,0 +1,97 @@
...
+                    dependencies, 'A dependency has',
+                    'Some dependencies have'),
+                formatting_utils.humanize_list(dependencies, 'and')))
+
+        return ''.join(messages)
+
+    def summary(self):
+        reasons = []
+
+        reason_count = 0
```
sergiusens: reasons_count?

snapcraft/internal/pluginhandler/_dirty_report.py
```diff
@@ -0,0 +1,97 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from snapcraft import formatting_utils
+
+
+class DirtyReport:
+    def __init__(self, *, dirty_properties=None, dirty_project_options=None,
```
sergiusens: types please

snapcraft/internal/pluginhandler/_dirty_report.py
```diff
@@ -0,0 +1,97 @@
...
+        if self.changed_dependencies:
+            reason_count += 1
+
+        if self.dirty_properties:
+            # Be specific only if this is the only reason
+            if reason_count > 1 or len(self.dirty_properties) > 1:
+                reasons.append('properties')
+            else:
+                reasons.append('{!r} property'.format(
+                    next(iter(self.dirty_properties))))
```
sergiusens: This confuses me and I couldn't fine a unit test; so given `dirty_properties={'foo', 'bar'}` (as in some test code below, this would reasons.append only `foo`? Is that the objective or am I missing something here?

kyrofa: Nah, this branch is only hit if there is a single property. This is just getting at the property in a manner that works for both lists and sets. You're absolutely right though, I'm missing tests for these, so I'll add those which should clarify further.

snapcraft/internal/lifecycle/__init__.py
```diff
@@ -19,3 +19,4 @@
 from ._init import init                   # noqa
 from ._packer import pack                 # noqa
 from ._runner import execute              # noqa
+from ._status_cache import StatusCache    # noqa
```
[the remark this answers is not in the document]
kyrofa: Fixed.

snapcraft/internal/lifecycle/_runner.py
kyrofa: Done.

snapcraft/internal/pluginhandler/_dirty_report.py
kyrofa: Done.

snapcraft/internal/lifecycle/_status_cache.py
kyrofa: Done.

snapcraft/internal/lifecycle/_status_cache.py
kyrofa: Done.

snapcraft/internal/lifecycle/_runner.py
kyrofa: Yeah good catch, fixed.

snapcraft/internal/lifecycle/_runner.py
kyrofa: Fixed as well.

snapcraft/internal/lifecycle/_runner.py
kyrofa: Fixed.

snapcraft/internal/lifecycle/_status_cache.py
kyrofa: Done, actually was able to bring it back two stops.

snapcraft/internal/pluginhandler/__init__.py
kyrofa: Yikes, fixed that a bit.

snapcraft/internal/pluginhandler/_dirty_report.py
kyrofa: Sure, done.

snapcraft/internal/pluginhandler/_dirty_report.py
kyrofa: Ah yes of course, fixed.

snapcraft/internal/pluginhandler/_dirty_report.py
kyrofa: Done.

kyrofa:
It seems rust is failing for every PR. Ignoring and merging.

PR #2167, opened 2018-06-23
many: detect local source changes

kyrofa:

-----

The snapcraft CLI doesn't notice changes to local sources once the parts have been pulled: one must clean and re-pull in order to take them into account. This PR resolves  [LP: #1583718](https://bugs.launchpad.net/snapcraft/+bug/1583718) by comparing local source timestamps to the pull state timestamp, and updating the pull (and all subsequent) steps as necessary.


sergiusens:
Hey, what a PR, it is quite big and lots of tests seem to be failing, so I'll leave it to you to fix them before moving forward with an actual review.

kyrofa:
> [...] lots of tests seem to be failing, so I'll leave it to you to fix them before moving forward with an actual review.

Yeah I had to move the XDG cache around, so integration tests are taking a few iterations. I'll ping you.

kyrofa:
Alright @sergiusens, this one should be good, now.

snapcraft/file_utils.py
```diff
@@ -106,18 +106,18 @@ def link_or_copy(source: str, destination: str,
...
             # remove it and try again.
             os.remove(destination)
             link_or_copy(source, destination, follow_symlinks)
         else:
-            _copy(source, destination, follow_symlinks)
+            copy(source, destination, follow_symlinks)
 
 
-def _link(source: str, destination: str, follow_symlinks: bool=False) -> None:
+def link(source: str, destination: str, follow_symlinks: bool=False) -> None:
```
sergiusens: now that they are public it would be nice to force kwargs and add some docstrings if you don't mind

snapcraft/file_utils.py
```diff
@@ -137,7 +137,7 @@ def _link(source: str, destination: str, follow_symlinks: bool=False) -> None:
         raise SnapcraftCopyFileNotFoundError(source)
 
 
-def _copy(source: str, destination: str, follow_symlinks: bool=False) -> None:
+def copy(source: str, destination: str, follow_symlinks: bool=False) -> None:
```
sergiusens: dito

snapcraft/internal/lifecycle/_runner.py
```diff
@@ -277,11 +285,18 @@ def _run_step(self, *, step: steps.Step, part, progress, hint=''):
 
         part = _replace_in_part(part)
 
+    def _run_step(self, *, step: steps.Step, part, progress, hint=''):
+        self._prepare_to_run(step=step, part=part)
```
sergiusens: maybe `_prepare_to_run_step`?

snapcraft/internal/lifecycle/_runner.py
```diff
@@ -277,11 +285,18 @@ def _run_step(self, *, step: steps.Step, part, progress, hint=''):
...
+
         notify_part_progress(part, progress, hint)
         getattr(part, step.name)()
 
         # We know we just ran this step, so rather than check, manually twiddle
         # the cache
+        self._step_complete(part, step)
+
+    def _step_complete(self, part, step):
```
sergiusens: `_complete_step` for parity?

snapcraft/internal/sources/_base.py
```diff
@@ -41,6 +41,36 @@ def __init__(self, source, source_dir, source_tag=None, source_commit=None,
...
+        """Update pulled source.
+
+        :raises RuntimeError: If this function is called before `check()`.
+        """
+        if not self._checked:
+            # This is programmer error
+            raise RuntimeError("source must be checked before it's updated")
+        self._update()
+
+    def _check(self, target: str):
```
sergiusens: why not just do the abc. thing here, you can leave the docstring, it will still be valid code.

kyrofa: Because then we'd need to add it to all sources, even those that don't actually support doing this.

snapcraft/internal/lifecycle/_runner.py
```diff
@@ -316,6 +331,28 @@ def _handle_dirty(self, part, step, dirty_report, cli_config):
...
+
+            # We know we just ran this step, so rather than check, manually
+            # twiddle the cache
+            self._step_complete(part, step)
+        else:
+            getattr(self, '_re{}'.format(step.name))(part, '({})'.format(
+                outdated_report.get_summary()))
+
 
 def notify_part_progress(part, progress, hint='', debug=False):
```
sergiusens: this is old, but what is the difference?

sergiusens: Just keep it green :smile:

snapcraft/internal/lifecycle/_runner.py
```diff
@@ -316,6 +331,28 @@ def _handle_dirty(self, part, step, dirty_report, cli_config):
         getattr(self, '_re{}'.format(step.name))(part, hint='({})'.format(
             dirty_report.get_summary()))
 
+    def _handle_outdated(self, part, step, outdated_report, cli_config):
```
sergiusens: same as handle_dirty, can this take a `dirty_action` as a parameter instead of passing down the entire object?

snapcraft/internal/lifecycle/_runner.py
```diff
@@ -202,10 +202,17 @@ def run(self, step: steps.Step, part_names=None):
                             getattr(self, '_re{}'.format(
                                 current_step.name))(part)
                         else:
-                            notify_part_progress(
-                                part,
-                                'Skipping {}'.format(current_step.name),
-                                '(already ran)')
+                            outdated_report = self._cache.get_outdated_report(
```
sergiusens: I have complications accepting this procedural flow, just the level of nesting.
How is `dirty_report` so different from `outdated_report`? How can we have a `dirty_report` if we haven't `has_step_run`? Do you think this can be shuffled a bit?

snapcraft/internal/lifecycle/_runner.py
sergiusens: With the introduction of the outdated concept, I think this is just another case of dirty. Can we add more precise qualifiers like `handle_dirty_definitions` and `handle_dirty_[local]_source`?

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -349,6 +371,21 @@ def pull(self, force=False):
         self._runner.pull()
         self.mark_pull_done()
 
+    def check_pull(self):
+        # Check to see if pull needs to be updated
+        state_file = states.get_step_state_file(
+            self.plugin.statedir, steps.PULL)
+
+        # Not all sources support checking for updates
+        with contextlib.suppress(NotImplementedError):
```
sergiusens: ok, I see the reason for raising now, still would be nice to have a different exception to avoid over catching (or check the payload) to not swallow exceptions like this.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -406,28 +443,48 @@ def prepare_build(self, force=False):
...
-            if directory == self.plugin.sourcedir:
-                snaps = glob(os.path.join(directory, '*.snap'))
-                if snaps:
-                    snaps = [os.path.basename(s) for s in snaps]
-                    return common.SNAPCRAFT_FILES + snaps
+        if not self.plugin.out_of_source_build:
+            if os.path.exists(self.plugin.build_basedir):
+                shutil.rmtree(self.plugin.build_basedir)
+
+            # FIXME: It's not necessary to ignore here anymore since it's now
```
sergiusens: wow, this is so old :sweat_smile:

snapcraft/internal/sources/_local.py
```diff
@@ -25,13 +26,79 @@
...
+                    else:
+                        self._updated_directories.add(relpath)
+
+        return (
+            len(self._updated_files) > 0 or
+            len(self._updated_directories) > 0)
+
+    def _update(self):
+        # First, copy the directories
+        for directory in self._updated_directories:
```
sergiusens: this feels a lot like the pluginhandlers "migration" code. Maybe we need some generalization there (in the future)

tests/fixture_setup/_fixtures.py
```diff
@@ -116,6 +116,13 @@ def setUp(self):
         patcher_dirs.start()
         self.addCleanup(patcher_dirs.stop)
 
+        self.useFixture(fixtures.EnvironmentVariable(
```
sergiusens: They are back!

tests/integration/__init__.py
```diff
@@ -93,40 +91,19 @@ def setUp(self):
         self.useFixture(temp_cwd_fixture)
         self.path = temp_cwd_fixture.path
 
-        self.useFixture(fixtures.EnvironmentVariable(
-            'XDG_CONFIG_HOME', os.path.join(self.path, '.config')))
-        self.useFixture(fixtures.EnvironmentVariable(
-            'XDG_DATA_HOME', os.path.join(self.path, 'data')))
+        # Use a separate path for XDG dirs, or changes there may be detected as
+        # source changes.
```
sergiusens: kudos on the code comment, makes sense

snapcraft/internal/lifecycle/_runner.py
kyrofa: Let me give you my thought process. A step (for a specific part) can be in one of two states:

1. Not run (we call this "clean")
2. Run, which is further broken down into sub-states:
   2.1. Complete
   2.2. Needs to be cleaned and run again (we call this "dirty", i.e. it needs to be cleaned)
   2.3. Needs to be run again, but does not need to be cleaned (in this PR we call this "outdated", i.e. it needs to be updated)

Both 2.2 and 2.3 are _caused_ by something, which needs to be recorded somewhere so we can tell the user what it was. We do that with the `{Dirty,Outdated}Report` classes. This allows us to isolate concerns, separating the "what makes this {dirty,outdated}" from "oh, this is {dirty,outdated}, I need to take appropriate action." If we combine these reports into one, we have a few issues:

1. We would need to come up with a new name, since "dirty" no longer works (i.e. the solution is not to clean it anymore).
2. Concerns leak. Now lifecycle needs to know "Oh, the project properties changed, okay, I know I need to clean" or "Oh, an earlier step happened, I know I need to update". This spread of knowledge seems error-prone.

(1) isn't a big deal, we can figure that one out. The word "outdated" could be used to refer to both situations, I suppose. (2) we can probably solve by keeping things similar to how they are today, but have a master class composed of both a dirty and outdated report, and take action appropriately. That at least gives us a single report to deal with in the `PluginHandler`'s API and thus the lifecycle, but other than that doesn't buy us much. Do you like that idea better?

snapcraft/internal/pluginhandler/__init__.py
kyrofa: Sure thing.

snapcraft/file_utils.py
kyrofa: This API needs to be compatible with copy2, so the only kwarg we can force is follow_symlinks, but I agree that it's an improvement.

snapcraft/internal/lifecycle/_runner.py
kyrofa: I'm afraid I'm not clear on what you're asking me to do, here.

snapcraft/internal/sources/_local.py
```diff
@@ -25,13 +26,79 @@
 
 class Local(Base):
 
+    def __init__(self, *args, copy_function=file_utils.link_or_copy, **kwargs):
```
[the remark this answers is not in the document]
kyrofa: Yeah, right [here](https://github.com/snapcore/snapcraft/pull/2167/files#diff-660a35e421f136410b0d3e5168c7a62cR477). Didn't want to duplicate functionality.

snapcraft/internal/sources/_local.py
kyrofa: There are definitely some similarities in that they both operate on directories and files, but the pluginhandler does some special stuff as well.

snapcraft/internal/pluginhandler/__init__.py
kyrofa: Fixed with a custom exception.

snapcraft/internal/lifecycle/_runner.py
kyrofa: There, that's quite a bit better. Thoughts?

snapcraft/internal/lifecycle/_runner.py
kyrofa: Agreed, done.

snapcraft/internal/lifecycle/_runner.py
kyrofa: Went with `_prepare_step` so we have `_{prepare,run,complete}_step`.

snapcraft/file_utils.py
kyrofa: Done.

snapcraft/file_utils.py
kyrofa: Fixed, thank you for catching the lack of docs!

snapcraft/internal/lifecycle/_runner.py
sergiusens: I did not say combine them, I just meant that "dirty" and "outdated" are easier to confuse and that we should narrow the scope of what it means down.

Your thought process is from the point of view of the action wrt lifecycle and my mind is trending towards the state of things (why is it dirty/outdated) whilst the report comes from a component that does not determine the future actions.

These two terms lead to confusions, is this dirty because it is outdated (so far that has been our logic, right?). The _what_ is outdated is what I am asking to more clearly state in the method calls, checks and class names.

But yeah, by no means, unless it gets converted into the state class merge these two.

snapcraft/internal/lifecycle/_runner.py
```diff
@@ -184,41 +184,56 @@ def run(self, step: steps.Step, part_names=None):
...
         self._create_meta(step, processed_part_names)
 
+    def _handle_step(self, requested_part_names: Sequence[str],
+                     part: pluginhandler.PluginHandler,
+                     requested_step: steps.Step,
+                     current_step: steps.Step, cli_config) -> None:
+        # If this step hasn't yet run, all we need to do is run it
+        if not self._cache.has_step_run(part, current_step):
+            getattr(self, '_run_{}'.format(current_step.name))(part)
+            return
```
sergiusens: instead of multiple returns here, it seems this could be handled very well with `elif` and a final `else`.

snapcraft/internal/sources/errors.py
```diff
@@ -72,8 +76,19 @@ def __init__(self, expected, calculated):
...
     fmt = ('The {deb_file} used does not contain valid data. '
            'Ensure a proper deb file is passed for .deb files '
            'as sources.')
+
+
+class SourceUpdateUnsupportedError(SnapcraftSourceError):
+
+    fmt = (
+        'Failed to update source: '
+        "{source!s} sources don't support updating"
```
sergiusens: Can you add a `.` to end the sentence please?

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -249,6 +250,32 @@ def is_clean(self, step):
         except errors.NoLatestStepError:
             return True
 
+    def is_outdated(self, step):
+        """Return true if the given step needs to be updated (no cleaning)."""
```
sergiusens: Can you clarify this doc string (what is in between parens).

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -249,6 +250,32 @@ def is_clean(self, step):
...
+
+        return self.get_outdated_report(step) is not None
+
+    def get_outdated_report(self, step: steps.Step):
+        """Return an OutdatedReport class describing why step is outdated.
```
sergiusens: Can you add a follow up paragraph on what it means to be outdated? Simil for what it means to be dirty in its counterpart.

snapcraft/internal/pluginhandler/_outdated_report.py
```diff
@@ -0,0 +1,63 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from snapcraft import formatting_utils
+from snapcraft.internal import steps
+
+
+class OutdatedReport:
+    """The OutdatedReport class explains why a given step is outdated."""
```
sergiusens: can you further expand in a paragraph what it means to be outdated (conditions that trigger it)? Simil for DirtyReport please.

snapcraft/internal/lifecycle/_runner.py
sergiusens: If you are really stuck with the names, I have added a couple more comments that would satisfy me. My intention was that from the names of variables and classes it was crystal clear what it meant without having to read the code, I can compromise with a bit more documentation under the _internal_ class names and methods called.

kyrofa: I agree that this should be cleaned up, but as we discussed on the call, let's see if we can get this PR in a decent state with docs and then take another pass (in another PR) at improving it further.

snapcraft/internal/lifecycle/_runner.py
kyrofa: We're fetching reports before checking them, which is heavy to do upfront if we don't actually need to, so it doesn't fit particularly well within an if/elif unless we want to nest within `else`s.

snapcraft/internal/sources/errors.py
kyrofa: Done.

snapcraft/internal/pluginhandler/__init__.py
kyrofa: Done, though mostly by directing the reader to `get_outdated_report()`.

snapcraft/internal/pluginhandler/__init__.py
kyrofa: Done, for both outdated and dirty.

snapcraft/internal/pluginhandler/_outdated_report.py
kyrofa: Done, for both outdated and dirty.

snapcraft/internal/lifecycle/_status_cache.py
```diff
@@ -34,8 +35,8 @@ def __init__(self, config: _config.Config) -> None:
         """
         self.config = config
         self._steps_run = dict()  # type: Dict[str, Set[steps.Step]]
-        self._outdated_reports = collections.defaultdict(dict)  # type: _Report
-        self._dirty_reports = collections.defaultdict(dict)  # type: _Report
+        self._outdated_reports = collections.defaultdict(dict)  # type: _OutdatedReport  # noqa
```
sergiusens: why noqa or why noqa with no qualifier, and if it is to prevent a line break, I'd rather change the line limit to 99 :wink:
Use your parens until then :laughing:

kyrofa: Fixed!

sergiusens:
atta boy :smile:

PR #2171, opened 2018-06-26
{catkin,python} plugin: support cleaning

kyrofa:

-----

Both the Catkin and Python plugins utilize the Pip class in order to clean their pull steps (specifically, they tell the Pip class to clean its packages). However, when cleaning the pull step, Python has already been removed from the installdir, which means the Pip class cannot be instantiated, causing an error about being unable to find Python.

It's reasonable to ask Pip to clean any fetched packages, and doing so doesn't require Python. This PR fixes LP: [#1778716](https://bugs.launchpad.net/snapcraft/+bug/1778716) by lazily looking for Python only when it's required instead of upon class instantiation.

### DIFF
--- a/snapcraft/plugins/_python/_pip.py
+++ b/snapcraft/plugins/_python/_pip.py
@@ -120,20 +120,32 @@ class Pip:
                                            staging or part's install area.
         """
         self._python_major_version = python_major_version
-        self._part_dir = part_dir
         self._install_dir = install_dir
         self._stage_dir = stage_dir
 
-        self._python_package_dir = os.path.join(
-            self._part_dir, 'python-packages')
+        self._python_package_dir = os.path.join(part_dir, 'python-packages')
         os.makedirs(self._python_package_dir, exist_ok=True)
 
-        self._python_command = get_python_command(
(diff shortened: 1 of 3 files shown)


snapcraft/plugins/_python/_pip.py
```diff
@@ -159,20 +171,20 @@ def _ensure_pip_installed(self):
         if not self._is_pip_installed():
             logger.info('Fetching and installing pip...')
 
-            real_python_home = self._python_home
+            real_python_home = self.__python_home
```
sergiusens: doesn't this have an extra _?

sergiusens: ah, this is weird :-P

kyrofa: Yeah it looks a tad off, but it maintains current behavior: "Ignore whatever you think you know, use this one. Okay, now you can go back to normal."

sergiusens: Yeah, it is made clear in the comments below, but not as I read line by line until I reached the end of it :smile:

PR #2172, opened 2018-06-26
many: add shellcheck to static tests

kyrofa:

-----

Also ensure all scripts pass. **Note:** part of the modification here involves autopkgtests, so we should probably run those once Travis is green.


bin/snapcraftctl
```diff
@@ -14,17 +14,29 @@
 # snapcraft). To that end, this script will use the interpreter defined within
 # the SNAPCRAFT_INTERPRETER environment variable.
 
+# Shell quote arbitrary string by replacing every occurrence of '
```
sergiusens: How is this related to shellcheck?

demos/gradle/gradlew
```diff
@@ -19,12 +20,17 @@ while [ -h "$PRG" ] ; do
         PRG=`dirname "$PRG"`"/$link"
     fi
 done
+# shellcheck disable=SC2006
```
sergiusens: if you are going to add these comments, why not just change them to `$()`?

sergiusens: Unless there is a way to skip the whole file without modifying it.

kyrofa: This file was automatically generated, so I'd rather not actually try to fix all its numerous issues or I'll have to pretend I understand what this script is doing and test it :stuck_out_tongue: . I'll hard-code a skip for it.

kyrofa: Done.

PR #2173, opened 2018-06-26
tests: create basic integration test spread infrastructure

kyrofa:

-----

This PR resolves [LP: #1779419](https://bugs.launchpad.net/snapcraft/+bug/1779419) by starting the process of reimplementing integration tests in spread. It proves the infrastructure out by reimplementing the Go plugin test from the integration "snapd" and "plugins" suites as a Spread suite.


tests/spread/run/go/task.yaml
```diff
@@ -0,0 +1,6 @@
+summary: build and run go snap
```
sergiusens: is this here just to avoid pushd and a trap with popd?

kyrofa: You mean as opposed to putting the snap in some sort of src/ directory? I suppose, just seemed to make sense to have it all at the same level unless there were multiple snaps.

spread.yaml
```diff
@@ -1,39 +1,91 @@
...
+  apt install -y snapd
+
+  # If $SNAPCRAFT_CHANNEL is defined, install snapcraft from that channel.
+  # Otherwise, look for it in /snapcraft/.
+  if [ -z "$SNAPCRAFT_CHANNEL" ]; then
+      if stat /snapcraft/*.snap 2>/dev/null; then
+          snap install --classic --dangerous /snapcraft/*.snap
+      else
+          echo "Expected a snap to exist in /snapcraft/. Did you forget to "
+          echo "define \$SNAPCRAFT_CHANNEL?"
```
sergiusens: if your intention was to install from the store you should set SNAPCRAFT_CHANNEL.

tests/spread/run/go/task.yaml
kyrofa: > Sounds reasonable. Shouldn't this test live in spread/run-go-hello for easier discovery though?

Are you expecting only a single Spread suite once it's fully implemented? I was expecting multiple, but it's difficult to know for sure without having an all-encompassing view of our integration tests, which I don't currently have. I was just focusing on eliminating an integration suite at a time. I kinda like the idea of the basic "hello" type sanity checks to be in their own suite, but I don't have strong feelings about it.

> Is the other suite going to live in spread/build/go-hello?

Following on the last question: Which other suite?

> Can we make a task depend on another? If so, can we just do the build one and then have this run one just do the installation and execution part?

No, not that I know of.

spread.yaml
kyrofa: Sure, done.

tests/spread/run/go/task.yaml
sergiusens: > Are you expecting only a single Spread suite once it's fully implemented? I was expecting multiple
Multiple are fine; but these tests I mention are too similar, one just builds on top of the other. If we draw the global picture in individual bug-tasks it might be easier to have this discussion as I have no idea what the direction is (we only have this https://bugs.launchpad.net/snapcraft/+bug/1779419 AFAIK)

> Following on the last question: Which other suite?

https://github.com/snapcore/snapcraft/blob/master/tests/integration/plugins/test_go_plugin.py#L30

>> Can we make a task depend on another? If so, can we just do the build one and then have this run one just do the installation and execution part?

> No, not that I know of.

Then let's do it with environment variables and stop short or continue if set.

I would also like to change the approach, instead of the meta bug that exists now that "builds on" to finally complete it, let's get individual migration bugs, start with moving the integration tests specific to plugins (as individual bugs/PRs) and leave these for last (which should just be a matter of environment check and just doing more on what was already done). This way, you can also gain knowledge of the global picture of what are in our integration tests.

kyrofa: > Then let's do it with environment variables and stop short or continue if set.

So run two tests, one that does X, and another that does both X and Y? Why not just write one test that does X and Y? I think that fits well with trying to move over the entire plugin suite, and essentially merge the plugin and snapd suite into just a "plugin" suite.

sergiusens: > So run two tests, one that does X, and another that does both X and Y? It would be easy to write using spread variants, but seems inefficient since it runs X twice: why not just write one test that does X and Y? I think that fits well with the changed approach: try to move over the entire plugin suite, and essentially merge the plugin and snapd suite into just a "plugin" suite.

Yeah, that is what I have been trying to express, which is why I proposed:
- dependencies
- environment variable

I think we just agreed on this in some form, and yes, most of the snapd suite is just a repeat of what is elsewhere, _except_ it installs. I never understood why Leo did it this way and since we both consider this not the best approach now that we know everything involved, I say we just get rid of the past and move forward.

kyrofa:
Alright @sergiusens just pushed another take on this. It combines the Go plugin and snapd integration test into a single spread test, but does so in a way that still allows for the use of shellcheck as well as some structure. Thoughts?

kyrofa:
@sergiusens I've re-implemented this as a new suite specific to the Go plugin, instead. Unfortunately this model means we'll be adding a new suite for each plugin, but that's probably workable.

sergiusens:
I approve, but would like to see a new/different/re-purposed bug for this specific task possible

PR #2176, opened 2018-07-05
[WIP] tests: use pytest as test runner

kyrofa:

-----

Don't rewrite unittest classes to use pytest, but simply start using pytest as the runner, which gives us the ability to use xdist, running tests in parallel.

This PR halves the unit test run time on my local machine. The savings won't be that dramatic on Travis, but it should help. Integration test time savings are unknown, but should be non-negligible as well. The other changes here are a side effect of using pytest: removal of some mocks that messed with it, no longer relying on a specific cwd, etc.


sergiusens:
Yeah, I like pytest, we just have this niggle
```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

marcelotrevisani:
If some testing sometime is freezing, I would suggest to mark that as flaky. After that would be good to create a task to solve that

kyrofa:
> Just a suggestion here, it would be good if you add pytest-xdist. It's a plugin for pytest to run your tests in parallel, it may save some time in your CI.

Thanks @marcelotrevisani, but we [already are](https://github.com/snapcore/snapcraft/pull/2176/files#diff-64f0cb1baad9d62832796d4bd16a9a53R14). It's one of the reasons we're looking to switch. The long-running test is due to our usage of testscenarios, which unittest turned into separate tests, but pytest runs as one long test. We're still playing about with this.

marcelotrevisani:
Could you please tell me which test is that? I may take a look, just for fun  :)
I have some experience with ``pytest``, I may help you with that

kyrofa:
@marcelotrevisani please help yourself! We use testscenarios in a number of places, but the really long-running ones are the ones in `tests/unit/lifecycle/test_order.py`.

marcelotrevisani:
Hi @kyrofa , I just saw that the test was created using unittests. And I also saw some stuffs like
```
dict(part="main", step=steps.STAGE),
                    dict(part="dependent", step=steps.PULL),
                    dict(part="dependent", step=steps.BUILD),
                    dict(part="dependent", step=steps.STAGE),
                    dict(part="nested-dependent", step=steps.PULL),
                    dict(part="nested-dependent", step=steps.BUILD),
                    dict(part="nested-dependent", step=steps.STAGE),
                    dict(part="main", step=steps.PRIME),
                    dict(part="dependent", step=steps.PRIME),
                    dict(part="nested-dependent", step=steps.PRIME)
```

So, you will select some tests to execute depending in which steps you are, right!?
How cool will be if you mark your tests with pytest and those tests which you want to run you just have to pass this information to pytest. 

More information about that you can find here
https://docs.pytest.org/en/latest/example/markers.html

If you want to fully change unittest to pytest you also can use this tool [unittest2pytest](https://github.com/pytest-dev/unittest2pytest)

PR #2177, opened 2018-07-09
pluginhandler: use uname from the host

kyrofa:

-----

Currently, when recording the manifest from the `PluginHandler`, the snapcraft CLI runs `uname` using `common.run_output`, which is made for running binaries out of parts or the staging area. That's never desired in this case, so this PR fixes LP: [#1780825](https://bugs.launchpad.net/snapcraft/+bug/1780825) by using `subprocess` directly to always run `uname` from the host.

### DIFF
--- a/snapcraft/internal/pluginhandler/__init__.py
+++ b/snapcraft/internal/pluginhandler/__init__.py
@@ -572,9 +572,22 @@ class PluginHandler:
     def _get_machine_manifest(self):
         # Use subprocess directly here. common.run_output will use binaries out
         # of the snap, and we want to use the one on the host.
-        uname = subprocess.check_output(['uname', '-srvmpio'])
+        try:
+            output = subprocess.check_output(['uname', '-srvmpio'])
+        except subprocess.CalledProcessError as e:
+            logger.warning(
+                "'uname' exited with code {}: unable to record machine "
+                "manifest".format(e.returncode))
+            return {}
+
+        try:
+            uname = output.decode(sys.getfilesystemencoding()).strip()
+        except UnicodeEncodeError:
+            logger.warning("Could not decode output for 'uname' correctly")
+            uname = output.decode('latin-1', 'surrogateescape').strip()
+
         return {
-            'uname': uname.decode(sys.getfilesystemencoding()).strip(),
+            'uname': uname,
             'installed-packages': repo.Repo.get_installed_packages(),
             'installed-snaps': repo.snaps.get_installed_snaps()
         }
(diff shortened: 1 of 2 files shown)


snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -569,8 +570,11 @@ def mark_build_done(self):
             scriptlet_metadata=self._scriptlet_metadata[steps.BUILD]))
 
     def _get_machine_manifest(self):
+        # Use subprocess directly here. common.run_output will use binaries out
+        # of the snap, and we want to use the one on the host.
+        uname = subprocess.check_output(['uname', '-srvmpio'])
```
sergiusens: We will need a specific error for this now (it was catched by the general wrapper until this change).

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -569,8 +570,11 @@ def mark_build_done(self):
...
         return {
-            'uname': common.run_output(['uname', '-srvmpio']),
+            'uname': uname.decode(sys.getfilesystemencoding()).strip(),
```
sergiusens: We will also need to catch unicode decode errors and use a surrogate on errors.

kyrofa: Not sure how uname would cause this, but seems harmless. I added the extra case.

kyrofa:
> My suggestion would be to create a module/class for uname that is self contained (like the multipass module/class).

That feels a bit premature/overkill for running a single command in a single location. I opted for simply making it more robust in-place, let me know if you still want it extracted.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -569,8 +570,24 @@ def mark_build_done(self):
             scriptlet_metadata=self._scriptlet_metadata[steps.BUILD]))
 
     def _get_machine_manifest(self):
+        # Use subprocess directly here. common.run_output will use binaries out
+        # of the snap, and we want to use the one on the host.
+        try:
+            output = subprocess.check_output(['uname', '-srvmpio'])
```
sergiusens: mind using uname's long options for clarity and purpose

kyrofa: Sure, done.

sergiusens:
Thank you for the added clarity. My network is bonkers, but tests seem to fail, feel free to land if no major changes take place to fix them :smile:

PR #2184, opened 2018-07-30
cli: add inspect subcommand

kyrofa:

-----

There are three features that would be useful to snapcraft CLI developers that are not supported today:

  1. A quick-look summary of the current project state.
  2. A way to determine which lifecycle step of which part was most recently completed.
  3. A way to determine which part provided a specific file within the staging or priming area.

This PR resolves [LP: #1759700](https://bugs.launchpad.net/snapcraft/+bug/1759700) by adding support for all three of these use-cases by introducing a new, hidden subcommand called `inspect`.

  1. `snapcraft inspect` will print a quick summary of the current project state.
  2. `snapcraft inspect --latest-step` will print the part, step, and timestamp of the latest step that was run.
  3. `snapcraft inspect --provides <path>` will print the part that provided the path.

Default to human-readable output, but also support a `--json` to print json for programmatic consumption.


snapcraft/cli/parts.py
```diff
@@ -68,3 +76,178 @@ def search(ctx, query):
...
+
+    By default, this command will print a tabulated summary of the current state of the
+    project.
+    """
+    count = 0
+    for option in (provides, latest_step):
+        if option:
+            count += 1
+
+    if count > 1:
```
sergiusens: I wonder if it is worth expanding click like in this last comment on this issue https://github.com/pallets/click/issues/257
Totally not a requirement for this PR, just a thought.

kyrofa:
> There seems to be a problem when running on a not yet run project

Whoops, sorry, that was silly. Fixed.

> For the json output at least we would need to get the working directory for that step too.

Sure, also fixed.

snapcraft/cli/parts.py
kyrofa: I think this is simple enough not to warrant more code, but if we find we need this functionality elsewhere, it may be worth doing.

snapcraft/internal/errors.py
```diff
@@ -185,6 +185,18 @@ class SnapcraftPartMissingError(SnapcraftError):
     )
 
 
+class NoSuchFileError(SnapcraftError):
```
sergiusens: the error message seems to specific for this class' name. How about `NoSuchFileInPartError`.
These messages as with ToolError are getting a bit too un-namespaced btw. We should fix that.

snapcraft/internal/errors.py
```diff
@@ -632,3 +644,34 @@ class InvalidMountinfoFormat(SnapcraftError):
...
+        "Failed to find part that provides path: {path!r} is not in the "
+        "staging or priming area.\n"
+        "Ensure the path is in the staging or priming area and try again."
+    )
+
+    def __init__(self, path):
+        super().__init__(path=path)
+
+
+class UntrackedFileError(SnapcraftInspectError):
```
sergiusens: this is specific to parts too

snapcraft/internal/errors.py
```diff
@@ -632,3 +644,34 @@ class InvalidMountinfoFormat(SnapcraftError):
...
+
+    fmt = "No known parts provided {path!r}. It may have been provided by a scriptlet."
+
+    def __init__(self, path):
+        super().__init__(path=path)
+
+
+class NoStepsRunError(SnapcraftInspectError):
```
sergiusens: ok, I see the pattern, we should prepend SnapcraftInspect to the class names (or put them in their own inspect package with most of the logic in the cli

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -126,24 +126,27 @@ def __init__(
 
     def get_pull_state(self) -> states.PullState:
         if not self._pull_state:
-            self._pull_state = states.get_state(self.plugin.statedir, steps.PULL)
+            self._pull_state = cast(states.PullState, self.get_state(steps.PULL))
```
sergiusens: why is this necessary?

kyrofa: Because this PR adds static checks to the return type of `get_state`. This now means casting if it's a subclass.

snapcraft/cli/parts.py
```diff
@@ -68,3 +76,188 @@ def search(ctx, query):
...
+                latest_step = step
+                latest_timestamp = timestamp
+
+    if not latest_part:
+        raise errors.NoStepsRunError()
+
+    return (latest_part, latest_step, latest_timestamp)
+
+
+def _lifecycle_status(config):
```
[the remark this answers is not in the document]
kyrofa: Done.

PR #2188, opened 2018-08-02
tests: support running spread suite in autopkgtests

kyrofa:

-----

The spread suite should run as part of the release process. Add an autopkgtest suite to run it.


TESTING.md
```diff
@@ -149,18 +149,30 @@ For manual exploratory testing, the team has one mac machine available.
...
 
-It's possible to select only one of the suites, with:
 
-    adt-run --unbuilt-tree . --apt-upgrade --setup-commands="apt install squashfuse -y" --testname=integrationtests --- lxd ubuntu:xenial
+### How to run on Bionic
+
+The easiest way is to use a LXC container. From the root of the project, run:
+
+    sudo apt install autopkgtest
+    autopkgtest . -U -- lxd ubuntu:xenial
```
sergiusens: autopkgtest instead of adt becuase the name changed in the archive?

debian/tests/integrationtests-spread
```diff
@@ -0,0 +1,45 @@
+#!/bin/sh
+
+set -ex
+
+# required for the debian adt host
+mkdir -p /etc/systemd/system/snapd.service.d/
+if [ "${http_proxy:-}" != "" ]; then
+	cat <<-EOF | tee /etc/systemd/system/snapd.service.d/proxy.conf
+		[Service]
+		Environment=http_proxy=$http_proxy
```
sergiusens: would this be a harmless unit if $http_proxy is not defined?

debian/tests/integrationtests-spread
```diff
@@ -0,0 +1,45 @@
...
+echo ubuntu:ubuntu | chpasswd
+sed -i 's/\(PermitRootLogin\|PasswordAuthentication\)\>.*/\1 yes/' /etc/ssh/sshd_config
+systemctl reload sshd.service
+
+# Spread will only build with recent go
+snap install --classic go
+
+# and now run spread against localhost
+# shellcheck disable=SC1091
+. /etc/os-release
```
sergiusens: maybe "source" instead of "."?

spread.yaml
```diff
@@ -66,18 +154,26 @@ prepare: |
 
   apt install -y snapd
 
-  # If $SNAPCRAFT_CHANNEL is defined, install snapcraft from that channel.
-  # Otherwise, look for it in /snapcraft/.
-  if [ -z "$SNAPCRAFT_CHANNEL" ]; then
-      if stat /snapcraft/*.snap 2>/dev/null; then
-          snap install --classic --dangerous /snapcraft/*.snap
+  if [ "$SNAPCRAFT_PACKAGE_TYPE" = "deb" ]; then
```
sergiusens: maybe not now, but this might be nice to have in helper scripts to be able to use shellcheck? Or did we discuss this already and am I going on in circles

debian/tests/integrationtests-spread
kyrofa: It wouldn't be defined at all (see the `if` statement above).

debian/tests/integrationtests-spread
kyrofa: `source` is a bashism, this will be using dash.

spread.yaml
kyrofa: I agree 100%, sadly, see [this comment](https://github.com/snapcore/snapcraft/pull/2188/files/909bef12452e1022faf774ed3c609a937c577679#diff-3c11e56e5f7f82b0f352d0fe1851a107R129) for why it has to be this way.

TESTING.md
kyrofa: Indeed, there is no `adt-run` now.

debian/tests/integrationtests-spread
sergiusens: duh, I completely missed that; my suggestion was going to be to gate it just like you did!

kyrofa:
Judging from our snapd suite, I suspect armhf will fail due to an inability to install snaps.

kyrofa:
Haha, spread tests pass, plugin suite fails.

sergiusens:
as expected @kyrofa the one that fails though is the cmake one which is now in master as a spread test, seems like a missing gcc build-packages entry satisfied by the migrated autotools one

kyrofa:
@sergiusens indeed, it requires g++, which I fixed in the migration to spread. The only other fix is to add openssh-server to the test deps so armhf can actually be used by spread, so we'll see how it is now.

sergiusens:
@kyrofa maybe just disable on armhf for now, let's run it on the other green arches to make sure it is isolated before making the change though (i386 and arm64), this might be a discrepancy between native run versus container run scenario that we are seeing.

kyrofa:
Jeez, finally. \o/

PR #2189, opened 2018-08-02
project_loader: add basic template support

kyrofa:

-----

This PR resolves [LP: #1785109](https://bugs.launchpad.net/snapcraft/+bug/1785109) by adding support for templates. Templates have the ability to add YAML snippets to apps, YAML snippets to parts, and indeed add entire parts to a project. Don't introduce any actual templates, just the groundwork for them.


setup.py
```diff
@@ -148,6 +149,10 @@
                 "share/snapcraft/libraries",
                 ["libraries/" + x for x in os.listdir("libraries")],
             ),
+            (
+                "share/snapcraft/template-data",
+                ["templates/" + x for x in os.listdir("templates")],
```
sergiusens: I wanted to get rid of this mechanism and do something a bit more pkg_resource driven as this fails when creating a source package to push over to pypi (unless you edit that other file https://github.com/snapcore/snapcraft/blob/master/MANIFEST.in). If possible, we should get away from that early.

snapcraft/cli/templates.py
```diff
@@ -0,0 +1,68 @@
...
+    pass
+
+
+@templatecli.command()
+def templates(**kwargs):
+    """List available templates."""
+
+    template_names = os.listdir(common.get_templatesdir())
+    if not template_names:
+        click.echo("No templates supported")
```
sergiusens: maybe even tie it to the base? So then different people running it don't get confused with, hey I get some, what about you? Or is this going to list everything regardless of base?

snapcraft/cli/templates.py
```diff
@@ -0,0 +1,68 @@
...
+        return
+
+    # Wrap the output depending on terminal size
+    width = common.get_terminal_width()
+    for line in common.format_output_in_columns(
+        sorted(template_names), max_width=width
```
sergiusens: alternatively we should list what bases it supports here

snapcraft/internal/project_loader/_templates.py
```diff
@@ -0,0 +1,218 @@
...
+def apply_templates(yaml_data: Dict[str, Any]) -> Dict[str, Any]:
+    """Apply all templates.
+
+    :param dict yaml_data: Loaded, unprocessed snapcraft.yaml
+    :returns: Modified snapcraft.yaml data with templates applied
+    """
+    # Don't modify the dict passed in
+    yaml_data = copy.deepcopy(yaml_data)
+
+    # Get the base being used for this project (defaults to "core")
```
sergiusens: core is not a base :-)
I would default to nothing and if needed, introduce this later

sergiusens: IOW, let's tie this to the use of bases.

sergiusens: Remember that today, you can run snapcraft on arch and if no stage packages or build-packages are introduced, chances are, it might just work. This would introduce another vector of confusion.

setup.py
kyrofa: It's probably beyond the scope of this PR to change this, so I'll edit the manifest here, but yeah, let's change that.

snapcraft/internal/project_loader/_templates.py
kyrofa: I disagree on this point, for a few reasons:

1. Remote parts are a problem today, with a ton of assumed knowledge and documentation issues. They are hard to use, and templates will be better. They are also very widely used (desktop helpers in particular).
2. Templates are the only option for folks using remote parts that move to bases (as remote parts are not supported). By not allowing folks to transition to templates before they transition to using bases, we make the move to using bases quite a lot more work than is actually necessary.
3. Running snapcraft today on things other than Ubuntu is not something we support without bases. Making decisions based on how snapcraft runs on other distros today sounds a bit like support :wink: .

Unless supporting templates before bases actually makes our lives harder (I don't think it does), I think it's a good idea. We can chat about this more in-person, though.

snapcraft/cli/templates.py
kyrofa: I agree, let me try some different formatting, here.

snapcraft/internal/project_loader/_templates.py
kyrofa: We chatted about this in person, and the big concern is simply that the core16 base, once it exists, cannot guarantee that it will continue building on trusty as well, which supporting templates without bases would entail (i.e. defaulting to core16 templates). As a result, I've started raising an exception if not base is specified.

templates/README.md
```diff
@@ -0,0 +1,26 @@
+# How to write a template
+
+Create a new directory alongside this README. The name of the directory is the
+name of the template. Within that directory, create a `template.yaml` that
+follows this syntax:
+
+```yaml
+# Keys in the root of the yaml specify the supported bases
+core16:
```
sergiusens: change this to core18, it is the new and flashy (and currently existing) base.

snapcraft/internal/project_loader/_templates.py
```diff
@@ -121,6 +116,10 @@ def load_template(template_name: str) -> Dict[str, Any]:
 
 
 def _find_template(base: str, template_name: str) -> Dict[str, Any]:
+    # A bade is required in order to use templates, so raise an error if not specified.
```
sergiusens: bade?

templates/README.md
```diff
@@ -0,0 +1,26 @@
...
+  apps:
+    '*':
+      # YAML snippet you want to add to apps using this template (e.g. `plugs`)
+
+  parts:
+    '*':
+      # Yaml snippet you want to add to parts in the `snapcraft.yaml` (e.g.
+      # `after`)
+
+    # You can also put whatever parts here you want, and they will be put
```
sergiusens: whatever parts you want here?

templates/README.md
```diff
@@ -0,0 +1,26 @@
...
+    # alongside the other parts in the `snapcraft.yaml`. Take note of the
+    # `$SNAPCRAFT_TEMPLATES_DIR` variable that is pointing to the directory
+    # containing this README.
+    my-template-part:
+      plugin: dump
+      source: $SNAPCRAFT_TEMPLATES_DIR/my-template-name/src
```
sergiusens: ok, this might be the reason. Maybe this should be more encapsulated to the actual template instead?

snapcraft/internal/project_loader/_templates.py
kyrofa: Well duh. What would YOU call it? :wink: 

Fixed.

templates/README.md
kyrofa: Yikes, awkward wording indeed. Fixed.

templates/README.md
kyrofa: Well, remember that templates can only alter the `snapcraft.yaml`. They're not particularly special beasts, we can't really do any magic there or we kill the ability for someone to copy/paste them into their local `snapcraft.yaml` to tweak.

sergiusens: oh, I mean SNAPCRAFT_TEMPLATE_DIR specific to the template and not the entire namespace

kyrofa:
ftp.gnu.org is down right now, which caused the autotools spread tests to fail, but I just tested them locally, so I'm merging.

PR #2193, opened 2018-08-03
file_utils: get_tool_path to always return an absolute path

sergiusens:
Change get_tool_path to always return an absolute path and fail if
the required command cannot be found.

The unit tests have been adapted to not be path dependent and tests
for retrieving environment dependent paths have been added.

LP: #1785320

-----


snapcraft/internal/errors.py
```diff
@@ -295,6 +295,17 @@ def __init__(self, message):
         super().__init__(message=message)
 
 
+class ToolMissingError(SnapcraftError):
+
+    fmt = (
+        "A tool snapcraft depends on could not be found: {command_name!r}.\n"
+        "Ensure the tool is installed and available, and try again."
```
kyrofa: This is a workaround, but in reality, for all the use-cases of this function so far (e.g. xdelta3, mksquashfs, etc.) all of these tools should ALWAYS be found, and if they're not, it's a bug we should know about. Worth adding "please log a bug" to this?

tests/unit/commands/test_push.py
```diff
@@ -79,64 +77,6 @@ def test_push_a_snap(self):
         )
         mock_upload.assert_called_once_with("basic", self.snap_file)
 
-    def test_push_a_snap_running_from_snap(self):
```
kyrofa: This test seems useful, is it duplicated somewhere?

sergiusens:
That last change enables this -> https://asciinema.org/a/hGSsKKhmPSPA92l8LyVDBAiBg

tests/unit/commands/test_push.py
kyrofa: We clarified this offline: the tests covering get_tool_path here cover most of what is removed here. We need a suite that runs in docker to cover everything, which we can introduce later.

sergiusens:
spread 17.10 tests failed, ignoring those.

PR #2194, opened 2018-08-06
plugins: do not install go debs if go snap is used

mvo5:
Right now the go plugin will always pull in the golang-go
build-package. This is problematic when the project also has
a "go" build-snap. The deb package comes before in PATH and it
seems like there is no easy way to override this. This PR fixes
it by simply not installing the golang-go deb if there is a
build-snap already defined.


snapcraft/plugins/go.py
```diff
@@ -93,7 +93,8 @@ def get_pull_properties(cls):
 
     def __init__(self, name, options, project):
         super().__init__(name, options, project)
-        self.build_packages.append("golang-go")
+        if "go" not in self.build_snaps:
```
sergiusens: You might want to account for channels too, `go/<channel>` is the syntax, so maybe something like this would be better:
```python
if "go" not in self.build_snaps and not any((s.startswith("go/") for s in self.build_snaps)):
    self.build_packages.append("golang-go")
```

I will take care of bases and such in a future branch.

mvo5: Ok, I updated the code to support "go/foo" and "go=foo" in the go plugin.

snapcraft/plugins/go.py
```diff
@@ -93,7 +93,10 @@ def get_pull_properties(cls):
 
     def __init__(self, name, options, project):
         super().__init__(name, options, project)
-        self.build_packages.append("golang-go")
+        if ("go" not in self.build_snaps and
+            not any((s.startswith("go/") for s in self.build_snaps)) and
+            not any((s.startswith("go=") for s in self.build_snaps))):
```
sergiusens: is this agreed upon with Gustavo? I'd rather introduce this later as you will need to fix our internal API for this to work too (snapcraft.internal.repo.snap) and also add a deprecation message for use of /

sergiusens:
And another thing, adding a build-snap might mean we need to mock that internal API as it does checks to see if the snap exists and this will fail during unittests while building the deb

mvo5:
Fwiw, the unit test error looks unrelated:
```
$lxc exec "$name" -- apt update
E: Type 'curity' is not known on line 50 in source list /etc/apt/sources.list
E: The list of sources could not be read.
```

sergiusens:
We will need smarter logic to handle a situation when one part installs the snap and another the deb, but this works for now

PR #2195, opened 2018-08-07
Stop expecting the deprecation message in test.

cprov:
Out of the blue we started getting this weird error in the 'Pre-merging' tests:

https://travis-ci.org/snapcore/snapcraft/builds/412858741

The proposed fix stop matching a chunk of opaque data before the revisions table. I suspect it was there for coping with the 'deprecation' message when the `history` command is used, but I can't tell for sure if anything related changed in snapcraft.


cprov:
@sergiusens thanks for clarifying that it was not about the deprecation message (which does not affect the command in question, `revisions` is not deprecated). The opaque chunk of text was for the debug messages that are not disabled.

PR #2196, opened 2018-08-07
build providers: better injection logic

sergiusens:

-----


kyrofa:
Conflicts here.

snapcraft/internal/build_providers/_base_provider.py
```diff
@@ -53,6 +47,9 @@ def __init__(self, *, project, echoer) -> None:
             self.snap_filename = "{}_{}.snap".format(
                 project.info.name, project.deb_arch
             )
+        self.provider_project_dir = os.path.join(
+            BaseDirectory.xdg_data_home, "snapcraft", "projects", project.info.name
```
kyrofa: Perhaps `os.path.join(xdg.BaseDirectory.save_data_path("snapcraft"), "projects", project.info.name)` ?

snapcraft/internal/build_providers/_snap.py
```diff
@@ -0,0 +1,296 @@
...
+        # - if there is no snapcraft on the host, used to determine if
+        #   a refresh is needed (instead of a fresh install).
+        # - if there is a snapcraft on the host, used to determine if
+        #   re-injecting is required.
+        snap_info = snap_repo.get_local_snap_info()
+
+        if not snap_repo.installed and self._latest_revision is None:
+            op = _SnapOp.INSTALL
+        elif not snap_repo.installed and self._latest_revision is not None:
+            op = _SnapOp.REFRESH
```
kyrofa: I'm a little confused on this one-- how can a refresh be what's needed if it's not installed?

snapcraft/internal/build_providers/_snap.py
```diff
@@ -0,0 +1,296 @@
...
+        return self.__install_cmd
+
+
+_STORE_ASSERTION = [
+    "account-key",
+    "public-key-sha3-384=BWDEoaqyr25nF5SNCvEv2v7QnM9QsfCc0PBMYD_i2NGSQ32EF2d4D0hqUel3m8ul",
+]
+
+
+class SnapInjector:
```
kyrofa: Mind adding some sphinx docs here?

snapcraft/internal/build_providers/_snap.py
```diff
@@ -0,0 +1,296 @@
...
+    def __init__(
+        self,
+        *,
+        snap_dir: str,
+        registry_filepath: str,
+        runner,
+        snap_dir_mounter,
+        snap_dir_unmounter,
+        file_pusher
```
kyrofa: Would like to see some types specified for these callbacks.

snapcraft/internal/build_providers/_snap.py
```diff
@@ -0,0 +1,296 @@
...
+
+            self._file_pusher(
+                source=assertion_file.name, destination=assertion_file.name
+            )
+            self._runner(["sudo", "snap", "ack", assertion_file.name])
+
+    def _get_latest_revision(self, snap_name):
+        self._load_registry()
+        try:
+            return self._registry_data[snap_name][-1]["revision"]
```
kyrofa: Rather than having an explicit `_load_registry()` function as well as a private `_registry_data` attribute, this usage seems better suited to simple lazy evaluation of the property (or a `get_registry()` function, if you prefer) to encapsulate the loading and returning of the registry.

snapcraft/internal/build_providers/_snap.py
```diff
@@ -0,0 +1,296 @@
...
+    def _load_registry(self):
+        if self._registry_filepath is None or self._registry_data:
+            return
+        if not os.path.exists(self._registry_filepath):
+            return
+
+        with open(self._registry_filepath) as registry_file:
+            self._registry_data = yaml.load(registry_file)
+
+    def _dump_registry(self):
```
kyrofa: Take it or leave it, but I feel like this should be called `_save_registry`.

snapcraft/internal/build_providers/_snap.py
```diff
@@ -0,0 +1,296 @@
...
+            self._runner(["sudo", "snap", "ack", assertion_file.name])
+
+    def _get_latest_revision(self, snap_name):
+        self._load_registry()
+        try:
+            return self._registry_data[snap_name][-1]["revision"]
+        except (IndexError, KeyError):
+            return None
+
+    def _add_latest_revision(self, snap_name: str, snap_revision: str) -> None:
```
kyrofa: This function name isn't a good indicator of what this does. It's not at all similar to the similarly-named `add()` function, and it doesn't do anything with the "latest revision": it takes the one you give it. Perhaps `_record_revision()` would be better?

snapcraft/internal/build_providers/_base_provider.py
sergiusens: Sure, why not? Is there a difference?

snapcraft/internal/build_providers/_snap.py
sergiusens: If you are running from the deb, it will not be installed, if the environment has already been created it may have a revision installed.
snap_repo is for the host, the operation for the build environment.

snapcraft/internal/build_providers/_snap.py
sergiusens: yes, I forgot!

snapcraft/internal/build_providers/_snap.py
sergiusens: I sort of started to dislike properties when I do not need them (like replacing an already in use attribute). They are hard to mock and this clearly does something under the hood. Also,
> Explicit is better than implicit" is in direct conflict with using a property. (It looks like a simple assignment, but it calls a function).

After that philosophical point, if you still want me to change, I will :-)

sergiusens: Oh, but I am down with get_registry, I had that originally and was passing around too many params for my taste, but I prefer that indeed

snapcraft/internal/build_providers/_snap.py
sergiusens: ok, was going with parity with yaml and json.load/dump :-)

snapcraft/internal/build_providers/_snap.py
sergiusens: sure, this was mostly a parity thing, but I even got that wrong I think :-)

snapcraft/internal/build_providers/_snap.py
sergiusens: I will document with code comments

snapcraft/internal/build_providers/_snap.py
sergiusens: done

snapcraft/internal/build_providers/_snap.py
sergiusens: there is no comprehensive way to define a Callable that takes optional arguments nor one that takes kwargs https://docs.python.org/3.5/library/typing.html#callable

snapcraft/internal/build_providers/_snap.py
sergiusens: I like the symmetry on this one and not sure I understand the change now that I am re-reading.

snapcraft/internal/build_providers/_snap.py
```diff
@@ -0,0 +1,347 @@
...
+            self._set_data()
+        return self.__revision
+
+    def get_install_cmd(self) -> List[str]:
+        if self.__install_cmd is None:
+            self._set_data()
+        return self.__install_cmd
+
+
+_STORE_ASSERTION = [
```
kyrofa: Can we move this up to the top of the file, please?

snapcraft/internal/build_providers/_snap.py
```diff
@@ -0,0 +1,347 @@
...
+        :param runner: a callable which can run commands in the build environment.
+        :param snap_dir_mounter: a callable which can mount the directory where snaps live
+                                 on the host into the build environment. This callable
+                                 must mount into snap_dir.
+        :param snap_dir_unmounter: a callable which can unmount snap_dir from the environment.
+        :param file_pusher: a callable that can push file from the host into the build
+                            environment.
+        """
+
+        self._snaps = []  # type: List["_SnapManager"]
```
kyrofa: Does this type need to be quoted? It's defined earlier in this file.

sergiusens: only because I moved things around, probably not

snapcraft/internal/build_providers/_snap.py
sergiusens: sure

PR #2207, opened 2018-08-09
kernel plugin: install .config as config-$kernelversion

piso77:
In Ubuntu classic we ship the kernel config as config-$kernelversion in /boot, and the same we do for kernel snaps available in the store: they contain the kernel config file as config-$kernelversion in the root of the kernel snap.

Unfortunately the kernel plugin never shipped the .config used to build the kernel, so fix it by installing it as config-$kernelversion in the root of the freshly generated kernel snap.

This fixes [LP1786190](https://bugs.launchpad.net/snapcraft/+bug/1786190)

Signed-off-by: Paolo Pisati <p.pisati@gmail.com>

-----

### DIFF
--- a/snapcraft/plugins/kernel.py
+++ b/snapcraft/plugins/kernel.py
@@ -531,3 +531,8 @@ class KernelPlugin(kbuild.KBuildPlugin):
                 os.path.join(self.installdir, "lib", "firmware"), self.installdir
             )
         os.rmdir(os.path.join(self.installdir, "lib"))
+        # install .config as config-$version
+        config = "config-{}".format(self.kernel_release)
+        config_path = os.path.join(self.installdir, config)
+        dot_config_path = self.get_config_path()
+        os.link(dot_config_path, config_path)


snapcraft/plugins/kernel.py
```diff
@@ -531,3 +531,8 @@ def do_install(self):
                 os.path.join(self.installdir, "lib", "firmware"), self.installdir
             )
         os.rmdir(os.path.join(self.installdir, "lib"))
+        # install .config as config-$version
+        config = "config-{}".format(self.kernel_release)
+        config_path = os.path.join(self.installdir, config)
```
sergiusens: how about installing it under `(self.installdir, "snap", config)`? This way it sits together with every other manifesty thing.

ogra1: it sits next to the Symbols file with the above code and also matches where the config lives in the deb based kernel snaps, if we change it to go to the other dir, we should change the deb based snap builds alongside for consistency.

PR #2216, opened 2018-08-15
spread tests: keep sources local

kyrofa:

-----

We've been having a few issues with FTP servers going down that we're using in spread tests. Stop using remote sources in spread tests, opting instead to carry tar archives of the sources in-tree.


tests/spread/plugins/autotools/snaps/autotools-hello/snap/snapcraft.yaml
```diff
@@ -18,4 +18,4 @@ build-packages: [gcc]
 parts:
   make-project:
     plugin: autotools
-    source: http://ftp.gnu.org/gnu/hello/hello-2.10.tar.gz
+    source: hello-2.10.tar.gz
```
sergiusens: why not unpack the tar here and use `source: .`?
this would make for a bit more of a pleasant deb.

kyrofa: Because each one of these cause about a million static test failures, and our ignore list will balloon. Thoughts?

sergiusens: fine then, if we get push back we can try and do it

PR #2218, opened 2018-08-16
yaml: replace yaml.safe_load() with CSafeLoader

chipaca:
that is, yaml.safe_load(...) -> yaml.load(..., Loader=yaml.CSafeLoader)
This fixes [lp:1787373](https://bugs.launchpad.net/snapcraft/+bug/1787373).

the tests fail, but they're failing for me on master as well (I'm probably missing something).


snapcraft/storeapi/_client.py
```diff
@@ -11,7 +11,7 @@
 
 # Set urllib3's logger to only emit errors, not warnings. Otherwise even
 # retries are printed, and they're nasty.
-logging.getLogger(requests.packages.urllib3.__package__).setLevel(logging.ERROR)
+logging.getLogger(requests.packages.urllib3.__package__).setLevel(logging.DEBUG)
```
sparkiegeek: stray debugging accidentally left in?

sparkiegeek:
Overall this looks good, but I think it's worth extracting a shared `load_yaml` function so that this is easier to identify and have future naive `yaml.load` additions spotted in review

chipaca:
I've been bothered by this all the way to physio and back: I should've called out @sparkiegeek for pointing this whole thing  out to me. Thanks dude.

snapcraft/storeapi/_client.py
chipaca: I have no idea what you're talking about :innocent:

chipaca:
@sparkiegeek agreed, a helper function would be better. I'll wait for a go-ahead from the snapcraft gods though.

sergiusens:
Thanks @chipaca 
You might want to add yaml-dev to the snapcraft.yaml here as well. If it uses ctypes you might want to look at patches/ctypes_init.diff too
WRT the loader mechanism that @sparkiegeek mentions, we have indeed already talked about doing so to also wrap the yaml representation bits (used to setup ordered writes for OrderedDict among others) into a single method call instead of doing it globally. If you have the time, that would be a welcoming treat!

sergiusens:
There are a couple of issues here @chipaca, care to take a look?

kyrofa:
I like this idea, but yeah, we've been wanting to extract this behavior into a module anyway, which is a little more than we'd ask you to do @chipaca. I hope you don't mind, but I've just pushed a commit to this PR.

snapcraft/yaml_utils.py
```diff
@@ -0,0 +1,81 @@
...
+    return OrderedDict(loader.construct_pairs(node))
+
+
+def _str_presenter(dumper, data):
+    if len(data.splitlines()) > 1:  # check for multiline string
+        return dumper.represent_scalar("tag:yaml.org,2002:str", data, style="|")
+    return dumper.represent_scalar("tag:yaml.org,2002:str", data)
+
+
+yaml.add_representer(str, _str_presenter)
```
sergiusens: can we add these to our own private class like https://stackoverflow.com/questions/5121931/in-python-how-can-you-load-yaml-mappings-as-ordereddicts
Affecting yaml globally has the effect of things maybe working and maybe not when running tests depending on when this module is imported

kyrofa: Sure, done.

sergiusens:
these errors might mean you probably want a private load and dump for the states like the GlobalState has. This would allow us to use the C dump logic everywhere else.

snapcraft/extractors/_metadata.py
```diff
@@ -32,7 +33,8 @@ def __init__(
         version: str = "",
         grade: str = "",
         icon: str = "",
-        desktop_file_paths: List[str] = None
+        desktop_file_paths: List[str] = None,
+        _data: Dict[str, Union[str, List[str]]] = None,
```
sergiusens: remove the, at the end of this line to get rid of the flake8 error :-/

snapcraft/extractors/_metadata.py
```diff
@@ -48,10 +47,7 @@ def __init__(
         :param list desktop_file_paths: Extracted desktop file paths
         """  # noqa
 
-        if _data:
-            self._data = _data
-        else:
-            self._data = {}
+        self._data = {}  # type: Dict[str, Union[str, List[str]]]
```
sergiusens: this is just me being blind, but I really started to prefer using `dict()` for these cases

sergiusens:
Can we apply whatever strategy we would do when moving to spread instead?

sergiusens:
It's finally happening!!!!!

sergiusens:
@kyrofa keep an eye on the weird architectures for this one.

PR #2220, opened 2018-08-21
schema: allow license field

mvo5:
We want snaps to also allow setting a license field. This was
discussed in https://forum.snapcraft.io/t/856/53

We got customer complains that `snap info core` shows
`license: unset` for locally installed snaps.

So it would be great if snapcraft could start allowing this.

When using `snap pack` this will also be validated.

### DIFF
--- a/schema/snapcraft.yaml
+++ b/schema/snapcraft.yaml
@@ -156,6 +156,9 @@ properties:
   description:
     type: string
     description: long description of the package
+  license:
+    type: string
+    description: SPDX conform license of the package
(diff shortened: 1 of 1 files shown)


sergiusens:
The agreement was that we would only allow for a license field after we could validate it properly and that is waiting on a proper snap isolated command that could be installed as described on https://forum.snapcraft.io/t/development-sprint-march-5th-2018/4345/6?u=sergiusens

Please feel free to reopen the PR once that is available.

PR #2222, opened 2018-08-22
lxd: support new style snap injection

sergiusens:
Use the same logic as the one implemented in the new provider logic.
Take the opportunity to normalize and untangle the code and also
fix the tests to remove duplication and unnecessary over mocking.

-----

### DIFF
--- a/snapcraft/internal/build_providers/_snap.py
+++ b/snapcraft/internal/build_providers/_snap.py
@@ -52,12 +52,12 @@ class _SnapManager:
         snap_dir: str,
         latest_revision: str,
         snap_arch: str,
-        store_only: bool
+        inject_from_host: bool = True
     ) -> None:
         self.snap_name = snap_name
         self._snap_dir = snap_dir
         self._snap_arch = snap_arch
-        self._store_only = store_only
+        self._inject_from_host = inject_from_host
 
         self._latest_revision = latest_revision
         self.__required_operation = None  # type: Optional[_SnapOp]
@@ -77,9 +77,7 @@ class _SnapManager:
         # From the point of view of multiple architectures if the target host (this)
         # is different than that of where these snaps run from, then we always need to
         # install from the store
-        if self._store_only:
-            is_installed = False
-        else:
+        if self._inject_from_host:
             # Get information from the host.
             host_snap_repo = self._get_snap_repo()
             try:
@@ -88,6 +86,8 @@ class _SnapManager:
             except repo.errors.SnapdConnectionError:
                 # This maybe because we are in a docker instance or another OS.
                 is_installed = False
+        else:
+            is_installed = False
 
         # The evaluations for the required operation is as follows:
         # - if the snap is not installed on the host (is_installed == False),
@@ -230,10 +230,12 @@ class SnapInjector:
         *,
         snap_dir: str,
         registry_filepath: str,
+        snap_arch: str,
         runner: Callable[..., None],
         snap_dir_mounter: Callable[[], None],
         snap_dir_unmounter: Callable[[], None],
-        file_pusher: Callable[..., None]
+        file_pusher: Callable[..., None],
+        inject_from_host: bool = True
     ) -> None:
         """
         Initialize a SnapInjector instance.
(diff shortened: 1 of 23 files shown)


sergiusens:
One thing to notice here is that we aren't really injecting anymore, any form of injecting would require a call to sudo, which could potentially break any scripts out there in the wild.
An option to solve this would be to setup unprivileged containers to run instead of these privileged ones.

sergiusens:
snap refresh snapcraft --channel edge/lxd

popey:
I tested this with a `snapcraft cleanbuild` of a c program snap. Worked as expected!

snapcraft/internal/build_providers/_snap.py
```diff
@@ -46,11 +46,18 @@ class _SnapOp(enum.Enum):
...
 class _SnapManager:
     def __init__(
-        self, *, snap_name: str, snap_dir: str, latest_revision: str, snap_arch: str
+        self,
+        *,
+        snap_name: str,
+        snap_dir: str,
+        latest_revision: str,
+        snap_arch: str,
+        store_only: bool
```
kyrofa: The name of this parameter (both here and in `SnapInjector.add()`) does not really indicate what it does. It's literally saying "don't inject this snap from the host", right? Perhaps `inject_from_host: bool` would better indicate what it does?

snapcraft/internal/build_providers/_snap.py
```diff
@@ -316,13 +329,14 @@ def _record_revision(self, snap_name: str, snap_revision: str) -> None:
         else:
             self._registry_data[snap_name].append(entry)
 
-    def add(self, snap_name: str, snap_arch: str) -> None:
+    def add(self, snap_name: str, snap_arch: str, store_only: bool = False) -> None:
```
kyrofa: Yeah, `inject_from_host: bool = True` feels like it makes more sense, here. However, do you ever envision creating a `SnapInjector` and calling this function while changing this parameter? I suspect not, which means perhaps this parameter should move to `__init__`. Honestly, that probably goes for `snap_arch` as well.

snapcraft/internal/lxd/_containerbuild.py
```diff
@@ -52,31 +54,35 @@
...
+            output = "{}_{}.snap".format(project.info.name, project.deb_arch)
+        self.snap_filename = output
+
         self._source = os.path.realpath(source)
-        self._project_options = project_options
-        self._metadata = metadata
+        self._project = project
         self._user = "root"
-        self._project_folder = "/root/build_{}".format(metadata["name"])
+        self._project_folder = "/root/build_{}".format(project.info.name)
```
kyrofa: Love it!

snapcraft/internal/lxd/_cleanbuilder.py
```diff
@@ -50,24 +50,26 @@ def _ensure_container(self):
...
-        # os.sep needs to be `/` and on Windows it will be set to `\`
-        src = "{}/{}".format(self._project_folder, self._snap_output)
-        self._pull_file(src, self._snap_output)
-        logger.info("Retrieved {}".format(self._snap_output))
+    def _finish(self, success=True):
+        super()._finish(success=success)
+
+        # remove project data as this is a cleanbuild
+        if os.path.exists(self.provider_project_dir):
+            shutil.rmtree(self.provider_project_dir)
```
kyrofa: I like this, but I suggest we also put the tarball we generated of the project in there so we can toast it when we're done. If that doesn't work for some reason, can we at least also remove the tarball here?

snapcraft/internal/build_providers/_snap.py
sergiusens: certainly, we can move it back if needed.

snapcraft/internal/build_providers/_snap.py
sergiusens: I'll switch to this naming convention

snapcraft/internal/lxd/_cleanbuilder.py
sergiusens: I am going to say no to this one, this is not where the tarball goes either.
This is the directory of the provider for private project things like the snap-registry.yaml (which if left would cause bad results) and such. The tarball could have usefulness on good or bad outcomes and is part of the original experience.

PR #2226, opened 2018-08-24
templates: reimplement templates as python classes

kyrofa:

-----

This gives them more flexibility than YAML.


snapcraft/internal/project_loader/_templates/_template.py
```diff
@@ -0,0 +1,59 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from typing import Any, Dict, Set
+
+
+class Template:
```
sergiusens: how about abc?

snapcraft/internal/project_loader/_templates/_template.py
```diff
@@ -0,0 +1,59 @@
...
+    @classmethod
+    def supported_bases(cls) -> Set[str]:
```
sergiusens: If we need to add parens, this reads better and leads to less pythonic errors:
```python
if "core18" in <template>.get_supported_bases()
```

If not, use a property:
```python
if "core18" in <template>supported_bases:
```

Or just a class attribute and rely on python's duck typing.

snapcraft/internal/project_loader/_templates/_template.py
kyrofa: We want default implementations here. Otherwise writing a template to simply add a plug involves implementing all these methods that do nothing.

snapcraft/internal/project_loader/_templates/_template.py
kyrofa: I thought about making this a class attribute, but I think using a method makes for a much easier-to-implement API. I find adding "get" actually decreases readability ("if core18 in template get supported bases" doesn't read as well as "if core18 in template supported bases") and be somewhat meaningless (particularly since there's no "set"), but will add it if you like. However, if we add it to this method, we should probably add it to the rest of the methods in the class as well. Thoughts?

sergiusens: making it an attribute at least fails fast and in your face, adding get_ makes it sort of obvious you need to add parens as it represents an action.

kyrofa: As you requested offline, the `Template` API has been moved entirely to attributes.

sergiusens:
can we get a bugtask created for this too?

PR #2229, opened 2018-08-29
elf: extract build ID and presence of debug info

jhenstridge:

-----
This is related to my proposal about collecting debug information as part of a snapcraft build here: https://forum.snapcraft.io/t/collecting-debug-symbols/7017

The patch simply updates the ElfFile class to extract the build ID and presence of debug info in a file.  This is needed to identify candidates to have their debug info separated, and to identify the debug info corresponding to installed ELF files.

I also took the opportunity to simplify `ElfFile._extract` to remove dependence on section names where they aren't actually relied on.  For example, while the ELF interpreter is usually found in the `.interp` section, the kernel actually looks for it in the `PT_INTERP` section.  You could store the name in a differently named section of that segment and it would still work.


jhenstridge:
@sergiusens: any chance of a review on this?  It's been sitting here for almost a year, and I've had to rebase it a few times to fix conflicts.

It has failed CI on `google:ubuntu-16.04-64:tests/spread/plugins/plainbox/provider-basic`, but that appears to be unrelated to the changes in this branch (seems to be an unreliable "git clone" invocation).

snapcraft/internal/elf.py
```diff
@@ -292,25 +295,29 @@ def _extract(self, path: str) -> ElfDataTuple:  # noqa: C901
...
+                        libs[needed] = NeededLibrary(name=needed)
+                    for tag in segment.iter_tags("DT_SONAME"):
+                        soname = _ensure_str(tag.soname)
+                elif segment["p_type"] == "PT_GNU_STACK":
+                    # p_flags holds the bit mask for this segment.
+                    # See `man 5 elf`.
+                    mode = segment["p_flags"]
+                    if mode & elftools.elf.constants.P_FLAGS.PF_X:
+                        execstack_set = True
+                elif segment["p_type"] == "PT_INTERP":
```
cjp256: It's effectively equivalent after looking at the pyelf code, but I think it should key off the instance types rather than checking the p_type.

e.g. `if isinstance(segment, InterpSegment)` or `if isinstance(segment, NoteSegment)`.

I'm sure the library is sane, but the code here for these two segments are using the interfaces provided by those types.

snapcraft/internal/elf.py
```diff
@@ -268,13 +266,18 @@ def __init__(self, *, path: str) -> None:
         self.needed = elf_data[3]
         self.execstack_set = elf_data[4]
         self.is_dynamic = elf_data[5]
+        self.build_id = elf_data[6]
```
cjp256: nit: add typing
```
self.build_id: str = elf_data[6]
self.has_debug_info: bool = elf_data[7]
```

But it is consistent with everything around it.

snapcraft/internal/elf.py
```diff
@@ -282,15 +289,22 @@ def _extract(self, path: str) -> ElfDataTuple:  # noqa: C901
                     for version in versions:
                         lib.add_version(_ensure_str(version.name))
 
-            for segment in elf.iter_segments():
```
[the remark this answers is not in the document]
jhenstridge: Each segment contains data from one or more named section.  For some of the data we were extracting, it is the segment that matters rather than section name.

For example, while GNU LD writes a `.interp` section containing the dynamic loader to use, the kernel looks at the `PT_INTERP` segment to determine the loader.  An alternative compiler/linker could use a different section name and everything would work provided that the section's data was written to the `PT_INTERP` segment.

snapcraft/internal/elf.py
jhenstridge: Can mypy not infer the types from the `_extract` return type?

snapcraft/internal/elf.py
jhenstridge: It looks like there are special `InterpSegment`, `DynamicSegment`, and `NoteSegment` classes, but the `PT_GNU_STACK` segment shows up as a generic `elftools.elf.segments.Segment` instance.

I can make the change, but it would only cover three of the four branches of the if statement.  That's not obviously a readability win, although I guess it could help with static analysis?

snapcraft/internal/elf.py
```diff
@@ -46,21 +46,19 @@ def add_version(self, version: str) -> None:
 
 ElfArchitectureTuple = Tuple[str, str, str]
 ElfDataTuple = Tuple[
-    ElfArchitectureTuple, str, str, Dict[str, NeededLibrary], bool, bool
+    ElfArchitectureTuple, str, str, Dict[str, NeededLibrary], bool, bool, str, bool
```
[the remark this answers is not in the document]
jhenstridge: The class you're describing is essentially the existing `ElfFile` class.  Convert the constructor to something that simply sets data members, and move the data extraction to either a factory function or class method, and we should have something that can be mocked without a massive tuple filled with repeated types.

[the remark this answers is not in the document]
cjp256: Agreed! :+1:

snapcraft/internal/elf.py
cjp256: I will say the code is effectively OK either way (based on https://github.com/eliben/pyelftools/blob/master/elftools/elf/elffile.py#L452). 

I think it would help analysis. 

But I think if you are going to use the interfaces of specific subclass type in python, I would ensure it's an instance of it (say the library incorrectly returned a Segment instead of NoteSegment for whatever reason...).  And while the readability might look a little more awkward at a glance, it's more obvious that you are now interfacing with a more-specific subclass, and provides the developer with a strong hint as to where to find iter_notes(), get_interp_name(), iter_tags(), etc. which are not in the base Segment class.  I'm not familiar with pyelf, so I had to look it up to and sort out why these calls were OK as-is.

FWIW pyelf's example scripts rely on this approach: https://github.com/eliben/pyelftools/blob/master/examples/elf_notes.py#L25

It's redundant, but an option to keep it both ways is:
`if segment["p_type"] == "PT_DYNAMIC" and isinstance(segment, DynamicSegment):`

snapcraft/internal/elf.py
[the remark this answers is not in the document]
cjp256: Makes sense, thank you for the explanations!

snapcraft/internal/elf.py
cjp256: Ah yes, silly me.  I'm already imaging the future ElfFile where everything is defined in init and not by "the" Tuple! :) Thanks!

PR #2240, opened 2018-09-04
file_utils: find tool when using docker and deb

sergiusens:
The case of returning an empty string was missed. More so,
the case of using the deb inside docker was not covered.

LP: #1790665
Fixes SNAPCRAFT-2E

-----

### DIFF
--- a/snapcraft/file_utils.py
+++ b/snapcraft/file_utils.py
@@ -334,8 +334,9 @@ def calculate_hash(path: str, *, algorithm: str) -> str:
 def get_tool_path(command_name: str) -> str:
     """Return the path to the given command
 
-    Return a path to command_name, if Snapcraft is running out of the snap or out
-    of Docker, it ensures it is using the one in the snap, not the host.
+    Return a path to command_name, if Snapcraft is running out of the snap
+    or in legacy mode (snap or sources), it ensures it is using the one in
+    the snap, not the host.
     If a path cannot be resolved, ToolMissingError is raised.
 
     : param str command_name: the name of the command to resolve a path for.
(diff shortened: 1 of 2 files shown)


snapcraft/file_utils.py
```diff
@@ -343,23 +343,24 @@ def get_tool_path(command_name: str) -> str:
...
     """
+    snapcraft_snap_path = os.path.join(os.sep, "snap", "snapcraft", "current")
+
     if common.is_snap():
         command_path = _command_path_in_root(os.getenv("SNAP"), command_name)
-    elif common.is_docker_instance():
-        command_path = _command_path_in_root(
-            os.path.join(os.sep, "snap", "snapcraft", "current"), command_name
-        )
+    elif common.is_docker_instance() and os.path.exists(snapcraft_snap_path):
```
kyrofa: `is_snap()` is true if `$SNAP_NAME == "snapcraft"`. Due to [this file](https://github.com/snapcore/snapcraft/blob/master/docker/bin/snapcraft-wrapper) it will also be true in the snap-based Docker images, which means I don't think this `elif` will ever be true.

sergiusens: Good catch, removed the condition as discussed during our quick call about it.

snapcraft/file_utils.py
```diff
@@ -334,8 +334,9 @@ def calculate_hash(path: str, *, algorithm: str) -> str:
 def get_tool_path(command_name: str) -> str:
     """Return the path to the given command
 
-    Return a path to command_name, if Snapcraft is running out of the snap or out
-    of Docker, it ensures it is using the one in the snap, not the host.
+    Return a path to command_name, if Snapcraft is running out of the snap
+    or in legacy mode (snap or sources), it ensures it is using the one in
```
kyrofa: Hmm, not sure I'd call it legacy mode, but regardless, I assume you mean to say "(deb or sources)" here?

sergiusens: considering the snap will be the standard way we want to promote wherever your are, yes, all others are indeed legacy from that point of view

sergiusens: fwiw, I am not changing the text here to wait another hour or so for tests to run, that might make us miss a release window for this week.

psiinon:
When will this be available in a docker image?
I've just tried  snapcore/snapcraft:latest and thats still failing with the same stack trace.

PR #2248, opened 2018-09-07
plugin: update catkin plugin to support melodic

NickZ:
ROS Melodic uses bionic as a base. 

I note that I'm required to `stage-packages['libc6']` now using this, which I wasn't before with kinetic, because the GLIBC version of the targeted core was 2.23, and most of the included libraries required 2.27.  I'm not sure how to change this.

Here's my snapcraft.yaml. I am building on Bionic 18.04.1. https://gist.github.com/NickZ/309fc17d5341c86e35f16a726cf48706

### DIFF
--- a/snapcraft/plugins/catkin.py
+++ b/snapcraft/plugins/catkin.py
@@ -89,6 +89,7 @@ _ROS_RELEASE_MAP = {
     "jade": "trusty",
     "kinetic": "xenial",
     "lunar": "xenial",
+    "melodic": "bionic",
 }
 
 _SUPPORTED_DEPENDENCY_TYPES = {"apt", "pip"}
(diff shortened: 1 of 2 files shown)


kyrofa:
Hey there @NickZ, thanks for this! However, given that not even the OSRF wants to take on the burden of ensuring Melodic builds against Xenial, this merge will have to hold off until base support has fully landed (i.e. we need to ensure that if one is using Melodic, they're building against base18, which is based on Bionic).

NickZ:
Yeah, that's what I was confused about; how can you tell which core snap you're building against?

kyrofa:
Yeah that feature hasn't completely landed yet, but once it has we can take this over the line.

kyrofa:
Closing in favor of #2333, thanks @NickZ!

PR #2250, opened 2018-09-07
project_loader: add preflight check

kyrofa:

-----

This is a mechanism that can be used for further quick sanity checks, but currently the only one implemented is ensuring that the snap directory doesn't include unexpected things, which makes this PR fix [LP: #1791368](https://bugs.launchpad.net/snapcraft/+bug/1791368).

### DIFF
--- a/snapcraft/internal/project_loader/_config.py
+++ b/snapcraft/internal/project_loader/_config.py
@@ -35,6 +35,7 @@ from ._env import (
     snapcraft_global_environment,
     environment_to_replacements,
 )
+from ._preflight_check import conduct_preflight_check
 from . import errors, grammar_processing, replace_attr
 
 
@@ -236,6 +237,8 @@ class Config:
         self.data["architectures"] = _process_architectures(
             self.data.get("architectures"), project.deb_arch
         )
+
+        conduct_preflight_check(self.project)
 
     def _ensure_no_duplicate_app_aliases(self):
         # Prevent multiple apps within a snap from having duplicate alias names
--- a/snapcraft/internal/project_loader/_config.py
+++ b/snapcraft/internal/project_loader/_config.py
@@ -35,7 +35,6 @@ from ._env import (
     snapcraft_global_environment,
     environment_to_replacements,
 )
-from ._preflight_check import conduct_preflight_check
 from . import errors, grammar_processing, replace_attr
 
 
@@ -237,8 +236,6 @@ class Config:
         self.data["architectures"] = _process_architectures(
             self.data.get("architectures"), project.deb_arch
         )
-
-        conduct_preflight_check(self.project)
 
     def _ensure_no_duplicate_app_aliases(self):
         # Prevent multiple apps within a snap from having duplicate alias names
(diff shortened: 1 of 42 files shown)


snapcraft/internal/project_loader/_preflight_check.py
```diff
@@ -0,0 +1,63 @@
...
+import snapcraft
+
+logger = logging.getLogger(__name__)
+
+
+_EXPECTED_SNAP_DIR_PATTERNS = {
+    re.compile(r"^snapcraft.yaml$"),
+    re.compile(r"^.snapcraft(/state)?$"),
+    re.compile(r"^plugins(/.*)?$"),
+    re.compile(r"^gui(/.*\.(png|svg|desktop))?$"),
```
sergiusens: missing hooks?

sergiusens: yup, most certainly https://github.com/snapcore/snapcraft/blob/master/snapcraft/internal/meta/_snap_packaging.py#L387

tests/unit/project_loader/test_preflight_check.py
```diff
@@ -0,0 +1,105 @@
...
+                "The snap/ directory is meant specifically for snapcraft, but it "
+                "contains the following non-snapcraft-related paths, which is "
+                "unsupported and will cause unexpected behavior:\n"
+                "- .snapcraft/baz\n"
+                "- dir1\n"
+                "- dir1/foo\n"
+                "- dir1/plugins\n"
+                "- dir2\n"
+                "- dir2/bar\n"
+                "- dir2/gui\n"
```
sergiusens: this and the plugins one are the checks I would of asked for as regex's are easy to mess up, good anticipating there :-)

snapcraft/internal/project_loader/_config.py
```diff
@@ -237,6 +238,8 @@ def __init__(self, project: project.Project) -> None:
             self.data.get("architectures"), project.deb_arch
         )
 
+        conduct_preflight_check(self.project)
```
sergiusens: mind adding a comment of why this is done inside project loading instead of earlier in the game like before project loading?

There is one downside to this, for build VMs and lxd, you will only see this after the environment is created (which is also good for other flight checks we might have) but bad for things like checking if multipass or qemu are there.

We might need 2 flight checks, and this one should be done a bit earlier

snapcraft/internal/project_loader/_preflight_check.py
kyrofa: Fixed.

snapcraft/internal/project_loader/_config.py
kyrofa: Fixed, should happy much earlier now, good call.

kyrofa:
> [...] why is this not in snapcraft.project instead?

We chatted about this offline, and decided to indeed move these checks into the project package, but continue doing the checks in cli.lifecycle. We're also renaming them to "project sanity checks" which makes way for "environment sanity checks" to run in the build environment (e.g. build VM).

This has been pushed.

sergiusens:
broken test is related to https://forum.snapcraft.io/t/snaps-are-broken-in-bionic-lxd-container/7339

PR #2251, opened 2018-09-07
pluginhandler: stop using alias for snapcraftctl

kyrofa:

-----

Once the snapcraft CLI supports plugins providing command chains, scriptlets will need to be run in a subshell, a boundary across which aliases cannot cross. Instead of using an alias when running from the snap, this PR makes progress on [LP: #1790978](https://bugs.launchpad.net/snapcraft/+bug/1790978) by simply installing snapcraftctl into its own directory and adding it to the PATH.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -44,6 +44,10 @@ parts:
         - patchelf
         - squashfs-tools
         - xdelta3
+    organize:
+        # Put snapcraftctl into its own directory that can be included in the PATH
+        # without including other binaries.
+        bin/snapcraftctl: bin/snapcraft-utils/snapcraftctl
     override-pull: |
         version="$(git describe --dirty --always | sed -e 's/-/+git/;y/-/./')"
         [ -n "$(echo $version | grep "+git")" ] && grade=devel || grade=stable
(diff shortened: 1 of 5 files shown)


snap/snapcraft.yaml
```diff
@@ -44,6 +44,10 @@ parts:
         - patchelf
         - squashfs-tools
         - xdelta3
+    organize:
+        # Put snapcraftctl into its own directory that can be included in the PATH
+        # without including other binaries.
+        bin/snapcraftctl: bin/snapcraft-utils/snapcraftctl
```
sergiusens: can we name this "internal-bin" at the top level instead? so it looks like,
    bin/snapcraftctl: internal-bin/snapcraftctl

Is this in anticipation of more tooling?

kyrofa: > can we name this "internal-bin" at the top level instead?

Definitely, no opinion on this whatsoever, as long as it's a directory that only contains tools _intended_ for use in scriptlets. Oh... actually, maybe it should be called "scriptlet-bin" then, since it's actually more of a public bin within a private than a private within a public (if that makes sense).

> Is this in anticipation of more tooling?

No, I don't see any more snapcraftctl-like things in the future, but adding e.g. `$SNAP/bin` to the path would have exposed all sorts of internal things to scriptlets.

sergiusens: scriptlet-bin is fine, I wasn't advocating for internal-bin in any way :-)

Also, not advocating for $SNAP/bin either, it has to be a nice subset of things we want to expose.

PR #2253, opened 2018-09-08
build-providers: add support for --shell-after 

sergiusens:

-----

### DIFF
--- a/snapcraft/cli/lifecycle.py
+++ b/snapcraft/cli/lifecycle.py
@@ -68,13 +68,13 @@ def _execute(
                     instance.execute_step(step)
                 if shell_after:
                     instance.shell()
-            except Exception as exc:
+            except Exception:
                 if project.debug:
                     instance.shell()
                 else:
                     echo.warning("Run the same command again with --debug to shell into the environment "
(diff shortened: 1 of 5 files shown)


snapcraft/cli/lifecycle.py
```diff
@@ -65,6 +66,8 @@ def _execute(
                     instance.pack_project(output=output)
                 else:
                     instance.execute_step(step)
+                if shell_after:
+                    instance.shell()
```
chipaca: This is a little surprising. What happens if the shell exits with error status, at  this point?
I'd expect the shell_after check to be done after we'd cleared the `try`/`except` block -- if necessary with its own exception handler.

sergiusens: it will raise a proper exception that is captured correctly in our exit handler

chipaca: is that what you want? the step itself didn't fail, just the shell after. Are the same messages still applicable? Do you want shell and debug to interact in this way?

sergiusens: I see your point now. Will fix too

PR #2264, opened 2018-09-12
Fix issue where Rust cross compile target is ignored

eberkund:

I have tested that I can still compile the snap for my native arch and as well as fixing the issue I was having. Please see https://forum.snapcraft.io/t/bug-rust-plugin-not-using-correct-pkg-config-path/7302/4 for more information.


eberkund:
Just FYI, I believe the tests that are failing are because they are timing out and not because of anything to do with the PR contents.

eberkund:
Thanks for reviewing this.

The problem was that since I had made it so that snapcraft does not write out a `.cargo/config` file, if one is not provided then it fails when cross compiling. I have modified my PR so that it will only write out the config file (same as original behavior) if a config file does not already exist. If a `.cargo/config` file is already provided by the project, then it will use that.

tests/unit/plugins/test_rust.py
```diff
@@ -104,8 +104,6 @@ def test_cross_compile(self, mock_download):
         )
 
         plugin.build()
-        self.assertThat(os.path.join(plugin._cargo_dir, "config"), FileExists())
```
sergiusens: Given the recent changes, do you mind adding this check back?

eberkund:
Added back `self.assertThat(os.path.join(plugin._cargo_dir, "config"), FileExists())`

PR #2267, opened 2018-09-12
build providers: refresh packages on bring up

sergiusens:
Run packaging repo refresh tasks when bringing up an environment.
This requires adding start to the BaseProvider and bringing up logic
that was in the Multipass provider.

LP: #1792214

-----

### DIFF
--- a/snapcraft/cli/containers.py
+++ b/snapcraft/cli/containers.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright (C) 2017 Canonical Ltd
+# Copyright (C) 2017-2018 Canonical Ltd
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License version 3 as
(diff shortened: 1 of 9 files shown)


snapcraft/cli/containers.py
```diff
@@ -32,25 +32,17 @@ def containerscli():
...
-        )
-
-    project = get_project(**kwargs, debug=debug)
-    lxd.Project(project=project, output=None, source=os.path.curdir).refresh()
+
+    if build_environment.is_lxd:
+        project = get_project(**kwargs, debug=debug)
+        lxd.Project(project=project, output=None, source=os.path.curdir).refresh()
+    else:
+        repo.Repo.refresh_build_packages()
```
kyrofa: It looks like this will run `sudo apt update` on the host instead of warning that refresh only applies to containers, is that expected?

sergiusens: yes, refresh is a hidden command
we basically assume that running snapcraft at all might do sudo apt install too (these days)
We can polish this for 3.0 and remove all sudo calls as snapcraft will rerun itself with sudo (we cannot do that today as we will break)

kyrofa: Works for me, just want to ensure it was on purpose.

sergiusens:
theb broken scenario is still broken https://forum.snapcraft.io/t/snaps-are-broken-in-bionic-lxd-container/7339/2

PR #2289, opened 2018-09-24
local source: only filter out snapcraft artifacts

kyrofa:

-----

Currently, the local source filters out essentially everything related to the snapcraft CLI, including the entire `snap/` directory. However, this is over-zealous: the `snap/` directory and its contents are often committed to version control, and the process of filtering them out makes changes as far as version control is concerned, resulting in a dirty tree. This really becomes problematic if one then uses `git describe --dirty` to determine the version: it's always dirty.

This PR fixes [LP: #1662388](https://bugs.launchpad.net/snapcraft/+bug/1662388) by taking a more fine-grained approach to filtering out files related to the snapcraft CLI: only filter artifacts, i.e. things generated by the snapcraft CLI. This includes `parts/` `stage/`, `prime/` and `snap/.snapcraft/`.

### DIFF
--- a/snapcraft/internal/common.py
+++ b/snapcraft/internal/common.py
@@ -31,15 +31,7 @@ from typing import Callable, List
 from snapcraft.internal import errors
 
 
-SNAPCRAFT_FILES = [
-    "snapcraft.yaml",
-    ".snapcraft.yaml",
-    ".snapcraft",
-    "parts",
-    "stage",
-    "prime",
-    "snap",
-]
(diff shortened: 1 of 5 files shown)


kyrofa:
As a side effect of this change, the snapcraft.yaml is now considered just another file in the source of the part. If it changes, the source for the part will be considered out of date. That makes sense since it's literally part of the source, but it might be surprising. Just a heads up.

sergiusens:
I think I am so far -1 on this until we have rebuilding without cleaning before. It does not affect small projects but seems like a major burden for the bigger ones.

PR #2317, opened 2018-10-03
tests: add spread suite for plainbox plugin

sergiusens:
Move the plugin and snapd integration tests for the CMake plugin into a
Spread suite.
Also pin the checkbox-ng package in the tests checkbox snap.

LP: #1783839

-----


sergiusens:
reminder, amend the commit message

tests/spread/plugins/plainbox/run/task.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+restore: |
+  cd "$SNAP_DIR"
+  snapcraft clean
+  rm -f ./*.snap
+
+execute: |
+  cd "$SNAP_DIR"
+  snapcraft
+  sudo snap install checkbox-simple_*.snap --dangerous
```
sergiocazzolato: sudo is not needed here

tests/spread/plugins/plainbox/provider-basic/task.yaml
```diff
@@ -0,0 +1,17 @@
+summary: Ensure a basic provider works
+
+systems: [ubuntu-16*, ubuntu-18*]
+
+environment:
+  SNAP_DIR: ../snaps/provider-basic
```
sergiocazzolato: I would define an env called SNAPS_DIR  (or similar) at spread.yaml level pointing to ../snaps/ (use absolute path if possible as we go with the tests_lib path https://github.com/snapcore/snapd/blob/master/spread.yaml#L11), and then you can define SNAP_DIR: $SNAPS_DIR/provider-basic

tests/spread/plugins/plainbox/run/task.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+systems: [ubuntu-16*]
+
+environment:
+  SNAP_DIR: ../snaps/checkbox
+
+restore: |
+  cd "$SNAP_DIR"
+  snapcraft clean
+  rm -f ./*.snap
```
sergiocazzolato: are you uninstalling the snap in other place? otherwise you should do it here

tests/spread/plugins/plainbox/provider-basic/task.yaml
sergiusens: I will consider this for a future refactor, as all our snaps live within the suites themselves (this relative path does not leave the suite level)

tests/spread/plugins/plainbox/run/task.yaml
sergiusens: good catch, we are not doing this.

tests/spread/plugins/plainbox/provider-basic/task.yaml
sergiusens: We can do this in a future PR that is more widespread, I don't want to deviate from the current layout in just this PR.

PR #2323, opened 2018-10-05
go plugin: support for bases

sergiusens:
Revamp the go plugins installation candidates to prefer the go snap
from the Snap Store first and provide a mechanism to be able to
install a build-packages golang-go instead on the supported bases.

LP: #1794785

-----


snapcraft/internal/errors.py
```diff
@@ -250,6 +250,14 @@ def __init__(self, message):
         super().__init__(message=message)
 
 
+class PluginBaseError(SnapcraftError):
+
+    fmt = "The {plugin_name!r} plugin does not support snaps targeting base {base!r}."
```
kyrofa: Perhaps "using" is a better word here, what do you think? When I think "target" I think "what do I run on?" which is typically Windows, Mac OS, or Linux, or perhaps Ubuntu, or even specific versions of Ubuntu. Specifying a base isn't really the same thing: nothing "runs on" core18, in fact, by virtue of being a snap it runs in a number of places and just happens to use core18 to get that functionality.

snapcraft/plugins/go.py
```diff
@@ -93,15 +103,22 @@ def get_pull_properties(cls):
...
         self._gopath_bin = os.path.join(self._gopath, "bin")
         self._gopath_pkg = os.path.join(self._gopath, "pkg")
 
+    def _setup_base_tools(self, go_channel, base):
+        if go_channel != "":
+            self.build_snaps.append("go/{}".format(go_channel))
+        elif base in ("core16", "core18"):
+            self.build_packages.append("golang-go")
+        else:
+            raise errors.PluginBaseError(plugin_name=self.name, base=base)
```
kyrofa: To be clear: you anticipate this plugin will run fine on any base as long as the Go snap is being used?

tests/spread/plugins/go/classic/task.yaml
```diff
@@ -1,32 +1,24 @@
...
-# Classic snaps don't build on 14.04
-systems: [-ubuntu-14*]
-
 environment:
   SNAP_DIR: ../snaps/go-gotty
 
 prepare: |
+  . "$TOOLS_DIR/snapcraft-yaml.sh"
+  set-base "$SNAP_DIR/snap/snapcraft.yaml"
   apt install patchelf -y
```
kyrofa: This isn't available in Trusty. You can either leave the `systems` filter there but change the comment, or use `apt install patchelf -y || true` and only check the rpaths `if which patchelf &> /dev/null`.

tests/spread/plugins/go/cross_compile_cgo/task.yaml
```diff
@@ -1,15 +1,18 @@
 summary: Cross-compile Go snaps using cgo
 
-# cgo doesn't build on 14.04
```
kyrofa: Is this comment no longer valid? Did something change?

tests/spread/plugins/go/classic/task.yaml
```diff
@@ -1,32 +1,24 @@
 summary: Build and run a basic Go snap
 
-# Classic snaps don't build on 14.04
```
kyrofa: Is this comment no longer valid?

tests/spread/plugins/go/legacy-build-run/task.yaml
```diff
@@ -0,0 +1,17 @@
+summary: Build and run a basic Go snap with no base
+
+environment:
+  SNAP_DIR: ../snaps/go-hello
+
+systems: [ubuntu-*]
```
kyrofa: What are you filtering out with this?

tests/spread/plugins/go/snaps/go-gotty/snap/snapcraft.yaml
```diff
@@ -5,18 +5,13 @@ description: |
...
   applications.
 
-grade: devel # must be 'stable' to release into candidate/stable channels
+grade: devel
 confinement: classic
 
 parts:
-  go:
-    source-tag: go1.9.2
-    source-depth: 1
```
kyrofa: Nice. Should speed things up a _lot_.

tests/spread/plugins/go/snaps/go-use-build-packages/snap/snapcraft.yaml
```diff
@@ -0,0 +1,20 @@
...
+apps:
+  go-hello:
+    command: go-hello
+
+parts:
+  go-hello:
+    plugin: go
+    source: go-hello
+    build-packages:
+      -golang-go
```
kyrofa: No space after the hyphen is bothering my OCD. It looks like a negative.

tests/spread/plugins/go/use-build-packages/task.yaml
```diff
@@ -0,0 +1,26 @@
...
+  snapcraft clean
+  rm -f ./*.snap
+  restore-yaml snap/snapcraft.yaml
+
+execute: |
+  cd "$SNAP_DIR"
+  snapcraft build
+
+  if snap list go; then
+    echo "The go snap from the Snap Store should not have been installed."
```
kyrofa: Hey, thanks for the nice message! :+1:

tests/spread/tools/snapcraft-yaml.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash -e
+
+set-base()
```
kyrofa: Would you mind using underscores here, please? Hyphens [are non-standard](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_235), and don't follow the convention already established in these scripts.

tests/spread/plugins/go/classic/task.yaml
sergiusens: the filter is at the top level in spread.yaml though and only special casing legacy.

tests/spread/plugins/go/legacy-build-run/task.yaml
sergiusens: I am considering someone will add a base called fedora, we will add tests for it and not run this.

snapcraft/plugins/go.py
sergiusens: I anticipate it is a good starter for anyone trying to start a port to a new device.

tests/spread/plugins/go/classic/task.yaml
sergiusens: Trusty is dead when dealing with bases btw; it should only ever go into that mode with `--destructive` and at that point, all bets are off.

tests/spread/plugins/go/cross_compile_cgo/task.yaml
sergiusens: trusty does not have a base, we are not testing it, so the comment is just as good as specifying where else it does not work

tests/spread/tools/snapcraft-yaml.sh
sergiusens: I just dislike underscores, but very well

tests/spread/plugins/go/snaps/go-use-build-packages/snap/snapcraft.yaml
sergiusens: yeah, it is not even an array :-)

tests/spread/plugins/go/snaps/go-gotty/snap/snapcraft.yaml
sergiusens: it does!

tests/spread/plugins/go/classic/task.yaml
sergiusens: Given we do not support _running_ on trusty with bases it is not even a test target.
We have to probably remove it as the default run and keep it for legacy triggers only (and a lxd/multipass smoke)

snapcraft/plugins/go.py
```diff
@@ -22,6 +22,11 @@
 
 Additionally, this plugin uses the following plugin-specific keywords:
 
+    - go-channel:
+      (string, default: latest/stable)
+      The Snap Store channel to install go from. If set to an empty string,
+      go will be installed using the systems's traditional package manager.
```
diddledan: Typo: change `systems's` to `system's`

tests/spread/plugins/go/snaps/go-use-build-packages/snap/snapcraft.yaml
```diff
@@ -0,0 +1,20 @@
+name: go-hello
+version: 0.1
+summary: A simple go project
+description: |
+  This is a basic go snap. It just prints a hello world and checksthat it is
```
diddledan: missing space in `checksthat`

tests/unit/plugins/test_go.py
```diff
@@ -110,6 +66,30 @@ def test_schema(self):
                 "Expected {!r} to be included in properties".format(expected),
             )
 
+        # Check go-channel
+        go_channel = properties["go-channel"]
+        for expected in ["type", "default"]:
+            self.assertTrue(
+                expected in go_channel,
+                "Expected {!r} to be included in 'go-channel'".format(expected),
```
diddledan: note that here we're checking the string for single quotes.... (to be continued)

tests/unit/plugins/test_go.py
```diff
@@ -110,6 +66,30 @@ def test_schema(self):
...
+            )
+
+        go_channel_type = go_channel["type"]
+        self.assertThat(
+            go_channel_type,
+            Equals("string"),
+            'Expected "go-channel" "type" to be "string", but '
```
diddledan: ... but here we're checking for double quotes ... (to be continued)

tests/unit/plugins/test_go.py
```diff
@@ -110,6 +66,30 @@ def test_schema(self):
...
+            'it was "{}"'.format(go_channel_type),
+        )
+
+        go_channel_default = go_channel["default"]
+        self.assertThat(
+            go_channel_default,
+            Equals("latest/stable"),
+            'Expected "go-channel" "default" to be '
+            '"latest/stable", but it was "{}"'.format(go_channel_default),
```
diddledan: ... and we're also checking for double quotes here. (fin)

These three comments of mine are more a question of should these messages be unified in their usage of quotes? Here we're checking the output in a test so they need to match what is actually being output, but that output should IMO be consistent with it's usage of souble vs single quotes. Maybe make a note to go back and re-do those in a separate PR?

tests/unit/plugins/test_go.py
```diff
@@ -605,27 +608,113 @@ class Options:
...
+        self.assertThat(plugin.build_packages, Not(Contains("golang-go")))
+        self.assertThat(plugin.build_snaps, Contains("go/latest/stable"))
+
+    def test_build_packages(self):
+        self.options.go_channel = ""
+
+        plugin = go.GoPlugin("test-part", self.options, self.project)
+
+        self.assertThat(plugin.build_packages, Contains("golang-go"))
+        self.assertThat(plugin.build_snaps, Not(Contains("go/latest/stable")))
```
diddledan: Would it be advisable to add an extra test for a specific non-latest channel to ensure that we're able to accept arbitrary input in the `go-channel` setting? I suggest adding an extra test that sets, and checks, for `go-channel: 1.11/stable`. This will be almost identical to the `test_snap` function but with the added override on `self.options.go_channel`:

```python
    def test_alternative_channel(self):
        self.options.go_channel = "1.11/stable"

        plugin = go.GoPlugin("test-part", self.options, self.project)

        self.assertThat(plugin.build_packages, Not(Contains("golang-go")))
        self.assertThat(plugin.build_snaps, Contains("go/1.11/stable"))
```

tests/unit/plugins/test_go.py
sergiusens: This is help text for when an assertion fails, black, the autoformatter we use would have changed the outer quotes to be double if there hadn't been any inner single quotes.

This is all rather legacy.

tests/spread/plugins/go/snaps/go-use-build-packages/snap/snapcraft.yaml
sergiusens: Good catch

tests/unit/plugins/test_go.py
sergiusens: Yes, they should and yes they are, this is just test code though with brief summaries of why things fail which should not reach an end user :smile:

tests/unit/plugins/test_go.py
sergiusens: Thanks for the catch, the validity of the channel is checked by a different component, the same one that checks for build-snaps.

sergiusens: An integration test however might be a good idea.

PR #2327, opened 2018-10-09
pluginhandler: remove prepare, build and install scriptlets

sergiusens:
These have long since been deprecated in favour of the
override-{step-name} syntax.

The integration tests remain to ensure backwards compatibility is in
order.

LP: #1794854

-----

### DIFF
--- a/tests/integration/general/test_scriptlets.py
+++ b/tests/integration/general/test_scriptlets.py
@@ -14,49 +14,14 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import multiprocessing
 import os
 
-from testtools.matchers import FileContains, FileExists, Not
+from testtools.matchers import FileExists, Not
 
 from tests import integration
 
 
 class ScriptletTestCase(integration.TestCase):
-    def test_prepare_scriptlet(self):
-        self.run_snapcraft("build", "scriptlet-prepare")
-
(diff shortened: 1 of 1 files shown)


schema/snapcraft.yaml
```diff
@@ -583,29 +488,13 @@ properties:
             items:
               type: string
             default: ['*']
-          snap: # FIXME: This is deprecated, replaced by `prime`, below.
```
kyrofa: This should probably be done in its own PR, no?

tests/unit/lifecycle/test_lifecycle.py
```diff
@@ -385,77 +385,6 @@ def test_non_prime_and_no_version(self):
         lifecycle.execute(steps.PULL, project_config)
 
 
-class DirtyBuildScriptletTestCase(LifecycleTestBase):
```
kyrofa: This is actually quite an important test, can you just change it to use the override scriptlets instead?

sergiusens: sure, I just saw old methods and decided to wipe. good catch

schema/snapcraft.yaml
sergiusens: it even has its own bug, let me fix that slip

tests/unit/lifecycle/test_lifecycle.py
sergiusens: wait, these same tests already exist

PR #2331, opened 2018-10-09
meson plugin: add support for bases

sergiusens:
Revamp the meson plugin to be base aware. Install meson straight from
PyPI instead of from the system's repositories.

LP: #1794794

-----


snapcraft/plugins/meson.py
```diff
@@ -24,15 +24,22 @@
 
 Additionally, this plugin uses the following plugin-specific keywords:
 
+    - meson-version:
+      (string, defaul: latest)
```
kyrofa: s/defaul/default/

snapcraft/plugins/meson.py
```diff
@@ -45,20 +52,50 @@ def schema(cls):
             "items": {"type": "string"},
             "default": [],
         }
+        schema["properties"]["meson-version"] = {"type": "string", "default": ""}
```
kyrofa: This default doesn't match up with the docs above. Should the docs say "if unspecified, the latest will be used"?

sergiusens: I'll fix, thanks

PR #2336, opened 2018-10-10
schema, meta: support layout

kyrofa:

-----

snapd has experimental support for the `layout` property, which provides the ability to reorganize the environment seen by the snap using bind-mounts, etc. This PR resolves [LP: #1754639](https://bugs.launchpad.net/snapcraft/+bug/1754639) by exposing this feature in the snapcraft CLI, but it warns that it's still experimental.

### DIFF
--- a/snapcraft/internal/meta/_snap_packaging.py
+++ b/snapcraft/internal/meta/_snap_packaging.py
@@ -56,6 +56,7 @@ _OPTIONAL_PACKAGE_KEYS = [
     "epoch",
     "grade",
     "hooks",
+    "layout",
     "license",
     "plugs",
     "slots",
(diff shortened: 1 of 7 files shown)


snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -532,6 +533,11 @@ def _compose_snap_yaml(self):
 
         self._process_passthrough_properties(snap_yaml)
 
+        # FIXME: Remove this warning once layouts are stable and no longer behind a
+        # feature flag in snapd.
```
sergiusens: is there a way to dynamically check this?

kyrofa: We chatted about this in IRC. `snap pack --check-skeleton` will catch invalid layouts, so @sergiusens requested that we not check the contents of the layout at all and just pass it through to the snap.yaml. This means we can't catch mistakes before the snap is built, but it gives snapd the freedom to change without requiring changes in snapcraft. I'll propose a new PR for this.

PR #2337, opened 2018-10-10
pluginhandler: library detection instead of injection

sergiusens:
With bases, the logic to inject binaries from the host will no longer be
at play. Instead, the base snap for the target project will always be
installed and necessity of the library will be determined throught there
instead

If a library is missing a warning will be issued instead of an error
given the existence of the content snap and its potential use.

LP: #1794556

-----


snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -860,32 +858,13 @@ def _handle_dependencies(self, all_dependencies: Set[str]):
...
-                    "stage-packages entry or through a part:\n{}".format(
-                        formatted_system
-                    )
-                )
+        if system:
+            formatted_system = "\n".join(sorted(system))
+            # We cannot error if we consider the content interface...
+            logger.warning(
+                "Files from the build host were use to build this snap "
+                "but are not part of the snap or base: \n{}".format(formatted_system)
```
kyrofa: I suggest "The {part_name!r} part needs the following libraries that are not included in the snap or base: \n{}\n\nThese dependencies can be satisfied via other stage-packages, other parts, or content sharing.".format(formatted_system)

snapcraft/internal/project_loader/_parts_config.py
```diff
@@ -252,23 +249,26 @@ def build_env_for_part(self, part, root_part=True) -> List[str]:
...
         if root_part:
             # this has to come before any {}/usr/bin
             env += part.env(part.plugin.installdir)
             env += runtime_env(part.plugin.installdir, self._project.arch_triplet)
             env += runtime_env(stagedir, self._project.arch_triplet)
             env += build_env(
-                part.plugin.installdir, self._snap_name, self._project.arch_triplet
+                part.plugin.installdir,
+                self._project.info.name,
+                self._project.arch_triplet,
```
kyrofa: There are a lot of changes in this PR that don't actually seem to relate to the issue being fixed, it makes the review more difficult.

snapcraft/internal/elf.py
```diff
@@ -369,7 +363,7 @@ def load_dependencies(
         # Return a set useful only for fetching libraries from the host
         library_paths = set()  # type: Set[str]
         for l in libs:
-            if os.path.exists(l.path) and not l.in_base_snap and not l.system_lib:
+            if os.path.exists(l.path) and not l.in_base_snap:
```
kyrofa: This seems problematic. We're now injecting the base, which is nice, but we never actually check to see if a library is contained within it now. `l.is_base_snap` only checks the linker path (before patchelf), which is not the same as checking to see if e.g. `/usr/lib/libfoo.so` is contained within the base. As a result, every system dependency will be tracked, regardless of whether or not it's in the base.

We should probably keep this attribute, setting its value using something similar to `_get_system_libs()` (which is removed below) to check the base snap instead of the pre-generated library list.

sergiusens: we have logic that still executes this check

snapcraft/internal/project_loader/_parts_config.py
sergiusens: I changed base as it was now technically incorrect and thought I'd remove the other two props

sergiusens: I did the code count @kyrofa, and I believe you are probably exagerating, it is 7 lines of code that really changed do to using `self._project.info.name` and `self._project.info.confinement`.

Not defaulting to `core` in the base is somewhat of a requirement as it is not a base.

snapcraft/internal/project_loader/_config.py
```diff
@@ -218,6 +218,18 @@ def __init__(self, project: project.Project) -> None:
...
         self.build_tools |= set(project.additional_build_packages)
 
+        # Always add the base for building for non os and base snaps
+        if project.info.base is not None and project.info.type not in ("base", "os"):
+            self.build_snaps.add(project.info.base)
+        elif project.info.type not in ("base", "os"):
+            # This exception is here to help with porting issues with bases. In normal
+            # executions, when no base is set, the legacy snapcraft will be executed.
+            raise RuntimeError(
+                "A base is required for the scenario of snap type {!r}".format(
```
kyrofa: "A base is required for {!r} snaps." conveys the same information in less words.

PR #2339, opened 2018-10-10
lifecycle: switch to multipass by default

sergiusens:
Add a destructive mode to trigger building on the host.

LP: #1791201

-----

### DIFF
--- a/bin/snapcraft
+++ b/bin/snapcraft
@@ -37,6 +37,11 @@ if os.path.isdir(common.get_legacy_snapcraft_dir()):
             base_dir = os.path.expanduser(os.path.join("~", "project"))
         else:
             base_dir = None
+
+        # Backwards compatibility to allow legacy to keep working
+        if os.getenv("SNAPCRAFT_BUILD_ENVIRONMENT") == "destructive-host":
+            del os.environ["SNAPCRAFT_BUILD_ENVIRONMENT"]
+
         snapcraft_yaml_path = project.get_snapcraft_yaml(base_dir=base_dir)
         with open(snapcraft_yaml_path, "r") as f:
             data = yaml_utils.load(f)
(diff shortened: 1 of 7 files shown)


snapcraft/cli/env.py
```diff
@@ -28,31 +28,32 @@ class BuilderEnvironmentConfig:
 
     Valid values are:
 
-    - host: the host will drive the build.
```
kyrofa: Above, the docstring still includes "If it is not set, a value of `host` is assumed."

bin/snapcraft
```diff
@@ -37,6 +37,11 @@ if os.path.isdir(common.get_legacy_snapcraft_dir()):
             base_dir = os.path.expanduser(os.path.join("~", "project"))
         else:
             base_dir = None
+
+        # Backwards compatibility to allow legacy to keep working
+        if os.getenv("SNAPCRAFT_BUILD_ENVIRONMENT") == "destructive-host":
+            del os.environ["SNAPCRAFT_BUILD_ENVIRONMENT"]
```
kyrofa: In general, I don't like how ingrained legacy has already become in master. I think workarounds should be added to legacy as much as possible rather than master, or it becomes impossible to remove. I would rather see `destructive-host` added to legacy so we didn't need to hack around it here. I understand this is done for time-saving as we approach the end of our cycle, but we need to be realistic and careful: not everything we punt now will be fixed.

sergiusens: I agree, but also consider that `bin/snapcraft` in its entirety should be removed per [LP: #1795680](https://bugs.launchpad.net/snapcraft/+bug/1795680). Legacy is built as part of snapcraft so we can do this in concert.

sergiusens: going back to host as the envvar value to avoid this discussion and it makes for a better name anyways.

PR #2343, opened 2018-10-11
schema, meta: add "full" app adapter

kyrofa:

-----

Today, the snapcraft CLI supports both "none" and "legacy" as app adapters, "none" meaning no wrappers or environment at all, and "legacy" meaning wrappers. This PR resolves [LP: #1794532](https://bugs.launchpad.net/snapcraft/+bug/1794532) by adding a "full" app adapter that means no more wrappers, but fully utilizes snap'd command-chain to accomplish the same thing in a much nicer way. This requires snapd 2.36 which is not yet released, so continue using "legacy" as the default for now.


snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -605,9 +610,40 @@ def _wrap_apps(self, apps: Dict[str, Any]) -> Dict[str, Any]:
...
+    def _add_command_chain_to_app(self, app_name, app):
+        # First, validate the commands. Without a wrapper, we're bound by snapd's
+        # restrictions. No quotes, and the first string must be a relative path from
+        # the root of the snap.
+        commands = (app[k] for k in ("command", "stop-command") if k in app)
+        for command in commands:
+            if not _APP_COMMAND_PATTERN.match(command):
+                raise errors.InvalidAppCommandFormatError(command, app_name)
+
+            command_without_args = command.split()[0]
```
sergiusens: Do you want to check that this does not start with slash for a fine grained error message? The regex does not capture that.

snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -605,9 +610,40 @@ def _wrap_apps(self, apps: Dict[str, Any]) -> Dict[str, Any]:
...
+                    mode & stat.S_IXUSR or mode & stat.S_IXGRP or mode & stat.S_IXOTH
+                )
+
+            if not is_executable:
+                raise errors.InvalidAppCommandError(command_without_args, app_name)
+
+        # Finally, assuming there is a meta runner, add it to the BEGINNING of the
+        # command chain. This is to ensure all subsequent commands in the chain get
+        # a valid environment.
+        if os.path.isfile(self._meta_runner):
```
sergiusens: this made me scratch my head a bit in the sense that we might not have picked the right name. meta is very meta :-)

snapcraft/internal/meta/_snap_packaging.py
kyrofa: Good catch-- actually no, we don't want to error, it seems snapd supports that, so we must as well. I've fixed the check and added a test to ensure this works.

snapcraft/internal/meta/_snap_packaging.py
kyrofa: Haha, yeah it's not the best. The naming challenge comes from the fact that eventually plugins will be supplying command chain binaries as well, so we need to differentiate the "base" one coming from meta, so I went with... meta. I'm open to suggestions.

snapcraft/internal/meta/_snap_packaging.py
kyrofa: We chatted about this in IRC, and decided that supporting this would end up confusing people, so I've pushed an update that disallows leading slashes.

PR #2359, opened 2018-10-16
[WIP] extensions: add glib

kyrofa:

-----

This PR resolves [LP: #1797627](https://bugs.launchpad.net/snapcraft/+bug/1797627) by adding a glib extension.

**This is a work in progress.** It works, but it's mostly a fixed-up copy/paste job of the remote part. There's a lot that it does that I don't understand, and I'd like to flesh out the spread test(s) a _lot_ more to ensure it behaves as expected.


sergiusens:
@kenvandine and @oSoMoN mind reviewing this?

kyrofa:
> can we inject `assumes: snapd_236` [...]

Sure, but I'll make another PR adding that functionality to meta instead of specific extensions.

extensions/desktop-common/desktop-exports
```diff
@@ -0,0 +1,416 @@
...
+    source "${XDG_CONFIG_HOME:-$HOME/.config}/user-dirs.dirs"
+  fi
+
+  if [ -z "${XDG_SPECIAL_DIRS+x}" ]; then
+    save_initial_values=true
+  fi
+
+  for d in $XDG_DIRS; do
+    var="XDG_${d}_DIR"
+    value="${!var}"
```
oSoMoN: This looks like a clever and elegant replacement for the horrible nested echo and eval magic. Thanks!

tests/spread/extensions/glib/run/task.yaml
```diff
@@ -0,0 +1,38 @@
...
+  snapcraft
+  sudo snap install glib-hello_*.snap --dangerous
+  [ "$(glib-hello)" = "hello world" ]
+
+  # Verify that the extension command chain went through the proper setup procedure
+  snap_user_data="$HOME/snap/glib-hello/current"
+  [ -d "$snap_user_data/.config" ]
+  [ -d "$snap_user_data/.local" ]
+  [ -h "$snap_user_data/.themes" ]
+
```
oSoMoN: It would be useful to test the contents of:

 - $snap_user_data/.config/user-dirs.dirs
 - $snap_user_data/.config/fontconfig/fonts.conf
 - $snap_user_data/.config/ibus

snapcraft/internal/project_loader/_extensions/glib.py
```diff
@@ -0,0 +1,77 @@
...
+        self.app_snippet = {
+            "command-chain": [
+                "snap/command-chain/desktop-init",
+                "snap/command-chain/desktop-exports",
+                "snap/command-chain/desktop-mark-and-exec",
+            ]
+        }
+
+        # Add the following snippet to each part in the snapcraft.yaml
+        self.part_snippet = {"after": ["glib-extension-init"]}
```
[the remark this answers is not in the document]
kyrofa: Ha! You just [discovered a bug](https://bugs.launchpad.net/snapcraft/+bug/1799596). Indeed, I went through a few renditions of this extension and that needed an update. Good catch, thank you. Fixed.

snapcraft/internal/project_loader/_extensions/glib.py
```diff
@@ -0,0 +1,77 @@
...
+
+        # Add the following part
+        self.parts = {
+            "glib-extension": {
+                "plugin": "dump",
+                "source": "$SNAPCRAFT_EXTENSIONS_DIR/desktop-common",
+                "organize": {"desktop-*": "snap/command-chain/"},
+                "build-packages": ["libglib2.0-dev"],
+                "stage-packages": ["libglib2.0-0"],
```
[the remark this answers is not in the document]
kyrofa: Ah, no indeed, I totally missed that you guys started added a stage-package at all. I've changed this to `libglib2.0-bin`.

extensions/desktop-common/desktop-exports
kyrofa: Sure thing, I busted my noggin just trying to figure out what was happening with that other bit! :rofl:

tests/spread/extensions/glib/run/task.yaml
kyrofa: Alright I'll add that in. Can you think of a way to test that snap upgrades work as expected? What new content should be copied over?

snapcraft/internal/project_loader/_extensions/glib.py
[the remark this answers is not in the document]
kyrofa: Although it looks to me like the extension itself doesn't use anything from -bin, do you agree? Is it necessary to include?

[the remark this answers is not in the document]
oSoMoN: You're right. I thought the glib-compile-schemas and gio-querymodules binaries were installed by libglib2.0-bin, but they are mere symlinks to the binaries installed by libglib2.0-0, and the extension invokes the binaries with their full path (`$RUNTIME/usr/lib/$SNAP_LAUNCHER_ARCH_TRIPLET/glib-2.0/`). So this can be reverted to libglib2.0-bin.

tests/spread/extensions/glib/run/task.yaml
oSoMoN: On upgrade there are a number of files under $SNAP_USER_DATA that are deleted and regenerated. I just verified that reinstalling the same snap from a local file results in a new revision (from x1 to x2), which will exercise the upgrade code path. So we could install the snap once, run it, then intentionally modify/delete some files under $SNAP_USER_DATA, then install it and run it again, and verify that the files have been regenerated.
We could also verify that the contents of `.last_revision` are updated after the second run.

kyrofa: I agree, but at least as this test snap stands, I don't think it puts anything into the snap itself that gets copied over, is that true? What would I need to add to the snap in order to make sure that works as expected?

snapcraft/internal/project_loader/_extensions/glib.py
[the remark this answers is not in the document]
kyrofa: Alright very good, done.

tests/spread/extensions/glib/run/task.yaml
oSoMoN: No, that snap doesn't copy over any custom content to $SNAP_USER_DATA, however the desktop helpers still create some common files/symlinks that can be overwritten for testing purposes:
 - `.config/fontconfig/fonts.conf`
 - `.config/ibus/bus` (symlink)
 - `.config/dconf/user` (symlink)
 - `.last_revision`

kyrofa:
We've been asked to start with a gnome extension instead, so I'm going to rework this a bit and re-propose.

PR #2370, opened 2018-10-19
ant plugin: add support for bases

sergiusens:
Update the Maven plugin to be base-aware, only supporting core16 and
core18.

LP: #1794776

-----


snapcraft/plugins/ant.py
```diff
@@ -59,22 +102,88 @@ def schema(cls):
             "items": {"type": "string"},
             "default": [],
         }
+
+        schema["properties"]["ant-version"] = {"type": "string"}
+
+        schema["properties"]["ant-version-checksum"] = {"type": "string"}
+
+        schema["required"] = ["source"]
+        schema["properties"]["ant-openjdk-version"] = {"type": "string", "default": ""}
```
kyrofa: This default isn't actually necessary, is it?

snapcraft/plugins/ant.py
```diff
@@ -59,22 +102,88 @@ def schema(cls):
             "items": {"type": "string"},
             "default": [],
         }
+
+        schema["properties"]["ant-version"] = {"type": "string"}
+
+        schema["properties"]["ant-version-checksum"] = {"type": "string"}
+
+        schema["required"] = ["source"]
```
kyrofa: This is already two lines below.

snapcraft/plugins/ant.py
```diff
@@ -59,22 +102,88 @@ def schema(cls):
...
+        return ["ant-version", "ant-version-checksum", "ant-openjdk-version"]
 
     @classmethod
     def get_build_properties(cls):
         # Inform Snapcraft of the properties associated with building. If these
         # change in the YAML Snapcraft will consider the build step dirty.
         return ["ant-build-targets", "ant-properties"]
 
+    @property
+    def _ant_tar(self):
```
kyrofa: What is the benefit of doing this lazily instead of just setting it in `_setup_ant`?

snapcraft/plugins/ant.py
```diff
@@ -83,15 +192,44 @@ def build(self):
...
             jardir = os.path.join(self.installdir, "jar")
             os.makedirs(jardir)
             for f in files:
                 base = os.path.basename(f)
                 os.link(f, os.path.join(jardir, base))
 
+        self._create_symlinks()
+
+    def _create_symlinks(self):
+        if self.project.info.base not in ("core18", "core19"):
```
kyrofa: This is in a couple of places. Would it be better to just put this once in `__init__`?

tests/unit/plugins/test_ant.py
```diff
@@ -122,21 +200,120 @@ def test_env(self):
...
+                "-Dhttp.proxyHost=localhost -Dhttp.proxyPort=3132 "
+                "-Dhttp.proxyUser=user -Dhttp.proxyPassword=pass "
+                "-Dhttps.proxyHost=localhost2 -Dhttps.proxyPort=3133 "
+                "-Dhttps.proxyUser=user2 -Dhttps.proxyPassword=pass2"
+            ),
+        )
+
+
+class AntPluginUnsupportedBase(unit.TestCase):
+    def setUp(self):
```
kyrofa: Perhaps it would be better to have a `AntPluginBase` with a `setUp` like this that can be used for all tests in here? Then this one test can just make a `Project` instance instead of duplicating this.

snapcraft/plugins/ant.py
sergiusens: just copying from maven

snapcraft/plugins/ant.py
sergiusens: yeah, in maven too :-/
want me to remove it there?

snapcraft/plugins/ant.py
sergiusens: Legacy today most likely, when project_loader was called regardless, so if on a weird system we would fail as all the plugin's __init__'s were called. I say we keep the safe measures to not load the module until needed.

snapcraft/plugins/ant.py
sergiusens: To make porting easier, image you now come in with a fedora base.

tests/unit/plugins/test_ant.py
sergiusens: it is a different project options though

snapcraft/plugins/ant.py
kyrofa: That's actually exactly what I'm imagining: adding "fedora" to one of those tuples and forgetting the other one. Can we at least extract this to a method or some other way to only need to modify the list of supported bases once?

snapcraft/plugins/ant.py
kyrofa: Darn, I missed that. Might as well.

snapcraft/plugins/ant.py
kyrofa: Haha, how can this be legacy? You just added it :laughing: . The risk here is that future devs may try to use `self._ant_tar` from within `__init__` before `setup_ant()` is called, which will blow up.

snapcraft/plugins/ant.py
sergiusens: ok, but this goes into the TODO bucket.

snapcraft/plugins/ant.py
sergiusens: ok, but this goes into the TODO bucket, as this is the same pattern in:
- nodejs
- maven
- ant
- gradle

PR #2375, opened 2018-10-19
sanity checks: verify that command-chain is not used with legacy adapter

kyrofa:

-----

This PR fixes [LP: #1798918](https://bugs.launchpad.net/snapcraft/+bug/1798918) by introducing sanity checks that run in the build environment after all extensions have been applied, and ensuring that `command-chain` is not used in combination with the `legacy` app adapter (they are incompatible).


snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -608,10 +600,10 @@ def _wrap_apps(self, apps: Dict[str, Any]) -> Dict[str, Any]:
             if os.path.splitext(f)[1] == ".desktop":
                 os.remove(os.path.join(gui_dir, f))
         for app_name, app in apps.items():
-            adapter = Adapter[app.pop("adapter").upper()]
-            if adapter == Adapter.LEGACY:
+            adapter = project_loader.Adapter[app.pop("adapter").upper()]
```
sergiusens: what if adapter is empty, don't you want to default to "" instead of None so .upper() doesn't go nuts? This may be pre-validated elsewhere, but maybe still worth doing here.

kyrofa: Nah, the default is added literally right before this function is called. Check the PR where the Adapter enum is actually added (this one just moved it) for reference.

PR #2381, opened 2018-10-23
tools: copy in spread-shellcheck from snapd

chipaca:
We're considering packing spread-shellcheck into a snap, or maybe
seeing if shellcheck is interested, but meanwhile I thought I'd offer
it to you guys.

Here's how you use it:

1. `snap install --edge shellcheck`

2. `./tools/spread-shellcheck.py spread.yaml tests/spread/`

---
Most of the changes are minor; I purposefully didn't stop to think why you were comparing fixed strings using a regexp matcher. However if you search for `*/` and `*p` you'll find some tests I had to change a bit because of mis-using globs; I hope I understood what you thought you were doing :-)


tools/spread-shellcheck.py
```diff
@@ -0,0 +1,286 @@
+#!/snap/core/current/usr/bin/python3
```
sergiusens:
```suggestion
#!/usr/bin/env python3
```
In our case, we do not always have core installed.

chipaca: hmmm. I changed `.travis.yml` to install shellcheck from the snap, as the one in travis (and whatever you were using locally, apparently) had trouble when run via `xargs` (note how it didn't catch everything). So now `run_static_tests` has
```sh
find . ... blah | blah ... | xargs /snap/bin/shellcheck
```
if that change is ok, then this one would be ok as well :-)

of course the reason it's like this is because _this_ `spread-shellcheck` is copied from the snap I was preparing, which is classic but uses python3 from core...

but, you tell me.

.travis.yml
```diff
@@ -22,27 +22,32 @@ jobs:
   include:
     # Tests, only when not triggered by the daily cron.
     - stage: static
+      name: static tests
```
kyrofa: *gasp* what is this madness?

sergiusens: it is, it evens export the name into the environment to do neat things

tests/spread/general/license-valid/task.yaml
```diff
@@ -26,4 +26,4 @@ execute: |
 
   snapcraft  
 
-  cat meta/snap.yaml | MATCH -v "license: MIT"
+  MATCH -v "license: MIT" < meta/snap.yaml
```
sergiusens:
```
Snapped my-snap-name_0.1_amd64.snap
+ MATCH -v 'license: MIT'
/bin/bash: line 33: meta/snap.yaml: No such file or directory
```
You uncovered something here, the path should be prime/meta/snap.yaml

chipaca: well spotted, thanks!

tests/spread/plugins/python/stage-packages-deps/task.yaml
```diff
@@ -19,12 +21,13 @@ execute: |
...
-  python2="$(ls parts/python2/install/lib | grep python2)"
-  [ -d "parts/python2/install/lib/$python2/site-packages/yamllint" ]
-  [ -d "parts/python2/install/usr/lib/$python2/dist-packages/yaml" ]
-  [ ! -d "parts/python2/install/lib/$python2/site-packages/yaml" ]
-
-  python3="$(ls parts/python3/install/lib | grep python3)"
-  [ -d "parts/python3/install/lib/$python3/site-packages/yamllint" ]
-  [ -d "parts/python3/install/usr/lib/python3/dist-packages/yaml" ]
-  [ ! -d "parts/python3/install/lib/$python3/site-packages/yaml" ]
+  for d in parts/python2/install/lib/*python2*; do
```
sergiusens: this now fails...
```
Collecting yamllint
Collecting pathspec
Installing collected packages: yamllint, pathspec
Successfully installed pathspec-0.5.9 yamllint-1.12.1
+ for d in parts/python2/install/lib/*python2*
+ test -d parts/python2/install/lib/python2.7/site-packages/yamllint
+ test -d parts/python2/install/lib/python2.7/dist-packages/yaml
-----
```

PR #2385, opened 2018-10-24
cli: consolidate re-execution

sergiusens:
Re-execution was done in two locations, minimize that by moving the
re-execution logic back into a console_scripts entry and leaving the
root user handling to multipass' implementation of the build provider
APIs.

LP: #1795680
LP: #1799780

-----


snapcraft/cli/_runner.py
```diff
@@ -52,6 +53,38 @@
 ]
 
 
+def _is_legacy_reexec() -> bool:
```
kyrofa: I think `_is_legacy` would read better, the function itself doesn't have anything to do with re-execing.

snapcraft/cli/_runner.py
```diff
@@ -62,18 +95,8 @@
...
-        # Use -E to keep the environment.
-        cmd = ["usr/bin/sudo", "-E"]
-        # Pick the correct interpreter if running from a snap (in managed-host mode, this
-        # is most likely the only scenario).
-        if os.getenv("SNAP_NAME") == "snapcraft":
-            cmd.append(os.path.join(snap_path, "usr", "bin", "python3"))
-        cmd.extend(sys.argv)
-        os.execve("/usr/bin/sudo", cmd, os.environ)
+    if _is_legacy_reexec():
+        _legacy_reexec()
```
kyrofa: Given that this calls os.execv, mind adding a comment here that this will not return so future readers don't get confused?

kyrofa:
I've been testing this out, and it works quite well. The only odd thing is this: when I run `snapcraft --debug` and end up in the VM, if I run ls -l this is what I see:

```
snapcraft-my-snap-name # ls -l
total 24
drwxr-xr-x 3 root      root      4096 Oct 24 21:46 parts
drwxr-xr-x 3 root      root      4096 Oct 24 21:46 prime
drwxr-xr-x 1 multipass multipass 4096 Oct 24 21:48 project
drwxr-xr-x 3 root      root      4096 Oct 24 21:46 snap
drwxr-xr-x 2 root      root      4096 Oct 24 21:46 stage
-rw-r--r-- 1 root      root        72 Oct 24 21:50 state
```

Uh oh... what happens if I make a new file as root in `project/`?

```
snapcraft-my-snap-name # touch project/file
snapcraft-my-snap-name # ls -l project/
total 12
-rw-r--r-- 1 multipass multipass   11 Oct 24 21:52 file
-rw-r--r-- 1 multipass multipass 4096 Oct 24 21:46 my-snap-name_0.1_amd64.snap
drwxr-xr-x 1 multipass multipass 4096 Oct 24 21:47 snap
```

Wait, it's still owned by multipass. That seems odd to me-- would it be difficult to just make that root?

My last complaint: I'm root. I can run `apt update` and `apt install` without sudo. But the only thing that actually tells me that is the "#" in the prompt, and if I happen to run `whoami` or something. How would you feel about adding the user back to PS1?

PR #2398, opened 2018-11-08
gnome extension

diddledan:
> I discussed this with @kyrofa directly at snapcraft summit.
> I have successfully run unit and spread tests relating to this feature.
-----

Please review with prejudice :-)


extensions/desktop-common/desktop-exports
```diff
@@ -0,0 +1,436 @@
...
+    source "${XDG_CONFIG_HOME:-$HOME/.config}/user-dirs.dirs"
+  fi
+
+  if [ -z "${XDG_SPECIAL_DIRS+x}" ]; then
+    save_initial_values=true
+  fi
+
+  for d in $XDG_DIRS; do
+    var="XDG_${d}_DIR"
+    value="$(eval echo "\$${var}")"
```
kyrofa: You're missing some improvements from the work done for the [glib extension](https://github.com/snapcore/snapcraft/pull/2359).

kyrofa:
I think it's important for us to start small, here. Rather than simply turning the desktop helper into an extension, we should (at least somewhat) cleanroom it. Create a snap of a few gnome applications with no remote parts or extensions, and see what common bits fall out.

kenvandine:
This extension should also connect to the platform snap's content interface.  Ideally, for core16 connect to gnome-3-26-1604 and for core18 connect to gnome-3-28-1804.

kenvandine:
If I include the environment keyword in my yaml with LD_LIBRARY_PATH it will override the LD_LIBRARY_PATH set by the gnome extension yaml.  This wasn't an issue when the desktop-launch script would set these variables.  I think it would be best to move LD_LIBRARY_PATH (and other runtime variables) to the command-chain.

snapcraft/internal/project_loader/_extensions/gnome.py
```diff
@@ -0,0 +1,167 @@
...
+        dependency_part = {}  # type: Dict[str, Any]
+        if base is not None:
+            if base == "core16":
+                platform_snap = "gnome-3-26-1604"
+                after_dependencies = ["gnome-extension-platform-dependencies"]
+                dependency_part = {
+                    "gnome-extension-platform-dependencies": {
+                        "plugin": None,
+                        "override-pull": """
+                            add-apt-repository ppa:
```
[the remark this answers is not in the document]
diddledan: fixed. I failed to remember to fill that before pushing :-)

snapcraft/internal/project_loader/_extensions/gnome.py
```diff
@@ -0,0 +1,167 @@
...
+        base = yaml_data.get("base")
+        after_dependencies = []  # type: List[str]
+        dependency_part = {}  # type: Dict[str, Any]
+        if base is not None:
+            if base == "core16":
+                platform_snap = "gnome-3-26-1604"
+                after_dependencies = ["gnome-extension-platform-dependencies"]
+                dependency_part = {
+                    "gnome-extension-platform-dependencies": {
+                        "plugin": None,
```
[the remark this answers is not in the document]
diddledan: fixed. I figured `None` would turn into `nil` in the yaml but I misunderstood that `nil` was actually a string and not `nothing`.

extensions/desktop-common/desktop-init
```diff
@@ -0,0 +1,390 @@
...
+    source "${XDG_CONFIG_HOME:-$HOME/.config}/user-dirs.dirs"
+  fi
+
+  if [ -z ${XDG_SPECIAL_DIRS+x} ]; then
+    save_initial_values=true
+  fi
+
+  for d in $XDG_DIRS; do
+    var="XDG_${d}_DIR"
+    value="$(eval echo \$"${var}")"
```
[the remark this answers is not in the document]
diddledan: fixed.

extensions/desktop-common/desktop-exports
diddledan: fixed.

extensions/desktop-common/desktop-common
```diff
@@ -0,0 +1,434 @@
+
```
oSoMoN: desktop-common needs to be marked executable (snapcraft will fail otherwise)

extensions/desktop-common/desktop-common
```diff
@@ -0,0 +1,434 @@
...
+export GST_PLUGIN_SYSTEM_PATH="$SNAP_DESKTOP_RUNTIME/usr/lib/$SNAP_LAUNCHER_ARCH_TRIPLET/gstreamer-1.0"
+# gst plugin scanner doesn't install in the correct path: https://github.com/ubuntu/snapcraft-desktop-helpers/issues/43
+export GST_PLUGIN_SCANNER="$SNAP_DESKTOP_RUNTIME/usr/lib/$SNAP_LAUNCHER_ARCH_TRIPLET/gstreamer1.0/gstreamer-1.0/gst-plugin-scanner"
+
+# XDG Config
+[ "$WITH_SNAP_DESKTOP_RUNTIME" = yes ] && prepend_dir XDG_CONFIG_DIRS "$SNAP_DESKTOP_RUNTIME/etc/xdg"
+prepend_dir XDG_CONFIG_DIRS "$SNAP/etc/xdg"
+
+# Define snaps' own data dir
+[ "$WITH_SNAP_DESKTOP_RUNTIME" = yes ] && prepend_dir XDG_DATA_DIRS "$SNAP_DESKTOP_WAYLAND_AVAILABLE/usr/share"
```
oSoMoN: Wrong variable? That should probably be `"$SNAP_DESKTOP_RUNTIME/usr/share"`. Same in various other places throughout this file.

diddledan: oh dear. *facepalm* silly search-replace fail!

extensions/desktop-common/desktop-common
diddledan: it is now executable :-)

extensions/desktop-common/desktop-common
```diff
@@ -0,0 +1,434 @@
...
+# Testability support
+append_dir LD_LIBRARY_PATH "$SNAP/testability"
+append_dir LD_LIBRARY_PATH "$SNAP/testability/$SNAP_LAUNCHER_ARCH_TRIPLET"
+append_dir LD_LIBRARY_PATH "$SNAP/testability/$SNAP_LAUNCHER_ARCH_TRIPLET/mesa"
+
+# Gdk-pixbuf loaders
+if [ "$SNAP_DESKTOP_COMPONENTS_NEED_UPDATE" = "true" ]; then
+  rm -f "$GDK_PIXBUF_MODULE_FILE"
+  if [ -f "$SNAP_DESKTOP_RUNTIME/usr/lib/$SNAP_LAUNCHER_ARCH_TRIPLET/gdk-pixbuf-2.0/gdk-pixbuf-query-loaders" ]; then
+    "$SNAP_DESKTOP_RUNTIME/usr/lib/$SNAP_LAUNCHER_ARCH_TRIPLET/gdk-pixbuf-2.0/gdk-pixbuf-query-loaders" > "$GDK_PIXBUF_MODULE_FILE"
```
oSoMoN: `GDK_PIXBUF_MODULE_FILE` is not defined anywhere. When I build and run a test snap, I get a "No such file or directory" error, and the launcher stops here.

extensions/desktop-common/desktop-init
```diff
@@ -0,0 +1,54 @@
...
+# If the user has modified their user-dirs settings, force an update
+if [[ -f "$XDG_CONFIG_HOME/user-dirs.dirs.md5sum" && -f "$XDG_CONFIG_HOME/user-dirs.locale.md5sum" ]]; then
+  if [[ "$(md5sum < "$REALHOME/.config/user-dirs.dirs")" != "$(cat "$XDG_CONFIG_HOME/user-dirs.dirs.md5sum")" ||
+        "$(md5sum < "$REALHOME/.config/user-dirs.locale")" != "$(cat "$XDG_CONFIG_HOME/user-dirs.locale.md5sum")" ]]; then
+    SNAP_DESKTOP_COMPONENTS_NEED_UPDATE="true"
+  fi
+fi
+
+# Create config folder
+mkdir -p "$XDG_CONFIG_HOME" && chmod 700 "$XDG_CONFIG_HOME"
```
oSoMoN: At this point in the command chain, `XDG_CONFIG_HOME` hasn't been defined yet, and this results in:

    mkdir: cannot create directory '': No such file or directory

oSoMoN:
I rebuilt the gnome-calculator snap with your latest changes, and this seems to work fine.
I am now in the process of rebuilding the libreoffice snap to test this further.

oSoMoN:
I successfully rebuilt the libreoffice snap, and for the most part it works, but I've noticed a couple of regressions:
 - ibus input doesn't work (tested with ibus-pinyin). I unpacked the snap and added the contents of ibus-gtk3 and libibus-1.0, but this doesn't seem to be enough.
 - Some elements in the GTK file open/save dialogs are not localized (buttons in the header bar, column headers in the list view). The libreoffice snap has a [gtk3-locales part](https://git.launchpad.net/~libreoffice/+git/libreoffice-snap/tree/snapcraft.yaml?h=6.1-gnome-extension#n356) that should make this work, but when built with the gnome extension something appears to be missing.

extensions/desktop-common/desktop-common
```diff
@@ -0,0 +1,436 @@
...
+
+# XKB config
+export XKB_CONFIG_ROOT="$SNAP_DESKTOP_RUNTIME/usr/share/X11/xkb"
+
+# Give XOpenIM a chance to locate locale data.
+# This is required for text input to work in SDL2 games.
+export XLOCALEDIR="$SNAP_DESKTOP_RUNTIME/usr/share/X11/locale"
+
+# Set XCursors path
+export XCURSOR_PATH="$SNAP_DESKTOP_RUNTIME/usr/share/icons"
```
oSoMoN: This is missing a recent fix in the desktop helpers: https://github.com/ubuntu/snapcraft-desktop-helpers/commit/da4d8f878eaeb40c1788789cd034b7142f7d6749

extensions/desktop-common/desktop-common
```diff
@@ -0,0 +1,436 @@
...
+
+# Mesa Libs for OpenGL support
+append_dir LD_LIBRARY_PATH "$SNAP_DESKTOP_RUNTIME/usr/lib/$SNAP_LAUNCHER_ARCH_TRIPLET/mesa"
+append_dir LD_LIBRARY_PATH "$SNAP_DESKTOP_RUNTIME/usr/lib/$SNAP_LAUNCHER_ARCH_TRIPLET/mesa-egl"
+
+# Tell libGL where to find the drivers
+export LIBGL_DRIVERS_PATH="$SNAP_DESKTOP_RUNTIME/usr/lib/$SNAP_LAUNCHER_ARCH_TRIPLET/dri"
+append_dir LD_LIBRARY_PATH "$LIBGL_DRIVERS_PATH"
```
oSoMoN: This is missing a recent fix in the desktop helpers: https://github.com/ubuntu/snapcraft-desktop-helpers/commit/cafb789b48dc572f17fea630df18e93f507a7dff

extensions/desktop-common/desktop-common
```diff
@@ -0,0 +1,436 @@
...
+  done
+}
+
+function is_subpath() {
+  dir="$(realpath "$1")"
+  parent="$(realpath "$2")"
+  [ "${dir##$parent/}" != "$dir" ] && return 0 || return 1
+}
+
+WITH_RUNTIME=no
```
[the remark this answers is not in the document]
diddledan: I've fixed the duplicated functionality and replaced any tests for `$WITH_RUNTIME` with appropriate alternative tests for `$WITH_SNAP_DESKTOP_RUNTIME`

extensions/gnome/desktop-gnome-specific
```diff
@@ -0,0 +1,30 @@
...
+##############################
+# GTK launcher specific part #
+##############################
+
+if [ "$SNAP_DESKTOP_WAYLAND_AVAILABLE" = "true" ]; then
+  export GDK_BACKEND="wayland"
+  export CLUTTER_BACKEND="wayland"
+  # Does not hurt to specify this as well, just in case
+  export QT_QPA_PLATFORM=wayland-egl
+  export GTK_PATH="$SNAP_DESKTOP_RUNTIME/usr/lib/$SNAP_LAUNCHER_ARCH_TRIPLET/gtk-3.0"
```
oSoMoN: This is not wayland-specific. Moving the export of `$GTK_PATH` outside of this block fixes ibus input in the libreoffice snap.

extensions/desktop-common/desktop-init
```diff
@@ -0,0 +1,55 @@
...
+fi
+
+SNAP_LAUNCHER_ARCH_TRIPLET="$ARCH"
+
+if [ -n "$SNAP_DESKTOP_RUNTIME" ]; then
+  WITH_SNAP_DESKTOP_RUNTIME="yes"
+fi
+
+if [ -f "$SNAP/lib/bindtextdomain.so" ]; then
+  export LD_PRELOAD="$LD_PRELOAD:$SNAP/lib/bindtextdomain.so"
```
oSoMoN: bindtextdomain.so is not built. I guess the extension needs to add a part that builds it.

diddledan: that's a direct copy+pasta from the desktop-helpers. Is `bindtextdomain.so` in the platform snap I wonder...

diddledan: aha. found it :-) added in 47973a98.

diddledan:
sorry about the double force-push, I missed off the new bindtextdomain source files :-)

oSoMoN:
When building the desktop-common-bindtextdomain part, I get the following error:

    Building desktop-common-bindtextdomain 
    make -j4
    rm -f bindtextdomain.so
    make install DESTDIR=/mnt/libreoffice/libreoffice-snap/parts/desktop-common-bindtextdomain/install
    gcc -Wall -O2 -o bindtextdomain.so -fPIC -shared /bindtextdomain.c -ldl
    gcc: error: /bindtextdomain.c: No such file or directory
    Makefile:9: recipe for target 'bindtextdomain.so' failed
    make: *** [bindtextdomain.so] Error 1
    Failed to run 'make install DESTDIR=/mnt/libreoffice/libreoffice-snap/parts/desktop-common-bindtextdomain/install' for 'desktop-common-bindtextdomain': Exited with code 2.

It looks like `$SRC_DIR` is not defined.

diddledan:
> Simply removing the `$(SRC_DIR)/` prefix from `extensions/bindtextdomain/Makefile` fixes the error I reported above, the libreoffice snap builds successfully and everything I tested works as expected.

I've pushed this change. Shall I go ahead, now, and squash the commits ready for merge?

oSoMoN:
> I've pushed this change. Shall I go ahead, now, and squash the commits ready for merge?

I'd like to rebuild and test a few more snaps to ensure that everything works well (the results on gnome-calculator and libreoffice are promising) before declaring it ready to be merged, but you can go ahead and squash the commits.

oSoMoN:
I'm currently rebuilding the chromium snap with the gnome extension, this will take a while to complete, I'll report back when done.
Can you suggest other interesting snaps to test against this PR ?

extensions/bindtextdomain/bindtextdomain.c
```diff
@@ -0,0 +1,119 @@
...
+        if (dirname == NULL || dirname[0] != '/')
+                goto orig;
+
+        snap = getenv ("SNAP");
+
+        if (snap == NULL || strcmp (snap, "") == 0)
+                goto orig;
+
+        char * paths[] = {
+                "gnome-platform/usr/share/locale",
```
oSoMoN: Please add "gnome-platform/usr/share/locale-langpack" to the list of paths. @kenvandine included the gtk3 translations in the gnome-3-28-1804 snap (currently in the candidate channel) so that individual snaps don't have to do it.

oSoMoN: Please add "gnome-platform/usr/share/locale-langpack" to the list of paths. @kenvandine included the gtk3 translations in the gnome-3-28-1804 snap (currently in the candidate channel) so that individual snaps don't have to do it.

diddledan: added in d3afb53

kenvandine:
Something else the extension should provide a layout for, iso-codes

+  layout:
+    /usr/share/xml/iso-codes:
+      symlink: $SNAP/gnome-platform/usr/share/xml/iso-codes
+

diddledan:
Ken: I've added that layout definition and updated the branch to match master.
Sergio: it looks like travis failed on the CLA check after my merge of master pulling in every user ever some of whom seem to have not signed the CLA.

sergiusens:
I think you confused the light weight system by force pushing followed by a merge

diddledan:
> force pushing followed by a merge

I like doing things weirdly :-p

kenvandine:
Before we merge this, we need to have a discussion about handling different versions of the gnome platform.  We're going to need a way for snap packagers to use newer versions of libraries.  This will result in newer platform snaps.  For example we'll probably end up with a gnome-3-32-1804 platform snap soon that includes newer glib, gtk, etc.  

Perhaps we should name the extension accordingly?  gnome-3-28-1084 and later add more as we have new platform snaps available, always matching the platform snap name?

cmatsuoka:
Also it should be discussed what happens when a new base is available and the user tries to build on one base when the platform was built with a different base.

snapcraft/internal/project_loader/_extensions/gnome.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+class GnomeExtension(DesktopCommonExtension):
+    """The Gnome extension.
+    This extension is to be used by applications that require GTK+.
+    Examples might include productivity applications or utilities.
+    Note that this extension does not support classically-confined snaps at this time.
+    """
+
+    supported_bases = ("core16", "core18")
```
sergiusens: Potentially missing feature
supported_architectures = ....

kenvandine:
> Before we merge this, we need to have a discussion about handling different versions of the gnome platform. We're going to need a way for snap packagers to use newer versions of libraries. This will result in newer platform snaps. For example we'll probably end up with a gnome-3-32-1804 platform snap soon that includes newer glib, gtk, etc.
> 
> Perhaps we should name the extension accordingly? gnome-3-28-1084 and later add more as we have new platform snaps available, always matching the platform snap name?

As discussed this week, we will rename this to gnome-3-28 and infer the platform version from the base.  For example, if base is core18 we'll connect to the gnome-3-28-1804 platform snap.  In the future, we'll add support for additional gnome versions like gnome-3-32 which would connect to gnome-3-32-1804.  This also allows us to handle other bases, like fedora perhaps which could use a different platform snap or even just stage packages.

We also discussed dropping support for core, only supporting core18 from the start.  If there is enough demand we can discuss adding support for core.

extensions/desktop-common/desktop-init
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash -e
+
+#################
+# Launcher init #
+#################
+
+# On Fedora $SNAP is under /var and there is some magic to map it to /snap.
+# # We need to handle that case and reset $SNAP
+SNAP="${SNAP/\/var\/lib\/snapd/}"
```
sergiusens: this is not exported nor used in this script, might need to be moved or exported

oSoMoN:
I successfully rebuilt the chromium snap with this extension and published it under the `candidate/gnome-3-28-extension` channel (amd64 only).
Successfully tested on Ubuntu 18.04 and Fedora 28.

oSoMoN:
@diddledan please consider backporting https://github.com/ubuntu/snapcraft-desktop-helpers/pull/176.

extensions/desktop-common/desktop-init
```diff
@@ -0,0 +1,57 @@
...
+elif [ "$SNAP_ARCH" == "ppc64el" ]; then
+  ARCH="powerpc64le-linux-gnu"
+else
+  ARCH="$SNAP_ARCH-linux-gnu"
+fi
+
+SNAP_LAUNCHER_ARCH_TRIPLET="$ARCH"
+
+if [ -n "$SNAP_DESKTOP_RUNTIME" ]; then
+  WITH_SNAP_DESKTOP_RUNTIME="yes"
```
niemeyer: My apologies in advance, as I lack quite a bit of context on this review, so I'll be making some silly questions.

In the case at hand, why do we have a variable stating that another variable is set?  Isn't the first variable being set in the first place enough?

extensions/desktop-common/desktop-init
```diff
@@ -0,0 +1,57 @@
...
+fi
+
+if [ -f "$SNAP/lib/bindtextdomain.so" ]; then
+  export LD_PRELOAD="$LD_PRELOAD:$SNAP/lib/bindtextdomain.so"
+fi
+
+export REALHOME
+export WITH_SNAP_DESKTOP_RUNTIME
+export SNAP_DESKTOP_COMPONENTS_NEED_UPDATE
+export SNAP_LAUNCHER_ARCH_TRIPLET
```
niemeyer: It seems worth picking a single prefix, and sticking with it, so that all content from the desktop extension is under a unique, clear, and hopefully safe namespace.

In that sense, most things are using $SNAP_DESKTOP_* here, which follows that and is a nice and readable pattern. $SNAP_LAUNCHER_* breaks that, though. Can we use $SNAP_DESKTOP_* here too?

$SNAP_* alone risks getting into interaction trouble between internal snapd logic.

snapcraft/internal/project_loader/_extensions/_desktop_common.py
```diff
@@ -0,0 +1,80 @@
...
+from typing import Any, Dict
+
+from ._extension import Extension
+
+
+class DesktopCommonExtension(Extension):
+    """The Common Desktop extension.
+    This extension is to be subclassed by the gnome and kde extensions.
+    It includes common code that each desktop extension would otherwise
+    need to duplicate.
```
niemeyer: Nice abstraction.

Also, the whole file here feels quite clean and readable. It's not clear how the underlying Extension type will handle that data (will it override or append? when? etc), but it looks good indeed.

extensions/desktop-common/desktop-init
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash -e
+
+#################
+# Launcher init #
+#################
+
+# On Fedora $SNAP is under /var and there is some magic to map it to /snap.
+# # We need to handle that case and reset $SNAP
+SNAP="${SNAP/\/var\/lib\/snapd/}"
```
niemeyer: This seems very strange on first sight. There are reasons why $SNAP points elsewhere in Fedora, and those reasons are trusted by snaps in general. That change breaks the common interface of snapd itself, and can easily break down other things in the system. On the other hand, internally in the snap, /snap still exists, and $SNAP should point to /snap _already_ if it's not a classic snap, so if that's not the case today we have a bug in snapd. If it is a classic snap, then it points elsewhere because /snap may not exist, which means there's not a good reason for it to point there anyway.

Can someone please clarify these issues with the snapd team?

extensions/desktop-common/desktop-init
```diff
@@ -0,0 +1,57 @@
...
+if [ -f "$SNAP/lib/bindtextdomain.so" ]; then
+  export LD_PRELOAD="$LD_PRELOAD:$SNAP/lib/bindtextdomain.so"
+fi
+
+export REALHOME
+export WITH_SNAP_DESKTOP_RUNTIME
+export SNAP_DESKTOP_COMPONENTS_NEED_UPDATE
+export SNAP_LAUNCHER_ARCH_TRIPLET
+
+exec "$@"
```
niemeyer: Besides these minor comments, the code overall in this wrapper feels reasonable and easy to maintain working over time.

snapcraft/internal/project_loader/_extensions/gnome-3-28.py
```diff
@@ -0,0 +1,111 @@
...
+        if yaml_data.get("confinement") == "classic":
+            raise GnomeExtensionClassicConfinementError()
+
+        platform_snap = ""
+        base = yaml_data.get("base")
+        after_dependencies = []  # type: List[str]
+        dependency_part = {}  # type: Dict[str, Any]
+        if base is not None:
+            if base == "core18":
+                platform_snap = "gnome-3-28-1804"
```
niemeyer: What happens if it's not? Should it raise an error to make sure this won't go out of sync with supported_bases above, which I assume is also being enforced?

snapcraft/internal/project_loader/_extensions/gnome-3-28.py
```diff
@@ -0,0 +1,111 @@
...
+        super().__init__(yaml_data)
+
+        if yaml_data.get("confinement") == "classic":
+            raise GnomeExtensionClassicConfinementError()
+
+        platform_snap = ""
+        base = yaml_data.get("base")
+        after_dependencies = []  # type: List[str]
+        dependency_part = {}  # type: Dict[str, Any]
+        if base is not None:
```
niemeyer: This seems irrelevant given what's below.

snapcraft/internal/project_loader/_extensions/gnome-3-28.py
```diff
@@ -0,0 +1,111 @@
...
+                },
+            },
+            "environment": {
+                **self.environment,
+                "SNAP_DESKTOP_RUNTIME": "$SNAP/gnome-platform",
+            },
+        }
+
+        command_chain = self.app_snippet["command-chain"]
+        exec_command = command_chain.pop()
```
niemeyer: This logic seems fragile. It assumes the implementation of a complete separate piece which can change somewhat easily. We might drop the command from the base, and put it here by name explicitly instead.

snapcraft/internal/project_loader/_extensions/gnome-3-28.py
```diff
@@ -0,0 +1,111 @@
...
+
+        command_chain = command_chain + [
+            "snap/command-chain/desktop-gnome-specific",
+            exec_command,
+        ]
+
+        self.app_snippet = {
+            **self.app_snippet,
+            "command-chain": command_chain,
+            "adapter": "full",
```
niemeyer: What does that mean nowadays? I thought adapters were gone now that we have the command-chain.

snapcraft/internal/project_loader/_extensions/gnome-3-28.py
```diff
@@ -0,0 +1,111 @@
...
+
+from snapcraft.internal import errors
+from ._desktop_common import DesktopCommonExtension
+
+
+class GnomeExtensionClassicConfinementError(errors.SnapcraftError):
+    fmt = "The gnome extension doesn't support classic confinement."
+
+
+class Gnome_3_28Extension(DesktopCommonExtension):
```
niemeyer: This is a question for the general design of extensions in Snapcraft: is the gnome extension able to cope with a different and unrelated extension below it, or does it assume that it's it override any other extension command-chain, etc?

snapcraft/internal/project_loader/_extensions/gnome-3-28.py
diddledan: I like your idea. I felt dirty when I wrote that :-p

snapcraft/internal/project_loader/_extensions/gnome-3-28.py
diddledan: I'd like some guidance from @sergiusens about the feasibility of combining extensions. I.e. does snapcraft have the capability to do so or does the yaml schema guarantee only a single extension per command? Also, what is the likelihood of our creating multiple extensions with orthogonal behaviour that would warrant the inclusion of more than one in a project? As it stands with the plan to implement a KDE and a Gnome extension these two are almost exclusive to each other where they are doing very similar things tailored specifically for each desktop environment, so we are unlikely to (but not inevitably) include both in a single project.

snapcraft/internal/project_loader/_extensions/gnome-3-28.py
diddledan: yes, you're quite right. I do think adding an extra exception here would be a good idea. Thanks :-)

extensions/desktop-common/desktop-init
diddledan: This bit of code is from the [ubuntu/snapcraft-desktop-helpers](https://github.com/ubuntu/snapcraft-desktop-helpers) repo. I don't know why it's needed but I also don't know for certain that it isn't, so I left it in place.

extensions/desktop-common/desktop-init
diddledan: This is a throwback - I improved the code surrounding the use of `$SNAP_DESKTOP_RUNTIME` but didn't come back to swap out the tests and this variable. 364960e1 fixes it :-)

extensions/desktop-common/desktop-init
diddledan: I've renamed `$SNAP_LAUNCHER_ARCH_TRIPLET` to `$SNAP_DESKTOP_ARCH_TRIPLET` and made relevant updates to the other scripts to suit via 364960e1

snapcraft/internal/project_loader/_extensions/gnome-3-28.py
diddledan: I've added an exception when the check falls-through. I also moved the exception for classic confinement into the extension loader itself rather than as extension-specific code. It made sense to allow other extensions to have declarative support for denying classic confinement. These are both in 3f5ec4f2.

snapcraft/internal/project_loader/_extensions/gnome-3-28.py
diddledan: gone *poof* like magic :-) in 31b92f50.

snapcraft/internal/project_loader/_extensions/gnome-3-28.py
diddledan: sorted this out in ab206008. Now the last item which I was popping off the list is gone entirely in favour of duplicating the three lines in one script for each of the desktop extensions. (If we follow my current pattern of a single `desktop-gnome-specific` script then the three lines go right before the `exec`.

snapcraft/internal/project_loader/_extensions/_desktop_common.py
diddledan: The idea now is to just append. The base class settings will be accessible in the subclasses allowing them to either replace or augment as they desire. The way I have done this in the gnome extension is to append the gnome-specific values via code like:

```python
        self.parts = {
            **self.parts,
            "gnome-extension": {
                "after": after_dependencies,
                "plugin": "dump",
                "source": "$SNAPCRAFT_EXTENSIONS_DIR/gnome",
                "source-type": "local",
                "organize": {"desktop-*": "snap/command-chain/"},
                "build-packages": ["build-essential", "libgtk-3-dev"],
            },
        }
```

Here you see I directly reference the `self.parts` when constructing a replacement value to insert the previous content as-is into the new value. This results in the equivalent combined definition of:

```python
        self.parts = {
            "desktop-common-extension": { # expanded from **self.parts when setting new value
                "plugin": "dump",
                "source": "$SNAPCRAFT_EXTENSIONS_DIR/desktop-common",
                "organize": {"desktop-*": "snap/command-chain/"},
            },
            "desktop-common-bindtextdomain": { # expanded from **self.parts when setting new value
                "plugin": "make",
                "source": "$SNAPCRAFT_EXTENSIONS_DIR/bindtextdomain",
            },
            "gnome-extension": { # defined explicitly when setting new value
                "after": after_dependencies,
                "plugin": "dump",
                "source": "$SNAPCRAFT_EXTENSIONS_DIR/gnome",
                "source-type": "local",
                "organize": {"desktop-*": "snap/command-chain/"},
                "build-packages": ["build-essential", "libgtk-3-dev"],
            },
        }
```

snapcraft/internal/project_loader/errors.py
```diff
@@ -84,6 +84,17 @@ def __init__(self, extension_name: str, base: str) -> None:
         super().__init__(extension_name=extension_name, base=base)
 
 
+class ExtensionUnsupportedClassicError(ProjectLoaderError):
+    fmt = (
+        "Failed to load extension {extension_name!r}: "
+        "this extensino does not support classic confinement.\n"
```
cjp256: extensino => extension

extensions/bindtextdomain/bindtextdomain.c
```diff
@@ -0,0 +1,120 @@
...
+                "usr/share/locale-langpack",
+                NULL
+        };
+        for(int i = 0; paths[i] != NULL; i++)
+        {
+                if (asprintf (&snap_path, "%s/%s", snap, paths[i]) < 0)
+                        continue;
+
+                if (access (snap_path, F_OK) < 0)
+                        continue;
```
cjp256: Leaking snap_path?

extensions/bindtextdomain/bindtextdomain.c
```diff
@@ -0,0 +1,120 @@
...
+                                free (snap_locale_path);
+                                snap_locale_path = NULL;
+                                continue;
+                        }
+                }
+
+                closedir (dir);
+                free (snap_path);
+                snap_path = NULL;
+                free (snap_locale_path);
```
cjp256: I don't think this is required. Looks like a potential double free (though snap_locale_path would be NULL as it is explicitly freed and reset in above loop on all paths after allocation).

extensions/bindtextdomain/bindtextdomain.c
```diff
@@ -0,0 +1,120 @@
...
+                        continue;
+
+                /*
+                 * if the mo file exists for one language we assume it exists for them
+                 * all, or at least that we're not going to find it anywhere else. we
+                 * don't know at this point what locale the application is actually
+                 * going to use, so we can't look in any particular directory.
+                 */
+                dir = opendir (snap_path);
+                if (dir == NULL)
```
cjp256: Leaking snap_path?

extensions/bindtextdomain/bindtextdomain.c
```diff
@@ -0,0 +1,120 @@
...
+        goto orig;
+
+ok:
+        ret = r_bindtextdomain (domainname, snap_path);
+        goto out;
+orig:
+        ret = r_bindtextdomain (domainname, dirname);
+        goto out;
+out:
+        free (snap_path);
```
cjp256: Maybe already freed/NULL.

snapcraft/internal/project_loader/_extensions/gnome-3-28.py
sergiusens: Interesting question, the answer is in your PR https://github.com/snapcore/snapcraft/pull/2398/files#diff-3a1ce052529b9a6c68d667ca4b50c02dR554 that is, multiple extension are indeed allowed.

Now for handling extension conflicts and parametrization, extensions have access to yaml_data, so they know what other extensions are used and should fallback in behavior or error out when needed (we do not have that problem yet as we have no extensions, but we can control the story).

When implemented, it was implemented as a stream (iterate and apply each extension), instead of loading all of the extensions and creating a final processor out of that (https://github.com/snapcore/snapcraft/blob/master/snapcraft/internal/project_loader/_extensions/_utils.py#L43)

extensions/desktop-common/desktop-init
niemeyer: If we are going to err, let's err on the safe side and drop this logic. Per rationale above, this is a fundamental setting made available to every snap by snapd. It shouldn't be hacked like this by first-class command chain scripts.

snapcraft/internal/project_loader/_extensions/gnome-3-28.py
niemeyer: > so they know what other extensions are used

A key reason we designed extensions the way we did is composability. It shouldn't matter if the data was entered into snapcraft.yaml by hand, or by an extension that preceded the extension currently being processed. In both cases, the result is the snapcraft.yaml data that is supposed to be handled by the extension at hand. If there are conflicts, it should conflict with the data, not with the extension name that is in the list.

sergiusens:
El domingo, 24 de marzo de 2019 12:54:44 -03 Daniel Llewellyn escribió:
> diddledan commented on this pull request.
> 
> > +        Note that this extension does not support classic snaps.
> 
> +        :param dict yaml_data: Loaded snapcraft.yaml data.
> +        """
> +
> +        super().__init__(yaml_data)
> +
> +        if yaml_data.get("confinement") == "classic":
> +            raise GnomeExtensionClassicConfinementError()
> +
> +        platform_snap = ""
> +        base = yaml_data.get("base")
> +        after_dependencies = []  # type: List[str]
> +        dependency_part = {}  # type: Dict[str, Any]
> +        if base is not None:
> +            if base == "core18":
> +                platform_snap = "gnome-3-28-1804"
> 
> yes, you're quite right. I do think adding an extra exception here would be
> a good idea. Thanks :-)

base cannot be none, the schema wouldn't allow it, the supported bases are 
taken into account on load, but it is good to be defensive.

niemeyer:
> base cannot be none, the schema wouldn't allow it, the supported bases are taken into account on load, but it is good to be defensive.

The logic here can still be broken regardless, by adding a base to supported_bases without reflecting what the respective platform snap should be. It is defensiveness because the logic is correct as is, but it's critical defensiveness rather than redundant because it's easy to change the logic and produce silent non-working code otherwise.

snapcraft/internal/project_loader/_extensions/gnome-3-28.py
sergiusens: That will be need to be taken into account here https://github.com/snapcore/snapcraft/blob/master/snapcraft/internal/project_loader/_extensions/_utils.py#L70 where the original_yaml is used instead of the processed one.

@kyrofa, given you were more involved with extensions, do you want to share your thoughts?

sergiusens:
> The logic here can still be broken regardless, by adding a base to supported_bases without reflecting what the respective platform snap should be. It is defensiveness because the logic is correct as is, but it's critical defensiveness rather than redundant because it's easy to change the logic and produce silent non-working code otherwise.

I was agreeing with the requirement, we need to have the right guards and be defensive so there are no surprises :smile:

snapcraft/internal/project_loader/_extensions/gnome-3-28.py
diddledan: Given my updated code in 364960e and ab20600, I think the Gnome extension can properly cope with being included with other extensions, but I'm not 100% certain. The bit I'm unsure about is how multiple extensions' command-chains will be handled/merged.

extensions/desktop-common/desktop-common
```diff
@@ -0,0 +1,434 @@
...
+  for d in $XDG_DIRS; do
+    var="XDG_${d}_DIR"
+    value="${!var}"
+
+    if [ "$save_initial_values" = true ]; then
+      XDG_SPECIAL_DIRS+=("$var")
+      XDG_SPECIAL_DIRS_INITIAL_PATHS+=("$value")
+    fi
+
+    XDG_SPECIAL_DIRS_PATHS+=("$value")
```
cmatsuoka: Perhaps we could check if `value` is not empty before adding to `XDG_SPECIAL_DIRS_PATHS`? Otherwise we can see errors like `realpath: '': No such file or directory` when creating links.

extensions/desktop-common/desktop-common
```diff
@@ -58,8 +58,8 @@ function is_subpath() {
 
 if [ "$WITH_SNAP_DESKTOP_RUNTIME" = "yes" ]; then
   # add general paths not added by snapcraft due to runtime snap
-  append_dir LD_LIBRARY_PATH "$SNAP_DESKTOP_RUNTIME/lib/$SNAP_LAUNCHER_ARCH_TRIPLET"
-  append_dir LD_LIBRARY_PATH "$SNAP_DESKTOP_RUNTIME/usr/lib/$SNAP_LAUNCHER_ARCH_TRIPLET"
+  append_dir LD_LIBRARY_PATH "$SNAP_DESKTOP_RUNTIME/lib/$SNAP_DESKTOP_ARCH_TRIPLET"
```
cmatsuoka: And when this directory is not added to `LD_LIBRARY_PATH`, pixbuf loaders can't find `libgdk_pixbuf-2.0.so.0` when it's in a content provider such as `kde-frameworks-5-core18`.

extensions/desktop-common/desktop-init
diddledan: changeset fd9cdd9 removes this.

extensions/desktop-common/desktop-common
```diff
@@ -58,8 +58,8 @@ function is_subpath() {
 
 if [ "$WITH_SNAP_DESKTOP_RUNTIME" = "yes" ]; then
```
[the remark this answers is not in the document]
diddledan: changeset f3d1f5d nukes that last instance of `WITH_SNAP_DESKTOP_RUNTIME`.

extensions/desktop-common/desktop-common
diddledan: does the fix in 6c46d4b7 make this work correctly for the KDE runtime?

extensions/desktop-common/desktop-common
diddledan: I've added a not-empty check in 03478540.

extensions/desktop-common/desktop-init
diddledan: This has been dropped for now in changeset fd9cdd9.

extensions/bindtextdomain/bindtextdomain.c
diddledan: I _think_ I've solved this in 136e2108, but my C is not good so I might have royally mucked it up.

extensions/bindtextdomain/bindtextdomain.c
diddledan: I _think_ I've solved this in 136e2108, but my C is not good so I might have royally mucked it up.

extensions/bindtextdomain/bindtextdomain.c
diddledan: I'm not sure what you are referring to about leaking, but does 136e2108 fix it?

extensions/desktop-common/desktop-common
cmatsuoka: Yes, issue fixed, thanks.

extensions/gnome/desktop-gnome-specific
```diff
@@ -28,4 +28,8 @@ if [ "$SNAP_DESKTOP_COMPONENTS_NEED_UPDATE" = "true" ]; then
   fi
 fi
 
+if [ "$SNAP_DESKTOP_COMPONENTS_NEED_UPDATE" = "true" ]; then
```
cmatsuoka: `SNAP_DESKTOP_COMPONENTS_NEED_UPDATE` is tested in `desktop-init` but only updated in `desktop-gnome-specific`, so if using only `desktop-init` and `desktop-common` as parts of the command chain `$SNAP_USER_DATA/.last_revision` will never be updated. The result is that mime cache generation and other first-time initialization procedures will run at every execution, not only after an update. I think this should be updated in the common part of the desktop scripts.

diddledan: check changeset e6e6503c. This has the downside that if the user kills the launch scripts before completion it could leave them in an inconsistent state where the `.last-revision` file is marked as current but the full caching has not been completed due to the aborted launch.

cmatsuoka: Hmm, yes, you're right. So perhaps we could update it at the end of `desktop-common` after the operations are done (but still in the common part)? The other possibility would be to trap early interruptions and restore the original state, but that would add unneeded complexity IMHO.

Lin-Buo-Ren:
You might want to port https://github.com/ubuntu/snapcraft-desktop-helpers/pull/184 and https://github.com/ubuntu/snapcraft-desktop-helpers/pull/185 .

snapcraft/internal/project_loader/_extensions/gnome-3-28.py
cmatsuoka: The current implementation of the Gnome extension also applies the Desktop Common extension, which conflicts with the KDE Neon extension adding the same common parts. One possible solution would be to let the user manually compose the common desktop and the specific desktop extensions (such as `extensions: ["desktop-common", "gnome-3-28"]`). The specific case of applying both KDE and Gnome extensions at the same time seems unlikely, but this solution would allow us to make it conceptually possible instead of flagging them as mutually exclusive.

Some additional infrastructural work would still be necessary to add entries to the command chain, especially for situations where an extension needs to specifically prepend or append elements to the existing command chain.

extensions/bindtextdomain/bindtextdomain.c
cmatsuoka: It seems to be leaking indeed, if `access()` fails it continues the loop without freeing `snap_path`.

snapcraft/internal/project_loader/_extensions/gnome-3-28.py
cmatsuoka: Another way to handle this issue is to use a simple dependency system so that an extension could depend on other extensions to be applied. That way, both Gnome and KDE extensions would depend on Desktop Common, but the user won't need to declare it explicitly and the dependency would be applied only once. This would be easy to implement (in fact I already have a working PoC) and also determine application order instead of sorting by extension name like it's done today. What do you think, @sergiusens , @niemeyer, @kyrofa?

extensions/bindtextdomain/bindtextdomain.c
diddledan: aah, I see what you mean. Thanks, I'll poke it.. :-)

diddledan: changeset c18b4749 has an attempt to fix it :-)

snapcraft/internal/project_loader/_extensions/gnome-3-28.py
niemeyer: As we discussed on IRC earlier in the week, these seem like orthogonal concerns. Both extensions should be able to cope with data that is already in the file, whether it was entered by hand or by a prior extension. We shouldn't force the user to understand that there is a dependency chain across extensions.. this is unnecessary additional complexity. Internally, it probably makes sense for the extensions to reuse some code, though, but that should be completely invisible to people using either or both of them.

snapcraft/internal/project_loader/_extensions/_desktop_common.py
```diff
@@ -0,0 +1,79 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from typing import Any, Dict
+
+from ._extension import Extension
+
+
+class DesktopCommonExtension(Extension):
+    """The Common Desktop extension.
+    This extension is to be subclassed by the gnome and kde extensions.
```
niemeyer: If it's sole purpose is being subclassed, can we please make sure that people cannot use this by hand?  Otherwise this will spawn new unanticipated scenarios that need to be supported forever.

snapcraft/internal/project_loader/_extensions/gnome-3-28.py
niemeyer: Missing a response here.

snapcraft/internal/project_loader/errors.py
```diff
@@ -84,6 +84,17 @@ def __init__(self, extension_name: str, base: str) -> None:
         super().__init__(extension_name=extension_name, base=base)
 
 
+class ExtensionUnsupportedClassicError(ProjectLoaderError):
+    fmt = (
+        "Failed to load extension {extension_name!r}: "
```
niemeyer: As a suggestion, we have a rule in snapd where we use "cannot" strictly for every error message whenever that meaning is required. This makes the user experience just a tiny bit more consistent, instead of having "can't", "could not", "couldn't", "failed to", "unable to", etc. We also use lowercase on all error messages, so that they can be composed, as in "cannot do foo: the file doesn't exist", rather than Something: Like: This.

snapcraft/internal/project_loader/_extensions/_desktop_common.py
diddledan: the underscore in the file name should prevent snapcraft from calling it directly. That is what I understood from reading the extension loading code, at least. Perhaps @sergiusens or @cmatsuoka can confirm?

snapcraft/internal/project_loader/errors.py
diddledan: This follows the pattern of all the other errors in the file. `Failed` is used several times in other messages, and all the other messages in the file are `Sentence case`.

snapcraft/internal/project_loader/_extensions/gnome-3-28.py
diddledan: This is from the glib example that I used as inspiration. I do not know the state of adaptors and their existence or not. @sergiusens or @cmatsuoka do you know whether this is still needed or should I drop the `adapter` setting?

snapcraft/internal/project_loader/_extensions/gnome-3-28.py
diddledan: @cmatsuoka is your PoC somewhere that I can merge it into this PR? I think it is a good idea and covers @niemeyer's concerns appropriately.

cmatsuoka: I discussed this with @niemeyer and we'll keep the Gnome extension as it is, without dealing with dependencies. Each extension has access to its context and will check if e.g. a part already exists before adding them again so they can coexist peacefully. I can add the checks along with the KDE Neon extension.

snapcraft/internal/project_loader/_extensions/gnome-3-28.py
sergiusens: command-chain arrived a couple of months after we did the 3.0 release, so we sticked to "legacy" as an adapter by default.

We found a couple of scenarios that broke classic confinement (mostly when using shebang/interpreters) so the transition needs a bit more work on our side to complete it.

snapcraft/internal/project_loader/_extensions/_desktop_common.py
cmatsuoka: It's true that the base desktop-common extension can't be loaded, but it fails in a rather ugly way because the class name doesn't match the expected extension module name. I'll propose a minor patch to handle it.

snapcraft/internal/project_loader/_extensions/gnome-3-28.py
niemeyer: We're talking about a extension being created just now. We shouldn't release this with legacy code that we know is obsolete and needs to be removed as people will start trusting on that. The behavior on any extensions we create need to be supported forever inside the same base.

sergiusens: This is why we are forcing the extension to use the new code (it is not the default yet), `adapter: full` makes use of `command-chain` and the vanilla `command` passed on from `snapcraft.yaml` to `snap.yaml`.
Once we make this the default behavior (`adapter: full`) the `adapter` keyword can just go away.

niemeyer: I must be missing something. "adapter: full" certainly doesn't sound like "use **no** adapter".

cmatsuoka: Changed to `adapter: none` and added the variable setting to the existing command chain, as discussed with @sergiusens.

sergiusens:
[gnome extension](https://trello.com/c/i5xI1YOT/26-gnome-extension)

cjp256:
As it stands today after merging this from master, adapter=full is required for users of this extension to build (use of command-chain appears to require it explicitly).

How are the common stage-packages expected to be handled? Building gnome-calculator, I get a number of packages required that I would expect to be common for gnome applications.

Built with:
https://github.com/cjp256/gnome-calculator/commit/6ac843d5ac4eb216402a144d3fec1505792f0335
https://github.com/cjp256/snapcraft/commit/bda55d9bc813513bf77ebc9f855b29ec813f1c64

```
The 'gnome-calculator' part needs the following libraries that are not included in the snap or base: 
usr/lib/x86_64-linux-gnu/libX11.so.6
usr/lib/x86_64-linux-gnu/libXau.so.6
usr/lib/x86_64-linux-gnu/libXcomposite.so.1
usr/lib/x86_64-linux-gnu/libXcursor.so.1
usr/lib/x86_64-linux-gnu/libXdamage.so.1
usr/lib/x86_64-linux-gnu/libXdmcp.so.6
usr/lib/x86_64-linux-gnu/libXext.so.6
usr/lib/x86_64-linux-gnu/libXfixes.so.3
usr/lib/x86_64-linux-gnu/libXi.so.6
usr/lib/x86_64-linux-gnu/libXinerama.so.1
usr/lib/x86_64-linux-gnu/libXrandr.so.2
usr/lib/x86_64-linux-gnu/libXrender.so.1
usr/lib/x86_64-linux-gnu/libatk-1.0.so.0
usr/lib/x86_64-linux-gnu/libatk-bridge-2.0.so.0
usr/lib/x86_64-linux-gnu/libatspi.so.0
usr/lib/x86_64-linux-gnu/libcairo-gobject.so.2
usr/lib/x86_64-linux-gnu/libcairo.so.2
usr/lib/x86_64-linux-gnu/libdatrie.so.1
usr/lib/x86_64-linux-gnu/libepoxy.so.0
usr/lib/x86_64-linux-gnu/libfontconfig.so.1
usr/lib/x86_64-linux-gnu/libfreetype.so.6
usr/lib/x86_64-linux-gnu/libgdk-3.so.0
usr/lib/x86_64-linux-gnu/libgdk_pixbuf-2.0.so.0
usr/lib/x86_64-linux-gnu/libgee-0.8.so.2
usr/lib/x86_64-linux-gnu/libgraphite2.so.3
usr/lib/x86_64-linux-gnu/libgtk-3.so.0
usr/lib/x86_64-linux-gnu/libharfbuzz.so.0
usr/lib/x86_64-linux-gnu/libicudata.so.60
usr/lib/x86_64-linux-gnu/libicuuc.so.60
usr/lib/x86_64-linux-gnu/libmpc.so.3
usr/lib/x86_64-linux-gnu/libmpfr.so.6
usr/lib/x86_64-linux-gnu/libpango-1.0.so.0
usr/lib/x86_64-linux-gnu/libpangocairo-1.0.so.0
usr/lib/x86_64-linux-gnu/libpangoft2-1.0.so.0
usr/lib/x86_64-linux-gnu/libpixman-1.so.0
usr/lib/x86_64-linux-gnu/libpng16.so.16
usr/lib/x86_64-linux-gnu/libsoup-2.4.so.1
usr/lib/x86_64-linux-gnu/libthai.so.0
usr/lib/x86_64-linux-gnu/libwayland-client.so.0
usr/lib/x86_64-linux-gnu/libwayland-cursor.so.0
usr/lib/x86_64-linux-gnu/libwayland-egl.so.1
usr/lib/x86_64-linux-gnu/libxcb-render.so.0
usr/lib/x86_64-linux-gnu/libxcb-shm.so.0
usr/lib/x86_64-linux-gnu/libxcb.so.1
usr/lib/x86_64-linux-gnu/libxkbcommon.so.0
usr/lib/x86_64-linux-gnu/libxml2.so.2
These dependencies can be satisfied via more stage-packages, more parts, or content sharing.
```

cjp256:
Should gsettings be included in the gnome plugs?  I would imagine it would be commonly required by gnome apps?

diddledan:
> As it stands today after merging this from master, adapter=full is required for users of this extension to build (use of command-chain appears to require it explicitly).

See the discussion about this with @neimeyer above at https://github.com/snapcore/snapcraft/pull/2398#discussion_r268393331

diddledan:
> How are the common stage-packages expected to be handled? Building gnome-calculator, I get a number of packages required that I would expect to be common for gnome applications.

I believe all of those are handled by the gnome-3-28-1604 snap

kenvandine:
> > How are the common stage-packages expected to be handled? Building gnome-calculator, I get a number of packages required that I would expect to be common for gnome applications.
> 
> I believe all of those are handled by the gnome-3-28-1604 snap

Those are provided by the content interface in gnome-3-28-1804

diddledan:
> Should gsettings be included in the gnome plugs? I would imagine it would be commonly required by gnome apps?

I unilaterally elected to let the snap author determine whether they need it or not rather than forcing them to use it. We want to default to secure/closed, so only the absolutely necessary interfaces should be plugged IMO.

diddledan:
> Those are provided by the content interface in gnome-3-28-1804

Thanks for the correction on my typo :-)

cjp256:
> > As it stands today after merging this from master, adapter=full is required for users of this extension to build (use of command-chain appears to require it explicitly).
> 
> See the discussion about this with @neimeyer above at [#2398 (comment)](https://github.com/snapcore/snapcraft/pull/2398#discussion_r268393331)

Thanks - I'm going to look at setting adapter=full (instead of assuming legacy) if command chain is used.

cjp256:
> > Should gsettings be included in the gnome plugs? I would imagine it would be commonly required by gnome apps?
> 
> I unilaterally elected to let the snap author determine whether they need it or not rather than forcing them to use it. We want to default to secure/closed, so only the absolutely necessary interfaces should be plugged IMO.

While that's a sensible security approach, I'm curious if there really are any gnome apps that don't use gsettings...?

kenvandine:
> > > Should gsettings be included in the gnome plugs? I would imagine it would be commonly required by gnome apps?
> > 
> > 
> > I unilaterally elected to let the snap author determine whether they need it or not rather than forcing them to use it. We want to default to secure/closed, so only the absolutely necessary interfaces should be plugged IMO.
> 
> While that's a sensible security approach, I'm curious if there really are any gnome apps that don't use gsettings...?

This extension isn't only for GNOME snaps.  It's really for any snap targetted for the desktop that uses things like GTK.  For example all the electron snaps should use this extension, those generally won't need gsettings.

diddledan:
can I get a re-review please? I've added async execution support based on ubuntu/snapcraft-desktop-helpers#190 so it needs a thorough check with a real application.

sergiusens:
I have created https://github.com/snapcore/snapcraft/pull/2655 which builds on top of this, but is rebased with conflicts and removes the latest commit which do not pass shellcheck

sergiusens:
#2655 did a bit more than just fix the tests, making async work required a bit of rewriting.

PR #2408, opened 2018-11-14
multipass: avoid stdin where possible

sergiusens:
multipass commands that do not require stdin should have it point to
subprocess.DEVNULL to avoid multipass' metrics UI

LP: #1803445

-----

### DIFF
--- a/snapcraft/internal/build_providers/_multipass/_multipass_command.py
+++ b/snapcraft/internal/build_providers/_multipass/_multipass_command.py
@@ -26,9 +26,9 @@ from snapcraft.internal.build_providers import errors
 logger = logging.getLogger(__name__)
 
 
-def _run(command: Sequence[str]) -> None:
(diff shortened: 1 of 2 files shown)


snapcraft/internal/build_providers/_multipass/_multipass_command.py
```diff
@@ -26,9 +26,9 @@
 logger = logging.getLogger(__name__)
 
 
-def _run(command: Sequence[str]) -> None:
+def _run(command: Sequence[str], stdin=None) -> None:
```
chipaca: I'd make this default to `subprocess.DEVNULL`, if the intent really is to not use it where possible. Then when you want the default behaviour you do `stdin=None` (as you are already below). The diff should shrink quite a bit, which is a bonus.

sergiusens: sounds like a good idea

PR #2413, opened 2018-11-21
kernel plugin: introduce a _get_kernel_version() helper

piso77:
Extract the kernel version from the Linux source code tree and make decision using this extracted data: at the moment, i'm skipping the 'make firmware_install' step if 
 kernel version >= 4.14 (since that target was retired in 4.14 and compilation would fail there) but in the future i want to exploit this feature to adjust the kernel config based off the specific kernel version we are building.

-----


sergiusens:
Hi, thanks for the contribution, mind creating a bug for this, as you probably want to target it to our legacy branch as well so it works in a sans base snapcraft.yaml

snapcraft/plugins/kernel.py
```diff
@@ -387,6 +387,25 @@ def _parse_kernel_release(self):
...
             )
 
+    def _get_kernel_version(self):
+        kernelversion = (
+            subprocess.check_output("make kernelversion", shell=True, cwd=self.builddir)
+            .decode("utf-8")
+            .strip()
+        )
+        if not kernelversion:
+            raise ValueError("make kernelversion didn't produce any version")
```
sergiusens: you need to raise an exception Inheriting from `SnapcraftError`. Look at the `dump` plugin for an example. Missing a new sentence on how to fix the issue.

snapcraft/plugins/kernel.py
```diff
@@ -387,6 +387,25 @@ def _parse_kernel_release(self):
                 )
             )
 
+    def _get_kernel_version(self):
+        kernelversion = (
+            subprocess.check_output("make kernelversion", shell=True, cwd=self.builddir)
```
sergiusens: what if this fails? How about catching the exception and raising one that inherits from `SnapcraftError`. Look at the `dump` plugin for inspiration

snapcraft/plugins/kernel.py
```diff
@@ -387,6 +387,25 @@ def _parse_kernel_release(self):
...
+        try:
+            # Linux kernel version is composed of up to 4 elements: version,
+            # patchlevel, sublevel and extraversion - the first three elements
+            # are numbers, while the last one is a an arbitrary ascii
+            # string e.g. 4.18.3-rc6 => '4', '18', '3', 'rc6'
+            self.kernel_version = re.split("\W+", kernelversion)
+            for i in range(0, 3):
+                self.kernel_version[i] = int(self.kernel_version[i])
+        except Exception as e:
+            raise ValueError("Malformed kernel_version: {!r}".format(kernelversion))
```
sergiusens: ditto on exception handling. Missing a new sentence on how to fix the issue or excepted format string.

tests/unit/plugins/test_kernel.py
```diff
@@ -1270,6 +1353,7 @@ class Options:
             ),
         )
         plugin = kernel.KernelPlugin("test-part", self.options, project)
+        plugin._set_kernel_targets()
```
sergiusens: The fact that you are doing this here makes me think that  if you run `pull` and `build` separately, the wrong thing will happen. How about making this implementation idempotent and calling where necessary since it is not `__init__`?

cjp256:
@piso77 @sergiusens Is this PR still desirable or was there another approach taken?  The source repository seems to have been deleted...  It'd be nice to close the PR if it's no longer useful, or revisit the topic if it still needs to be addressed.

sergiusens:
I am closing this due to inactivity

PR #2437, opened 2019-01-11
repo,baseplugin: support trusting repo keys

kyrofa:

-----

Today, snapcraft allows insecure repositories (i.e. repos without a key) because that is the default behavior in Xenial (core16). This behavior changed in Zesty, and stays changed in Bionic (core18). This PR resolves [LP: #1811304](https://bugs.launchpad.net/snapcraft/+bug/1811304) by adding the ability to trust keys to the plugin API, taking advantage of it in the Catkin plugin, and flipping the switch to only allow snapcraft to use secure repositories.

For history, see #1614 as well as [this forum topic](https://forum.snapcraft.io/t/insecure-apt-repos/2464).


kyrofa:
Hmm, I'm not getting that test failure locally, I'll need to look into that.

snapcraft/internal/repo/_deb.py
```diff
@@ -58,15 +58,22 @@
...
-    def __init__(self, deb_arch, *, sources_list=None, use_geoip=False):
+    def __init__(self, deb_arch, *, sources_list=None, keyrings=None, use_geoip=False):
         self._deb_arch = deb_arch
         self._sources_list = sources_list
         self._use_geoip = use_geoip
 
+        self._keyrings = list()
+        if keyrings:
+            self._keyrings = keyrings
+
```
sergiusens: same lines of code counter proposal:
```python
if not keyrings:
    keyrings = list()
self._keyrings = keyrings
```

or
```python
self._keyrings = keyrings if keyrings else list()
```

snapcraft/plugins/_ros/wstool.py
```diff
@@ -59,6 +59,7 @@ def __init__(
         ros_package_path: str,
         wstool_path: str,
         ubuntu_sources: str,
+        ubuntu_keyrings: List[str],
```
sergiusens: `List` should be reserved for return types, it is preferable to use `Sequence` here (https://docs.python.org/3.5/library/typing.html#typing.List)

snapcraft/internal/repo/_deb.py
kyrofa: Sure, fixed.

PR #2450, opened 2019-01-24
pluginhandler: handle removal of inconsistent files

cmatsuoka:
When a part is cleaned, the system does its housekeeping by removing
all the files or directories it prepared for stage or prime. If the
user manually removes a file (say, in another part's override-prime),
we should ignore the non-existent file and issue a warning instead
of halting and catching fire.

See https://bugs.launchpad.net/snapcraft/+bug/1813033 for more details.

-----

### DIFF
--- a/snapcraft/internal/pluginhandler/__init__.py
+++ b/snapcraft/internal/pluginhandler/__init__.py
@@ -1134,8 +1134,8 @@ def _clean_migrated_files(snap_files, snap_dirs, directory):
             os.remove(os.path.join(directory, snap_file))
         except FileNotFoundError:
             logger.warning(
-                "File {name!r} not found. Did you remove it in a different "
-                "part?".format(name=snap_file)
+                "Attempted to remove file {name!r}, but it didn't exist. "
+                "Skipping...".format(name=snap_file)
             )
 
     # snap_dirs may not be ordered so that subdirectories come before
@@ -1150,8 +1150,8 @@ def _clean_migrated_files(snap_files, snap_dirs, directory):
                 os.rmdir(migrated_directory)
         except FileNotFoundError:
             logger.warning(
-                "Directory {name!r} not found. Did you remove it in a "
-                "different part?".format(name=snap_dir)
+                "Attempted to remove directory {name!r}, but it didn't exist. "
+                "Skipping...".format(name=snap_dir)
             )
 
 
--- a/snapcraft/internal/pluginhandler/__init__.py
+++ b/snapcraft/internal/pluginhandler/__init__.py
@@ -1130,7 +1130,13 @@ def _organize_filesets(part_name, fileset, base_dir, overwrite):
 
 def _clean_migrated_files(snap_files, snap_dirs, directory):
     for snap_file in snap_files:
-        os.remove(os.path.join(directory, snap_file))
+        try:
+            os.remove(os.path.join(directory, snap_file))
+        except FileNotFoundError:
+            logger.warning(
+                "File {name!r} not found. Did you remove it in a different "
+                "part?".format(name=snap_file)
+            )
 
     # snap_dirs may not be ordered so that subdirectories come before
     # parents, and we want to be able to remove directories if possible, so
@@ -1139,8 +1145,14 @@ def _clean_migrated_files(snap_files, snap_dirs, directory):
 
     for snap_dir in snap_dirs:
         migrated_directory = os.path.join(directory, snap_dir)
-        if not os.listdir(migrated_directory):
-            os.rmdir(migrated_directory)
+        try:
+            if not os.listdir(migrated_directory):
+                os.rmdir(migrated_directory)
+        except FileNotFoundError:
+            logger.warning(
+                "Directory {name!r} not found. Did you remove it in a "
+                "different part?".format(name=snap_dir)
+            )
 
 
 def _get_file_list(stage_set):
--- a/tests/unit/pluginhandler/test_pluginhandler.py
+++ b/tests/unit/pluginhandler/test_pluginhandler.py
@@ -1810,15 +1810,17 @@ class StateTestCase(StateBaseTestCase):
     def test_clean_prime_state_inconsistent_files(self):
         self.assertRaises(errors.NoLatestStepError, self.handler.latest_step)
         self.assertThat(self.handler.next_step(), Equals(steps.PULL))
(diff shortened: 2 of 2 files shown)


tests/unit/pluginhandler/test_pluginhandler.py
```diff
@@ -1807,6 +1807,25 @@ def test_clean_prime_state(self):
         self.assertThat(self.handler.next_step(), Equals(steps.PRIME))
         self.assertFalse(os.path.exists(bindir))
 
+    def test_clean_prime_state_inconsistent_files(self):
```
kyrofa: This test doesn't cover the [directory case](https://github.com/snapcore/snapcraft/pull/2450/files#diff-660a35e421f136410b0d3e5168c7a62cR1150). Mind either adding a new test, or also covering that here?

cmatsuoka: The idea was to test removal of non-existent directories by passing directory "foo" (in line 1820 below), Effectively, when the test runs on an unpatched pluginhandler, `os.listdir()` raises `FileNotFoundError`.  Should I refactor to cover more cases/scenarios?

tests/unit/pluginhandler/test_pluginhandler.py
```diff
@@ -1807,6 +1807,25 @@ def test_clean_prime_state(self):
...
+        self.assertRaises(errors.NoLatestStepError, self.handler.latest_step)
+        self.assertThat(self.handler.next_step(), Equals(steps.PULL))
+        bindir = os.path.join(self.prime_dir, "bin")
+        os.makedirs(bindir)
+        open(os.path.join(bindir, "1"), "w").close()
+
+        self.handler.mark_done(steps.STAGE)
```
sergiusens: in practice, prime is a subset of staged files so mark_done should probably reflect that in this test.
This error could also be triggered by self.handler.clean_stage.

tests/unit/pluginhandler/test_pluginhandler.py
```diff
@@ -1807,6 +1807,25 @@ def test_clean_prime_state(self):
...
+
+        self.handler.mark_done(
```
sergiusens: mind adding a comment about adding a file bin/2 and dir foo as primed but having them non existent?
Alternatively, invert the order and add a comment to the file creation part
```diff
diff --git a/tests/unit/pluginhandler/test_pluginhandler.py b/tests/unit/pluginhandler/test_pluginhandler.py
index 930ea82..4ceda2c 100644
--- a/tests/unit/pluginhandler/test_pluginhandler.py
+++ b/tests/unit/pluginhandler/test_pluginhandler.py
@@ -1810,9 +1810,6 @@ class StateTestCase(StateBaseTestCase):
     def test_clean_prime_state_inconsistent_files(self):
         self.assertRaises(errors.NoLatestStepError, self.handler.latest_step)
         self.assertThat(self.handler.next_step(), Equals(steps.PULL))
-        bindir = os.path.join(self.prime_dir, "bin")
-        os.makedirs(bindir)
-        open(os.path.join(bindir, "1"), "w").close()
 
         self.handler.mark_done(steps.STAGE)
 
@@ -1820,6 +1817,11 @@ class StateTestCase(StateBaseTestCase):
             steps.PRIME, states.PrimeState({"bin/1", "bin/2"}, {"bin", "foo"})
         )
 
+        # Only create a subset of the files and directories that were marked as primed
+        bindir = os.path.join(self.prime_dir, "bin")
+        os.makedirs(bindir)
+        open(os.path.join(bindir, "1"), "w").close()
+
         self.handler.clean_prime({})
 
         self.assertThat(self.handler.latest_step(), Equals(steps.STAGE))
```

tests/unit/pluginhandler/test_pluginhandler.py
kyrofa: Ha! _Totally_ missed `foo`. Yeah, just add a comment and I'm happy.

sergiusens: Yeah, that is the reason I asked for the comment, hard to spot... maybe changing the name would also be good `inexistent-file` and `inexistent-dir` (or use nonexistent, both are valid according to Merriam Webster at least)

cmatsuoka: Refactored test to change order and add comment as per Segio's suggestion. If good to go and nothing more to add, I'll squash the commits for merge.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -1130,7 +1130,13 @@ def _organize_filesets(part_name, fileset, base_dir, overwrite):
 
 def _clean_migrated_files(snap_files, snap_dirs, directory):
     for snap_file in snap_files:
-        os.remove(os.path.join(directory, snap_file))
+        try:
+            os.remove(os.path.join(directory, snap_file))
+        except FileNotFoundError:
+            logger.warning(
+                "File {name!r} not found. Did you remove it in a different "
+                "part?".format(name=snap_file)
```
kyrofa: Asking this question is rubbing me the wrong way. I'd prefer the warning just stated the problem instead of guessing why the problem exists. There are other ways this happens, too (e.g. manually removing files or moving them around). I suggest rephrasing these warnings to be statements, for example: "Attempted to remove file {!r}, but it didn't exist. Skipping..."

sergiusens: I had this in my draft notes, then saw you reviewed and did not bring it up so I thought I was being picky :smile:. I would like to reuse the error pattern for warnings as much as possible https://github.com/snapcore/snapcraft/blob/master/CODE_STYLE.md#errors

cmatsuoka: Agreed. That's something I considered for the same reasons, but (reluctantly) ended up with the question because it was stylistically similar to another message already in the code. Given the uncertainty/guessing factor there, the fact statement is indeed a better option. New patch on the way.

kyrofa: > That's something I considered for the same reasons, but (reluctantly) ended up with the question because it was stylistically similar to another message already in the code.

Trust your gut. Parts of this codebase have been around for a _long_ time, and pluginhandler is one of the worst offenders. Established patterns aren't always the best, if you feel something isn't quite the way it should be, always feel free to say so!

PR #2453, opened 2019-01-28
ant, maven and gradle plugins: use correct defaults for jre

sergiusens:
Ensure that the expected java version is properly set when the default
openjdk option is used so that the globbing mechanism can format the
path to the java binary correctly.

Also ensure that the sytem works end to end for core16 and core18,
removing the typo of core19 which was supposed to be core16.

LP: #1813637
LP: #1813636
Fixes SNAPCRAFT-D1

-----

### DIFF
--- a/snapcraft/plugins/maven.py
+++ b/snapcraft/plugins/maven.py
@@ -256,7 +256,6 @@ class MavenPlugin(snapcraft.BasePlugin):
                 "lib",
                 "jvm",
                 "java-{}-openjdk-*".format(self._java_version),
-                "jre",
                 "bin",
                 "java",
             )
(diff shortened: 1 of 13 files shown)


snapcraft/plugins/maven.py
```diff
@@ -255,7 +255,7 @@ def _create_symlinks(self):
                 "usr",
                 "lib",
                 "jvm",
-                "java-{}-openjdk-*".format(self.options.maven_openjdk_version),
+                "java-{}-openjdk-*".format(self._java_version),
                 "jre",
```
cmatsuoka: Also remove `jre` here, see comment in gradle.py.

kubiko: correct there is no 'jre' directory when using OpenJDK 11, code should be more forgiving

sergiusens: just added tests to prove this error, going to fix tomorrow

cmatsuoka:
Paths are correct now to work with both 8 and 11, and the rest seems to be fine.

sergiusens:
hmm, thanks @cmatsuoka ... timeouts ftw!

PR #2455, opened 2019-01-30
cli: retrieve error data from provider

cmatsuoka:
Upon facing an unhandled exception, snapcraft generates traceback data and stores it in the filesystem. If running in a build provider, the filesystem location is not accessible from host, and to give user access to error details we need to retrieve this data from the managed provider.
    
This fix addresses this issue by letting inner instance to save error data in a known location, which is retrieved by outer instance before proceeding.
    
LP: #1793082

**Note:** this patch refactors CLI error handling to deal with formatted trace data from provider. Raw traceback data can't be easily pickled up for transferring, hence the pre-formatting. It's also hard to test since this version must also be injected in the build provider, but I strongly recommend it to be tested before merging. I ran my tests by manually setting up this version inside the multipass instance.

Expected behavior is as follows:
- Crashes happening inside the build provider are silent, but generate traceback data that's handled by the managing (outer) instance.
- When running non-interactively (no tty), the traceback data should also be printed.
- No temporary files used for traceback data retrieval should be present after building.
- Non-crashing builds should (obviously) not produce error data.
- Legacy modes are handled by legacy snapcraft and should not be affected by this change.
- When running without build providers (destructive mode), error data should be presented as usual.

-----

-----

### DIFF
--- a/snapcraft/cli/_errors.py
+++ b/snapcraft/cli/_errors.py
@@ -17,6 +17,8 @@
 import distutils.util
 import os
 import sys
+import shutil
+import logging
 import tempfile
 import traceback
 from textwrap import dedent
@@ -28,6 +30,7 @@ from . import echo
 import snapcraft
 from snapcraft.config import CLIConfig as _CLIConfig
 from snapcraft.internal import errors
+from snapcraft.internal.build_providers.errors import ProviderExecError
 
 # raven is not available on 16.04
 try:
@@ -66,6 +69,11 @@ _YES_VALUES = ["yes", "y"]
 _NO_VALUES = ["no", "n"]
 _ALWAYS_VALUES = ["always", "a"]
 
+TRACEBACK_FILEPATH = ".snapcraft_provider_traceback"
+
+
+logger = logging.getLogger(__name__)
+
 
 def exception_handler(  # noqa: C901
     exception_type, exception, exception_traceback, *, debug=False
@@ -86,16 +94,18 @@ def exception_handler(  # noqa: C901
         - a snapcraft handled error occurs, debug=False so only the
           exception message is shown
     """
-    # TODO pickle the traceback if on a manged host so the actual host can deal with it.
     exc_info = (exception_type, exception, exception_traceback)
     exit_code = 1
-    is_snapcraft_error = issubclass(exception_type, errors.SnapcraftError)
+    is_snapcraft_error = (
+        issubclass(exception_type, errors.SnapcraftError)
+        and exception_type is not ProviderExecError
+    )
     is_snapcraft_reportable_error = issubclass(
         exception_type, errors.SnapcraftReportableError
     )
     is_raven_setup = RavenClient is not None
     is_snapcraft_managed_host = (
-        distutils.util.strtobool(os.getenv("SNAPCRAFT_MANAGED_HOST", "n")) == 1
+        os.getenv("SNAPCRAFT_BUILD_ENVIRONMENT") == "managed-host"
     )
     is_connected_to_tty = sys.stdout.isatty()
     ask_to_report = False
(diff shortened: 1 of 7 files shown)


sergiusens:
```
Running flake8
./snapcraft/cli/lifecycle.py:165:1: E302 expected 2 blank lines, found 1
./snapcraft/cli/lifecycle.py:172:1: E302 expected 2 blank lines, found 1
./snapcraft/cli/lifecycle.py:178:1: E302 expected 2 blank lines, found 1
```

snapcraft/cli/_errors.py
```diff
@@ -86,16 +94,18 @@ def exception_handler(  # noqa: C901
         - a snapcraft handled error occurs, debug=False so only the
           exception message is shown
     """
-    # TODO pickle the traceback if on a manged host so the actual host can deal with it.
     exc_info = (exception_type, exception, exception_traceback)
     exit_code = 1
-    is_snapcraft_error = issubclass(exception_type, errors.SnapcraftError)
+    is_snapcraft_error = (
+        issubclass(exception_type, errors.SnapcraftError)
+        and exception_type is not ProviderExecError
```
sergiusens: Ah, this is where the magic happens. We do however lose the traceback information to what led to this exception from sentry's point of view though, right?

snapcraft/cli/_errors.py
```diff
@@ -66,6 +69,11 @@
 _NO_VALUES = ["no", "n"]
 _ALWAYS_VALUES = ["always", "a"]
 
+TRACEBACK_FILEPATH = ".snapcraft_provider_traceback"
```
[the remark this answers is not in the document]
cmatsuoka: The reason I didn't place it in an absolute location is to allow more than one build to happen on the machine at the same time, but we can protect that adding a PID to the file name (which also would prevent us from the unlikely situation of reusing an old trace file). so /tmp/snapcraft_provider_traceback.PID?

snapcraft/cli/_errors.py
```diff
@@ -115,29 +125,46 @@ def exception_handler(  # noqa: C901
...
     if ask_to_report:
-        if not is_raven_setup or is_snapcraft_managed_host:
+        if not is_raven_setup:
             click.echo(_MSG_TRACEBACK_PRINT)
             traceback.print_exception(*exc_info, file=sys.stdout)
             click.echo(_MSG_MANUALLY_REPORT)
         else:
-            if is_connected_to_tty:
-                click.echo(_MSG_TRACEBACK_FILE)
+            if is_snapcraft_managed_host:
```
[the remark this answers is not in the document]
cmatsuoka: Mmm, yes, I think this was the result of moving logical blocks around. Will change to elif.

cmatsuoka:
A number of changes were made to this PR:
- Sending crash reports to sentry is now done by the inner snapcraft instance, if a build provider is used. Externalizing the traceback using [tblib](https://pypi.org/project/tblib/) strips a few details off so only the instance where the crash originates is able to report it with full.
- Inner instance must know whether outer instance is attached to a terminal or not, to print the crash dump to screen. This information is passed through an environment variable on provider command execution.
- Provider method `retrieve_file()` was renamed to `pull_file()`, to keep it consistent with `_push_file()`. Also a parameter for destination name was added to this call.
- Files used for transfer now created on `/tmp`. Outer file has a PID suffix to prevent name clashes with other concurrent instances.
- The traceback file name is announced by the user-facing instance (outer instance in a provider-based build).

The following known problems were not addressed:
- Setting crash report to "Always" is only valid for the instance where the crash happens, meaning that in a provider-based build only the inner instance will persist that value. If the instance is recreated, snapcraft will ask the user again.

The changes were tested in provider-based and destructive mode builds, with and without terminal attachment.

snapcraft/cli/_errors.py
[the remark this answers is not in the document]
cmatsuoka: Changed to use an absolute location in both inner and outer environments.

snapcraft/internal/build_providers/_multipass/_multipass.py
```diff
@@ -251,6 +251,22 @@ def retrieve_snap(self) -> str:
         self._multipass_cmd.copy_files(source=source, destination=self.snap_filename)
         return self.snap_filename
 
+    def retrieve_file(self, name: str, delete: bool = False):
```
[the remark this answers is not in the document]
cmatsuoka: Changed to `pull_file`.

snapcraft/cli/_errors.py
```diff
@@ -86,18 +96,26 @@ def exception_handler(  # noqa: C901
...
     is_raven_setup = RavenClient is not None
+    is_snapcraft_host = exc_info is not None and not os.path.isfile(TRACEBACK_HOST)
     is_snapcraft_managed_host = (
-        distutils.util.strtobool(os.getenv("SNAPCRAFT_MANAGED_HOST", "n")) == 1
+        os.getenv("SNAPCRAFT_BUILD_ENVIRONMENT") == "managed-host"
+    )
+    is_connected_to_tty = (
+        # used by inner instance, variable set by outer instance
+        (distutils.util.strtobool(os.getenv("SNAPCRAFT_HAS_TTY", "n")) == 1)
+        if is_snapcraft_managed_host
```
sergiusens: this looks really weird when read with this indentation :confused: 
Oh black...

nothing to do about this comment, this is we airing my thoughts.

snapcraft/cli/_errors.py
```diff
@@ -86,18 +96,26 @@ def exception_handler(  # noqa: C901
...
-    is_snapcraft_error = issubclass(exception_type, errors.SnapcraftError)
+    is_snapcraft_error = (
+        issubclass(exception_type, errors.SnapcraftError)
+        and exception_type is not ProviderExecError
+    )
     is_snapcraft_reportable_error = issubclass(
         exception_type, errors.SnapcraftReportableError
     )
     is_raven_setup = RavenClient is not None
+    is_snapcraft_host = exc_info is not None and not os.path.isfile(TRACEBACK_HOST)
```
sergiusens: can you add a comment above on why this is `is_snapcraft_host`? It doesn't parse quickly as the checks seem unrelated.

snapcraft/cli/_errors.py
```diff
@@ -86,18 +96,26 @@ def exception_handler(  # noqa: C901
         - a snapcraft handled error occurs, debug=False so only the
           exception message is shown
     """
-    # TODO pickle the traceback if on a manged host so the actual host can deal with it.
     exc_info = (exception_type, exception, exception_traceback)
     exit_code = 1
-    is_snapcraft_error = issubclass(exception_type, errors.SnapcraftError)
+    is_snapcraft_error = (
```
sergiusens: mind adding a comment above explaining why we filter out ProviderExecError for the casual future code contributor (including ourselves :smile:)?

snapcraft/cli/_errors.py
cmatsuoka: I agree, black indents are strange sometimes.

snapcraft/cli/_errors.py
cmatsuoka: Testing `exc_info` for `None` is unnecessary since we always get here when an exception is raised. Adding a comment to explain the rest.

snapcraft/cli/_errors.py
cmatsuoka: Added. It's better to do that while things still make sense.

PR #2473, opened 2019-02-15
plugins: improve python and go schema validation

cmatsuoka:
When using build providers, validate our yaml file schema prior to
executing ourselves inside the provider instance. This allows us
to quickly interrupt execution if validation fails. Also added
extra validation entries to the Go and Python plugins, to require
either sources or packages and prevent crashes in the test case
listed in the original bug report.

LP: #1806055

-----

### DIFF
--- a/snapcraft/plugins/go.py
+++ b/snapcraft/plugins/go.py
@@ -83,6 +83,7 @@ class GoPlugin(snapcraft.BasePlugin):
             "items": {"type": "string"},
             "default": [],
         }
+        schema["anyOf"] = [{"required": ["source"]}, {"required": ["go-packages"]}]
 
         return schema
 
--- a/snapcraft/plugins/python.py
+++ b/snapcraft/plugins/python.py
@@ -122,6 +122,7 @@ class PythonPlugin(snapcraft.BasePlugin):
             "default": "python3",
             "enum": ["python2", "python3"],
         }
+        schema["anyOf"] = [{"required": ["source"]}, {"required": ["python-packages"]}]
 
         return schema
 
--- a/tests/unit/plugins/test_go.py
+++ b/tests/unit/plugins/test_go.py
@@ -15,6 +15,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import os
+import jsonschema
 from textwrap import dedent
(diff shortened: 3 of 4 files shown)


snapcraft/cli/lifecycle.py
```diff
@@ -117,6 +117,10 @@ def _execute(  # noqa: C901
         if pack_project:
             _pack(project.prime_dir, output=output)
     else:
+        # load project and validate schema before invoking a provider
+        if project.info.base is not None:
+            project_config = project_loader.load_config(project)
```
sergiusens: this will load a lot of code on the host, the host being Ubuntu or any other linux variant, Windows and OSX, we should think about alternatives to this.

cmatsuoka: Changed to validate plugin schemas by directly calling the plugin loader.

sergiusens:
Can we split this in two PRs? one with the new logic and another with the loading one?
I think we want to land the schema now and give another twist and turn of knobs on loading.
We are also missing a plugin schema test for mixing these schema values.

cmatsuoka:
Rolled back the logic change for a separate PR, per maintainer's request. Added a set of tests to validate go and python schemas regarding only one of "source" or "{python,go}-package" being allowed to be in a given part.

cmatsuoka:
Both source and {go,python}-packages should be allowed at the same time in the plugins. Fixed schema and tests accordingly.

snapcraft/cli/lifecycle.py
```diff
@@ -117,6 +117,10 @@ def _execute(  # noqa: C901
         if pack_project:
             _pack(project.prime_dir, output=output)
     else:
+        # load project and validate schema before invoking a provider
```
sergiusens: in the comment you mention having rolled back this logic, but I still see it here?

sergiusens:
The second part of your bug resolution is https://bugs.launchpad.net/snapcraft/+bug/1807875 and could be used as the reference to what is being fixed.

snapcraft/cli/lifecycle.py
cmatsuoka: Commits split at the wrong point. Re-splitting.

PR #2479, opened 2019-02-21
store: support registering to a specific store

sergiusens:
Add a --store-id option to the register command.

-----

### DIFF
--- a/snapcraft/cli/store.py
+++ b/snapcraft/cli/store.py
@@ -101,7 +101,7 @@ def _human_readable_acls(store: storeapi.StoreClient) -> str:
 @click.argument("snap-name", metavar="<snap-name>")
 @click.option("--private", is_flag=True, help="Register the snap as a private one")
 @click.option("--store", metavar="<store>", help="Store to register with")
-def register(snap_name, private, store_id):
+def register(snap_name, private, store):
     """Register <snap-name> with the store.
 
     You can use this command to register an available <snap-name> and become
@@ -114,7 +114,7 @@ def register(snap_name, private, store_id):
     if private:
         click.echo(_MESSAGE_REGISTER_PRIVATE.format(snap_name))
     if click.confirm(_MESSAGE_REGISTER_CONFIRM.format(snap_name)):
-        snapcraft.register(snap_name, is_private=private, store_id=store_id)
+        snapcraft.register(snap_name, is_private=private, store_id=store)
         click.echo(_MESSAGE_REGISTER_SUCCESS.format(snap_name))
     else:
         click.echo(_MESSAGE_REGISTER_NO.format(snap_name))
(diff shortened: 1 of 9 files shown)


snapcraft/cli/store.py
```diff
@@ -100,7 +100,8 @@ def _human_readable_acls(store: storeapi.StoreClient) -> str:
 @storecli.command()
 @click.argument("snap-name", metavar="<snap-name>")
 @click.option("--private", is_flag=True, help="Register the snap as a private one")
-def register(snap_name, private):
+@click.option("--store-id", metavar="<store>", help="Store ID to register with")
```
sparkiegeek: I think `--store` is better

```suggestion
@click.option("--store", metavar="<store>", help="Store to register with")
```

snapcraft/cli/store.py
```diff
@@ -100,7 +100,8 @@ def _human_readable_acls(store: storeapi.StoreClient) -> str:
...
+def register(snap_name, private, store_id):
```
sparkiegeek:
```suggestion
def register(snap_name, private, store):
```

snapcraft/cli/store.py
```diff
@@ -113,7 +114,7 @@ def register(snap_name, private):
     if private:
         click.echo(_MESSAGE_REGISTER_PRIVATE.format(snap_name))
     if click.confirm(_MESSAGE_REGISTER_CONFIRM.format(snap_name)):
-        snapcraft.register(snap_name, private)
+        snapcraft.register(snap_name, is_private=private, store_id=store_id)
```
sparkiegeek:
```suggestion
        snapcraft.register(snap_name, is_private=private, store=store)
```

snapcraft/storeapi/_sca_client.py
```diff
@@ -91,9 +91,13 @@ def register_key(self, account_key_request):
         if not response.ok:
             raise errors.StoreKeyRegistrationError(response)
 
-    def register(self, snap_name, is_private, series):
+    def register(
+        self, snap_name: str, *, is_private: bool, series: str, store_id: str
```
sparkiegeek:
```suggestion
        self, snap_name: str, *, is_private: bool, series: str, store: str
```

tests/integration/store/test_store_register.py
```diff
@@ -32,6 +32,11 @@ def test_successful_private_registration(self):
         snap_name = self.get_unique_name()
         self.register(snap_name, private=True)
 
+    def test_successful_registration_with_store(self):
+        self.login()
+        snap_name = self.get_unique_name(name_prefix="snapcraft-test")
```
cprov: What's the reason for using a different name prefix here ? it implies in additional cleanup steps on our staging service for no particular gain. We can continue to differentiate the name (in case it falls into the global store) by extending the agreed prefix, e.g.`test-snapcraft-in-store`.

sergiusens: I only did that because brand stores now require the prefix for the registered names to be that of the brand store.

cprov: Ah, forgot about the store name prefix, which works both ways (also blocking the use in the global store).Any objections to set the prefix to `test-snapcraft-in-store` ?

cprov: I've discussed with Adam and we updated the store name prefix to `test-snapcraft-in-store` to simplify the cleanup.

PR #2491, opened 2019-02-28
Fix multipass error handling spread test

cmatsuoka:
Fix the multipass error handler spread test but don't remove the manual switch as multipass tests in gcloud don't seem to be stable enough.

-----

### DIFF
--- a/tests/spread/multipass/error-handling/task.yaml
+++ b/tests/spread/multipass/error-handling/task.yaml
@@ -1,4 +1,4 @@
-summary: Build a basic snap using multipass and ensure that it runs
+summary: A faulty snap to test error handling
 warn-timeout: 9m  # Keep less than 10 minutes so Travis can't timeout
 priority: 90  # Run this test relatively early since fetching images can take time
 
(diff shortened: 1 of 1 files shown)


cmatsuoka:
Multipass tests on gcloud are not stable. Error handling is fixed but will keep multipass tests disabled to prevent creating random test failures in every PR.

tests/spread/multipass/error-handling/task.yaml
```diff
@@ -25,5 +25,6 @@ execute: |
   # Ensure that snapcraft uses build VMs
   unset SNAPCRAFT_BUILD_ENVIRONMENT
 
-  snapcraft
-  [ $? -eq 2 ]
+  err=0
```
sergiusens: how about using the same pattern as everywhere else

```suggestion
  if ! snapcraft; then
    echo "Expected <fill the blank>"
    exit 1
  fi
```

If we care about the code, then use your construct, but add some messaging please

cmatsuoka: The expected result is, in this case, 2. Result will be 1 if the error is not properly handled, so I'll keep the construct and add an error message.

PR #2500, opened 2019-03-12
cli: add remote build

cmatsuoka:
Add the capability of dispatching a source tree to be built remotely
in Launchpad. After the build is complete, retrieve the resulting snap
packages or the log files in case of error.

LP: #1805220

-----


snapcraft/__init__.py
```diff
@@ -339,6 +339,8 @@ def _get_version():
 
 from snapcraft._baseplugin import BasePlugin  # noqa
 
+from snapcraft._info_file import InfoFile  # noqa
```
sergiusens: please put this in internal or keep it here with the _ but do not global import it into the main namespace.

sergiusens: if this is generic enough, putting the class in something like file_utils should be good (not sure that is the best place yet)

sergiusens: Ah, maybe this could like in the config module (or it could be converted into a package)

snapcraft/cli/remote.py
```diff
@@ -0,0 +1,245 @@
...
+import uuid
+
+from snapcraft.internal.remote_build import Worktree, LaunchpadClient, Repo, errors
+from typing import List, Tuple
+from xdg import BaseDirectory
+from . import echo
+from ._options import get_project
+from snapcraft import InfoFile
+
+_SUPPORTED_ARCHS = ["amd64", "arm64", "armhf", "i386", "ppc64el", "s390x"]
```
sergiusens: I have this same structure for the details to info migration....
Maybe this should go into storeapi.constants

snapcraft/cli/remote.py
```diff
@@ -0,0 +1,245 @@
...
+        work_dir,
+        ignore=[name + "_*.snap", "buildlog_*.txt*", "parts", "stage", "prime"],
+    )
+    url = wt.add_remote(provider, user, build_id)
+    wt.sync()
+    wt.push()
+    return url
+
+
+def _check_supported_archs(archs: List[str]) -> None:
```
sergiusens: this could be part of storeapi

snapcraft/cli/remote.py
sergiusens: and made public

snapcraft/cli/remote.py
```diff
@@ -0,0 +1,245 @@
...
+
+    Command remote-build sends the current project to be built remotely. After the build
+    is complete, packages for each architecture are retrieved and will be available in
+    the local filesystem.
+
+    If the local files are under git revision control, use option --git to build the
+    current branch of this repository. Be aware, however, that untracked files and
+    uncommitted changes will not be used.
+
+    If not specified in the snapcraft.yaml file, the list of architectures to build
```
sergiusens: Feels like it should be the other way around, right?
CLI params like --arch should override what is pre-defined, don't you think?

snapcraft/cli/remote.py
```diff
@@ -0,0 +1,245 @@
...
+        build_id = remote_info["id"]
+        if "provider" in remote_info:
+            provider = remote_info["provider"]
+    else:
+        build_id = "snapcraft-{}".format(uuid.uuid4().hex)
+        remote_info["id"] = build_id
+        remote_info["provider"] = provider
+        remote_info.save()
+
+    # TODO: change login strategy after launchpad infrastructure is ready
```
sergiusens: can you add a bug number and list it here?

snapcraft/cli/remote.py
```diff
@@ -0,0 +1,245 @@
...
+    lp = LaunchpadClient(project, build_id)
+    lp.login(user)
+
+    if status:
+        # Show build status
+        lp.recover_build(status)
+        lp.show_build_status()
+        return
```
sergiusens: I have a bad feeling when I see returns like this that are not presented as guards at the beginning, how about removing it and as a simple solution do `if not status:` for the block at the end (add a code comment perhaps).

Alternatively split these out into 3 methods

snapcraft/cli/remote.py
```diff
@@ -0,0 +1,245 @@
...
+        echo.info("Sending data to remote builder...")
+        if git:
+            url, branch = _send_current_tree(provider, lp.user, build_id + "-git")
+        else:
+            url = _copy_and_send_tree(
+                provider, lp.user, remote_dir, build_id, project.info.name
+            )
+
+        # Create build recipe
+        lp.delete_snap()
```
sergiusens: Can you comment on why we are deleting first?

snapcraft/internal/remote_build/_launchpad.py
```diff
@@ -0,0 +1,227 @@
...
+        self._cache_dir = os.path.join(
+            BaseDirectory.save_cache_path("snapcraft"), "launchpad"
+        )
+
+        os.makedirs(self._data_dir, mode=0o700, exist_ok=True)
+        self._credentials = os.path.join(self._data_dir, "credentials")
+
+    def login(self, user: str) -> None:
+        info = InfoFile(os.path.join(self._data_dir, "config.yaml"))
+        if user:
```
sergiusens: I would prefer this to look more like the config class we have for error handling and other preferences.

We already have credentials config available, maybe we just need to add the launchpad stuff there and not require Info

sergiusens: this user needs to be escaped

snapcraft/__init__.py
cmatsuoka: Changed to be local to remote-build, then we can decide later how to generalize this for remote build and build providers.

snapcraft/cli/remote.py
cmatsuoka: Refactored final block into a different function and removed return from build status case.

snapcraft/cli/remote.py
cmatsuoka: Add comment to code (delete any existing snap to remove leftovers from previous builds).

snapcraft/cli/remote.py
cmatsuoka: I propose we land this first, and then refactor to use a common infrastructure for both cases.

snapcraft/cli/remote.py
cmatsuoka: Ditto for this one.

snapcraft/cli/remote.py
cmatsuoka: Agree that parameters should override, but the precedence is set by the remote builder in this case. As discussed in our meeting, will refactor to not allow both set at the same time.

snapcraft/cli/remote.py
cmatsuoka: Add reference to [LP bug 1827679](https://bugs.launchpad.net/snapcraft/+bug/1827679).

snapcraft/internal/remote_build/_launchpad.py
cmatsuoka: Changed configuration data storage to use the existing infrastructure used for credentials, under a new "Launchpad" section.

cmatsuoka: Encoding with urllib before storing in the config file. Perhaps sanity-checking usernames would make escaping unnecessary? (assuming reasonably well-behaved valid usernames)

snapcraft/cli/remote.py
```diff
@@ -224,9 +228,7 @@ def _list_architectures(project, arch: str) -> List[str]:
     arch_from_yaml = _project_architectures(project)
     if arch_from_yaml:
         if arch:
-            echo.warning(
-                "Architecture list already set in snapcraft.yaml, ignoring --arch option."
-            )
+            raise errors.ConflictingArchListError()
```
sergiusens: this could be click.BadOptionUsage instead of a full blown exception in errors (given you are doing the check here and not internally in the remote_build package

snapcraft/cli/remote.py
```diff
@@ -0,0 +1,252 @@
...
+    required=False,
+    help="Display remote build status.",
+)
+@click.option(
+    "--arch",
+    metavar="<arch-list>",
+    type=str,
+    nargs=1,
+    required=False,
+    help="Set architectures to build.",
```
sergiusens: the list of valid arches could be passed here as a param to click.option

snapcraft/cli/remote.py
```diff
@@ -0,0 +1,252 @@
...
+            url = _copy_and_send_tree(
+                provider, lp.user, remote_dir, build_id, project.info.name
+            )
+
+        # Create build recipe
+        # (delete any existing snap to remove leftovers from previous builds)
+        lp.delete_snap()
+        lp.create_snap(url, branch, archs)
+
+        targets = " for {}".format(", ".join(archs)) if archs else ""
```
sergiusens: we have a humanize_list in formatting_utils which should be used here

snapcraft/cli/remote.py
```diff
@@ -0,0 +1,252 @@
...
+        lp.show_build_status()
+    elif recover:
+        # Recover from interrupted build
+        echo.info("Recover build {}...".format(recover))
+        lp.recover_build(recover)
+        _monitor_build(lp)
+    else:
+        # If build architectures not set in snapcraft.yaml, let the user override
+        # Launchpad defaults using --arch.
+        archs = _list_architectures(project, arch)
```
sergiusens: before this, it would be good to do and
project_architectures = _get_project_architectures(....)
if  project_architectures and arch:
    raise click.BadOptionUsage(....)

snapcraft/cli/remote.py
```diff
@@ -0,0 +1,252 @@
...
+        archs = arch_from_yaml
+    elif arch == "all":
+        archs = _SUPPORTED_ARCHS
+    else:
+        archs = arch.split(",") if arch else []
+
+    return archs
+
+
+def _project_architectures(project) -> List[str]:
```
sergiusens: can you add an action to this, get seems appropriate, so _get_project_architectures

snapcraft/internal/remote_build/_launchpad.py
```diff
@@ -0,0 +1,233 @@
...
+import urllib.request
+import urllib.error
+import urllib.parse
+
+from lazr import restfulclient
+from launchpadlib.launchpad import Launchpad
+from typing import Any, Dict, List
+from xdg import BaseDirectory
+from snapcraft.project import Project
+from .errors import (
```
sergiusens: why not import "errors" instead? Optional, but it follows the pattern used in other packages we have

snapcraft/internal/remote_build/_launchpad.py
```diff
@@ -0,0 +1,233 @@
...
+                arch = build["arch_tag"]
+                build_state = build["buildstate"]
+                logger.debug("{} state: {}".format(arch, build_state))
+                if arch in self._waiting:
+                    if build_state == _LP_SUCCESS_STATUS:
+                        self._process_build(build)
+                    elif build_state == _LP_FAIL_STATUS:
+                        self._process_fail(build)
+
+    def show_build_status(self) -> None:
```
sergiusens: can this be "get_build_status" and have the user of this method print it instead?

snapcraft/cli/remote.py
cmatsuoka: It makes sense. Fixed this along with the code refactoring describe three comments below.

snapcraft/cli/remote.py
cmatsuoka: Renamed as described.

snapcraft/cli/remote.py
cmatsuoka: List humanized.

snapcraft/cli/remote.py
cmatsuoka: Unfortunately `type=click.Choice()` doesn't seem to fit this situation without changing the parameter format. We can either leave it without the choice list, or change the list separator to space instead of comma. Comma-separated seems more natural to me, other members of a space-separated list feel like they're other arguments:
```
$ snapcraft remote-build --arch=i386,amd64,armhf --git

$ snapcraft remote-build --arch=i386 amd64 armhf --git
```

snapcraft/internal/remote_build/_launchpad.py
cmatsuoka: Done.

snapcraft/internal/remote_build/_launchpad.py
cmatsuoka: I'm all for consistency, changed to import errors.

snapcraft/cli/remote.py
```diff
@@ -139,10 +139,15 @@ def remote_build(
     else:
         # If build architectures not set in snapcraft.yaml, let the user override
         # Launchpad defaults using --arch.
-        archs = _list_architectures(project, arch)
+        project_architectures = _get_project_architectures(project)
+        if project_architectures and arch:
+            raise click.BadOptionUsage(
+                "Can't use --arch, architecture list is already set in snapcraft.yaml."
```
sergiusens: these should not be contracted, instead of Can't we should use Cannot

sergiusens:
```suggestion
                "Cannot use --arch, architecture list is already set in snapcraft.yaml."
```

snapcraft/cli/remote.py
```diff
@@ -0,0 +1,256 @@
...
+    required=False,
+    help="Set architectures to build.",
+)
+@click.option(
+    "--git", is_flag=True, required=False, help="Build a local git repository."
+)
+@click.option(
+    "--user", metavar="<username>", nargs=1, required=False, help="Launchpad username."
+)
+def remote_build(
```
sergiusens: can you pop this command just like we do for "promote"?

snapcraft/cli/remote.py
```diff
@@ -0,0 +1,256 @@
...
+
+    \b
+    Examples:
+        snapcraft remote-build
+        snapcraft remote-build --arch=all
+        snapcraft remote-build --git --arch=arm64,armhf
+        snapcraft remote-build --recover 47860738
+        snapcraft remote-build --status 47860738
+    """
+
```
sergiusens: can you add a similar warning as we have for promote right here?

snapcraft/cli/remote.py
cmatsuoka: Done.

snapcraft/cli/remote.py
cmatsuoka: Warning added.

snapcraft/cli/remote.py
```diff
@@ -101,6 +101,9 @@ def remote_build(
         snapcraft remote-build --recover 47860738
         snapcraft remote-build --status 47860738
     """
+    echo.warning(
+        "snapcraft remote-build does not have a stable CLI interface. Use with caution in scripts."
```
sergiusens: When I said similar I was thinking about mentioning upcoming workflow changes, like the authentication mechanism and ssh keys.

cmatsuoka: Oh ok rewording that.

sergiusens:
Seems like we need to set "GIT_PYTHON_GIT_EXECUTABLE".... how we got to have a bad git executable I do not know

snapcraft/cli/remote.py
```diff
@@ -0,0 +1,260 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import click
+import os
+import uuid
+
+from snapcraft.internal.remote_build import (
+    Worktree,
+    LaunchpadClient,
+    Repo,
```
sergiusens: try lazy loading this before use, before loading, check if git is available and print out a nicer message.

PR #2502, opened 2019-03-14
meta: fix management of snap/local

sergiusens:
Do not copy snap/local into the snap, furthermore, improve the copy
logic to work better with embedded directories.

Fixes SNAPCRAFT-J2
Fixes SNAPCRAFT-JB

-----

### DIFF
--- a/snapcraft/internal/meta/_snap_packaging.py
+++ b/snapcraft/internal/meta/_snap_packaging.py
@@ -433,33 +433,31 @@ class _SnapPackaging:
                 yaml_utils.dump(annotated_snapcraft, stream=manifest_file)
 
     def write_snap_directory(self) -> None:
+        assets_dir = self._project_config.project._get_snapcraft_assets_dir()
         # First migrate the snap directory. It will overwrite any conflicting
         # files.
-        for root, directories, files in os.walk(
-            self._project_config.project._get_snapcraft_assets_dir()
-        ):
-            with contextlib.suppress(ValueError):
-                directories.remove(".snapcraft")
-            with contextlib.suppress(ValueError):
-                # The snapcraft.yaml is migrated later
-                files.remove("snapcraft.yaml")
+        for root, directories, files in os.walk(assets_dir):
+            if root == assets_dir:
+                with contextlib.suppress(ValueError):
+                    directories.remove(".snapcraft")
+                with contextlib.suppress(ValueError):
+                    # The objective of local inside the snap dir is to
+                    # use it as an additional source so there is little
+                    # sense in copying it in here.
+                    directories.remove("local")
+                with contextlib.suppress(ValueError):
+                    # The snapcraft.yaml is migrated later
+                    files.remove("snapcraft.yaml")
 
             for directory in directories:
-                source = os.path.relpath(
-                    os.path.join(root, directory),
-                    self._project_config.project._work_dir,
-                )
-                # Also build-aux from destination
-                destination = os.path.join(self._prime_dir, source.lstrip("build-aux/"))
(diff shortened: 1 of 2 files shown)


snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -433,43 +433,39 @@ def _record_manifest_and_source_snapcraft_yaml(self):
...
         # files.
-        for root, directories, files in os.walk(
-            self._project_config.project._get_snapcraft_assets_dir()
-        ):
-            with contextlib.suppress(ValueError):
-                directories.remove(".snapcraft")
-            with contextlib.suppress(ValueError):
-                # The snapcraft.yaml is migrated later
-                files.remove("snapcraft.yaml")
+        for root, directories, files in os.walk(assets_dir):
```
kyrofa: So let's see... at this point, I think all this is doing is migrating local plugins over. I don't like the catch-all copy, I think that a targeted approach is probably better. The important stuff is the snapcraft.yaml, the gui dir, and the hooks dir. If we're copying local plugins, that should probably only happen if the snapcraft.yaml is being copied as well. Anything else we care about?

sergiusens: I agree, but I also think, why are we copying at all?

sergiusens: I think we do not care about anything in there, the "manifest recording" logic is backwards as it wipes instead making everything all so complicated.

snap/hooks installed through a part is migrated through the pluginhandler, the logic here would only be useful for some form of consistency to say that snap/hooks takes precedence.

Aside from that, we just want GUI and potentially snapcraft.yaml, but I'd just copy that if needed instead.

To avoid great changes, I will just whitelist this tomorrow morning and also look into the PR this landed in to see if there is any indication of why this was done this way.

Thanks for taking a look.

sergiusens: ok, looking at when it was introduced https://github.com/snapcore/snapcraft/pull/1026/files#diff-6ddd7f2a83b2f307ff953c3fb60ae35dR126 I think the original reason to copy to snap/hooks was that, in the case of hooks, we would wrap them up in prime/meta/hooks.

We ended up only doing this for things that were primed, so there is no reason to copy these into prime/snap/hooks at all, except for parities sake and for the convention that anything defined in <project-root>/snap/hooks takes precedence over anything in prime/snap.

For this reason only it makes sense to keep copying these in, but not as broadly as proposed in the original implementation where we end up deleting most things after the fact.

kyrofa: Agreed.

snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -433,40 +433,15 @@ def _record_manifest_and_source_snapcraft_yaml(self):
...
-        for origin in ["gui", "hooks"]:
-            src_dir = os.path.join(assets_dir, origin)
-            dst_dir = os.path.join(self.meta_dir, origin)
+        snap_assets_dir = self._project_config.project._get_snapcraft_assets_dir()
+        prime_snap_dir = os.path.join(self._prime_dir, "snap")
+
+        snap_dir_iter = itertools.product([prime_snap_dir], ["hooks"])
+        meta_dir_iter = itertools.product([self._meta_dir], ["hooks", "gui"])
+
+        for origin in itertools.chain(snap_dir_iter, meta_dir_iter):
```
cmatsuoka: This is a bit convoluted way to set src_dir and dst_dir, I hope we still understand the logic in the future.

PR #2504, opened 2019-03-18
cli: validate plugin schema before provider

cmatsuoka:
When using build providers, validate our yaml file schema prior to
executing ourselves inside the provider instance. This allows us
to quickly interrupt execution if validation fails.

LP: #1807875

Note: this is the second part of the original PR #2473 which was split in two parts.

-----


sergiusens:
I think we should close this, loading of plugins can fail on non linux systems, we need to rather move the plugin schemas to the general schema.

PR #2509, opened 2019-03-21
build providers: initial support for LXD

sergiusens:
Allow running snapcraft with --use-lxd for all the lifecycle steps and
cleaning.

LP: #1805221

-----


Lin-Buo-Ren:
I noticed the following warning message while running `snapcraft-pr.init 2509`:

```
pylxd 2.2.9 has requirement requests-toolbelt>=0.8.0, but you'll have requests-toolbelt 0.6.0 which is incompatible.
```

Lin-Buo-Ren:
You probably have already noticed, but the Snapcraft still printing `Launching a VM.` when being run with `--with-lxd`

Lin-Buo-Ren:
I noticed that snapcraft doesn't pick up the default profile in LXD, is it possible to at least making the cloud-init config customizable so that we can use a local APT mirror?

sergiusens:
Changes to this branch should make it to the beta/use-lxd channel.
If interested `snap refresh --channel beta/use-lxd snapcraft`

Lin-Buo-Ren:
```
$ snap refresh --channel beta/use-lxd snapcraft
error: requested a non-existing branch on latest/beta for snap
       "snapcraft": use-lxd
```

Lin-Buo-Ren:
Currently the build will error if the container isn't created in advance:

```
  File "/home/brlin/snap/snapcraft-pr/10/venvs/2509/lib/python3.6/site-packages/snapcraft/internal/build_providers/_base_provider.py", line 220, in launch_instance
    self._start()
  File "/home/brlin/snap/snapcraft-pr/10/venvs/2509/lib/python3.6/site-packages/snapcraft/internal/build_providers/_lxd/_lxd.py", line 91, in _start
    raise errors.ProviderInstanceNotFoundError(instance_name=self.instance_name)
snapcraft.internal.build_providers.errors.ProviderInstanceNotFoundError: Cannot find an instance named 'snapcraft-links'.
```

Lin-Buo-Ren:
I created a new container using a template that has set the `SNAPCRAFT_BUILD_ENVIRONMENT=host` environment variable in `/etc/profile` but the multipass installation prompt still appears, is Snapcraft not using a login shell?

sergiusens:
El viernes, 22 de marzo de 2019 18:47:06 -03 林博仁(Buo-ren Lin) escribió:
> ```
> $ snap refresh --channel beta/use-lxd snapcraft
> error: requested a non-existing branch on latest/beta for snap
>        "snapcraft": use-lxd
> ```

The branch is just lxd, so `snap refresh snapcraft --channel beta/lxd`

evandandrea:
It is really confusing that this option does not exist when you're in a directory with a snapcraft.yaml that does not use bases. I know the snapcraft-inside-snapcraft is the reason for this, but I think we should fix it.

sergiusens:
El miércoles, 27 de marzo de 2019 08:01:47 -03 Evan escribió:
> It is really confusing that this option does not exist when you're in a
> directory with a snapcraft.yaml that does not use bases. I know the
> snapcraft-inside-snapcraft is the reason for this, but I think we should
> fix it.

I have plans to fix it, but requires some refactoring to get rid of where the 
re-execution takes place.

snapcraft/internal/build_providers/_lxd/_lxd.py
```diff
@@ -0,0 +1,277 @@
...
+    _PROJECT_DEVICE_NAME = "snapcraft-project"
+
+    @classmethod
+    def ensure_provider(cls):
+        try:
+            if repo.snaps.SnapPackage.is_snap_installed("lxd"):
+                return
+        except repo.errors.SnapdConnectionError as snapd_error:
+            if sys.platform == "linux":
+                error_message = "snap support is required to continue: {}".format(
```
evandandrea: Why are we showing the full error when we know exactly what the problem is? How about:
>Snap support and the lxd snap are required to continue: https://docs.snapcraft.io/installing-snapd/6735

snapcraft/internal/build_providers/_lxd/_lxd.py
```diff
@@ -0,0 +1,277 @@
...
+                    snapd_error
+                )
+            else:
+                error_message = "LXD is not supported on this platform"
```
evandandrea: How about:
>LXD is not supported on this platform. Use multipass to run snapcraft on non-Linux systems: https://github.com/CanonicalLtd/multipass

snapcraft/internal/build_providers/_lxd/_lxd.py
```diff
@@ -0,0 +1,277 @@
...
+                prompt_installable=False,
+                error_message=error_message,
+            ) from snapd_error
+
+        # At this point, we know that if we are on linux that we have
+        # snap support.
+        raise errors.ProviderNotFound(
+            provider=cls._get_provider_name(),
+            prompt_installable=True,
+            error_message="The LXD snap is required to continue",
```
evandandrea: How about:
>The lxd snap is required to continue: snap install lxd

snapcraft/internal/build_providers/_lxd/_lxd.py
```diff
@@ -0,0 +1,277 @@
...
+            subprocess.check_output(["lxd", "waitready", "--timeout=30"])
+        except subprocess.CalledProcessError as call_error:
+            raise SnapcraftEnvironmentError(
+                "Timeout reached waiting for LXD to start."
+            ) from call_error
+
+        try:
+            subprocess.check_output(["lxd", "init", "--auto"])
+        except subprocess.CalledProcessError as call_error:
+            raise SnapcraftEnvironmentError("Failed to initialize LXD.") from call_error
```
evandandrea: How about:
>Failed to initialize LXD. Try manually initializing before trying again: lxd init --auto

snapcraft/internal/build_providers/_lxd/_lxd.py
```diff
@@ -0,0 +1,277 @@
...
+
+    def shell(self) -> None:
+        self._run(command=["/bin/bash"])
+
+    def _ensure_container_running(self) -> None:
+        self._container.sync()
+        if self._container.status.lower() != "running":
+            raise errors.ProviderFileCopyError(
+                provider_name=self._get_provider_name(),
+                error_message="container is not running, it is current state is: {!r].".format(
```
evandandrea: "it is current state is"?

What does this error mean, that the LXD instance died? What might cause this?

snapcraft/internal/build_providers/_lxd/_lxd.py
```diff
@@ -0,0 +1,277 @@
...
+            self._container.delete(wait=True)
+        return was_cleaned
+
+    def pull_file(self, name: str, destination: str, delete: bool = False) -> None:
+        self._ensure_container_running()
+
+        # TODO: better handling of larger files.
+        try:
+            source_data = self._container.files.get(name)
+        except pylxd.exceptions.LXDAPIException as lxd_api_error:
```
evandandrea: If pylxd.exceptions.LXDAPIException happens it's a bug in LXD, right? Should we instruct them to file one?

sergiusens: This can be anything, not necessarily a bug... They only have one exception so we cannot know exactly what is going on, but they use the stdlib exceptions for anything not return from the lxd socket.

Examples of this are:
- Failed to reach the lxd remote
- Failed to create the container
- Failed to save the config

Think of it as multipass exiting with != 0

snapcraft/internal/build_providers/_lxd/_lxd.py
sergiusens: good typo catch, could happen if someone stops the container or removes LXD mid run.

snapcraft/internal/build_providers/_lxd/_lxd.py
sergiusens: Sounds good.

snapcraft/internal/build_providers/_lxd/_lxd.py
sergiusens: This message is what makes it to click.confirm, so you will be prompted with a yes/no with that message. Let me see if I can fit this in somehow

snapcraft/internal/build_providers/_lxd/_lxd.py
sergiusens: Yeah, that looks good, although we could also tell people to just run without --use-lxd, what do you think?

snapcraft/internal/build_providers/_lxd/_lxd.py
sergiusens: Indeed

snapcraft/internal/build_providers/_lxd/_lxd.py
```diff
@@ -45,30 +45,49 @@ class LXD(Provider):
...
+        # If we reach this point, it means the lxd snap is properly setup.
+        # Now is the time for additional sanity checks to ensure the provider
+        # will work.
+        try:
+            # TODO: add support for more distributions. Maybe refactor a bit so that Repo behaves
+            # similar to a build provider.
+            if repo.Repo.is_package_installed("lxd") or repo.Repo.is_package_installed(
+                "lxd-client"
+            ):
+                raise SnapcraftEnvironmentError(
```
cmatsuoka: A comment on what happened to me when I allowed snapcraft to install the lxd snap (I was previously using the deb packages): it installed lxd, and successfully built my package. Then in the next attempt to run snapcraft, it refused to run because I had the deb packages installed -- but it ran the first time. Also I lost my custom default profile in the process and had to recreate it in the new lxd installation -- I wonder what else from my previous configuration has changed.

sergiusens: The first time, before doing the iterative PRs to make this smaller involved setting LXD_DIR, then with a refactor I removed it as it was mostly a hack. So a mixup of deb and snap went into place.

pylxd does not support both installations at the same time in an official capacity but can be made to work.

I want to work the lxd team to get the best mechanisms to have lxd tell me where its socket lived, alas it was determined that we needed to hardcode the path to it (seen in later versions of this PR).

We do nothing special with the profile. But I do suspect having both installed at a point in time triggered `lxd init --auto` from /usr/bin instead of /snap/bin and that is now solved with the abspath to the binary and that process might have wiped your profile.

To solve the deb problem, we now just decide to falloff if the deb is installed.

Lastly, I want to add that to not lose stuff you need to run lxd.migrate from the snap to move away from the deb.

snapcraft/internal/build_providers/_lxd/_lxd.py
sergiusens: I am not sure about this one now... this might be the wrong way to solve this.

sergiusens: I will fix in a different PR provided I get a bug for this :smile:
But the gist of it, is that we should solve this at a higher layer to keep a good separation of concerns.

snapcraft/internal/build_providers/_lxd/_lxd.py
sergiusens: lxd.migrate as part of commit 13cf4b9b

snapcraft/internal/build_providers/_lxd/_lxd.py
evandandrea: https://bugs.launchpad.net/snapcraft/+bug/1822545

snapcraft/cli/lifecycle.py
```diff
@@ -90,8 +90,10 @@ def _execute(  # noqa: C901
             build_provider_class.ensure_provider()
         except build_providers.errors.ProviderNotFound as provider_error:
             if provider_error.prompt_installable:
-                click.echo(str(provider_error))
-                if click.confirm("Would you like to install it now?"):
+                if click.confirm(
+                    "Support for {!r} needs to be setup. "
```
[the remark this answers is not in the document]
sergiusens: true http://www.future-perfect.co.uk/grammar-tip/is-it-setup-set-up-or-set-up/

PR #2512, opened 2019-03-24
many: better handling of appstream icons

cmatsuoka:
Appstream icons can be specified as a full path ("local") or just the
application name to be searched in the standard locations ("stock").
The system search for stock icons doesn't include directories that are
part of snap packages, so support stock icons by resolving the path
at packaging time. If icons are not specified in the AppStream file,
use the desktop file definition instead.

LP: #1814898

-----

### DIFF
--- a/snapcraft/extractors/appstream.py
+++ b/snapcraft/extractors/appstream.py
@@ -17,7 +17,7 @@
 import operator
 import os
 from io import StringIO
-from typing import List
+from typing import List, Optional
 
 import lxml.etree
 
@@ -77,21 +77,6 @@ def extract(relpath: str, *, workdir: str) -> ExtractedMetadata:
     summary = _get_value_from_xml_element(dom, "summary")
     description = _get_value_from_xml_element(dom, "description")
 
-    node = dom.find("icon")
-    icon = node.text.strip() if node is not None else None
-
-    if node is not None and "type" in node.attrib:
-        node_type = node.attrib["type"]
-    else:
-        node_type = None
-
-    if node_type == "remote":
-        # TODO Currently we only support local and stock icons.
-        # See bug https://bugs.launchpad.net/snapcraft/+bug/1742348
-        # for supporting remote icons.
-        # --elopio -20180109
-        icon = None
-
     desktop_file_paths = []
     desktop_file_ids = _get_desktop_file_ids_from_nodes(dom.findall("launchable"))
     # if there are no launchables, use the appstream id to take into
@@ -104,10 +89,7 @@ def extract(relpath: str, *, workdir: str) -> ExtractedMetadata:
         if desktop_file_path:
             desktop_file_paths.append(desktop_file_path)
 
-    if node_type == "stock":
-        icon = _get_icon_from_theme(workdir, "hicolor", icon)
-    else:
-        icon = _extract_icon(workdir, icon, desktop_file_paths)
+    icon = _extract_icon(dom, workdir, desktop_file_paths)
 
     return ExtractedMetadata(
         common_id=common_id,
@@ -140,7 +122,7 @@ def _get_xslt():
     return lxml.etree.XSLT(xslt)
 
 
-def _get_value_from_xml_element(tree, key) -> str:
+def _get_value_from_xml_element(tree, key) -> Optional[str]:
     node = tree.find(key)
     if node is not None:
         return node.text.strip()
@@ -156,7 +138,7 @@ def _get_desktop_file_ids_from_nodes(nodes) -> List[str]:
     return desktop_file_ids
 
 
-def _desktop_file_id_to_path(desktop_file_id: str, *, workdir: str) -> str:
+def _desktop_file_id_to_path(desktop_file_id: str, *, workdir: str) -> Optional[str]:
     # For details about desktop file ids and their corresponding paths, see
     # https://standards.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#desktop-file-id  # noqa
     for xdg_data_dir in ("usr/local/share", "usr/share"):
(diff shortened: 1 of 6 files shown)


snapcraft/extractors/appstream.py
```diff
@@ -75,14 +77,20 @@ def extract(relpath: str, *, workdir: str) -> ExtractedMetadata:
     summary = _get_value_from_xml_element(dom, "summary")
     description = _get_value_from_xml_element(dom, "description")
 
-    icon = None
     node = dom.find("icon")
-    if node is not None and "type" in node.attrib and node.attrib["type"] == "local":
-        # TODO Currently we only support local icons.
+    icon = node.text.strip() if node is not None else None
```
sergiusens: how about moving this code block for icon management closer to where you use it again after the desktop file parsing?
Maybe even consider renaming node to icon_node, node_type to icon_node_type OR even better, extract all this into its own method or better yet and Icon-like class

snapcraft/internal/meta/_desktop.py
```diff
@@ -38,9 +39,9 @@ def __init__(
                 filename, "does not exist (defined in the app {!r})".format(name)
             )
 
-    def parse_and_reformat(self):
+    def parse_and_reformat(self, extracted_metadata: _metadata.ExtractedMetadata):
         self._parser = configparser.ConfigParser(interpolation=None)
-        self._parser.optionxform = str
+        self._parser.optionxform = str  # type: ignore
```
sergiusens: why is this type ignored?

snapcraft/internal/meta/_desktop.py
```diff
@@ -56,7 +57,15 @@ def parse_and_reformat(self):
             exec_value = "{}.{} %U".format(self._snap_name, self._name)
         self._parser[section]["Exec"] = exec_value
         if "Icon" in self._parser[section]:
-            icon = self._parser[section]["Icon"]
+            # Extracted metadata (e.g. from the AppStream) can override the
+            # icon location.
+            if extracted_metadata:
+                metadata_dict = extracted_metadata.to_dict()
```
sergiusens: why not use `get_icon()`? It should return None if not found or set.

snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -129,7 +129,7 @@ def _update_yaml_with_extracted_metadata(
     config_data: Dict[str, Any],
     parts_config: project_loader.PartsConfig,
     prime_dir: str,
-) -> None:
+) -> extractors.ExtractedMetadata:
     if "adopt-info" in config_data:
```
sergiusens: can we instead of `return None` at the very end, just add a check here,
```suggestion
    if "adopt-info" not in config_data:
        return None
```

And move everything beneath this one level up?

snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -129,7 +129,7 @@ def _update_yaml_with_extracted_metadata(
     config_data: Dict[str, Any],
     parts_config: project_loader.PartsConfig,
     prime_dir: str,
-) -> None:
+) -> extractors.ExtractedMetadata:
```
sergiusens: this should be type wrapped with Optional given that None can be returned

snapcraft/internal/pluginhandler/_metadata_extraction.py
```diff
@@ -30,7 +30,7 @@
 
 
 def extract_metadata(
-    part_name: str, file_relpath: str, workdir: str
+    part_name: str, file_relpath: str, workdir: str, primedir: str = None
```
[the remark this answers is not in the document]
cmatsuoka: Leftover from previous implementation that was refactored. Cleaning up.

snapcraft/internal/meta/_desktop.py
cmatsuoka: Otherwise mypy complains about "Incompatible types in assignment", see https://github.com/python/mypy/issues/5062

However, I don't remember seeing this error before, and mypy remains at version 0.590.

snapcraft/internal/meta/_desktop.py
cmatsuoka: Changed to use get_icon(), but also refactored the override logic to prevent icon to remain unassigned.

snapcraft/extractors/appstream.py
cmatsuoka: Refactored to move node parsing into the icon extraction function, and cleaned up the logic a little bit there.

snapcraft/internal/meta/_snap_packaging.py
cmatsuoka: Yes, it looks better. Changed this one, but I think meta has still many other places where it can be improved in a future refactoring.

snapcraft/internal/meta/_desktop.py
sergiusens: Can you just add a comment above that line, something like

    # mypy type checking ignored, see python/mypy#5062

snapcraft/internal/meta/_desktop.py
```diff
@@ -57,14 +57,14 @@ def parse_and_reformat(self, extracted_metadata: _metadata.ExtractedMetadata):
...
             # icon location.
             if extracted_metadata:
-                metadata_dict = extracted_metadata.to_dict()
-                if "icon" in metadata_dict:
-                    icon = "/" + str(metadata_dict["icon"])
-            else:
-                icon = self._parser[section]["Icon"]
+                metadata_icon = extracted_metadata.get_icon()
+                if metadata_icon is not None:
+                    icon = "/" + metadata_icon
```
sergiusens: I wonder why we add this slash here to strip it a line below. :man_shrugging: 
Not that you did this, but maybe worth a fix :smile:

snapcraft/internal/meta/_desktop.py
```diff
@@ -57,14 +57,14 @@ def parse_and_reformat(self, extracted_metadata: _metadata.ExtractedMetadata):
             exec_value = "{}.{} %U".format(self._snap_name, self._name)
         self._parser[section]["Exec"] = exec_value
         if "Icon" in self._parser[section]:
+            icon = self._parser[section]["Icon"]
```
sergiusens: Maybe this should have `lstrip("/")` and we can remove the if below.

snapcraft/internal/meta/_desktop.py
sergiusens: if metadata_icon even had a leading /, by adding one, the logic below makes no sense.

cmatsuoka: Indeed it doesn't. Refactored existing code to remove unnecessary maneuvering with initial slash.

cmatsuoka: Oh, there is a situation where the original code makes sense: stock icons in classic confinement. In this case, the base name of the icon is specified instead of the full path and the system will search on the standard system theme locations.

apachelogger:
_get_icon_from_theme is not spec compliant unfortunately.

Any folder may contain any file format. And there is nothing preventing an application from installing svgs into e.g. 22x22/ or PNGs in scalable/. In fact, it's very likely even, smaller resolutions call for less detail, so install a less detailed SVG into a resolution folder isn't unheard of.

Also the folders may be called anything and have any amount nesting so long as they are defined in the index.theme.

apachelogger:
It also seems to pick PNG over SVG, even though given the fact that the meta icon is used for the store (I think?) a scalable variant is likely more suitable in just about all cases.

e.g. https://github.com/apachelogger/snap-appstreamtest seems to pick up usr/share/icons/hicolor/64x64/apps/kbruch.png rather than the svg

PR #2515, opened 2019-03-25
build providers: support for provider setup

sergiusens:
Move checks and setup to the provider implementation.

LP: #1821586

-----

### DIFF
--- a/snapcraft/cli/lifecycle.py
+++ b/snapcraft/cli/lifecycle.py
@@ -14,12 +14,9 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import subprocess
 import logging
 import typing
-import sys
 import os
-from time import sleep
 
 import click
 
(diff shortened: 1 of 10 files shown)


snapcraft/internal/build_providers/errors.py
```diff
@@ -36,17 +36,19 @@ def __init__(self, *, provider: str) -> None:
...
-class ProviderCommandNotFound(ProviderBaseError):
+class ProviderNotFound(ProviderBaseError):
 
-    fmt = (
-        "{command!r} command not found: this command is necessary to build in "
-        "this environment.\n"
-        "Install {command!r} or if already installed, ensure it is "
-        "on the system PATH, and try again."
-    )
+    fmt = "You need {provider!r} setup to build snaps: {error_message}."
```
cmatsuoka: Shouldn't this be "set up"?

snapcraft/cli/lifecycle.py
```diff
@@ -14,32 +14,23 @@
...
     project_loader,
     steps,
-    repo,
 )
-from snapcraft.project._sanity_checks import (
-    conduct_project_sanity_check,
-    conduct_build_environment_sanity_check,
-)
-from snapcraft.project.errors import MultipassMissingInstallableError
+from snapcraft.project._sanity_checks import conduct_project_sanity_check
```
cmatsuoka: I believe this would be cleaner if conduct_project_sanity_check is exported by snapcraft.project instead of accessing project._sanity_checks directly from a different module.

sergiusens: the option is to either keep the module private or make the method private (with _).
I am not sure which I prefer, but considering that this is not part of the PR, I think we should leave as is :wink:

snapcraft/internal/build_providers/errors.py
sergiusens: Yeah, I guess

popey: "set-up" IMO

Issue #66, opened 2015-11-02
catkin plugin catches the wrong exception


sergiusens:
The plugin is catching `os.FileNotFound` when it should be catching `IOError`

PR #2519, opened 2019-03-29
packaging: use local patchelf

cmatsuoka:
Pull patchelf from the latest stable tag from the upstream repository,
which contains patches critical for snapcraft to work properly with
Go ELF binaries.

Noteworthy patches and issues fixed:

- Avoid inflating file sizes needlessly and allow binaries to be stripped
- patchelf --set-interpreter fails with `cannot find section` (#66)
- --set-interpreter gives "cannot find section" for go binary (#138)

LP: #1805205

-----


cmatsuoka:
This will require additional work. `--set-interpreter` is now working with patchelf bug 66 fixed, but bug 146/152 isn't and `--set-rpath` is making some binaries segfault.

Checking with readelf, we see the section layout in the gotty binary changing from:
```
  LOAD           0x00000000007e6000 0x0000000000be6000 0x0000000000be6000
                 0x0000000000083840 0x00000000000a5538  RW     0x1000
  DYNAMIC        0x00000000007e6140 0x0000000000be6140 0x0000000000be6140
                 0x0000000000000130 0x0000000000000130  RW     0x8
```
to
```
  LOAD           0x0000000000000000 0x0000000000400000 0x0000000000400000
                 0x00000000003b1330 0x00000000003b1330  R E    0x1000
  DYNAMIC        0x0000000000000270 0x0000000000400270 0x0000000000400270
                 0x0000000000000140 0x0000000000000140  RW     0x8
```
which will still cause problems with cgo binaries.

cmatsuoka:
Instead of fixing patchelf to be able to set rpath in Go binaries, we changed the strategy to relink dynamic executables with the system linker so that patchelf will be able to handle them. Patchelf 0.10 is still needed to set the interpreter without the workarounds put in place for patchelf 0.9. This solves the rpath setting problem for binaries we build (but not for pre-built dynamic Go binaries already linked with the Go linker).

sergiusens:
Looks good! Just wondering if you can add a spread test. Feel free to use the current cgo snap with a different spread task to make it classic, you can use the tooling to set a base and reset the yaml on restore as a template for setting `confinement` to `classic`.

snap/snapcraft.yaml
```diff
@@ -34,6 +34,18 @@ parts:
     stage:
       - snapcraft-completion
 
+  patchelf:
+    plugin: autotools
+    source: https://github.com/NixOS/patchelf
```
sergiusens: we can now use snapcore and it might be wise to do so now.

snapcraft/internal/elf.py
```diff
@@ -376,6 +377,16 @@ def load_dependencies(
                 library_paths.add(l.path)
         return library_paths
 
+    def is_dynamic_executable(self) -> bool:
+        # This way of answering the question is perhaps a bit OTT. But it works.
+        try:
+            ef = elftools.elf.elffile.ELFFile(open(self.path, "rb"))
+        except ELFError as e:
```
sergiusens: `as e` can be better expressed as `as elf_error`

cmatsuoka: Done.

snap/snapcraft.yaml
cmatsuoka: Done.

tests/spread/plugins/go/rpath-rewrite-cgo/task.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+  snapcraft prime
+
+  if ! snap list go; then
+    echo "The go snap from the Snap Store should have been installed."
+    exit 1
+  fi
+
+  # Ensure binaries are properly patched
+  patchelf --print-rpath prime/bin/go-cgo | MATCH "/snap/core[0-9]*/current"
```
sergiusens: make this `/snap/snapcraft/current/bin/patchelf`

sergiusens: and if you really wanted some independence here, you you install binutils and use readelf for this :-)

snapcraft/internal/elf.py
```diff
@@ -376,6 +377,16 @@ def load_dependencies(
...
 
+    def is_dynamic_executable(self) -> bool:
+        # This way of answering the question is perhaps a bit OTT. But it works.
+        try:
+            ef = elftools.elf.elffile.ELFFile(open(self.path, "rb"))
+        except ELFError as elf_error:
+            logger.debug("ELFFile({}) failed: {}".format(self.path, elf_error))
+            return False
+        else:
+            return "PT_DYNAMIC" in [s.header.p_type for s in ef.iter_segments()]
```
sergiusens: just for reference, we also have https://github.com/snapcore/snapcraft/pull/2519/files#diff-f7cf307285f042f35434bbe680606c77R258 and those are all set as properties here https://github.com/snapcore/snapcraft/pull/2519/files#diff-f7cf307285f042f35434bbe680606c77R219 so it might be good to just do it on load, we can leave as is for now if we have a bug for this and a code comment TODO with the LP: # here

snapcraft/plugins/go.py
```diff
@@ -177,6 +178,16 @@ def build(self):
         for package in packages:
             binary = os.path.join(self._gopath_bin, self._binary_name(package))
             self._run(["go", "build", "-o", binary] + tags + [package])
+            # Relink with system linker if executable is dynamic in order to be
+            # able to set rpath later on. This workaround can be removed after
+            # https://github.com/NixOS/patchelf/issues/146 is fixed.
+            if self._is_classic:
+                if elf.ElfFile(path=binary).is_dynamic:
```
sergiusens: is there a reason for double nesting instead of using `and` here?

tests/fixture_setup/_unittests.py
```diff
@@ -128,37 +128,72 @@ def _fake_elffile_extract(self, path):
...
         glibc.add_version("GLIBC_2.2.5")
         glibc.add_version("GLIBC_2.26")
-        return (arch, "/lib64/ld-linux-x86-64.so.2", "", {glibc.name: glibc}, False)
+        return (
+            arch,
+            "/lib64/ld-linux-x86-64.so.2",
+            "",
+            {glibc.name: glibc},
+            False,
+            True,
```
sergiusens: maybe worth, maybe not, but a case for this last part of the Tuple returning False might be good.

snapcraft/plugins/go.py
cmatsuoka: No special reason, will turn into a single conditional.

tests/fixture_setup/_unittests.py
cmatsuoka: There is a test for static binaries where `is_dynamic` is set to False. It's the `fake-elf-static` test in line 163 (harder to see because indentation make it visually smaller than the other test cases).

sergiusens: Great.

PR #2522, opened 2019-04-04
build providers: add API for friendly instance type names

sergiusens:

-----

### DIFF
--- a/snapcraft/cli/lifecycle.py
+++ b/snapcraft/cli/lifecycle.py
@@ -21,8 +21,7 @@ import os
 import click
 
 from . import echo
-from . import env
-from ._options import add_build_options, get_project
+from ._options import add_build_options, get_build_environment, get_project
 from snapcraft.internal import (
     build_providers,
     deprecations,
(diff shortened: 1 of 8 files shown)


snapcraft/cli/lifecycle.py
```diff
@@ -100,7 +100,11 @@ def _execute(  # noqa: C901
             else:
                 raise provider_error
 
-        echo.info("Launching a VM.")
+        echo.info(
+            "Launching a {}.".format(
```
cmatsuoka: I wonder if we should include the action itself in the per-provider supplied string. "Launching a VM" and "Launching a provider" sound correct, but if we have host or remote build providers, "Building on local host" could sound better than "Launching a local build". This could be supplied by e.g. get_instance_type_friendly_action() or something like that. It also could help translations if we get into this path later on.

sergiusens: Sounds like a plan, but let me come up with a better classmethod name :-)

PR #2524, opened 2019-04-05
Snapcraft try

sergiusens:

-----


sergiusens:
Hah, I will have to split out those lxd snapcraft try tests as they fail require snapcraft to come from the store which is unfortunate as we still have no injection for lxd

snapcraft/cli/lifecycle.py
```diff
@@ -272,7 +298,8 @@ def pack(directory, output, **kwargs):
     required=False,
     help="Forces snapcraft to use LXD for this clean command.",
 )
-def clean(parts, use_lxd):
+@click.option("--unprime", is_flag=True, required=False, cls=HiddenOption)
```
cmatsuoka: Just a note, newer click versions added hidden=true so we can use that when we upgrade click.

sergiusens:
Any time we need to deal with state inside the container/VM it becomes a matter of re-execution and knowing where you are standing, so yeah, you are correct, this has not been the easiest :-)

PR #2525, opened 2019-04-08
catkin plugin: check stage-snaps for ROS dependencies

kyrofa:

-----

Similar to building snaps with a separate underlay (i.e. content sharing), the introduction of `stage-snaps` means that ROS dependencies can already be satisfied by the time rosdep runs to determine what dependencies are required. This PR resolves [LP: #1823788](https://bugs.launchpad.net/snapcraft/+bug/1823788) by reusing the functionality that prevents fetching already-satisfied dependencies in an underlay to ensure that dependencies already satisfied by stage-snaps are treated the same way (i.e. not fetched again).


snapcraft/plugins/catkin.py
```diff
@@ -1062,14 +1073,14 @@ class _Catkin:
     def __init__(
```
sergiusens: minor, but can we add some type annotations to __init__?

tests/spread/plugins/catkin/snaps/fake-ros-package-kyrofa/snap/snapcraft.yaml
```diff
@@ -0,0 +1,24 @@
+name: fake-ros-package-kyrofa
```
sergiusens: can you rename this to `test-snapcraft-fake-ros-package-core18`? Or will this work with other bases?

kyrofa: It will not work with other bases given that "melodic" is hard-coded (which snapcraft will only use on core18). I've renamed it.

snapcraft/plugins/catkin.py
kyrofa: Sure, done.

PR #2537, opened 2019-04-23
dotnet plugin: fix parsing of newer sdk releases

ianmetcalf:
- [ ] Have you successfully run `./runtests.sh static`? **Couldn't install black
- [ ] Have you successfully run `./runtests.sh tests/unit`? **Couldn't get tests to finish

-----

In the process of setting up snapcraft for a dotnet core project I ran into an issue when using newer runtime versions. The releases metadata file that is being parsed to find the right sdk file to download has newer releases in a slightly different format. They now include the full URL instead of just the filename. They also changed the checksum file to use upper case instead of lower case letters in the checksum values. You can see this [here](https://raw.githubusercontent.com/dotnet/core/master/release-notes/releases.json) if you compare the releases at the top of the file to the ones lower down.

This updates the plugin to handle both release formats.

------

**I'm not developing on an ubuntu machine so running this test suite proved to be a challenge.

### DIFF
--- a/snapcraft/plugins/dotnet.py
+++ b/snapcraft/plugins/dotnet.py
@@ -32,6 +32,7 @@ The plugin will take into account the following build-attributes:
 import os
 import shutil
 import fnmatch
+import urllib.parse
 import urllib.request
 import json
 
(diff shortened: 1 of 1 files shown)


snapcraft/plugins/dotnet.py
```diff
@@ -211,15 +211,20 @@ def _get_sdk_info(self, version):
...
+            sdk_package_url = metadata[0]["sdk-ubuntu.14.04"]
         else:
             raise DotNetBadReleaseDataError(version)
 
-        sdk_package_url = "{}{}".format(metadata[0]["blob-sdk"], sdk_packge_name)
+        # for older releases prepend the base sdk url
+        if not sdk_package_url.startswith("http"):
+            sdk_package_url = "{}{}".format(metadata[0]["blob-sdk"], sdk_package_url)
+
+        filename = sdk_package_url.rsplit("/", 1).pop()
```
sergiusens: can you change this to use urlparse?
Something like:
```
parsed_url = parse.urlparse(sdk_package_url)
filename = os.path.basename(parse_url.path)
```
?

ianmetcalf: sure, just pushed up that change

PR #2544, opened 2019-04-25
Remove legacy --no-compile option given to pip

tonysimpson:
The python plugin will now byte compile (create *.pyc files)
when installing packages.

Byte compiling python source in the snap will allow it to start.

The use of --no-commit was marked for removal in the original
commit from more than a year ago so its a good time to remove it.

-----

### DIFF
--- a/snapcraft/plugins/_python/_pip.py
+++ b/snapcraft/plugins/_python/_pip.py
@@ -335,9 +335,6 @@ class Pip:
         #
         # --user: Install packages to PYTHONUSERBASE, which we've pointed to
         #         the installdir.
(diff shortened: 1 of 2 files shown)


sergiusens:
This will need some trial, bases are new-ish so I guess folks would cope with issues here, issue being, multiple parts using the python plugin will start to fail (and need to run manual compilation anyways)

cjp256:
@sergiusens This might be a good candidate for pip plugin on core20?

sergiusens:
This is the default working scenario when using core20 with the new plugin infrastructure. Thanks for the proposal

PR #2545, opened 2019-04-30
catkin plugin: use build-packages for compilers

kyrofa:

-----

Currently, the Catkin plugin pulls compilers off to the side rather than requesting that they be installed on the host. The reason for this is historical at this point: before bases existed, one could build a ROS snap on any version of Ubuntu, even versions unsupported by the ROS distribution being used. In an effort to ensure this worked as well as possible, the Catkin plugin did as much as it could without utilizing `build-packages` to ensure all components were compatible.

With the introduction of bases (and the fact that the only ROS1 releases available correspond to LTS releases, and thus bases), this is no longer a concern. It's no longer possible to build a ROS snap on an unsupported Ubuntu version in a supported manner. Add to that the fact that there's a bug in the plugin's use of the compilers that prevents rebuilds from working properly, and it's time to get rid of them.

Rather than pulling compilers off to the side, this PR fixes [LP: #1827148](https://bugs.launchpad.net/snapcraft/+bug/1827148) by simply specifying them as required `build-packages`.

### DIFF
--- a/snapcraft/plugins/catkin.py
+++ b/snapcraft/plugins/catkin.py
@@ -278,7 +278,7 @@ class CatkinPlugin(snapcraft.BasePlugin):
 
         self._rosdistro = _BASE_TO_ROS_RELEASE_MAP[project.info.base]
 
-        self.build_packages.extend(["libc6-dev", "make", "python-pip"])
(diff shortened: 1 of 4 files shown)


kyrofa:
(#2548 is needed before tests will pass)

PR #2548, opened 2019-05-01
integration tests: use correct series in get_package_version

kyrofa:

-----

Currently, `get_package_version` returns the results from the latest Ubuntu release, regardless of the series requested. Until recently, this happened to work out, whether that's because the versions all lined up (doubtful) or because people.canonical.com only recently started using HTTPS.
    
However, now tests are breaking. This is due to a [bug in requests](https://github.com/kennethreitz/requests/issues/5077) that double-encodes URL parameters upon redirect. Simply switching over to HTTPS works around the bug (i.e. no redirect needed) and solves the problem.

### DIFF
--- a/tests/integration/__init__.py
+++ b/tests/integration/__init__.py
@@ -741,7 +741,7 @@ def get_package_version(package_name, series, deb_arch):
         "text": "on",
     }
     query = requests.get(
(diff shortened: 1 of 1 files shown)


kyrofa:
There is, but the bug exists in the most recent release as well.

PR #2550, opened 2019-05-02
storeapi: move from details (v1) to info (v2)

sergiusens:
Instead of accessing each element from the returned calls as dictionary
elements, this implementation offers classes with attributes and some
methods to interact with the API results.

-----
This implementation strays away from the existing implementation pattern of what lives in the storeapi package but I personally think, even if more work initially, will provide easier ways to interact with the results when the Snap Store API is not fresh in your head.

### DIFF
--- a/snapcraft/internal/build_providers/_snap.py
+++ b/snapcraft/internal/build_providers/_snap.py
@@ -42,7 +42,37 @@ class _SnapOp(enum.Enum):
     REFRESH = 3
 
 
-def _get_snap_channel(snap_name: str) -> str:
+_VALID_RISKS = ["stable", "candidate", "beta", "edge"]
+
+
+class _Channel:
+    def __str__(self) -> str:
+        return self._channel
+
+    def __init__(self, channel: str) -> None:
+        channel_parts = channel.split("/")
+        if len(channel_parts) == 1:
+            self.track = None
+            self.risk = channel_parts[0]
+            self.branch = None
+        elif len(channel_parts) == 3:
+            self.track = channel_parts[0]
+            self.risk = channel_parts[1]
+            self.branch = channel_parts[2]
+        elif len(channel_parts) == 2 and channel_parts[0] in _VALID_RISKS:
+            self.track = None
+            self.risk = channel_parts[0]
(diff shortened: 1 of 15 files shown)


snapcraft/storeapi/_snap_index_client.py
```diff
@@ -46,40 +63,37 @@ def get_default_headers(self):
...
-        }
-        if channel is not None:
-            params["channel"] = channel
-        logger.debug("Getting details for {}".format(snap_name))
-        url = "api/v1/snaps/details/{}".format(snap_name)
+
+        params = dict()
+        params[
+            "fields"
+        ] = "channel-map, snap,snap-id,name,publisher,confinement,revision"
```
cprov: There is a stray space before `snap`, but this field doesn't not exist and it's not needed anyway.

snapcraft/storeapi/_snap_index_client.py
```diff
@@ -46,40 +63,37 @@ def get_default_headers(self):
...
 
-        :return Details for the snap.
-        :rtype: dict
+        :return information for the snap.
+        :rtype: SnapInfo
         """
         headers = self.get_default_headers()
         headers.update(
             {
                 "Accept": "application/hal+json",
```
cprov: v2 responses are technically only `application/json`, we can drop the `hal+` part

snapcraft/internal/build_providers/_snap.py
```diff
@@ -144,7 +174,7 @@ def get_assertions(self) -> List[List[str]]:
         if host_snap_info["revision"].startswith("x"):
             return []
 
-        assertions = []  # type: List[List[str]]
+        assertions = list()  # type: List[List[str]]
```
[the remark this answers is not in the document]
sergiusens: pycharms recommends this, it is easier to read I guess

snapcraft/storeapi/_snap_index_client.py
sergiusens: removed, valid fields are not documented though, are they? they also seem to be flattened, correct?

PR #2564, opened 2019-05-15
remote_build: handle git push in detached head state

cmatsuoka:
When dispatching a remote build job using --git and we're currently in
detached head state, branch name retrieval raises an exception. Fix that
by returning HEAD if detached head is detected.
    
-----

### DIFF
--- a/tests/unit/remote_build/test_repo.py
+++ b/tests/unit/remote_build/test_repo.py
@@ -108,6 +108,15 @@ class RepoTestCase(unit.TestCase):
         )
         self.assertThat(str(raised), Contains("is not supported"))
 
+    def test_branch_name_detached_head(self):
+        repo = Repo(self._dir.path)
+        repo.add("foo")
+        repo.commit()
+        commit = self._check_output("git", "rev-parse", "HEAD")
+        commit = commit.decode("ascii").strip()
+        self._run("git", "checkout", commit)
+        self.assertThat(repo.branch_name, Equals("HEAD"))
+
     @mock.patch("git.Repo")
     def test_push_remote(self, mock_repo):
         repo = Repo(self._dir.path)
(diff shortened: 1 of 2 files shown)


sergiusens:
I think you need to rebase first

snapcraft/cli/remote.py
```diff
@@ -159,7 +159,13 @@ def remote_build(
         branch = "master"
 
         # Send local data to the remote repository
+        if not click.confirm(
```
sergiusens: why is this in this PR?

cmatsuoka: It shouldn't, of course. Mixed branches in the hurry just before the presentation. Fixing that.

sergiusens:
[Remote build](https://trello.com/c/2t7tjtLd/2-remote-build)

tests/unit/remote_build/test_repo.py
```diff
@@ -108,6 +108,15 @@ def test_repo_push_remote_error(self):
         )
         self.assertThat(str(raised), Contains("is not supported"))
 
+    def test_branch_name_detached_head(self):
+        repo = Repo(self._dir.path)
+        repo.add("foo")
+        repo.commit()
+        commit = self._check_output("git", "rev-parse", "HEAD")
+        commit = commit.decode("ascii").strip()
```
sergiusens: random question, why decode to ascii?

cmatsuoka: It could have been utf-8, but since it's only a commit hash ascii will do. The decoding itself was used to get a string from bytes.

sergiusens: yeah, the question was, why ascii :-)

cmatsuoka: It could be nothing and use the default, I can change it if you think it's neater.

PR #2569, opened 2019-05-21
lifecycle: warn about bases

sergiusens:
LP: #1774012

-----


snapcraft/project/_sanity_checks.py
```diff
@@ -45,6 +47,25 @@ def conduct_project_sanity_check(project: snapcraft.project.Project):
...
+def _migration_notice(base: Optional[str]) -> None:
+    # We only want to encourage developers to use bases if using the snapcraft snap.
+    if base is None and common.is_snap():
+        logger.warning(
+            "This snapcraft project does not use bases which means this project "
+            "does not take advantage of the latest snapcraft features.\n"
+            "Read more about bases at https://docs.snapcraft.io/t/base-snaps/11198"
+        )
+    # Or if they are on the deb (this legacy codebase) and using bases.
+    elif base is not None:
```
diddledan: Does this match the incorrect "base is set AND we're a snap"?

Consider the first `if`: if `base` is set then it is not `None` and so this is false leading us to the next test which is guaranteed to match in this instance because `base` is definitely not `None`. Does this make better sense?

```suggestion
    elif base is not None and not common.is_snap():
```

snapcraft/project/_sanity_checks.py
```diff
@@ -45,6 +47,25 @@ def conduct_project_sanity_check(project: snapcraft.project.Project):
     if os.path.isdir(snap_dir_path):
         _check_snap_dir(snap_dir_path)
 
+    _migration_notice(project.info.base)
+
+
+def _migration_notice(base: Optional[str]) -> None:
+    # We only want to encourage developers to use bases if using the snapcraft snap.
+    if base is None and common.is_snap():
```
diddledan: Do we want to limit bases to **only** snap-packaged snapcraft? What about snapcraft running on Windows or macOS?

sergiusens: You will install the snap in the VM and find yourself in this use case. I could add a comment if it makes it clearer. But at the end of the day you are going to be running the snap if running snapcraft from a native OSX and future Windows environment.

snapcraft/project/_sanity_checks.py
sergiusens: If this is the snap, the 3.x snap, and this situation happens the world will fall apart! There is overly complex re-execution logic (more than necessary) that make it the case that if running from the snap, you will most likely not fall into this situation.

I guess this suggestion could tie it to the comment above.

snapcraft/project/_sanity_checks.py
```diff
@@ -45,6 +47,27 @@ def conduct_project_sanity_check(project: snapcraft.project.Project):
...
+        logger.warning(
+            "This snapcraft project does not use bases which means this project "
+            "does not take advantage of the latest snapcraft features.\n"
+            "Read more about bases at https://docs.snapcraft.io/t/base-snaps/11198"
+        )
+    # Or if they are on the deb (this legacy codebase) and using bases.
+    elif base is not None and not common.is_snap():
+        logger.warning(
```
evandandrea: Can we document in a comment what broadly are the differences between base support in the deb and in the snap? I can imagine revisiting this part of the code when someone is using bases with the deb, it breaks, and we want to know why.

snapcraft/project/_sanity_checks.py
```diff
@@ -45,6 +47,27 @@ def conduct_project_sanity_check(project: snapcraft.project.Project):
...
+            "This version of snapcraft only has experimental support for bases.\n"
+            "Please move to snapcraft 3.0 or greater for proper base support.\n"
```
evandandrea: "Proper base support" sounds vague. How about: "Please move to snapcraft 3.0 or greater for the fully supported implementation of bases."

snapcraft/project/_sanity_checks.py
sergiusens: Bases in the deb is something only few Canonical folks used. I am fine with expanding the warning, but would rather this never make it to documentation.

sergiusens:
[Bases migration notification and documentation](https://trello.com/c/7aidHsRg/17-bases-migration-notification-and-documentation)

cjp256:
these warnings break tests checking output:
https://github.com/snapcore/snapcraft/blob/legacy/tests/unit/commands/test_snap.py#L431

cjp256:
Took it for a test run and got some conflicting warnings ("not use bases" vs "targets the core base"):
```
~/git/snapcraft-legacy/demos/base-example$ snapcraft 
This snapcraft project does not use bases.
Switch to using bases to take advantage of the latest snapcraft features.
Read more about bases at https://docs.snapcraft.io/t/base-snaps/11198
This project targets the 'core' base but is building on a different host (detected host: Ubuntu 19.04).
```

cjp256:
Can the message detail acceptable base platforms to targets, or link to documentation?

This project targets the 'core18' base but is building on a different host (detected host: Ubuntu 19.04).

I'm assuming based on `_HOST_CODENAME_FOR_BASE = {"core18": "bionic", "core": "xenial"}`,
you can safely build core on Ubuntu Xenial or core18 on Ubuntu Bionic?

snapcraft/internal/common.py
```diff
@@ -134,6 +134,25 @@ def is_docker_instance() -> bool:
...
+    # If not on linux, not worth checking.
+    if platform != "linux":
+        return False
+    # Check the path snapcraft is installed to as a deb.
+    if argv0 != "/usr/bin/snapcraft":
+        return False
+    # Use that path to verify it is part of a deb.
+    try:
+        output = (
+            subprocess.check_call(["dpkg", "-S", "/usr/bin/snapcraft"]).decode().strip()
```
cjp256: I suspect it's always installed to /usr/bin/snapcraft for any deb variant, but if possibly not, you could maybe skip the above path check and jump to:
`subprocess.check_call(["dpkg", "-S", argv0]).decode().strip()`

cjp256: one advantage to this would be to support a relative path (if used), otherwise this check is likely to break on relative pathing?

snapcraft/project/_environment_checks.py
```diff
@@ -0,0 +1,132 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from typing import List, Optional  # noqa: F401
+from typing import Dict
+
+from snapcraft import project
+from snapcraft.internal import common, errors, os_release
+
+_MSG_NO_BASE_FOR_SNAP = (
```
cjp256: If using a snap (new enough to get this check), the no-base check will be invoked from the 3.x version before calling into this legacy mode.  Would it make more sense to just provide this error there?

snapcraft/project/_environment_checks.py
```diff
@@ -0,0 +1,132 @@
...
+        # Any setup run from the debian package and not created from docker
+        # should get the upgrade message.
+        if not common.is_docker_instance() and common.is_deb():
+            self._messages.append(_MSG_UPGRADE)
+
+    def _check_host(self) -> None:
+        base = self._project.info.base
+        # core is not really a base, so it will not be set (legacy).
+        expected_release = self._project.get_expected_release_info_for_base(
+            base if base else "core"
```
cjp256: Would it perhaps be more intuitive to add:
`None: dict(codename="xenial", name="Ubuntu", version_id="16.04"),`
to _HOST_INFO_FOR_BASE?

Then I won't scratch my head wondering about the setting of base when there is no base? :)

snapcraft/internal/common.py
```diff
@@ -134,6 +134,25 @@ def is_docker_instance() -> bool:
     return os.path.exists(_DOCKERENV_FILE)
 
 
+def is_deb(*, argv0: str = sys.argv[0], platform: str = sys.plaform) -> bool:
```
cjp256: s/plaform/platform

cjp256:
Fixed the build issue and ran a few cases locally:
```
snap install snapcraft_2.43.1+git63.g493fc13.dirty_amd64.snap --dangerous --classic
```

building a snap without base set:
```
(snapcraft) ubuntu@snapcraft-legacy-dev:~/snapcraft-legacy$ /snap/bin/snapcraft snap
This snapcraft project does not specify the base keyword, explicitly setting the base keyword enables the latest snapcraft features.
Read more about bases at https://docs.snapcraft.io/t/base-snaps/11198
```

building a snap with base set:
```
(snapcraft) ubuntu@snapcraft-legacy-dev:~/snapcraft-legacy$ /snap/bin/snapcraft snap
<No enviroment messages>
...
```

snapcraft/internal/common.py
sergiusens: argv0 is converted to a full path even if you pass a relative one

snapcraft/project/_environment_checks.py
sergiusens: The checks on non legacy will be much more lighter, we don't print any message there so things should be mostly fine (and when bases become mandatory, we can throw away the old stuff more easily)

snapcraft/internal/common.py
sergiusens: good catch

sergiusens:
@cjp256 the snap test is an interesting one, because we will not be building a snap from this codebase, for a proper test you will need to checkout `master` and in its `snap/snapcraft.yaml` point to this repo/branch for the `snapcraft-legacy` part.

snapcraft/project/_environment_checks.py
sergiusens: done

snapcraft/internal/common.py
cjp256: ran on test:
```
chris@blinkpad:~/git/snapcraft$ cat /usr/bin/python-print-path 
#!/usr/bin/env python3

import sys

print("argv[0]=", sys.argv[0])
chris@blinkpad:~/git/snapcraft$ /usr/bin/python-print-path 
argv[0]= /usr/bin/python-print-path
chris@blinkpad:~/git/snapcraft$ python-print-path 
argv[0]= /usr/bin/python-print-path
chris@blinkpad:~/git/snapcraft$ cd /usr
chris@blinkpad:/usr$ ./bin/python-print-path 
argv[0]= ./bin/python-print-path
chris@blinkpad:/usr$ cd bin/
chris@blinkpad:/usr/bin$ ./python-print-path 
argv[0]= ./python-print-path
```

cjp256:
needs to run `black` formatter to pass static tests

cjp256:
Built and tested the full snap.  

```
(snapcraft) ubuntu@snapcraft-dev:~/snapcraft/tests/spread/plugins/make/snaps/make-hello$ /snap/bin/snapcraft snap
This snapcraft project does not specify the base keyword, explicitly setting the base keyword enables the latest snapcraft features.
Read more about bases at https://docs.snapcraft.io/t/base-snaps/11198
Pulling make-project 
...

Modified os-release to fake 18.04 install:
```
This snapcraft project does not specify the base keyword, explicitly setting the base keyword enables the latest snapcraft features.
This project is best built on 'Ubuntu 16.04', but is building on a 'Ubuntu 18.04' host.
Read more about bases at https://docs.snapcraft.io/t/base-snaps/11198
...

The one thing that really strikes me is that the silent switch to legacy mode seems bad.  The help menu, syntax, etc. changes depending on the directory I'm in (and whether there is a base: core in snap/snapcraft.yaml). There really should be a message informing the user, otherwise I wouldn't get it (if I didn't know what was going on under the hood).

Lastly I didn't have a good way of forcing legacy mode when base is created, so testing those will come next (wish I saved that 2.x snap :))

cjp256:
Tested snap message, and deb message using a 2.x snap and:
```
$ sudo apt install snapcraft
$ sudo mv /usr/bin/snapcraft /usr/bin/snapcraft.real_deb_installed
$ sudo ln -sf /snap/bin/snapcraft /usr/bin/snapcraft
$ /usr/bin/snapcraft snap
This project targets the 'core18' base, best built on 'Ubuntu 18.04', but is building on a 'Ubuntu 16.04' host.
Read more about bases at https://docs.snapcraft.io/t/base-snaps/11198
Skipping pull make-project (already ran)
```

PR #2570, opened 2019-05-22
cli: do not prompt when there is no tty available on stdin

mvo5:
This is a bit of a drive-by. I noticed that inside travis
snapcraft 3 hangs with the following message:
```
Support for 'LXD' needs to be set up. Would you like to do that it now? [y/N]:
```
It seems like we should only prompt here if stdin is attached
to a tty and otherwise just fail early. This should fix hangs
in e.g. scripts.


snapcraft/cli/lifecycle.py
```diff
@@ -85,7 +85,7 @@ def _execute(  # noqa: C901
             build_provider_class.ensure_provider()
         except build_providers.errors.ProviderNotFound as provider_error:
             if provider_error.prompt_installable:
-                if click.confirm(
+                if os.isatty(sys.stdin.fileno()) and click.confirm(
```
zyga: @mvo5 you need to `import sys`

mvo5: Indeed, silly me, I had this in my editor but not saved :( force-pushed.

sergiusens:
Interesting, last I read `click.confirm` was supposed to take this into account. Mind if I take this further and centralize this change considering that we have https://github.com/snapcore/snapcraft/blob/master/snapcraft/internal/indicators.py#L107?

mvo5:
@sergiusens When I played around with redirection it seems like click.confirm() does not take tty into account. But caveats apply, I did not look deeply and did not report an upstream bug (probably a good idea though). Go wild on this, this PR is just a drive-by as I was slightly annoyed by this hang in travis.

PR #2585, opened 2019-06-07
catkin plugin: check workspace for dependencies

kyrofa:

-----

Currently the Catkin plugin doesn't check the workspace for dependencies, which causes the build to fail if a declared dependency isn't contained within the rosdep index (even though it's satisfied by another package in the workspace). Until rosdep [grows the ability to do this on its own](https://github.com/ros-infrastructure/rosdep/issues/685), this PR resolves [LP: #1832044](https://bugs.launchpad.net/snapcraft/+bug/1832044) by adding the ability for the snapcraft CLI to do this manually, using `rospack` to enumerate the packages contained in the workspace and filtering dependencies through that list.

### DIFF
--- a/snapcraft/plugins/_ros/rospack.py
+++ b/snapcraft/plugins/_ros/rospack.py
@@ -1,0 +1,146 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2019 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+import logging
+import subprocess
+import tempfile
+
+from snapcraft.internal import common, repo
+from snapcraft import formatting_utils
+
+logger = logging.getLogger(__name__)
+
+
+class Rospack:
+    def __init__(
+        self,
+        *,
+        ros_distro,
+        ros_package_path,
+        rospack_path,
+        ubuntu_sources,
+        ubuntu_keyrings,
+        project
+    ):
+        self._ros_distro = ros_distro
+        self._ros_package_path = ros_package_path
+        self._rospack_path = rospack_path
+        self._ubuntu_sources = ubuntu_sources
+        self._ubuntu_keyrings = ubuntu_keyrings
+        self._project = project
+
+        self._rospack_install_path = os.path.join(self._rospack_path, "install")
+        self._rospack_cache_path = os.path.join(self._rospack_path, "cache")
+
+    def setup(self):
+        # Make sure we can run multiple times without error
+        os.makedirs(self._rospack_install_path, exist_ok=True)
+
+        # rospack isn't necessarily a dependency of the project, so we'll unpack
+        # it off to the side and use it from there.
+        logger.info("Preparing to fetch rospack...")
+        ubuntu = repo.Ubuntu(
+            self._rospack_path,
+            sources=self._ubuntu_sources,
+            keyrings=self._ubuntu_keyrings,
+            project_options=self._project,
+        )
+
+        logger.info("Fetching rospack...")
+        ubuntu.get(
+            [
+                "ros-{}-rospack".format(self._ros_distro),
+                "ros-{}-catkin".format(self._ros_distro),
+            ]
+        )
+
+        logger.info("Installing rospack...")
+        ubuntu.unpack(self._rospack_install_path)
+
+    def list_names(self):
+        """Obtain list of packages present in the workspace."""
+        output = self._run(["list-names"]).strip()
+        if output:
+            return set(output.split("\n"))
+        else:
+            return set()
+
+    def _run(self, arguments):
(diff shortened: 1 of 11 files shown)


artivis:
:+1:

miguelprada:
+1 to this.

I just stumbled upon this issue when graduating from the simplest possible ROS-snap packaging examples myself.

Out of curiousity, how is it that this hasn't surfaced before? Has there been any change to the internals of the catkin plugin? Is there an alternate way to build N inter(build)dependent packages as a single snap, e.g. building as separate _parts_?

kyrofa:
@miguelprada I'm quite surprised myself. I [specifically accounted for this](https://github.com/snapcore/snapcraft/blob/master/snapcraft/plugins/_ros/rosdep.py#L146) when writing it initially, and only recently learned that my assumptions were completely wrong. I wish rosdep would have errored out instead of letting me continue assuming I did things the right way!

Most of our uses involve the turtlebot, which has its ROS packages in the rosdep index. It masked this issue since dependencies were always resolved, even though it ended up using the index instead of the dependencies contained within the workspace. I didn't hit this until trying to build a workspace with dependencies that were ONLY satisfied by other packages contained in the workspace. That type of scenario is obviously very common, we just honestly fell short on testing it.

snapcraft/plugins/_ros/rospack.py
```diff
@@ -0,0 +1,146 @@
...
+from snapcraft import formatting_utils
+
+logger = logging.getLogger(__name__)
+
+
+class Rospack:
+    def __init__(
+        self,
+        *,
+        ros_distro,
```
sergiusens: can you add type annotations for this and maybe docstrings too?

snapcraft/plugins/_ros/rospack.py
```diff
@@ -0,0 +1,146 @@
...
+        self._rospack_cache_path = os.path.join(self._rospack_path, "cache")
+
+    def setup(self):
+        # Make sure we can run multiple times without error
+        os.makedirs(self._rospack_install_path, exist_ok=True)
+
+        # rospack isn't necessarily a dependency of the project, so we'll unpack
+        # it off to the side and use it from there.
+        logger.info("Preparing to fetch rospack...")
+        ubuntu = repo.Ubuntu(
```
sergiusens: can you call this `ubuntu_repo`?

snapcraft/plugins/_ros/rospack.py
```diff
@@ -0,0 +1,146 @@
...
+                    "export {}".format(
+                        formatting_utils.format_path_variable(
+                            "LD_LIBRARY_PATH", lib_paths, prepend="", separator=":"
+                        )
+                    )
+                )
+
+            # Source our own workspace so we have all of rospack's dependencies
+            lines.append(
+                "_CATKIN_SETUP_DIR={} source {}".format(
```
sergiusens: why are you sourcing this way if you are writing a script?
Aside from the need to source, why not construct the rest so that is passed on with `env`?

cjp256: Could perhaps use Popen with `env` and `shell=True` and drop this script wrapper?
e.g.
`Popen("source ...; do something", shell=True, env=env...)`

snapcraft/plugins/_ros/rospack.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def _run(self, arguments):
+        with tempfile.NamedTemporaryFile(mode="w+") as f:
+            lines = [
+                'export PYTHONPATH="{}"'.format(
+                    os.path.join(
+                        self._rospack_install_path,
+                        "usr",
+                        "lib",
+                        "python2.7",
```
cjp256: Is this always going to be python2.7?

snapcraft/plugins/_ros/rospack.py
kyrofa: Of course, done.

snapcraft/plugins/_ros/rospack.py
kyrofa: Sure, done.

snapcraft/plugins/_ros/rospack.py
kyrofa: Yeah, python2 is EOL at 2.7 (even if it bumped, it won't be SRU'd back to old Ubuntu releases).

snapcraft/plugins/_ros/rospack.py
kyrofa: `_CATKIN_SETUP_DIR` is only a temporary variable that we don't want to set beyond the sourcing of this script. This pattern is already well-established in other wrapper classes.

miguelprada:
Thanks for your response @kyrofa!

> Until fairly recently, the catkin plugin required folks to specify all packages to be built in the workspace (which was silly). Which means the ability to build the entire workspace when no packages are requested is relatively new. When folks specify the package names, we filter dependencies based on that. This only becomes an issue when folks don't specify them, and most users still do because of that historical requirement.

Will try that as a workaround until this gets released :+1: 

> That type of scenario is obviously very common, we just honestly fell short on testing it.

Is there a way to contribute `catkin` plugin-specific tests anywhere? Not making any promises here, but I'm certainly missing more examples and I can try to share if I get around at producing some myself.

kyrofa:
> Will try that as a workaround until this gets released +1

You're welcome to use the `edge` channel as well, anything that gets merged is available there:

    $ sudo snap install --edge --classic snapcraft

> Is there a way to contribute catkin plugin-specific tests anywhere? Not making any promises here, but I'm certainly missing more examples and I can try to share if I get around at producing some myself.

Definitely. There are two sets of tests that will be most interesting to you. Fast, unit-level ones are [here](https://github.com/snapcore/snapcraft/blob/master/tests/unit/plugins/test_catkin.py), integration-level ones are [here](https://github.com/snapcore/snapcraft/tree/master/tests/spread/plugins/catkiny). Contributing either kind would be welcome!

PR #2590, opened 2019-06-10
remote build: add option to skip public upload question

cmatsuoka:
Add an option to prevent user interaction when running on a noninteractive
environment while still acknowledging that uploaded code will be public.

-----


cmatsuoka:
Pushed a commit that adds tty-aware versions of prompt and confirm, and uses it for the prompt parameter of click.option. This makes the confirmation to be displayed earlier, before any other remote build message, so also add an error message explaining what is happening if the check fails.

PR #2597, opened 2019-06-13
ant plugin: switch to using ant snap for building (by default)

cjp256:

-----

### DIFF
--- a/snapcraft/plugins/ant.py
+++ b/snapcraft/plugins/ant.py
@@ -33,11 +33,16 @@ Additionally, this plugin uses the following plugin-specific keywords:
       (list of strings)
       Run the given ant targets.
 
+    - ant-channel:
+      (string)
+      The channel to use for ant in the snap store, if not using tarball from
+      the ant archive (see ant-version and ant-version-checksum).
+      Defaults to latest/edge.
+
     - ant-version:
       (string)
-      The version of ant you want to use to build the source artifacts.
-      Defaults to the current release downloadable from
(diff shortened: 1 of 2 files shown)


cjp256:
@bodewig Hey Stefan, maybe you can take a look at these updates? :)

Tested ant hello world using:
(1) default, unspecified resulting in ant snap from latest/edge
(2) ant-version set to 1.10.5 with checksum
(3) ant-version set to 1.10.6 with wrong checksum (fails as expected)
(4) ant-snap-channel: "1.9/candidate"

cjp256:
(rebased)

snapcraft/plugins/ant.py
```diff
@@ -33,11 +33,16 @@
       (list of strings)
       Run the given ant targets.
 
+    - ant-snap-channel:
```
sergiusens: can we call this `ant-channel`? Just to be consistent with all the other plugins.

cjp256: Good call, fixed.

cjp256:
(rebased with spread fix)

bodewig:
Sorry for being late to the party. This looks good to me, and of course you know how to add a snap dependency while building better than me. :-)

You may want to make the doc for `ant-checksum` more explicit, but then again it already says it is the checksum for `ant-version` which already says it is an alternative to using snaps.

bodewig:
BTW, you could have switched to `candiate` rather than `edge`, but I'll open a PR to make `stable/latest` the default once we've made that official - which I hope to be a question of days.

PR #2598, opened 2019-06-13
crystal plugin: initial version

bcardiff:
Initial plugin for crystal applications.

### DIFF
--- a/tests/spread/plugins/crystal/snaps/crystal-hello/hello/src/hello.cr
+++ b/tests/spread/plugins/crystal/snaps/crystal-hello/hello/src/hello.cr
@@ -1,0 +1,1 @@
+puts "hello world"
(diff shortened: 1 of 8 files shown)


tests/spread/plugins/crystal/snaps/crystal-hello/hello/src/hello.cr
```diff
@@ -0,0 +1 @@
+puts "hello world"
```
cmatsuoka: Could we use a slightly more complex test with a `require`, just to make sure modules are correctly loaded?

bcardiff: The prelude of the std-lib is implicitly required. Otherwise the `puts` usage will not even compile.

Using an external dependency will probably will not be wanted since it will require some external connections, but I could extend the example if wanted. Requiring the sqlite3 connection could be an example (that will also require some addition build_package)

Another alternative is to require something from the std-lib that is outside the prelude (like `Complex`) but that will add no value IMO. Since the prelude is found.

The other thing that can be required is relative files to the project. But I don't think this is something that could break.

Let me know what is preferred.

cmatsuoka: Thanks, if just using `puts` already exercises this, I think it's enough.

PR #2602, opened 2019-06-18
errors: refactor exception/error handling

cjp256:
- Ensure that exception message is _always_ printed.

- Tweak rules of engagement
  - Print exception message (always)

  - Capture trace in temporary file and print path, conditionally if:
    (1) Exception is SnapcraftReportableError or
    (2) Exception is non-Snapcraft error
    (3) debug is true

  - Automatically send bug report if:
    (1) SNAPCRAFT_ENABLE_SILENT_REPORT=y
    (2) "Always" send has been previously set.

  - Prompt to send bug report, conditionally if connected to TTY and:
    (1) Exception is SnapcraftReportableError or
    (2) Exception is non-Snapcraft error

  - Print trace, conditionally if:
    (1) debug is true, or
    (2) Reportable (see above) and either:
      (a) no TTY
      (b) not running as snap

  - Use exit code from snapcraft error (if available), otherwise 1.

- Fixup unit tests to comply with changes.

-----

### DIFF
--- a/snapcraft/cli/_errors.py
+++ b/snapcraft/cli/_errors.py
@@ -30,7 +30,6 @@ from . import echo
 import snapcraft
 from snapcraft.config import CLIConfig as _CLIConfig
 from snapcraft.internal import errors
-from snapcraft.internal.build_providers.errors import ProviderExecError
 
 # raven is not available on 16.04
 try:
@@ -43,7 +42,6 @@ except ImportError:
 # - annotate the part and lifecycle step in the message
 # - add link to privacy policy
 _MSG_TRACEBACK_PRINT = "Sorry, an error occurred in Snapcraft:"
-_MSG_TRACEBACK_FILE = "Sorry, an error occurred in Snapcraft."
 _MSG_TRACEBACK_LOCATION = "You can find the traceback in file {!r}."
 _MSG_SEND_TO_SENTRY_TRACEBACK_PROMPT = dedent(
     """\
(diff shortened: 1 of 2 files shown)


cjp256:
The easiest method to test this is to point to invalid source (e.g. /tmp/does-not-exist.tar.gz).  Specifically this error will be handled properly after this is merged.

Tested using snapcraft with destructive mode (no inner), and as a regular snap (using multipass).  With and without SNAPCRAFT_ENABLE_DEVELOPER_DEBUG=y

cjp256:
(Includes commit from https://github.com/snapcore/snapcraft/pull/2603 to pass static checks)

snapcraft/cli/_errors.py
```diff
@@ -78,6 +76,128 @@
...
+    if not _is_reportable_error(exc_info):
+        return
+
+    # Only attempt if running as snap (with Raven requirement).
+    if not snapcraft.internal.common.is_snap():
+        # Suggest manual reporting instead.
+        click.echo(_MSG_MANUALLY_REPORT)
+        return
+
+    # Only if Raven is installed.
```
sergiusens: this check we can probably remove as we will always have raven installed in the snap.

snapcraft/cli/_errors.py
```diff
@@ -78,6 +76,128 @@
...
+    echo.error(str(exc_info[1]))
+
+
+def _process_exception(exc_info, debug, trace_filepath):
+    _print_exception_message(exc_info, debug)
+
+    if _is_printable_traceback(exc_info, debug):
+        _print_trace_output(exc_info)
+
+    _handle_sentry_submission(exc_info)
```
sergiusens: I was not aware we had a scenario for printing the trace output AND sending to sentry. Should this be an else for the above conditional?

snapcraft/cli/_errors.py
```diff
@@ -78,6 +76,128 @@
...
+            )
+            click.echo(_MSG_MANUALLY_REPORT)
+        else:
+            # Sent successfully.
+            click.echo(_MSG_SEND_TO_SENTRY_THANKS)
+
+
+def _print_exception_message(exc_info, debug) -> None:
+    click.echo(_MSG_TRACEBACK_PRINT)
+    echo.error(str(exc_info[1]))
```
sergiusens: Can you add a code comment of what exc_info[1] holds?

cjp256: Will do

snapcraft/cli/_errors.py
cjp256: I decoupled them with the tweaked ROE. My thought was there may be a case where you do want to report automatically as well as see trace (e.g. SNAPCRAFT_ENABLE_ERROR_REPORTING=y SNAPCRAFT_ENABLE_SILENT_REPORT=y and tty=no).

However, that change in logic does cause a potential issue where tty=no and click.prompt is reached without those variables. So I'll need to fix that regardless.

If you want to maintain the either-or condition, let me know and I'll fix it up.

snapcraft/cli/_errors.py
cjp256: Will do.

snapcraft/cli/_errors.py
cjp256: Fixed.

snapcraft/cli/_errors.py
cjp256: No longer prompt to report if not connected to TTY.  Maintained conditionals as discussed offline.

snapcraft/cli/_errors.py
cjp256: Fixed.

cjp256:
I also fixed the XXX comments.  The any_order issues were due to the use of assertThat in __eq__ for _Tracefile (and copied into _Stdout).  Removed assertions from those methods.

cjp256:
(rebased)

cjp256:
(rebased again with spread fix)

cjp256:
build test failures are for unrelated tests:
- google:ubuntu-16.04-64:tests/spread/plugins/catkin/legacy-pull
- google:ubuntu-18.04-64:tests/spread/general/cwd_not_on_sys_path

https://travis-ci.org/snapcore/snapcraft/jobs/551491295#L1208

snapcraft/cli/_errors.py
```diff
@@ -43,7 +42,6 @@
 # - annotate the part and lifecycle step in the message
 # - add link to privacy policy
 _MSG_TRACEBACK_PRINT = "Sorry, an error occurred in Snapcraft:"
```
sergiusens: it would be great to see a follow-up PR that concats the error message and this on the same line

PR #2612, opened 2019-06-25
tty-aware echo prompt and confirm

cjp256:
Adapted from @cmatsuoka's remote-build work.

-----

### DIFF
--- a/tests/integration/__init__.py
+++ b/tests/integration/__init__.py
@@ -100,6 +100,9 @@ class TestCase(testtools.TestCase):
         # Use a dumb terminal for tests
         self.useFixture(fixtures.EnvironmentVariable("TERM", "dumb"))
 
+        # Force appearance of TTY for pexpect to work with echo.is_connected_tty()
+        self.useFixture(fixtures.EnvironmentVariable("SNAPCRAFT_HAS_TTY", "y"))
+
         # Disable Sentry reporting for tests, otherwise they'll hang waiting
         # for input
         self.useFixture(
--- a/tests/unit/commands/__init__.py
+++ b/tests/unit/commands/__init__.py
@@ -13,6 +13,7 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+import fixtures
 import json
 import os
 from textwrap import dedent
@@ -22,7 +23,6 @@ from click.testing import CliRunner
 from snapcraft import storeapi
 from snapcraft.cli._runner import run
 from tests import fixture_setup, unit
-
 
 _sample_keys = [
     {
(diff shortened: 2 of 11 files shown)


sergiusens:
You need to remove the sys import

sergiusens:
Seems some tests require passing a set of answers to the runner so it gets them correctly

cjp256:
Yes... it due to CommandBaseTestCase.run_command() use of click.testing.CliRunner.invoke(), which feeds the input to click.  It does this by remapping stdio streams, which would cause the side effect of isatty to fail.

Though interestingly enough, I would have expected mock patching isatty() to return True would fix these test cases, but it does not...

cjp256:
Actually the patch probably isn't working precisely because it saves off stdio fds and sets new ones.

cjp256:
Alright, push some changes that should hopefully address the tests.

tests/unit/commands/__init__.py
```diff
@@ -73,6 +73,12 @@ def setUp(self):
         self.runner = CliRunner()
 
     def run_command(self, args, **kwargs):
+        # For click testing, runner will overwrite the descriptors for stdio -
+        # ensure TTY always appears connected.
+        patcher = mock.patch("snapcraft.cli.echo.is_tty_connected", return_value=True)
```
sergiusens: I have been lately preferring the more linguistic friendly use of fixtures,
```python
self.useFixture(fixtures.MockPatchObject("snapcraft.cli.echo.is_tty_connected", return_value=True))
```
If you really want to make the fixture/mock available (although I see it not being used anywhere), you can do:
```python
self.fake_is_tty = fixtures.MockPatchObject("snapcraft.cli.echo.is_tty_connected", return_value=True)
self.useFixture(self.fake_is_tty)
```
And access the mock object using with `.mock`

cjp256: I like that approach - updated to use the fixture.  Thanks!

cjp256:
Pushed updates with fixtures.

cjp256:
This may need some work still.. some store tests failed but I am not yet sure if related.  I restarted the spread tests.

cjp256:
(rebased)

tests/integration/__init__.py
```diff
@@ -448,6 +448,11 @@ def setUp(self):
         self.useFixture(self.test_store)
         self.useFixture(fixtures.EnvironmentVariable("SNAPCRAFT_TEST_INPUT", "1"))
 
+        # Ensure TTY always appears connected.
```
sergiusens: I did not spot this one before, this will not work, you cannot patch in integration tests as they run from the built snap.

cjp256:
Pushed a commit (https://github.com/snapcore/snapcraft/pull/2612/commits/fb1cd4ff7ce1c6063fb654e42521810d703dabb2) that should hopefully address the store failures by setting SNAPCRAFT_HAS_TTY (and honoring it in echo.is_tty_connected().

cjp256:
Build test failures are for unrelated issue (see https://github.com/snapcore/snapcraft/pull/2616#issuecomment-506700424)

PR #2613, opened 2019-06-26
rust plugin: use rust-toolchain by default if present

sergiusens:
If the plugin options are not used and a rust-toolchain file
is found in the source, use that and only default to the
stable toolchain channel as a last resort.

-----

### DIFF
--- a/snapcraft/plugins/rust.py
+++ b/snapcraft/plugins/rust.py
@@ -30,10 +30,16 @@ Additionally, this plugin uses the following plugin-specific keywords:
       select rust channel (stable, beta, nightly)
     - rust-revision
       (string)
-      select rust version
+      select rust version, this is preferred over rust-channel
     - rust-features
       (list of strings)
       Features used to build optional dependencies
+
+If a rust-toolchain file is found, the toolchain specified there will
+be used unless rust-channel or rust-revision are set which will
+override the rust-toolchain file. If neither rust-toolchain exists nor
+rust-channel or rust-revision are set, the latest stable toolchain
+will be used.
 """
 
 import collections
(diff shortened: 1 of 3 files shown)


snapcraft/plugins/rust.py
```diff
@@ -223,13 +235,27 @@ def _build_env(self):
...
 
-    def _get_toolchain(self) -> str:
+    def _get_toolchain(self) -> Optional[str]:
+        """Return the toolchain to use.
+
+        If a rust-toolchain file is present, None will be returned.
+        If a rust-toolchain file is not present and neither rust-version
+        nor rust-channel are set, then stable will be returned.
+        """
+        if self.options.source_subdir:
```
cjp256: NIT - could be simplified a bit:
```
sourcedir = self.sourcedir
    sourcedir = os.path.join(sourcedir, self.options.source_subdir)
```

snapcraft/plugins/rust.py
```diff
@@ -223,13 +235,27 @@ def _build_env(self):
...
+            sourcedir = os.path.join(self.sourcedir, self.options.source_subdir)
+        else:
+            sourcedir = self.sourcedir
+        rust_toolchain_path = os.path.join(sourcedir, "rust-toolchain")
+
         if self.options.rust_revision:
```
cjp256: I wonder if it should error if both revision and channel are specified in config?

snapcraft/plugins/rust.py
cjp256: NIT - could be simplified:

```
sourcedir = self.sourcedir
    sourcedir = os.path.join(sourcedir, self.options.source_subdir)
```

snapcraft/plugins/rust.py
cjp256: I wonder if it should error if both revision and channel are specified in the config?  Otherwise it should probably spell out the precedence in the documentation.

snapcraft/plugins/rust.py
sergiusens: this is a a problem with all plugins, we should fix it across instead of just here

snapcraft/plugins/rust.py
sergiusens: I will do a follow-up for this, it is documented though

sergiusens:
Failed unrelated test:

- google:ubuntu-16.04-64:tests/spread/plugins/catkin/legacy-pull

PR #2625, opened 2019-07-05
tools: let environment-setup.sh skip unnecessary steps

clobrano:
Current environment-setup.sh implementation fails in some points when
the lxc container exists and it is configured - even only partially.

Added some more checks to skip already performed configuration steps.

    No, but the change does not affect source code
    No, but the change does not affect source code

-----


tools/environment-setup.sh
```diff
@@ -14,9 +14,15 @@ if ! grep -q '^name: snapcraft$' snap/snapcraft.yaml; then
 fi
 
 # Create the container
-lxc init ubuntu:16.04 snapcraft-dev
-lxc config set snapcraft-dev raw.idmap "both $UID 1000"
-lxc start snapcraft-dev
+if ! lxc list | grep -q "snapcraft-dev"; then
```
cjp256: suggested improvement to improve accuracy:

`if ! lxc info snapcraft-dev >/dev/null 2>&1`

tools/environment-setup.sh
```diff
@@ -14,9 +14,15 @@ if ! grep -q '^name: snapcraft$' snap/snapcraft.yaml; then
...
+    lxc init ubuntu:16.04 snapcraft-dev
+fi
+if ! lxc config get snapcraft-dev raw.idmap | grep -q "both $UID 1000"; then
+    lxc config set snapcraft-dev raw.idmap "both $UID 1000"
+fi
+if ! lxc list | grep "snapcraft-dev" | grep -q "RUNNING"; then
```
cjp256: suggested improvement to improve accuracy:

`if ! lxc info snapcraft-dev | grep -q "Status: Running"; then`

tools/environment-setup.sh
clobrano: Thanks! I will update the PR

cjp256: Thank you for the updates!

cjp256:
Thanks for the updates! LGTM

PR #2627, opened 2019-07-09
synchronizing snapd supported schema to snapcraft

cjp256:
synchronizing snapd supported schema to snapcraft
    
    - Explicitly specify type string for existing options with unspcecified type:
      - `stop-command`
      - `reload-command`
    - Add daemon type `dbus`
    - Add daemon options:
      - `bus-name` (for use with `daemon: dbus`) with regex pattern found in snapd
      - `restart-delay`
      - `start-timeout`
      - `timer`
      - `watchdog-timeout`
    - Add `on-watchdog` for `restart-condition`
    - Add `autostart` for apps installing autostart desktop files
    - Specify dependencies on daemon for new daemon options, as well
      as existing options (`stop-timeout`, `restart-condition`)
    - Add some tests for newly exposed schema items.
    - Update some descriptions to match https://docs.snapcraft.io/snap-format
    - Add regex patterns for stop timeout (to match introduced timeouts).

### DIFF
--- a/schema/snapcraft.json
+++ b/schema/snapcraft.json
@@ -339,6 +339,9 @@
                         "command"
                     ],
                     "dependencies": {
+                        "bus-name": [
+                            "daemon"
+                        ],
                         "refresh-mode": [
                             "daemon"
                         ],
@@ -348,10 +351,25 @@
                         "stop-command": [
                             "daemon"
                         ],
+                        "start-timeout": [
+                            "daemon"
+                        ],
+                        "stop-timeout": [
+                            "daemon"
+                        ],
+                        "watchdog-timeout": [
+                            "daemon"
+                        ],
+                        "restart-delay": [
+                            "daemon"
+                        ],
                         "post-stop-command": [
                             "daemon"
                         ],
                         "reload-command": [
+                            "daemon"
+                        ],
+                        "restart-condition": [
                             "daemon"
                         ],
                         "before": [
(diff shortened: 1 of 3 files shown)


cjp256:
> Looks pretty good, just spotted a potential issue that can cause headaches if everything goes silent and that is that you should check for the existence of the autostart desktop file in meta and error out appropriately if the file is not found.

Will do.

cjp256:
> > Looks pretty good, just spotted a potential issue that can cause headaches if everything goes silent and that is that you should check for the existence of the autostart desktop file in meta and error out appropriately if the file is not found.
> 
> Will do.

Sorted out that the desktop file doesn't need to exist, it's installable by the application itself.  Snapd just uses the file basename as a lookup to figure out which app to use in launching it.  I added a regex which should match pretty much any sensibly named desktop file.

cjp256:
Pushed other changes (doc updates, new pattern matches).

cjp256:
sergiusens/degville: an overview of the schema regex changes

**autostart**:

- _snapcraft_: ^[A-Za-z0-9. _#:$-]+\\.desktop$
- _snapd_: any string

NOTE: I created the regex to a broad range of possible desktop file names.  It would have 100% coverage for more esoteric names/characters one might find in a path.  We could improve this.

**bus-name**:
- _snapcraft_: ^[A-Za-z0-9/. _#:$-]*$
- _snapd_: "^[A-Za-z0-9/. _#:$-]*$",
  https://github.com/snapcore/snapd/blob/master/snap/validate.go#L610
  https://github.com/snapcore/snapd/blob/master/snap/validate.go#L621
  https://github.com/snapcore/snapd/blob/master/snap/validate.go#L582

NOTE: snapd has another regex within its code base that is probably more appropriate:
- see https://github.com/snapcore/snapd/blob/master/interfaces/core.go#L243

**start-timeout, stop-timeout, watchdog-timeout, restart delay**:
- _snapcraft_: ^[0-9]+(ns|us|ms|s|m)*$
- _snapd_: non-negative, parseable with golang's time.ParseDuration
  https://github.com/snapcore/snapd/blob/master/snap/validate.go#L514
  https://golang.org/src/time/format.go?s=37792:37838#L1271

NOTE: time.ParseDuration will additionally handle <n>µs, <n>μs, and <n>h, but I made a regex to match the documentation I read at https://snapcraft.io/docs/snap-format.

**timer**:
- _snapcraft_: any string
- _snapd_: parseable via timeutil.ParseSchedule
  https://github.com/snapcore/snapd/blob/master/snap/validate.go#L546
  https://github.com/snapcore/snapd/blob/master/timeutil/schedule.go#L522

NOTE: the format is quite complex so I didn't bother with a regex.

PR #2628, opened 2019-07-09
pluginhandler: honour symlink directory paths for filesets (LP: #1833408)

cjp256:
Addresses the usr-merge issues discussed at https://bugs.launchpad.net/snapcraft/+bug/1833408

Adds tests to reproduce issue, fixup/remove faulty tests that are exposed with these changes.

### DIFF
--- a/snapcraft/internal/pluginhandler/__init__.py
+++ b/snapcraft/internal/pluginhandler/__init__.py
@@ -1035,12 +1035,12 @@ def _migratable_filesets(fileset, srcdir):
     include_files = _generate_include_set(srcdir, includes)
     exclude_files, exclude_dirs = _generate_exclude_set(srcdir, excludes)
 
-    # And chop files, including whole trees if any dirs are mentioned
+    # Chop files, including whole trees if any dirs are mentioned.
     snap_files = include_files - exclude_files
     for exclude_dir in exclude_dirs:
         snap_files = set([x for x in snap_files if not x.startswith(exclude_dir + "/")])
 
-    # Separate dirs from files
+    # Separate dirs from files.
     snap_dirs = set(
         [
             x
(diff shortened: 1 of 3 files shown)


cjp256:
Due to a conversation with  @kyrofa, I ended up leaving follow_symlinks option in _migrate_files(), even though seemingly unused by code.  While I'm confident that cleanup can, and perhaps should, happen - it could be done separately in the future.

cjp256:
Also added tests to expose bug (and similar scenarios).  Fixed other faulty tests exposed by this (in some cases, removing redundant faulty tests).

sergiusens:
What was the conversation with @kyrofa about?

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -1029,18 +1029,34 @@ def _expand_part_properties(part_properties, part_schema):
...
+
+    :param str relative_path: Path of target, relative to base_directory.
+
+    :param str base_directory: Base path of target.
+    """
+    parent, filename = os.path.split(relative_path)
+    parent = os.path.realpath(os.path.join(base_directory, parent))
+
+    resolved = os.path.join(parent, filename)
+    resolved = os.path.relpath(resolved, base_directory)
```
sergiusens:
```suggestion
    filename_relpath = os.path.relpath(filename_abspath, base_directory)
```

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -1029,18 +1029,34 @@ def _expand_part_properties(part_properties, part_schema):
     return properties
 
 
+def _resolve_path_parents(relative_path: str, base_directory: str) -> str:
```
sergiusens: How about moving this to fileutils (removing the _) and maybe calling it something closer to the docstring description (e.g.; `get_relative_path`)

I am almost certain we have an implementation for this somewhere in our code too, so this could _eventually_ replace that and its uses

cjp256:
> What was the conversation with @kyrofa about?

The one on IRC where he requested to not break the follow_symlink tests.  The discussion was a bit inconclusive so I just restored the flag for _migrate_files and can open up a PR specifically scoped for that in the future.

snapcraft/internal/pluginhandler/__init__.py
cjp256: Good idea, will do.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -1029,18 +1029,34 @@ def _expand_part_properties(part_properties, part_schema):
...
+    """Resolves symlink path components, returning relative path.
+
+    :param str relative_path: Path of target, relative to base_directory.
+
+    :param str base_directory: Base path of target.
+    """
+    parent, filename = os.path.split(relative_path)
```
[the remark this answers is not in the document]
cjp256: will do

cjp256:
Pushed changes to address comments

snapcraft/internal/pluginhandler/__init__.py
cjp256: I went with get_resolved_relative_path

PR #2629, opened 2019-07-11
RFC: test_pluginhandler: expose potential issue with usr-merge

cjp256:
With the upcoming usr-merge in core20, there may be an issue with existing snapcraft configs using filesets.  Specifically, what to do if a fileset includes say, `bin/*` - that now points to `usr/bin/*`?

Options I can think of:
(1) Require user to update paths (ex: change `bin/*` -> `usr/bin/*`)
(2) Resolve include paths and seamlessly include `usr/bin/*` as well as include any dependent symlinks (ex: `bin -> usr/bin`) so everything just functions.

As a potential bug (?) this PR's intent is to demonstrate the issue and decide on how to address it, if at all.


sergiusens:
As discussed through conversation, this isn't really a problem as the fileset just needs to include what is really wanted.

PR #2632, opened 2019-07-16
pluginhandler, repo: find stage-packages from DT_NEEDED on host

sergiusens:
After priming, DT_NEEDED is checked for in all the files that were primed
and a list of missing files is listed, `stage-packages` are now listed
for the cases where these can be satisfied by adding them.

-----

### DIFF
--- a/snapcraft/internal/pluginhandler/_dependencies.py
+++ b/snapcraft/internal/pluginhandler/_dependencies.py
@@ -1,0 +1,91 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2019 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+from typing import Sequence, Set
+
+from snapcraft.internal import repo
+
+
+_MSG_EXTEND_STAGE_PACKAGES = (
+    "The {part_name!r} part is missing libraries that are not "
+    "included in the snap or base. They can be satisfied by adding the following "
+    "entries to the existing stage-packages for this "
+    "part:\n{stage_packages}"
+)
+_MSG_ADD_STAGE_PACKAGES = (
+    "The {part_name!r} part is missing libraries that are not "
+    "included in the snap or base. They can be satisfied by adding the following "
+    "entry for this part\nstage-packages:\n{stage_packages}"
+)
+_MSG_UNHANDLED_DEPENDENCIES = (
+    "This part is missing libraries that cannot be satisfied with "
+    "any available stage-packages known to snapcraft:\n{unhandled_list}\n"
+    "These dependencies can be satisfied via additional parts or content sharing. "
+    "Consider validating configured filesets if this dependency was built."
+)
+
+
+def _get_formatted_list(items: Set[str]) -> str:
+    return "".join(["- {}\n".format(s) for s in sorted(items)]).strip()
+
+
+class MissingDependencyResolver:
+    """A missing DT_NEEDED resolver."""
+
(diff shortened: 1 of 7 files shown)


snapcraft/internal/repo/_deb.py
```diff
@@ -270,6 +270,24 @@ def get_package_libraries(cls, package_name):
 
         return _library_list[package_name].copy()
 
+    @classmethod
+    def get_package_for_file(cls, file_path: str) -> str:
+        try:
+            output = (
+                subprocess.check_output(
+                    ["dpkg", "-S", file_path], stderr=subprocess.STDOUT
```
cjp256: if there is a case where this may be called without the leading '/' (i.e. relative instead of absolute path), the -S will effectively wildcard the search term potentially yielding undesired results.  may be safest to add leading '/' if not present?

cjp256: Though i see now that's how you invoked it.... maybe move the join(os.sep, file_path) into here?

snapcraft/internal/pluginhandler/_dependencies.py
```diff
@@ -0,0 +1,108 @@
...
+                )
+            )
+
+        unhandled_list = "".join(
+            ["- {}\n".format(s) for s in sorted(self._unhandled)]
+        ).strip()
+        if self._unhandled:
+            logger.warning(
+                "This part is missing libraries that cannot be satisfied with "
+                "any known stage-packages to snapcraft:\n{}\n"
```
cjp256: perhaps: "...any available stage-packages known to snapcraft:"?

snapcraft/internal/repo/_deb.py
sergiusens: Good idea, it is idempotent to do that so better to do it always

sergiusens:
@cjp256 about the merge, the reason for having 2 is that the resolver will stay alive across all of snapcraft and thought I'd store any query done to the host (snaps, stage-packages, what not) in the resolver and avoid doing dpkg -S N times for each part that reports the missing elf file.

What I was thinking of while working on other code though was to just use the lru_cache decorator on dpkg -S and avoid those repeated subprocess calls

cjp256:
> @cjp256 about the merge, the reason for having 2 is that the resolver will stay alive across all of snapcraft and thought I'd store any query done to the host (snaps, stage-packages, what not) in the resolver and avoid doing dpkg -S N times for each part that reports the missing elf file.
> 
> What I was thinking of while working on other code though was to just use the lru_cache decorator on dpkg -S and avoid those repeated subprocess calls

that's a pretty slick decorator there, TIL :)

sergiusens:
How does this look:

```
In [1]: from snapcraft.internal.pluginhandler import MissingDependencyResolver                                                           

In [2]: from snapcraft.cli import echo                                                                                                   

In [3]: d = MissingDependencyResolver(elf_files=["/bin/bash", "/bin/sh", "/bin/foo", "/bin/bar"])                                        

In [4]: d.print_resolutions(part_name="foo", stage_packages_exist=False, echoer=echo)                                                    
The 'foo' part is missing libraries that are not included in the snap or base. They can be satisfied by adding the following entry for this part
stage-packages:
- bash
- dash
This part is missing libraries that cannot be satisfied with any known stage-packages to snapcraft:
- /bin/bar
- /bin/foo
These dependencies can be satisfied via more parts (ensure filesets are correct), or content sharing.

In [5]: d.print_resolutions(part_name="foo", stage_packages_exist=True, echoer=echo)                 
The 'foo' part is missing libraries that are not included in the snap or base. They can be satisfied by adding the following entries to the existing stage-packages for this part:
- bash
- dash
This part is missing libraries that cannot be satisfied with any known stage-packages to snapcraft:
- /bin/bar
- /bin/foo
These dependencies can be satisfied via more parts (ensure filesets are correct), or content sharing.
```
?

snapcraft/internal/pluginhandler/_dependencies.py
```diff
@@ -0,0 +1,92 @@
...
+)
+_MSG_ADD_STAGE_PACKAGES = (
+    "The {part_name!r} part is missing libraries that are not "
+    "included in the snap or base. They can be satisfied by adding the following "
+    "entry for this part\nstage-packages:\n{stage_packages}"
+)
+_MSG_UNHANDLED_DEPENDENCIES = (
+    "This part is missing libraries that cannot be satisfied with "
+    "any available stage-packages known to snapcraft:\n{unhandled_list}\n"
+    "These dependencies can be satisfied via more parts (ensure filesets are correct), "
```
cjp256: NIT: (ensure filesets are correct) ... 

perhaps:
"...via additional parts or content sharing.  Consider validating configured filesets if this dependency built."

PR #2633, opened 2019-07-18
remote-build: increase number of launchpad start_build request attempts

cjp256:
- Add some informative logging for user as well.
- Don't sleep after last attempt.

It seems to be happening a lot, even when the builders are not busy... there might be an underlying reason?


cjp256:
This is actually due to an error on the backend that's not being caught... closing this for now.

PR #2634, opened 2019-07-18
remote build: fully preserve local sources and support packaging all sources

cjp256:


requirements.txt
```diff
@@ -27,3 +27,4 @@ https://files.pythonhosted.org/packages/16/4a/b085a04d6dad79aa5c00c65c9b2bbcb2c6
 lxml==4.3.0; sys_platform == 'win32'
 GitPython==2.1.11
 launchpadlib==1.10.6
+yamlordereddictloader==0.4.0
```
sergiusens: why this and not enhance snapcraft.yaml_utils?

cjp256: Oh that would be much more appropriate :)

snapcraft/internal/remote_build/_launchpad.py
```diff
@@ -246,3 +264,87 @@ def _download_file(self, url: str, name: str) -> None:
...
+        if os.path.exists(self._repo_dir):
+            shutil.rmtree(self._repo_dir)
+
+        os.makedirs(self._repo_dir)
+        self._git_repo = git.Repo.init(self._repo_dir)
+
+    def _create_source_archive(self, source_dir: str, archive_path: str):
+        """Archive sources at source_dir into archive_path."""
+        logger.debug("creating source archive: {}".format(archive_path))
+        command = ["tar", "cJf", archive_path, "-C", source_dir, "."]
```
sergiusens: so source will contain the original snap directory, which is desired, but how will it behave wrt git commands?

cjp256: Yeah, `snapcraft` itself appears to build fine (with its use of git-describe)

cjp256: My first spin of this was to extend the source classes to support mirroring.  It was a lot more optimal for build deltas, but I kept winding up with git-inside-of-git issues.

At first I just did the tarball hack for the obvious .git case, but reconsidered that any `local` sources may have other git-tracking-unfriendly files.  I don't currently know of any data loss issues with tarballing things up in this manner, but it's possible...

cjp256:
Reworked to:
- only package `local` source types by default.  this should eliminate cross-platform issues in the default configuration.
- add `--package-all-sources` to pull and package all source types for the builder.
- implemented source selector support (and will package up all arch variants)
- eliminated source project cache and restored a semblance of worktree, decoupled from launchpad-specifics

cjp256:
Added local plugins support if plugin is not in `/snap` (snapd)
Tested against figlet, multipass, nextcloud, katacontainers, snapcraft, and snapd.

`snapd` is currently erroring out on:
```
If you must store these files within the snap/ directory, move them to snap/local/, which is ignored by snapcraft.
DEPRECATED: The 'build' keyword has been replaced by 'override-build'
See http://snapcraft.io/docs/deprecation-notices/dn8 for more information.
Searching for local plugin for x-builddeb
DEPRECATED: The 'build' keyword has been replaced by 'override-build'
See http://snapcraft.io/docs/deprecation-notices/dn8 for more information.
Skipping pull fontconfig-bionic (already ran)
Skipping pull fontconfig-xenial (already ran)
Skipping pull libc6 (already ran)
Skipping pull liblzma5 (already ran)
Skipping pull snapd (already ran)
Skipping pull squashfs-tools (already ran)
Skipping pull xdelta3 (already ran)
Building fontconfig-bionic 
  File "./build-from-security.py", line 3, in <module>
    import apt
ImportError: No module named 'apt'
Failed to run 'build': Exit code was 1.
Build failed
  File "/usr/lib/python2.7/dist-packages/lpbuildd/target/build_snap.py", line 268, in run
    self.build()
  File "/usr/lib/python2.7/dist-packages/lpbuildd/target/build_snap.py", line 257, in build
    env=env)
  File "/usr/lib/python2.7/dist-packages/lpbuildd/target/build_snap.py", line 104, in run_build_command
    return self.backend.run(args, env=full_env, **kwargs)
  File "/usr/lib/python2.7/dist-packages/lpbuildd/target/lxd.py", line 502, in run
    subprocess.check_call(cmd, **kwargs)
  File "/usr/lib/python2.7/subprocess.py", line 541, in check_call
CalledProcessError: Command '['lxc', 'exec', 'lp-xenial-amd64', '--env', 'LANG=C.UTF-8', '--env', 'SHELL=/bin/sh', '--env', 'SNAPCRAFT_BUILD_INFO=1', '--env', 'SNAPCRAFT_IMAGE_INFO={"build-request-id": "lp-50402220", "build-request-timestamp": "2019-07-22T14:20:38Z", "build_url": "https://launchpad.net/~cjp256/+snap/snapcraft-f1b0caf3e3cb4582857c41e998d09b08/+build/625075"}', '--env', 'SNAPCRAFT_BUILD_ENVIRONMENT=host', '--env', 'http_proxy=http://10.10.10.1:8222/', '--env', 'https_proxy=http://10.10.10.1:8222/', '--env', 'GIT_PROXY_COMMAND=/usr/local/bin/snap-git-proxy', '--', '/bin/sh', '-c', 'cd /build/snapcraft-f1b0caf3e3cb4582857c41e998d09b08 && linux64 snapcraft']' returned non-zero exit status 2
```

I wonder if python-apt should be in build-packages? Kicked off another clean build in an attempt to repro...

On another note, "git: version" ends up mismatching the expected value.

cjp256:
Added support for `version: git` and `version-script`

sergiusens:
Please, let's not support version-script and tell people to use `snapcraftctl set-version` instead (through a DN which we should do if we haven't already).

sergiusens:
Can `add --package-all-sources to pull and package all source types for the builder.` be a separate commit?

snapcraft/internal/remote_build/_worktree.py
```diff
@@ -14,100 +14,263 @@
...
+            source_commit=part.get("source-commit", None),
+        )
+
+    def _printable_part_name(self, part_name: str, selector=None) -> str:
+        if selector:
+            return "{}[{}]".format(part_name, selector)
+        return part_name
+
+    def _get_part_cache_dir(self, part_name: str, selector=None) -> str:
+        """Get source cache directory for part_name."""
```
sergiusens: can you add a docstring for what the selector is?

snapcraft/internal/remote_build/_worktree.py
```diff
@@ -14,100 +14,263 @@
...
+        if "source" not in part_yaml:
+            logger.debug("no source specified, skipping...")
+            return part_yaml
+
+        source = part_yaml["source"]
+        source_modified = False
+        if isinstance(source, list):
+            # Grammar case.
+            part_yaml["source"] = []
+            for s in source:
```
sergiusens: can you add a comment here about needing to take into account advanced grammar and maybe a snippet on how it looks?

sergiusens: Oh you do a bit above...

snapcraft/internal/remote_build/_worktree.py
```diff
@@ -14,100 +14,263 @@
...
+            source,
+            source_dir,
+            source_checksum=part.get("source-checksum", None),
+            source_branch=part.get("source-branch", None),
+            source_tag=part.get("source-tag", None),
+            source_depth=part.get("source-depth", None),
+            source_commit=part.get("source-commit", None),
+        )
+
+    def _printable_part_name(self, part_name: str, selector=None) -> str:
```
sergiusens: this one does not need to be a class member

snapcraft/internal/remote_build/_worktree.py
```diff
@@ -14,100 +14,263 @@
...
+        to snapcraft which parcat t to execute within, or to extract
+        a part to project root instead of its part directory.
+        """
+        if (
+            "version" in self._snapcraft_yaml
+            or "version-script" in self._snapcraft_yaml
+        ):
+            version = self._snapcraft_yaml.get("version", None)
+            version_script = self._snapcraft_yaml.get("version-script", None)
+            version = _version.get_version(version, version_script)
```
sergiusens: version-script can be anything, it should just not be supported in this case

sergiusens:
can you reassess some of the logger.info calls? some seem to be implementation detail user facing messages.

snapcraft/internal/remote_build/_worktree.py
```diff
@@ -14,100 +14,263 @@
...
+            part_yaml["source"] = self._pull_source(part_name, source)
+            if part_yaml["source"] != source:
+                source_modified = True
+
+        if source_modified:
+            # It's now an archive.
+            part_yaml["source-type"] = "tar"
+
+            # Strip source attributes no longer relevant.
+            for key in [
```
sergiusens: we should add a TODO to make this more structured and pick these up from the actual source handler or schema considering that we will be finding ourselves in a potential place where we add a new key and forget to pop it here.

cjp256:
> Please, let's not support version-script and tell people to use `snapcraftctl set-version` instead (through a DN which we should do if we haven't already).

DN?

snapcraft/internal/remote_build/_worktree.py
cjp256: > Can `add --package-all-sources to pull and package all source types for the builder.` be a separate commit?



> Can `add --package-all-sources to pull and package all source types for the builder.` be a separate commit?

Will do.

snapcraft/internal/remote_build/_worktree.py
cjp256: Yeah I'll add some detail for good measure, it is a bit of an unusual construct.

sergiusens:
@cjp256 Deprecation Notice

cjp256:
> Please, let's not support version-script and tell people to use `snapcraftctl set-version` instead (through a DN which we should do if we haven't already).

Added DN for version-script:
https://github.com/snapcore/snapcraft/pull/2639

snapcraft/internal/remote_build/_worktree.py
cjp256: I changed it to instead just pop all keys that start with "source-", and then set the source attributes we dictate here (i.e. source-type).

snapcraft/internal/remote_build/_worktree.py
cjp256: Added error for remote build, opened PR to add DN #2639

cjp256:
@sergiusens Pushed updates which should address your prior review.  Left one comment open unresolved for you to verify.

- Cleaned up user-facing output, add new docstrings
- broke out commit for --package-all-sources
- rebased onto origin/remote-build after 3.7 update
- error out when using `version-script` with UnsupportedVersionScriptError
- error out when not using a supported base.

cjp256:
Added/updated unit tests and addressed some static warnings.

snapcraft/cli/remote.py
```diff
@@ -124,6 +111,9 @@ def remote_build(
 
     project = get_project()
 
+    if project.info.base not in ("core", "core16", "core18"):
```
sergiusens: instead of this check, check for `is None` or we will automatically block dispatching "bare" and "core20" builds when they are ready....

snapcraft/cli/remote.py
```diff
@@ -124,6 +111,9 @@ def remote_build(
...
+        raise errors.InvalidBaseError(base=project.info.base)
```
sergiusens: Maybe "requires" base?
Can you quickly try if the builder errors out appropriately if you try and do something like "base: core20"?

sergiusens: Add a TODO to use Project.is_legacy, I will work on that.

snapcraft/internal/remote_build/_launchpad.py
```diff
@@ -132,6 +134,8 @@ def start_build(self, timeout: int = 5, attempts: int = 5) -> int:
             )
             if builds.entries:
                 break
+
+            # TODO: there may actually be a real build error here!
```
sergiusens: this change seems unnecessary considering #2642

snapcraft/internal/remote_build/_worktree.py
```diff
@@ -14,100 +14,321 @@
...
+    def __init__(self, worktree_dir: str, project: Project) -> None:
+        """Create remote-build WorkTree.
+
+        :param str worktree_dir: Directory to use for working tree.
+        :param Project project: Snapcraft project.
+        """
+        self._project = project
+
+        # Get snapcraft yaml (as OrderedDict).
+        self._snapcraft_config = (
```
sergiusens: why store this if you are saving project?
If it is to avoid a second call, you are doing it right below.

snapcraft/cli/remote.py
cjp256: How about for an error: 
```
class UnsupportedBaseError(RemoteBuildBaseError):

    fmt = (
        "Remote build currently requires that the project uses bases.\n"
        "Please specify an appropriate `base` keyword and try again."
    )
```

snapcraft/internal/remote_build/_launchpad.py
cjp256: Indeed, removed.

snapcraft/cli/remote.py
cjp256: Makes sense, updated.

snapcraft/internal/remote_build/_worktree.py
cjp256: It's really to avoid several calls to it on anything iterating over / working on parts as the `parts` info is not directly available in project (or .info) AFAICT.  I figured it made sense two grab two copies, one original `self._snapcraft_config` and one working copy to freely modify `self._preopared_snapcraft_config`.

I could replace the refs to `self._snapcraft_config` with `project.__raw_snapcraft`?  I was just trying to avoid use of internal properties.

snapcraft/cli/remote.py
cjp256: `base` is really unprocessed with project.info.  So whatever I put into base is considered valid here (e.g. core20, core21) until the builder itself determines it isn't valid (which you'll see with the other PR #2642).

I updated to use get_build_base() for the time being, I think that's most appropriate?

cjp256:
Pushed updates to hopefully address comments.

cjp256:
Fixed launchpad test case.

cjp256:
Build failures for various 16.04 images I expect are unrelated:
```
+ snapcraft stage
  File "/snap/snapcraft/x1/lib/python3.5/site-packages/git/__init__.py", line 83, in <module>
    refresh()
  File "/snap/snapcraft/x1/lib/python3.5/site-packages/git/__init__.py", line 73, in refresh
    if not Git.refresh(path=path):
  File "/snap/snapcraft/x1/lib/python3.5/site-packages/git/cmd.py", line 290, in refresh
    raise ImportError(err)
ImportError: Bad git executable.
The git executable must be specified in one of the following ways:
    - be included in your $PATH
    - be set via $GIT_PYTHON_GIT_EXECUTABLE
    - explicitly set via git.refresh()
All git commands will error until this is rectified.
This initial warning can be silenced or aggravated in the future by setting the
$GIT_PYTHON_REFRESH environment variable. Use one of the following values:
    - quiet|q|silence|s|none|n|0: for no warning or exception
    - warn|w|warning|1: for a printed warning
    - error|e|raise|r|2: for a raised exception
Example:
    export GIT_PYTHON_REFRESH=quiet
  File "/snap/snapcraft/x1/bin/snapcraft", line 11, in <module>
    load_entry_point('snapcraft==3.7', 'console_scripts', 'snapcraft')()
  File "/snap/snapcraft/x1/lib/python3.5/site-packages/pkg_resources/__init__.py", line 489, in load_entry_point
  File "/snap/snapcraft/x1/lib/python3.5/site-packages/pkg_resources/__init__.py", line 2843, in load_entry_point
    return ep.load()
  File "/snap/snapcraft/x1/lib/python3.5/site-packages/pkg_resources/__init__.py", line 2434, in load
    return self.resolve()
  File "/snap/snapcraft/x1/lib/python3.5/site-packages/pkg_resources/__init__.py", line 2440, in resolve
  File "/snap/snapcraft/x1/lib/python3.5/site-packages/snapcraft/cli/__main__.py", line 23, in <module>
    from ._runner import run
  File "/snap/snapcraft/x1/lib/python3.5/site-packages/snapcraft/cli/_runner.py", line 33, in <module>
    from .remote import remotecli
  File "/snap/snapcraft/x1/lib/python3.5/site-packages/snapcraft/cli/remote.py", line 21, in <module>
    from snapcraft.internal.remote_build import WorkTree, LaunchpadClient, InfoFile, errors
  File "/snap/snapcraft/x1/lib/python3.5/site-packages/snapcraft/internal/remote_build/__init__.py", line 17, in <module>
    from ._worktree import WorkTree  # noqa: F401
  File "/snap/snapcraft/x1/lib/python3.5/site-packages/snapcraft/internal/remote_build/_worktree.py", line 18, in <module>
    import git
  File "/snap/snapcraft/x1/lib/python3.5/site-packages/git/__init__.py", line 85, in <module>
    raise ImportError('Failed to initialize: {0}'.format(exc))
ImportError: Failed to initialize: Bad git executable.
The git executable must be specified in one of the following ways:
    - be included in your $PATH
    - be set via $GIT_PYTHON_GIT_EXECUTABLE
    - explicitly set via git.refresh()
All git commands will error until this is rectified.
```

cjp256:
Going to late-import git (as it was before I foolishly lost it).

cjp256:
- Introduced a new error (GitNotFoundVersionError) specifically for git required for `version: git`.
- Late-import git
- Move all repo git operations into launchpad (as it is currently launchpad-specific and simplifies)
- Rename InvalidVersionGitError -> GitNotFoundProviderError (and align error messages)

snapcraft/internal/remote_build/_launchpad.py
```diff
@@ -31,7 +30,9 @@
 from . import errors
 
 import snapcraft
+
```
sergiusens: what is this empty line here?

cjp256:
> You seem to have forgotten to remove git from requirements.txt?

Hah! I was a bit perplexed earlier when I couldn't find "git" in the requirements.txt.  I wondered how that package got installed... now I know.  It's one of _those_ packages that use camelcase...

cjp256:
Pushed fixes for above comments (spacing, GitPython in requirements.txt)

PR #2651, opened 2019-08-01
Extension confinement

sergiusens:

-----

### DIFF
--- a/snapcraft/internal/project_loader/_extensions/_utils.py
+++ b/snapcraft/internal/project_loader/_extensions/_utils.py
@@ -121,17 +121,13 @@ def supported_extension_names() -> List[str]:
 def _load_extension(
     base: str, extension_name: str, yaml_data: Dict[str, Any]
 ) -> Extension:
-    # A base is required in order to use extensions, so raise an error if not specified.
-    if not base:
-        raise errors.ExtensionBaseRequiredError()
-
     extension_class = find_extension(extension_name)
-    if base not in extension_class.get_supported_bases():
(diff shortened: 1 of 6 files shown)


snapcraft/internal/project_loader/_extensions/_utils.py
```diff
@@ -121,13 +121,8 @@ def supported_extension_names() -> List[str]:
...
     base: str, extension_name: str, yaml_data: Dict[str, Any]
 ) -> Extension:
-    # A base is required in order to use extensions, so raise an error if not specified.
-    if not base:
-        raise errors.ExtensionBaseRequiredError()
-
     extension_class = find_extension(extension_name)
-    if base not in extension_class.supported_bases:
-        raise errors.ExtensionUnsupportedBaseError(extension_name, base)
+    extension_class.sanity_check(extension_name=extension_name, yaml_data=yaml_data)
```
kyrofa: Given that this is run literally right before the class is constructed, a case could be made for just doing this check in the base class's `__init__()`. Thoughts? It feels nice to have a class that can only be constructed with compatible YAML as opposed to leaving the compatibility check to the caller.

cmatsuoka: Agree with @kyrofa here, having the check in the constructor instead of calling explicitly looks better.

cjp256: I agree that it makes more sense to move that error check into __init__.  Otherwise LGTM.

Spread test timed out so I restarted it.

sergiusens: this is now done

PR #2653, opened 2019-08-02
scriplets: run override-pull on update_pull

cjp256:
```
    Once a project is re-run, snapcraft loses metadata generated
    from scripts (i.e. snapcraftctl set-version and set-grade).
    Snapcraft itself will reproduce this issue when the sources change:
    
    ```
    Skipping stage bash-completion (already ran)
    Skipping stage snapcraft-libs (already ran)
    Skipping stage legacy-snapcraft (already ran)
    Skipping stage patchelf (already ran)
    Cleaning later steps and re-staging snapcraft ('build' step changed)
    Skipping prime bash-completion (already ran)
    Skipping prime snapcraft-libs (already ran)
    Skipping prime legacy-snapcraft (already ran)
    Skipping prime patchelf (already ran)
    Priming snapcraft
    Compiling pyc files...
    Failed to generate snap metadata: 'adopt-info' refers to part 'snapcraft', but that part is lacking the 'parse-info' property.
    ```
    
    - Since version information may have changed between
    runs, we now re-execute the override-pull scriplets when
    the pull step is re-run.
    
    - Track current step in pluginhandler when invoking a step
    inside the runner - we can use this to determine which
    scriplet metadata to update once called.  The previous logic
    assumed that the steps were cleaned and not being re-run when
    checking for metadata duplication.
    
    - Add test to reproduce failure (and pass it!)
    
    LP #1790651
```

### DIFF
--- a/snapcraft/internal/pluginhandler/__init__.py
+++ b/snapcraft/internal/pluginhandler/__init__.py
@@ -126,6 +126,7 @@ class PluginHandler:
         )
 
         self._migrate_state_file()
+        self._current_step = None  # type: steps.Step
 
     def get_pull_state(self) -> states.PullState:
         if not self._pull_state:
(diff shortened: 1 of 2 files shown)


cjp256:
Update comments and add LP # in commit.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -774,9 +788,13 @@ def clean_stage(self, project_staged_state):
 
         self.mark_cleaned(steps.STAGE)
 
+    def _do_runner_step(self, step: steps.Step):
```
sergiusens: You can also put self.makedirs in here too (pull, stage and prime call that and should be ok for build)

cjp256: will do

PR #2654, opened 2019-08-05
click: update to 7.0

cjp256:
cli: drop custom HiddenOption and use click's hidden flag

Hidden flag now available as first-class option in Click 7.0.

While here, consolidate add_build_options()'s two lists,
_BUILD_OPTION_NAMES and _BUILD_OPTIONS, as they are zipped
together anyhow.

### DIFF
--- a/snapcraft/cli/_options.py
+++ b/snapcraft/cli/_options.py
@@ -95,7 +95,8 @@ def get_build_environment(**kwargs):
 
     if force_destructive_mode and force_use_lxd:
         raise click.BadOptionUsage(
-            "--use-lxd and --destructive-mode cannot be used together."
+            "--destructive-mode",
+            "--use-lxd and --destructive-mode cannot be used together.",
         )
     elif force_use_lxd:
         provider = "lxd"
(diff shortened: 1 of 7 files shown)


sergiusens:
Looks good, but you missed a couple of new messages that came in with the update

cjp256:
Pushed update to fix minor formatting changes for click exceptions.

snapcraft/cli/_options.py
```diff
@@ -48,35 +43,46 @@ def prompt_for_value(self, ctx):
...
     def _add_build_options(func):
-        for name, params in zip(
-            reversed(_BUILD_OPTION_NAMES), reversed(_BUILD_OPTIONS)
-        ):
-            option = click.option(
-                name, **params, cls=HiddenOption if hidden else click.Option
-            )
+        for build_option in reversed(_BUILD_OPTIONS):
+            build_option = build_option.copy()
+            param_decls = build_option.pop("param_decls")
```
sergiusens: any reason to pop this instead of just passing in as kwargs?

cjp256: Because param_decls is a non-keyworded variadic args: https://github.com/pallets/click/blob/master/click/decorators.py#L156

It breaks if I keyword it, I tried :D

sergiusens: :disappointed:

cjp256: @sergiusens pushed a commit to add a code comment, it does deserve one.

PR #2659, opened 2019-08-11
Add support for building inside podman containers

abitrolly:
https://github.com/containers/libpod/blob/master/docs/podman-run.1.md

-----

### DIFF
--- a/snapcraft/internal/common.py
+++ b/snapcraft/internal/common.py
@@ -44,7 +44,7 @@ _DEFAULT_LEGACY_SNAPCRAFT_DIR = os.path.join(sys.prefix, "legacy_snapcraft")
 _legacy_snapcraft_dir = _DEFAULT_LEGACY_SNAPCRAFT_DIR
 
 _DOCKERENV_FILE = "/.dockerenv"
-_PODMAN_FLAG = "/run/.containerenv"
+_PODMAN_FILE = "/run/.containerenv"
 
 MAX_CHARACTERS_WRAP = 120
 
@@ -127,7 +127,7 @@ def is_snap() -> bool:
 
 
 def is_in_container() -> bool:
-    return os.path.exists(_DOCKERENV_FILE) or os.path.exists(_PODMAN_FLAG)
+    return any([os.path.exists(p) for p in (_DOCKERENV_FILE, _PODMAN_FILE)])
 
 
 def set_plugindir(plugindir):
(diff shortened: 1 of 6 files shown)


snapcraft/internal/common.py
```diff
@@ -124,8 +126,8 @@ def is_snap() -> bool:
     return is_snap
 
 
-def is_docker_instance() -> bool:
-    return os.path.exists(_DOCKERENV_FILE)
+def is_in_container() -> bool:
```
sergiusens: can you call this differently, the word container is a bit vague and certainly is not relevant to LXD containers

snapcraft/internal/common.py
```diff
@@ -124,8 +126,8 @@ def is_snap() -> bool:
...
+    return os.path.exists(_DOCKERENV_FILE) or os.path.exists(_PODMAN_FLAG)
```
sergiusens: How about
```suggestion
    return any([os.path.exists(p) for p in (_DOCKERENV_FILE, _PODMAN_FLAG)])
```

snapcraft/internal/common.py
```diff
@@ -42,7 +42,9 @@
 _keyringsdir = _DEFAULT_KEYRINGSDIR
 _DEFAULT_LEGACY_SNAPCRAFT_DIR = os.path.join(sys.prefix, "legacy_snapcraft")
 _legacy_snapcraft_dir = _DEFAULT_LEGACY_SNAPCRAFT_DIR
+
 _DOCKERENV_FILE = "/.dockerenv"
+_PODMAN_FLAG = "/run/.containerenv"
```
sergiusens: Why did you call this FLAG and not FILE?

snapcraft/internal/common.py
abitrolly: Ephemeral LXD containers behave the same. `def container_mode() -> bool` is ok? With explanation of what container mode is.

snapcraft/internal/common.py
abitrolly: Just to remind that it is just an empty file. Fixed.

abitrolly: Renamed to `is_process_container` to be consistent with `is_snap` above and to distinguish between LXD full OS containers.

snapcraft/internal/common.py
abitrolly: Renamed to `is_process_container` to be consistent with `is_snap` above and to distinguish between LXD full OS containers.

abitrolly:
@sergiusens rebased and ready to review.

PR #2665, opened 2019-08-15
Plugin catkin: disable parallel option

artivis:

-----

Catkin plugin : add an option to disable parallel building.

Note: running tests using `spread`, some bits are failing. However the same bits are failing on `master` branch as well.

### DIFF
--- a/snapcraft/plugins/catkin.py
+++ b/snapcraft/plugins/catkin.py
@@ -68,6 +68,10 @@ Additionally, this plugin uses the following plugin-specific keywords:
       (string)
       The URI to ros master setting the env variable ROS_MASTER_URI. Defaults
       to http://localhost:11311.
+    - disable-parallel:
+      (boolean)
+      Disable parallel build, effectively using a single worker.
+      Defaults to False.
 """
 
 import contextlib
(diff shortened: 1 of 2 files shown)


artivis:
@kyrofa FYI

snapcraft/plugins/catkin.py
```diff
@@ -68,6 +68,10 @@
       (string)
       The URI to ros master setting the env variable ROS_MASTER_URI. Defaults
       to http://localhost:11311.
+    - disable-parallel:
```
kyrofa: This functionality actually already exists, just use [this](https://github.com/snapcore/snapcraft/blob/master/snapcraft/_baseplugin.py#L174) to enable it. The [make](https://github.com/snapcore/snapcraft/blob/master/snapcraft/plugins/make.py#L108) plugin is a great example.

artivis: I see. I'll let CI finish the current run then open another PR to replace this.

kyrofa: No need to close the PR, just revamp it a bit. Make sure you add a test that verifies that specifying the option actually does result in `-j1`.

artivis: Well, the scope of the feature slightly change and most of the introduced code goes down the drain. I think opening a new PR is proper.  
Also, any tip on testing ?

PR #2668, opened 2019-08-15
Restore cmake artifacts

cjp256:

-----

### DIFF
--- a/snapcraft/file_utils.py
+++ b/snapcraft/file_utils.py
@@ -414,3 +414,23 @@ def get_resolved_relative_path(relative_path: str, base_directory: str) -> str:
     filename_relpath = os.path.relpath(filename_abspath, base_directory)
 
     return filename_relpath
+
+
+def link_or_copy_files(
+    src_base_dir: str, dst_base_dir: str, file_list: List[str]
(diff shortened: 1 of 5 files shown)


sergiusens:
Are we bringing this back from legacy? I would suggest not and relying on use of override-build or adding the proper targets to the makefiles.

cjp256:
@sergiusens That's certainly an option and I'll defer to your judgement.  I assumed when you handed me the ticket that you wanted me to "fix" it :)

While I think the rationale for adding the install target is fair:
(1) I think the challenge may be if it's an upstream outside of your control... 
(2) The question of consistency between `make` and `cmake` plugin?  Why one and not the other?

Either way, I would suggest we keep the two file_utils.link_or_copy_files commits.  I bet there are other areas in the codebase that could use it as well.

sergiusens:
OK on commit 1 and 2... the adding an install target rationale can apply to all build systems, so having it for a few does not seem to provide a consistent picture, it is quite unfortunate that we failed to remove this from the make plugin before transitioning

cjp256:
Closing, will open new one for file_utils.link_or_copy_files

PR #2670, opened 2019-08-16
Plugin colcon: forward parallel build count

artivis:

-----
Plugin colcon : forward the `parallel-build-count`/`disable-parallel` options down to the build invocation command.


artivis:
@kyrofa FYI

cjp256:
Looks like an issue... 
```
Installing apt dependencies...
Building colcon-part 
Preparing to build colcon packages...
Building colcon packages...
usage: /snapcraft/tests/spread/plugins/colcon/snaps/colcon-talker-listener/parts/colcon-part/install/usr/bin/colcon
       [-h] [--log-base LOG_BASE] [--log-level LOG_LEVEL]
       {build,extension-points,extensions,info,list,metadata,test,test-result,version-check}
       ...
/snapcraft/tests/spread/plugins/colcon/snaps/colcon-talker-listener/parts/colcon-part/install/usr/bin/colcon: error: unrecognized arguments: --parallel-workers 1
colcon build --merge-install --build-base /snapcraft/tests/spread/plugins/colcon/snaps/colcon-talker-listener/parts/colcon-part/build --base-paths /snapcraft/tests/spread/plugins/colcon/snaps/colcon-talker-listener/parts/colcon-part/src/src --install-base /snapcraft/tests/spread/plugins/colcon/snaps/colcon-talker-listener/parts/colcon-part/install/opt/ros/snap --parallel-workers 1 --cmake-args -DCMAKE_BUILD_TYPE=Release
Failed to run 'colcon build --merge-install --build-base /snapcraft/tests/spread/plugins/colcon/snaps/colcon-talker-listener/parts/colcon-part/build --base-paths /snapcraft/tests/spread/plugins/colcon/snaps/colcon-talker-listener/parts/colcon-part/src/src --install-base /snapcraft/tests/spread/plugins/colcon/snaps/colcon-talker-listener/parts/colcon-part/install/opt/ros/snap --parallel-workers 1 --cmake-args -DCMAKE_BUILD_TYPE=Release' for 'colcon-part': Exited with code 2.
```

artivis:
Humm this is annoying as the command runs locally. Maybe is it a colcon version issue. Looking into it.

snapcraft/plugins/colcon.py
```diff
@@ -632,6 +632,9 @@ def _build_colcon_packages(self):
         # Specify that the packages should be installed into the overlay
         colconcmd.extend(["--install-base", self._ros_overlay])
 
+        # Specify the number of workers
+        colconcmd.append("--parallel-workers {}".format(self.parallel_build_count))
```
cjp256: This is combining the flag and the parameter as a single arg.

Simply switch:
`colconcmd.append("--parallel-workers {}".format(self.parallel_build_count))`

with:
`colconcmd.extend(["--parallel-workers", self.parallel_build_count])`

Now it's appending the flag, then appending the count.

cjp256: Alternatively, you could have done:
`colconcmd.append("--parallel-workers={}".format(self.parallel_build_count))`

artivis: Plain curiosity, why is it actually a problem ? The generated command, as printed in CI terminal looks fine.

Any preference on which solution to use ? I'd go with the second one.

cjp256: It's due to argument parsing and the log isn't showing the quotes around the arguments.

It's the equivalent of you running:
`colcon build "--parallel-workers 1"`

The application (colcon) parses the string `--parallel-workers 1` as one argument, not as two.

cjp256: https://github.com/snapcore/snapcraft/blob/master/snapcraft/internal/errors.py#L577 is where it loses the quotes.

If we swapped out:
`command = " ".join(command)`

for:
`command = " ".join([shlex.quote(c) for c in command])`

We'd see the correct output.

sergiusens: We use `shell=False` in our eventual `Popen` calls (default behavior) to avoid injection (https://docs.python.org/2/library/subprocess.html#frequently-used-arguments) so your arguments need to be sliced up.

artivis: Ok I see, thanks !

artivis:
There are a bunch of tests 'running late', resulting in:
`The job exceeded the maximum time limit for jobs, and has been terminated.`

cjp256:
Restarted spread, and wrt to the error:

Created: https://github.com/snapcore/snapcraft/pull/2677/

PR #2674, opened 2019-08-21
 file_utils: introduce link_or_copy_files

cjp256:
And update cmake plugin to use this common code for collecting artifacts.  There are likely other areas in the code base that can use this too.

-----


snapcraft/file_utils.py
```diff
@@ -414,3 +414,23 @@ def get_resolved_relative_path(relative_path: str, base_directory: str) -> str:
     filename_relpath = os.path.relpath(filename_abspath, base_directory)
 
     return filename_relpath
+
+
+def link_or_copy_files(
+    src_base_dir: str, dst_base_dir: str, file_list: List[str]
```
sergiusens: can you make the arguments mandatory by adding a leading *? As I saw the make plugin I noticed I had to come back to this to look at what the 3rd argument was.

How about just using src_dir and dst_dir, they seem pretty standard names already.

sergiusens: What if I want to copy ALL files? Should file_list default to None? That might mean os.walk should be used instead.

PR #2677, opened 2019-08-21
erorrs: preserve quotes when printing SnapcraftPluginCommandError

cjp256:
The exception will not show any effective quotes in the command.
This uses shlex.quote() to print a more accurate description of
the command that failed.

-----

### DIFF
--- a/snapcraft/internal/errors.py
+++ b/snapcraft/internal/errors.py
@@ -14,6 +14,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import shlex
+
 from abc import ABC, abstractmethod
 from snapcraft import formatting_utils
 from snapcraft.internal import steps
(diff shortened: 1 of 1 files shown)


cjp256:
@zyga Did not see any tests that are specific on the command that fails.  I ran `test_errors.py` and `tests/unit/plugins/*.py` locally.  I was surprised myself :)  We'll see if anything else fails I guess, I'm sure something is lurking...

cjp256:
@sergiusens while that may be the case, I matched the approach used in common._run() to execute the command.  I remember you talking about some trouble you faced, could you provide an example?  Should the approach in _run() be changed?

cjp256:
Rebased

> LGTM
> 
> No test needed changing? Should there be one?

It did not appear to affect any tests.

cjp256:
> The quoting here is not the best as you can probably verify from #2686

@sergiusens 
Wasn't that more to do with splitting (posix=False)? Here the command is already split for us?

PR #2678, opened 2019-08-21
cli: introduce --provider

cjp256:
Valid options include `host`, `lxd`, and `multipass`.  This approach
is more extendable than the --use-<provider> flags for future providers.

- Split _BUILD_OPTIONS into _BUILD_OPTIONS and _PROVIDER_OPTIONS
- Split add_build_options into add_build_options and add_provider_options
- Convert `clean` command to use `add_provider_options` so it doesn't
  duplicate provider flags (--use-lxd, --destructive-mode, etc.).

- Minor cleanup for BuilderEnvironmentConfig:
-- drop is_multipass, no-one uses it
-- document lxd as valid provider
-- rename force_provider to build_provider

-----


snapcraft/cli/_options.py
```diff
@@ -65,49 +65,97 @@ def prompt_for_value(self, ctx):
...
     ),
     dict(
         param_decls="--use-lxd",
         is_flag=True,
-        help="Forces snapcraft to use LXD to build.",
+        help="Forces snapcraft to use LXD to build (implies `--provider=lxd`).",
+    ),
+    dict(
+        param_decls="--provider",
+        help="Specify build provider - options include `lxd`, `host`, or `multipass`.",
```
sergiusens: better use the Choice type for this

snapcraft/cli/_options.py
```diff
@@ -65,49 +65,97 @@ def prompt_for_value(self, ctx):
...
             "--destructive-mode",
             "--use-lxd and --destructive-mode cannot be used together.",
         )
-    elif force_use_lxd:
-        provider = "lxd"
-    elif force_destructive_mode:
-        provider = "host"
+
+    if provider:
+        if provider not in ["lxd", "host", "multipass"]:
```
sergiusens: by using the Choice type, this would not be needed

snapcraft/cli/_options.py
```diff
@@ -65,49 +65,97 @@ def prompt_for_value(self, ctx):
...
+                "--provider",
+                "--destructive-mode cannot be used with `--provider={}`".format(
+                    provider
+                ),
+            )
+
+        if provider == "host" and not destructive_mode:
+            raise click.BadOptionUsage(
+                "--provider",
+                "`--provider=host` requires `--destructive-mode` to acknowledge side effects.",
```
sergiusens: good!

snapcraft/cli/_options.py
```diff
@@ -65,49 +65,97 @@ def prompt_for_value(self, ctx):
...
-        provider = None
-    return env.BuilderEnvironmentConfig(force_provider=provider)
+        if use_lxd:
+            provider = "lxd"
+
+        if destructive_mode:
+            provider = "host"
+
+    # BuilderEnvironmentConfig will determine provider if None.
+    return env.BuilderEnvironmentConfig(build_provider=provider)
```
sergiusens: getting rid of BuildEnvironmentConfig might also be fun :-)

cjp256: Done! And with that also managed to consolidate SNAPCRAFT_BUILD_ENVIRONMENT handling in with the --provider option. Much nicer :D

snapcraft/cli/_options.py
cjp256: Thanks for that "DUHH" moment.  Done!

snapcraft/cli/_options.py
cjp256: Thanks for that "DUHH" moment. Done!

cjp256:
Still need to add some tests...

cjp256:
Still a TODO in the env for the tests... don't merge :D

snapcraft/cli/_options.py
```diff
@@ -65,49 +65,112 @@ def prompt_for_value(self, ctx):
...
+                "--use-lxd cannot be used with --provider={}".format(provider)
+            )
+
+        if provider != "host" and destructive_mode:
+            raise click.BadArgumentUsage(
+                "--destructive-mode cannot be used with --provider={}".format(provider)
+            )
+
+        # Specifying --provider=host requires the use of --destructive-mode.
+        # SNAPCRAFT_BUILD_ENVIRONMENT=host does not.
```
sergiusens: maybe it should, our tests would need to adapt and we would need an MP for launchpad to use --destructive-mode

snapcraft/cli/_options.py
```diff
@@ -65,49 +65,112 @@ def prompt_for_value(self, ctx):
...
     ),
     dict(
         param_decls="--use-lxd",
         is_flag=True,
-        help="Forces snapcraft to use LXD to build.",
+        help="Forces snapcraft to use LXD to build (implies `--provider=lxd`).",
+    ),
+    dict(
+        param_decls="--provider",
+        envvar="SNAPCRAFT_BUILD_ENVIRONMENT",
```
sergiusens: If https://github.com/snapcore/snapcraft/pull/2678/files#diff-1b5470a9f2e177fc1606479628098324R123 is done anyways, it probably shouldn't be here... I'd rather keep this envvar as internal, we should probably log.warning when being used (not this PR), but not expose it here either

cjp256: https://click.palletsprojects.com/en/7.x/api/
`show_envvar – controls if an environment variable should be shown on the help page. Normally, environment variables are not shown.`

cjp256:
Rebased

tests/spread/build-providers/lxd-basic/task.yaml
```diff
@@ -2,6 +2,7 @@ summary: Build a basic snap using LXD and ensure that it runs
 
 environment:
   SNAP_DIR: ../snaps/make-hello
+  SNAPCRAFT_BUILD_ENVIRONMENT: lxd
```
sergiusens: why do we need this? --use-lxd should be enough

snapcraft/cli/_options.py
sergiusens: let's use this key

tests/spread/build-providers/lxd-basic/task.yaml
cjp256: IIRC (and by that, I mean if the commit message is accurate lol)

SNAPCRAFT_BUILD_ENVIRONMENT: "host" is set by spread.yaml.  Which is now an error due to mismatch.  The option would be to either unset SNAPCRAFT_BUILD_ENVIRONMENT or setting it to "lxd".

sergiusens: Can we unset it then? It might bring the expectation that this is the way to do things

cjp256: Sure! Looks like that matches what was done for multipass too.

cjp256:
Rebased.

PR #2679, opened 2019-08-21
Colcon

danielwangksu:

-----
plugin colcon: add ability to ignore packages in workspace to colcon plugin

`colcon build` supports `--packages-ignore` to explicitly specify packages
to skip for this build. This PR add the `packages-ignore`ability to snapcraft
colcon plugin

### DIFF
--- a/snapcraft/plugins/colcon.py
+++ b/snapcraft/plugins/colcon.py
@@ -178,6 +178,14 @@ class ColconPlugin(snapcraft.BasePlugin):
             "default": [],
         }
 
+        schema["properties"]["colcon-packages-ignore"] = {
+            "type": "array",
+            "minitems": 1,
+            "uniqueItems": True,
+            "items": {"type": "string"},
+            "default": [],
+        }
+
         schema["required"] = ["source"]
 
         return schema
@@ -196,6 +204,7 @@ class ColconPlugin(snapcraft.BasePlugin):
             "colcon-cmake-args",
             "colcon-catkin-cmake-args",
             "colcon-ament-cmake-args",
+            "colcon-packages-ignore",
         ]
 
     @property
@@ -642,6 +651,11 @@ class ColconPlugin(snapcraft.BasePlugin):
 
         # Finally, add any cmake-args requested from the plugin options
         colconcmd.extend(self.options.colcon_cmake_args)
+
+        if self.options.colcon_packages_ignore:
+            colconcmd.extend(
+                ["--packages-ignore"] + self.options.colcon_packages_ignore
+            )
 
         if self.options.colcon_catkin_cmake_args:
             colconcmd.extend(
--- a/tests/unit/plugins/test_colcon.py
+++ b/tests/unit/plugins/test_colcon.py
@@ -73,6 +73,7 @@ class ColconPluginTestBase(unit.TestCase):
             colcon_catkin_cmake_args = []
             colcon_ament_cmake_args = []
             build_attributes = []
+            colcon_packages_ignore = []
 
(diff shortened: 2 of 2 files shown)


danielwangksu:
@kyrofa FYI

tests/unit/plugins/test_colcon.py
```diff
@@ -1053,6 +1081,14 @@ def __eq__(self, args):
                             ".*--ament-cmake-args.*{}".format(re.escape(expected_args))
                         ),
                     )
+                if colcon_packages_ignore:
+                    expected_args = "".join(colcon_packages_ignore)
```
kyrofa: These are space-separated, are they not?

tests/unit/plugins/test_colcon.py
```diff
@@ -951,6 +970,7 @@ class BuildTestCase(ColconPluginTestBase):
 
     package_scenarios = [
         ("one package", {"colcon_packages": ["my_package"]}),
+        ("two packages", {"colcon_packages": ["my_package", "other_package"]}),
```
kyrofa: Is there a reason you added this scenario?

snapcraft/plugins/colcon.py
```diff
@@ -643,6 +652,11 @@ def _build_colcon_packages(self):
         # Finally, add any cmake-args requested from the plugin options
         colconcmd.extend(self.options.colcon_cmake_args)
 
+        if self.options.colcon_packages_ignore:
```
kyrofa: This is a bit of a nit, but can you move this above the `--cmake-args`? I'm not sure how they're written those consuming types of flags and would rather avoid this option being eaten by that one. It probably makes the most sense to have it directly following `--packages-select`.

snapcraft/plugins/colcon.py
```diff
@@ -178,6 +178,14 @@ def schema(cls):
             "default": [],
         }
 
+        schema["properties"]["colcon-packages-ignore"] = {
```
kyrofa: Please add docs for this new option as well, following the `colcon-packages` option on line 25 of this file.

tests/unit/plugins/test_colcon.py
danielwangksu: yeah you are right I will fix it

tests/unit/plugins/test_colcon.py
danielwangksu: I was trying to test if the `package-ignore` actually works by having multiple packages at first. But it is not really necessary I guess. should I remove it?

snapcraft/plugins/colcon.py
danielwangksu: sure, I will move it

snapcraft/plugins/colcon.py
danielwangksu: OK!

snapcraft/plugins/colcon.py
```diff
@@ -27,6 +27,10 @@
       List of colcon packages to build. If not specified, all packages in the
       workspace will be built. If set to an empty list ([]), no packages will
       be built, which could be useful if you only want ROS debs in the snap.
+    - colcon-packages-ignore:
+      (list of strings)
+      List of colcon packages to ignore. If not specified or set to an empty
+      list ([]), no packages will be ignore.
```
kyrofa:
```suggestion
      list ([]), no packages will be ignored.
```

sergiusens:
Thanks for the PR, mind resolving the conflicts in preparation for a merge?

PR #2692, opened 2019-08-29
internal: unknown apt packages aren't installed

stefanor:
The `is_package_installed()` function was assuming the package was known to APT. There are cases where this is not true, e.g. `lxd` isn't currently available in Debian.

Rather than having to explicitly `check build_package_is_valid()` first, just return False from `is_package_installed()`. It's almost certainly the answer you want.

-----


cjp256:
Makes sense, thank you!

Though the static tests called out an issue:
```
./snapcraft/internal/repo/_deb.py:446:16: E713 test for membership should be 'not in'
```

stefanor:
Urgh, yeah. fixed.

PR #2694, opened 2019-08-30
multipass: fix setup exception when multipass is not found in PATH 

cjp256:
In setup_multipass, there is a case where multipass installs OK, but then
the call to check the version info fails with FileNotFoundError.

The most plausible explanation is that the multipass snap installs OK,
but /snap/bin is not in the PATH, so multipass is not found. This is
easily reproduced.

When this occurs, raise SnapcraftEnvironmentError() with
an error message suggesting the user check their PATH.

Catching this exception was the straw that broke the flake8 "is too
complex" check, so move the wait loop into _wait_for_multipass_ready() -
and flip "multipass" to cls.provider_cmd for consistency.

Fixes SNAPCRAFT-10S

-----

### DIFF
--- a/snapcraft/internal/build_providers/_multipass/_multipass_command.py
+++ b/snapcraft/internal/build_providers/_multipass/_multipass_command.py
@@ -73,6 +73,31 @@ class MultipassCommand:
         )
 
     @classmethod
+    def _wait_for_multipass_ready(cls, *, echoer):
+        echoer.wrapped("Waiting for multipass...")
+        retry_count = 20
+        while retry_count:
+            try:
+                output = subprocess.check_output([cls.provider_cmd, "version"]).decode()
+            except subprocess.CalledProcessError:
+                output = ""
+            except FileNotFoundError:
+                raise SnapcraftEnvironmentError(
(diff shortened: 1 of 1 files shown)


cjp256:
Tested:
```
chris@blinkpad:~/git/snapcraft$ PATH=/home/chris/.local/bin:/home/chris/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games
chris@blinkpad:~/git/snapcraft$ /snap/bin/snapcraft snap
Support for 'multipass' needs to be set up. Would you like to do that it now? [y/N]: y
snapd is not logged in, snap install commands will use sudo
multipass (latest/beta) 0.8.0 from Canonical✓ installed
Waiting for multipass...
Failed to install Multipass: 'multipass not found - please check that it can be found in the configured PATH'
Please install manually. You can find the latest release at:
https://multipass.run
```

cjp256:
I changed the error to SnapcraftEnvironment exception, because multipass actually installed OK, just couldn't be found.

PR #2696, opened 2019-09-02
shell_utils: Add SNAPCRAFT_DEBUG_SCRIPTS env variable

mkg20001:
This allows, if enabled, to see the commands being run inside the script files.

Signed-off-by: Maciej Krüger <mkg20001@gmail.com>

-----


snapcraft/shell_utils.py
```diff
@@ -38,4 +39,8 @@ def run_script(script, **kwargs):
         print("#!/bin/sh", file=tempf)
         print(script, file=tempf)
         tempf.flush()
-        return common.run_output(["/bin/sh", tempf.name], **kwargs)
+        if "SNAPCRAFT_DEBUG_SCRIPTS" in os.environ:
+                return common.run_output(["/bin/bash", "-x", tempf.name], **kwargs)
```
cjp256: any reason for switching from `/bin/sh` to `/bin/bash`?

I wonder if something like this would be appropriate under SNAPCRAFT_ENABLE_DEVELOPER_DEBUG rather than introducing a new variable?

mkg20001: I need to use bash instead of sh, because sh doesn't support the `-x` option AFAIK (I'm wondering which one's better, `-x` (prints every command as executed) or `-v` (prints part of the script being executed as written in source)

`SNAPCRAFT_ENABLE_DEVELOPER_DEBUG` this exists? If so then that might be a better choice

mkg20001: Note that that does print every command that's being executed in a script. This might be too verbose for some cases (since it makes the general debug log less useful), so a separate flag might make sense.
I'll leave that up to the team, but separating it makes sense IMO

mkg20001: Note that that does print every command that's being executed in a script. This might be too verbose for some cases (since it makes the general debug log less useful), so a separate flag might make sense.
I'll leave that up to the team, but separating it makes sense IMO

cjp256: Makes sense, thanks.

`sh -x` should work. You may have unintended consequences by changing the shell, so I'd suggest trying it. Or simply writing the `set -x` at the top of the script.

cjp256: Also, your formatting needs to be corrected (see static test output).

mkg20001: I've turned it into setting -x inline instead of changing the shell
The variable name's fine then?

cjp256: Sorry I dropped the ball on this thread.

@sergiusens Would you agree that it would be fair to enable this for SNAPCRAFT_ENABLE_DEVELOPER_DEBUG?

snapcraft/shell_utils.py
```diff
@@ -36,6 +37,8 @@ def run_script(script, **kwargs):
     """Run the script and return the output."""
     with tempfile.NamedTemporaryFile("w+") as tempf:
         print("#!/bin/sh", file=tempf)
+        if "SNAPCRAFT_DEBUG_SCRIPTS" in os.environ:
```
cjp256, line 40:
```suggestion
        if "SNAPCRAFT_ENABLE_DEVELOPER_DEBUG" in os.environ:
```

snapcraft/shell_utils.py
```diff
@@ -36,6 +37,8 @@ def run_script(script, **kwargs):
     """Run the script and return the output."""
     with tempfile.NamedTemporaryFile("w+") as tempf:
         print("#!/bin/sh", file=tempf)
+        if "SNAPCRAFT_ENABLE_DEVELOPER_DEBUG" in os.environ:
+            return print("set -x", file=tempf)
```
sergiusens, line 41: why return?

sergiusens (changes requested):

snapcraft/shell_utils.py
sergiusens:
```suggestion
            print("set -x", file=tempf)
```
```suggestion
            return print("set -x", file=tempf)
```

sergiusens:
I see that this functionality landed in another PR already, but it is not quite right, so will need to fix considering what was discussed here

PR #2697, opened 2019-09-03
Neon extension

sergiusens:

-----

### DIFF
--- a/snapcraft/internal/project_loader/_extensions/kde_neon.py
+++ b/snapcraft/internal/project_loader/_extensions/kde_neon.py
@@ -1,0 +1,75 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2018-2019 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Import types and tell flake8 to ignore the "unused" List.
+
+from typing import Any, Dict, Tuple
+
+from ._extension import Extension
+
+
+_PLATFORM_SNAP = dict(core18="kde-frameworks-5-core18")
+
+
+class ExtensionImpl(Extension):
+    """The KDE Neon extension.
+
+    This extension makes it easy to assemble KDE based applications
+    using the Neon stack.
+
+    The version of Neon used is dependent on the base used to create
+    the snap.
+    """
+
+    @staticmethod
+    def get_supported_bases() -> Tuple[str, ...]:
+        return ("core18",)
+
+    @staticmethod
+    def get_supported_confinement() -> Tuple[str, ...]:
+        return ("strict", "devmode")
+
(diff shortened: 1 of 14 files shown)


snapcraft/internal/project_loader/_extensions/_utils.py
```diff
@@ -98,12 +98,9 @@ def find_extension(extension_name: str) -> Type[Extension]:
...
 
-    # This may throw an AttributeError, but that would be programmer error of whoever
-    # is hacking on extensions.
-    extension_class_name = "{}Extension".format(
-        extension_name.capitalize().replace("-", "_")
-    )
-    return getattr(extension_module, extension_class_name)
+    # Extensions have an attribute with the name of the extension
+    # class.
+    return getattr(extension_module, "ExtensionImpl")
```
cjp256: Seems a little odd to require them all to share the same name, but it works and simplifies the naming rule.

snapcraft/internal/project_loader/_extensions/kde_neon.py
```diff
@@ -0,0 +1,88 @@
...
+    def __init__(self, *, extension_name: str, yaml_data: Dict[str, Any]) -> None:
+        super().__init__(extension_name=extension_name, yaml_data=yaml_data)
+
+        platform_snap = _PLATFORM_SNAP[yaml_data.get("base")]
+        self.root_snippet = {
+            "plugs": {
+                "gtk-3-themes": {
+                    "interface": "content",
+                    "target": "$SNAP/data-dir/themes",
+                    "default-provider": "gtk-common-themes:gtk-3-themes",
```
cjp256: I assume these are shared between gtk and kde?

sergiusens: I am not totally sold on this, just put it out there, it was a proposal from @kenvandine and part of the original grand scheme.

snapcraft/internal/project_loader/_extensions/_utils.py
cjp256: As a follow up to your point on IRC.  I think the naming scheme is fine.  As implemented, it's already a requirement on the module name, with the added tweak to capitalize the class name.

I kinda liked how the plugin system found it by class type, but that's overkill and has some potential issues of its own.

I do often use class names to navigate source at times, so I'd lean to towards the naming format unless there's something technically wrong with it.  However, as it's largely a personal preference, I defer to your judgement.   :)

sergiusens:
Thanks @apachelogger I will take a look at the locale-gen bits.
We can certainly add usr/sbin and use/games.
And I will make sure I find where vlc's pulse plugins live inside the content snap and add those.

I guess this would benefit GUI users more as is now in almost all cases, we can work on another  neon-cli extension for that specific case, are there apps around that are in need of this?

apachelogger:
FWIW I'm not objecting to landing as-is, merely pointing out the deficiencies.

usr/games is used by some of the kde snaps because they internally call some helper binaries we get from staged debs which use usr/games for some reason. The vlc plugin needs to be found for output to actually work which impacts also some kde games, `kblinken` specifically comes to mind. In both cases @jriddell would know better how widely used these things are right now.
locale-gen is required for a feature that is in just about every KDE application (Help -> Switch application language)

sergiusens:
Thanks @apachelogger, I added locale-gen functionality (missing the one feature which you apparently want so I will work on that now, I wanted to understand it better), vlc plugin paths and the paths for games

cjp256:
The one thought I did have, is the environment info for SNAP_DESKTOP_RUNTIME - is that an interface?  Or would it perhaps be better if that configuration was provided by the runtime?

sergiusens:
on $SNAP_DESKTOP_RUNTIME, I thought about it, and was going for it initially, but that means that most of the common code related to snaps gets split out into all the _frameworks_, this would also require a wrapper and not be able to be offered as a command-chain as it would be an out-of-snap runnable for which we might get away with by using a wrapper and by doing so add one more level of indirection.

PR #2703, opened 2019-09-06
build provider: allow configuration of primary apt mirror

cjp256:
Update cloud-init template to specify the primary mirror.

The template is populated with the (current) default of
`http://archive.ubuntu.com/ubuntu`.

The user may now override this with SNAPCRAFT_APT_PRIMARY_MIRROR.

Only works if used on the command that creates the build VM/conainer,
which executes the cloud-init.

-----

### DIFF
--- a/snapcraft/internal/build_providers/_base_provider.py
+++ b/snapcraft/internal/build_providers/_base_provider.py
@@ -45,6 +45,10 @@ def _get_tzdata(timezone_filepath=os.path.join(os.path.sep, "etc", "timezone")) 
     return timezone
 
 
+def _get_apt_primary_mirror() -> str:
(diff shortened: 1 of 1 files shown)


cjp256:
Due to the complex nature of the solution as-is (most specifically, the environment variable needs to be declared at creation of the container/VM), I don't think it's suitable for documenting as a supported feature (yet).

cjp256:
Another potential option would be to leave out the `apt` section entirely unless the environment variable is found.  I don't know if there are cases where it would cause harm as written with the default...

sergiusens:
we are moving away from cloud-init from snapcraft's PoV, this should be done through the configuration task we discussed.

PR #2704, opened 2019-09-06
extensions: create the gnome-platform directory

sergiusens:
This is the target directory for the gnome "platform" content interface plug.

-----

### DIFF
--- a/extensions/desktop/gnome/Makefile
+++ b/extensions/desktop/gnome/Makefile
@@ -2,9 +2,10 @@
 
 SRC_DIR ?= .
 
-DATA_DIR	:= $(DESTDIR)/data-dir
-BIN_DIR		:= $(DESTDIR)/snap/command-chain
-LIB_DIR		:= $(DESTDIR)/lib
+PLATFORM_DIR := $(DESTDIR)/gnome-platform
+DATA_DIR     := $(DESTDIR)/data-dir
+BIN_DIR      := $(DESTDIR)/snap/command-chain
+LIB_DIR	     := $(DESTDIR)/lib
 DEST_LAUNCHER	:= desktop-launch
 BINDTEXTDOMAIN	:= bindtextdomain.so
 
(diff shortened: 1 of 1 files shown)


extensions/desktop/gnome/Makefile
```diff
@@ -17,14 +17,15 @@ clean:
 
 $(DEST_LAUNCHER):
 	@cat $(SRC_DIR)/init >> $(DEST_LAUNCHER)
-	# tail -n +2 to remove the shebang
+# tail -n +2 to remove the shebang
```
cjp256: lost the tab?

extensions/desktop/gnome/Makefile
```diff
@@ -17,14 +17,15 @@ clean:
...
 	@tail -n +2 $(SRC_DIR)/desktop-exports >> $(DEST_LAUNCHER)
 	@tail -n +2 $(SRC_DIR)/launcher-specific >> $(DEST_LAUNCHER)
 	@tail -n +2 $(SRC_DIR)/mark-and-exec >> $(DEST_LAUNCHER)
 	gcc -Wall -O2 -o $(BINDTEXTDOMAIN) -fPIC -shared $(SRC_DIR)/../src/bindtextdomain.c -ldl
-		
+
 install: $(DEST_LAUNCHER)
 	install -d $(DATA_DIR)
+	install -d $(DATA_DIR)/gnome-platform
```
cjp256: I don't think this goes under data-dir?

extensions/desktop/gnome/Makefile
sergiusens: the tab showed as a syntax error on emacs so I removed it, yes.

extensions/desktop/gnome/Makefile
sergiusens: good catch!

extensions/desktop/gnome/Makefile
cjp256: Now it looks like a syntax error :P I'm not sure the comment is really necessary anyhow.

If these scripts are not intended to be used as-is, perhaps it makes sense to just drop the shebangs from the source?

extensions/desktop/gnome/Makefile
```diff
@@ -17,13 +18,14 @@ clean:
 
 $(DEST_LAUNCHER):
 	@cat $(SRC_DIR)/init >> $(DEST_LAUNCHER)
-	# tail -n +2 to remove the shebang
+# tail -n +2 to remove the shebang
```
zyga: Technically the tab is not a syntax error here. Though this is not the essential part.

sergiusens: Oh, it is not, as it was working, just visual tolerance on my side

PR #2709, opened 2019-09-10
incorporate content provider snaps in dependency resolution

cjp256:
This removes the false-positive warnings associated with use of the gnome extension (and hopefully any users of content snaps).

Tested with gnome-calculator using gnome extension:
https://github.com/cjp256/gnome-calculator/commits/master

Still needs some tests, but open for feedback! :)

-----

### DIFF
--- a/snapcraft/internal/elf.py
+++ b/snapcraft/internal/elf.py
@@ -111,35 +111,23 @@ class Library:
         self,
         *,
         soname: str,
-        path: str,
-        root_path: str,
+        soname_path: str,
+        search_paths: List[str],
         core_base_path: str,
         arch: ElfArchitectureTuple,
         soname_cache: SonameCache
     ) -> None:
+
         self.soname = soname
+        self.soname_path = soname_path
+        self.search_paths = search_paths
+        self.core_base_path = core_base_path
+        self.arch = arch
+        self.soname_cache = soname_cache
 
-        # We need to always look for the soname inside root first,
-        # and after exhausting all options look in core_base_path.
-        if path.startswith(root_path):
-            self.path = path
-        else:
-            self.path = _crawl_for_path(
-                soname=soname,
-                root_path=root_path,
-                core_base_path=core_base_path,
-                arch=arch,
-                soname_cache=soname_cache,
-            )
+        # Resolve path, if possible.
+        self.path = self._crawl_for_path()
 
-        if not self.path and path.startswith(core_base_path):
-            self.path = path
-
-        # Required for libraries on the host and the fetching mechanism
-        if not self.path:
-            self.path = path
-
-        # self.path has the correct resulting path.
         if self.path.startswith(core_base_path):
             self.in_base_snap = True
         else:
@@ -148,44 +136,48 @@ class Library:
         logger.debug(
             "{soname} with original path {original_path} found on {path} in base: {in_base}".format(
                 soname=soname,
-                original_path=path,
+                original_path=soname_path,
                 path=self.path,
                 in_base=self.in_base_snap,
             )
         )
 
+    def _update_soname_cache(self, resolved_path: str) -> None:
+        self.soname_cache[self.arch, self.soname] = resolved_path
 
-def _crawl_for_path(
-    *,
-    soname: str,
-    root_path: str,
-    core_base_path: str,
-    arch: ElfArchitectureTuple,
-    soname_cache: SonameCache
-) -> str:
-    # Speed things up and return what was already found once.
-    if (arch, soname) in soname_cache:
-        return soname_cache[arch, soname]
+    def _is_valid_elf(self, resolved_path: str) -> bool:
+        return (
+            os.path.exists(resolved_path)
+            and ElfFile.is_elf(resolved_path)
+            and ElfFile(path=resolved_path).arch == self.arch
+        )
 
-    logger.debug("Crawling to find soname {!r}".format(soname))
-    for path in (root_path, core_base_path):
-        if not os.path.exists(path):
-            continue
-        for root, directories, files in os.walk(path):
-            for file_name in files:
-                if file_name == soname:
-                    file_path = os.path.join(root, file_name)
-                    if ElfFile.is_elf(file_path):
-                        # We found a match by name, anyway. Let's verify that
-                        # the architecture is the one we want.
-                        elf_file = ElfFile(path=file_path)
-                        if elf_file.arch == arch:
-                            soname_cache[arch, soname] = file_path
-                            return file_path
+    def _crawl_for_path(self) -> str:
+        # Speed things up and return what was already found once.
+        if (self.arch, self.soname) in self.soname_cache:
+            return self.soname_cache[self.arch, self.soname]
 
-    # If not found we cache it too
-    soname_cache[arch, soname] = None
-    return None
+        logger.debug("Crawling to find soname {!r}".format(self.soname))
+
+        if self._is_valid_elf(self.soname_path):
+            self._update_soname_cache(self.soname_path)
+            return self.soname_path
+
+        for path in self.search_paths:
+            if not os.path.exists(path):
+                continue
+            for root, directories, files in os.walk(path):
+                if self.soname not in files:
+                    continue
+
+                file_path = os.path.join(root, self.soname)
+                if self._is_valid_elf(file_path):
+                    self._update_soname_cache(file_path)
+                    return file_path
+
+        # Required for libraries on the host and the fetching mechanism.
+        self._update_soname_cache(self.soname_path)
+        return self.soname_path
 
 
 # Old versions of pyelftools return bytes rather than strings for
@@ -323,7 +315,12 @@ class ElfFile:
         return version_required
 
     def load_dependencies(
-        self, root_path: str, core_base_path: str, soname_cache: SonameCache = None
+        self,
+        root_path: str,
+        core_base_path: str,
+        content_dirs: Set[str],
+        arch_triplet: str,
+        soname_cache: SonameCache = None,
     ) -> Set[str]:
         """Load the set of libraries that are needed to satisfy elf's runtime.
 
@@ -343,11 +340,26 @@ class ElfFile:
 
         logger.debug("Getting dependencies for {!r}".format(self.path))
         ldd_out = []  # type: List[str]
+
+        search_paths = [root_path, *content_dirs, core_base_path]
+
+        ld_library_paths: List[str] = list()
+        for path in search_paths:
+            ld_library_paths.extend(common.get_library_paths(path, arch_triplet))
+
+        # Set environment for ldd with LD_LIBRARY_PATH.
+        env = os.environ.copy()
+        env["LD_LIBRARY_PATH"] = ":".join(ld_library_paths)
+
         try:
             # ldd output sample:
             # /lib64/ld-linux-x86-64.so.2 (0x00007fb3c5298000)
             # libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fb3bef03000)
-            ldd_out = common.run_output(["ldd", self.path]).split("\n")
+            ldd_out = (
+                subprocess.check_output(["ldd", self.path], env=env)
+                .decode()
+                .split("\n")
+            )
         except subprocess.CalledProcessError:
             logger.warning(
                 "Unable to determine library dependencies for {!r}".format(self.path)
@@ -360,8 +372,8 @@ class ElfFile:
                 libs.add(
                     Library(
                         soname=ldd_line[0],
-                        path=ldd_line[2],
-                        root_path=root_path,
+                        soname_path=ldd_line[2],
+                        search_paths=search_paths,
                         core_base_path=core_base_path,
                         arch=self.arch,
                         soname_cache=soname_cache,
--- a/snapcraft/internal/meta/application.py
+++ b/snapcraft/internal/meta/application.py
@@ -15,14 +15,15 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import os
+
 from copy import deepcopy
-from typing import Any, Dict, Optional, Sequence
+from snapcraft import yaml_utils
+from typing import Any, Dict, List, Optional, Sequence  # noqa: F401
 
 from . import errors
 from ._utils import _executable_is_valid
 from .command import Command
 from .desktop import DesktopFile
-from snapcraft import yaml_utils
 
 
 _COMMAND_ENTRIES = ["command", "stop-command"]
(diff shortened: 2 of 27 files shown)


snapcraft/internal/repo/snaps.py
```diff
@@ -94,6 +94,77 @@ def in_store(self):
...
+        if not os.path.exists(yaml_path):
+            return None
+
+        return yaml_utils.load(open(yaml_path, "r"))
+
+    def _translate_installed_path(self, path):
+        if path.startswith("$SNAP"):
+            return path.replace("$SNAP", self._get_installed_path())
+        elif path.startswith("/"):
+            return os.path.join(self._get_installed_path(), path.lstrip("/"))
```
sergiusens: `path[1:]` instead

snapcraft/internal/repo/snaps.py
```diff
@@ -94,6 +94,77 @@ def in_store(self):
                 self._is_in_store = False
         return self._is_in_store
 
+    def _get_installed_path(self):
+        return os.path.join("/snap", self.name, "current")
```
sergiusens: Add a note mentioning that this requires building on a system that lays out snaps into /snap

snapcraft/internal/repo/snaps.py
```diff
@@ -94,6 +94,77 @@ def in_store(self):
...
+    def _translate_installed_dirs(self, paths: Set[str]):
+        translated = set()
+
+        for path in paths:
+            path = self._translate_installed_path(path)
+            translated.add(path)
+
+        return translated
+
+    def get_slot_content_dirs(self, slot_name: str, installed_path=True) -> Set[str]:
```
sergiusens: This feels a bit too specific for this class, maybe this can be get_slots and returns a collection class of slot objects?
Maybe too much work, but let's think about it

snapcraft/internal/project_loader/_plugs_config.py
```diff
@@ -0,0 +1,74 @@
...
+
+import logging
+from typing import Dict, Set  # noqa: F401
+
+from snapcraft.internal.repo.snaps import SnapPackage
+
+logger = logging.getLogger(__name__)
+
+
+class Plug:
```
sergiusens: from the first commit, something like this I guess

snapcraft/internal/repo/snaps.py
cjp256: Are there other known paths?   Or a better way to query this...? I couldn't find anything obvious with a quick search.

cjp256: Another option would be to skip the install, and download the content snap and extract it (with some common code in SnapCache._get_snap_deb_arch).  Or install it as first resort, if we can't find it, fall back to downloading it?

snapcraft/internal/repo/snaps.py
cjp256: I agree.  I was thinking to formalize some classes to handle the marshal/unmarshal meta/snap.yaml.  In the case of the snap the could just request the "SnapMeta" or whatever, then the user can sort out what they want from it.  Could be shared with the various users of snap.yaml (_packer, snapcache, etc.).

snapcraft/internal/project_loader/_plugs_config.py
cjp256: After responding below, now I get this.  And I do believe we are in agreement :)

snapcraft/internal/repo/snaps.py
cjp256: Went with common.get_base_path() for now as it's the same logic, will generalize the name later.

cjp256: Added commit to do rename.

cjp256:
All tests should be passing now *crosses fingers* :D

cjp256:
Added unit tests for newly introduced objects, which rooted out some minor issues the other tests did not catch.

snapcraft/internal/meta/hook.py
```diff
@@ -0,0 +1,53 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from copy import deepcopy
+from typing import Any, Dict, List, Optional, Sequence  # noqa: F401
```
sergiusens: You should not need the noqa anymore by switching to the non-comment inline annotation

snapcraft/internal/meta/hook.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class Hook:
+    """Representation of a generic snap hook."""
+
+    def __init__(self, *, hook_name: str = "") -> None:
+        self.hook_name = hook_name
+        self._hook_properties = dict()  # type: Dict[str, Any]
```
sergiusens: self._hook_properties: Dict[str, Any] = dict()

and so on...

snapcraft/internal/meta/hook.py
```diff
@@ -0,0 +1,53 @@
...
+        self.passthrough = dict()  # type: Dict[str, Any]
+
+    def validate(self):
+        pass
```
sergiusens: Please add a comment on why this is just a `pass`

snapcraft/internal/meta/plugs.py
```diff
@@ -0,0 +1,134 @@
...
+    def __init__(
+        self,
+        *,
+        plug_name: str = "",
+        content: str = "",
+        default_provider: str = "",
+        target: str = ""
+    ) -> None:
+        super().__init__(plug_name=plug_name)
+        self.interface = "content"  # type: str
```
sergiusens: All plugs have an interface, why not pass it to the parent instead?
"content" could also be a enum like we have for build steps

snapcraft/internal/meta/plugs.py
```diff
@@ -0,0 +1,134 @@
...
+        if not self.content:
+            raise PlugValidationError(message="`content` is required for dbus slot")
+
+        if not self.target:
+            raise PlugValidationError(message="`target` is required for dbus slot")
+
+        # Note: default-provider is optional.
+
+    @classmethod
+    def from_dict(cls, plug_dict: Dict[str, Any], plug_name: str = ""):
```
sergiusens: return type missing

snapcraft/internal/meta/plugs.py
```diff
@@ -0,0 +1,134 @@
...
+
+        if "target" in plug_dict:
+            plug.target = plug_dict["target"]
+
+        if "default-provider" in plug_dict:
+            plug.default_provider = plug_dict["default-provider"]
+
+        return plug
+
+    def to_yaml(self):
```
sergiusens: return type missing

snapcraft/internal/meta/plugs.py
```diff
@@ -0,0 +1,134 @@
...
+        target: str = ""
+    ) -> None:
+        super().__init__(plug_name=plug_name)
+        self.interface = "content"  # type: str
+        self._content = content
+        self.default_provider = default_provider
+        self.target = target
+
+    @property
+    def content(self):
```
sergiusens: return type missing

snapcraft/internal/meta/plugs.py
```diff
@@ -0,0 +1,134 @@
...
+import logging
+
+from collections import OrderedDict
+from snapcraft.internal.meta.errors import PlugValidationError
+from typing import Any, Dict, List, Set  # noqa: F401
+
+logger = logging.getLogger(__name__)
+
+
+class Plug:
```
sergiusens: good candidate for abc, abstract base class

snapcraft/internal/meta/hook.py
```diff
@@ -0,0 +1,53 @@
...
+        return hook
+
+    def to_yaml(self) -> Dict[str, Any]:
+        hook_dict = deepcopy(self._hook_properties)
+
+        # Apply passthrough keys.
+        hook_dict.update(self.passthrough)
+        return hook_dict
+
+    def __repr__(self):
```
sergiusens: missing return type

snapcraft/internal/meta/hook.py
```diff
@@ -0,0 +1,53 @@
...
+        return repr(self.to_yaml())
+
+    def __str__(self):
```
sergiusens: missing return type

snapcraft/internal/meta/plugs.py
sergiusens: scratch this, took a second look at the from_dict classmethod

snapcraft/internal/meta/plugs.py
```diff
@@ -0,0 +1,134 @@
...
+        self._plug_dict = dict()  # type: Dict[str, Any]
+        self.plug_name = plug_name
+
+    def validate(self):
+        pass
+
+    @classmethod
+    def from_dict(cls, plug_dict: Dict[str, Any], plug_name: str = ""):
+        # If we explicitly support the type, use it instead.
+        if plug_dict.get("interface") == "content":
```
sergiusens: "content" being an enum or better yet a dictionary of interface names holding class handlers may simplify this.

snapcraft/internal/meta/slots.py
```diff
@@ -0,0 +1,195 @@
...
+            raise SlotValidationError(message="`read` or `write` is required for slot")
+
+    def get_content_dirs(self, installed_path: str) -> Set[str]:
+        content_dirs = set()  # type: Set[str]
+
+        for path in self.read + self.write:
+            # Strip leading "$SNAP" and "/".
+            path = re.sub(r"^\$SNAP", "", path)
+            path = re.sub(r"^/", "", path)
+            content_dirs.add(os.path.join(installed_path, path))
```
sergiusens: just for prettiness, do you want to also strip `./`?

snapcraft/internal/meta/slots.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def __init__(self, *, slot_name: str = "") -> None:
+        self._slot_dict = dict()  # type: Dict[str, Any]
+        self.slot_name = slot_name
+
+    def validate(self):
+        pass
+
+    @classmethod
+    def from_dict(cls, slot_dict: Dict[str, Any], slot_name: str = ""):
```
sergiusens: return type. ... you get the picture :)

snapcraft/internal/meta/slots.py
```diff
@@ -0,0 +1,195 @@
...
+
+        if write is None:
+            self.write = list()  # type: List[str]
+        else:
+            self.write = write
+
+        self.use_source_key = use_source_key
+
+    def validate(self):
+        if self.interface != "content":
```
sergiusens: how is this scenario even possible considering the setting in __init__ above?

snapcraft/internal/meta/slots.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+class DbusSlot(Slot):
+    """Representation of a snap dbus slot."""
+
+    def __init__(self, *, slot_name: str = "", bus: str = "", name: str = "") -> None:
+        super().__init__(slot_name=slot_name)
+
+        self._plug_dict = dict()  # type: Dict[str, Any]
+        self.interface = "dbus"  # type: str
```
sergiusens: I might pass this to parent and make it a read only property there.

snapcraft/internal/meta/snap.py
```diff
@@ -0,0 +1,279 @@
...
+from copy import deepcopy
+from snapcraft.internal import common
+from snapcraft.internal.meta import errors
+from snapcraft.internal.meta.application import Application
+from snapcraft.internal.meta.hook import Hook
+from snapcraft.internal.meta.plugs import ContentPlug, Plug
+from snapcraft.internal.meta.slots import Slot
+
+from snapcraft import yaml_utils
+from typing import Dict, Set, Sequence, List, Any  # noqa: F401
```
sergiusens: use newer annotations and get rid of noqa

snapcraft/project/_project_options.py
```diff
@@ -214,6 +214,7 @@ def __init__(
         logger.debug("Prime dir {}".format(self._prime_dir))
 
         self._set_machine(target_deb_arch)
+        self.snap_meta = Snap()
```
sergiusens: should this not happen in Project? Similar to Info?

snapcraft/internal/meta/application.py
```diff
@@ -32,120 +33,141 @@
...
+
+            app.commands[command_name] = Command(
+                app_name=app_name,
+                command_name=command_name,
+                command=app._app_properties[command_name],
+            )
+
+        return app
+
+    def to_yaml(self) -> Dict[str, Any]:
```
sergiusens: should this not be to_dict?

snapcraft/internal/project_loader/_config.py
```diff
@@ -209,6 +210,15 @@ def __init__(self, project: project.Project) -> None:
         snapcraft_yaml = self._expand_filesets(snapcraft_yaml)
 
         self.data = self._expand_env(snapcraft_yaml)
+
+        self.data["architectures"] = _process_architectures(
+            self.data.get("architectures"), project.deb_arch
+        )
+
+        # TODO: convergence of project_loader and config.
+        snap_meta = Snap.from_dict(self.data)
```
sergiusens: This should probably be the holder of snap_meta, we should work towards this goal.
Can you make the on in Project private?

sergiusens: Let's discuss this more

tests/fixture_setup/_fixtures.py
```diff
@@ -693,6 +693,14 @@ def _setUp(self):
         mock_core_path.return_value = self.core_path
         self.addCleanup(patcher.stop)
 
+        self.content_dirs = set([])
+        patcher = mock.patch(
```
sergiusens: maybe start using `fixtures.MockPatch` here

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -24,12 +24,13 @@
...
 import sys
 from glob import glob, iglob
-from typing import cast, Dict, List, Set, Sequence
+from typing import Any, cast, Dict, List, Set, Sequence  # noqa: F401
 
 import snapcraft.extractors
 from snapcraft import file_utils, yaml_utils
 from snapcraft.internal import common, elf, errors, repo, sources, states, steps
 from snapcraft.internal.mangling import clear_execstack
+from snapcraft.project import Project
```
sergiusens: only import this if TYPE_CHECKING

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -53,7 +54,7 @@ def __init__(
         *,
         plugin,
         part_properties,
-        project_options,
+        project_options: Project,
```
sergiusens: quote this so it is only used during TYPE_CHECKING

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -63,11 +64,11 @@ def __init__(
         confinement,
         snap_type,
         soname_cache
-    ):
+    ) -> None:
         self.valid = False
         self.plugin = plugin
         self._part_properties = _expand_part_properties(part_properties, part_schema)
-        self.stage_packages = []
+        self.stage_packages = list()  # type: List[str]
```
sergiusens: use type annotations instead.

snapcraft/internal/elf.py
```diff
@@ -114,6 +114,7 @@ def __init__(
         path: str,
         root_path: str,
         core_base_path: str,
+        content_dirs: Set[str],
```
sergiusens: core_base_path and content_dirs could be generalized into one, can it not? It would need to stop being a set for content_dirs to be preferred over core_base_path.

sergiusens: When doing this add a docstring explaining the importance of the order, then you do not need to do it in the three other places it is used.

You can make this type a Tuple instead of a set or list to not allow mutations.

I have not looked into even adding "root_path" to this, if it has more meanings than just a search path like the others, then keep it separate.

sergiusens:
Prepend a _ to all attributes and methods added under the project module as we might move these

snapcraft/internal/meta/slots.py
cjp256: x = ContentSlot(..)
x.interface = "foo"

I can prevent that with a read-only property.

snapcraft/internal/meta/slots.py
cjp256: As in, the input path may list `./` like `./bin/...` ?

snapcraft/internal/meta/slots.py
cjp256: agreed

snapcraft/internal/pluginhandler/__init__.py
cjp256: what's the rationale for that? that goes back to that other thread I forget which...

cjp256: (I assume it was a dependency issue, but I didn't see any resulting problems?)

tests/fixture_setup/_fixtures.py
cjp256: will do.

snapcraft/internal/project_loader/_config.py
cjp256: Made private, and consolidated to the instance in Project() being overriden by Config.

cjp256:
> Prepend a _ to all attributes and methods added under the project module as we might move these

Done

snapcraft/internal/elf.py
cjp256: Done.

snapcraft/internal/meta/plugs.py
cjp256: I made a dict of interfaces (SLOT_MAPPINGS and PLUG_MAPPINGS), that good?

snapcraft/internal/meta/slots.py
cjp256: Added ./ as well, but do correct me if I'm mistaking the intent!

snapcraft/internal/meta/slots.py
cjp256: Done

snapcraft/internal/pluginhandler/__init__.py
cjp256: Done.

cjp256:
With gnome-calculator @ github.com/cjp256/gnome-calculator:
```
Priming gnome-calculator
Skipping prime gnome-extension (already ran)
Determining the version from the project repo (version: git).
The version has been set to '3.33.3+git9.fcfaf675'
Snapping 'gnome-calculator' \
Snapped gnome-calculator_3.33.3+git9.fcfaf675_amd64.snap
```

vs

```
Priming gnome-calculator
The 'gnome-calculator' part is missing libraries that are not included in the snap or base. They can be satisfied by adding the following entries to the existing stage-packages for this part:
- libatk-bridge2.0-0
- libatk1.0-0
- libatspi2.0-0
- libcairo-gobject2
- libcairo2
- libdatrie1
- libepoxy0
- libfontconfig1
- libfreetype6
- libgdk-pixbuf2.0-0
- libgee-0.8-2
- libgraphite2-3
- libgtk-3-0
- libharfbuzz0b
- libicu60
- libmpc3
- libmpfr6
- libpango-1.0-0
- libpangocairo-1.0-0
- libpangoft2-1.0-0
- libpixman-1-0
- libpng16-16
- libsoup2.4-1
- libthai0
- libwayland-client0
- libwayland-cursor0
- libwayland-egl1
- libx11-6
- libxau6
- libxcb-render0
- libxcb-shm0
- libxcb1
- libxcomposite1
- libxcursor1
- libxdamage1
- libxdmcp6
- libxext6
- libxfixes3
- libxi6
- libxinerama1
- libxkbcommon0
- libxml2
- libxrandr2
- libxrender1
Priming gnome-extension
Determining the version from the project repo (version: git).
The version has been set to '3.33.3+git9.fcfaf675'
Snapping 'gnome-calculator' |
Snapped gnome-calculator_3.33.3+git9.fcfaf675_amd64.snap
```

Should run faster too (if dependent snaps are installed)!

cjp256:
Even if not in this PR, we probably should add a couple of spread tests.  One with missing libs, one with satisfied deps in content snap.

cjp256:
On my machine, this patch series cut the time to snap gnome-calculator from ~3m50s to ~1m.

cjp256:
For clarification, I went back and ran some fresh tests:

[pip installed]
master: 3m 21s
with PR: 0m 59s

[snap installed]
master: 3m 35s
with PR: 1m 14s

So the PR does make a significant difference, and the snap is introducing ~14s of overhead on the build when compared to pip installed.

sergiusens:
@cjp256 and the contents of the snap?

cjp256:
```
diff --git a/snap/gnome-calculator/x1/meta/snap.yaml b/snap/gnome-calculator/x2/meta/snap.yaml
index 43675ba..ffe72e6 100644
--- a/snap/gnome-calculator/x1/meta/snap.yaml
+++ b/snap/gnome-calculator/x2/meta/snap.yaml
@@ -19,9 +19,26 @@ description: |
   Programming mode supports conversion between common bases (binary, octal,
   decimal, and hexadecimal), boolean algebra, one’s and two’s complementation,
   character to character code conversion, and more.
-base: core18
+apps:
+  gnome-calculator:
+    command: usr/bin/gnome-calculator
+    command-chain:
+    - snap/command-chain/snapcraft-runner
+    - snap/command-chain/desktop-launch
+    environment:
+      LD_LIBRARY_PATH: $LD_LIBRARY_PATH:$SNAP/lib/gnome-calculator
+    plugs:
+    - gsettings
+    - network
+    - desktop
+    - desktop-legacy
+    - wayland
+    - x11
 architectures:
 - amd64
+assumes:
+- command-chain
+base: core18
 confinement: strict
 environment:
   SNAP_DESKTOP_RUNTIME: $SNAP/gnome-platform
@@ -33,40 +50,23 @@ layout:
     bind: $SNAP/gnome-platform/usr/share/xml/iso-codes
 plugs:
   gnome-3-28-1804:
-    default-provider: gnome-3-28-1804:gnome-3-28-1804
     interface: content
     target: $SNAP/gnome-platform
+    default-provider: gnome-3-28-1804
   gtk-3-themes:
-    default-provider: gtk-common-themes:gtk-3-themes
     interface: content
     target: $SNAP/data-dir/themes
+    default-provider: gtk-common-themes
   icon-themes:
-    default-provider: gtk-common-themes:icon-themes
     interface: content
     target: $SNAP/data-dir/icons
+    default-provider: gtk-common-themes
   sound-themes:
-    default-provider: gtk-common-themes:sound-themes
     interface: content
     target: $SNAP/data-dir/sounds
+    default-provider: gtk-common-themes
 slots:
   gnome-calculator:
-    bus: session
     interface: dbus
+    bus: session
     name: org.gnome.Calculator
-apps:
-  gnome-calculator:
-    command: usr/bin/gnome-calculator
-    command-chain:
-    - snap/command-chain/snapcraft-runner
-    - snap/command-chain/desktop-launch
-    environment:
-      LD_LIBRARY_PATH: $LD_LIBRARY_PATH:$SNAP/lib/gnome-calculator
-    plugs:
-    - gsettings
-    - network
-    - desktop
-    - desktop-legacy
-    - wayland
-    - x11
-assumes:
-- command-chain
```

The two differences are:
- ordering of snap yaml
- default-provider always prints the slot (even though it's almost always, but not guaranteed??, to be the name of the plug)

cjp256:
(and file lists are identical)

sergiusens:
Thanks for checking!

PR #2710, opened 2019-09-11
Add setting to remove the "jar" option in the gradle command

LyzardKing:
This is needed to build using the new java11 module launcher
Is it something that can be included in snapcraft? 
I've made the default to be as it is now, so nothing has to be changed.

At the moment I've added a "custom plugin" in the project I need this in (https://github.com/JabRef/jabref/pull/5305)

The static tests are succesful in the test environment.
The other tests are having issues mainly because of my internet connection.
I'm running them again...

-----


LyzardKing:
@sergiusens I added a single unit test to verify if setting the option to false actually removes the jar parameter...Is it enough?
About the spread test...I'm not sure how to add that...I'll look into it a bit.

LyzardKing:
For the spread tests, is it cheating if I developed a test in which I have a bash script that prints hello gradle, and use the gradle copy method to simply copy that (so I don't need the "jar", but I'm not technically compiling java...)

LyzardKing:
@sergiusens This plugin change is needed to enable the jlink java builds, which are bundles (folders)
with a launcher and all the necessary java libs (taken from the build system).
So another thing needed is to remove the openjdk jre from the stage-packages list.

Locally I added the same condition build-jre to not automatically add the openjdk-jre as a stage-package if it's set as true(like the default) but it could be set as a different option (something like no_system_jvm?)

I changed the spread test to actually compile a jlink java hello world...and it works with the changes I made..

LyzardKing:
This has been superseded by packaging the snap, but it might be useful for other applications using gradle without building jar files.
I hit a roadblock with the tests, if there are suggestions I'm open for any tips ;)

PR #2718, opened 2019-09-16
Update README - remove suggestion to use forum

cevap:
Forum is bad place to send anybody to disucuss anything and should not be used. They ban and censor criticism which is in no way proper addressing of the issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,9 @@ dependencies, they work on all major Linux systems without modification.
 
 ## Get support
 
-We’re here to help. Ask your questions at the [Snapcraft Forum](https://forum.snapcraft.io). Report bugs on [Launchpad](https://bugs.launchpad.net/snapcraft/+filebug).
+~~We’re here to help. Ask your questions at the [Snapcraft Forum](https://forum.snapcraft.io)~~. Report bugs on [Launchpad](https://bugs.launchpad.net/snapcraft/+filebug).
+
+Snapcraft forum should not be used anymore for any questions, forum staff is incompetent in many regard, good place to ask is stackexchange, there you will find competent staff and proper replies.
 
 Learn about the latest features by following Snapcraft on
 [Twitter](https://twitter.com/snapcraftio),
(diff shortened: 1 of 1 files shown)


cevap:
your staff banning for nothing accounts, please reply here why is it happening. Forum should be moved to competent place, to stackexchange, not shady forum censoring and banning due to personal problems, unbelieveable!

cevap:
mod who uses such silencing methods:
grafik

these are other mods, not sure what they think about such methods:


If you keep current forum, please make sure that your staff is not abusing their positions for personal reasons.

cevap:
@chipaca Mr. John R. Lenton, could you please provide reasons for a ban as well as a proof of it. Additionally explain what was purpose of you trying to provocate a user as well to write your personal insults.

chipaca:
You're not banned, you're silenced until next week, because you don't seem to be able to engage in productive discourse without demeaning or insulting people. You were repeatedly warned about your behaviour by other forum members as well as myself, and instead of pausing and reflecting you doubled down, toxically. On being silenced you have resorted to filing [a bug on launchpad](https://bugs.launchpad.net/snapcraft/+bug/1844160) as well as this ridiculous attempt here, both of which I think make the case for me rather well.

cevap:
@chipaca you seem to be crazy. So, you guys broke it intentionaly, correct? I am waiting for reply from those who PR'ed. 

Actually only you were insulting, nobody else and that after you posted 3 offtopic provocations, you were not able to reply anything on topic, you were unable to show whatever was wrong and you are unable to write it here.

Conclusion is at the end as addition that you are a LIER. Why do you lie Mr. John R. Lenton? 

Putting on intention such bugs to break software should distance you from ubuntu, you are untrustworthy person and you break this project. I would have not though, but you are incompetent and gently said not the brightest from, at least that is how you have presented yourself. Lies have short legs, you should have learned it.

cevap:
@chipaca you were told to delete the account as you are incapable to do the work. Just do it then.

PR #2727, opened 2019-09-19
storeapi: use the channels attribute in push

sergiusens:
Switch from releasing  manually to adding the channels intent attribute
from the snap-push endpoint.
https://dashboard.staging.snapcraft.io/docs/api/snap.html#push-a-snap-build-to-the-store

Additional fixes for mypy were added to accomodate the new code.

-----

### DIFF
--- a/snapcraft/plugins/_ros/rospack.py
+++ b/snapcraft/plugins/_ros/rospack.py
@@ -18,11 +18,14 @@ import os
 import logging
 import subprocess
 import tempfile
-from typing import List, Set
+from typing import List, Set, TYPE_CHECKING
 
-import snapcraft
 from snapcraft.internal import common, repo
 from snapcraft import formatting_utils
+
+if TYPE_CHECKING:
+    from snapcraft.project import Project
+
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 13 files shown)


tomwardill:
The store API use looks good to me.
The API changes are live, so this should be verifiable against the production store.

sergiusens:
thanks @tomwardill, the only gotcha with this change is that we lose the possibility to know if a channel will be opened with this action (the release api returns the newly opened channels in the resulting payload).

The benefit of this change does outweigh that loss though.

snapcraft/plugins/_ros/rospack.py
```diff
@@ -18,12 +18,15 @@
...
 import tempfile
-from typing import List, Set
+from typing import List, Set, TYPE_CHECKING
 
-import snapcraft
 from snapcraft.internal import common, repo
 from snapcraft import formatting_utils
 
+if TYPE_CHECKING:
+    from snapcraft.project import Project
```
cjp256: why conditional?

sergiusens: the type is only used for type checking and quoted where used.

cjp256: I guess I mean, why change this at all? :)  I've done a similar annotation and am curious what the problem was?

sergiusens: Oh, I cannot remember now and this probably was not needed in what the code ended up looking like :man_shrugging:

PR #2734, opened 2019-09-30
SnapcraftError refactoring

cjp256:


snapcraft/internal/errors.py
```diff
@@ -23,22 +23,80 @@
...
+    fmt_resolution: str = ""
+    fmt_details: str = ""
+    docs_url: str = ""
+    exit_code: int = 2
+    reportable: bool = False
 
     def __init__(self, **kwargs) -> None:
+        self._kwargs = kwargs
+
+        # DEPRECATED: Please define real properties when creating classes.
```
sergiusens: Maybe just create a new error class and make it an AbstractBaseClass while at it that we can start inheriting from and keep the old class as is.

snapcraft/internal/errors.py
```diff
@@ -23,22 +23,80 @@
...
+
+    :cvar fmt: (DEPRECATED) Legacy format string.
+    :cvar fmt_brief: Keyworded format string for error.
+    :cvar fmt_details: Keyworded format string for details.
+    :cvar fmt_resolution: Keyworded format string for resolution.
     """
 
-    fmt = "Daughter classes should redefine this"
+    fmt: str = ""
+    fmt_brief: str = ""
```
sergiusens: if using AbstractBaseClass we can make the below properties (brief, ...) that are required to return formatted strings for this the base class to not be concerned about potentially unknown attributes.

cjp256: Something like?

```
class SnapcraftABCError(Exception, abc.ABC):
    """Base class for all snapcraft exceptions.

    :cvar brief: Concise, single-line description of the error.
    :cvar details: Detailed information, if considered necessary for user.
    :cvar resolution: Concise suggestion for user to resolve error.
    :cvar docs_url: docs.snapcraft.io link, if applicable.
    :cvar exit_code: Exit code to use if this exception causes Snapcraft to exit.
    :cvar reportable: Prompt user to report this error to sentry.
    """

    resolution: str = ""
    details: str = ""
    docs_url: str = ""
    exit_code: int = 2
    reportable: bool = False

    @property
    @abc.abstractmethod
    def brief(self) -> str:
        """Formatted error string using attributes as keywords."""
        pass

    @property
    def printable_message(self) -> str:
        """Formatted string suited to print to the user."""
        parts = [self.brief]

        if self.resolution:
            parts.extend(["", "Recommended resolution:", self.resolution])

        if self.details:
            parts.extend(["", "Detailed information:", self.details])

        if self.docs_url:
            parts.extend(["", "For more information, check out:", self.docs_url])

        return "\n".join(parts)

    def __str__(self):
        return self.printable_message

    def get_exit_code(cls) -> int:
        """Deprecated. Use exit_code class variable instead."""
        return cls.exit_code
```

cjp256: ^ I'm unsure what else can be designated as abstract/required.  Perhaps resolution?

sergiusens: Yes, but make those cvars actual properties too.

```python
class SanpcraftABCError(Exception, meta=abc.ABC):
    @property
    @abc.abstractmethod
    def resolution(self) -> str:
    """Return a string with the resolution... blah blah"""

same for details, docs_url, exit_code and reportable

File: snapcraft/internal/errors.py [hunk 3]
```diff
@@ -42,15 +39,61 @@ def get_exit_code(self):
...
-    :cvar fmt: A format string that daughter classes override
+class SnapcraftAbstractError(SnapcraftError, ABC):
+    """Abstract class for all snapcraft exceptions."""
 
-    """
+    @property
+    @abstractmethod
+    def brief(self) -> str:
+        """Concise, single-line description of the error."""
+        pass
```
sergiusens: this line is not needed

snapcraft/internal/errors.py
```diff
@@ -42,15 +39,61 @@ def get_exit_code(self):
...
+
+    @property
+    def details(self) -> str:
+        """Detailed information, if considered necessary for user."""
+        return None
```
sergiusens:
```suggestion
        return ""
```

Or change the return annotation to `Optional[str]`

snapcraft/cli/_errors.py
```diff
@@ -198,6 +220,18 @@ def _process_outer_exception(exc_info, debug):
     click.echo(_MSG_TRACEBACK_LOCATION.format(trace_filepath))
 
 
+def _get_exception_exit_code(exception: Exception):
+    # SnapcraftException has `exit_code` attribute.
+    if isinstance(exception, errors.SnapcraftException):
+        return exception.exit_code
```
sergiusens: when making this `get_exit_code` things might get cleaner here.

snapcraft/internal/errors.py
```diff
@@ -42,15 +39,35 @@ def get_exit_code(self):
...
-    An exception class inheriting from this will get a prompt to report
-    to sentry if raised and is handled by the exception handler on program
-    exit.
 
-    :cvar fmt: A format string that daughter classes override
+class SnapcraftException(Exception, ABC):
+    """Base class for Snapcraft Exceptions."""
 
-    """
+    docs_url: Optional[str] = None
```
sergiusens: get_

snapcraft/internal/errors.py
```diff
@@ -42,15 +39,35 @@ def get_exit_code(self):
...
+    """Link to documentation on docs.snapcraft.io, if applicable."""
+
+    exit_code: int = 2
```
sergiusens: get_

snapcraft/internal/errors.py
```diff
@@ -42,15 +39,35 @@ def get_exit_code(self):
...
+    """Exit code to use when exiting snapcraft due to this exception."""
+
+    reportable: bool = False
```
sergiusens: get_

snapcraft/internal/errors.py
```diff
@@ -42,15 +39,35 @@ def get_exit_code(self):
...
 class SnapcraftReportableError(SnapcraftError):
-    """Base class for all snapcraft exceptions that integrate with Sentry.
+    """DEPRECATED: Use SnapcraftException instead with reportable=True."""
 
-    An exception class inheriting from this will get a prompt to report
-    to sentry if raised and is handled by the exception handler on program
-    exit.
 
-    :cvar fmt: A format string that daughter classes override
+class SnapcraftException(Exception, ABC):
```
sergiusens: if ABC isn't really working here due to inheriting from Exception, then raising NotImplementedError instead of wrapping with @abstractmethod might be better.

In any case, doesn't this need to be
```python
class SnapcraftException(Exception, meta=ABC):
```
?

snapcraft/cli/_errors.py
```diff
@@ -143,14 +147,33 @@ def _handle_sentry_submission(exc_info) -> None:
...
-def _print_exception_message(exc_info, debug) -> None:
-    # Print exception in exc_info[1].
-    message = str(exc_info[1])
-    echo.error(message)
+def _print_snapcraft_exception_message(exception: errors.SnapcraftException):
+    parts = [exception.get_brief()]
+
+    resolution = exception.get_resolution()
+    if resolution:
+        parts.extend(["", "Recommended resolution:", resolution])
```
sergiusens: why is the first element here ""?

cjp256: To add a newline.  It could be "\nRecommended..." as well

snapcraft/internal/errors.py
cjp256: ABC does work here, with mypy detecting the issue:
`error: Cannot instantiate abstract class 'TestSnapcraftException' with abstract attribute 'get_brief'`

WRT meta=ABC, I think you may be thinking of the other option to use `metaclass=ABCMeta`.  ABC is a helper class which does this for you: 
https://docs.python.org/3/library/abc.html

snapcraft/cli/_errors.py
sergiusens: We can leave it like this for now

PR #2736, opened 2019-10-01
cli: prompt for login if required

sergiusens:
Promp to login if it is required instead of exiting with an error.

-----

### DIFF
--- a/snapcraft/_store.py
+++ b/snapcraft/_store.py
@@ -25,6 +25,7 @@ import re
 import subprocess
 import tempfile
 from datetime import datetime
+from functools import wraps
 from subprocess import Popen
 from typing import Any, Dict, Iterable, List, Optional, TextIO
 
@@ -163,7 +164,7 @@ def _try_login(
     acls: Iterable[str] = None,
     channels: Iterable[str] = None,
     expires: str = None,
-    config_fd: TextIO = None
+    config_fd: TextIO = None,
 ) -> None:
     try:
         store.login(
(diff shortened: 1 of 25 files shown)


snapcraft/_store.py
```diff
@@ -248,21 +249,42 @@ def login(
...
-    try:
-        yield
-    except storeapi.errors.InvalidCredentialsError:
-        logger.error("No valid credentials found." ' Have you run "snapcraft login"?')
-        raise
+def _login_wrapper(instance, method):
+    @wraps(method)
+    def wrapped(*args, **kwargs):
+        try:
+            return method(*args, **kwargs)
```
cjp256: Is there a way to simply detect if you're logged in (or not), or does the target operation have to be performed?

I'm wondering if it would be instead of adding a wrapper class, it would make sense to refactor StoreClient a bit?

I would consider reworking `login()` and moving it into StoreClient.  Then it can be aware of whether or not its already logged in for anything touching the store. Each function can be decorate with `requires_login`, or simply call login() at the beginning to ensure it's logged in before proceeding.

As far as other changes to StoreClient, anything that is being passed store_client (e.g. login(), try_login(), _push_delta, etc.), or creates a new instance of StoreClient appear to be good candidates to get moved right into StoreClient.

Thoughts?  I might be way off target, I haven't got a good picture on the store stuff yet.

sergiusens: I actually want to go the other way and remove the conf from there and pass it the credentials to use from this new class meant for orchestration, which could be a good next step, but I am not looking into refactoring the StoreClient just right now (end of cycle is not a good time to do this).

cjp256:
https://github.com/snapcore/snapcraft/compare/a3cf6d8e1c715e43648a4f54761a4bd15d5c9c64..23b0f1176c8e4dd8696c3c521b39daa5f824bc9e

Looks great!

PR #2738, opened 2019-10-03
mypy: add coverage to tests

cjp256:
mypy.ini: set python version to 3.6
tests: minor fixups for mypy to run successfully
runtests: add mypy coverage of unit tests for static target

### DIFF
--- a/runtests.sh
+++ b/runtests.sh
@@ -36,7 +36,7 @@ run_static_tests() {
     python3 -m flake8 .
 
     echo "Running mypy"
-    mypy --python-version 3.6 -p snapcraft
+    mypy .
 
     echo "Running codespell"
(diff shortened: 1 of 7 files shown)


runtests.sh
```diff
@@ -36,7 +36,7 @@ run_static_tests() {
     python3 -m flake8 .
 
     echo "Running mypy"
-    mypy --python-version 3.6 -p snapcraft
+    mypy --python-version 3.6 snapcraft tests
```
sergiusens: you might be able to drop `--python-version 3.6`, I missed the change on mypy.ini (couldn't find where it was back then)

tests/fixture_setup/_unittests.py
```diff
@@ -532,14 +532,14 @@ def __str__(self):
...
 
-    @property
-    def version(self):
-        return self._version
-
     @property
     def is_auto_removable(self):
         return self.marked_install and self.is_auto_installed
 
+    @property
```
sergiusens: is there a reason for this change?

cjp256: > is there a reason for this change?

Yeah, mypy was erroring out on the following setter.  It appeared to need the @property next to the setter.

sergiusens: gotcha

cjp256:
Should be good to go.

PR #2740, opened 2019-10-04
crystal plugin: add flags to use during shards build

mamantoha:

-----


snapcraft/plugins/crystal.py
```diff
@@ -80,7 +94,12 @@ def build(self):
         super().build()
 
         self.run(["shards", "install", "--production"], self.builddir)
-        self.run(["shards", "build", "--production"], self.builddir)
+
+        build_flags = ["--production"]
```
sergiusens: we should probably let the defaults take place here, how would I build a non production build if I wanted to?

mamantoha: Do you mean here?

```
schema["properties"]["crystal-buildflags"] = {
            "type": "array",
            "minitems": 1,
            "uniqueItems": True,
            "items": {"type": "string"},
            "default": [], # << HERE
        }
```

bcardiff:
Regarding the crystal management of options, it looks good to me.

Thanks @mamantoha !

snapcraft/plugins/crystal.py
sergiusens: That is a good place, please make mention of the default in the docstring.

mamantoha: Done

sergiusens: You seem to have missed adding it to "default" in the schema

mamantoha: Done

sergiusens: ok, so now this line is not needed as it should be taken into account in the conditional right below

PR #2742, opened 2019-10-04
cli: use click utilities for login prompts

sergiusens:

-----

### DIFF
--- a/tests/unit/commands/__init__.py
+++ b/tests/unit/commands/__init__.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright (C) 2015-2018 Canonical Ltd
+# Copyright (C) 2015-2019 Canonical Ltd
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License version 3 as
@@ -13,9 +13,11 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
 import fixtures
 import json
 import os
+import subprocess
 from textwrap import dedent
 
 from click.testing import CliRunner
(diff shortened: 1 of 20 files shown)


tests/unit/commands/test_list_registered.py
```diff
@@ -15,30 +15,41 @@
...
+class ListRegisteredTestCase(FakeStoreCommandsBaseTestCase):
 
     scenarios = [
         ("list-registered", {"command_name": "list-registered"}),
         ("registered alias", {"command_name": "registered"}),
     ]
 
     def test_command_without_login_must_ask(self):
-        result = self.run_command([self.command_name])
+        # TODO: look into why this many calls are done inside snapcraft.storeapi
```
cjp256: is this intended to be fixed in this PR? is this a regression in behavior or consistent because it wasn't checked before?

tests/unit/commands/__init__.py
```diff
@@ -43,8 +45,13 @@ def get_sample_key(name):
     raise KeyError(name)
 
 
-def mock_snap_output(command, *args, **kwargs):
-    if command == ["snap", "keys", "--json"]:
+original_check_output = subprocess.check_output
```
cjp256: This could use an explanation :)

tests/unit/commands/__init__.py
```diff
@@ -124,3 +133,69 @@ def setUp(self):
...
+        )
+        self.useFixture(self.fake_store_release)
+
+        self.fake_store_register_key = fixtures.MockPatchObject(
+            storeapi._sca_client.SCAClient, "register_key"
+        )
+        self.useFixture(self.fake_store_register_key)
+
+        # Mock the snap command
+        self.fake_check_output = fixtures.MockPatch(
```
cjp256: Just the `snap` command?

tests/unit/commands/__init__.py
sergiusens: Sure, it is a pattern we have been using across our tests, but I can certainly doc it

tests/unit/commands/__init__.py
sergiusens: Yes, the side_effect mocks the snap command, let's unsquashfs pass and fails on every other call

tests/unit/commands/test_list_registered.py
sergiusens: Not related to the PR, I bet it is "refresh_if_necessary" in the storeapi. We never checked this before, we never seem to have checked many things before which now are properly checked.

PR #2743, opened 2019-10-07
meta: warn about command mangling

sergiusens:
Warn when a wrapper is needed due to the command used or when the command
has changed to something more compatible with snapd's accepted format.

-----

### DIFF
--- a/snapcraft/internal/meta/command.py
+++ b/snapcraft/internal/meta/command.py
@@ -15,6 +15,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import contextlib
+import logging
 import os
 import re
 import shlex
@@ -25,7 +26,29 @@ from . import errors
 from ._utils import _executable_is_valid
 from snapcraft.internal import common
 
+
+logger = logging.getLogger(__name__)
 _COMMAND_PATTERN = re.compile("^[A-Za-z0-9. _#:$-][A-Za-z0-9/. _#:$-]*$")
+_FMT_COMMAND_FROM_BASE = (
+    "The command {!r} was not found in the prime directory, it has been "
+    "changed to {!r}."
+)
+_FMT_SNAP_STRIP = (
+    "Stripped '$SNAP/' from command {!r}. "
+    "Relative paths from the prime are required by snapd."
+)
+_FMT_SNAPD_WRAPPER = (
+    "A shell wrapper will be generated for command {!r} as it does not conform "
+    "with the command pattern expected by snapd. Commands must be relative to "
+    "the prime directory and can only consist of alphanumeric characters, "
+    "spaces, and the following special characters: / . _ # : $ -"
+)
+_FMT_COMMAND_CHANGED = (
+    "Command {!r} has been changed to {!r} to conform with the "
+    "command pattern expected by snapd. Commands must be relative to "
+    "the prime directory and can only consist of alphanumeric characters, "
+    "spaces, and the following special characters: / . _ # : $ -"
+)
 
 
 def _get_shebang_from_file(file_path: str) -> List[str]:
@@ -157,6 +180,7 @@ class Command:
     def __init__(self, *, app_name: str, command_name: str, command: str) -> None:
         self._app_name = app_name
         self._command_name = command_name
+        self.original_command = command
         self.command = command
         self.wrapped_command: Optional[str] = None
         self.massaged_command: Optional[str] = None
@@ -195,9 +219,24 @@ class Command:
         if self.requires_wrapper:
             if not can_use_wrapper:
                 raise errors.InvalidAppCommandFormatError(self.command, self._app_name)
+            if self.original_command.startswith("$SNAP"):
+                logger.warning(_FMT_SNAP_STRIP.format(self.original_command))
+            if self.command.startswith("/") and not self.original_command.startswith(
+                "/"
+            ):
+                logger.warning(
+                    _FMT_COMMAND_FROM_BASE.format(self.original_command, self.command)
+                )
+            if not _COMMAND_PATTERN.match(self.command):
+                logger.warning(_FMT_SNAPD_WRAPPER.format(self.command))
             self.wrapped_command = self.command
             self.command = self.wrapped_command_name
             return self.command
+
+        if self.command != self.original_command:
+            logger.warning(
+                _FMT_COMMAND_CHANGED.format(self.original_command, self.command)
+            )
 
         command_parts = shlex.split(self.command)
         command_path = os.path.join(prime_dir, command_parts[0])
--- a/tests/unit/meta/test_command.py
+++ b/tests/unit/meta/test_command.py
@@ -14,8 +14,10 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import logging
 import os
 
+import fixtures
 from testtools.matchers import Equals, Is, FileContains, FileExists
 
 from snapcraft.internal.meta import command, errors
@@ -28,7 +30,18 @@ def _create_file(file_path: str, *, mode=0o755) -> None:
     os.chmod(file_path, mode)
 
 
-class CommandWithoutWrapperTest(unit.TestCase):
+class UnmodifiedCommandWithoutWrapperTest(unit.TestCase):
+    def setUp(self):
+        super().setUp()
+
+        self.fake_logger = fixtures.FakeLogger(level=logging.WARNING)
+        self.useFixture(self.fake_logger)
+
+    def tearDown(self):
+        super().tearDown()
+
+        self.assertThat(self.fake_logger.output, Equals(""))
+
     def test_command(self):
         _create_file(os.path.join(self.path, "foo"))
         cmd = command.Command(app_name="foo", command_name="command", command="foo")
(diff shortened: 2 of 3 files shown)


cjp256:
Looks good, though passing echoer around seems increasingly painful.  What do you think about simply allowing:
```
import echo

echo.warning(..) ?
```

sergiusens:
I really want these inner modules to dependency inject anything that is more of a cli thing.

This will make it easy to then choose if we want to log to a file or print out everything or duplex it.

the build_provider stuff works similarly to this, so it is at least consistent

cjp256:
From the user perspective, I don't know that they would really care about command wrappers, etc?  Why would / should they care?

I haven't fully thought it through, at a glance, I would think you maybe could save off the command before `prime_command()` in `finalize_snap_meta_commands()` and check if the command is modified.  Then just inform the user that the command is being rewritten and suggest they incorporate the rewrite into their snapcraft.yaml?  It would avoid the need for echoer propagation, but perhaps more importantly, provide a simpler warning/suggestion?

cjp256:
> I really want these inner modules to dependency inject anything that is more of a cli thing.
> 
> This will make it easy to then choose if we want to log to a file or print out everything or duplex it.
> 
> the build_provider stuff works similarly to this, so it is at least consistent

Is it there documentation somewhere for snapcraft's usage of logger vs. click vs. echoer and the well-supported log levels?  From what I've seen, logger is nearly exclusively used for debug() or perhaps as a stand-in where echoer is not available (and echo isn't used)?

My assumption was that that echo is a wrapper around click to handle snapcraft-specific CLI.  It could take care of duplex or any output configuration desired between levels or snapcraft-specific details (e.g. is_tty_connected).

If echoer would expanded to support debug(), what gaps remain to unifying logger & click & echo into a single pattern across snapcraft that best supports library & CLI code?

sergiusens:
No documentation. Our use of logger is deprecated, use of logger.info, logger.warning and logger.warn are left overs, we want to move away from random logging from packages and modules.
logger.debug will not be related to echo.debug, one is developer focused while the latter will be, when it exists, user focused.

The cli comment makes little sense as we could have re-written the logger writer to use click.echo too, but we chose to go down the path of having everything new inject the "console writer" if needed (the insight and desire came after fixing bugs in our dependencies that printed even when we did not want them to)

tests/unit/meta/test_command.py
```diff
@@ -108,6 +143,32 @@ def test_command(self):
...
+        )
+        cmd.prime_command(
+            can_use_wrapper=True, massage_command=True, prime_dir=self.path
+        )
+
+        self.assertThat(cmd.command, Equals("command-foo.wrapper"))
+        self.assertThat(
+            self.fake_logger.output.strip(),
+            Equals(
+                "Command '$SNAP/foo !option' has been changed to 'foo !option' "
```
cjp256: This doesn't look quite right due to the mixed messaging `($ is allowed, ! is not, but  you lost $ and kept !)`.  But looking at it in this context is helpful :D

When use of "$SNAP" is detected, I would maybe just independently indicate that it's being dropped because it's not required:
```
"Stripping '$SNAP/' from command '<command>'.  Relative paths are assumed to be relative to $SNAP."
```

A suggested format for wrappers due to special characters:
```
Snapcraft created a wrapper script to execute '<command>' .  Commands may only consist of alphanumeric characters, spaces, and the following special characters: `/ . _ # : $ -'` and the configured command contains unsupported characters.  To resolve this issue for you automatically, snapcraft configured a wrapper script '<relative path>' to execute '<wrapped command>'.  To remove the need for this wrapper script, you must remove unsupported characters from the command.
```

tests/unit/meta/test_command.py
```diff
@@ -151,6 +221,14 @@ def test_command_starts_with_slash(self):
         self.expectThat(cmd.command, Equals("command-foo.wrapper"))
         self.assertThat(wrapper_path, FileExists())
         self.assertThat(wrapper_path, FileContains('#!/bin/sh\nexec /foo "$@"\n'))
+        self.assertThat(
+            self.fake_logger.output.strip(),
+            Equals(
```
cjp256: A suggestion for absolute paths:

```
Snapcraft created a wrapper script to execute '<command>' .  Commands must be relative to the snap, and this command was not found.  To resolve this issue for you automatically, snapcraft configured a wrapper script '<relative path>' to execute '<wrapped command>'.  To remove the need for this wrapper script, you must use a relative path that can be found in the snap.
```

snapcraft/internal/meta/command.py
```diff
@@ -25,7 +26,29 @@
...
+
+logger = logging.getLogger(__name__)
 _COMMAND_PATTERN = re.compile("^[A-Za-z0-9. _#:$-][A-Za-z0-9/. _#:$-]*$")
+_FMT_COMMAND_FROM_BASE = (
+    "The command {!r} was not found in the prime directory, it has been "
+    "changed to {!r}."
+)
+_FMT_SNAP_STRIP = (
+    "Stripped '$SNAP/' from command {!r}. "
+    "Relative paths from the prime are required by snapd."
```
cjp256: Is it really that it needs to be relative?  I figured it was just be simplified, or $VARIABLES aren't resolved?

snapcraft/internal/meta/command.py
```diff
@@ -195,10 +219,25 @@ def prime_command(
         if self.requires_wrapper:
             if not can_use_wrapper:
                 raise errors.InvalidAppCommandFormatError(self.command, self._app_name)
+            if self.original_command.startswith("$SNAP"):
+                logger.warning(_FMT_SNAP_STRIP.format(self.original_command))
+            if self.command.startswith("/") and not self.original_command.startswith(
+                "/"
+            ):
+                logger.warning(
```
cjp256: I think this error will be odd if this happens due to the interpreter being picked up?

tests/unit/meta/test_command.py
```diff
@@ -163,6 +238,19 @@ def test_command_relative_command_found_in_slash(self):
         self.expectThat(cmd.command, Equals("command-foo.wrapper"))
         self.assertThat(wrapper_path, FileExists())
         self.assertThat(wrapper_path, FileContains('#!/bin/sh\nexec /bin/sh "$@"\n'))
+        self.assertThat(
+            self.fake_logger.output.strip(),
+            Equals(
+                "The command 'sh' was not found in the prime directory, it "
+                "has been changed to '/bin/sh'."
+                "\n"
+                "A shell wrapper will be generated for command '/bin/sh' "
```
cjp256: ...but you just changed it to '/bin/sh' for me! :)  my command "sh" pattern was relative and did not have those problems!! :thinking: :laughing:

sergiusens: It would have only been a non-working snap :shrug:

snapcraft/internal/meta/command.py
sergiusens: Good catch

snapcraft/internal/meta/command.py
sergiusens: Yes it does, they NEED to be relative, variables are resolved, but the command's first argument needs to be **the** path itself

snapcraft/internal/meta/command.py
```diff
@@ -25,7 +26,29 @@
...
+    "The command {!r} was not found in the prime directory, it has been "
+    "changed to {!r}."
+)
+_FMT_SNAP_STRIP = (
+    "Stripped '$SNAP/' from command {!r}. "
+    "Relative paths from the prime are required by snapd."
+)
+_FMT_SNAPD_WRAPPER = (
+    "A shell wrapper will be generated for command {!r} as it does not conform "
+    "with the command pattern expected by snapd. Commands must be relative to "
```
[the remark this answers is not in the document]
sergiusens: It is what the exception where this is logged returns

snapcraft/internal/meta/command.py
sergiusens: Good catch, this might be tricky

sergiusens: This will not happen, we raise `errors.ShebangInRoot` for that scenario, which avoids the wrapper

snapcraft/internal/meta/command.py
```diff
@@ -195,10 +219,25 @@ def prime_command(
...
+                )
+            if not _COMMAND_PATTERN.match(self.command):
+                logger.warning(_FMT_SNAPD_WRAPPER.format(self.command))
             self.wrapped_command = self.command
             self.command = self.wrapped_command_name
             return self.command
 
+        if self.command != self.original_command:
+            logger.warning(
+                _FMT_COMMAND_CHANGED.format(self.original_command, self.command)
```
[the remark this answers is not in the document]
sergiusens: This is snapcraft,
```
Command 'bin/snapcraft' has been changed to 'usr/bin/python3 $SNAP/bin/snapcraft' to conform with the command pattern expected by snapd. Commands must be relative to the prime directory and can only consist of alphanumeric characters, spaces, and the following special characters: / . _ # : $ -
Snapping 'snapcraft' \                                                           
Snapped snapcraft_3.8+git55.gbd3ebe88b_amd64.snap
sergiusens@gotham:~/projects/snapcore/snapcraft$ grep command /snap/snapcraft/current/meta/snap.yaml 
    command: usr/bin/python3 $SNAP/bin/snapcraft
```
Probably needs fixing.

sergiusens:
Closing in favour of #2748

PR #2745, opened 2019-10-08
tests: update spread tests to account for content snaps

cjp256:
tests: update gnome-3-28 extension spread test to use gtk

tests: update gnome extension spread task to account for content snaps

tests: update kde extension spread task to account for content snaps


cjp256:
For future use, I PR'd a NOMATCH helper for spread to simplify those checks.

tests/spread/extensions/gnome-3-28/task.yaml
```diff
@@ -28,18 +32,32 @@ restore: |
   . "$TOOLS_DIR/snapcraft-yaml.sh"
   restore_yaml "snap/snapcraft.yaml"
 
+  # Remove content snaps installed by test.
```
sergiusens: This is not needed, it is done in the general restore (`restore-each` in `spread.yaml`)

tests/spread/extensions/gnome-3-28/task.yaml
```diff
@@ -12,13 +12,17 @@ systems:
...
 environment:
-  SNAP_DIR: ../snaps/glib-hello
+  SNAP_DIR: ../snaps/gtk3-hello
 
 prepare: |
   #shellcheck source=tests/spread/tools/snapcraft-yaml.sh
   . "$TOOLS_DIR/snapcraft-yaml.sh"
   set_base "$SNAP_DIR/snap/snapcraft.yaml"
 
+  # Remove content snaps to be installed by test.
```
sergiusens: This really shouldn't be needed, I suspect this returning 0 may even be a bug and will hit us in the future.

tests/spread/extensions/kde-neon/task.yaml
```diff
@@ -16,6 +16,9 @@ prepare: |
   . "$TOOLS_DIR/snapcraft-yaml.sh"
   set_base "$SNAP_DIR/snap/snapcraft.yaml"
 
+  # Remove content snaps to be installed by test.
```
sergiusens: ditto

tests/spread/extensions/kde-neon/task.yaml
```diff
@@ -16,6 +16,9 @@ prepare: |
...
+  sudo snap remove kde-frameworks-5-core18
```
sergiusens: ditto

tests/spread/extensions/kde-neon/task.yaml
```diff
@@ -25,9 +28,12 @@ restore: |
   . "$TOOLS_DIR/snapcraft-yaml.sh"
   restore_yaml "snap/snapcraft.yaml"
 
+  # Remove content snaps installed by test.
```
sergiusens: ditto

tests/spread/extensions/gnome-3-28/task.yaml
cjp256: Thanks!

cjp256:
> Just remove the logic that removes, we should expect a clean environment already.
> 
> If you want to be sure, grep $output and verify the snaps were installed or just check if they are installed in `prepare`, we will have a bigger problem if the snap was installed prior

Ah, I did realize `restore-each` purged all the snaps. Fixed up unnecessary removes! Thanks!

PR #2747, opened 2019-10-08
nodejs plugin: fix errors when building in confinement

NickZ:

-----
This PR fixes errors when running the nodejs plugin with the base keyword.

I am now running `npm install` with `--unsafe-perm` and `SUDO_UID` unset. When building in a VM, node will crash upon running certain scripts because it drops the root UID when running them. It will sometimes ignore `--unsafe-perm` on certain commands, such as running `git` during a `npm install` that has a `git:` dependency, and the only way to override that is to drop `SUDO_UID`.

I am no longer running `npm install` during the pull step because 1) it adds a significant amount of build time between the 'pull' and 'build' step, and 2) it is often the case that changes need to be made during the build process in between downloading node/npm/yarn and running `npm install`, such as importing private packages, setting custom a registry, etc.

I also skip running `npm ls` and populating the `manifest["node-packages"]` if yarn is used, because yarn often creates different dependency trees and it will crash the build if `npm ls` detects that the dependency graph is different from one that npm would have generated.

### DIFF
--- a/snapcraft/plugins/nodejs.py
+++ b/snapcraft/plugins/nodejs.py
@@ -147,8 +147,8 @@ class NodePlugin(snapcraft.BasePlugin):
         if self.options.nodejs_package_manager == "yarn":
             self._yarn_tar.download()
 
-        # do the install in the pull phase to download all dependencies.
-        self._install(rootdir=self.sourcedir)
+        # install node and yarn.
+        self._install_node_and_yarn(rootdir=self.sourcedir)
 
     def clean_pull(self):
         super().clean_pull()
@@ -178,19 +178,27 @@ class NodePlugin(snapcraft.BasePlugin):
                 self._manifest["yarn-lock-contents"] = lock_file.read()
 
         # Get the names and versions of installed packages
-        installed_node_packages = self._get_installed_node_packages(self.installdir)
-        self._manifest["node-packages"] = [
-            "{}={}".format(name, installed_node_packages[name])
-            for name in installed_node_packages
-        ]
+        if self.options.nodejs_package_manager == "npm":
+            installed_node_packages = self._get_installed_node_packages(self.installdir)
+            self._manifest["node-packages"] = [
+                "{}={}".format(name, installed_node_packages[name])
+                for name in installed_node_packages
+            ]
+        # Skip this step if yarn is used, as it may produce different
+        # dependency trees than npm
+        else:
+            self._manifest["node-packages"] = []
 
-    def _install(self, rootdir):
+    def _install_node_and_yarn(self, rootdir):
         self._nodejs_tar.provision(self._npm_dir, clean_target=False, keep_tarball=True)
         if self.options.nodejs_package_manager == "yarn":
             self._yarn_tar.provision(
                 self._npm_dir, clean_target=False, keep_tarball=True
             )
+        # Check to see if package.json exists in pull step
+        self._get_package_json(rootdir)
 
+    def _install(self, rootdir):
         cmd = [os.path.join(self._npm_dir, "bin", self.options.nodejs_package_manager)]
 
         if self.options.nodejs_package_manager == "yarn":
@@ -200,9 +208,11 @@ class NodePlugin(snapcraft.BasePlugin):
                 cmd.extend(["--https-proxy", os.getenv("https_proxy")])
 
         flags = []
-        if rootdir == self.builddir:
-            flags = ["--offline", "--prod"]
 
+        if self.options.nodejs_package_manager == "npm":
+            flags.append("--unsafe-perm")
+
+        # Run once to download dependencies and run install scripts
         self.run(cmd + ["install"] + flags, rootdir)
 
         package_json = self._get_package_json(rootdir)
@@ -232,6 +242,9 @@ class NodePlugin(snapcraft.BasePlugin):
                 os.path.join(package_dir, "yarn.lock"),
             )
 
+        flags.append("--offline")
+        flags.append("--prod")
+
         self.run(cmd + ["install"] + flags, package_dir)
 
         return package_dir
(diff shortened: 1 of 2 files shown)


snapcraft/plugins/nodejs.py
```diff
@@ -178,19 +178,27 @@ def build(self):
...
+            self._manifest["node-packages"] = []
 
-    def _install(self, rootdir):
+    def _install_node_and_yarn(self, rootdir):
         self._nodejs_tar.provision(self._npm_dir, clean_target=False, keep_tarball=True)
         if self.options.nodejs_package_manager == "yarn":
             self._yarn_tar.provision(
                 self._npm_dir, clean_target=False, keep_tarball=True
             )
+        # Check to see if package.json exists in pull step
```
cjp256: Curious, what's the point of this?

snapcraft/plugins/nodejs.py
```diff
@@ -251,6 +264,15 @@ def _build_environment(self):
         else:
             new_path = npm_bin
 
+        # npm has the behavior that, if it detects that SUDO_UID is set,
+        # it will then set the uid of some of its child processes (such as
+        # git when installing a package that specifies a git repository)
+        # to that of the sudoer, ignoring --unsafe-perm entirely.
+        # We have to unset SUDO_UID and SUDO_GID to prevent this.
+
+        env.pop("SUDO_UID", None)
```
cjp256: Did you hit this case by chance?  If so, presumably using destructive-mode?  I'm curious how/when it's set in the environment :)

snapcraft/plugins/nodejs.py
NickZ: I wanted it to crash during the pull step if it couldn't find the package.json, since that's what would happen previously, but it's not strictly necessary.

snapcraft/plugins/nodejs.py
NickZ: No, this is a known bug/feature in npm. I've made a report about it here, it's fairly easy to replicate: https://npm.community/t/npm-ignores-unsafe-perm-and-user-params-when-running-git-as-part-of-install/10437

NickZ:
> I'll leave it to folks with more knowledge of nodejs to further review... but I'm curious if you have any examples of broken builds you could provide?

Not publicly, no. Most of my broken builds have been private packages that require private dependencies. Although this package is useful for demonstrating the SUDO_UID bug: https://github.com/NickZ/demonstrate-unsafe-perm-git

snapcraft/plugins/nodejs.py
sergiusens: This is reproduceable @cjp256 and we have a bug for it too https://bugs.launchpad.net/snapcraft/+bug/1816204

snapcraft/plugins/nodejs.py
cjp256: Great, thanks - makes sense!

To clarify, by "crash", you are referring to the error message "Could not find a 'package.json' in the source tree...", correct?

snapcraft/plugins/nodejs.py
cjp256: Thanks!

cjp256:
> > I'll leave it to folks with more knowledge of nodejs to further review... but I'm curious if you have any examples of broken builds you could provide?
> 
> Not publicly, no. Most of my broken builds have been private packages that require private dependencies. Although this package is useful for demonstrating the SUDO_UID bug: https://github.com/NickZ/demonstrate-unsafe-perm-git

Thank you :)

snapcraft/plugins/nodejs.py
NickZ: Yes, that's right.

snapcraft/plugins/nodejs.py
```diff
@@ -178,19 +178,27 @@ def build(self):
...
-            "{}={}".format(name, installed_node_packages[name])
-            for name in installed_node_packages
-        ]
+        if self.options.nodejs_package_manager == "npm":
+            installed_node_packages = self._get_installed_node_packages(self.installdir)
+            self._manifest["node-packages"] = [
+                "{}={}".format(name, installed_node_packages[name])
+                for name in installed_node_packages
+            ]
+        # Skip this step if yarn is used, as it may produce different
```
sergiusens: Can you expand on this?

NickZ: Basically, I've noticed that most builds that I'm doing with yarn will crash at this step in the build because `npm ls` returns an error after it can't find all the dependencies in the tree. Yarn structures its dependencies differently from npm, so it's common for the dependency tree for npm to be different from the one for yarn. Since we can't get the output of `yarn ls` easily, I just skip this step for yarn.

snapcraft/plugins/nodejs.py
```diff
@@ -232,6 +242,9 @@ def _install(self, rootdir):
                 os.path.join(package_dir, "yarn.lock"),
             )
 
+        flags.append("--offline")
```
palavrov: What's the reason for '--offline' here?
Seems that it breaks my build:

> 
> /root/parts/server/npm/bin/npm install --unsafe-perm --offline --prod
> npm ERR! code ENOTCACHED
> npm ERR! request to https://registry.npmjs.org/@feathersjs%2fexpress failed: cache mode is 'only-if-cached' but no cached response available.
>

palavrov: The previous code
```
        if rootdir == self.builddir:
            flags = ["--offline", "--prod"]
```
actually was never executed because it was only called `self._install(rootdir=self.sourcedir)`

palavrov: Workaround was to implicitly install globally all packages with preinstall hook i.e.

```json
  "scripts": {
    "preinstall": "npm install --global @feathersjs/express express socket.io avahi-dbus github:Andersbakken/node-posix#node12fix"
  },  
```

which in its own turn failed (because `SUDO_UID` fix doesn't cower `github:...` modules) with this error:

> \> server@1.0.0 preinstall /root/parts/server/build
> \> npm install --global @feathersjs/express express socket.io avahi-dbus github:Andersbakken/node-posix#node12fix
> 
> npm ERR! Error while executing:
> npm ERR! /usr/bin/git ls-remote -h -t ssh://git@github.com/Andersbakken/node-posix.git
> npm ERR! 
> npm ERR! fatal: failed to stat '/root/parts/server/build': Permission denied
> npm ERR! 
> npm ERR! exited with error code: 128

which remind me that [already fixed that issue few months ago](https://forum.snapcraft.io/t/multiple-problems-when-build-on-mac/12693/5?u=palavrov) - the proper fix is to set these environment variables in `snapcraft.yaml`

```yaml
        build-environment:
            - SUDO_UID: '0'
            - SUDO_GID: '0'
            - SUDO_USER: 'root'
```

and finally my build was able to succeed.

PR #2752, opened 2019-10-16
errors: migrate handful of errors to SnapcraftException

cjp256:
- Add SnapcraftReportableException helper class for reportable
  errors, which simply flips the default value for get_reportable().
- Convert MissingStateClean, StepOutdatedError,
  SnapcraftEnvironmentError, SnapcraftDataDirectoryMissingError,
  and SnapcraftMissingLinkerInBaseError and associated tests.
- Add missing test for SnapcraftDataDirectoryMissingError.
- Update tests to verify expected "reportable" value as well.

The error messages are being shortened and I am also using
this opportunity to simplify them (and the resolution messages).

For example, when StepOutdatedError occurs, instead of suggesting
that the user clean just one part/step when something is out of date,
ask them to instead clean the whole project.  Something unusual has
happened and it's probably a good idea to clean the whole tree.

-----

### DIFF
--- a/snapcraft/internal/errors.py
+++ b/snapcraft/internal/errors.py
@@ -18,7 +18,15 @@ from abc import ABC, abstractmethod
 from snapcraft import formatting_utils
 from snapcraft.internal import steps
 from subprocess import CalledProcessError
-from typing import Dict, List, Union, Optional
+from typing import Dict, List, Union, Optional, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from snapcraft.internal.pluginhandler._dirty_report import DirtyReport
+    from snapcraft.internal.pluginhandler._outdated_report import OutdatedReport
+
+
+# Commonly used resolution message to clean and retry build.
+CLEAN_RESOLUTION = "Run `snapcraft clean` and retry build."
 
 
 class SnapcraftError(Exception):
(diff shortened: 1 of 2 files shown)


snapcraft/internal/errors.py
```diff
@@ -73,91 +73,105 @@ def __str__(self) -> str:
...
-        "{report}"
-        "To continue, clean that part's {step.name!r} step by running "
-        "`snapcraft clean {parts_names} -s {step.name}`."
-    )
 
+class StepOutdatedError(SnapcraftException):
     def __init__(
         self, *, step, part, dirty_report=None, outdated_report=None, dependents=None
     ):
+        self._step = step
```
sergiusens: `AttributeError: 'StepOutdatedError' object has no attribute 'step'`
Maybe leave what was once a public attribute as public?

snapcraft/internal/errors.py
```diff
@@ -73,91 +73,105 @@ def __str__(self) -> str:
...
-        "Failed to reuse files from previous run: "
-        "The {step.name!r} step of {part!r} is out of date:\n"
-        "{report}"
-        "To continue, clean that part's {step.name!r} step by running "
-        "`snapcraft clean {parts_names} -s {step.name}`."
-    )
 
+class StepOutdatedError(SnapcraftException):
     def __init__(
         self, *, step, part, dirty_report=None, outdated_report=None, dependents=None
```
sergiusens: want to add type annotations while at it?

cjp256: Sounds like a good idea :D

snapcraft/internal/errors.py
cjp256: Ahh good catch

cjp256: Done!

snapcraft/internal/errors.py
cjp256: Done!

cjp256:
Added 'CLEAN_RESOLUTION' constant and fixed up members for StepOutdatedError.

snapcraft/internal/errors.py
```diff
@@ -18,7 +18,15 @@
...
 from subprocess import CalledProcessError
-from typing import Dict, List, Union, Optional
+from typing import Dict, List, Union, Optional, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from snapcraft.internal.pluginhandler._dirty_report import DirtyReport
+    from snapcraft.internal.pluginhandler._outdated_report import OutdatedReport
+
+
+"""Commonly used resolution message to clean and retry build."""
```
sergiusens: this should be a code comment (#), as it is now, it is seen as the module's docstring

cjp256: I was thinking it should be doc'd like that as it could be useful for other error modules too... but I'm indifferent so I changed to a code comment anyhow :)

sergiusens: Ah, but this would be the help for the module, not the string following.

PR #2753, opened 2019-10-16
cli: use click utilities for registering on push

sergiusens:
Add registration to the push workflow via use echo.confirm
by adding a decorator to the push related methods that catches
the resource-not-found error code and prompts for registration
before trying to push again.

LP: #1805211

-----

### DIFF
--- a/snapcraft/_store.py
+++ b/snapcraft/_store.py
@@ -266,6 +266,28 @@ def _login_wrapper(method):
     return login_decorator
 
 
+def _register_wrapper(method):
+    def register_decorator(self, *args, snap_name: str, **kwargs):
+        try:
+            return method(self, *args, snap_name=snap_name, **kwargs)
+        except storeapi.errors.StorePushError as push_error:
+            if "resource-not-found" not in push_error.error_list:
+                raise
+            echo.wrapped(
+                "You are required to register this snap before continuing. "
+                "Refer to 'snapcraft help register' for more options."
+            )
+            if echo.confirm(
+                "Would you like to register {!r} with the Snap Store?".format(snap_name)
+            ):
+                self.register(snap_name=snap_name)
+                return method(self, *args, snap_name=snap_name, **kwargs)
+            else:
+                raise
+
+    return register_decorator
+
+
 class StoreClientCLI(storeapi.StoreClient):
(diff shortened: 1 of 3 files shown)


tests/unit/commands/test_push.py
```diff
@@ -152,10 +152,41 @@ def json(self):
...
-        self.fake_store_push_precheck.mock.side_effect = StorePushError(
-            "basic", MockResponse()
+        self.fake_store_push_precheck.mock.side_effect = [
+            StorePushError("basic", MockResponse()),
+            None,
+        ]
+
+        result = self.run_command(["push", self.snap_file], input="y\n")
+
+        self.assertThat(result.exit_code, Equals(0))
```
cjp256: out of curiosity, why not use the short form, self.assertEquals(reference, actual) ?  I tend to use the short-form versions, but notice you use assertThat instead...

```
self.assertEquals(0, result.exit_code)
self.assertRegex(result.output, "You are required to register this snap before continuing.")
```
seems simpler to me... but maybe you prefer the sentence-like structure? :D

snapcraft/_store.py
```diff
@@ -266,6 +266,28 @@ def login_decorator(self, *args, **kwargs):
...
+def _register_wrapper(method):
+    def register_decorator(self, *args, snap_name: str, **kwargs):
+        try:
+            return method(self, *args, snap_name=snap_name, **kwargs)
+        except storeapi.errors.StorePushError as push_error:
+            if "resource-not-found" not in push_error.error_list:
+                raise
+            echo.wrapped(
+                "You are required to register this snap before continuing. "
+                "Refer to 'snapcraft help register' for more options."
```
cjp256: Maybe only print `Refer to 'snapcraft help register' for more options.` if they say no to registration?

tests/unit/commands/test_push.py
sergiusens: Consistency, this is probably another thing we should add to the style guide, something agreed as a team way back...

snapcraft/_store.py
sergiusens: Let's see how this plays out, I do not want to give the impression that the option shown is the only way to register a snap and coerce people into registering something that they may have wanted to be private.

PR #2755, opened 2019-10-17
extensions: kde-neon: add icon and sound themes

galgalesh:

-----

This PR adds the common cursor, icon and sound themes to the `kde-neon` extension using the `gtk-common-themes` content snap. Unlike its name implies, the cursor, icon and sound themes are also applicable to Qt/KDE applications.

This fixes the issue where applications using this extension have a weird pixelated cursor on default Ubuntu.

I tested this by building the Krita snap using this extension: https://invent.kde.org/merlijnsebrechts/krita/blob/neon-extension/packaging/linux/snap/snapcraft.yaml Without this PR, the cursor is correct on Kubuntu, but incorrect on Ubuntu. With this PR, the cursor is correct on both flavors.

### DIFF
--- a/snapcraft/internal/project_loader/_extensions/kde_neon.py
+++ b/snapcraft/internal/project_loader/_extensions/kde_neon.py
@@ -30,8 +30,22 @@ class ExtensionImpl(Extension):
     This extension makes it easy to assemble KDE based applications
     using the Neon stack.
 
-    The version of Neon used is dependent on the base used to create
-    the snap.
+    It configures each application with the following plugs:
+
+    \b
+    - Common Icon Themes.
+    - Common Sound Themes.
+    - The Qt5 and KDE Frameworks runtime libraries and utilities.
+
+    For easier desktop integration, it also configures each application
+    entry with these additional plugs:
+
+    \b
+    - desktop (https://snapcraft.io/docs/desktop-interface)
+    - desktop-legacy (https://snapcraft.io/docs/desktop-legacy-interface)
+    - gsettings (https://snapcraft.io/docs/gsettings-interface)
+    - wayland (https://snapcraft.io/docs/wayland-interface)
+    - x11 (https://snapcraft.io/docs/x11-interface)
     """
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


snapcraft/internal/project_loader/_extensions/kde_neon.py
```diff
@@ -30,8 +30,22 @@ class ExtensionImpl(Extension):
     This extension makes it easy to assemble KDE based applications
     using the Neon stack.
 
-    The version of Neon used is dependent on the base used to create
-    the snap.
+    It configures each application with the following plugs:
+
+    \b
+    - Common Icon Themes.
+    - Common Sound Themes
```
cjp256: nit: missing period to match line before & after  (either way, but consistent)

galgalesh: Fixed, thanks! The gnome extension also had the same issue, I fixed it there too.

apachelogger:
looks good to us :+1: 

I do rather wonder why one of the snapd interfaces doesn't forward both of these data. Any snap would only need to contain one sound theme and one icon theme (to deal with potential absence of either on the host system, which already calls the icon theme into question IMHO) but outside that we'd want to use the host resources so I am not sure why snaps need a sound theme, or in point of fact why they need  250Mib of portable host->confinement assets. Anyway, something to muse on perhaps, nothing wrong with the PR.

sergiusens:
@apachelogger this is a temporary measure, there is a longer term plan that I cannot hold in my head but @jhenstridge or @kenvandine could dive into.

snapcraft/internal/project_loader/_extensions/kde_neon.py
```diff
@@ -48,14 +62,24 @@ def __init__(self, *, extension_name: str, yaml_data: Dict[str, Any]) -> None:
         platform_snap = _PLATFORM_SNAP[yaml_data.get("base")]
         self.root_snippet = {
             "plugs": {
+                "icon-themes": {
+                    "interface": "content",
+                    "target": "$SNAP/data-dir/icons",
+                    "default-provider": "gtk-common-themes:icon-themes",
```
sergiusens: Please take a look at #2763

galgalesh: I removed the slot specification but I don't see any difference in the resulting snap. It seems snapcraft strips the slot specification before writing `snap.yaml`.

Is there a bug report or more information about @kenvandine his PR?

PR #2756, opened 2019-10-20
WIP: extractors: match setuptools metadata keys

galgalesh:

-----

* Use setuptools' `description` as snapcraft's summary
* Use setuptools' `long_description` as snapcraft's description

Fixes LP https://bugs.launchpad.net/snapcraft/+bug/1813364
See also https://forum.snapcraft.io/t/using-external-metadata/4642/10

This was first proposed in March: https://github.com/snapcore/snapcraft/pull/2518, but the tests failed and the author didn't respond.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -97,19 +97,19 @@ For example:
   * To run only the unit tests in the test_nodejs module:
 
     ```
-    python3 -m unittest tests.unit.plugins.tests_nodejs
+    python3 -m unittest tests.unit.plugins.test_nodejs
     ```
 
-  * To run only the unit tests in the NodePluginTestCase:
+  * To run only the unit tests in the NodePluginTest:
 
     ```
-    python3 -m unittest tests.unit.plugins.tests_nodejs.NodePluginTestCase
+    python3 -m unittest tests.unit.plugins.test_nodejs.NodePluginTest
     ```
 
-  * To run only the unit test named test_pull_executes_npm_run_commands:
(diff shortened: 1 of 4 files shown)


galgalesh:
This functionality has a number of issues.

1. **`long_description` is often the Github README file** which starts with explaining how to install the app on Mac, Windows and Linux. Examples: [glances](https://github.com/nicolargo/glances/blob/develop/README.rst), [term2048](https://github.com/bfontaine/term2048/blob/master/README.rst). This is very common because PyPi shows `long-description` as the homepage of the project.
2. `long_description` is often in an incompatible format (rst).
3. `long_description` is often too long for the `description` field
4. Some project don't use `long_description`. Which means that this PR might cause existing snaps to fail their builds, if they don't have a `description` key and `setup.py` doesn't have `long_description`. Debugging these issues is non-intuitive given the description/long_description mismatch.
5. Getting the metadata from `setup.py` is also an issue in itself because the `setup.py` file needs to be executed in the exact same way as when the application is built/installed, because this file can require arbitrary modules as dependencies. Snapcraft currently doesn't do that so if the `setup.py` file imports non-standard modules, then the `parse-info` logic fails. Moreover, the [term2048](https://github.com/bfontaine/term2048/blob/master/README.rst) project simply refuses to build when I use `parse-info`. I didn't investigate this further.

In my opinion, issue 1 is bad enough to make this a WONTFIX, because `long_description` clearly has a different purpose than a snap's `description`.

PR #2761, opened 2019-10-22
Core20 base

sergiusens:

-----

### DIFF
--- a/schema/snapcraft.json
+++ b/schema/snapcraft.json
@@ -886,7 +886,7 @@
                     ]
                 },
                 {
-                    "usage": "base: <base> and type: <app|gadget> (without a build-base)",
+                    "usage": "base: <base> and type: <app|gadget>",
                     "properties": {
                         "type": {
                             "enum": [
@@ -900,13 +900,6 @@
                             "required": [
                                 "base"
                             ]
-                        },
-                        {
-                            "not":{
-                                "required": [
-                                    "build-base"
-                                ]
-                            }
                         }
                     ]
                 },
--- a/snapcraft/internal/meta/_snap_packaging.py
+++ b/snapcraft/internal/meta/_snap_packaging.py
@@ -249,16 +249,12 @@ def _update_yaml_with_defaults(config_data, schema):
     # necessary. Defaults are taken from the schema. Technically these are the
     # only two optional keywords currently WITH defaults, but we don't want to
     # risk setting something that we add later on accident.
-    for key in ("confinement", "grade"):
-        if key not in config_data:
-            with contextlib.suppress(KeyError):
-                default = schema[key]["default"]
-                config_data[key] = default
-                logger.warn(
-                    "{!r} property not specified: defaulting to {!r}".format(
-                        key, default
-                    )
-                )
+    if "confinement" not in config_data:
+        default = schema["confinement"]["default"]
+        config_data["confinement"] = default
+        logger.warning(
+            f"'confinement' property not specified: defaulting to {default!r}"
+        )
 
     # Set default adapter
     app_schema = schema["apps"]["patternProperties"]["^[a-zA-Z0-9](?:-?[a-zA-Z0-9])*$"][
@@ -313,7 +309,6 @@ class _SnapPackaging:
         self._parts_dir = project_config.project.parts_dir
 
         self._arch_triplet = project_config.project.arch_triplet
-        self._global_state_file = project_config.project._get_global_state_file_path()
         self._is_host_compatible_with_base = (
             project_config.project.is_host_compatible_with_base
         )
--- a/snapcraft/internal/meta/errors.py
+++ b/snapcraft/internal/meta/errors.py
@@ -198,3 +198,14 @@ class HookValidationError(errors.SnapcraftError):
 
     def __init__(self, *, hook_name: str, message: str) -> None:
         super().__init__(hook_name=hook_name, message=message)
+
+
+class GradeDevelRequiredError(errors.SnapcraftException):
+    def __init__(self, *, set_grade: str) -> None:
+        self.set_grade = set_grade
+
+    def get_brief(self) -> str:
+        return f"Snap 'grade' was set to {self.set_grade!r} but must be 'devel'."
+
+    def get_resolution(self) -> str:
+        return "Set 'grade' to 'devel' or use a stable base for this snap."
--- a/snapcraft/internal/meta/snap.py
+++ b/snapcraft/internal/meta/snap.py
@@ -19,7 +19,7 @@ import os
 
 from collections import OrderedDict
 from copy import deepcopy
-from snapcraft.internal import common
+from snapcraft.internal import common, repo
 from snapcraft.internal.meta import errors
 from snapcraft.internal.meta.application import Application
 from snapcraft.internal.meta.hooks import Hook
@@ -27,13 +27,12 @@ from snapcraft.internal.meta.plugs import ContentPlug, Plug
 from snapcraft.internal.meta.slots import ContentSlot, Slot
 
 from snapcraft import yaml_utils
-from typing import Dict, Set, Sequence, List, Any
+from typing import Dict, Set, Sequence, List, Any, Optional
 
 logger = logging.getLogger(__name__)
 
 
 _MANDATORY_PACKAGE_KEYS = ["name", "version", "summary", "description"]
-
 _OPTIONAL_PACKAGE_KEYS = [
     "apps",
     "architectures",
@@ -78,6 +77,8 @@ class Snap:
         self.hooks: Dict[str, Hook] = dict()
         self.passthrough: Dict[str, Any] = dict()
         self.layout: Dict[str, Any] = dict()
+
+        self.__required_grade: Optional[str] = None
 
     @classmethod
     def from_file(cls, snap_yaml_path: str) -> "Snap":
@@ -127,6 +128,28 @@ class Snap:
 
         return provider_dirs
 
+    def _get_required_grade(self) -> str:
+        # Take into account loading from snap.yaml, those can be baseless (core),
+        # or the case of working with types of snap that do not require a base.
+        if self.base is None:
+            return "stable"
+
+        if self.__required_grade is not None:
+            return self.__required_grade
+
+        base_snap_pkg = repo.snaps.SnapPackage(self.base)
+        if base_snap_pkg.get_local_snap_info()["channel"] == "stable":
+            self.__required_grade = "stable"
+        else:
+            self.__required_grade = "devel"
+
+        return self.__required_grade
+
+    def _validate_grade(self) -> None:
+        """Verify that the grade is properly set."""
+        if self.grade == "stable" and self._get_required_grade() == "devel":
+            raise errors.GradeDevelRequiredError(set_grade=self.grade)
+
     def _validate_required_keys(self) -> None:
         """Verify that all mandatory keys have been satisfied."""
         missing_keys: List[str] = []
@@ -143,6 +166,7 @@ class Snap:
     def validate(self) -> None:
         """Validate snap, raising exception on error."""
         self._validate_required_keys()
+        self._validate_grade()
 
         for app in self.apps.values():
             app.validate()
@@ -219,8 +243,8 @@ class Snap:
         self._ensure_command_chain_assumption()
 
         for key in _MANDATORY_PACKAGE_KEYS + _OPTIONAL_PACKAGE_KEYS:
-            # Skip unset fields.
-            if self.__dict__[key] is None:
+            # Skip unset fields except the ones we can update with defaults.
+            if self.__dict__[key] is None and key != "grade":
                 continue
 
             # Skip fields that are empty lists/dicts.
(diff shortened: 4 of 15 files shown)


schema/snapcraft.json
```diff
@@ -922,6 +915,19 @@
                     "required": [
                         "build-base"
                     ]
+                },
+                {
+                    "usage": "type: gadget (with a base)",
```
cjp256: This section looks redundant now with:
`"usage": "base: <base> and type: <app|gadget>",`

snapcraft/internal/lifecycle/_runner.py
```diff
@@ -83,6 +84,12 @@ def execute(
         )
     else:
         installed_snaps = repo.snaps.install_snaps(required_snaps)
+        # We do not want snaps built from a non stable risk to be automatically
+        # set with a stable grade as this base might still suffer ABI or API
+        # changes.
+        base_snap_pkg = repo.snaps.SnapPackage(project_config.project.info.base)
+        if "stable" not in base_snap_pkg.get_local_snap_info()["channel"]:
+            recommended_grade = "devmode"
```
cjp256: s/devmode/devel ?

snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -244,12 +249,12 @@ def _desktop_file_exists(app_name: str) -> bool:
...
 
 
-def _update_yaml_with_defaults(config_data, schema):
+def _update_yaml_with_defaults(config_data, schema, global_state_file_path: str):
     # Ensure that grade and confinement have their defaults applied, if
     # necessary. Defaults are taken from the schema. Technically these are the
     # only two optional keywords currently WITH defaults, but we don't want to
     # risk setting something that we add later on accident.
-    for key in ("confinement", "grade"):
+    for key in "confinement":
```
cjp256: For loop seems unnecessary now...

snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -260,6 +265,23 @@ def _update_yaml_with_defaults(config_data, schema):
...
+    if "grade" not in config_data:
+        config_data["grade"] = recommended_grade
+        logger.warning(
+            "'grade' property not specified: defaulting to {!r}.".format(
+                recommended_grade
+            )
+        )
+    elif config_data["grade"] == "stable" and recommended_grade != "stable":
+        logger.warning(
+            "The  recommended 'grade' for this snap is 'devmode' as it relies "
```
cjp256: devmode v. devel

cjp256: Perhaps simplify to `f"The  recommended 'grade' for this snap is {recommended_grade!r} as it relies on a base that is not available on the stable channel."`

snapcraft/internal/repo/snaps.py
```diff
@@ -270,14 +275,25 @@ def download_snaps(*, snaps_list: Sequence[str], directory: str) -> None:
...
     """Install snaps of the format <snap-name>/<channel>.
 
     :return: a list of "name=revision" for the snaps installed.
     """
     snaps_installed = []
     for snap in snaps_list:
         snap_pkg = SnapPackage(snap)
+
+        # Allow bases to be installed from non stable channels.
+        snap_pkg_channel = snap_pkg.get_store_snap_info()["channel"]
```
cjp256: Does this work if the snap is not installed?  Which "channel" would be the default provided? stable, if available?

cjp256:
Hrmm, just checked the schema and the default is already set to stable there (but perhaps should be devel?)

snapcraft/internal/lifecycle/_runner.py
sergiusens: I always confuse those, good catch

sergiusens:
@cjp256 default should still be "stable", "devel" only for this case.

schema/snapcraft.json
sergiusens: True!

xnox:
I am confused why i wouldn't be publishing core20 based snaps to stable channels?

Yes it's not out yet, but we will need to publish all the things into stable (20/stable) prior to April release. And build all Ubuntu Core images out of stable prior to April release.

I.e. Ubuntu development series uses "-proposed => -release", why would I not be using "edge => stable" for the core20 snap, and other snaps that use 20/stable tracks?

I'd rather snapcraft enforce distro_series to be focal.

snapcraft/internal/lifecycle/_runner.py
sergiusens: Changed

snapcraft/internal/meta/_snap_packaging.py
sergiusens: done

schema/snapcraft.json
sergiusens: done

snapcraft/internal/meta/_snap_packaging.py
sergiusens: This gives no indication of why we are showing this message though (the fact that it was forced to stable in snapcraft.yaml). I will use it anyways and let you think on it :-)

snapcraft/internal/repo/snaps.py
sergiusens: This is information from the store, not local

sergiusens: Here's an illustration for various snaps and the last one showing the usual workflow when adding an invalid snap:
```
In [2]: from snapcraft.internal.repo import snaps                                                          

In [3]: snaps.SnapPackage("lxd").get_store_snap_info()["channel"]                                          
Out[3]: 'stable'

In [4]: snaps.SnapPackage("core20").get_store_snap_info()["channel"]                                       
Out[4]: 'edge'

In [5]: snaps.SnapPackage("core18").get_store_snap_info()["channel"]                                       
Out[5]: 'stable'

In [6]: snaps.SnapPackage("snapcraft").get_store_snap_info()["channel"]                                    
Out[6]: 'stable'

In [7]: snaps.SnapPackage("foo").get_store_snap_info()["channel"]                                          
---------------------------------------------------------------------------
HTTPError                                 Traceback (most recent call last)
~/snapcraft/snapcraft/internal/repo/snaps.py in get_store_snap_info(self)
    113                 try:
--> 114                     self._store_snap_info = _get_store_snap_info(self.name)
    115                     break

~/snapcraft/snapcraft/internal/repo/snaps.py in _get_store_snap_info(snap_name)
    385     snap_info = requests_unixsocket.get(url)
--> 386     snap_info.raise_for_status()
    387     return snap_info.json()["result"][0]

~/.venv/snapcraft/lib/python3.6/site-packages/requests/models.py in raise_for_status(self)
    939         if http_error_msg:
--> 940             raise HTTPError(http_error_msg, response=self)
    941 

HTTPError: 404 Client Error: Not Found for url: http+unix://%2Frun%2Fsnapd.socket/v2/find?name=foo

SnapUnavailableError                      Traceback (most recent call last)
<ipython-input-7-b9705a1c80c5> in <module>
----> 1 snaps.SnapPackage("foo").get_store_snap_info()["channel"]

~/snapcraft/snapcraft/internal/repo/snaps.py in get_store_snap_info(self)
    123                     if http_error.response.status_code == 404:
    124                         raise errors.SnapUnavailableError(
--> 125                             snap_name=self.name, snap_channel=self.channel
    126                         )
    127                     retry_count -= 1

SnapUnavailableError: Failed to install or refresh a snap: 'foo' does not exist or is not available on the desired channel 'latest/stable'. Use `snap info foo` to get a list of channels the snap is available on.
```

sergiusens:
@xnox it is overridable thing, a warning, defaulting to "grade: devel" only if not set so the causal developer does not accidentally use base: core20 BEFORE release only to find out someone removed libraries from underneath. In addition to that, the snapd runtime does not install a base UNLESS it is on stable, so you will need to manually install the base yourself, which from a user experience point of view is somewhat broken (but ok for a developer).

snapcraft/internal/meta/snap.py
```diff
@@ -53,6 +52,36 @@
...
+    elif not is_base_from_non_stable and user_set_grade is not None:
+        grade = user_set_grade
+    elif is_base_from_non_stable and user_set_grade is None:
+        logger.warning(
+            "'grade' property not specified: defaulting to 'devel' as "
+            "it relies on a base that is not available on the stable channel."
+        )
+        grade = "devel"
+    elif is_base_from_non_stable and user_set_grade == "stable":
+        logger.warning(
```
sergiusens: @cjp256 this is the case for when it might be appropriate to fail.

cjp256: What about `validate()`?

snapcraft/internal/meta/snap.py
```diff
@@ -210,6 +239,13 @@ def from_dict(cls, snap_dict: Dict[str, Any]) -> "Snap":
         if "adopt-info" in snap_dict:
             snap.adopt_info = snap_dict["adopt-info"]
 
+        # Set grade to "stable" unless building from an non-stable base.
+        # We do not want snaps built from a non stable risk to have a stable
+        # grade as this base might still suffer ABI or API changes.
+        snap.grade = _get_grade_for_snap(
```
cjp256: What do you think about:
1. set `grade: str = "stable"` in `__init__()` so the default is always set (if one were to create a Snap not from dict)
2. which then lets allows grade to be moved into `_MANDATORY_PACKAGE_KEYS`
3. decoupling validation/warning logic into `validate_grade()`, called by `validate()`
4. with that decoupled, it's then something like:
```
if "grade" not in snap_dict:
    snap.grade = _get_base_grade(base)
```

sergiusens: we should have a conversation again, I don't think this is possible

snapcraft/internal/meta/snap.py
```diff
@@ -127,6 +128,21 @@ def get_provider_content_directories(self) -> Set[str]:
 
         return provider_dirs
 
+    def _get_required_grade(self) -> str:
+        if self.__required_grade is not None:
+            return self.__required_grade
+
+        if self.base is None:
```
cjp256: is it possible base is None, if project is built without a base?

sergiusens: hmm, correct, this can happen for type: kernel, base and snapd now... for apps and gadget now, you would be reexecuting into the old legacy code base for that

snapcraft/internal/meta/snap.py
```diff
@@ -140,7 +156,13 @@ def _validate_required_keys(self) -> None:
         if missing_keys:
             raise errors.MissingSnapcraftYamlKeysError(keys=missing_keys)
 
-    def validate(self) -> None:
+        required_grade = self._get_required_grade()
```
cjp256: Mind just bunching this into `validate_grade()`?

snapcraft/internal/meta/snap.py
```diff
@@ -259,6 +281,15 @@ def to_dict(self):  # noqa: C901
                 snap_dict[key] = dict()
                 for name, slot in sorted(self.slots.items()):
                     snap_dict[key][name] = slot.to_dict()
+            elif key == "grade":
+                if self.grade is None:
```
cjp256: I think it'd be cleanest if you took this section and just chunked it into a `@property` for `grade`.  Then there wouldn't be any special casing in `to_dict()`.  And it would just return the required grade if unset.

snapcraft/internal/meta/snap.py
sergiusens: sure, public method or private though?

cjp256: Ah, private (whatever is consistent)

snapcraft/internal/meta/errors.py
```diff
@@ -198,3 +198,17 @@ class HookValidationError(errors.SnapcraftError):
 
     def __init__(self, *, hook_name: str, message: str) -> None:
         super().__init__(hook_name=hook_name, message=message)
+
+
+class RequiredGradeIncompatibleError(errors.SnapcraftException):
```
cjp256: Loving the use of SnapcraftException ;)

snapcraft/internal/meta/snap.py
cjp256: You're right, that's gonna break the __dict__... pass on this

sergiusens: I don't think this is doable with the current solution, I am planning on changing this dict dance a bit, let's see how that plays out

snapcraft/internal/meta/snap.py
xnox: Hmmmm, type:kernel should be declaring at least build-base. When it includes an initrd, and binaries come from ~= matching build-base.

sergiusens: kernel's do not have bases though, the build-base is just to determine the build environment

snapcraft/internal/meta/snap.py
```diff
@@ -140,9 +163,10 @@ def _validate_required_keys(self) -> None:
         if missing_keys:
             raise errors.MissingSnapcraftYamlKeysError(keys=missing_keys)
 
-    def validate(self) -> None:
+    def validate(self, *, required_grade: Optional[str] = None) -> None:
```
cjp256: drop `required_grade: Optional[str] = None)`

tests/unit/meta/test_errors.py
```diff
@@ -168,3 +168,29 @@ def test_error_formatting(self):
         self.assertThat(
             str(self.exception(**self.kwargs)), Equals(self.expected_message)
         )
+
```
cjp256: accidental paste?

tests/unit/meta/test_errors.py
```diff
@@ -168,3 +168,29 @@ def test_error_formatting(self):
...
+
+class SnapcraftExceptionTests(unit.TestCase):
+
+    scenarios = (
+        (
+            "StrangeExceptionSimple",
```
cjp256: nit: `GradeDevelRequiredError` ?

PR #2762, opened 2019-10-22
appstream: extract title and version

galgalesh:

-----

This PR extracts `title` and `version` from AppStream's `name` tag and
from the `version` property of the first `release` tag.

According to the AppStream docs:

> `<name/>`: A human-readable name for this software component.

> Each release of the software component should have a `<release/>`
> tag describing it, but at least one release child must be present
> for the current release of the software. **The release children
> should be sorted in a latest-to-oldest order to simplify reading
> the metadata file.**

[source](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-releases)

This PR assumes the AppStream follows the spec in that the first `<release>` element describes the latest release. It's possible to check the `date` and `timestamp` properties of each release to find the latest, however, parsing the dates is non-trivial:

> The date property can have any time in ISO 8601 format as its value and should be present for every release. At least day-level granularity is required, which means that the ISO 8601 string must contain at least a full date (e.g. 2019-08-12).

As a result, we can't simply use `strptime` because we don't know what the granularity of the date is beforehand. `python-dateutil` could be used to parse this date, however, I didn't want to add another dependency just to support AppStream files who don't follow the spec.

I looked at a bunch of AppStream files in the wild and they all seem to follow the spec.

* [GNOME Calculator](https://gitlab.gnome.org/GNOME/gnome-calculator/blob/master/data/org.gnome.Calculator.appdata.xml.in)
* [Foliate](https://github.com/johnfactotum/foliate/blob/master/data/com.github.johnfactotum.Foliate.appdata.xml.in)
* [Krita](https://invent.kde.org/kde/krita/blob/master/krita/org.kde.krita.appdata.xml)
* [Nautilus](https://github.com/GNOME/nautilus/blob/mainline/data/org.gnome.Nautilus.appdata.xml.in.in)

I also tested this PR manually with the Foliate snaps: https://github.com/snapcrafters/foliate/pull/4

### DIFF
--- a/snapcraft/extractors/_metadata.py
+++ b/snapcraft/extractors/_metadata.py
@@ -28,6 +28,7 @@ class ExtractedMetadata(yaml_utils.SnapcraftYAMLObject):
         self,
         *,
         common_id: str = "",
+        title: str = "",
         summary: str = "",
         description: str = "",
         version: str = "",
@@ -39,6 +40,7 @@ class ExtractedMetadata(yaml_utils.SnapcraftYAMLObject):
 
         :param str: common_id: The common identifier across multiple packaging
             formats
+        :param str title: Extracted title
         :param str summary: Extracted summary
         :param str description: Extracted description
         :param str version: Extracted version
@@ -52,6 +54,8 @@ class ExtractedMetadata(yaml_utils.SnapcraftYAMLObject):
 
         if common_id:
             self._data["common_id"] = common_id
+        if title:
+            self._data["title"] = title
         if summary:
             self._data["summary"] = summary
         if description:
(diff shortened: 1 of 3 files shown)


snapcraft/extractors/_metadata.py
```diff
@@ -39,6 +40,7 @@ def __init__(
 
         :param str: common_id: The common identifier across multiple packaging
             formats
+        :param str icon: Extracted title
```
sergiusens: s/icon/title/

galgalesh:
@sergiusens 

From what I understand, the Snapcraft code is already generic enough to use any metadata you extract. Specifically, the `_adopt_info` and `_adopt_keys` functions in `_snap_packaging.py` copy every key from the `extracted_metadata` to the top-level metadata. Building the Foliate snap with this PR seems to confirm that.

(It does require you to use `adopt-info` in your `snapcraft.yaml` though)

PR #2764, opened 2019-10-23
Snapd doesn't actually honor the slot specified in the default-provider.

kenvandine:
The default-provider should just be the content snap name.

-----


cjp256:
On a related note, with the changes already in the next version of snapcraft (available on edge), that slot name is never written to meta/snap.yaml. :)  But @sergiusens is going to backport to 3.8.x anyhow, and the cleanup is good IMO.  But there is a static test failure (run `black` on the sources)

PR #2766, opened 2019-10-23
project: truncate project directory hash

cjp256:
A full sha1 hash is admittedly a bit excessive, truncate
it to 6 characters.  Security is irrelevant and collisions
are not a realistic concern for these use cases.

Shorten it before it proliferates beyond remote-build :)

-----

### DIFF
--- a/snapcraft/project/_project.py
+++ b/snapcraft/project/_project.py
@@ -70,7 +70,7 @@ class Project(ProjectOptions):
     def _get_project_directory_hash(self) -> str:
         m = hashlib.sha1()
         m.update(self._project_dir.encode())
-        return m.hexdigest()
+        return m.hexdigest()[:6]
 
     def _get_content_snaps(self) -> Set[str]:
         """Return the set of content snaps from snap_meta."""


snapcraft/project/_project.py
```diff
@@ -70,7 +70,7 @@ def __init__(
     def _get_project_directory_hash(self) -> str:
         m = hashlib.sha1()
         m.update(self._project_dir.encode())
-        return m.hexdigest()
+        return m.hexdigest()[:6]
```
sergiusens: is 6 a good number? why not 8?

cjp256: I did eight at first, then I looked at it.  I figured with the format snapcraft-<project>-<id>, it's getting kinda long.  

Given that 6 chars (3 bytes) is something like 16million values...  I imagine it's something like winning the lottery if you can find a collision of project paths on the same system :)

As a reference point, I looked at git and they use 9 chars for short-form.

Either way, I'm thinking there will be some component for system-wide state in the upcoming config work that could be used to track pointers to all projects under active development (i.e. not cleaned).  That could (should) be utilized to guarantee collisions won't happen.

I'll go with whatever number you think is best though, no problemo :)

PR #2767, opened 2019-10-24
rust plugin: add rustup profile

dalance:
On some environment, some components can't be downloaded by rustup.
It causes the following error on build.snapcraft.io:

error: some components unavailable for download for channel stable:
'clippy', 'rustfmt'

To fix it, I added _get_rustup_profile method to adjust profile.
If profile is minimal, 'clippy' and 'rustfmt' will be skipped.

-----


protometa:
Somewhat related, that error also happens for me when the `RUSTUP_HOME` env var is not set to the default `~/.rustup`. [It's being set to `~/.cargo` in this plugin.](https://github.com/snapcore/snapcraft/blob/master/snapcraft/plugins/rust.py#L234)

cjp256:
Thanks for the report and PR @protometa.

I managed to reproduce this issue and have a few questions:

(1) Is it because there are no "stable" rustup components for clippy, etc.?  I see that they build nightlies, but it seems to be hard to find out what's available in their stable channel without running that arch...

(2) is there really a need for profile=default when building a snap?  is clippy/rustfmt ever invoked on behalf of the plugin?  It might make sense to just set it to minimal across the board to support the greatest common denominator (rather than supporting mixed-modes and allowing builds to fail for specific architectures due to a missing component).

(3) In the future, I'd expect "default" profile to be supported on newer rustup versions.  At that point, we could set it to "default" profile based on a version number?

Thanks!

cjp256:
Related launchpad bug: https://bugs.launchpad.net/snapcraft/+bug/1852545

protometa:
I made a note on the launchpad bug that I was getting this in `--destructive-mode` probably due to the unexpected `RUSTUP_HOME` default value.

I'm no Rust expert, but I think the minimal profile would actually be best across the board for the plugin. Documentation says the minimal profile is best for CI builds because it doesn't include extra documentation, linting, or formatting tools. I don't think those are ever needed in snapcraft builds. Even if setting the env var to the expected value fixes my issue, we probably want minimal profile regardless. I think we should probably avoid the default profile even in the future.

cjp256:
Related PR to split: https://github.com/snapcore/snapcraft/pull/2849/files

cjp256:
Can this PR be updated by chance to just set the profile to minimal?  (and fixup a few strings in tests/unit/plugins/test_rust.py)?

Thanks!

dalance:
I changed all profiles to minimal, and fixed test.

PR #2769, opened 2019-10-26
extensions: skip icon cache creation for theme and runtime snaps 

galgalesh:

-----

Some icon themes, such as cursor-only themes, can't have a cache. However, desktop-exports still tries to create a cache for every icon theme, which is surprisingly expensive. It takes up to 30% of the time of a cold desktop-launch run. This is partly due to the script creating additional directories and softlinks during this process.

This commit skips icon cache creation for everything in the theme and runtime snaps because those should already contain caches.

With this commit, the cold start time of the desktop-launch script is reduced by ~30%.

The KDE Neon apps also have an improved boot time with this fix, although it's less pronounced.

| App | without fix | with fix | reduction |
| --- | --- | --- | --- |
| [SDLPoP](https://github.com/galgalesh/sdlpop/tree/gnome-extension) | 1.146s | 0.675s | 40% |
| [Foliate](https://github.com/snapcrafters/foliate) | 0.848s | 0.619s | 27% |
| [KCalc](https://github.com/galgalesh/kcalc) | 2.980s | 2.673s | 10% |

Full benchmark results: https://gist.github.com/galgalesh/c6ff9b261668c0ce41f06661822c9fc3

### DIFF
--- a/extensions/desktop/common/desktop-exports
+++ b/extensions/desktop/common/desktop-exports
@@ -393,6 +393,11 @@ if [ "$needs_update" = true ]; then
   rm -rf "$XDG_DATA_HOME/icons"
   mkdir -p "$XDG_DATA_HOME/icons"
   for ((i = 0; i < ${#data_dirs_array[@]}; i++)); do
+    # The runtime and theme content snaps should already contain icon caches
+    # so we skip them to optimise app start time.
+    if [[ "${data_dirs_array[$i]}" == "$SNAP/data-dir" || "${data_dirs_array[$i]}" == "$SNAP_DESKTOP_RUNTIME/"* ]]; then
+      continue
+    fi  
     for theme in "${data_dirs_array[$i]}"/icons/*; do
       if [ -f "$theme/index.theme" ] && [ ! -f "$theme/icon-theme.cache" ]; then
         theme_dir="$XDG_DATA_HOME/icons/$(basename "$theme")"


galgalesh:
@cjp256 

> Out of curiosity, is there a way that we could programmatically determine which themes won't yield caches (safely)?

There isn't an efficient way, afaik. You could try to figure out if the theme contains any cache-able icons, but `/snap/gtk-common-themes/current/share/icons/hicolor/`, for example, appears to have an entire directory tree without any actual icons, so you'd have to traverse the entire tree in order to figure out that it doesn't need caching.

I don't know if that's a bug in that theme or if it's supposed to be that way. Maybe @jhenstridge can comment on this?

jhenstridge:
So if I understand the problem, correctly, the cases where this triggers are `$datadir/themes/name` directories that contain an `index.theme` file but no image files that `gtk-update-icon-theme` recognises.  And this is common for cursor themes, which usually contain only Xcursor format image files.

A better long term fix would be to get rid of the runtime gtk-update-icon-theme calls in the wrapper script completely.  If the extension could ensure that gtk-update-icon-cache was called on all the relevant primed icon directories at snap creation time, then there would be no need to check things on first run.

From the look of it, this would require additional changes on the snapcraft extension system though, so I agree that it is worth doing something else in the short term.  Given the nature of the extension, I wonder if we could do something a little less invasive though.

We know that icon caches have been built for gtk-common-themes and the platform snap.  We also know what paths that content will be mounted under.  So we should be able to safely skip over those directories in the `desktop-launch` script.

galgalesh:
Excluding the icon themes from the platform snap seems easy. However, excluding the icon themes from `gtk-common-themes` is more tricky because multiple snaps can provide icons in `$datadir/icons/`.

1. Excluding `$datadir/icons` entirely risks excluding icon themes from other content snaps that don't provide the caches. (This will become more relevant when the "greedy" interfaces stuff is implemented)
2. Generating the exclusion list at build-time risks having an exclusion list that is out of sync with `gtk-common-themes`. Any icon theme added to `gtk-common-themes` after an application snap is built will not be excluded.
3. Generating the exclusion list at run-time seems tricky to me. I'm not aware of any way a snap can query which directories originate from which content snap. Moreover, we need to make sure that this query code doesn't add a significant overhead to the startup of the snap.

As far as I know, the icon cache is not actually required. It just makes the icon theme more efficient. So, in a sense, option `1.` does not pose a significant risk. Moreover, I don't think it's a stretch to assume all snaps providing icon themes will include caches for those themes.

@jhenstridge  Do you see any issues with just excluding `$datadir/icons` entirely?

jhenstridge:
An icon theme will still work without an `icon-theme.cache` file, so I wouldn't worry about trying to distinguish which directories come from `gtk-common-themes` and which ones don't.

If someone publishes an icon theme snap that doesn't bake in an `icon-theme.cache` file, I think it would be better to solve that through linting tools or developer advocacy rather than have the desktop-launch script try to generate one at runtime.

galgalesh:
I changed the PR according to @jhenstridge's comments and retested the performance improvements.

kenvandine:
@jhenstridge can you give this another review?

PR #2770, opened 2019-10-26
extensions: speed up desktop-launch by not calling mkdir -p

galgalesh:

-----

Calling `mkdir -p` is relatively expensive because it starts an external process. This commit uses bash to check if the directory exists to avoid calling `mkdir -p` altogether in a hot start. This saves ~20% execution time in a hot start and has no significant impact on a cold start.

As an example; the hot startup time of [the SDLPoP snap using `gtk-3-28-1804`](https://github.com/galgalesh/sdlpop/tree/gnome-extension) with fix:

```
0.069s /snap/core/7917/usr/lib/snapd/snap-confine
0.008s /usr/bin/realpath
0.008s /usr/bin/realpath
0.008s /usr/bin/realpath
0.008s /usr/bin/realpath
0.008s /usr/bin/realpath
0.008s /usr/bin/realpath
0.009s /usr/bin/realpath
0.008s /bin/rm
0.147s /snap/sdlpop/x1/snap/command-chain/desktop-launch
```

without fix:

```
0.071s /snap/core/7917/usr/lib/snapd/snap-confine
0.009s /bin/mkdir
0.009s /bin/mkdir
0.009s /bin/mkdir
0.008s /usr/bin/realpath
0.008s /usr/bin/realpath
0.008s /usr/bin/realpath
0.008s /usr/bin/realpath
0.009s /bin/mkdir
0.183s /snap/sdlpop/x1/snap/command-chain/desktop-launch
```

### DIFF
--- a/extensions/desktop/common/init
+++ b/extensions/desktop/common/init
@@ -6,6 +6,15 @@
 # shellcheck disable=SC2034
 START=$(date +%s.%N)
 
+# ensure_dir_exists calls `mkdir -p` if the given path is not a directory.
+# This speeds up execution time by avoiding unnecessary calls to mkdir.
+#
+# Usage: ensure_dir_exists <path> [<mkdir-options>]...
+#
+function ensure_dir_exists() {
+  [ -d "$1" ] ||  mkdir -p "$@"
+}
+
 declare -A PIDS
 function async_exec() {
   "$@" &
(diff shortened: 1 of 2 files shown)


extensions/desktop/common/init
```diff
@@ -6,6 +6,10 @@
 # shellcheck disable=SC2034
 START=$(date +%s.%N)
 
+function ensure_dir_exists() {
+  [ -d "$1" ] ||  mkdir -p "$@"
```
cjp256: Very simple, nice!

Beware that `$1` vs `$@` behavior is different.

I would suggest either allowing it to take multiple args (to match mkdir), or require just one.  A quick grep looks like all of the current use cases are single directory.

galgalesh: I use `$@` so that this function can pass additional flags to mkdir. This is needed for the following line.

https://github.com/snapcore/snapcraft/blob/45e6b956cf8f9904f7994bb5492c4aec97a82604/extensions/desktop/common/desktop-exports#L136

This means that the first argument to `ensure_dir_exists` must always be the path, and the following arguments are not allowed to be paths. Is it ok to just document this in a comment above the function?

PR #2771, opened 2019-10-28
python plugin: first try processing setup.py without PyPI

sergiusens:
Try to install setup.py with dependencies satisfied from:
- python-packages
- requirements

If that fails, resort to using PyPI to satisfy missing install_requires
from setup.py

LP: #1841861

-----

### DIFF
--- a/snapcraft/plugins/_python/_pip.py
+++ b/snapcraft/plugins/_python/_pip.py
@@ -54,7 +54,7 @@ def _process_package_args(
     *,
     packages: Optional[Sequence[str]],
     requirements: Optional[Sequence[str]],
-    setup_py_dir: Optional[str]
+    setup_py_dir: Optional[str],
 ) -> List[str]:
     args = []
     if requirements:
@@ -243,7 +243,7 @@ class Pip:
         setup_py_dir: Optional[str] = None,
         constraints: Optional[Set[str]] = None,
         requirements: Optional[Sequence[str]] = None,
-        process_dependency_links: bool = False
+        process_dependency_links: bool = False,
     ):
         """Download packages into cache, but do not install them.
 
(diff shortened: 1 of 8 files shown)


basak:
This appears to work with Adrien's reproducer. Thanks!

I wasn't able to check it against certbot proper, because they're not currently affected by the issue. They will be when they next add a feature to acme that they need from certbot, at which my daily builds will fail and I'll be able to test this against the certbot snap build.

But it looks like you have tests in this branch that will provide enough confidence in the fix anyway?

bmw:
Hi. I'm a developer of Certbot.

We've deliberately avoided merging https://github.com/certbot/certbot/pull/7442 which will create the issue this resolves until our release on Tuesday. If it would be useful, we can merge the PR.

adferrand:
@rbasak, @sergiusens, if you want a second check besides my showcase, Certbot repository is currently in the situation where the bug occurs and this PR could fix it. If you want I can also open a branch from master that create on purpose dev dependencies to keep a real world case available.

bmw:
What needs to happen for this PR to be merged?

We were wanting to move forward with snaps as the primary packaging solution for our project and this is slowing us down quite a bit.

sergiusens:
Hi there, we just need to run some more tests and clean up a bit to merge this, that could happen this week.

adferrand:
By the way @sergiusens, I think this PR will have a side effect that is beneficial in my opinion: pip will reuse the existing packages downloaded and compiled from the previous parts, instead of do it for all of them in each part, then merge afterwards.

It will save some bandwidth when there are several dependencies that are reused in several part, and even more important, will save computation time by not recompiling them.

sergiusens:
@cjp256 correct, this is what I added in the commit comment. I am not decided yet on how to satisfy both camps, but this one is the technically correct one whilst the current behavior is a bad one, but it is also the expected one for existing users.

PR #2774, opened 2019-10-28
remote-build: architecture handling

cjp256:


sergiusens:
Just took notice that we have no unit tests for the remote-build cli though.

cjp256:
> Just took notice that we have no unit tests for the remote-build cli though.

Yeah, there are still some gaps in remote-build on my list along with the test_errors bit...  I will be back with a follow up PR to introduce tests for all of the CLI remote-build stuff once the dust settles here. Sound good?

PR #2780, opened 2019-10-30
remote-build: option renames for arch, user, and accept-public-upload

cjp256:


cjp256:
This will break https://github.com/snapcore/snapcraft/pull/2775.

If accepted, merge this first and I'll rebase 2775.

sergiusens:
How about we use the language we have for architectures and call it "--build-on" or "--build-on-architectures"?

cjp256:
> How about we use the language we have for architectures and call it "--build-on" or "--build-on-architectures"?

Sounds good, `build-on` it is.

PR #2782, opened 2019-10-30
snapcraft: introduce click-based YAML configuration file support

cjp256:

-----

### DIFF
--- a/snapcraft/cli/lifecycle.py
+++ b/snapcraft/cli/lifecycle.py
@@ -22,6 +22,7 @@ import click
 
 from . import echo
 from ._command import SnapcraftProjectCommand
+from ._config import enable_snapcraft_config_file
 from ._options import add_build_options, get_build_environment, get_project
 from snapcraft.internal import (
     build_providers,
@@ -151,6 +152,7 @@ def lifecyclecli(ctx, **kwargs):
 
 
 @lifecyclecli.command()
+@enable_snapcraft_config_file()
 def init():
     """Initialize a snapcraft project."""
     snapcraft_yaml_path = lifecycle.init()
@@ -162,6 +164,7 @@ def init():
 
 
 @lifecyclecli.command(cls=SnapcraftProjectCommand)
+@enable_snapcraft_config_file()
 @click.pass_context
 @add_build_options()
 @click.argument("parts", nargs=-1, metavar="<part>...", required=False)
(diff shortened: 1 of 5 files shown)


cjp256:
Requires #2781  and #2780

cjp256:
Rebased.

This PR enables it for remote-build as a limited test case.

I tested it with:
echo -e "launchpad-user: cjp256\nlaunchpad-accept-public-upload: true" > ~/.config/snapcraft/config.yaml

Then running:
`snapcraft remote-build` without args works without prompts.

cjp256:
On a related note, there is an option of utilizing Click's `auto_envvar_prefix` for command groups to enable coverage of environment variable mappings to parameters. https://click.palletsprojects.com/en/7.x/options/#values-from-environment-variables

It would incorporate the command name, e.g. `remote-build` for mapping environment variables, e.g. `SNAPCRAFT_REMOTE_BUILD_LAUNCHPAD_USER=cjp256`.

With the Click wrapper, `_click_load_config()`, we could rely on cmd_name to produce similar behavior if there is some foreseen conflicts among parameter names between commands.  Presumably we would short circuit the lifecycleci commands as the common case.

snapcraft/cli/remote.py
```diff
@@ -22,6 +22,7 @@
 from typing import List
 from xdg import BaseDirectory
 from . import echo
+from .config import enable_snapcraft_config_file
```
sergiusens: why isn't this from ._config ?

sergiusens: It is not meant to be used outside of the cli package, right?

cjp256: Correct, changed to _config.

cjp256:
Enabled config file for all commands except legacy...

cjp256:
Added info that config file is being used.  Added warning that config file is experimental.

```
$ snapcraft remote-build
Using config file: /home/ubuntu/.config/snapcraft/config.yaml
Configuration file is experimental and is subject to change.
snapcraft remote-build is experimental and is subject to change - use with caution.
```

cjp256:
Rebased

cjp256:
Accidentally forced push wrong branch over this. Fixed.

cjp256:
Scaled it back to just lifecycle and remote commands.

snapcraft/cli/lifecycle.py
```diff
@@ -262,6 +270,7 @@ def snap(directory, output, **kwargs):
 
 
 @lifecyclecli.command(cls=SnapcraftProjectCommand)
+@enable_snapcraft_config_file()
```
sergiusens: pack shouldn't need this

cjp256: Removed.

cjp256:
rebased/squashed

PR #2783, opened 2019-10-30
cli: run review tools before pushing to the store if available

sergiusens:
Consider only snap v2 lint warnings and errors. On discovery of any issues,
emit warnings, but do not fail as we have no easy way to override
any of these issues.

-----

### DIFF
--- a/snapcraft/cli/store.py
+++ b/snapcraft/cli/store.py
@@ -13,6 +13,7 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
 import contextlib
 import os
 import functools
@@ -29,6 +30,7 @@ import snapcraft
 from snapcraft import storeapi, formatting_utils
 from snapcraft.storeapi.constants import DEFAULT_SERIES
 from . import echo
+from ._review import review_snap
 
 
 _MESSAGE_REGISTER_PRIVATE = dedent(
@@ -171,6 +173,7 @@ def push(snap_file, release):
     else:
         channel_list = None
 
+    review_snap(snap_file=snap_file)
     snapcraft.push(snap_file, channel_list)
 
 
--- a/snapcraft/internal/review_tools/_runner.py
+++ b/snapcraft/internal/review_tools/_runner.py
@@ -1,0 +1,49 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2019 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import json
+import pathlib
+import subprocess
+
+from . import errors
+
+
+_REVIEW_TOOLS_PATH = pathlib.Path("/snap/bin/review-tools.snap-review")
+
+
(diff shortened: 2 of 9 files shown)


snapcraft/cli/store.py
```diff
@@ -171,6 +172,24 @@ def push(snap_file, release):
...
+        echo.info(
+            "Running the review tools before pushing this snap to the Snap " "Store."
+        )
+        # TODO just raise when we can override.
+        try:
+            review_tools.run(snap_filename=snap_file)
+        except review_tools.errors.ReviewError as review_error:
+            echo.warning(review_error.get_brief())
+            echo.warning(review_error.get_resolution())
+            print(review_error.get_details())
```
cjp256: s/print/echo.warning?

snapcraft/cli/store.py
```diff
@@ -171,6 +172,24 @@ def push(snap_file, release):
     else:
         channel_list = None
 
+    # Review the snap.
```
cjp256: Might be best to pull this out into its own function, then it could be re-used for a possible `review` command future maybe?

snapcraft/internal/review_tools/_runner.py
```diff
@@ -0,0 +1,50 @@
...
+from . import errors
+
+
+def is_available() -> bool:
+    review_tools_bin = pathlib.Path("/snap/bin/review-tools.snap-review")
+    return review_tools_bin.exists()
+
+
+def run(*, snap_filename: str) -> None:
+    review_tools_bin = pathlib.Path("/snap/bin/review-tools.snap-review")
```
cjp256: Feels slightly redundant, but it's safe :)

snapcraft/internal/review_tools/_runner.py
```diff
@@ -0,0 +1,50 @@
...
+        subprocess.check_output(command, stderr=subprocess.STDOUT, env=env)
+    except subprocess.CalledProcessError as call_error:
+        # Error codes:
+        # 2 -> warnings.
+        # 3 -> warnings and errors.
+        # Any of these trigger a store review.
+        if call_error.returncode not in (2, 3):
+            raise
+        # We only really care about v2_lint now.
+        raise errors.ReviewError(review_json=json.loads(call_error.output.decode()))
```
cjp256: I assume the presence of v2_lint is guaranteed with a return code of 2/3?

cjp256:
Maybe add a couple of tests?

snapcraft/cli/store.py
sergiusens: too much yellow!

snapcraft/internal/review_tools/_runner.py
sergiusens: yes, the payload is always there, @jdstand can confirm

snapcraft/cli/store.py
```diff
@@ -171,6 +172,24 @@ def push(snap_file, release):
...
+    if review_tools.is_available():
+        echo.info(
+            "Running the review tools before pushing this snap to the Snap " "Store."
```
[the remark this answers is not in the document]
sergiusens: black!

snapcraft/internal/review_tools/_runner.py
sergiusens: I did not want to spend time with a join as Path knows how to deal with this internally :-)

snapcraft/internal/review_tools/_runner.py
```diff
@@ -0,0 +1,50 @@
...
+
+def is_available() -> bool:
+    review_tools_bin = pathlib.Path("/snap/bin/review-tools.snap-review")
+    return review_tools_bin.exists()
+
+
+def run(*, snap_filename: str) -> None:
+    review_tools_bin = pathlib.Path("/snap/bin/review-tools.snap-review")
+    if not review_tools_bin.exists():
+        raise errors.ReviewToolNotAvailableError()
```
cjp256: FWIW, I think the proper way to do this is to catch FileNotFoundError on check_output()

cjp256: In fact, do that because `/snap/bin` may not be in path as we have seen before...

cjp256:
Lookin good

PR #2785, opened 2019-10-30
remote-build: add initial command unit tests

cjp256:
Add some tests test for `remote-build` and `build-on`
arch handling.

-----

### DIFF
--- a/tests/unit/commands/test_remote.py
+++ b/tests/unit/commands/test_remote.py
@@ -1,0 +1,90 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2019 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License remote_build 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from unittest import mock
+
+import fixtures
+from testtools.matchers import Contains, Equals
+
+import snapcraft.internal.remote_build.errors as errors
+from . import CommandBaseTestCase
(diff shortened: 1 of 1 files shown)


cjp256:
On a related note, there is an option of utilizing Click's `auto_envvar_prefix` for command groups to enable coverage of environment variable mappings to parameters. https://click.palletsprojects.com/en/7.x/options/#values-from-environment-variables

It would incorporate the command name, e.g. `remote-build` for mapping environment variables, e.g. `SNAPCRAFT_REMOTE_BUILD_LAUNCHPAD_USER=cjp256`.

With the Click wrapper, `_click_load_config()`, we could rely on cmd_name to produce similar behavior if there is some foreseen conflicts among parameter names between commands.  Presumably we would short circuit the lifecycleci commands as the common case.

tests/unit/commands/test_remote.py
```diff
@@ -0,0 +1,122 @@
...
+import snapcraft.internal.remote_build.errors as errors
+from . import CommandBaseTestCase
+from unittest import mock
+
+
+class RemoteBuildTests(CommandBaseTestCase):
+    def setUp(self):
+        super().setUp()
+
+    def create_project(self):
```
sergiusens: use the SnapcraftYaml fixture instead.

cjp256: Ah, that's right!

Fixed and rebased.

cjp256:
> Looks good, just use the fixture we have instead to make future changes easier to manage

Done and rebased!

tests/unit/commands/test_remote.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+class RemoteBuildTests(CommandBaseTestCase):
+    def setUp(self):
+        super().setUp()
+        self.useFixture(fixture_setup.SnapcraftYaml(self.path))
+
+    @mock.patch("snapcraft.cli.remote.LaunchpadClient")
+    def test_remote_build(self, mock_lc):
+        lc = mock_lc.return_value
```
sergiusens: why don't you convert this into a fixtues.MockPatch in setup with all the default values, then just apply the variations in the tests which will make obvious what the change for the individual test is.

This particular assignment is also very confusing for a casual read,

cjp256: Will do

cjp256: Done!

cjp256:
Resolved issues and rebased.

PR #2786, opened 2019-10-31
cli: add support for 'http-proxy' and 'https-proxy' parameters

cjp256:
Some modules support configuration of http(s) proxies today (e.g.
ant, maven, nodejs) - but they are only honoured if building
with host provider and the environment variables are already in
the environment.

Add two explicit build options for http_proxy and https_proxy,
with matching environment variables to support typical application
behavior.

Only add these to the environment if provider is one of 'lxd',
'multipass', 'host', or 'managed-host'. Otherwise, either:
(a) warn if found in the user's environment.
(b) error if not found in environment, and therefor must have
    been configured with an explicit command-line option.

Add tests for coverage of get_build_environment_flags() with
new http_proxy and https_proxy handling.

### DIFF
--- a/snapcraft/cli/_options.py
+++ b/snapcraft/cli/_options.py
@@ -21,7 +21,6 @@ import click
 from snapcraft.project import Project, get_snapcraft_yaml
 from snapcraft.cli.echo import confirm, prompt
 from snapcraft.internal import common, errors
-from typing import Tuple
 
 
 class PromptOption(click.Option):
@@ -116,7 +115,7 @@ def add_provider_options(hidden=False):
     return _add_provider_options
 
 
-def _sanity_check_build_environment_flags(**kwargs):
+def _sanity_check_build_provider_flags(**kwargs):
     provider = kwargs.get("provider")
     use_lxd = kwargs.get("use_lxd")
     destructive_mode = kwargs.get("destructive_mode")
@@ -156,14 +155,12 @@ def _sanity_check_build_environment_flags(**kwargs):
         )
 
 
-def get_build_environment(
-    skip_sanity_checks: bool = False, **kwargs
-) -> Tuple[str, bool]:
+def get_build_provider(skip_sanity_checks: bool = False, **kwargs) -> str:
     """Get build provider and determine if running as managed instance."""
 
     # Sanity checks may be skipped for the purpose of checking legacy.
     if not skip_sanity_checks:
-        _sanity_check_build_environment_flags(**kwargs)
+        _sanity_check_build_provider_flags(**kwargs)
 
     provider = kwargs.get("provider")
 
@@ -178,14 +175,7 @@ def get_build_environment(
             # Default is multipass.
             provider = "multipass"
 
-    # Handle special case for managed-host.
-    if provider == "managed-host":
-        provider = "host"
-        is_managed_host = True
-    else:
-        is_managed_host = False
-
-    return provider, is_managed_host
+    return provider
 
 
 def get_project(*, is_managed_host: bool = False, **kwargs):
--- a/snapcraft/cli/_options.py
+++ b/snapcraft/cli/_options.py
@@ -115,7 +115,7 @@ def add_provider_options(hidden=False):
     return _add_provider_options
 
 
-def _sanity_check_build_provider_flags(**kwargs):
+def _sanity_check_build_provider_flags(**kwargs) -> None:
     provider = kwargs.get("provider")
     use_lxd = kwargs.get("use_lxd")
     destructive_mode = kwargs.get("destructive_mode")
--- a/snapcraft/cli/_options.py
+++ b/snapcraft/cli/_options.py
@@ -15,6 +15,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import os
+import sys
 
 import click
 
@@ -68,16 +69,19 @@ _BUILD_OPTIONS = [
 
 _SUPPORTED_PROVIDERS = ["host", "lxd", "multipass"]
 _HIDDEN_PROVIDERS = ["managed-host"]
+_ALL_PROVIDERS = _SUPPORTED_PROVIDERS + _HIDDEN_PROVIDERS
 _PROVIDER_OPTIONS = [
     dict(
         param_decls="--destructive-mode",
         is_flag=True,
         help="Forces snapcraft to try and use the current host to build (implies `--provider=host`).",
+        supported_providers=["host", "managed-host"],
     ),
     dict(
         param_decls="--use-lxd",
         is_flag=True,
         help="Forces snapcraft to use LXD to build (implies `--provider=lxd`).",
+        supported_providers=["lxd"],
     ),
     dict(
         param_decls="--provider",
@@ -85,7 +89,8 @@ _PROVIDER_OPTIONS = [
         show_envvar=False,
         help="Build provider to use.",
         metavar="[{}]".format("|".join(_SUPPORTED_PROVIDERS)),
-        type=click.Choice(_SUPPORTED_PROVIDERS + _HIDDEN_PROVIDERS),
+        type=click.Choice(_ALL_PROVIDERS),
+        supported_providers=_ALL_PROVIDERS,
     ),
 ]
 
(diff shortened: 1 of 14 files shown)


cjp256:
While some tests need to be updated pending initial review, it seems to work.

The most significant side effect I encounter is due to the introduction of `sudo -i env <...>` as part of common _run.  `sudo -i` exec'd on LXD will print some cruft like `mesg: ttyname failed: No such device`, due to `mesg n` in `/root/.profile`.

I could add a property for the provider for it to suggest the default user, and decide whether or not `sudo` is needed?  Or just leave that part to the multipass-specific `_run_prepared_command()`...?

cjp256:
> Couple of things.
> 
> Why the rename from _run to _run_prepared? Just call it run, _run was already "preparing" anyways. Also, you probably want to use this for shell as well.
> 
> Keep sudo multipass specific or add a method in the provider to check everytime.

_run is now common doing the extra prep, so `shell` inherits that.  _run_prepared_command could be something like _run_provider()?

Thinking about it some more maybe I'll just keep the run as-is and just update each to call _wrap_command_with_environment.

cjp256:
This should largely be complete now, but I do want to include three spread tests to verify lxd, multipass, host.  I'll add those Monday morning.

cjp256:
> Couple of things.
> 
> Why the rename from _run to _run_prepared? Just call it run, _run was already "preparing" anyways. Also, you probably want to use this for shell as well.
> 
> Keep sudo multipass specific or add a method in the provider to check everytime.

Done.

snapcraft/cli/_options.py
```diff
@@ -121,6 +133,29 @@ def get_build_environment_flags(provider: str, **kwargs) -> Dict[str, str]:
...
+        if key not in kwargs:
+            continue
+
+        value = kwargs[key]
+        if not value:
+            continue
+
+        if provider in ["host", "managed-host", "lxd", "multipass"]:
+            provider_env[key] = value
+        elif not os.getenv(key):
```
sergiusens: if this is the case, why not make the http-proxy and https-proxy params hidden?

snapcraft/cli/_options.py
```diff
@@ -110,6 +111,19 @@ def get_build_environment(**kwargs):
     return env.BuilderEnvironmentConfig(force_provider=provider)
 
 
+def apply_host_build_environment_flags(build_environment_flags: Dict[str, str]) -> None:
```
sergiusens: should this not live in cli.env?

sergiusens: it can probably even be a method from the class that lives there, same for get_build_environment_flags

tests/spread/build-providers/host-env-passthrough/task.yaml
```diff
@@ -0,0 +1,27 @@
...
+  SNAP_DIR: ../snaps/env-passthrough
+
+prepare: |
+  #shellcheck source=tests/spread/tools/snapcraft-yaml.sh
+  . "$TOOLS_DIR/snapcraft-yaml.sh"
+  set_base "$SNAP_DIR/snap/snapcraft.yaml"
+
+restore: |
+  cd "$SNAP_DIR"
+  snapcraft clean --use-lxd
```
sergiusens: this clean should omit the --use-lxd

sergiusens: Change it to --destructive-mode

snapcraft/internal/build_providers/_base_provider.py
```diff
@@ -172,7 +172,9 @@ def destroy(self) -> None:
         """
 
     @abc.abstractmethod
-    def _run(self, command: Sequence[str]) -> Optional[bytes]:
+    def _run(
+        self, command: Sequence[str], hide_output: bool = False
```
sergiusens: strange that implementing classes did not barf on this

snapcraft/cli/_options.py
cjp256: I'm indifferent, but this is really an artifact that "host" is a special case provider and does not get a provider class.  Ideally we can refactor the provider bits to support launchpad and host as well...

My avoidance of cli.env / BuilderEnvironmentConfig is that it's blown away in: https://github.com/snapcore/snapcraft/pull/2678/files

Should I put it in env anyways?

tests/spread/build-providers/host-env-passthrough/task.yaml
cjp256: Ah good catch.

snapcraft/internal/build_providers/_base_provider.py
cjp256: Yes I thought so as well...

snapcraft/cli/_options.py
cjp256: They can certainly be hidden.  Though I'm not sure I understand the purpose of making it harder to find valid options though...

I would consider the cases of hiding options to be good when feature is very experimental or deprecated.

In the case of http_proxy and https_proxy, I don't know that it is either.

If the concern is that it may flooding the user with advanced, unlikely to use options, I would suggest either:
(1) introduce a new "provider-specific options:" list below the standard options in `help`, or
(2) effectively group them by prefacing the names with `provider`, e.g. `--provider-http-proxy`.

tests/spread/build-providers/host-env-passthrough/task.yaml
cjp256: Fixed

cjp256: Fixed

snapcraft/cli/_options.py
sergiusens: this comment is outdated from reading the code top-down and then bottom-up. It might be easier to grasp the objective if instead of a big loop some list comprehension is done to reduce the sets of what is trying to be accomplished and assigning to nice variable names.

Alternatively, more code comments can help.

snapcraft/cli/_options.py
sergiusens: ah, forgot about that PR. Keep it here for now.

cjp256:
Requires #2678

sergiusens:
This needs conflict resolution

sergiusens:
I am regretting agreeing to 82df8a451d3c5edfc7f8e2a9817e2f306ca43bd1 ... let's punt it for later

cjp256:
> We should be using the features of the backend to our benefit, maybe instead keep using what LXD has to offer [acf46ce#diff-ce8f136dec24256964db704d798d83cdR184](https://github.com/snapcore/snapcraft/commit/acf46ce71ebcacb93d3e2a814668212db046860b#diff-ce8f136dec24256964db704d798d83cdR184)
> 
> and apply the env loading only to multipass.

I considered it both ways, but is there anything to be gained over the `env` method?  Utilizing the backend-specfic features does require another code path to maintain/test... I would prefer a common solution unless there are measurable gains to be had.  The `env` method at least guarantees `_run()`-time configuration without adding container environment config synchronization AFAICT.

cjp256:
> I am regretting agreeing to [82df8a4](https://github.com/snapcore/snapcraft/commit/82df8a451d3c5edfc7f8e2a9817e2f306ca43bd1) ... let's punt it for later

Can you please clarify what it is that you do not like? :man_facepalming: Do you want to go back to a common set of options that have a defined set of supported providers?  The change is not insignificant, so I want to be sure we sort this out first :slightly_smiling_face:

cjp256:
Single set of options, with a set of `supported_providers`.  I like this much better :D

Should be complete, but hasn't been tested too extensively yet.  This wasn't a good "merge" option, so let's reset with this respin and review forward...

snapcraft/cli/_options.py
```diff
@@ -148,10 +149,10 @@ def _sanity_check_build_provider_flags(build_provider: str, **kwargs) -> None:
     # change defaults, so they are safe to ignore due to filtering
     # in get_build_provider_flags().
     for option in _PROVIDER_OPTIONS:
-        key = option["param_decls"]
-        supported_providers = option["supported_providers"]
-        unsupported = build_provider not in supported_providers  # type: ignore
-        if key in sys.argv and unsupported:
+        key: str = option["param_decls"]  # type: ignore
```
sergiusens: why annotate and then ignore?

PR #2791, opened 2019-11-04
introduce bind-ssh support

cjp256:
Introduce support for binding user's .ssh to local LXD and Multipass build environments.

Depends on #2786 and #2078

### DIFF
--- a/snapcraft/internal/build_providers/_base_provider.py
+++ b/snapcraft/internal/build_providers/_base_provider.py
@@ -193,6 +193,14 @@ class Provider(abc.ABC):
         """Push a file into the instance."""
 
     @abc.abstractmethod
+    def _is_mounted(self, target: str) -> bool:
+        """Query if there is a mount at target mount point."""
+
+    @abc.abstractmethod
+    def _mount(self, host_source: str, target: str) -> None:
+        """Mount host source directory to target mount point."""
+
+    @abc.abstractmethod
     def mount_project(self) -> None:
         """Provider steps needed to make the project available to the instance.
         """
--- a/snapcraft/internal/build_providers/_base_provider.py
+++ b/snapcraft/internal/build_providers/_base_provider.py
@@ -16,6 +16,7 @@
 
 import abc
 import os
+import pathlib
 import logging
 import shlex
 import shutil
@@ -126,6 +127,8 @@ class Provider(abc.ABC):
         if build_provider_flags is None:
             build_provider_flags = dict()
         self.build_provider_flags = build_provider_flags.copy()
+
+        self._cached_home_directory: pathlib.Path = None
 
     def __enter__(self):
         try:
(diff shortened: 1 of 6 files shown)


cjp256:
Should we mark this as experimental?

snapcraft/internal/build_providers/_base_provider.py
```diff
@@ -126,6 +127,7 @@ def __init__(
         if environment_passthrough is None:
             environment_passthrough = dict()
         self.environment_passthrough = environment_passthrough
+        self._cached_home_directory: pathlib.Path = None
```
sergiusens: :100:

sergiusens:
Can you split 9c77cb69e8ef368327963f76f38078d99881abddin two? One without changing existing signatures and another with signature changes/concentration of code?

sergiusens:
It feels like we need to rename this base provider to LocalBaseProvider (to make room for the remote one) considering moving most of the only local things into this class.

sergiusens:
Does this indeed work? It would be interesting to see (albeit might not be possible today), to mount this path read only

cjp256:
> Can you split [9c77cb6](https://github.com/snapcore/snapcraft/commit/9c77cb69e8ef368327963f76f38078d99881abdd) two? One without changing existing signatures and another with signature changes/concentration of code?

Absolutely

cjp256:
> Does this indeed work? It would be interesting to see (albeit might not be possible today), to mount this path read only

Yeah I tested in LXD and multipass.  A read-only option would be good, but might cause grief with known_hosts...

cjp256:
Requires https://github.com/snapcore/snapcraft/pull/2678

cjp256:
Depends on #2786.

snapcraft/internal/build_providers/_base_provider.py
```diff
@@ -190,16 +193,39 @@ def _push_file(self, *, source: str, destination: str) -> None:
         """Push a file into the instance."""
 
     @abc.abstractmethod
+    def is_mounted(self, target: str) -> bool:
```
sergiusens: the model we decided to follow for this base class is that if we do not intend to use these methods outside of the components inheriting from it, they should be private/protected (start with `_`)

snapcraft/internal/build_providers/_base_provider.py
```diff
@@ -190,16 +193,39 @@ def _push_file(self, *, source: str, destination: str) -> None:
...
+        """Query if there is a mount at target mount point."""
+
+    @abc.abstractmethod
+    def mount(self, host_source: str, target: str) -> None:
```
sergiusens: the model we decided to follow for this base class is that if we do not intend to use these methods outside of the components inheriting from it, they should be private/protected (start with `_`)

cjp256:
Added commit: https://github.com/snapcore/snapcraft/pull/2791/commits/005119a9bb406215ee946db3f6f501197f6dba26

Force push diff:
https://github.com/snapcore/snapcraft/compare/259bfcf17e95110dbaee955b2aeaebb91c55b6e2..857851bf47e198211be5c474437b2f4eabdf99a1

Should be good to go.

PR #2794, opened 2019-11-06
Add gnome 3 34 extension

hellsworth:
  - would reformat snapcraft/plugins/dotnet.py. I didn't make this change since I know the CI doesn't use the latest version of black.
  - yes, there are still 5 failing tests I'll need to look at

-----

### DIFF
--- a/snapcraft/internal/project_loader/_extensions/_utils.py
+++ b/snapcraft/internal/project_loader/_extensions/_utils.py
@@ -175,8 +175,18 @@ def _apply_extension(
         parts[part_name] = part_definition
 
 
-def _apply_extension_property(existing_property: Any, extension_property: Any) -> dict[str, Any]:
-    if existing_property:
+def _apply_extension_property(existing_property: Any, extension_property: Any) -> Dict[str, Any]:
+    """Take the user-defined yaml properties and apply any missing
+    extension properties. If there is a property defined in both, 
+    the user-defined property is applied.
+    """
+
+    # If there is no user-defined property, then just add the extension-defined property. 
+    if not existing_property:
+        return extension_property
+
+    # If there is a user-defined property, then take care when merging with the extension-defined property.
+    elif existing_property:
 
         if isinstance(existing_property, list) and isinstance(extension_property, list):
            
@@ -215,9 +225,7 @@ def _apply_extension_property(existing_property: Any, extension_property: Any) -
                 existing_property[key] = _apply_extension_property(
                     existing_property.get(key), value
                 )
-            return existing_property
         return existing_property
-    return extension_property
 
 
 def _merge_lists(list1: List[str], list2: List[str]) -> List[str]:
--- a/snapcraft/internal/project_loader/_extensions/_utils.py
+++ b/snapcraft/internal/project_loader/_extensions/_utils.py
@@ -51,7 +51,6 @@ def apply_extensions(yaml_data: Dict[str, Any]) -> Dict[str, Any]:
     yaml_data = copy.deepcopy(yaml_data)
     original_yaml_data = copy.deepcopy(yaml_data)
     base: Optional[str] = yaml_data.get("base")
-    #base = yaml_data.get("base")
 
     # Mapping of extension names to set of app names to which the extension needs to be
     # applied.
@@ -247,7 +246,7 @@ def _merge_lists(list1: List[str], list2: List[str]) -> List[str]:
     seen = set()  # type: Set[str]
     merged = list()  # type: List[str]
 
-    for item in list1+list2:
+    for item in list1 + list2:
         if item not in seen:
             seen.add(item)
             merged.append(item)
--- a/snapcraft/internal/project_loader/_extensions/_utils.py
+++ b/snapcraft/internal/project_loader/_extensions/_utils.py
@@ -15,6 +15,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import collections
+from collections import OrderedDict
 import contextlib
 import copy
 import jsonschema
@@ -130,6 +131,41 @@ def _load_extension(
     )
 
 
+def _convert_to_ordereddict(
+    property_list: List[Dict[str, str]]
+) -> OrderedDict:
+    od = OrderedDict()
+    for elem in property_list:
+        if not isinstance(elem, dict) or len(elem) != 1:
+            raise RuntimeError("Badly disguised ordered dictionary!")
+        for k, v in elem.items():
+            od[k] = v
+    return od
+
+
+def _convert_to_disguised_ordereddict(
+    od: OrderedDict
+) -> List[Dict[str, str]]:
+    disguised: List[Dict[str, str]] = list()
+    for k, v in od.items():
+        disguised.append({k: v})
+    return disguised
+
+
+def _merge_build_environment(
+    existing: List[Dict[str, str]], extension: List[Dict[str, str]]
+) -> List[Dict[str, str]]:
+    # build-environment is an OrderedDict in disguise.
+    if not existing:
+        return extension
+    if not extension:
+        return existing
+    existing_dict = _convert_to_ordereddict(existing)
+    extension_dict = _convert_to_ordereddict(extension)
+    extension_dict.update(existing_dict)
+    return _convert_to_disguised_ordereddict(extension_dict)
+
+
 def _apply_extension(
     yaml_data: Dict[str, Any],
     app_names: Set[str],
@@ -158,9 +194,14 @@ def _apply_extension(
     for part_name, part_definition in parts.items():
 
         for property_name, property_value in part_extension.items():
-            part_definition[property_name] = _apply_extension_property(
-                part_definition.get(property_name), property_value
-            )
+            if property_name == "build-environment":
+                part_definition[property_name] = _merge_build_environment(
+                    part_definition.get(property_name), property_value
+                )
+            else:
+                part_definition[property_name] = _apply_extension_property(
+                    part_definition.get(property_name), property_value
+                )
 
         # Stores the extension's list of part_snippets in each part
         parts[part_name] = part_definition
--- a/snapcraft/internal/project_loader/_extensions/_utils.py
+++ b/snapcraft/internal/project_loader/_extensions/_utils.py
@@ -68,7 +68,7 @@ def apply_extensions(yaml_data: Dict[str, Any]) -> Dict[str, Any]:
         with contextlib.suppress(KeyError):
             del yaml_data["apps"][app_name]["extensions"]
 
-    # Process extensions in a consistent order
+    # Process extensions in a consistent order.
     for extension_name in sorted(declared_extensions.keys()):
         extension = _load_extension(base, extension_name, original_yaml_data)
         _apply_extension(
@@ -112,7 +112,7 @@ def supported_extension_names() -> List[str]:
 
     extension_names = []  # type: List[str]
     for _, modname, _ in pkgutil.iter_modules([os.path.dirname(__file__)]):
-        # Only add non-private modules/packages to the extension list
+        # Only add non-private modules/packages to the extension list.
         if not modname.startswith("_"):
             extension_names.append(modname)
 
@@ -172,14 +172,14 @@ def _apply_extension(
     extension_name: str,
     extension: Extension,
 ):
-    # Apply the root components of the extension (if any)
+    # Apply the root components of the extension (if any).
     root_extension = extension.root_snippet
     for property_name, property_value in root_extension.items():
         yaml_data[property_name] = _apply_extension_property(
             yaml_data.get(property_name), property_value
         )
 
-    # Apply the app-specific components of the extension (if any)
+    # Apply the app-specific components of the extension (if any).
     app_extension = extension.app_snippet
     for app_name in app_names:
         app_definition = yaml_data["apps"][app_name]
@@ -188,7 +188,7 @@ def _apply_extension(
                 app_definition.get(property_name), property_value
             )
 
-    # Next, apply the part-specific components
+    # Next, apply the part-specific components.
     part_extension = extension.part_snippet
     parts = yaml_data["parts"]
     for part_name, part_definition in parts.items():
@@ -203,16 +203,16 @@ def _apply_extension(
                     part_definition.get(property_name), property_value
                 )
 
-        # Stores the extension's list of part_snippets in each part
+        # Stores the extension's list of part_snippets in each part.
         parts[part_name] = part_definition
 
-    # Finally, add any parts specified in the extension
+    # Finally, add any parts specified in the extension.
     for part_name, part_definition in extension.parts.items():
         # If a extension part name clashes with a part that already exists, error.
         if part_name in parts:
             raise errors.ExtensionPartConflictError(extension_name, part_name)
 
-        # Stores the extension's list of parts in the parts section
+        # Stores the extension's list of parts in the parts section.
         parts[part_name] = part_definition
 
 
--- a/snapcraft/internal/project_loader/_extensions/_utils.py
+++ b/snapcraft/internal/project_loader/_extensions/_utils.py
@@ -227,19 +227,17 @@ def _apply_extension_property(existing_property: Any, extension_property: Any) -
         return extension_property
 
     # If there is a user-defined property, then take care when merging with the extension-defined property.
-    elif existing_property:
+    if isinstance(existing_property, list) and isinstance(extension_property, list):
+        return _merge_lists(existing_property, extension_property)
 
-        if isinstance(existing_property, list) and isinstance(extension_property, list):
-            return _merge_lists(existing_property, extension_property)
-
-        elif isinstance(existing_property, dict) and isinstance(
-            extension_property, dict
-        ):
-            for key, value in extension_property.items():
-                existing_property[key] = _apply_extension_property(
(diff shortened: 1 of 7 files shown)


hellsworth:
There is an [issue with the way the build-environment is set](https://bugs.launchpad.net/snapcraft/+bug/1853040) that will need to be solved before this PR is merged.

hellsworth:
@cjp256 It would be great if we could create a temporary branch in the store to host a version of snapcraft that has the gnome-3-34 extension, until it's provided to the store in a latest channel. If there is a version of snapcraft with the extension in the store, then I can have launchpad auto builds of various snaps. Until then, I just have to build locally.

What do you think?

snapcraft/internal/project_loader/_extensions/_utils.py
```diff
@@ -156,25 +158,58 @@ def _apply_extension(
...
+        # Stores the extension's list of parts in the parts section
         parts[part_name] = part_definition
 
 
 def _apply_extension_property(existing_property: Any, extension_property: Any):
     if existing_property:
-        # If the property is not scalar, merge them
+
         if isinstance(existing_property, list) and isinstance(extension_property, list):
-            return _merge_lists(existing_property, extension_property)
```
cjp256: Could you provide some additional context / examples on what's being done here?

While outside the scope of this PR, I _really_ think we should reconsider the approach for extensions.  Why not operate on the YAML directly (and eventually, being a Snap object)? There's a lot of complexity around what amounts to a handful trivial operations on a dictionary and a base validation check. :disappointed:

cjp256:
> @cjp256 It would be great if we could create a temporary branch in the store to host a version of snapcraft that has the gnome-3-34 extension, until it's provided to the store in a latest channel. If there is a version of snapcraft with the extension in the store, then I can have launchpad auto builds of various snaps. Until then, I just have to build locally.
> 
> What do you think?

I think we should just get this into master :smiley:.  Let's issue a warning to the user indicating that this extension is still experimental/unsupported and forcing grade to "devel".  IMO, that'd be sufficient.  When you're comfortable the extension and the gnome snaps are ready for prime-time, we can remove those lines.

Related question: is gnome-3-34-1804 supposed to have a "stable" grade?

snapcraft/internal/project_loader/_extensions/_utils.py
```diff
@@ -156,25 +158,58 @@ def _apply_extension(
...
         # If a extension part name clashes with a part that already exists, error.
         if part_name in parts:
             raise errors.ExtensionPartConflictError(extension_name, part_name)
 
+        # Stores the extension's list of parts in the parts section
         parts[part_name] = part_definition
 
 
 def _apply_extension_property(existing_property: Any, extension_property: Any):
     if existing_property:
```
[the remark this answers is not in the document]
hellsworth: Ok yep I tried to simplify it a bit and added some clarifying comments. Let me know if you think this works.

snapcraft/internal/project_loader/_extensions/_utils.py
hellsworth: I added clarity and tried to not be too wordy so please take a look and let me know if that makes sense.

In the long term, we're planning to use the freedesktop-sdk and the corresponding gnome runtimes (1-2 years). So for now, I think this method of applying the extension could hold us over until then, but I'm open to changing things sooner if we feel the need to do so.

snapcraft/internal/project_loader/_extensions/gnome_3_34.py
```diff
@@ -0,0 +1,135 @@
...
+    At build time it ensures the right build dependencies are setup and for
+    the runtime it ensures the application is run in an environment catered
+    for GNOME applications.
+
+    It configures each application with the following plugs:
+
+    \b
+    - GTK3 Themes.
+    - Common Icon Themes.
+    - Common Sound Themes
```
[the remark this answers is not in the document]
hellsworth: done

snapcraft/internal/project_loader/_extensions/gnome_3_34.py
```diff
@@ -0,0 +1,135 @@
...
+            },
+        }
+
+        self.app_snippet = {
+            "command-chain": ["snap/command-chain/desktop-launch"],
+            "plugs": ["desktop", "desktop-legacy", "gsettings", "wayland", "x11"],
+        }
+
+        self.part_snippet = {
+            "build-environment": [
```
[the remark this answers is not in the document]
hellsworth: > It'd be great to see a list of apps this extension has been tested against?

I've tested gedit, gnome-calculator, evince, and I feel like there were others too.. One thing is I'd like to get a snapcraft in the store with this extension (by adding a gnome-3-34-extension branch or something) so I can have LP build many more apps.

snapcraft/internal/project_loader/_extensions/gnome_3_34.py
```diff
@@ -0,0 +1,135 @@
...
+        return ("core18",)
+
+    @staticmethod
+    def get_supported_confinement() -> Tuple[str, ...]:
+        return ("strict", "devmode")
+
+    def __init__(self, *, extension_name: str, yaml_data: Dict[str, Any]) -> None:
+        super().__init__(extension_name=extension_name, yaml_data=yaml_data)
+
+        platform_snap = _PLATFORM_SNAP[yaml_data.get("base")]
```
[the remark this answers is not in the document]
cjp256: To confirm, in conversation with @kenvandine, this is expected to be the case.

snapcraft/internal/project_loader/_extensions/_utils.py
cjp256: Thank you! I think I understand the problem now.  Everything is complicated because of the use of a list of dictionaries for `build-environment`, instead of a simple dict/object.

I expect the use of the list in the case of `build-environment` is to guarantee ordering.  The irony here of course is that each element in build-environment gets parsed into its own OrderedDict, but it's really effectively a combined OrderedDict.  If it were, then you could just use '.update()' to merge the dictionaries, e.g.:

```
>>> from collections import OrderedDict
>>> od = OrderedDict()
>>> od['a'] = 1
>>> od['b'] = 2
>>> od['c'] = 3
>>> od['d'] = 4
>>>
>>> od2 = OrderedDict()
>>> od2['e'] = 5
>>> od2['f'] = 6
>>> od2['a'] = 7
>>> od2['d'] = 8
>>>
>>> merged = od.copy()
>>> merged.update(od2)
>>> merged
OrderedDict([('a', 7), ('b', 2), ('c', 3), ('d', 8), ('e', 5), ('f', 6)])
```

Which I think satisfies what you're trying to do?


I pondered it for a bit and I would suggest something like:
https://github.com/cjp256/snapcraft/commit/576ccae204c5784e01b2907bc8698aed12ee8504

Namely, eliminate any assumptions that will come back to bite us later (special case handling only for build-environment rather than assuming what to do based on some incomplete type information and context available here).  And I have a preference towards using the OrderedDict update operation, because that's what the list really is :smile: 

But whatever you decide, can you add a few of unit tests to cover some scenarios regarding this merge functionality?  It's fairly complex, and deserves a bit of attention to ensure we're not going to introduce some regressions.

snapcraft/internal/project_loader/_extensions/_utils.py
```diff
@@ -152,44 +152,91 @@ def _apply_extension(
...
+            # defined variable. 
+            # See https://bugs.launchpad.net/snapcraft/+bug/1853040 for more info.
+            temp_list = []
+            seen = set()
+
+            # Safely copy the list of OrderedDicts to a list of dicts.
+            for item in existing_property:
+                if isinstance(item, collections.OrderedDict):
+                    temp_list.append(dict(item))
+                if len(temp_list) > 0:
```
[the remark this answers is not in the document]
hellsworth: >Is this indent correct?

Yes, this indent is correct. This for loop is supposed to only be triggered if we're dealing with an existing_property that is a list and an extension_property that is a list.

>I think you would lose data if existing_property was a list containing non-ordereddict.

I was just being cautious here... maybe one day we have to deal with a list with some combination of dicts and ordereddicts. I wanted all ordereddicts to be put into a dict, one item at a time. Alternatively, we could make the assumption that if there is a list whose first item is an ordereddict, then all of the lists items should be ordereddicts and replace this block with:

```
if isinstance(existing_property[0], collections.OrderedDict):
    for item in existing_property:
        temp_list.append(dict(item))
    if len(temp_list) > 0:
        existing_property = temp_list
```

Now, that being said, this block isn't needed at all if we go the route of making the build-environment an ordereddict like you have in [cjp256@576ccae](https://github.com/cjp256/snapcraft/commit/576ccae204c5784e01b2907bc8698aed12ee8504). I actually like your solution better because it reads better and now all lists are of ordereddicts (well, or strings) and we don't have a mixture of lists of ordered dicts and lists of dicts.

snapcraft/internal/project_loader/_extensions/_utils.py
```diff
@@ -152,44 +152,91 @@ def _apply_extension(
...
+                    existing_property = temp_list
+
+            # Keep track of user-defined items.
+            for item in existing_property:
+                if isinstance(item, collections.OrderedDict):
```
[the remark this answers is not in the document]
hellsworth: >Won't the above loop ensure that won't happen?
Ah good point.. I hadn't realized that but yes you're right.

Anyways, I've removed this block along with all other blocks for the case of merging two lists.

snapcraft/internal/project_loader/_extensions/_utils.py
```diff
@@ -152,44 +152,91 @@ def _apply_extension(
...
+    extension properties. If there is a property defined in both, 
+    the user-defined property is applied.
+    """
+
+    # If there is no user-defined property, then just add the extension-defined property. 
+    if not existing_property:
+        return extension_property
+
+    # If there is a user-defined property, then take care when merging with the extension-defined property.
+    elif existing_property:
```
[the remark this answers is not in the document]
hellsworth: I really left this elif, just for readability but you're right so I've removed it and shifted the block left.

snapcraft/internal/project_loader/_extensions/_utils.py
```diff
@@ -152,44 +152,91 @@ def _apply_extension(
...
+                if len(temp_list) > 0:
+                    existing_property = temp_list
+
+            # Keep track of user-defined items.
+            for item in existing_property:
+                if isinstance(item, collections.OrderedDict):
+                    for key, value in item.items():
+                        seen.add(key)
+                elif isinstance(item, str):
+                    return _merge_lists(existing_property, extension_property)
```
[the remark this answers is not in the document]
hellsworth: This section is now gone.

hellsworth:
Ah yes, I will add tests. Thanks!!

sergiusens:
@cjp256 @hellsworth  I am going to rather put this branch on the store for testing instead of landing on master as we might iterate on it a lot depending on what the other desktop peers find.

Additionally, while I do this, may I asked for the stylistic changes (sentence endings, whitespace additions, etc) be moved to a separate PR (or PRs)?

hellsworth:
Ok @sergiusens I believe I did what you asked: move the formatting commits to a different PR.

In detail though, here's what I did....

I went ahead and rebased on master and then had this topmost history:

history

I identified the commits that were really just white space/punctuation/formatting related - they are highlighted in blue. So I set this branch's HEAD at `1042eba6 Add spread test for gnome-3-34 extension` and then cherry-pick'd each subsequent non-blue commits and fixed merge conflicts along the way. That is why you'll see that the number of commits in this PR has been reduced from 31 to 24.

Then I created a new branch called cleanups and cherry-pick'd all of the blue commits. Then I merged this cleanups branch on my fork onto the add-gnome-3-34-extension branch on my fork (which is the proposed branch for this PR).

So all in all, the end result is unchanged but the formatting changes have been moved to other commits in a different PR.

snapcraft/internal/project_loader/_extensions/_utils.py
```diff
@@ -130,20 +131,51 @@ def _load_extension(
     )
 
 
+def _convert_to_ordereddict(property_list: List[Dict[str, str]]) -> OrderedDict:
+    od = collections.OrderedDict()
```
cjp256: should be type annotated with `od: OrderedDict = collections.OrderedDict()`

cjp256: also you'll have to rename `od` because the spell checker will complain... :(

snapcraft/internal/project_loader/_extensions/gnome_3_34.py
```diff
@@ -0,0 +1,135 @@
...
```
cjp256: the static checker is complaining because it's possible that yaml_data.get("base") returns a non-string (or "Any").

Given the logic to ensure that only supported bases are allowed and you have a complete dictionary above for the supported bases...  you have two OK options:

```
base = yaml_data.get("base")
assert isinstance(base, str)
```

or

```
base = yaml_data["base"]
platform_snap = _PLATFORM_SNAP[base]
```

The second one is a bit of a cheat, but I probably use it most commonly :wink:

hellsworth: I went with the second option so thanks!!

snapcraft/internal/project_loader/_extensions/_utils.py
hellsworth: thank you so much!!! I didn't see the spell checker error (and still don't) but i renamed it to `myod`- hope this is ok.

hellsworth:
I'm still finding and fixing the comparison errors in my added tests, but I've checked in some fixes. Should be finished with these test fixes tomorrow.

snapcraft/internal/project_loader/_extensions/gnome_3_34.py
```diff
@@ -0,0 +1,136 @@
...
+            },
+        }
+
+        self.app_snippet = {
+            "command-chain": ["snap/command-chain/desktop-launch"],
+            "plugs": ["desktop", "desktop-legacy", "gsettings", "wayland", "x11"],
+        }
+
+        self.part_snippet = {
+            "build-environment": [
```
sergiusens: why is this not part of the gnome extension part?

sergiusens: This would solve the merging issues, the core of snapcraft should do the right thing with "after" and build-environment entries.

hellsworth: Because we want the build-environment part snippet to be added to each part, while the gnome-3-34-extension is one whole part that should be added at the part level, and only once.

hellsworth: >This would solve the merging issues, the core of snapcraft should do the right thing with "after" and build-environment entries.

Because of the reason listed in my prev comment, I don't see a way to accomplish having the build-environment section added to each part if we combine the part_snippet and part sections.

tests/unit/project_loader/extensions/test_gnome_3_34.py
```diff
@@ -0,0 +1,418 @@
...
+
+        self.assertThat(
+            result.output,
+            Contains(
+                textwrap.dedent(
+                    """\
+                        parts:
+                          test-part-1:
+                            plugin: nil
+                            build-environment: &id001
```
cjp256: Hrmm, I wonder what this `&id001` is about...?

hellsworth: it's referencing the first build-environment section. Since we didn't give it a name, one was assigned to it: "id001".

A similar thing is done in [the build-snap](https://gitlab.gnome.org/Community/Ubuntu/gnome-sdk/blob/gnome-3-34-1804-sdk/snapcraft.yaml) where the builld-environment section is used in every part, but only explicitly defined in the first part.

hellsworth:
The build-environment behavior needs to be modified such that:
1.  a build-env section can have more than one key (same var) definition. For example, a build-environment section with two PATH lines should be allowed
2. if a key is defined by the user and by the extension, then just add this extension defined line
3. all of these lines are visible in expand extensions

So for example, a user could define two PATH lines in one build-environment section. The extension defines a PATH line too. So expand-extensions should show 3 PATH lines.

These changes will mean significant changes to the _utils logic as well as the relevant tests.

sergiusens:
test time

hellsworth:
Would you please add this snapcraft to the store in a temp branch? Than I'll setup a handful of launchpad tests. If those all pass, then can we consider merging this pr?

sergiusens:
Sorry for the delays, published on edge/pr-2794

This still seems to need review from @cjp256

hellsworth:
This snapcraft extension pulls the gnome-3-34-1804-sdk from the snap store and it was recently decided to add gtkmm to the build snap. So in order to make sure the extension (and it's build snap) are being tested in its final form, I'm working on adding gtkmm to the build snap before porting several snaps to use this extension. The testing of this extension will be completed when several snaps have been updated to use it. While it would be ideal to port all of [these snaps](https://wiki.ubuntu.com/DesktopTeam/GNOMESnaps) to complete the test effort of this snapcraft extension, a select set has been chosen: 
* gnome-characters (java)
* gnome-system-monitor (C++, also test gtkmm)
* eog (C)
* simple-scan (vala)
* evince
* gnome-calculator
* gedit

hellsworth:
Closing in favor of another PR to be made

PR #2795, opened 2019-11-07
manifest: track and annotate `primed-stage-packages`

cjp256:
Use extended attributes to implement a low-overhead file-origin
tracking solution.  As debs are extracted, they are marked with
an attribute that indicates their source origin.

Primed files that carry the marker are expected to have originated
from the source package that it was extracted from.

Add `xattr` version 0.9.6 to requirements.txt.

-----

### DIFF
--- a/snapcraft/internal/errors.py
+++ b/snapcraft/internal/errors.py
@@ -685,3 +685,26 @@ class SnapDataExtractionError(SnapcraftError):
 
 class ProjectNotFoundError(SnapcraftReportableError):
     fmt = "Failed to find project files."
+
+
+class XAttributeTooLongError(SnapcraftReportableException):
+    def __init__(self, *, key: str, value: str, path: str) -> None:
+        self._key = key
+        self._value = value
+        self._path = path
+
+    def get_brief(self) -> str:
+        return "Unable to write extended attribute as the key and/or value is too long."
+
+    def get_details(self) -> str:
+        return (
+            f"Failed to write attribute to {self._path}:\n"
+            f"key={self._key!r} value={self._value!r}"
+        )
+
+    def get_resolution(self) -> str:
+        return (
+            "This issue is generally resolved by addressing/truncating "
+            "the data source of the long data value. In some cases, the "
+            "filesystem being used will limit the allowable size."
+        )
--- a/snapcraft/internal/repo/_base.py
+++ b/snapcraft/internal/repo/_base.py
@@ -23,10 +23,11 @@ import os
 import re
 import shutil
 import stat
-from typing import List
+
+from typing import List, Set
 
 from snapcraft import file_utils
-from snapcraft.internal import mangling
+from snapcraft.internal import mangling, xattrs
 from . import errors
 
 
@@ -216,6 +217,23 @@ class BaseRepo:
         self._fix_artifacts(unpackdir)
         self._fix_xml_tools(unpackdir)
         self._fix_shebangs(unpackdir)
+
+    def _mark_origin_stage_package(
+        self, sources_dir: str, stage_package: str
+    ) -> Set[str]:
+        """Mark all files in sources_dir as coming from stage_package."""
+        file_list = set()
+        for (root, dirs, files) in os.walk(sources_dir):
+            for file_name in files:
+                file_path = os.path.join(root, file_name)
+
+                # Mark source.
+                xattrs.write_origin_stage_package(file_path, stage_package)
+
+                file_path = os.path.relpath(root, sources_dir)
+                file_list.add(file_path)
+
+        return file_list
 
     def _remove_useless_files(self, unpackdir):
         """Remove files that aren't useful or will clash with other parts."""
--- a/tests/unit/lifecycle/test_lifecycle.py
+++ b/tests/unit/lifecycle/test_lifecycle.py
@@ -607,17 +607,16 @@ class RecordManifestBaseTestCase(LifecycleTestBase):
         check_output_patcher.start()
         self.addCleanup(check_output_patcher.stop)
 
-        original_check_call = subprocess.check_call
-
-        def _fake_dpkg_deb(command, *args, **kwargs):
-            if "dpkg-deb" not in command:
-                return original_check_call(command, *args, **kwargs)
-
-        check_call_patcher = mock.patch(
-            "subprocess.check_call", side_effect=_fake_dpkg_deb
+        self.useFixture(
+            fixtures.MockPatch(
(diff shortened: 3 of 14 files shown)


snapcraft/internal/repo/_base.py
```diff
@@ -217,6 +222,49 @@ def normalize(self, unpackdir):
         self._fix_xml_tools(unpackdir)
         self._fix_shebangs(unpackdir)
 
+    def read_source_package_attr(self, path: str) -> Optional[str]:
```
sergiusens: Is there a reason for this to be public? I would make it private even if used in pluginhandler as the Repo class is allowed for use by plugins.

snapcraft/internal/repo/_base.py
```diff
@@ -217,6 +222,49 @@ def normalize(self, unpackdir):
...
+
+    def write_source_package_attr(self, path: str, source: str) -> None:
+        if sys.platform != "linux":
+            raise RuntimeError("xattr support only available for Linux")
+
+        # Extended attributes do not apply to symlinks.
+        if not os.path.islink(path):
+            xattr.setxattr(path, "user.snapcraft.source_package", source.encode())
+
+    def mark_source_package_contents(
```
sergiusens: Is there a reason for this to be public? I would make it private.

snapcraft/internal/repo/_base.py
```diff
@@ -217,6 +222,49 @@ def normalize(self, unpackdir):
         self._fix_xml_tools(unpackdir)
         self._fix_shebangs(unpackdir)
 
+    def read_source_package_attr(self, path: str) -> Optional[str]:
+        if sys.platform != "linux":
+            raise RuntimeError("xattr support only available for Linux")
+
+        # Extended attributes do not apply to symlinks.
+        if not os.path.islink(path):
+            return None
```
sergiusens: lamest nit of all, but mind adding a newline after this return call so it these early guards are visually separated from the rest of the logic?

snapcraft/internal/repo/_base.py
cjp256: I haven't quite grepped the rationale for public/private in snapcraft.  Generally I would expect private to be for internal class (or package) usage only... 

If called externally from outside class/package, I would expect that it should be public?  Then it's obvious you shouldn't modify it without considering impact to all (external) users.

Regardless, I was thinking it should be pulled out of repo.  I don't think it's best suited to live in repo, just put it here out of convenience in my first stab.  Maybe file_utils `read_snapcraft_xattrs(file_path)` and write_snapcraft_xattr(file_path, snapcraft_user_key, value)`?

snapcraft/internal/repo/_base.py
cjp256: agreed

snapcraft/internal/repo/_base.py
cjp256: will do

snapcraft/internal/repo/_base.py
cjp256: Pulled it out into `snapcraft.internal.xattrs`

cjp256:
I was thinking before this went gold, it might be useful to include the mapping of packages/files:

something like:
```
primed-stage-packages:
  - libyaml-0-2:amd64=0.1.7-2ubuntu3:
    /usr/lib/x86_64-linux-gnu/libyaml-0.so.2.0.5: /usr/lib/x86_64-linux-gnu/libyaml-0.so.2.0.5
  - something=0.0:
    /usr/bin/something: /usr/bin/something-renamed
    /usr/lib/x86_64-linux-gnu/libsomething-0.so.0.0: /usr/lib/x86_64-linux-gnu/libsomething-0.so.0.0
```

Then if the files are moved/organized, we know where they are if the manifest is present.

cjp256:
is there a formal schema for the manifest.yaml anywhere?

sergiusens:
@cjp256 we can in the future move to having objects if we do the review-tools part for it to be safe https://git.launchpad.net/review-tools/tree/reviewtools/store.py#n205

snapcraft/internal/errors.py
```diff
@@ -685,3 +685,26 @@ def __init__(self, snap):
 
 class ProjectNotFoundError(SnapcraftReportableError):
     fmt = "Failed to find project files."
+
+
+class XAttributeTooLongError(SnapcraftReportableException):
```
sergiusens: this should probably be a reportable error

tests/unit/lifecycle/test_lifecycle.py
```diff
@@ -607,17 +607,16 @@ def fake_uname(cmd, *args, **kwargs):
         check_output_patcher.start()
         self.addCleanup(check_output_patcher.stop)
 
-        original_check_call = subprocess.check_call
-
-        def _fake_dpkg_deb(command, *args, **kwargs):
-            if "dpkg-deb" not in command:
-                return original_check_call(command, *args, **kwargs)
+        self.mock_extract_deb_name_version = fixtures.MockPatch(
```
sergiusens: thought we agreed to call these fake_<foo>, in any case, you do not need to carry this in `self` if not going to be accessed in the tests, same for `self.mock_extract_deb` -> `fake_extract_deb`

tests/unit/pluginhandler/test_pluginhandler.py
```diff
@@ -976,6 +977,11 @@ def setUp(self):
         self.get_elf_files_mock.return_value = frozenset()
         self.addCleanup(patcher.stop)
 
+        self.mock_read_xattr = fixtures.MockPatch(
```
sergiusens: just change this to
```
self.useFixture(fixtures.MockPatch("snapcraft.internal.xattrs.read_origin_stage_package", return_value=None))
```

snapcraft/internal/errors.py
cjp256: It is SnapcraftReportableException...? Did you mean it shouldn't be?

tests/unit/lifecycle/test_lifecycle.py
cjp256: Agreed, Fixed!

tests/unit/pluginhandler/test_pluginhandler.py
cjp256: Agreed, Fixed!

snapcraft/internal/errors.py
sergiusens: oh, sorry, yes, this is good as is!

PR #2802, opened 2019-11-15
uprev devel python package requirements

cjp256:
Sync everything except mypy in requirements-devel.txt to the latest, making minor changes, as necessary.

### DIFF
--- a/runtests.sh
+++ b/runtests.sh
@@ -39,7 +39,7 @@ run_static_tests() {
     mypy .
 
     echo "Running codespell"
-    codespell -S "*.tar,*.xz,*.zip,*.bz2,*.7z,*.gz,*.deb,*.rpm,*.snap,*.gpg,*.pyc,*.png,*.ico,*.jar,changelog,.git,.hg,.mypy_cache,.tox,.venv,_build,buck-out,__pycache__,build,dist,.vscode,parts,stage,prime,test_appstream.py" -q4
+    codespell -S "*.tar,*.xz,*.zip,*.bz2,*.7z,*.gz,*.deb,*.rpm,*.snap,*.gpg,*.pyc,*.png,*.ico,*.jar,changelog,.git,.hg,.mypy_cache,.tox,.venv,_build,buck-out,__pycache__,build,dist,.vscode,parts,stage,prime,test_appstream.py,./snapcraft.spec" -q4
 
     echo "Running shellcheck"
     # Need to skip 'demos/gradle/gradlew' as it wasn't written by us and has
--- a/snapcraft/internal/project_loader/_config.py
+++ b/snapcraft/internal/project_loader/_config.py
@@ -21,7 +21,7 @@ import os.path
 import re
 
 import jsonschema
-from typing import Set  # noqa: F401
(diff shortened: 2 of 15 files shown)


snapcraft/internal/project_loader/_config.py
```diff
@@ -21,7 +21,7 @@
 import re
 
 import jsonschema
-from typing import Set  # noqa: F401
+from typing import List, Set  # noqa: F401
```
sergiusens: can you remove the noqa from here and make sure these are used explicitly?

runtests.sh
```diff
@@ -39,7 +39,7 @@ run_static_tests() {
     mypy .
 
     echo "Running codespell"
-    codespell -S "*.tar,*.xz,*.zip,*.bz2,*.7z,*.gz,*.deb,*.rpm,*.snap,*.gpg,*.pyc,*.png,*.ico,*.jar,changelog,.git,.hg,.mypy_cache,.tox,.venv,_build,buck-out,__pycache__,build,dist,.vscode,parts,stage,prime,test_appstream.py" -q4
+    codespell -S "*.tar,*.xz,*.zip,*.bz2,*.7z,*.gz,*.deb,*.rpm,*.snap,*.gpg,*.pyc,*.png,*.ico,*.jar,changelog,.git,.hg,.mypy_cache,.tox,.venv,_build,buck-out,__pycache__,build,dist,.vscode,parts,stage,prime,test_appstream.py,./snapcraft.spec,./tests/unit/test_common.py,./tests/integration/snaps/circular-dependencies/tree/snapcraft.yaml" -q4
```
sergiusens: why these latest additions?

cjp256: I should have put that in the commit.

snapcraft.spec: external Analysis object type uses "datas"
```
./snapcraft.spec:19: datas  ==> data
./snapcraft.spec:44: datas  ==> data
```

test_common: spell checking URL parts:
```
./tests/unit/test_common.py:36: fo  ==> of, for
self.assertFalse(common.isurl("/fo:o"))
```

snapcraft.yaml: letters for dependency names
```
./tests/integration/snaps/circular-dependencies/tree/snapcraft.yaml:31: ba  ==> by, be
./tests/integration/snaps/circular-dependencies/tree/snapcraft.yaml:51: ba  ==> by, be
```

I was hoping there was a way to disable the line line-by-line in python, but I don't see a way?

cjp256: Alternatively, I could rename the test in common, and part names for snapcraft.yaml.

Want me to do that?

snapcraft/internal/project_loader/_config.py
cjp256: done

runtests.sh
cjp256: I updated the commits to include more info, and adjusted the sources I could (test_common and snapcraft.yaml).

https://github.com/snapcore/snapcraft/compare/61e292aee2823e4bbf26d636528f0901bdce00ed..85bdba66fda9457610ffccaecf3317d9a2454b86

Only snapcraft.spec is now added to the ignore list.

PR #2805, opened 2019-11-18
plugins: address type errors for mypy uprev

cjp256:
Type fixes for plugins that should not affect runtime behavior.

### DIFF
--- a/snapcraft/plugins/cmake.py
+++ b/snapcraft/plugins/cmake.py
@@ -35,7 +35,7 @@ Additionally, this plugin uses the following plugin-specific keywords:
 
 import logging
 import os
-from typing import List
+from typing import List, Optional
 
 import snapcraft
 from snapcraft.internal import errors
@@ -56,7 +56,7 @@ class _Flag:
         parts = flag.split("=")
         self.name = parts[0]
         try:
-            self.value = parts[1]
+            self.value: Optional[str] = parts[1]
         except IndexError:
             self.value = None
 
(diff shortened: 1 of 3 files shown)


snapcraft/plugins/cmake.py
```diff
@@ -55,10 +55,11 @@ def __str__(self) -> str:
     def __init__(self, flag: str) -> None:
         parts = flag.split("=")
         self.name = parts[0]
+        self.value: Optional[str] = None
         try:
             self.value = parts[1]
         except IndexError:
-            self.value = None
+            pass
```
sergiusens: instead of passing, why not just do the assignment here and remove the assignment right before the try block?

cjp256: mypy complains that self.value is type string or whatever, and then it sees the except as trying to change it to optional.

but it probably would be better to do:

```
       try:
            self.value: Optional[str] = parts[1]
        except IndexError:
            self.value = None
```

cjp256: on a related note, would it be better form to check with len() rather than catch exception?

cjp256: Eh sorry, I fixed up commit and rebased against master at the same time.

```
diff --git a/snapcraft/plugins/cmake.py b/snapcraft/plugins/cmake.py
index 87d3a54a2..dbd826cb8 100644
--- a/snapcraft/plugins/cmake.py
+++ b/snapcraft/plugins/cmake.py
@@ -55,11 +55,10 @@ class _Flag:
     def __init__(self, flag: str) -> None:
         parts = flag.split("=")
         self.name = parts[0]
-        self.value: Optional[str] = None
         try:
-            self.value = parts[1]
+            self.value: Optional[str] = parts[1]
         except IndexError:
-            pass
+            self.value = None
```

Fixed!

PR #2810, opened 2019-11-19
build-providers: address errors from mypy uprev

cjp256:

-----

### DIFF
--- a/snapcraft/internal/build_providers/_lxd/_lxd.py
+++ b/snapcraft/internal/build_providers/_lxd/_lxd.py
@@ -138,12 +138,12 @@ class LXD(Provider):
         cmd.extend(command)
         self._log_run(cmd)
 
+        output = None
         try:
             if hide_output:
                 output = subprocess.check_output(cmd)
             else:
-                # output here will be None, and that is OK.
-                output = subprocess.check_call(cmd)
+                subprocess.check_call(cmd)
         except subprocess.CalledProcessError as process_error:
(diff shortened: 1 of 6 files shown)


snapcraft/internal/build_providers/_lxd/_lxd.py
```diff
@@ -350,6 +367,10 @@ def shell(self) -> None:
         self._run(command=["/bin/bash"])
 
     def _ensure_container_running(self) -> None:
+        # Sanity check - developer error if container not initialized.
```
sergiusens: would adding the check only here make it so that you don't need to do it in every method? If not, a decorator might be better

cjp256: I do not believe the check in _ensure_container_running was suficient unfortunately

cjp256: Confirmed.  And tried the decorator method but stumbled onto: https://github.com/python/mypy/issues/5836

sergiusens: gotcha, sucks a bit though

cjp256: I agree. I'm going to come back to this when we revisit local/remote provider.  I don't think _container should be optional, and just launched on init.  Everything is dependent on it anyhow... and if not, it's probably a classmethod.

PR #2818, opened 2019-11-22
project: pivot `info` from ProjectInfo to Snap

cjp256:
There's really three parts to this PR:

1. The detailing of Snap's from_dict() and to_dict() so as to not rely on the hackish `__dict__` usage.

2. Adding 'build-base' support to Snap.

3. Converging `info` and `_snap_meta` with Snap (and removing ProjectInfo) and moving the raw_snapcraft bits into Project.

The motivation for this PR was the mypy uprev.  project.info is Optional, and it's usage is extensive across the code base (assuming that's it's not None).  Instead of reworking ProjectInfo, I figured it was better to rip it out as it's mostly duplicating functionality of Snap.


cjp256:
I pulled out the Snap refactoring into https://github.com/snapcore/snapcraft/pull/2821 

I'll introduce build-base to Snap after, then revisit this PR.

PR #2819, opened 2019-11-22
isort: automatic formatting/sorting of imports

cjp256:
This commit adds an isort configuration that's pretty close to black with one exception.

- Add iscript config configuration that's pretty close to black with one exception. Ideally we'll enable this for a perfect match once available: https://github.com/psf/black/issues/251 

- Add a script to format source code (tools/format-code.sh) reliably.

- Add isort check (ignores whitespace) to static tests.

- Actually format code.  <-- this commit will need to be re-done once the current backlog of PRs is cleared (and this is ready to merge).

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,0 +1,7 @@
+[settings]
+multi_line_output=3
+include_trailing_comma=True
+force_grid_wrap=0
+use_parentheses=True
+line_length=88
+ensure_newline_before_comments=True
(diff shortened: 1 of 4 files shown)


.isort.cfg
```diff
@@ -0,0 +1,7 @@
+[settings]
```
sergiusens: does isort support pyproject.toml?

cjp256: It in fact looks like they do!

cjp256:
> the requirement for `tools/format-code.sh` makes this proposal a not ideal one as everyone using a code editor could be affected and require this.

I agree, but it's not really required if following the import style guide. `black` itself still does the actual formatting, `isort` will just ensure the imports are ordered and grouped appropriately. 

Regardless, it looks like `isort` will hopefully have a new release soon which will enable the option be consistent with black.  I think it would be just fine to wait until that release is available before doing anything with this PR.  Then we can drop the format-code script if desired.

PR #2821, opened 2019-11-25
meta: remove __dict__ usage in Snap 

cjp256:
- Remove `__dict__` usage in Snap and detail everything explicitly `from_dict()`, `to_dict()`, and `validate_required_keys()`.

- Make Snap fully configurable upon init with exhaustive __init__ parameters.

These changes will allow us to be more flexible going forward.  There should be no behavioral changes other than changing a few dictionaries to OrderedDict(), which they presumably should be since we expected them to be ordered (as can be seen with the paired `sorted()` use).

### DIFF
--- a/snapcraft/internal/meta/_snap_packaging.py
+++ b/snapcraft/internal/meta/_snap_packaging.py
@@ -19,6 +19,7 @@ import contextlib
 import itertools
 import logging
 import os
+from pathlib import Path
 import re
 import shlex
 import shutil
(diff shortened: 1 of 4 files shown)


cjp256:
To recap our conversation earlier (@sergiusens)

- remove/replace from_dict() with `from_snap_yaml()`, `from_snapcraft_yaml()`
- remove/replace to_dict() with `write_snap_yaml()`, with the possibility of a future `write_snapcraft_yaml()` that there is no case for (yet?)

cjp256:
@sergiusens added commit to split snap.yaml vs snapcraft.yaml handling: https://github.com/snapcore/snapcraft/pull/2821/commits/937da6c439c3d7f94c83ebea3252642c5f5adb0d

cjp256:
To ease review, I'm breaking this PR up into smaller ones.

PR #2822, opened 2019-11-26
xattrs: ignore errors if SNAPCRAFT_BUILD_INFO is unset

cjp256:
As SNAPCRAFT_BUILD_INFO is not always enabled, ignore errors
setting xattrs to record origin stage packages.  This data
will not be used if there is no manifest being generated.

This should prevent unnecessary errors being raised to the user
if their file system doesn't support xattrs (or are otherwise
limited from reading/modifying them).

Introduce XAttributeError() to improve the error messaging
to the user should reading/writing xattrs fail.  Mark as
reportable for now, but ultimately this should be False once
the feature is stable and extensively used by the community.

-----

### DIFF
--- a/snapcraft/internal/errors.py
+++ b/snapcraft/internal/errors.py
@@ -687,6 +687,22 @@ class ProjectNotFoundError(SnapcraftReportableError):
     fmt = "Failed to find project files."
 
 
+class XAttributeError(SnapcraftReportableException):
+    def __init__(self, *, action: str, key: str, path: str) -> None:
+        self._action = action
+        self._key = key
+        self._path = path
+
+    def get_brief(self) -> str:
+        return f"Unable to {self._action} extended attribute."
+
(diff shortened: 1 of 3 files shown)


snapcraft/internal/errors.py
```diff
@@ -687,6 +687,22 @@ class ProjectNotFoundError(SnapcraftReportableError):
...
+    def __init__(self, *, action: str, key: str, path: str) -> None:
+        self._action = action
+        self._key = key
+        self._path = path
+
+    def get_brief(self) -> str:
+        return f"Unable to {self._action} extended attribute."
+
+    def get_details(self) -> str:
+        return f"Failed to {self._action} attribute {self._key!r} on {self._path}."
```
sergiusens: self.path should also have !r

cjp256: Agreed, will do

PR #2830, opened 2019-12-04
elf: properly handle corrupted ELF files

kyrofa:

-----

Currently snapcraft dies during the priming step if the snap includes any corrupted ELF files (e.g ELF files where section names aren't strings). Fix this by catching the error and warning when it occurs.


snapcraft/internal/errors.py
```diff
@@ -708,3 +708,10 @@ def get_resolution(self) -> str:
             "the data source of the long data value. In some cases, the "
             "filesystem being used will limit the allowable size."
         )
+
+
+class CorruptedElfFile(SnapcraftError):
```
cjp256: Can you use SnapcraftException instead of SnapcraftError?  We're trying to migrate away from SnapcraftError.

snapcraft/internal/elf.py
```diff
@@ -619,6 +623,10 @@ def get_elf_files(root: str, file_list: Sequence[str]) -> FrozenSet[ElfFile]:
         except elftools.common.exceptions.ELFError:
             # Ignore invalid ELF files.
             continue
+        except errors.CorruptedElfFile as e:
```
sergiusens: we are trying to move away from "e", can you use a more descriptive var name?

snapcraft/internal/errors.py
kyrofa: Of course, thanks for the heads up: done. Note that I added a mitigation that, in the current codebase, will never been seen. Still seems sensible, though.

snapcraft/internal/elf.py
kyrofa: Sure, done.

kyrofa:
> Where did the corrupt test binaries come from?

From the Qt 5.13 codebase where this issue was originally discovered.

> Is it something that can be created dynamically?

I honestly don't know, but Qt doesn't, so I suspect it wouldn't be trivial.

cjp256:
> > Where did the corrupt test binaries come from?
> 
> From the Qt 5.13 codebase where this issue was originally discovered.
> 
> > Is it something that can be created dynamically?
> 
> I honestly don't know, but Qt doesn't, so I suspect it wouldn't be trivial.

Hrm, does the license match on the source files (GPL)? Can you annotate the canonical source for the files in the commit message for reference?

With that, +1 - thanks for changing the error! :)

kyrofa:
Yeah it's GPLv3. I noted the source of the files in the [test itself](https://github.com/snapcore/snapcraft/pull/2830/files#diff-6dfd146c9760b95a7bd9bb738a6592a6R6), you would like it in the commit message as well? Would you like me to squash then, or just add another no-change commit?

cjp256:
Ah, I didn't notice the note in the tests, but I think it's still best to attribute the source into the commit message. A squash/rebase would be great, thanks!  I think everything is ready to go...

E.g.:
```
corrupt_empty.so is copied from QT's source, available under GPLv3 at:
`https://code.qt.io/cgit/qt/qtbase.git/tree/<path-to-test-file>?h=5.13`
```

kyrofa:
> A squash/rebase would be great, thanks!

Done!

PR #2833, opened 2019-12-06
catkin: remove rospack workaround

Arnatious:

-----

Reverts a change made in #2585 now that rosdep has grown the ability to ignore source packages on its own.


cjp256:
If you do a git clean / remove the mypy cache, you will likely be able to repro the error in the tests.

I'd be willing to wager it's a bug or something related to import loops or something.  There's a pretty common pattern @sergiusens has been using used to import Project for type checking like done on:
https://github.com/snapcore/snapcraft/pull/2833/files#diff-9185414f4ff7052827d5727089e12823L26.

It also worked for me without the if TYPE_CHECKING.

snapcraft/plugins/catkin.py
```diff
@@ -87,6 +87,8 @@
 from snapcraft import common, file_utils, formatting_utils, repo
 from snapcraft.internal import errors, mangling
 
+from snapcraft.project import Project
```
sergiusens: Please make this import conditional to typing

Arnatious: Done in 9fe00f9

snapcraft/plugins/catkin.py
```diff
@@ -79,14 +79,17 @@
...
+from typing import List, Set, TYPE_CHECKING
 
 import snapcraft
 from snapcraft.plugins import _ros
 from snapcraft.plugins import _python
 from snapcraft import common, file_utils, formatting_utils, repo
 from snapcraft.internal import errors, mangling
 
+if TYPE_CHECKING:
+    from snapcraft.project import Project
```
kyrofa: This is odd. It wasn't required before we added rospack... not sure why it's needed now. :man_shrugging: .

tests/unit/plugins/test_catkin.py
```diff
@@ -1683,24 +1617,19 @@ def test_find_system_dependencies_system_only_no_packages(self):
...
-            ),
+            catkin._find_system_dependencies(None, self.rosdep_mock, self.catkin_mock),
             Equals(dict()),
         )
 
         self.rosdep_mock.get_dependencies.assert_called_once_with()
-        self.rospack_mock.list_names.assert_called_once_with()
-        self.rosdep_mock.resolve_dependency.assert_not_called()
-        self.catkin_mock.find.assert_not_called()
+        self.rosdep_mock.resolve_dependency.assert_called_once_with("bar")
```
kyrofa: This is no longer testing the same thing. Is it any different now than [test_find_system_dependencies_system_only_no_packages](https://github.com/snapcore/snapcraft/pull/2833/files#diff-e65da2ca12f490f5cab4eeedc84e6636R1607)? Looks the same to me, but it's missing the mock to make it actually return anything other than an empty dict. I think this entire test should be removed now, as the behavior is only present now within rosdep itself, the calling of which is already well-tested, and the spread test ensures this functionality still exists.

snapcraft/plugins/catkin.py
Arnatious: rospack had this import in it before

kyrofa: But we didn't need it BEFORE we added rospack either. Anyway, no changed needed here, I just find it confusing and wanted to whine.

Arnatious: Been needed since: https://github.com/snapcore/snapcraft/commit/d1c5f1b00abe312b7df6e5a7c84896de5657c992#diff-3cc26b49a70ee86523598a96dd385301R1081

kyrofa: Nice detective work. That makes more sense.

tests/unit/plugins/test_catkin.py
Arnatious: Removed this test in d823c5b

PR #2836, opened 2019-12-07
meta: fix command-chain handling when Application adapter == "none"

cjp256:
As reported by @chipaca, `command-chain` is being set when adapter is specified to none.  This should address that along with some minor related changes to clean things up a bit.

meta: fix command-chain handling when Application adapter == "none"
meta: remove Application's `prepend_command_chain`
meta: convert Application's `adapter` from string to enum
project: remove unused SanityCheckError and CommandChainWithLegacyAdapterError
schema: set app adapter default to full

### DIFF
--- a/schema/snapcraft.json
+++ b/schema/snapcraft.json
@@ -568,7 +568,7 @@
                                 "legacy",
                                 "full"
                             ],
-                            "default": "legacy"
+                            "default": "full"
                         },
                         "command-chain": {
                             "type": "array",
--- a/snapcraft/internal/meta/_snap_packaging.py
+++ b/snapcraft/internal/meta/_snap_packaging.py
@@ -367,8 +367,13 @@ class _SnapPackaging:
 
     def finalize_snap_meta_command_chains(self) -> None:
         prepend_command_chain = self._generate_command_chain()
+
+        # If there's nothing to prepend, we have nothing to do.
+        if not prepend_command_chain:
+            return
+
         for app_name, app in self._snap_meta.apps.items():
-            app.prepend_command_chain = prepend_command_chain
+            app.command_chain = prepend_command_chain + app.command_chain
 
     def finalize_snap_meta_version(self) -> None:
         # Reparse the version, the order should stick.
--- a/snapcraft/internal/meta/_snap_packaging.py
+++ b/snapcraft/internal/meta/_snap_packaging.py
@@ -83,7 +83,7 @@ def create_snap_packaging(project_config: _config.Config) -> str:
     packaging.cleanup()
     packaging.validate_common_ids()
     packaging.finalize_snap_meta_commands()
-    packaging.finalize_snap_meta_command_chains()
+    packaging.finalize_snap_meta_snapcraft_runner()
     packaging.finalize_snap_meta_version()
     packaging.write_snap_yaml()
     packaging.setup_assets()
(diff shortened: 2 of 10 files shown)


schema/snapcraft.json
```diff
@@ -568,7 +568,7 @@
                                 "legacy",
                                 "full"
                             ],
-                            "default": "legacy"
+                            "default": "full"
```
sergiusens: this entire entry should be invalid for anything NOT core or core18, changing this is not desired.

snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -367,8 +368,16 @@ def finalize_snap_meta_commands(self) -> None:
 
     def finalize_snap_meta_command_chains(self) -> None:
         prepend_command_chain = self._generate_command_chain()
+
+        # If there's nothing to prepend, we have nothing to do.
+        if not prepend_command_chain:
+            return
```
sergiusens: good catch

schema/snapcraft.json
cjp256: In master, there's no special handling AFAICT for "full" or "legacy".

I went back and looked at the behavior for 3.8 to compare:

```
if adapter == "full":
   command: <command>
   command-chain:
   - snap/command-chain/snapcraft-runner

elif adapter == "legacy":
   command: snap/command-chain/snapcraft-runner <command>

elif adapter == "none":`
  command: <command>
```

The current behavior I see (with this PR anyhow) is that if the adapter is "none", command-chain usage is disabled.

In master, it always uses command-chain.  Was that not intentional?  And doesn't that effectively mean "full" is synonymous with "legacy"? 

Should:
(1) the old legacy behavior be restored and
(2) and maintained as default?

cjp256: FWIW, testing with _classic_ confinement:

Stable 3.8:

```
if adapter == "full":
   command: <command>
   command-chain:
   - snap/command-chain/snapcraft-runner

elif adapter == "legacy":
   command: snap/command-chain/snapcraft-runner command-<command>.wrapper

elif adapter == "none":
  command: <command>
```

In master, today without this PR:

```
if adapter == "full":
   command: <command>

elif adapter == "legacy":
   command: <command>

elif adapter == "none":
   command: <command>
```

sergiusens: Look at the contents of the snapcraft-runner, it should be empty for classic, more so, not something snapcraft should set. 3.8 is probably buggy.

cjp256: Did you see a more extensive post prior to this?  It seems to be gone?

cjp256: Was in my email, but not showing on here...

```
In master, there's no special handling AFAICT for "full" or "legacy".

I went back and looked at the behavior for 3.8 to compare:

if adapter == "full":
   command: <command>
   command-chain:
   - snap/command-chain/snapcraft-runner

elif adapter == "legacy":
   command: snap/command-chain/snapcraft-runner <command>

elif adapter == "none":`
  command: <command>

The current behavior I see (with this PR anyhow) is that if the adapter is "none", command-chain usage is disabled.

In master, it always uses command-chain. Was that not intentional? And doesn't that effectively mean "full" is synonymous with "legacy"?

Should:
(1) the old legacy behavior be restored and
(2) and maintained as default?
```

cjp256:
While venturing beyond the scope of this PR, I added a commit which I believe should restore the behavior found in 3.8 wrt adapter, cleaning up and fixing a few things in the process (with the goal of making it a much easier to understand and maintain).  Subject for discussion anyhow, I can pull it if desired. https://github.com/snapcore/snapcraft/pull/2836/commits/49c71fcdda109e052f31687cc83cea8b9fd30474

cjp256:
Closing this in favor of a simpler PR suitable for back-porting: https://github.com/snapcore/snapcraft/pull/2853

And refactoring will occur separately once the above PR is merged.

PR #2839, opened 2019-12-12
plugs: add passthrough; check plug_dict is not empty

sd-hd:
* allow plugs to passthrough

* plugs don't need to have an interface

(sorry, tests are missing yet)


sd-hd:
sorry, there is an error. I will reopened it later

PR #2840, opened 2019-12-12
plugs: plugs can have no element

sd-hd:
if plug has no subelement, then the interface changes crash this

### DIFF
--- a/snapcraft/internal/meta/plugs.py
+++ b/snapcraft/internal/meta/plugs.py
@@ -29,7 +29,7 @@ class Plug:
 
     def __init__(self, *, plug_name: str) -> None:
         self._plug_name = plug_name
-        self._plug_dict: Dict[str, Any] = dict()
+        self._plug_dict: Dict[str, Any] = {"interface": plug_name}
 
     @property
     def plug_name(self) -> str:
@@ -54,15 +54,16 @@ class Plug:
     def from_dict(cls, *, plug_dict: Dict[str, Any], plug_name: str) -> "Plug":
         """Create plug from dictionary."""
 
-        interface = plug_dict.get("interface", None)
-        if interface is not None:
-            plug_class = PLUG_MAPPINGS.get(interface, None)
-            if plug_class is not None:
-                return plug_class.from_dict(plug_dict=plug_dict, plug_name=plug_name)
+        plug_dict = {} if plug_dict is None else plug_dict
+        interface = plug_dict.get("interface", plug_name)
+
+        plug_class = PLUG_MAPPINGS.get(interface, None)
+        if plug_class is not None:
+            return plug_class.from_dict(plug_dict=plug_dict, plug_name=plug_name)
 
         # Handle the general case.
         plug = Plug(plug_name=plug_name)
-        plug._plug_dict = plug_dict
+        plug._plug_dict.update(plug_dict)
         return plug
 
     def to_dict(self) -> Dict[str, Any]:
--- a/snapcraft/internal/meta/slots.py
+++ b/snapcraft/internal/meta/slots.py
@@ -31,7 +31,7 @@ class Slot:
 
     def __init__(self, *, slot_name: str) -> None:
         self._slot_name = slot_name
-        self._slot_dict: Dict[str, Any] = dict()
+        self._slot_dict: Dict[str, Any] = {"interface": slot_name}
 
     @property
     def slot_name(self) -> str:
@@ -55,16 +55,17 @@ class Slot:
     def from_dict(cls, *, slot_dict: Dict[str, Any], slot_name: str) -> "Slot":
         """Return applicable Slot instance from dictionary properties."""
 
+        slot_dict = {} if slot_dict is None else slot_dict
+        interface = slot_dict.get("interface", slot_name)
+
         # If we explicitly support the type, use it instead.
-        interface = slot_dict.get("interface", None)
-        if interface is not None:
-            slot_class = SLOT_MAPPINGS.get(interface, None)
-            if slot_class is not None:
-                return slot_class.from_dict(slot_dict=slot_dict, slot_name=slot_name)
+        slot_class = SLOT_MAPPINGS.get(interface, None)
+        if slot_class is not None:
+            return slot_class.from_dict(slot_dict=slot_dict, slot_name=slot_name)
 
         # Handle the general case.
         slot = Slot(slot_name=slot_name)
-        slot._slot_dict = slot_dict
+        slot._slot_dict.update(slot_dict)
         return slot
 
     def to_dict(self) -> Dict[str, Any]:
--- a/tests/unit/meta/test_plugs.py
+++ b/tests/unit/meta/test_plugs.py
@@ -39,18 +39,19 @@ class GenericPlugTests(unit.TestCase):
         plug_name = "plug-test"
 
         plug = Plug(plug_name=plug_name)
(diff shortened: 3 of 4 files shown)


sd-hd:
hi @cjp256,

i added a minimalistic test.



Today i got following documentation about plugs: 
https://forum.snapcraft.io/t/plugs-in-global-position-in-snapcraft-yaml/14618/2

So empty plugs are allowed. 
But the "test_invalid_from_dict_raises_exception" checks somehow of the plug is empty ?

sd-hd:
I wonder about this, too.

Must plugs have interfaces, or not ?


```
def validate(self):
       ....

        if "interface" not in self._plug_dict:
            raise PlugValidationError(
                plug_name=self.plug_name, message="plug has no defined interface"
            )
```

sd-hd:
Here is a example:

```
name: test

plugs:
   wayland:
   x11:
   login-session-observe:
   gtk-3-themes:
    interface: content
    target: $SNAP/data-dir/themes
    default-provider: gtk-common-themes

apps:
   myapp: bin/true

parts:
    mypart:
       plugin: nil


```

sd-hd:
I have another idea.

Would it make more sense to automatically added the missing "interface" if none is added?

cjp256:
Well I just learned something, I don't think I've seen that yet, but it's certainly valid!

snapcraft 3.8 will print out:
```
plugs:
  <plug-name>: null
```

And it functions, enabling that plug for all apps. So this is certainly a regression.

The interface name will just be the plug name.  So if we come across None, what do you think about creating a "default" plug_dict, where `plug_dict=dict(interface=plug_name)`?  The output snap.yaml would get a little more verbose, but that's not a problem (and we could address that separately).

cjp256:
> I have another idea.
> 
> Would it make more sense to automatically added the missing "interface" if none is added?

Yes :D Agreed.

sd-hd:
@cjp256 
ok, i will update this soon ;-)

We have to took a look for "slots". The also have automatically interface in it.

sd-hd:
Hi, Happy new year,

it's sometime ago. Is there anything more to do here?

snapcraft/internal/meta/slots.py
```diff
@@ -55,16 +55,16 @@ def validate(self) -> None:
     def from_dict(cls, *, slot_dict: Dict[str, Any], slot_name: str) -> "Slot":
         """Return applicable Slot instance from dictionary properties."""
 
+        iface = slot_dict["interface"] if "interface" in slot_dict else slot_name
```
cjp256: I think the more idiomatic way to do this would be to just update the default value for `.get()` like so:
`interface = slot_dict.get("interface", slot_name)`

sd-hd:
> Happy New Year @sd-hd! :D

Happy New Year :-)

snapcraft/internal/meta/slots.py
sd-hd: You are right.

snapcraft/internal/meta/plugs.py
```diff
@@ -54,15 +54,15 @@ def validate(self) -> None:
     def from_dict(cls, *, plug_dict: Dict[str, Any], plug_name: str) -> "Plug":
         """Create plug from dictionary."""
 
-        interface = plug_dict.get("interface", None)
-        if interface is not None:
-            plug_class = PLUG_MAPPINGS.get(interface, None)
-            if plug_class is not None:
-                return plug_class.from_dict(plug_dict=plug_dict, plug_name=plug_name)
+        iface = plug_dict.get("interface", plug_name)
```
sergiusens: may I ask why the variable "interface" was renamed to iface?

sd-hd: @sergiusens 
before i had a very long line. Now i can rename it back.

Please, don't merge it yet, because there is still a bug.

sd-hd:
@cjp256 
@sergiusens 
Sorry, because i didn't wrote a test for the "None" case. 
There is still a check missing.

My fault :-( 
Hadn't have time until now, to work with the newer snapcraft:3.9 and this patch ... Sorry

Tomorrow i will fix this too

sd-hd:
I added the initial and missing None check yesterday ;-)

snapcraft/internal/meta/plugs.py
sd-hd: Had renamed iface back to interface

zyga:
> Must plugs have interfaces, or not ?

Plug and slots don't need to have an interface definition. Implicitly the name of the plug or slot is used as the interface name in that case.

snapcraft/internal/meta/plugs.py
```diff
@@ -54,15 +54,16 @@ def validate(self) -> None:
     def from_dict(cls, *, plug_dict: Dict[str, Any], plug_name: str) -> "Plug":
         """Create plug from dictionary."""
 
-        interface = plug_dict.get("interface", None)
-        if interface is not None:
-            plug_class = PLUG_MAPPINGS.get(interface, None)
-            if plug_class is not None:
-                return plug_class.from_dict(plug_dict=plug_dict, plug_name=plug_name)
+        plug_dict = {} if plug_dict is None else plug_dict
```
zyga: Just my 0.02PLN, the assignment, especially using the rather verbose python conditional syntax is probably not saving much over the very explicit:

```
if plug_dict is None:
  plug_dict = {}
```

snapcraft/internal/meta/plugs.py
```diff
@@ -54,15 +54,16 @@ def validate(self) -> None:
...
+        interface = plug_dict.get("interface", plug_name)
```
zyga: Is this redundant given the change to `Plug.__init__`?

snapcraft/internal/meta/plugs.py
cjp256: I've got some other related changes in queue, I can update in that PR.

cjp256: Relatedly, the plug_dict needs to be labelled Optional if allowing None, but I'll be following up with that as well.

snapcraft/internal/meta/plugs.py
cjp256: Indeed

snapcraft/internal/meta/plugs.py
sd-hd: > Just my 0.02PLN

For me both assignments are ok. in the past I had always saved lines.

snapcraft/internal/meta/plugs.py
sd-hd: No - "`from_dict`" is a second constructor by the "`@classmethod`" decorator.

If we use, the classic "`if`" assignment, then it would become clearer. 



```
        if plug_dict is not None:
            interface = plug_dict.get("interface", plug_name)
            plug_class = PLUG_MAPPINGS.get(interface, None)
            if plug_class is not None:
                return plug_class.from_dict(plug_dict=plug_dict, plug_name=plug_name)
```

But here the indention-level increases a bit.

snapcraft/internal/meta/plugs.py
sd-hd: > Relatedly, the plug_dict needs to be labelled Optional if allowing None, but I'll be following up with that as well.

@cjp256 
In the past years i only used python 2.7 on a Solaris 10 machine. So for me, you are the master with "labels" ;-)

PR #2841, opened 2019-12-12
plugs: add passthrough

sd-hd:
allow plugs to passthrough


sd-hd:
An example yaml would like:
```
name: test

plugs:
   wayland:
   x11:
   login-session-observe:
   gtk-3-themes:
    interface: content
    target: $SNAP/data-dir/themes
    default-provider: gtk-common-themes

apps:
   myapp: bin/true

parts:
    mypart:
       plugin: nil

```

PR #2842, opened 2019-12-12
rust: add support for workspaces

cjp256:


cjp256:
Tested by building alacritty snap.

tests/spread/plugins/rust/snaps/rust-workspace-fib/fib/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "fib"
+version = "0.1.0"
+authors = ["Chris Patterson <chris.patterson@canonical.com>"]
+edition = "2018"
```
sergiusens: funny, you are behind the times.

snapcraft/plugins/rust.py
```diff
@@ -193,21 +194,74 @@ def _get_target(self) -> str:
...
+
     def build(self):
         super().build()
 
-        # Write a minimal config.
-        self._write_cargo_config()
+        uses_workspaces = self._project_uses_workspace()
+
+        if uses_workspaces:
+            # This is a bit ugly because `cargo install` does not yet support
```
sergiusens: seems some of the discrepancies we have when using go.mod can also be seen here :man_shrugging:

snapcraft/plugins/rust.py
```diff
@@ -193,21 +194,74 @@ def _get_target(self) -> str:
...
+        # Find artifacts in release directory.
+        release_dir = Path(self.builddir, "target", "release")
+
+        # Install binaries to bin/.
+        bins_dir = Path(self.installdir, "bin")
+        bins_dir.mkdir(parents=True, exist_ok=True)
+
+        # Install shared objects to usr/lib/<arch-triplet>.
+        # TODO: Dynamic library support needs to be properly added.
+        # Although weinstall libraries if we find them, they are most
```
sergiusens: missing space typo

snapcraft/plugins/rust.py
```diff
@@ -193,21 +194,74 @@ def _get_target(self) -> str:
             )
         return rust_target.format("unknown-linux", "gnu")
 
+    def _project_uses_workspace(self) -> bool:
+        path = Path(self.builddir, "Cargo.toml")
+        if not path.is_file():
```
sergiusens: can anything build without a Cargo.toml? Maybe an exception about not being a supported rust project would be better?

snapcraft/plugins/rust.py
cjp256: Hrm, odd that the `codespell` doesn't flag it... will fix

snapcraft/plugins/rust.py
cjp256: Ah yes, indeed! :)

snapcraft/plugins/rust.py
cjp256: I don't think it's valid without Cargo.toml according my minimal experience and testing, but I don't know that's saying much :smile: 

While I'm rather OK with the `cargo install` step failing after this as it should be rather apparent to the user when cargo complains (`error: <path> does not contain a Cargo.toml file. --path must point to a directory containing a Cargo.toml file.`), I do think it's reasonable to at least warn on conditions in which we believe the plugin will fail, particularly if it ends up being more helpful for the user to diagnose.

Is there an existing exception you would prefer? Or perhaps just `warn()`? Or add a new one like `SnapcraftPluginAssertionFailure()` so we can use it across all plugins consistently?

tests/spread/plugins/rust/snaps/rust-workspace-fib/fib/Cargo.toml
cjp256: lol, I think 2015 and 2018 are the valid rust editions :wink: 2021 might get some love, so we'll have to revisit this then!

sergiusens: Ha, for some reason I thought this was the "edition" for the package and not rust itself

snapcraft/plugins/rust.py
sergiusens: Failing here is fine, if it ever works it will work by chance, right?

We have multiple in-plugin exceptions already, but I am not opposed to a generic one

cjp256:
rebased/squashed

snapcraft/plugins/rust.py
```diff
@@ -194,21 +195,78 @@ def _get_target(self) -> str:
...
+        # likely going to be missing dependencies, e.g.:
+        # /home/ubuntu/.cargo/toolchains/stable-x86_64-unknown-linux-gnu/lib/libstd-fae576517123aa4e.so
+        libs_dir = Path(self.installdir, "usr", "lib", self.project.arch_triplet)
+        libs_dir.mkdir(parents=True, exist_ok=True)
+
+        # Cargo build marks binaries and shared objects executable...
+        # Search target directory to get these and install them to the
+        # correct location.
+        for path in release_dir.iterdir():
+            if not os.path.isfile(path):
```
sergiusens: I really wish we could start using list comprehension for these, it takes a bit more to write but it makes it much faster to read, using ( instead of [ makes them generators, so if concern over multiple loops taking place is what prevents it, the generator part should be the sell :-)

snapcraft/plugins/rust.py
```diff
@@ -194,21 +195,78 @@ def _get_target(self) -> str:
...
+                file_utils.link_or_copy(path.as_posix(), bins_dir.as_posix())
+
     def build(self):
         super().build()
 
-        # Write a minimal config.
-        self._write_cargo_config()
+        uses_workspaces = self._project_uses_workspace()
+
+        if uses_workspaces:
```
sergiusens: does _write_cargo_config not work with workspaces?

snapcraft/plugins/rust.py
cjp256: Okie dokie, your wish is my command :wink:

snapcraft/plugins/rust.py
cjp256: ah, i did miss that.  I guess that might cause trouble if trying to cross compile huh...

sergiusens: but, the more direct use case is the parallel build count

cjp256: Fixed!

snapcraft/plugins/rust.py
cjp256: Fixed!

cjp256:
Added two commits addressing comments

PR #2843, opened 2019-12-12
snap: only ship .pyc files, strip shared objects

om26er:
- [*] Have you followed the [guidelines for contributing](https://github.com/snapcore/snapcraft/blob/master/CONTRIBUTING.md)?
- [*] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [*] Have you successfully run `./runtests.sh static`?
- [*] Have you successfully run `./runtests.sh tests/unit`?

-----
To speed things up snapcraft ships .pyc files as part of its snap package, at the same times it also ships .py files. This is duplication and increases the size of the snap unneeded.

This PR also strips shared objects that are bundled with snapcraft.

The change reduces the download size of the snap by 15Mb

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -98,7 +98,9 @@ parts:
 
         echo "Compiling pyc files..."
         # This is the last step, let's now compile all our pyc files.
-        "$SNAPCRAFT_PART_INSTALL/usr/bin/python3" -m compileall -q .
+        "$SNAPCRAFT_PART_INSTALL/usr/bin/python3" -m compileall -b -q .
+        find . -type f -name '*.py' -exec rm {} \;
+        find . -type f -name "*.so" -exec strip -s {} \;
 
     after: [snapcraft-libs]
 
@@ -114,6 +116,7 @@ parts:
         $SNAPCRAFT_PROJECT_DIR/tools/snapcraft-override-build.sh
 
         sed -ri 's|(lib/.*/site-packages)|legacy_snapcraft/\1|' $SNAPCRAFT_PART_INSTALL/usr/lib/python3.6/sitecustomize.py
+        find . -type f -name "*.so" -exec strip -s {} \;
     organize:
         '*': legacy_snapcraft/
     after: [snapcraft-libs]


om26er:
Is there a way to force a CI rerun ? I have definitely signed the CLA last night

snap/snapcraft.yaml
```diff
@@ -98,7 +98,9 @@ parts:
 
         echo "Compiling pyc files..."
         # This is the last step, let's now compile all our pyc files.
-        "$SNAPCRAFT_PART_INSTALL/usr/bin/python3" -m compileall -q .
+        "$SNAPCRAFT_PART_INSTALL/usr/bin/python3" -m compileall -b -q .
+        find . -type f -name '*.py' -exec rm {} \;
+        find . -type f -name "*.so" -exec strip -s {} \;
```
cjp256: We do plan to introduce a feature flag to enable stripping in a more generic manner in the not-so-distant future, I'd perhaps wait on that for this type of change?

snap/snapcraft.yaml
```diff
@@ -98,7 +98,9 @@ parts:
 
         echo "Compiling pyc files..."
         # This is the last step, let's now compile all our pyc files.
-        "$SNAPCRAFT_PART_INSTALL/usr/bin/python3" -m compileall -q .
+        "$SNAPCRAFT_PART_INSTALL/usr/bin/python3" -m compileall -b -q .
+        find . -type f -name '*.py' -exec rm {} \;
```
cjp256: This is an interesting one.  Does this have any effect on exceptions or any other run-time side effects?

Seems a bit risky, but mostly because I don't know the set of actions/environment changes that would trigger a rebuild (or ignoring the pyc).  I assume the fact that we're packaging all the dependencies in the snap may lower the risk, but does it eliminate it?

Although this can be satisfied easily enough in override-build as you demonstrate, I wonder if it would be useful to support a [python plugin?] flag for projects wishing to do this?

om26er:
> Out of curiosity, do you know what the size difference was just for the .py, vs stripping?

Without removing .py files, the size difference is 5988Kb

om26er:
> But I'm honestly not really keen on removing the .py files. As an open-source development tool I don't think removing the source is worth the trade-off if users ever want to reference it for the version they have installed. Though certainly worth considering!

Technically, if the project was in any of the compiled languages (c, c++ or go), we wouldn't be shipping source as part of the package. I believe snapcraft is tagged well enough on github so that if anyone wants to actually see the code for a specific version they could do that on GitHub.

snap/snapcraft.yaml
om26er: This is project's (Snapcraft) decision, if it wants to have the size benefit now or later.

snap/snapcraft.yaml
om26er: > This is an interesting one. Does this have any effect on exceptions or any other run-time side effects?

That is indeed an issue i.e. if the relevant .py file is removed, the traceback only mentions the line number, see below
```
  File "main.py", line 10, in <module>
    thrower()
  File "main.py", line 6, in thrower
    raise Exception("OK")
Exception: OK
```
VS
```
  File "main.py", line 10, in <module>
  File "main.py", line 6, in thrower
Exception: OK
```


> Seems a bit risky, but mostly because I don't know the set of actions/environment changes that would trigger a rebuild (or ignoring the pyc).

In my understand the pyc only gets regenerated when the relevant py file changes and in case of a snap, due to its read-only nature that cannot ever happen once the package is created.

om26er:
ping @sergiusens

sergiusens:
Apologies for the delay, but this is quite a controversial change and I am sorry to say, this does not seem to fit our goals.

PR #2845, opened 2019-12-17
remote-build: configurable timeout/deadline for starting and monitoring build

cjp256:
The error was ambiguous when the timeout was hit, and there was no recourse for the user.

Re-instate a configurable timeout as `--launchpad-timeout` but apply it to both the start_build and monitor_build as a deadline.  When Launchpad is slow, it may take some time to process the request, but as long as its Pending, we keep waiting for a response (unless deadline is hit).

### DIFF
--- a/snapcraft/cli/remote.py
+++ b/snapcraft/cli/remote.py
@@ -15,6 +15,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import click
+import time
 
 from snapcraft.project import Project
 from snapcraft.internal.remote_build import WorkTree, LaunchpadClient, errors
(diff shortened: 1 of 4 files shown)


snapcraft/cli/remote.py
```diff
@@ -111,43 +125,47 @@ def remote_build(
...
+        project=project,
+        build_id=build_id,
+        architectures=architectures,
+        deadline=deadline,
     )
 
     if status:
         _print_status(lp)
+        return
     elif recover:
```
sergiusens: seems that if the above statement is going to be treated as a guard, this elif should start its own if block

cjp256: Fair enough, will fix

cjp256:
Fixed comments as well as some other minor issues I saw lurking.  
https://github.com/snapcore/snapcraft/pull/2845/commits/878860a4695cda5da9f047d4bedba1783675bcb3

PR #2847, opened 2019-12-17
Catkin plugin: consider only 'local' workspaces

artivis:
This PR fixes a bug when calling snapcraft with `--destructive-mode`.
Some installation paths from the host machine leak into the list of paths considered by the catkin plugin. It results in having dependencies being found (in the host path) and thus not being installed by the plugin in the expected 'parts' path.

In the 'pull' step, prior to building, the ROS dependencies of the package(s) to be built are retrieved.
When checking whether a dependency is available 'locally' or should be installed, the `catkin_find` tool will look into workspaces exported in `CMAKE_PREFIX_PATH` among other paths:
[`catkin_find` calls `find_in_workspaces`](https://github.com/ros/catkin/blob/kinetic-devel/bin/catkin_find#L41) which in turns [calls `get_workspaces`](https://github.com/ros/catkin/blob/kinetic-devel/python/catkin/find_in_workspaces.py#L114) which [retrieve those from `CMAKE_PREFIX_PATH`](https://github.com/ros/catkin/blob/kinetic-devel/python/catkin/workspace.py#L46).

Now, in order for `catkin_find` to work, some ROS-related environment variables are exported by sourcing an automatically generated (ag.) script.
The ag. script `opt/ros/melodic/_setup_util.py` (called by ag. `opt/ros/melodic/setup.sh`) [sets the variable `CMAKE_PREFIX_PATH` to `/opt/ros/<distro>`](https://github.com/ros/catkin/blob/kinetic-devel/cmake/templates/_setup_util.py.in#L273) by default. This results in having the path to the ROS install on the host being searched when looking for packages.
To prevent looking into `CMAKE_PREFIX_PATH`, the flag `--local` is passed to `opt/ros/<distro>/setup.sh` when sourcing it.

-----

### DIFF
--- a/snapcraft/plugins/catkin.py
+++ b/snapcraft/plugins/catkin.py
@@ -1038,14 +1038,14 @@ class _Catkin:
             # Source our own workspace so we have all of Catkin's dependencies,
             # then source the workspace we're actually supposed to be crawling.
             lines.append(
-                "_CATKIN_SETUP_DIR={} source {}".format(
+                "_CATKIN_SETUP_DIR={} source {} --local".format(
                     ros_path, os.path.join(ros_path, "setup.sh")
                 )
             )
 
             for workspace in self._workspaces:
                 lines.append(
-                    "_CATKIN_SETUP_DIR={} source {} --extend".format(
+                    "_CATKIN_SETUP_DIR={} source {} --local --extend".format(
                         workspace, os.path.join(workspace, "setup.sh")
                     )
                 )
--- a/snapcraft/plugins/catkin.py
+++ b/snapcraft/plugins/catkin.py
@@ -579,9 +579,10 @@ class CatkinPlugin(snapcraft.BasePlugin):
             source_script = textwrap.dedent(
                 """
                 if [ -f {underlay_setup} ]; then
+                    set -- --local
                     _CATKIN_SETUP_DIR={underlay} . {underlay_setup}
                     if [ -f {rosdir_setup} ]; then
-                        set -- --extend
+                        set -- --local --extend
                         _CATKIN_SETUP_DIR={rosdir} . {rosdir_setup}
                     fi
                 fi
@@ -596,6 +597,7 @@ class CatkinPlugin(snapcraft.BasePlugin):
             source_script = textwrap.dedent(
                 """
                 if [ -f {rosdir_setup} ]; then
+                    set -- --local
                     _CATKIN_SETUP_DIR={rosdir} . {rosdir_setup}
                 fi
             """
--- a/tests/unit/plugins/test_catkin.py
+++ b/tests/unit/plugins/test_catkin.py
@@ -931,7 +931,7 @@ class CatkinPluginTestCase(CatkinPluginBaseTest):
         lines_of_interest = [
             "set --",
             "if [ -f {} ]; then".format(underlay_setup_path),
-            "set -- --local"
+            "set -- --local",
             "_CATKIN_SETUP_DIR={} . {}".format("test-underlay", underlay_setup_path),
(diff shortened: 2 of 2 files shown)


artivis:
This fix isn't critical as I guess not so many people build their ROS snaps with `--destructive-mode`.  
I'd wait for a double check from @kyrofa.

snapcraft/plugins/catkin.py
```diff
@@ -1038,14 +1038,14 @@ def _run(self, arguments):
...
-                "_CATKIN_SETUP_DIR={} source {}".format(
+                "_CATKIN_SETUP_DIR={} source {} --local".format(
                     ros_path, os.path.join(ros_path, "setup.sh")
                 )
             )
 
             for workspace in self._workspaces:
                 lines.append(
-                    "_CATKIN_SETUP_DIR={} source {} --extend".format(
+                    "_CATKIN_SETUP_DIR={} source {} --local --extend".format(
```
kyrofa: I haven't looked at how this is implemented, but from [the docs](https://github.com/ros/catkin/blob/2f51d6618788267192b57fbbd24ff41e8e6b4118/cmake/templates/setup.sh.in#L7-L8), `--local` and `--extend` seem mutually exclusive. Specifically, `--extend` allows chaining, building on an underlay, whereas `--local` explicitly ignores the underlay. We need the ability to chain here, thus I doubt `--local` is what we need. That said, your investigation was tremendously well done, thank you, and leads us toward another solution.

Check [this](https://github.com/snapcore/snapcraft/blob/master/snapcraft/plugins/catkin.py#L717-L725) out. We massage _setup_util.py to make sure it doesn't try to pull ROS bits off the host, but _only after we finish building_. I suggest looking into doing something similar in [_prepare_build](https://github.com/snapcore/snapcraft/blob/master/snapcraft/plugins/catkin.py#L660) as well. Is that feasible?

artivis: You are right, `--local` will fully clear `CMAKE_PREFIX_PATH` including any underlay. Looking at your suggestion.

kyrofa: After @artivis's research, it seems that `--local` actually does exactly what we need here. They aren't mutually exclusive, in fact, `--local` has the exact same effect as our `_setup_util.py` massaging, which means we can probably get rid of that hack as well. Nice find!

snapcraft/plugins/catkin.py
```diff
@@ -579,9 +579,10 @@ def _source_setup_sh(self, root, underlay_path):
             source_script = textwrap.dedent(
                 """
                 if [ -f {underlay_setup} ]; then
+                    set -- --local
                     _CATKIN_SETUP_DIR={underlay} . {underlay_setup}
                     if [ -f {rosdir_setup} ]; then
-                        set -- --extend
+                        set -- --local --extend
```
cjp256: Forgive my ignorance, but what is the effect of `set -- --<flag>`?  I would naively assume it as the shell's `set`, but that must not be?  I'm missing something of great value :thinking:

kyrofa: `source` is a bashism that accepts parameters (e.g. `source <script> --flag`). Posix shell doesn't have that, and `.` doesn't support passing params, so this is a hack to get similar functionality by utilizing [set](https://pubs.opengroup.org/onlinepubs/009695399/utilities/set.html). We also [backup/restore](https://github.com/snapcore/snapcraft/pull/2847/files#diff-3cc26b49a70ee86523598a96dd385301R624-R625) any args passed to the ROS app as well so the script doesn't consume them instead.

cjp256: Oh wow, I have never seen `set` used to modify positional arguments.  TIL :wink:

PR #2848, opened 2019-12-18
common: generate run scripts which can execute independently

cjp256:
To support debugging, update run() to create a script that accounts
for the parameters `cwd` and `env` passed along to check_call(),
stripping those parameters (if present).

No longer remove script generated, but instead persist it in
/tmp/snapcraft-<pid>/run-<#>.sh for optional debugging.

Now a developer can look at the script and re-run it independently of
snapcraft for debugging/development purposes.

-----

### DIFF
--- a/snapcraft/internal/common.py
+++ b/snapcraft/internal/common.py
@@ -19,6 +19,7 @@ import glob
 import logging
 import math
 import os
+from pathlib import Path
 import shlex
 import shutil
 import subprocess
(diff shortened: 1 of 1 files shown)


snapcraft/internal/common.py
```diff
@@ -57,21 +58,73 @@ def assemble_env():
...
+
+    # Set shell.
+    lines.append("#!/bin/sh")
+
+    # Account for `env` parameter by populating exports.
+    # Ordering matters: assembled_env overrides `env` parameter.
+    cmd_env = kwargs.pop("env", None)
+    if cmd_env:
+        lines.append("#############################")
+        lines.append("# Exported via `env` parameter:")
```
sergiusens: do we really want something like this? I was wishing on moving away from this assembled_env thing completely, seems this change would tie us more to "common".

cjp256: I personally found it quite useful for debugging rust stuff, and will certainly use it again in the future.  The idea is just to provide a tool for debugging commands without trying to guess or waste a lot of time trying to reproduce the environment.  OpenEmbedded/bitbake does a similar thing to capture each build step, and I always have found that incredibly useful in my workflow.

The next best idea I had would be to setup the environment for --debug to reproduce the environment of the failed command that executed.  This was perhaps easier to inspect. :D

When things change, like removing assembled_env, this can be updated, or even removed, of course.  But there's always going to be an environment, specific to a part, step, and command.  It should be easy to reproduce and inspect it.

sergiusens: ok, we can keep it as long as we can remove assembled_env and you have a plan for that! :-)

PR #2850, opened 2019-12-20
hooks: enable command-chain

cjp256:
Enables command-chain regardless of base, but does not actively enable or otherwise use it. It just passes it through if the user configures it in their snapcraft yaml.

### DIFF
--- a/snapcraft/internal/meta/hooks.py
+++ b/snapcraft/internal/meta/hooks.py
@@ -14,11 +14,11 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-from collections import OrderedDict
 import re
+from collections import OrderedDict
+from typing import Any, Dict, List, Optional
 
 from snapcraft.internal.meta.errors import HookValidationError
-from typing import Any, Dict, List, Optional
 
 
 class Hook:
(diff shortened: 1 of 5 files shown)


snapcraft/internal/meta/hooks.py
```diff
@@ -14,20 +14,32 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+from collections import OrderedDict
```
sergiusens: I would prefer if the most specific import (`from ...`) came after the broader import of "re"

snapcraft/internal/meta/hooks.py
```diff
@@ -14,20 +14,32 @@
...
 import re
 
-from copy import deepcopy
 from snapcraft.internal.meta.errors import HookValidationError
-from typing import Any, Dict
+from typing import Any, Dict, List, Optional
```
sergiusens: This import here does not follow PEP8, there is no check for this, but we do follow it across the codebase: https://www.python.org/dev/peps/pep-0008/#imports

snapcraft/internal/meta/hooks.py
cjp256: Fixed!

snapcraft/internal/meta/hooks.py
cjp256: Fixed!

PR #2852, opened 2020-01-06
wstool: don't rely on host git

kyrofa:

-----

The Catkin plugin utilizes wstool to support rosinstall files. While wstool is fetched into its own install area so as to not contaminate either the host or the snap, it doesn't properly use the git that is fetched along with it. This means that merging rosinstall files that include git repositories only works if git is already installed on the host. If the host doesn't include git, it errors out.

This PR properly sets the variables necessary to run git from wstool's install area, thereby no longer relying on the host already having git installed.

### DIFF
--- a/snapcraft/plugins/_ros/wstool.py
+++ b/snapcraft/plugins/_ros/wstool.py
@@ -163,9 +163,12 @@ class Wstool:
 
         env["PATH"] += ":" + os.path.join(self._wstool_install_path, "usr", "bin")
 
-        if "LD_LIBRARY_PATH" not in env:
-            env["LD_LIBRARY_PATH"] = ""
-        env["LD_LIBRARY_PATH"] += ":" + snapcraft.formatting_utils.combine_paths(
+        if "LD_LIBRARY_PATH" in env:
+            env["LD_LIBRARY_PATH"] += ":"
+        ld_library_path = env.get("LD_LIBRARY_PATH", "")
+        env[
+            "LD_LIBRARY_PATH"
+        ] = ld_library_path + snapcraft.formatting_utils.combine_paths(
             snapcraft.common.get_library_paths(
                 self._wstool_install_path, self._project.arch_triplet
             ),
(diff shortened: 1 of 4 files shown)


snapcraft/plugins/_ros/wstool.py
```diff
@@ -162,6 +162,23 @@ def _run(self, arguments: List[str]) -> str:
         env = os.environ.copy()
 
         env["PATH"] += ":" + os.path.join(self._wstool_install_path, "usr", "bin")
+
+        if "LD_LIBRARY_PATH" not in env:
+            env["LD_LIBRARY_PATH"] = ""
+        env["LD_LIBRARY_PATH"] += ":" + snapcraft.formatting_utils.combine_paths(
```
cjp256: Not that it matters a great deal, but this will add an unnecessary ":" if LD_LIBRARY_PATH is unset.  

I would encourage the direct use of `":".join(...)` instead of combine_paths...

kyrofa: `":".join()` doesn't make this prettier if `LD_LIBRARY_PATH` is unset, you still need to filter out a `None` or an empty string, depending on how it's written. If your concern is the extra colon, however, I've pushed a commit that remedies it.

tests/spread/plugins/catkin/catkin-with-python/task.yaml
```diff
@@ -22,5 +22,9 @@ restore: |
 execute: |
   cd "$SNAP_DIR"
 
+  # rosinstall files shouldn't require git to be installed, and this test is only
+  # completely valid if it ISN'T installed
+  apt remove --yes git || true
```
sergiusens: should be apt-get

kyrofa: Don't they do the same thing in this case?

sergiusens: They use different backend implementations and we pollute the logs with the "do not use in scripts" message

kyrofa: Ah that makes sense, thank you for explaining. Fixed.

PR #2858, opened 2020-01-08
add support for system-usernames

cjp256:
LP #1845267

### DIFF
--- a/snapcraft/internal/meta/snap.py
+++ b/snapcraft/internal/meta/snap.py
@@ -27,6 +27,7 @@ from snapcraft.internal.meta.application import Application
 from snapcraft.internal.meta.hooks import Hook
 from snapcraft.internal.meta.plugs import ContentPlug, Plug
 from snapcraft.internal.meta.slots import ContentSlot, Slot
+from snapcraft.internal.meta.system_user import SystemUser, SystemUserScope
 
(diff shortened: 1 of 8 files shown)


cjp256:
Added the remaining functionality to implement and schema tests. It's a little ugly due to "system-usernames" not matching the __dict__ model, but that's going to be removed & cleaned up in an upcoming PR.

snapcraft/internal/meta/snap.py
```diff
@@ -267,6 +279,34 @@ def from_dict(cls, snap_dict: Dict[str, Any]) -> "Snap":
...
+                        )
+                    elif isinstance(raw_user, str):
+                        try:
+                            scope = SystemUserScope[raw_user.upper()]
+                        except KeyError:
+                            raise errors.SystemUsernamesValidationError(
+                                name=user_name, message=f"scope {raw_user!r} is invalid"
+                            )
+                        user = SystemUser(name=user_name, scope=scope)
+                    else:
```
sergiusens: if this is part of passthrough, it probably should be allowed to have other values here.

cjp256: I'm not sure what you mean?

sergiusens: no worries, we can cross that bridge later.

PR #2859, opened 2020-01-08
meson plugin: add support for core20

sd-hd:
add core20


sergiusens:
This plugin will come in the form of PluginV2
Thanks for the contribution.

PR #2861, opened 2020-01-09
meta: remove Application's `prepend_command_chain`

cjp256:
Remove `prepend_command_chain` member - it's ambigiuous by name and
makes more sense to just modify the command-chain directly.

Rename and simplify snap packaging's `_generate_command_chain` ->
`_generate_snapcraft_runner` as it really just creates the runner,
if required.

-----

### DIFF
--- a/tests/unit/meta/test_application.py
+++ b/tests/unit/meta/test_application.py
@@ -89,6 +89,34 @@ class AppCommandTest(unit.TestCase):
         self.expectThat("command-foo.wrapper", FileExists())
         self.expectThat("stop-command-foo.wrapper", FileExists())
 
+    def test_massaged_core(self):
+        app = application.Application.from_dict(
+            app_name="foo", app_dict={"command": "$SNAP/test-command"}
+        )
+        app.prime_commands(base="core", prime_dir=self.path)
+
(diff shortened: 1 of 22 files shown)


cjp256:
I pulled out the fix so we can apply this with new tests and backport the fix.

Requires https://github.com/snapcore/snapcraft/pull/2864

tests/unit/meta/test_application.py
```diff
@@ -64,33 +64,6 @@ def test_app_no_change(self):
         self.expectThat("command-foo.wrapper", Not(FileExists()))
         self.expectThat("stop-command-foo.wrapper", Not(FileExists()))
 
-    def test_mangling(self):
```
sergiusens: Not saying it is wrong to remove, but what is the rationale for removing them?

cjp256: Oh it is wrong to remove ;) Sorry, I meant to go back and reconsider these tests, but forgot!  I originally thought they were testing things that were already covered elsewhere and didn't make sense to me by name.   Looking at it again, I see they actually should remain, but rename mangling -> massaging.  It's really testing that $SNAP is stripped for this case for core18, but not anything outside of massaged bases (e.g. core20).

In actu

PR #2878, opened 2020-01-15
extensions: fix merging of build-environment

cjp256:
extensions: fix merging of build-environment
    
    When merging "build-environment", merge_lists will fail
    because OrderedDict is unhashable when calculating the "seen"
    set.
    
    In addition to this failure, build-environment should be merged
    with the extension definitions being listed first, followed by
    the user definitions.  So we append the lists in that order.
    
    Add tests to show coverage for "build-environment" using the
    part_snippet.

### DIFF
--- a/tests/unit/project_loader/extensions/test_utils.py
+++ b/tests/unit/project_loader/extensions/test_utils.py
@@ -32,6 +32,7 @@ class ExtensionTestBase(ProjectLoaderBaseTest):
         super().setUp()
 
         # Create a few fake extensions
+        self.useFixture(_build_environment_extension_fixture())
         self.useFixture(_environment_extension_fixture())
         self.useFixture(_plug_extension_fixture())
         self.useFixture(_plug2_extension_fixture())
@@ -250,6 +251,12 @@ class ExtensionMergeTest(ExtensionTestBase):
                     "command": "echo 'hello'",
                     "plugs": ["foo", "test-plug"],
                 },
+                "part_definition": textwrap.dedent(
+                    """\
+                    plugin: nil
+                    """
+                ),
+                "expected_part_definition": {"plugin": "nil", "prime": [], "stage": []},
             },
         ),
         (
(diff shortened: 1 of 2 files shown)


hellsworth:
I think this change improves the readability of the code. +1 for merging

cjp256:
Pushed update to slightly simplify the implementation and added tests.

tests/unit/project_loader/extensions/test_utils.py
```diff
@@ -675,6 +728,29 @@ def __init__(self, extension_name, yaml_data):
...
+        def get_supported_confinement() -> Tuple[str, ...]:
+            return ("strict",)
+
+        def __init__(self, extension_name, yaml_data):
+            super().__init__(extension_name=extension_name, yaml_data=yaml_data)
+            self.root_snippet = {}
+            self.app_snippet = {}
+            self.part_snippet = {
+                "after": ["extension-part"],
+                "build-environment": [{"PATH": "$PATH:/bar"}, {"EXTKEY": "EXTVAL"}],
```
sergiusens: assign $PATH/extensions-bin to make it easier to identify in the test.

tests/unit/project_loader/extensions/test_utils.py
```diff
@@ -267,6 +274,41 @@ class ExtensionMergeTest(ExtensionTestBase):
...
+                    command: echo 'hello'
+                    extensions: [buildenvironment]
+                    """
+                ),
+                "expected_app_definition": {"command": "echo 'hello'"},
+                "part_definition": textwrap.dedent(
+                    """\
+                    plugin: nil
+                    build-environment:
+                      - PATH: "$PATH:/foo"
```
sergiusens: s/foo/part-bin (or part-path), keep the suffix consistent with the part and extension suffix

cjp256:
> Minor fixes, and keep in mind that for build-environment we want the extension's environment to come first.

To clarify, you don't want this to be done generically for merge_lists?  Just add a special-case handling for the "build-environment" property?

tests/unit/project_loader/extensions/test_utils.py
cjp256: good idea, will do

tests/unit/project_loader/extensions/test_utils.py
cjp256: good idea, will do

cjp256:
Closing in favor of #2882

PR #2882, opened 2020-01-16
extensions: change extension merge-strategy to fix build-environment

cjp256:
When merging "build-environment", merge_lists will fail
because OrderedDict is unhashable when calculating the "seen"
set.

- When merging lists, just append the user-defined list after the
extension-defined list.  If the merged list is is comprised of just
strings, strip duplicates.

This ordering has changed from previous behavior, where user-defined
items were first, then the extension.  This ripples to multiple extensions
and seeing the culmative effects (i.e. ExtensionOrderConsistencyTest).

- Add tests for coverage for `build-environment`.

### DIFF
--- a/tests/unit/project_loader/extensions/test_utils.py
+++ b/tests/unit/project_loader/extensions/test_utils.py
@@ -32,6 +32,8 @@ class ExtensionTestBase(ProjectLoaderBaseTest):
         super().setUp()
 
         # Create a few fake extensions
+        self.useFixture(_build_environment_extension_fixture())
+        self.useFixture(_build_environment2_extension_fixture())
         self.useFixture(_environment_extension_fixture())
         self.useFixture(_plug_extension_fixture())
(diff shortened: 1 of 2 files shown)


tests/unit/project_loader/extensions/test_utils.py
```diff
@@ -195,7 +197,7 @@ def test_multiple_extensions(self):
         self.expectThat(config.data["environment"], Contains("TEST_EXTENSION"))
 
 
-class ExtensionOrderConsistencyTest(ExtensionTestBase):
+class MultipleExtensionsExtendPlugs(ExtensionTestBase):
```
sergiusens: we changed the order, but it does not mean we should change the test this way.
Ideally the scenario should include the expected result and it should represent the order in which the extensions were added.

cjp256: Done!

PR #2885, opened 2020-01-18
lifecycle: only warn when a default provider snap is missing

sergiusens:
Snapcraft gained the ability to check for artifacts from snaps listed as plugs
when using the content interface.

The side effect, is that the snap needs to be published on the main Snap Store,
this is a requirement brand stores cannot meet, so instead of failing,
Snapcraft resorts to warning instead.

-----

### DIFF
--- a/snapcraft/internal/lifecycle/_runner.py
+++ b/snapcraft/internal/lifecycle/_runner.py
@@ -82,10 +82,10 @@ def execute(
             "The repo backend is not returning the list of installed packages"
         )
 
+    build_snaps = project_config.build_snaps
     content_snaps = project_config.project._get_content_snaps()
-    required_snaps = project_config.build_snaps | content_snaps
 
-    if common.is_process_container():
+    if common.is_process_container() and build_snaps:
         installed_snaps: List[str] = []
         logger.warning(
             (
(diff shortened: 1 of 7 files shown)


snapcraft/internal/lifecycle/_runner.py
```diff
@@ -82,23 +82,32 @@ def execute(
...
         )
     else:
-        installed_snaps = repo.snaps.install_snaps(required_snaps)
+        installed_snaps = repo.snaps.install_snaps(build_snaps)
+        for content_snap in content_snaps:
+            try:
+                installed_snaps += repo.snaps.install_snaps([content_snap])
+            except repo.snaps.errors.SnapUnavailableError:
+                logger.warning(
+                    f"Could not install snap defined in a plug {content_snap!r}. "
```
cjp256: nit: s/in a plug/in plug/

snapcraft/internal/lifecycle/_runner.py
```diff
@@ -82,23 +82,32 @@ def execute(
...
+                    "The missing library report may have false positives listed if those "
+                    "libraries were to be provided by the content snap."
```
cjp256: nit: s/were to be/are

PR #2889, opened 2020-01-22
meta: always generate snapcraft-runner to workaround classic PATH bug

cjp256:
Snapcraft previously generated wrappers for most commands.  Once
snapcraft became more conserative about generating unnecessary wrappers,
issues began to surface with regard to PATH being set incorrectly
for classic snap apps that did not use shell.

To work around the issue, this commit adds an empty snapcraft-runner to
the command-chain for all apps, where they typically would have none.
Once this issue is resolved in snapd, we can probably remove this
workaround.

LP#: 1860369

-----

### DIFF
--- a/snapcraft/internal/meta/_snap_packaging.py
+++ b/snapcraft/internal/meta/_snap_packaging.py
@@ -438,8 +438,8 @@ class _SnapPackaging:
 
         Return path relative to prime directory, if created."""
 
-        # If there are no apps, no need to create a runner.
-        if not self._snap_meta.apps:
+        # If there are no apps, or type is snapd, no need to create a runner.
+        if not self._snap_meta.apps or self._config_data.get("type", "") == "snapd":
             return None
 
         # Classic confinement or building on a host that does not match the target base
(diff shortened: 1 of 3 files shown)


snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -438,36 +438,42 @@ def _generate_snapcraft_runner(self) -> Optional[str]:
 
         Return path relative to prime directory, if created."""
 
+        # If there are no apps, no need to create a runner.
+        if not self._snap_meta.apps:
```
sergiusens: can you take the snapd snap type into account as well?

cjp256: Done

sergiusens:
Can we add a couple of unit and integration tests for this?

cjp256:
Tests added.


Did not provider coverage of  self._is_host_compatible_with_base check, because that's simply broken.  It points to a function (which requires `base` parameter).  I can fix that in a separate PR, but I'm wondering if the environment really will actually be wrong?  Should we just remove that check or fix  `_is_host_compatible_with_base`?

PR #2890, opened 2020-01-22
extensions: add opengl extension to support classic and strict

cjp256:
Based off the work done by Alan Griffiths:
https://github.com/MirServer/egmde-snap/blob/master/snap/snapcraft.yaml

In my prior attempts to make a classic OpenGL snap, I lacked the
"build-attributes": ["no-patchelf"] requirement for libgl1-mesa-dri and
the crashes were hard to diagnose.

Properly adding OpenGL support for snaps is a non-trivial excercise.
This is an attempt to create a single extension which can support
multiple types of snaps (classic & strict) and do the right thing
for OpenGL as the optimal approach to OpenGL changes over time.

---

A test case I have used for it: 
https://github.com/cjp256/alacritty/blob/snap-updates/extra/linux/snap/snapcraft.yaml

I am very much hoping for feedback for improvements, I am no expert on this! :smile:

### DIFF
--- a/snapcraft/internal/project_loader/_extensions/opengl.py
+++ b/snapcraft/internal/project_loader/_extensions/opengl.py
@@ -1,0 +1,100 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2018-2020 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Import types and tell flake8 to ignore the "unused" List.
+
+import logging
+from typing import Any, Dict, Tuple
+
+from ._extension import Extension
+
+
+logger = logging.getLogger(__name__)
+
+
+class ExtensionImpl(Extension):
+    """This extension eases creation of snaps that integrate with OpenGL.
+
+    For easier desktop integration, it also configures each application
+    entry with these additional plugs if strictly confined (or devmode):
+
+    \b
+    - opengl (https://snapcraft.io/docs/opengl-interface)
+    """
+
+    @staticmethod
+    def get_supported_bases() -> Tuple[str, ...]:
+        return ("core18",)
+
+    @staticmethod
+    def get_supported_confinement() -> Tuple[str, ...]:
+        return ("strict", "devmode", "classic")
+
+    def __init__(self, *, extension_name: str, yaml_data: Dict[str, Any]) -> None:
+        super().__init__(extension_name=extension_name, yaml_data=yaml_data)
+
+        logger.warning(
+            "*EXPERIMENTAL*: The OpenGL extension is experimental and likely to change."
+        )
+
+        if yaml_data.get("confinement") == "classic":
+            # Use environment variables for classic snaps.
+            self.root_snippet = {
+                "environment": {
+                    "__EGL_VENDOR_LIBRARY_DIRS": "$SNAP/etc/glvnd/egl_vendor.d:$SNAP/usr/share/glvnd/egl_vendor.d",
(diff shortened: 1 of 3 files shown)


cjp256:
A test case I have used for it: 
https://github.com/cjp256/alacritty/blob/snap-updates/extra/linux/snap/snapcraft.yaml

AlanGriffiths:
It may be better to do things a bit differently in strict confinement: c.f. https://github.com/MirServer/mir-kiosk-apps/blob/use-layout/snap/snapcraft.yaml - which uses layouts instead of environment variables.

cjp256:
@AlanGriffiths awesome, thanks! I was going to try a strict snap today :D  That's a great use of layouts.

Out of curiosity, is there a reason not to use the same approach with classic?

AlanGriffiths:
> Out of curiosity, is there a reason not to use the same approach with classic?

Yes. With the classic egmde snap (for example) it is normal for to launch programs belonging to the host environment from the snap. Those need to get a clean environment.

It's easier to clean up environment variables to do this:
```yaml
  EGMDE_LAUNCH_PREFIX: "env -u __EGL_VENDOR_LIBRARY_DIRS -u LIBGL_DRIVERS_PATH -u MIR_SERVER_X11_DISPLAY -u MIR_SERVER_X11_DISPLAY_EXPERIMENTAL -u MIR_SERVER_PLATFORM_PATH -u MIR_SERVER_XWAYLAND_PATH -u EGMDE_LAUNCH_PREFIX --"
```

cjp256:
> > Out of curiosity, is there a reason not to use the same approach with classic?
> 
> Yes. With the classic egmde snap (for example) it is normal for to launch programs belonging to the host environment from the snap. Those need to get a clean environment.
> 
> It's easier to clean up environment variables to do this:
> 
> ```yaml
>   EGMDE_LAUNCH_PREFIX: "env -u __EGL_VENDOR_LIBRARY_DIRS -u LIBGL_DRIVERS_PATH -u MIR_SERVER_X11_DISPLAY -u MIR_SERVER_X11_DISPLAY_EXPERIMENTAL -u MIR_SERVER_PLATFORM_PATH -u MIR_SERVER_XWAYLAND_PATH -u EGMDE_LAUNCH_PREFIX --"
> ```

Great points, thanks!

snapcraft/internal/project_loader/_extensions/opengl.py
```diff
@@ -0,0 +1,100 @@
...
+            }
+        else:
+            # Use plugs and layouts for strict/devmode.
+
+            self.app_snippet = {"plugs": ["opengl"]}
+
+            self.root_snippet = {
+                "layout": {
+                    "/etc/glvnd": {"bind": "$SNAP/etc/glvnd"},
+                    "/usr/lib/${SNAPCRAFT_ARCH_TRIPLET}/dri": {
```
diddledan: Should there be a `stage-packages` added someplace to add these files? Without adding them in the extension we expect people to know that `extensions: [opengl]` is not enough to get opengl working, which is counter to their expectations.

snapcraft/internal/project_loader/_extensions/opengl.py
```diff
@@ -0,0 +1,100 @@
...
+            self.root_snippet = {
+                "environment": {
+                    "__EGL_VENDOR_LIBRARY_DIRS": "$SNAP/etc/glvnd/egl_vendor.d:$SNAP/usr/share/glvnd/egl_vendor.d",
+                    "LIBGL_DRIVERS_PATH": "$SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET/dri",
+                }
+            }
+
+            # Mesa must use no-patchelf, otherwise driver will crash.
+            # TODO: develop/upstream patch to fix mesa.
+            self.parts = {
```
diddledan: Why only adding this in classic confinement?

cjp256: In this case, mesa is being added in this way only to ensure `no-patchelf` attribute is set for classic.  AFAIK, it's only a problem for classic snaps..?

snapcraft/internal/project_loader/_extensions/opengl.py
cjp256: Good catch sir, I never really firmed up the strict support.

Which packages would you expect to see here? Same as classic? ["libgl1-mesa-dri", "libglx-mesa0"] ?

snapcraft/internal/project_loader/_extensions/opengl.py
cjp256: On second thought, this could (should?) be shared for both.   no-patchelf is basically a no-op in the strict case, and the `prime` cleanup would be nice to have.

PR #2891, opened 2020-01-24
meta: always generate snapcraft-runner to workaround classic (#2889)

sergiusens:
Snapcraft previously generated wrappers for most commands.  Once
snapcraft became more conserative about generating unnecessary wrappers,
issues began to surface with regard to PATH being set incorrectly
for classic snap apps that did not use shell.

To work around the issue, this commit adds an empty snapcraft-runner to
the command-chain for all apps, where they typically would have none.
Once this issue is resolved in snapd, we can probably remove this
workaround.

Add unit tests for snapcraft-runner generation

Add some additional parameters to the SnapcraftYaml fixture
to make it more configurable.

LP#: 1860369

-----

### DIFF
--- a/snapcraft/internal/meta/_snap_packaging.py
+++ b/snapcraft/internal/meta/_snap_packaging.py
@@ -432,7 +432,7 @@ class _SnapPackaging:
             )
 
     def _generate_command_chain(self) -> List[str]:
-        command_chain = list()
+        command_chain: List[str] = list()
 
         # command-chain is not required in these situations.
         if (
--- a/snapcraft/internal/meta/_snap_packaging.py
+++ b/snapcraft/internal/meta/_snap_packaging.py
@@ -433,38 +433,43 @@ class _SnapPackaging:
 
     def _generate_command_chain(self) -> List[str]:
         command_chain = list()
-        # Classic confinement or building on a host that does not match the target base
-        # means we cannot setup an environment that will work.
+
+        # command-chain is not required in these situations.
         if (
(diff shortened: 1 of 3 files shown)


snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -433,38 +433,43 @@ def setup_assets(self) -> None:
 
     def _generate_command_chain(self) -> List[str]:
         command_chain = list()
-        # Classic confinement or building on a host that does not match the target base
-        # means we cannot setup an environment that will work.
+
+        # command-chain is not required in these situations.
         if (
-            self._config_data["confinement"] == "classic"
+            self._config_data.get("type", "app") == "snapd"
```
cjp256: I don't think specifying a default for .get() matters.  `None` != `"snapd"`.

snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -433,38 +433,43 @@ def setup_assets(self) -> None:
 
     def _generate_command_chain(self) -> List[str]:
         command_chain = list()
```
cjp256: mypy wants the type info here now...

snapcraft/internal/meta/_snap_packaging.py
sergiusens: it does not, but the default behavior for not having it specified is "app", I thought this would make it clear what the outcome is when not set.

PR #2894, opened 2020-01-25
Fix issue with multipass mount on win32

NickZ:

-----

os.getuid() does not exist on win32, so it will crash when called. I just disable the UID and GID map by setting it to `{"0":"0"}` if it's win32.

### DIFF
--- a/snapcraft/internal/build_providers/_multipass/_multipass.py
+++ b/snapcraft/internal/build_providers/_multipass/_multipass.py
@@ -16,7 +16,6 @@
 
 import logging
 import os
-import platform
 import sys
 from typing import Dict, Optional, Sequence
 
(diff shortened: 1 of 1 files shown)


snapcraft/internal/build_providers/_multipass/_multipass.py
```diff
@@ -198,9 +199,12 @@ def _mount(self, host_source: str, target: str) -> None:
             return
 
         target = "{}:{}".format(self.instance_name, target)
-        uid_map = {str(os.getuid()): "0"}
-        gid_map = {str(os.getgid()): "0"}
-
+        if platform.system() != "Windows":
```
cjp256: nit: the rest of the code base is using `if sys.platform == "win32"` for this check, perhaps there's some value in consistency.

Either way, :+1: Thank you @NickZ!

I haven't given Windows attention in a while and need to circle back now...  Are you having success with it? :-)

PR #2909, opened 2020-02-04
elf: search for host libraries within search paths

sergiusens:
A regression was introduced with commit 76e8a3f3f1c3f69d957b2806c9387ac3c57e926a
which considered elf files on the host system to be valid. Additionally, since
the host-found libraries were absolute paths, an os.path.join suffixing the
search paths snapcraft cared about were ignored.

A unit test which considers this scenario was added.

LP: #1860766

-----

### DIFF
--- a/snapcraft/internal/elf.py
+++ b/snapcraft/internal/elf.py
@@ -211,7 +211,7 @@ class Library:
 
         valid_search_paths = [p for p in self.search_paths if os.path.exists(p)]
         in_search_paths = any(
-            [self.soname_path.startswith(p) for p in valid_search_paths]
+            self.soname_path.startswith(p) for p in valid_search_paths
         )
 
         # Expedite path crawling if we have a valid elf file that lives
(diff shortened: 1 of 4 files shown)


snapcraft/internal/elf.py
```diff
@@ -209,18 +209,23 @@ def _crawl_for_path(self) -> str:
 
         logger.debug("Crawling to find soname {!r}".format(self.soname))
 
-        if self._is_valid_elf(self.soname_path):
+        valid_search_paths = [p for p in self.search_paths if os.path.exists(p)]
+        in_search_paths = any(
+            [self.soname_path.startswith(p) for p in valid_search_paths]
```
cjp256: Not that it matters here, but I think in this would be a good case for a generator (lose the list brackets):
`any(self.soname_path.startswith(p) for p in valid_search_paths)`

sergiusens: I am a fan of generators in general

PR #2923, opened 2020-02-09
requirements: Update PyYAML requirement to 5.3

NickZ:

-----

They are no longer providing updated wheels for windows for newer python versions (>3.7). This updates snapcraft to use 5.3 instead, which is the current version and is being provided updated wheels for newer python versions.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,9 +3,7 @@ click-config-file==0.5.0
 configparser==3.5.0
 jsonschema==2.5.1
 progressbar33==2.4
-https://pyyaml.org/download/pyyaml/PyYAML-5.3-cp37-cp37m-win_amd64.whl; sys_platform == 'win32' and python_version == "3.7"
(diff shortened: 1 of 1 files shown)


cjp256:
I did some testing and i don't think you need to point to the wheels.

`PyYAML==5.3` worked just fine for me with Windows and python3.8.

NickZ:
It installs, yes, but I had previously gotten errors when using snapcraft with the version being installed from pip complaining that it wasn't compiled with libyaml bindings; was that not the case for you?

cjp256:
> It installs, yes, but I had previously gotten errors when using snapcraft with the version being installed from pip complaining that it wasn't compiled with libyaml bindings; was that not the case for you?

I think snapcraft raised a RuntimeError of that sort when I accidentally installed the old version that way (or something like that, I'm pretty sure I noted the old version at the time?).  I can try it again.

cjp256:
FWIW In my virtualenv I uninstalled pyyaml (re-)installed.  Then re-built the subedit snap OK.

NickZ:
ah, it must be because 5.3 actually has the windows wheels in pip; I'll
take those out then.

PR #2926, opened 2020-02-11
plugin handler: process elf files only if base is specified

cjp256:
ELF files are always processed when `type: app`, where `app`
is the default type.

This causes an issue when snapd is building without specifying
the type, as it gets treated as `type: app` and its files
are processed like typical app snaps.  The schema currently
allows for this because it fails to set the default type `app`
in time for validation.

Since snapd does not have a base, check for the existence of the
base before attempting to process ELF files.  This will prevent
the error reported in LP #1857019 and allow snapd to build without
a type specified.

Once snapd moves to `type: snapd`, we can fix the schema to
ensure that the invalid configuration that led to this error
does not pass verification.

LP #1857019
LP #1862642

-----

### DIFF
--- a/snapcraft/internal/pluginhandler/__init__.py
+++ b/snapcraft/internal/pluginhandler/__init__.py
@@ -815,7 +815,7 @@ class PluginHandler:
         snap_files, snap_dirs = self.migratable_fileset_for(steps.PRIME)
         _migrate_files(snap_files, snap_dirs, self.stagedir, self.primedir)
 
-        if self._snap_type == "app":
+        if self._snap_type == "app" and self._base is not None:
             dependency_paths = self._handle_elf(snap_files)
         else:
             dependency_paths = set()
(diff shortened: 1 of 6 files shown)


snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -815,7 +815,7 @@ def _do_prime(self) -> None:
         snap_files, snap_dirs = self.migratable_fileset_for(steps.PRIME)
         _migrate_files(snap_files, snap_dirs, self.stagedir, self.primedir)
 
-        if self._snap_type == "app":
+        if self._snap_type == "app" and self._base:
```
sergiusens: I don't know why, but lately I prefer `is not None` for explicit `None` checks.
This is fine though, just thought I'd mention it.

cjp256: Done

cjp256:
Added spread test.

anonymouse64:
I'm a little late to report this, but I built https://github.com/snapcore/snapd/pull/7904 with this PR (the `--channel=edge/pr-2926` thing is really cool :+1: ) and it worked fine, so we are unblocked on the snapd side as soon as this lands in the candidate channel.

PR #2927, opened 2020-02-11
requirements: uprev lxml to 4.5.0

cjp256:
Should now fetch wheel for Windows and Linux.

-----

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,8 +87,6 @@ jobs:
             - libnacl-dev
             - libsodium-dev
             - libssl-dev
-            - libxml2-dev
-            - libxslt-dev
             - libyaml-dev
             - python3.6-dev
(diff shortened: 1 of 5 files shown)


cjp256:
Looks it it builds fine without those stage packages.  Removed!

sergiusens:
Are we sure we are using C bindings and not some slower implementation in python by going down this path?

cjp256:
I don't think so.  But I think my original PR made the most sense, so I'm going to restore that.

Even if we don't need the `-dev` packages, I think we should still install them (in case the wheel becomes no longer available).

It's probably picking up libxml2 and xslt from the host.

cjp256:
Tested the latest push (w/o xslt and xml libs) against #2952 to verify appstream is OK after removal in a minimal env.

PR #2930, opened 2020-02-11
extensions: Handle case when user-dirs.dirs exits but not user-dirs.locale

galgalesh:
This is a fix for bug https://github.com/ubuntu/snapcraft-desktop-helpers/issues/173

This reduces hot run time for `desktop-launch` on *affected systems*. Depending on the snap, this can remove 1 - 10 seconds off its hot start time.

`desktop-launch` keeps needlessly setting `needs_xdg_reload=true` when `~/.config/user-dirs.dirs` exist but `~/.config/user-dirs.locale` doesn't. This happens when a user gets the `xdg-user-dirs-gtk` dialog (after changing the system language) and checks the "Don't ask me this again" option, which removes `~/.config/user-dirs.locale` .

This patch makes sure we only set `needs_xdg_reload=true` when something actually changes and removes the hash of `~/.config/user-dirs.locale` when the file itself is removed.

Note: depending on what logic runs when `needs_xdg_reload=true`, this will have more or less impact. Snaps using extensions don't generate the mime-cache, so they won't see a drastic change. Fixing this in the `desktop-helpers` repo will have a drastic impact on snaps such as VLC (- 10 seconds hot start time), but I want to get it merged here first.

-----

### DIFF
--- a/extensions/desktop/common/desktop-exports
+++ b/extensions/desktop/common/desktop-exports
@@ -155,14 +155,19 @@ if [ "$HOME" != "$SNAP_USER_DATA" ] && ! is_subpath "$XDG_CONFIG_HOME" "$HOME"; 
   done
 fi
 
-if can_open_file "$REALHOME/.config/user-dirs.dirs" && can_open_file "$REALHOME/.config/user-dirs.locale"; then
+if can_open_file "$REALHOME/.config/user-dirs.dirs"; then
   # shellcheck disable=SC2154
   if [ "$needs_update" = true ] || [ "$needs_xdg_reload" = true ]; then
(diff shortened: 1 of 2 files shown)


sergiusens:
Thanks for the contribution. Going to let the desktop-knowledgeable folks do a first pass though (requested as reviewers)

hellsworth:
LGTM! Just one minor comment on wording.

galgalesh:
@hellsworth I can't seem to find your comment on the wording. Where did you post it?

hellsworth:
comment

@sergiusens @galgalesh not sure why this isn't showing up for you two but my here's my comment :)

PR #2932, opened 2020-02-12
elf: resolve paths in `ldd()` to purge relative path components

cjp256:
elf: resolve paths in `ldd()` to purge relative path components

ldd may return paths with relative components such as "..", e.g.:
{
    'linux-vdso.so.1': '',
    'libc.so.6': '/snap/core/current/lib/x86_64-linux-gnu/libc.so.6',
    'libnvidia-container.so.1': '/root/prime/usr/bin/../lib/x86_64-linux-gnu/libnvidia-container.so.1',
    'libcap.so.2': '/snap/core/current/lib/x86_64-linux-gnu/libcap.so.2',
    'libdl.so.2': '/snap/core/current/lib/x86_64-linux-gnu/libdl.so.2',
    'libseccomp.so.2': '/snap/core/current/lib/x86_64-linux-gnu/libseccomp.so.2'
}

(1) Resolve the paths before using them to ensure we have a consistent
values that do not look odd in potential warnings/errors (e.g.
corrupted ELF).

This also appears to expose an issue with the FakeElf fixture which
was not creating a fake "barsnap.so.2" in the core base path.  This
updates the fixture to create one in the same manner as it does for
the root libraries.

(2) In the same example above, we can see linux-vdso has an empty string.
In Ubuntu 16.04, ldd will return a slightly different formatted string
which results in an empty path for the "found" shared object path. I do
not believe it has caused a problem before attempting to use abspath
on it.

Check for the empty-string case and ignore it.  We don't need to check
linux-vdso, and calling abspath on it will resolve to the current
working directory.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -104,19 +104,3 @@ parts:
         "$SNAPCRAFT_PART_INSTALL/usr/bin/python3" -m compileall -q .
 
     after: [snapcraft-libs]
-
-  legacy-snapcraft:
-    plugin: python
-    source: https://github.com/snapcore/snapcraft.git
-    source-branch: legacy
-    source-depth: 1
-    requirements:
-        - requirements.txt
-    override-build: |
-        snapcraftctl build
-        $SNAPCRAFT_PROJECT_DIR/tools/snapcraft-override-build.sh
-
-        sed -ri 's|(lib/.*/site-packages)|legacy_snapcraft/\1|' $SNAPCRAFT_PART_INSTALL/usr/lib/python3.6/sitecustomize.py
-    organize:
-        '*': legacy_snapcraft/
-    after: [snapcraft-libs]
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -104,3 +104,19 @@ parts:
         "$SNAPCRAFT_PART_INSTALL/usr/bin/python3" -m compileall -q .
 
     after: [snapcraft-libs]
+
+  legacy-snapcraft:
+    plugin: python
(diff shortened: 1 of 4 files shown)


cjp256:
Still debugging this error. For some reason there appears to be an empty string lib getting returned in the crystal case, which expands to the cwd by abspath.

snap/snapcraft.yaml
```diff
@@ -104,19 +104,3 @@ parts:
         "$SNAPCRAFT_PART_INSTALL/usr/bin/python3" -m compileall -q .
 
     after: [snapcraft-libs]
-
```
sergiusens: why?

cjp256: whoops, did not mean to include that.

cjp256:
> Seems to be a rogue change.
> Also, can you please do successive commits instead of amends to make it easier to review? Spotting the change is a task on itself.

Will do, sorry about that. I hadn't thought you reviewed it yet.

sergiusens:
I just did not assess, but I have been looking at this since it first came out :-)

snapcraft/internal/elf.py
```diff
@@ -35,6 +35,21 @@
 logger = logging.getLogger(__name__)
 
 
+def _ldd_resolve(soname: str, soname_path: str) -> Tuple[str, str]:
+    logger.debug(f"_ldd_resolve: {soname!r} {soname_path!r}")
+
+    # If found, resolve the path components.
+    if soname_path.startswith("/"):
```
sergiusens: can you add a comment on why this check is important? Or example output

cjp256: will do

PR #2933, opened 2020-02-12
remote-build: introduce --launchpad-snapcraft-channel option

cjp256:
It may be useful to configure Launchpad to use a specific
snapcraft channel.  This commit changes the default from "edge"
to "stable", and allows the user to configure it.

-----

### DIFF
--- a/snapcraft/cli/remote.py
+++ b/snapcraft/cli/remote.py
@@ -58,6 +58,15 @@ def remotecli():
     cls=PromptOption,
 )
 @click.option(
+    "--launchpad-snapcraft-channel",
+    default="stable",
+    metavar="<channel>",
+    nargs=1,
+    required=False,
+    type=str,
+    help="Specify Snapcraft channel for Launchpad to build with.",
+)
+@click.option(
     "--launchpad-timeout",
     metavar="<seconds>",
     type=int,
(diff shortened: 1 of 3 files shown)


cjp256:
> We agreed that we would default to the channel that snapcraft was on for this.

"channel that snapcraft was on" being "edge"? Or matching the user's host-installed channel, if available?  I just was under the impression edge was required at some point in order to function, but I'm probably wrong.

Pointing it to edge just feels like the wrong thing to do if the feature is eventually to be considered stable.  But I do like the effect of encouraging more use of edge :D

> We are currently under design review, so let's not change the CLI atm

:disappointed: Does it really matter?  I think exposing this feature is quite useful.  We can address the deltas later (it is after all still warning as experimental).

Regardless, I think it would be wise to share this change with the reviewers and allow them to provide feedback with this in mind before they've completed their review.  Though I don't know who that is.

sergiusens:
Let's just change the default to stable, I somehow missed this.
And the toggle we can discuss in that review.

sergiusens:
If you really want this, use `SNAPCRAFT_BUILD_ENVIRONMENT_CHANNEL_SNAPCRAFT` with a nice warning when used and default to stable.
This should be mostly a developer option.

cjp256:
I will open a new PR with SNAPCRAFT_BUILD_ENVIRONMENT_CHANNEL_SNAPCRAFT.

PR #2942, opened 2020-02-14
pluginhandler: do not search installdir or stagedir for dependencies

cjp256:
The search paths already include primedir which has all of the primed
and staged files at this point.  The installdir and stagedir may
contain files that are not shipped in the final snap, so limit search
to primedir as it should be fully populated.

- Replace _handle_dependencies() with a simpler _warn_missing_dependencies()
and _calculate_dependency_paths() called from _handle_elf().

- Add a couple useful debug prints.

- Add two tests to cover the case where the dependency is found, and
another where the dependency is missing.  For the latter, the lib is
removed using `stage` configuration.

This last test would have previously failed, because the dependency
was found in installdir, and ignored by the missing dependency resolver.

-----

### DIFF
--- a/snapcraft/internal/pluginhandler/__init__.py
+++ b/snapcraft/internal/pluginhandler/__init__.py
@@ -927,9 +927,9 @@ class PluginHandler:
         _clean_migrated_files(primed_files, primed_directories, shared_directory)
 
     def _calculate_dependency_paths(
-        self, split_depencies: Dict[str, Set[str]]
+        self, split_dependencies: Dict[str, Set[str]]
     ) -> Set[str]:
-        part_dependencies: Set[str] = split_depencies.get(self.primedir, set())
+        part_dependencies = split_dependencies.get(self.primedir, set())
         return {os.path.dirname(d) for d in part_dependencies}
 
     def _warn_missing_dependencies(self, split_dependencies: Dict[str, Set[str]]):
--- a/snapcraft/internal/pluginhandler/__init__.py
+++ b/snapcraft/internal/pluginhandler/__init__.py
@@ -65,7 +65,7 @@ class PluginHandler:
         base,
         confinement,
         snap_type,
-        soname_cache
+        soname_cache,
     ) -> None:
         self.valid = False
         self.plugin = plugin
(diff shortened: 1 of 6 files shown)


cjp256:
This is the starter fix, but I think we should take it further and eliminate use of _find_directory().  I don't think we should be attempting to (re)-search for dependencies.

If we should in fact be re-searching installdir and stagedir for things that were lost in priming (but are dependencies), perhaps we should move that logic into MissingDependencyResolver..?

cjp256:
@sergiusens just to confirm - we did agree that this check should simply be moved into MissingDependencyResolver?

What do you think about a message like:

> This part is missing libraries that cannot be satisfied with the currently primed files.  If this is not intended, ensure that these files are not filtered out using the `filesets`, `stage`, or `prime` keywords.  These missing libraries include:
> - libfoo.so.1, found at parts/foo/install/usr/lib/libfoo.so.1
> - libbar.so.1, found at stage/bar/usr/lib/libfoobar.so.1
>

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -921,39 +926,23 @@ def _clean_shared_area(self, shared_directory, part_state, project_state):
...
-    ):
-        # Split the necessary dependencies into their corresponding location.
-        # We'll only track the part and staged dependencies, since they should have
-        # already been primed by other means, and migrating them again could
-        # potentially override the `stage` or `snap` filtering.
-        dirs = [self.plugin.installdir, self.stagedir, self.primedir, *content_dirs]
+    def _calculate_dependency_paths(
+        self, split_depencies: Dict[str, Set[str]]
+    ) -> Set[str]:
+        part_dependencies: Set[str] = split_depencies.get(self.primedir, set())
```
kyrofa: mypy should be able to infer this type, no?

cjp256: Ah yes good point.

tests/spread/general/dependency-check-missing-lib/snaps/missing-lib/snap/snapcraft.yaml
```diff
@@ -0,0 +1,19 @@
...
+confinement: strict
+
+parts:
+  pciutils:
+    plugin: nil
+    build-packages:
+      - pciutils
+    override-build: |
+      mkdir -p $SNAPCRAFT_PART_INSTALL/usr/bin
+      install -m 0755 /usr/bin/lspci $SNAPCRAFT_PART_INSTALL/usr/bin
```
sergiusens: install -D and you can remove the line above

cjp256: did not know of that, done!

PR #2949, opened 2020-02-19
cli: fix clean on Windows

NickZ:

-----
This fixes the `clean` command on Windows; `is_snap_installed()` has been short circuited to return "false" because snapd does not currently work on Windows, and `_project_options()` uses os.cpu_count instead of os_sched_getaffinity because the latter does not exist on Windows.


snapcraft/internal/repo/snaps.py
```diff
@@ -61,6 +61,9 @@ def is_valid_snap(cls, snap):
 
     @classmethod
     def is_snap_installed(cls, snap):
+        # Hack for windows
```
cjp256: I have seen some snaps working under WSL.  But generally speaking, I think returning false here is fine here as they aren't supported on Windows (yet?), with or without WSL.

Why are we checking if snaps are installed on Windows..?  I'd like to see the call stack on this one because I am guessing there is something further up the stack we need to change.  

That said, I'd reword the comment ("Hack for Windows" => "Snaps are unsupported on Windows." or something like that

snapcraft/project/_project_options.py
```diff
@@ -131,6 +131,8 @@ def _get_platform_architecture():
 class ProjectOptions:
     @property
     def parallel_build_count(self) -> int:
+        if sys.platform == "win32":
```
cjp256: this LGTM

snapcraft/internal/repo/snaps.py
NickZ: Here's the callstack: 
image

Here's the most relevant bit: https://github.com/snapcore/snapcraft/blob/master/snapcraft/internal/project_loader/_config.py#L244
It calls it when it checks to see if the base is installed to determine if it should install it (it also checks if its in a container, which Windows will fail)

cjp256: Gotcha, thanks.

It boils down to snapcraft trying to clean up `prime` in case `snapcraft try` was executed.
https://github.com/snapcore/snapcraft/blob/master/snapcraft/cli/lifecycle.py#L355

This should be harmless.  Though `try` and this cleanup case should probably be disabled on Windows.

NickZ: Ok, I have disabled the clean step for the prime directory on Windows.

snapcraft/cli/lifecycle.py
```diff
@@ -352,8 +353,9 @@ def clean(ctx, parts, unprime, step, **kwargs):
                 instance.clean(part_names=parts)
         else:
             build_provider_class(project=project, echoer=echo).clean_project()
-            # Clear the prime directory on the host
-            lifecycle.clean(project, parts, steps.PRIME)
+            # Clear the prime directory on the host, unless on Windows
```
cjp256: nit: we try to punctuate our comments (https://github.com/snapcore/snapcraft/blob/master/CODE_STYLE.md#comments)

snapcraft/cli/lifecycle.py
```diff
@@ -336,8 +337,9 @@ def clean(ctx, parts, unprime, step, **kwargs):
...
```
sergiusens, line 340:
```suggestion
            # Clear the prime directory on the host, unless on Windows.
```

PR #2960, opened 2020-03-04
specification: base plugin and plugins for core20

sergiusens:

-----


cjp256:
Also, I think we should add something like get_package_management() to provide configuration.  If the plugin adds build/stage-packages that require additional repos, it should be able to configure them directly without requiring additional work.

20200304-core20-plugins.org
```diff
@@ -0,0 +1,190 @@
...
+        """Return a schema for the plugin properties."""
+
+    def __init__(self, properties: Dict[str, str]) -> None:
+        """
+        :param properties: plugin defined properties.
+        :raises PluginBadBuildEnvironmentError: when invalid build_environment was passed to the plugin.
+        """
+
+    @abstactmethod
+    def get_build_environment(cls) -> Dict[str, str]:
```
cjp256: s/cls/self

20200304-core20-plugins.org
```diff
@@ -0,0 +1,190 @@
...
+"""
+Plugin help
+"""
+
+from typing import Dict, List
+
+import snapcraft
+
+class PluginImpl(snapcraft.plugins.core20.Plugin):
+    def get_build_environment(cls) -> Dict[str, str]:
```
cjp256: s/cls/self

20200304-core20-plugins.org
```diff
@@ -17,8 +18,9 @@ When the =base= in snapcraft is set different one than the previously mentioned
 bases, then snapcraft will use the following path to search for plugins:
 =snapcraft.plugins.<base>= and search for =<plugin-name>.py=.
 
-Plugin class names must be named =Plugin=. The plugin help shall be implemented
-as the *doctring* for the plugin module.
+Plugin class names must be named =PluginImpl= (following the same pattern used
```
cjp256: This still feels wrong to me, at least for in-tree plugins.  I think it could be something as:

That can be used by whomever inits it, or as a factory in core20.Plugin itself, maybe something like
```
class Plugin:
  @classmethod
  def from_part(properties: Dict[str, Any]):
     plugin_name = properties["plugin"]
     plugin_class = CORE20_PLUGIN_MAPPINGS.get(plugin_name)
     if plugin_class is not None:
         return plugin_class(properties)
     plugin_class = cls.find_external_plugin(properties)
     return plugin_class
<snip>

CORE20_PLUGIN_MAPPINGS: Dict[str, Type[Plugin]] = dict(
    autotools=plugins.core20.autotools.AutotoolsPlugin,
    catkin=plugins.core20.catkin.CatkinPlugin,
    colcon=plugins.core20.colcon.ColconPlugin,
}


```

In the case of out-of-tree plugins, named "PluginImpl" seems fair, or whatever it is you want to require to make find_external_plugin() work.

sergiusens: Yeah, dictionary access like we do for source is an implementation detail I was entertaining. I was going for Impl for consistency, but yeah, let's remove

PR #2966, opened 2020-03-09
build providers: move to buildd images

sergiusens:
This enables support for all architectures and core20 for LXD

-----


snapcraft/internal/build_providers/_lxd/_images.py
```diff
@@ -30,7 +30,7 @@ def get_image_source(*, base: str) -> Dict[str, str]:
     return dict(
         type="image",
         mode="pull",
-        server="https://cloud-images.ubuntu.com/minimal/releases/",
+        server="https://cloud-images.ubuntu.com/buildd/daily",
```
cjp256: would this affect any images today (negatively)?

snapcraft/internal/build_providers/_lxd/_lxd.py
```diff
@@ -365,6 +382,49 @@ def pull_file(self, name: str, destination: str, delete: bool = False) -> None:
...
         self._run(command=["/bin/bash"])
 
+    def _wait_for_network(self) -> None:
+        while True:
+            if self._container is None:
+                raise RuntimeError("Attempted to use container before starting.")
+            if any(
+                a["address"]
+                for a in self._container.state().network["eth0"]["addresses"]
+                if a["family"] == "inet"
```
cjp256: Is ipv6 supported today?  or will lxd tunnel 4to6?

snapcraft/internal/build_providers/_lxd/_lxd.py
```diff
@@ -365,6 +382,49 @@ def pull_file(self, name: str, destination: str, delete: bool = False) -> None:
...
+            ):
+                break
+            sleep(1)
```
cjp256: Maybe make 0.1?  I don't think that'd be unreasonable - especially if it saves me a few seconds each day :stuck_out_tongue_winking_eye:

snapcraft/internal/build_providers/_lxd/_images.py
sergiusens: Yes, they will have missing packages in some cases, but will match closely what is on launchpad. This should be fine considering that we had the "preview" message when using "--use-lxd".

snapcraft/internal/build_providers/_lxd/_lxd.py
sergiusens: sure

snapcraft/internal/build_providers/_lxd/_lxd.py
sergiusens: there's always an ipv6 address assigned, so this is not the best mechanism to have if the network is ipv6 only and we might need to fallback to connectivity checks (but that would make all builds require being online). Not sure what is best at this point with this. So far, for snapcraft we have been recommending lxd defaults, which setup ipv4

sergiusens: @stgraber  or @Saviq do you have suggestions for this?

stgraber: LXD sets up a private IPv4 and IPv6 subnet by default.
As we use a private IPv6 subnet, the C library will prefer IPv4 over it, however should connecting over IPv4 fail, IPv6 will be attempted.

cjp256: Should we wait for both? Or just get started once ipv4 is configured?

snapcraft/internal/build_providers/_lxd/_images.py
cjp256: roger

snapcraft/internal/build_providers/_lxd/_lxd.py
stgraber: The normal recommend is to wait for what you care about. In most cases you don't care about an address, you care about being able to connect to stuff.

In our own CI, our go-to test is to run `getent hosts linuxcontainers.org` in the container. When that succeeds it means we have an IP and DNS working, doesn't matter if over v4 or v6.

sergiusens:
On jue, 2020-03-26 at 10:52:48 -0300, Stéphane Graber wrote:

> @stgraber commented on this pull request.
>
>
>
>> @@ -365,6 +382,49 @@ def pull_file(self, name: str, destination: str, delete: bool = False) -> None:
>      def shell(self) -> None:
>          self._run(command=["/bin/bash"])
>  
> +    def _wait_for_network(self) -> None:
> +        while True:
> +            if self._container is None:
> +                raise RuntimeError("Attempted to use container before starting.")
> +            if any(
> +                a["address"]
> +                for a in self._container.state().network["eth0"]["addresses"]
> +                if a["family"] == "inet"
>
> The normal recommend is to wait for what you care about. In most cases you don't care about an address, you care about being able to connect to stuff.
>
> In our own CI, our go-to test is to run `getent hosts linuxcontainers.org` in the container. When that succeeds it means we have an IP and DNS working, doesn't matter if over v4 or v6.

True, I was trying to think of the repeatable I-am-offline case and thinking
about this a bit more, we really don't need an IP at all at that point.

-- 
---
Sergio Schvezov

snapcraft/internal/build_providers/_lxd/_lxd.py
```diff
@@ -361,6 +378,91 @@ def pull_file(self, name: str, destination: str, delete: bool = False) -> None:
     def shell(self) -> None:
         self._run(command=["/bin/bash"])
 
+    def _wait_for_network(self) -> None:
+        for i in range(5):
```
cjp256, line 382: i think a longer timeout is desirable. this would be 2.5 secs?

PR #2972, opened 2020-03-13
catkin plugins: remove bash workaround for catkin cmake args

cjp256:
Parse the args like bash would using shlex.split().
    
    This parsing should match the behavior currently performed by
    pumping it all out to bash and executing it there.
    
    While I think any approach here is a bit fragile, we demonstrate
    various test cases with some unit tests and add a few cmake arg
    variants to ros-talker-listener:
    - no quotes
    - basic quotes
    - nested quotes using inner-escaped
    
    For the spread test, we don't really need to validate the variables,
    just that it managed to build.

### DIFF
--- a/snapcraft/plugins/catkin.py
+++ b/snapcraft/plugins/catkin.py
@@ -76,6 +76,7 @@ import os
 import tempfile
 import logging
 import re
+import shlex
 import shutil
 import subprocess
 import textwrap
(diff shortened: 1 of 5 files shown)


snapcraft/plugins/catkin.py
```diff
@@ -803,13 +842,9 @@ def _build_catkin_packages(self):
         catkincmd.append("-DCMAKE_BUILD_TYPE={}".format(build_type))
 
         # Finally, add any cmake-args requested from the plugin options
-        catkincmd.extend(self.options.catkin_cmake_args)
+        catkincmd.extend(self._parse_cmake_args())
 
-        # This command must run in bash due to a bug in Catkin that causes it
-        # to explode if there are spaces in the cmake args (which there are).
-        # This has been fixed in Catkin Tools... perhaps we should be using
-        # that instead.
```
kyrofa: It's been a long time since this was written. Any chance you tried just switching this to `self.run` to observe the breakage for yourself? I don't remember what exactly it was.

cjp256: I did, and at first thought it was OK.  but then I came up with a case that breaks:

-DCMAKE_C_FLAGS="-Wall -Werror -DIGNORE_ME_IM_JUST_TESTING_SPACES"

cmake /root/parts/ros-project/build/src/listener -DCATKIN_DEVEL_PREFIX=/root/parts/ros-project/build/devel_isolated/listener -DCMAKE_INSTALL_PREFIX=/root/parts/ros-project/install/opt/ros/melodic -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_FLAGS="-Wall -Werror -DIGNORE_ME_IM_JUST_TESTING_SPACES" -G Unix Makefiles in '/root/parts/ros-project/build/build_isolated/listener'

/usr/bin/cc   "-Wall -Werror -DIGNORE_ME_IM_JUST_TESTING_SPACES"    -o CMakeFiles/cmTC_64abc.dir/testCCompiler.c.o   -c /root/parts/ros-project/build/build_isolated/listener/CMakeFiles/CMakeTmp/testCCompiler.c

kyrofa: Ah excellent, okay. And I see you've used essentially that for the spread test as well, so this seems to work just fine.

PR #2976, opened 2020-03-16
cli: formalize developer debug as hidden build option

cjp256:
It will now be handled in the same manner as other build
provider options.  If it is set, it will automatically get
passed through to build environment if used on host with LXD
or Multipass providers.

-----


sergiusens:
Technically `managed-host` should not be an option, but we can fix that in a different PR

PR #2987, opened 2020-03-20
spread tests: set appropriate default base in snapcraft.yamls

cjp256:
Instead have spread tests replace the base with appropriate
target base.  This way we can easily build tests without having
to modify snapcraft.yaml locally.

Every snap was designated as core18 by default, except:
- ament is legacy, so no base added
- dotnet is core16 only according to spread.yaml
- all legacy tests

Added clear_base() helper for legacy tests outside of /legacy/
to specifically ensure that the base is removed when run.

This also populates grade and confinement in a couple cases
where they weren't specified explicitly, nor set with
set_confinement().

-----

### DIFF
--- a/tests/spread/build-providers/snaps/env-passthrough/snap/snapcraft.yaml
+++ b/tests/spread/build-providers/snaps/env-passthrough/snap/snapcraft.yaml
@@ -5,6 +5,7 @@ description: |
   Help verify environment passthrough variables are set.
 
 grade: devel
+base: core18
 confinement: devmode
 
 parts:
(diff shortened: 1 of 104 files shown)


tests/spread/plugins/copy/snaps/copy-plugin-test/snap/snapcraft.yaml
```diff
@@ -7,6 +7,7 @@ description: |
   Make a new one.
 
 grade: devel
+base: core18
```
sergiusens: copy plugin is only available when not using bases, we can probably remove these and others that are not bases related as we do test them for releases to legacy. That said, this is also a quick smoke test to make sure we do not break re-execution.

cjp256:
I fixed the legacy copy/tar-content plugins, but wonder if we should move them into /tests/spread/legacy?

cjp256:
Rebased to address conflicts.

PR #3000, opened 2020-03-30
repo: use python-apt's fetch_binary implementation

cjp256:
python-apt has been updated to be less verbose, so we can
use the fetch_binary() logic that was lifted from there.

Update mocks, fixture, and tests accordingly.

Depends on https://github.com/snapcore/snapcraft/pull/2999

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,7 @@ PyYAML==5.3
 pyxdg==0.26
 requests==2.20.0
 requests_unixsocket==0.1.5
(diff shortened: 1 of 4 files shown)


spread.yaml
```diff
@@ -241,6 +241,7 @@ suites:
        libxslt-dev \
        libyaml-dev \
        python3-venv \
+       python3-wheel \
```
sergiusens, line 244: do you need to remove this?

PR #3004, opened 2020-03-31
storeapi: add channel-map endpoint

sergiusens:
Only the data required to implement the Progressive Delivery feature are considered.

-----

### DIFF
--- a/snapcraft/storeapi/v2/_api_schema.py
+++ b/snapcraft/storeapi/v2/_api_schema.py
@@ -1,0 +1,221 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2020 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Attributes that the jsonschema from the Snap Store originally require but
+# Snapcraft does not have been commented out from this schema originally
+# imported from
+# https://dashboard.snapcraft.io/docs/v2/en/snaps.html#snap-channel-map
+CHANNEL_MAP_JSONSCHEMA = {
+    "additionalProperties": False,
+    "properties": {
+        "channel-map": {
+            "items": {
+                "additionalProperties": False,
+                "properties": {
+                    "architecture": {"type": "string"},
+                    "channel": {
+                        "description": 'The channel name, including "latest/" for the latest track.',
+                        "type": "string",
+                    },
+                    "expiration-date": {
+                        "description": "The date when this release expires, in RFC 3339 format. If null, the release does not expire.",
+                        "format": "date-time",
+                        "type": ["string", "null"],
+                    },
+                    "progressive": {
+                        "additionalProperties": False,
+                        "properties": {
+                            "key": {"type": ["string", "null"]},
+                            "paused": {"type": ["boolean", "null"]},
+                            "percentage": {"type": ["number", "null"]},
+                        },
+                        "required": ["key", "paused", "percentage"],
+                        "type": "object",
+                    },
+                    "revision": {"type": "integer"},
+                    "when": {
+                        "description": "The date when this release was made, in RFC 3339 format.",
+                        "format": "date-time",
+                        "type": "string",
+                    },
+                },
+                "required": [
+                    "architecture",
+                    "channel",
+                    "expiration-date",
+                    "progressive",
+                    "revision",
+                    # "when"
+                ],
+                "type": "object",
+            },
+            "minItems": 0,
+            "type": "array",
+        },
+        "revisions": {
+            "items": {
+                "additionalProperties": False,
+                "properties": {
+                    "architectures": {
+                        "items": {"type": "string"},
+                        "minItems": 1,
+                        "type": "array",
+                    },
+                    "attributes": {"type": "object"},
+                    "base": {"type": "string"},
+                    "build-url": {"type": "string"},
+                    "confinement": {
+                        "enum": ["strict", "classic", "devmode"],
+                        "type": "string",
+                    },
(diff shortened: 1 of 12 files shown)


snapcraft/storeapi/v2/channel_map.py
```diff
@@ -0,0 +1,266 @@
...
+            percentage=payload["percentage"],
+        )
+
+    def marshal(self) -> Dict[str, Union[str, bool]]:
+        return {"key": self.key, "paused": self.paused, "percentage": self.percentage}
+
+    def __repr__(self) -> str:
+        return f"<Progressive: {self.percentage!r}>"
+
+    def __init__(
```
cjp256: should __init__ be the first instance method?

snapcraft/storeapi/v2/channel_map.py
```diff
@@ -0,0 +1,266 @@
...
+"""
+
+
+class Progressive:
+    """
+    Represents Progressive information for a MappedChannel.
+    """
+
+    @classmethod
+    def unmarshal(cls, payload: Dict[str, Union[str, bool]]) -> "Progressive":
```
cjp256: While entirely correct, I wonder if trying to maintain the `Union`s for (un)mashaling is a bit unnecessary instead of say `Any`.... but your call :-)

snapcraft/storeapi/v2/channel_map.py
```diff
@@ -0,0 +1,266 @@
...
+        self.architectures = architectures
+
+
+class SnapChannel:
+    """
+    Represents a Channel from a channel List for snap.
+    """
+
+    @classmethod
+    def unmarshal(cls, payload: Dict[str, Optional[str]]):
```
cjp256: `.. -> "SnapChannel":`

snapcraft/storeapi/v2/channel_map.py
```diff
@@ -0,0 +1,266 @@
...
+
+    def marshal(self) -> Dict[str, Any]:
+        return {
+            "channel-map": [c.marshal() for c in self.channel_map],
+            "revisions": [r.marshal() for r in self.revisions],
+            "snap": self.snap.marshal(),
+        }
+
+    def __repr__(self) -> str:
+        return "<ChannelMap: {!r}>".format(self.snap.name)
```
cjp256: could use `self.__class__.__name__` in case you ever change the name

tests/unit/store/v2/test_channel_map.py
```diff
@@ -0,0 +1,319 @@
...
+        self.expectThat(mc.marshal(), Equals(self.payload))
+
+    def test_channel_with_expiration(self):
+        date_string = "2020-02-11T17:51:40.891996Z"
+        self.payload.update({"expiration-date": date_string})
+
+        mc = channel_map.MappedChannel.unmarshal(self.payload)
+
+        self.expectThat(
+            repr(mc),
```
cjp256: it's interesting that you are testing reprs, are they being used for anything other than debugging/development?

tests/fixture_setup/_fixtures.py
```diff
@@ -339,6 +344,12 @@ class FakeStoreSearchServerRunning(FakeServerRunning):
 class StagingStore(fixtures.Fixture):
     def setUp(self):
         super().setUp()
+        self.useFixture(
+            fixtures.EnvironmentVariable(
+                "SNAPU_STORE_DASHBOARD_ROOT_URL",
```
sergiusens: oh, this needs fixing. This was in preparation of more than anything (as we only test commands from integration tests).

snapcraft/storeapi/v2/channel_map.py
```diff
@@ -0,0 +1,266 @@
...
+
+class Progressive:
+    """
+    Represents Progressive information for a MappedChannel.
+    """
+
+    @classmethod
+    def unmarshal(cls, payload: Dict[str, Union[str, bool]]) -> "Progressive":
+        return cls(
+            key=payload["key"],
```
cjp256: The payload data does needs to be validated on unmarshal to make sure we received the expected types.

In the meta, I've been doing something like:
```
key = payload.get("key")
if not isinstance(key, str):
  raise SomeError()
...
```

But it could be done more concisely with a schema check...

tests/unit/store/v2/test_channel_map.py
sergiusens: yes, this is for debugging/devel

snapcraft/storeapi/v2/channel_map.py
sergiusens: done, with the schema, but I want it to raise on error to sentry as is if that is ok with you

snapcraft/storeapi/v2/channel_map.py
sergiusens: done

snapcraft/storeapi/v2/channel_map.py
sergiusens: Done

snapcraft/storeapi/v2/channel_map.py
sergiusens: if we need more, we can move to Any, but 2 values or 3 is still manageable

snapcraft/storeapi/v2/channel_map.py
sergiusens: we never discussed this, this is the order I usually see in the wild, I may be wrong or biased. What is your preference?

cjp256: Either way, I just thought of classmethods before `__init__` guideline, and just assumed that implied that __init__ should be top.

sergiusens: We should probably rename it to special attributes. Thoughts?

snapcraft/storeapi/v2/_api_schema.py
```diff
@@ -0,0 +1,221 @@
...
+                        "type": "string",
+                    },
+                },
+                "required": [
+                    "architecture",
+                    "channel",
+                    "expiration-date",
+                    "progressive",
+                    "revision",
+                    # "when"
```
cjp256, line 61: I assume fields were commented out that we don't care about?  I think it's fair to leave them, even if we don't care about them (yet).

snapcraft/storeapi/v2/channel_map.py
cjp256: Googling about, I found some interesting rationale wrt ordering:
https://stackoverflow.com/a/36967648

snapcraft/storeapi/v2/_api_schema.py
sergiusens: yes, exactly so, I added a comment mentioning this at the top.

PR #3008, opened 2020-04-01
cli: use the channel-map api for status

sergiusens:
Also:

- introduce --experimental-progressive-delivery
- fixup the channel-map schema

-----

### DIFF
--- a/snapcraft/cli/_channel_map.py
+++ b/snapcraft/cli/_channel_map.py
@@ -14,6 +14,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import itertools
 import os
 from collections import OrderedDict
 from typing import List, Optional
@@ -197,18 +198,23 @@ def _get_channel_lines_for_channel(
 def _has_channels_for_architecture(
     snap_channel_map, architecture: str, channels: List[str]
 ) -> bool:
-    for channel_name in channels:
-        for progressive in (False, True):
-            try:
-                snap_channel_map.get_mapped_channel(
-                    channel_name=channel_name,
-                    architecture=architecture,
-                    progressive=progressive,
-                )
-                return True
-            except ValueError:
-                continue
-    return False
+    progressive = (False, True)
+    # channel_query = (channel_name, progressive)
+    for channel_query in itertools.product(channels, progressive):
+        try:
+            snap_channel_map.get_mapped_channel(
+                channel_name=channel_query[0],
+                architecture=architecture,
+                progressive=channel_query[1],
+            )
+            found_architecture = True
+            break
+        except ValueError:
+            continue
+    else:
+        found_architecture = False
+
+    return found_architecture
 
 
 def get_tabulated_channel_map(
--- a/snapcraft/cli/_channel_map.py
+++ b/snapcraft/cli/_channel_map.py
@@ -1,0 +1,267 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2020 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+from collections import OrderedDict
+from typing import List, NamedTuple, Optional
+
+from tabulate import tabulate
+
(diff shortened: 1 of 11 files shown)


snapcraft/cli/_channel_map.py
```diff
@@ -0,0 +1,267 @@
...
+
+from snapcraft.storeapi.v2.channel_map import (
+    ChannelMap,
+    MappedChannel,
+    Revision,
+    SnapChannel,
+)
+
+
+_HINTS_T = NamedTuple(
```
cjp256, line 31: I'd suggest using mypy to for constant checks with `Final`.

Either something like `_SYMBOL_CLOSED: Final[str] = "-"`
Or If you prefer, bundle them into a class:
```
from typing import Final
class Hints:
  CLOSED: Final[str] = "-"
  FOLLOWING: Final[str] = "↑"
...
```

snapcraft/cli/store.py
```diff
@@ -360,19 +362,39 @@ def close(snap_name, channels):
...
     """Get the status on the store for <snap-name>.
 
     \b
     Examples:
         snapcraft status my-snap
-        snapcraft status my-snap --arch armhf
     """
-    snapcraft.status(snap_name, arch)
+    if experimental_progressive_delivery:
+        os.environ["SNAPCRAFT_EXPERIMENTAL_PROGRESSIVE_DELIVERY"] = "Y"
```
cjp256, line 383: This is already handled by apply_host_provider_flags(), but I have used strtobool, so they could be made consistent to 0|1 for boolean flags for checking throughout code.

sergiusens: We discussed this and agreed that this only applies to build commands (this is not one) and that one solution would be to implement a click.Option that does this for us.

snapcraft/cli/_channel_map.py
sergiusens: Sure.

snapcraft/cli/_channel_map.py
```diff
@@ -0,0 +1,267 @@
...
+    if revision is not None:
+        version_string = revision.version
+        revision_string = f"{revision.revision}"
+
+    if mapped_channel is not None:
+        if channel_info.branch is None and mapped_channel.progressive.percentage:
+            channel_string = ""
+        elif channel_info.branch is not None:
+            channel_string = f"{channel_info.risk}/{channel_info.branch}"
+            # This value can be None.
```
cjp256, line 102: Can probably drop the comment if check is:
`if mapped_channel.expiration_date is not None`

snapcraft/cli/_channel_map.py
```diff
@@ -0,0 +1,267 @@
...
+) -> bool:
+    for channel_name in channels:
+        for progressive in (False, True):
+            try:
+                snap_channel_map.get_mapped_channel(
+                    channel_name=channel_name,
+                    architecture=architecture,
+                    progressive=progressive,
+                )
+                return True
```
cjp256, line 212: uh oh, here's a case of multiple returns ;)

snapcraft/cli/_channel_map.py
```diff
@@ -0,0 +1,267 @@
...
+            except ValueError:
```
cjp256, line 213: I wonder if get_mapped_channel() should have returned a None.

snapcraft/storeapi/v2/_api_schema.py
```diff
@@ -198,7 +198,9 @@
                                 "type": ["string", "null"],
                             },
                         },
-                        "required": ["name", "creation-date", "pattern"],
+                        # pattern is documented as required but is not returned,
```
cjp256, line 201: So the API is not passing their own schema?  Is the store team aware?

sergiusens: yes, they are fixing it. I suspect (and hope) this is a bug in the documentation and not the schema itself

snapcraft/cli/_channel_map.py
sergiusens: we have been value based more than exception based so far in our duo team endeavor which might strike a cord with familiarity.

Let's discuss this later when we have more time and adapt our coding style to reflect that.

snapcraft/cli/_channel_map.py
sergiusens: done

snapcraft/cli/_channel_map.py
sergiusens: fixed

snapcraft/cli/_channel_map.py
```diff
@@ -0,0 +1,269 @@
...
+    if revision is not None:
+        version_string = revision.version
+        revision_string = f"{revision.revision}"
+
+    if mapped_channel is not None:
+        if channel_info.branch is None and mapped_channel.progressive.percentage:
+            channel_string = ""
+        elif channel_info.branch is not None:
+            channel_string = f"{channel_info.risk}/{channel_info.branch}"
+            # This value can be None.
```
cjp256, line 99: rm this comment?

sergiusens:
the spread failure is related to spread mixing up the new test in master for go.mod into the tests for this snap. Ignoring failure for this reason

PR #3021, opened 2020-04-07
remote-build: remove artifact sanity check

cjp256:
A snapcraft.yaml may have a configuration of `build-on` that allows
Launchpad to pick from a set of architectures that will run on
multiple architectures.

This will fail our sanity check which assumes the arch, which may
in fact end up as `multi`.  There isn't really any need to sanity
check the artifacts anyhow, so we'll remove this check and download
whatever Launchpad provides.

LP: #1870730

-----

### DIFF
--- a/snapcraft/internal/remote_build/_launchpad.py
+++ b/snapcraft/internal/remote_build/_launchpad.py
@@ -159,7 +159,7 @@ class LaunchpadClient:
 
         logger.debug("Downloading artifacts...")
         for build in builds:
-            self._download_snap(build)
+            self._download_build_artifacts(build)
             self._download_log(build)
 
     def _get_builds_collection_entry(self, snap: Entry) -> Optional[Entry]:
@@ -409,7 +409,7 @@ class LaunchpadClient:
         except urllib.error.HTTPError as e:
             logger.error(f"Error downloading {url}: {e.reason}")
 
-    def _download_snap(self, build: Dict[str, Any]) -> None:
+    def _download_build_artifacts(self, build: Dict[str, Any]) -> None:
         arch = build["arch_tag"]
         snap_build = self._lp_load_url(build["self_link"])
         urls = snap_build.getFileUrls()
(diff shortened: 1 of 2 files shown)


snapcraft/internal/remote_build/_launchpad.py
```diff
@@ -421,11 +421,6 @@ def _download_snap(self, build: Dict[str, Any]) -> None:
         for url in urls:
             snap_name = _get_url_basename(url)
```
sergiusens, line 422: side comment, this should probably be snap_file_name

snapcraft/internal/remote_build/_launchpad.py
```diff
@@ -421,11 +421,6 @@ def _download_snap(self, build: Dict[str, Any]) -> None:
...
 
-            # Sanity check file name.
-            if not snap_name.endswith("_{}.snap".format(arch)):
```
sergiusens, line 425: good... if anything you can keep a startswith(snap_name) (snap_name being the snap_name and not the file_name)

snapcraft/internal/remote_build/_launchpad.py
cjp256: `_get_url_basename()` is used to get the last component in the URL. I would agree if this was the only use of this function.  However, `_get_url_basename()` is also used to get the build # from the URL.  With that context, feel free to suggest a better name :+1:

snapcraft/internal/remote_build/_launchpad.py
cjp256: The last time I looked at this I didn't know if there were any other possible artifacts that we should be ignoring.  Though I haven't come across anything yet...  We could:
(1) filter for `*.snap`?
(2) just download it all?  if something else did get shipped, it's probably a good idea to download it, because it's about to get deleted...

snapcraft/internal/remote_build/_launchpad.py
cjp256: scratch my prior stupid comment, I misunderstood :man_facepalming: 
fixing

snapcraft/internal/remote_build/_launchpad.py
cjp256: went for #2 and just updated the log handling

snapcraft/internal/remote_build/_launchpad.py
cjp256: fixed

cjp256:
Forgive my force push, but it's all different now anyways.

PR #3022, opened 2020-04-08
plugins: introduce v2.PluginV2 and v2.NilPlugin

sergiusens:
There's just enough scaffolding in place to allow for the plugin to
load.

PluginHandler has been extended to determine when a PluginV2 is being
used (PluginHandler should be renamed to PartHandler).

Some extra attribute were added to PluginV2 to minimize the changes
required in PluginHandler.

Old code handling state file migration in PluginHandler._do_build was
removed as that was done before the migration to bases (meaning that
no base using Snapcraft project needs this).

-----


cjp256:
some failures: https://travis-ci.org/github/snapcore/snapcraft/jobs/672990847#L7981

sergiusens:
@cjp256 yeah, fixed them earlier, my push got stuck on my ssh keys...

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -548,33 +567,16 @@ def prepare_build(self, force=False):
...
-                        return common.SNAPCRAFT_FILES + snaps
-                    else:
-                        return common.SNAPCRAFT_FILES
-                else:
-                    return []
+        if not (
+            isinstance(self.plugin, plugins.v1.PluginV1)
+            and self.plugin.out_of_source_build
+        ):
+            if os.path.exists(self.part_build_dir):
```
cjp256, line 574: so this will always happen with v2?

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -584,18 +586,75 @@ def update_build(self):
...
+        print("## Plugin Environment", file=run_environment)
+        for env in plugin_build_environment:
+            for k, v in env.items():
+                print(f'export {k}="{v}"', file=run_environment)
+        print("## Part Environment", file=run_environment)
+        for env in part_build_environment:
+            for k, v in env.items():
+                print(f'export {k}="{v}"', file=run_environment)
+
+        run_environment.seek(0)
```
cjp256, line 630: you can use run_environment.getvalue() instead of `seek`/`read`.  i'd suggest using `with` to clean to make sure it is closed...

perhaps:
```
with io.StringIO() as run_environment:
  ..
  return run_environment.getvalue()
```

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -112,24 +118,33 @@ def __init__(
         self._scriptlet_metadata: Dict[
             steps.Step, snapcraft.extractors.ExtractedMetadata
         ] = collections.defaultdict(snapcraft.extractors.ExtractedMetadata)
+
+        if isinstance(plugin, plugins.v2.PluginV2):
+            build_step_function = self._do_v2_build
```
cjp256, line 123: maybe `run_build_step_function`?

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -806,17 +876,22 @@ def prime(self, force=False) -> None:
...
     def _do_prime(self) -> None:
         snap_files, snap_dirs = self.migratable_fileset_for(steps.PRIME)
-        _migrate_files(snap_files, snap_dirs, self.stagedir, self.primedir)
+        _migrate_files(
+            snap_files, snap_dirs, self._project.stage_dir, self._project.prime_dir
+        )
 
-        if self._snap_type == "app" and self._base is not None:
+        if (
+            self._project._snap_meta.type in ("app", None)
```
cjp256, line 892: is this an unrelated fix?  when would type be none?

snapcraft/internal/pluginhandler/__init__.py
sergiusens: I suppose this is what we just discussed not doing and you were going to look into. I can remove the rm entirely and we can see what happens

snapcraft/internal/pluginhandler/__init__.py
sergiusens: good call

snapcraft/internal/pluginhandler/__init__.py
sergiusens: this is from the previous PR

snapcraft/internal/pluginhandler/__init__.py
cjp256: i thought we discussed the opposite.  v2 would leave the build directory alone and incremental builds would work?

sergiusens: exactly, but I did this before that conversation, that's all I am saying :-)

tests/spread/plugins/v2/nil/snaps/nil-basic/snap/snapcraft.yaml
```diff
@@ -11,8 +11,8 @@ parts:
   nil-part:
     plugin: nil
     build-environment:
-            - foo: bar
+      - foo: bar
     override-build: |
-            env
-            snapcraftctl build
-            env
+      env | grep foo
```
cjp256, line 16: MATCH?

snapcraft/internal/pluginhandler/__init__.py
cjp256: I say let's make the change now :)

PR #3030, opened 2020-04-10
repo: drop _AptCache and add migrate to install_stage_packages()

cjp256:
This does remove the broken dependency handling... I need
a working test case to make sure that's solid.

- Move repo.get_pkg_name_parts to test_base, even though it's
  in __init__.

- Remove all instance methods from repo and Ubuntu, consolidating
  to install_stage_packages() classmethod.

- Move non-class methods in Repo into Repo class so they can be
  easily inherited.

- Remove custom _AptCache implementation.

  - Use a single instance of the system's apt cache, which should
    improve performance due to the penalty of recreating caches in
    numerous points.  It also simplifies usage without the requirement
    of copying/sharing instances for what is (now) a system-wide
    resource.

  - Remove old scheme for gathering stage packages, and simply do a scan
    ourselves for the packages and their dependencies.  Then fetch them
    using python-apt's fetch_binary().

- Rework apt tests.

- Rework of the ROS plugins to remove their current usage of Ubuntu
  and replace it with install_stage_packages().

-----

### DIFF
--- a/snapcraft/internal/pluginhandler/__init__.py
+++ b/snapcraft/internal/pluginhandler/__init__.py
@@ -431,28 +431,23 @@ class PluginHandler:
                 self.part_install_dir, clean_target=False, keep_snap=True
             )
 
-    def _fetch_stage_packages(self):
-        stage_packages = self._grammar_processor.get_stage_packages()
-        if stage_packages:
-            logger.debug("Fetching stage-packages {!r}".format(stage_packages))
-            try:
-                self.stage_packages = self._stage_packages_repo.get(stage_packages)
-            except repo.errors.PackageNotFoundError as e:
-                raise errors.StagePackageDownloadError(self.name, e.message)
-
-    def _unpack_stage_packages(self):
+    def _install_stage_packages(self):
         stage_packages = self._grammar_processor.get_stage_packages()
         if stage_packages:
             logger.debug(
-                "Unpacking stage-packages to {!r}".format(self.part_install_dir)
+                "Installing stage-packages to {!r}".format(self.plugin.installdir)
             )
-            self._stage_packages_repo.unpack(self.part_install_dir)
+            try:
+                self.stage_packages = self._stage_packages_repo.install_stage_packages(
+                    package_names=stage_packages, install_dir=self.plugin.installdir
+                )
+            except repo.errors.PackageNotFoundError as e:
+                raise errors.StagePackageDownloadError(self.name, e.message)
 
     def prepare_pull(self, force=False):
         self.makedirs()
-        self._fetch_stage_packages()
+        self._install_stage_packages()
         self._fetch_stage_snaps()
-        self._unpack_stage_packages()
         self._unpack_stage_snaps()
 
     def pull(self, force=False):
@@ -543,7 +538,7 @@ class PluginHandler:
         self.makedirs()
         # Stage packages are fetched and unpacked in the pull step, but we'll
         # unpack again here just in case the build step has been cleaned.
-        self._unpack_stage_packages()
+        self._install_stage_packages()
 
     def build(self, force=False):
         self.makedirs()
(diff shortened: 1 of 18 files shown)


snapcraft/internal/project_loader/_parts_config.py
```diff
@@ -201,9 +201,7 @@ def load_part(self, part_name, plugin_name, part_properties):
         for source in sources:
             repo.Repo.install_source(source)
 
-        # TODO: rename with migration strategy.
-        repo_dir = path.join(self._project.parts_dir, part_name, "ubuntu")
-        stage_packages_repo = repo.Repo(repo_dir)
+        stage_packages_repo = repo.Repo
```
sergiusens, line 204: repo_dir is used in PluginHandler, so you might need to adapt some tests or actual code.

snapcraft/internal/repo/_base.py
```diff
@@ -223,13 +193,14 @@ def normalize(self, unpackdir: str) -> None:
...
-        self._remove_useless_files(unpackdir)
-        self._fix_artifacts(unpackdir)
-        self._fix_xml_tools(unpackdir)
-        self._fix_shebangs(unpackdir)
+        cls._remove_useless_files(unpackdir)
+        cls._fix_artifacts(unpackdir)
+        cls._fix_xml_tools(unpackdir)
+        cls._fix_shebangs(unpackdir)
 
+    @classmethod
```
sergiusens, line 201: maybe
```suggestion
    @staticmethod
```

and drop cls

snapcraft/internal/repo/_base.py
```diff
@@ -269,14 +242,15 @@ def _fix_artifacts(self, unpackdir: str) -> None:
...
-                    self._fix_symlink(path, unpackdir, root)
+                    cls._fix_symlink(path, unpackdir, root)
                 elif os.path.exists(path):
                     _fix_filemode(path)
 
                 if path.endswith(".pc") and not os.path.islink(path):
                     fix_pkg_config(unpackdir, path)
 
-    def _fix_xml_tools(self, unpackdir: str) -> None:
+    @classmethod
```
sergiusens, line 252: static

snapcraft/internal/repo/_base.py
```diff
@@ -293,9 +267,10 @@ def _fix_xml_tools(self, unpackdir: str) -> None:
                 "prefix={}/usr".format(unpackdir),
             )
 
-    def _fix_symlink(self, path: str, unpackdir: str, root: str) -> None:
+    @classmethod
```
sergiusens, line 270: static

snapcraft/internal/repo/_base.py
```diff
@@ -307,7 +282,8 @@ def _fix_symlink(self, path: str, unpackdir: str, root: str) -> None:
         os.remove(path)
         os.symlink(os.path.relpath(target, root), path)
 
-    def _fix_shebangs(self, unpackdir: str) -> None:
+    @classmethod
```
sergiusens, line 285: static

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -517,10 +517,6 @@ def clean_pull(self):
         if self.is_clean(steps.PULL):
             return
 
-        # Remove stage-packages cache (where stage packages are fetched)
```
sergiusens, line 525: Does this mean that if you remove a stage-packages entry from a list that would remove the need to bring set packages in that those packages would still be staged?

cjp256: the cache directory is common for everything now (~/.cache/snapcraft/download).  install_stage_packages() is the one that extracts the list of packages.  So assuming the part was cleaned (that is, it detects the change in stage-packages and rebuilds), that stage package is never extracted.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -543,7 +534,7 @@ def prepare_build(self, force=False):
         self.makedirs()
         # Stage packages are fetched and unpacked in the pull step, but we'll
         # unpack again here just in case the build step has been cleaned.
-        self._unpack_stage_packages()
+        self._install_stage_packages()
```
sergiusens, line 537: This might not be needed anymore for v2... we need to cleanup snapcraft.internal.lifecycle._runner when we have time.

PR #3031, opened 2020-04-11
build_providers: pass through SNAPCRAFT_{BUILD,IMAGE}_INFO to container or VM

jhenstridge:

-----
In order to include a `snap/manifest.yaml` in a snap, you can set the `SNAPCRAFT_BUILD_INFO` environment variable.  However, this doesn't work with build providers that perform a build within a container/VM, since the variable is not passed to the child Snapcraft instance.

This should fix [bug #1871770][1].

[1]: https://bugs.launchpad.net/snapcraft/+bug/1871770

### DIFF
--- a/tests/unit/build_providers/test_base_provider.py
+++ b/tests/unit/build_providers/test_base_provider.py
@@ -339,7 +339,10 @@ class BaseProviderTest(BaseProviderBaseTest):
     def test_passthrough_environment_flags_all(self):
         provider = ProviderImpl(project=self.project, echoer=self.echoer_mock)
         provider.build_provider_flags = dict(
-            http_proxy="http://127.0.0.1:8080", https_proxy="http://127.0.0.1:8080"
+            http_proxy="http://127.0.0.1:8080",
+            https_proxy="http://127.0.0.1:8080",
+            SNAPCRAFT_BUILD_INFO=True,
+            SNAPCRAFT_IMAGE_INFO='{"build_url":"https://example.com"}',
         )
 
         results = provider._get_env_command()
@@ -355,6 +358,8 @@ class BaseProviderTest(BaseProviderBaseTest):
                     "SNAPCRAFT_HAS_TTY=False",
                     "http_proxy=http://127.0.0.1:8080",
                     "https_proxy=http://127.0.0.1:8080",
+                    "SNAPCRAFT_BUILD_INFO=True",
+                    'SNAPCRAFT_IMAGE_INFO={"build_url":"https://example.com"}',
                 ]
             ),
         )
--- a/snapcraft/cli/_options.py
+++ b/snapcraft/cli/_options.py
@@ -14,6 +14,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import distutils.util
 import os
 import sys
 
@@ -46,16 +47,21 @@ class PromptOption(click.Option):
         )
 
 
-class SimpleBoolParamType(click.ParamType):
+class BoolParamType(click.ParamType):
     name = "boolean"
 
     def convert(self, value, param, ctx):
         """Convert option string to value.
 
-        Unlike click's BoolParamType, any non-empty string is treated
-        as True.
+        Unlike click's BoolParamType, use distutils.util.strtobool to
+        convert values.
         """
-        return bool(value)
+        if isinstance(value, bool):
+            return value
+        try:
+            return bool(distutils.util.strtobool(value))
+        except ValueError:
+            self.fail("%r is not a valid boolean" % value, param, ctx)
 
     def __repr__(self):
         return "BOOL"
@@ -142,7 +148,7 @@ _PROVIDER_OPTIONS = [
     dict(
         param_decls="--enable-manifest",
         is_flag=True,
-        type=SimpleBoolParamType(),
+        type=BoolParamType(),
         help="Generate snap manifest.",
         envvar="SNAPCRAFT_BUILD_INFO",
         supported_providers=["host", "lxd", "managed-host", "multipass"],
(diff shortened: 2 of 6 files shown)


tests/unit/build_providers/test_base_provider.py
```diff
@@ -293,6 +293,7 @@ def test_clean_multiple_parts(self):
             ["snapcraft", "clean", "part1", "part2"]
         )
 
+    @patch.dict("os.environ", {}, clear=True)
     def test_passthrough_environment_flags_empty(self):
         provider = ProviderImpl(project=self.project, echoer=self.echoer_mock)
```
sergiusens, line 298:
```suggestion
        self.useFixture(fixture_setup.CleanEnvironment())
        provider = ProviderImpl(project=self.project, echoer=self.echoer_mock)
```

tests/unit/build_providers/test_base_provider.py
```diff
@@ -293,6 +293,7 @@ def test_clean_multiple_parts(self):
             ["snapcraft", "clean", "part1", "part2"]
         )
 
+    @patch.dict("os.environ", {}, clear=True)
```
sergiusens, line 296: mind using our already used fixture instead?

tests/unit/build_providers/test_base_provider.py
```diff
@@ -334,6 +336,60 @@ def test_passthrough_environment_flags_non_string(self):
             ),
         )
 
+    @patch.dict("os.environ", {"SNAPCRAFT_BUILD_INFO": "1"}, clear=True)
+    def test_passthrough_environment_flags_build_info(self):
+        provider = ProviderImpl(project=self.project, echoer=self.echoer_mock)
```
sergiusens, line 341:
```suggestion
        self.useFixture(fixture_setup.CleanEnvironment())
        self.useFixture(fixtures.EnvironmentVariable("SNAPCRAFT_BUILD_INFO, "1"))
        provider = ProviderImpl(project=self.project, echoer=self.echoer_mock)
```

tests/unit/build_providers/test_base_provider.py
```diff
@@ -334,6 +336,60 @@ def test_passthrough_environment_flags_non_string(self):
             ),
         )
 
+    @patch.dict("os.environ", {"SNAPCRAFT_BUILD_INFO": "1"}, clear=True)
```
sergiusens, line 339: mind using what I proposed instead?

snapcraft/internal/build_providers/_base_provider.py
```diff
@@ -385,6 +385,13 @@ def _get_env_command(self) -> Sequence[str]:
             value = str(value)
             env_list.append(f"{key}={value}")
 
+        # Pass through build information environment variables.
+        for key in ["SNAPCRAFT_BUILD_INFO", "SNAPCRAFT_IMAGE_INFO"]:
```
sergiusens, line 389: I think @cjp256 would agree, how about we make these self.build_provider_flags instead? That's in `snapcraft.cli._options._PROVIDER_OPTIONS`

cjp256: Yes, something like adding the following to provider options should be sufficient.
```
    dict(
        param_decls="--generate-manifest",
        is_flag=True,
        help="Generate snap manifest.",
        envvar="SNAPCRAFT_BUILD)INFO",
        supported_providers=["host", "lxd", "managed-host", "multipass"],
        hidden=True,
    ),
    dict(
        param_decls="--manifest-image-information",
        metavar="<json string>",
        help="Set snap manifest image-info",
        envvar="SNAPCRAFT_IMAGE_INFO",
        supported_providers=["host", "lxd", "managed-host", "multipass"],
        hidden=True,
    ),
```

Tests won't be required as it's covered with the generic provider option handling.

jhenstridge: I've updated the PR to make these options.

For compatibility with the existing call sites, I've updated `get_build_provider_flags` to not include `SNAPCRAFT_BUILD_INFO` if the option is turned off.  Without this, we'd set the environment variable to `"False"`, which be considered true by the various `if os.getenv("SNAPCRAFT_BUILD_INFO"):` calls.

jhenstridge:
I've updated the PR as suggested.  The CI seems to be failing for reasons unrelated to the changes in my branch.

snapcraft/internal/build_providers/_base_provider.py
cjp256: Good idea, works for the existing flags. Thanks!

jhenstridge:
Spread tests failed with:

```
2020-04-14 18:49:05 Error preparing google:ubuntu-18.04-64:tests/spread/general/build-aux (apr141843-278343) : 
-----
+ mkdir test-snap
+ cd test-snap
+ snapcraft init
Usage: snapcraft [OPTIONS] COMMAND [ARGS]...
Try 'snapcraft -h' for help.

Error: Invalid value for '--generate-manifest': on is not a valid boolean
```

I think I can fix this with a custom Click ParamType that does the simpler "anything non empty is true" interpretation we're using for the environment variable.

cjp256:
@sergiusens should "on" be supported?  Interesting that it's used in spread.yaml, but I think everyone uses `SNAPCRAFT_BUILD_INFO=1`. I think we should fix it in our spread.yaml.

Unrelated, would `--enable-manifest` make better sense than `--generate-manifest`?

sergiusens:
I have no strong opinion on --enable versus --generate although your proposal has a better ring.

Wrt on, 1, true... we already make use of https://docs.python.org/3.6/distutils/apiref.html#distutils.util.strtobool in out code.

jhenstridge:
I've renamed the flag to `--enable-manifest`, and adjusted the parameter type to use `strtobool`.  I also updated the call sites making use of `SNAPCRAFT_BUILD_INFO` to use `strtobool` too, to keep things in sync.

PR #3032, opened 2020-04-11
plugins: introduce v2.MakePlugin with rebuilding

sergiusens:
Environment is now also properly setup.

-----


snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -441,14 +441,26 @@ def _assemble_runtime_environment(self) -> str:
...
+            assembled_env = assembled_env.replace(self._prime_dir, "$SNAP")
+            env.append(self._install_path_pattern.sub("$SNAP", assembled_env))
+        else:
+            # TODO use something local to the meta package and
+            # only add paths for directory items that actually exist.
+            runtime_env = project_loader.runtime_env(
+                self._prime_dir, self._project_config.project.arch_triplet
+            )
+            for e in runtime_env:
+                env.append(re.sub(self._prime_dir, "$SNAP", e))
```
cjp256, line 459: Feels like runtime_env() should do that conversion.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -605,25 +610,29 @@ def _generate_build_env(self) -> str:
         if isinstance(self.plugin, plugins.v1.PluginV1):
             raise RuntimeError("PluginV1 not supported.")
 
-        # TODO add Snapcraft's environment.
+        # Snapcraft's say.
+        part_build_environment = get_part_build_environment(self)
```
cjp256, line 614: get_part_build_environment() feels misleading. 

Maybe use `get_snapcraft_part_environment()` and distinguish between "user" & "snapcraft"? or "common"?

spread.yaml
```diff
@@ -490,6 +490,10 @@ suites:
    systems:
    - ubuntu-16.04*
    - ubuntu-18.04*
+ tests/spread/plugins/v2/make/:
+   summary: tests of snapcraft's v2 Nil plugin
```
cjp256, line 494: s/Nil/make

tests/spread/plugins/v2/make/run/task.yaml
```diff
@@ -0,0 +1,24 @@
+summary: Build and run a basic make snap
```
cjp256, line 1: it might make more sense to allow the existing tests to be extended by adding the 20.04 target? otherwise we're going to have a lot of duplicate tests...  maybe use a symlink for things that can/should be shared?

sergiusens: yeah, I think I want to move all the snap dirs to a common location at the root of the spread tests.

snapcraft/internal/meta/_snap_packaging.py
sergiusens: This instance of runtime_env is used everywhere in a very different way, I import the logic here until we get rid of that.

snapcraft/internal/pluginhandler/__init__.py
sergiusens: sure, will do that.

snapcraft/internal/meta/_snap_packaging.py
sergiusens: this could probably be done while working on the TODO I added.

tests/spread/plugins/v2/make/run/task.yaml
sergiusens: I took a quick look at merging everything, that would require fine tuning all the individual tasks for things not supported in core20 if they do not live in their own suite.

snapcraft/plugins/v2/make.py
```diff
@@ -46,15 +46,7 @@ def get_schema(cls) -> Dict[str, Any]:
             "$schema": "http://json-schema.org/draft-04/schema#",
             "type": "object",
             "additionalProperties": False,
-            "properties": {
-                "make-parameters": {
```
cjp256: i thought we were going to maintain the schema where practical?

sergiusens: this piece makes no sense though

PR #3035, opened 2020-04-14
repo: fix resolution of virtual build packages

cjp256:
With my previous PR, virtual build packages are no longer found.
Ensure that build and stage packages are resolved the best we can,
and persist any potential versioning.

Log any build and stage packages that have been resolved so the
user can see any assumptions we're making to resolve it.

Add unit and spread test coverage.

-----

### DIFF
--- a/tests/spread/general/virtual-packages/task.yaml
+++ b/tests/spread/general/virtual-packages/task.yaml
@@ -1,0 +1,36 @@
+summary: Build a snap that has virtual stage and build packages
+
+environment:
+  SNAP_DIR: snaps/virtual-packages-test
+
+systems:
+  - ubuntu-16*
+  - ubuntu-18*
+
+restore: |
+  cd "$SNAP_DIR"
+  snapcraft clean
+  rm -f ./*.snap
+
+execute: |
+  test_file="/usr/lib/nodejs/uglify-js/bin/uglifyjs"
+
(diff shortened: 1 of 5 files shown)


cjp256:
Depends on #3036 . whoops :-(

.travis.yml
```diff
@@ -58,7 +58,7 @@ jobs:
       addons:
         snaps:
           - name: snapcraft
-            channel: edge
+            channel: stable
```
sergiusens, line 61: candidate at least :-)

tests/spread/general/virtual-packages/task.yaml
```diff
@@ -0,0 +1,32 @@
+summary: Build a snap that has virtual stage and build packages
```
sergiusens, line 1: this should probably be enabled for 20.04 as well.

.travis.yml
cjp256: The next step will be to build snapcraft from the resulting snap.  Any reason not to build from stable as the starting point? Seems to me that should always work...

tests/spread/general/virtual-packages/task.yaml
cjp256: Prior to enabling 20.04 suite for all of general?

cjp256: Done

PR #3039, opened 2020-04-14
build providers: setup initial apt source configuration

cjp256:
To move this into base provider nicely, some minor refactoring
included:

- _setup_snapcraft() is no longer responsibility of setup_environment,
  instead invoked on all launched instances.

- snapcraft refresh is no longer responsibility of setup_snapcraft(),
  instead invoked on all launched instances.  This will ensure the
  cache is always up-to-date.

- refactor _setup_snapcraft_files -> install_file(), modifying the
  use of NamedTemporaryFile to set the suffix as the file's basename,
  which is handy for testing.

- refactor list of dictionary of files into successive calls of
  install_file().

- add apt sources configuration, using
  SNAPCRAFT_BUILD_ENVIRONMENT_PRIMARY_MIRROR and
  SNAPCRAFT_BUILD_ENVIRONMENT_SECURITY_MIRROR flags

- move _get_code_name_from_build_base() from lxd to base provider
  as it's now required for all providers.

- fixes errorneous mode 06444 set for sources.list.

-----


snapcraft/internal/build_providers/_base_provider.py
```diff
@@ -274,10 +240,13 @@ def launch_instance(self) -> None:
...
-            # We always setup snapcraft after a start to bring it up to speed with
-            # what is on the host
-            self._setup_snapcraft()
+
+        # We always setup snapcraft after a start to bring it up to speed with
+        # what is on the host
+        self._setup_snapcraft()
+
+        # Ensure package cache is ready.
+        self._run(["snapcraft", "refresh"])
```
sergiusens, line 249: it is a requirement to only do this on project setup to avoid slow turnaround on multiple calls to snapcraft

snapcraft/internal/build_providers/_base_provider.py
```diff
@@ -290,34 +259,117 @@ def _ensure_base(self) -> None:
...
+            )
+            self._push_file(source=temp_file.name, destination=remote_file)
+            self._run(["mv", remote_file, path])
+            # This chown is not necessarily needed. but does keep things
+            # consistent.
+            self._run(["chown", "root:root", path])
+            self._run(["chmod", permissions, path])
+
+    def _get_code_name_from_build_base(self):
+        # TODO fix this with generalized mechanism.
```
sergiusens, line 281: this is now a generalized mechanism :laughing:

snapcraft/internal/build_providers/_base_provider.py
cjp256: But if the container is launched several days later, how are you sure that the cache is good?

snapcraft/internal/build_providers/_base_provider.py
```diff
@@ -271,13 +237,16 @@ def launch_instance(self) -> None:
...
+
+            # Configure environment for snapcraft use.
             self._setup_environment()
-        else:
-            # We always setup snapcraft after a start to bring it up to speed with
-            # what is on the host
-            self._setup_snapcraft()
+
+            # Refresh repository caches.
+            self._run(["snapcraft", "refresh"])
```
sergiusens, line 245: btw,  why not just apt-get update? Now I recall that "snapcraft refresh" is to be run outside the provider

sergiusens: you do not have snapcraft installed at this point for this to work. I would just remove this logic here (the else), as apt update was just run and we don't want to be refreshing on every instance start.

cjp256: I wondered that myself.  But looking at it, I suppose it can be.  And we can probably remove refresh altogether now?

PR #3043, opened 2020-04-15
package-repositories: initial schema and meta read/write support

cjp256:


cjp256:
> I might be missing some context, but how is repo architectures going to be used here?
> 
> While we do not need to do this now, we should contemplate integration with the grammar engine

If unspecified, architectures will default to the host.  If specified, it will add/restrict to the set specified.  AFAICT, it won't be harmful to add extra architectures, as it won't be used unless explicitly requested (e.g. <package>:<arch> which should be handled by the current grammar for build/stage packages)

cjp256:
Improved the schema docs. Rebased fixups so it's good to merge as-is.

Renamed reserved sources from 'main' and 'security' to 'default' and 'default-security', so upcoming work will rename those in the build provider.

PR #3044, opened 2020-04-15
build providers: use ubuntu-ports mirrors for non-x86 platforms

cjp256:
Non-x86 mirrors use http://ports.ubuntu.com/ubuntu-ports, not
the default archives.  Detect the host CPU and use the correct
mirror.

### DIFF
--- a/snapcraft/internal/build_providers/_base_provider.py
+++ b/snapcraft/internal/build_providers/_base_provider.py
@@ -18,6 +18,7 @@ import abc
 import base64
 import os
 import pathlib
+import platform
 import logging
 import shlex
 import shutil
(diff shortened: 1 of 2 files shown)


snapcraft/internal/build_providers/_base_provider.py
```diff
@@ -287,6 +288,28 @@ def _get_code_name_from_build_base(self):
             "core20": "focal",
         }[build_base]
 
+    def _get_primary_mirror(self) -> str:
+        primary_mirror = os.getenv("SNAPCRAFT_BUILD_ENVIRONMENT_PRIMARY_MIRROR", None)
+
+        if primary_mirror is None:
+            if platform.machine() in ["i686", "x86_64"]:
```
sergiusens, line 295: platform machine is not the right check, it might be minimal here, but amrhf workloads usually run on amr64 and i386 run on amd64.. you can reliably use Project.deb_arch for this.

sergiusens: A todo for supporting something different than the host should be added too.

cjp256: That's why I checked for i686 / amd64 instead of arm variants?  If running on anything non-x86, you get ubuntu-ports.  Am I missing something?  I can use deb_arch though, but I had reviewed it to ensure i686 and x86_64 were the known x86 options.

PR #3054, opened 2020-04-17
Add note about docker snap

sixcorners:
Could we please get a note somewhere that this isn't the project for the docker snap?


sergiusens:
Thank you for your contribution, but this doesn't seem to add up, the repo and organization have no references to docker.

PR #3068, opened 2020-04-21
cmake v2 plugin: default to -DCMAKE_BUILD_TYPE=Release

sergiusens:
Expose an environment variable to easily override and ignore
completely if part of cmake-parameters.

Additionally, remove the incorrect "minitems" in the schema.

-----


snapcraft/plugins/v2/cmake.py
```diff
@@ -28,6 +28,17 @@
     - cmake-parameters:
       (list of strings)
       parameters to pass to the build using the common cmake semantics.
+
+This plugin also interprets these specific build-environment entries:
+
+    - SNAPCRAFT_CMAKE_INSTALL_PREFIX
```
cjp256, line 34: I feel like the lines are a bit blurred between schema and the variables...

Just wondering - have you considered to instead set the default list of `cmake-parameters: ["-DCMAKE_INSTALL_PREFIX=/", "-DCMAKE_BUILD_TYPE=Release"] if unset?  It provides a path for the user to override it, with the documentation reinforcing their consideration if overriding?  Seems to me to cover the 90% case without complexity.

sergiusens: I have, the problem with that is exactly the overriding, most of the user errors and complaints we had from field is the fact that either things cannot be replaced or things cannot be added. This solution seems to cater for both those counter complaints.

sergiusens: And yes, the lines are a bit blurred, this was my main concern with exposing both environment variables and "options" in the first place.

snapcraft/plugins/v2/cmake.py
```diff
@@ -69,7 +70,12 @@ def _get_cmake_configure_command(self) -> str:
             c.startswith("-DCMAKE_INSTALL_PREFIX=")
             for c in self.options.cmake_parameters
         ):
-            cmd.append('-DCMAKE_INSTALL_PREFIX="${SNAPCRAFT_CMAKE_INSTALL_PREFIX}"')
+            cmd.append("-DCMAKE_INSTALL_PREFIX=/")
+
+        if not any(
+            c.startswith("-DCMAKE_BUILD_TYPE=") for c in self.options.cmake_parameters
+        ):
+            cmd.append("-DCMAKE_BUILD_TYPE=Release")
```
cjp256, line 78: Last Q, do we really need to even specify this at all?  Perhaps just leave it up to the project?

sergiusens: ok, will remove, but this is the point of no return.

PR #3069, opened 2020-04-21
make v2 plugin: make use of make-parameters

sergiusens:

-----

### DIFF
--- a/tests/spread/plugins/v2/snaps/make-hello/hello.c
+++ b/tests/spread/plugins/v2/snaps/make-hello/hello.c
@@ -1,7 +1,13 @@
 #include <stdio.h>
 
+#if HELLO == 1
+static char *say = "hello";
+#else
+static char *say = "goodbye";
+#endif
+
 int main()
 {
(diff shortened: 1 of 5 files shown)


tests/spread/plugins/v2/snaps/make-hello/hello.c
```diff
@@ -1,7 +1,13 @@
 #include <stdio.h>
 
+#if HELLO == 1
```
cjp256, line 3: The hello handling could be simplified a bit:

```
#ifndef HELLO
#define HELLO "hello"
#endif

static const char *hello = HELLO;
```
and
```
    make-parameters:
      - HELLO="goodbye"
```
Then you could reuse it for testing rebuild when you detect property changes :D

cjp256: I realize that's not quite wired for cflags, and my head was in cmake mode.

Ignore that.  But perhaps we should have a test case with quotes...

PR #3080, opened 2020-04-22
repo: fix for multi-arch stage-package scenario

cjp256:
When enabling multiple architectures, python-apt's BaseDependency
may yield target_versions for a package which are for the unintended
architecture when coming across a package with ':all' architecture.

For example: python3-pip has a dependency on "python3:any". On AMD64
it may point to the i386 package, when having i386 architectures
enabled:

```
>>> apt.Cache()["python3-pip"].candidate.dependencies[-1].target_versions

[
  <Version: package:'python3:i386' version:'3.6.7-1~18.04'>,
  <Version: package:'python3' version:'3.6.7-1~18.04'>,
  <Version: package:'python3:i386' version:'3.6.5-3'>,
  <Version: package:'python3' version:'3.6.5-3'>]
]
```

Instead, we'll look at python-apt's `or_dependencies` list, which is
used to calculate the allowable target_versions:

```
>>> apt.Cache()["python3-pip"].candidate.dependencies[-1].or_dependencies

<Dependency: [<BaseDependency: name:'python3:any' relation:'>=' version:'3.4~' rawtype:'Depends'>]>
```

apt.package.Dependency will provide us with the 'name', e.g.
'python3:any', and we can use this name to fetch the correct
dependencies.

To support installing the correct architecture, we now will pass
target_arch to _mark_package_dependencies() and
_get_resolved_package().  _get_resolved_package() will use the
now-introduced _get_package_for_arch() method to find the correct
package for the intended architecture:

1. First trying `<package-name>:<target-arch>`.

2. Then trying `<package-name>:all` in case it's a no-arch package.

3. Finally trying `<package-name>`, but logging a warning that
   it may be the incorrect architecture.  This shouldn't happen,
   but the warning is there just so we'll know about it if there
   is a case we didn't consider.

Update unit tests, cleaning out some old cruft left behind and
passing without purpose.

-----

### DIFF
--- a/snapcraft/internal/repo/_deb.py
+++ b/snapcraft/internal/repo/_deb.py
@@ -380,21 +380,72 @@ class Ubuntu(BaseRepo):
         cls._refresh_cache()
 
     @classmethod
-    def _get_resolved_package(cls, package_name: str) -> apt.package.Package:
-        # Strip ":any" that may come from dependency names.
-        if package_name.endswith(":any"):
-            package_name = package_name[:-4]
+    def _package_name_parts(
+        cls, package_name: str
+    ) -> Tuple[str, Optional[str], Optional[str]]:
+        split_version = package_name.split("=")
+        if len(split_version) > 1:
+            version: Optional[str] = split_version[1]
+        else:
+            version = None
+
+        # Purge possible ":any" architecture tag, then split for arch.
+        any_filtered = split_version[0].replace(":any", "")
+        split_arch = any_filtered.split(":")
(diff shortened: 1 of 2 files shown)


sergiusens:
unit test errors

AttributeError: 'FakeAptPackage' object has no attribute 'architecture'

Saviq:
There's a bug incoming, but: https://paste.ubuntu.com/p/5vZWdC692P/

Apparently `dpkg-architecture` isn't available by default.

PR #3086, opened 2020-04-24
repo: restore marked-install strategy for apt-cache

cjp256:
Re-introduce refactored AptCache to wrap apt-cache usage.
Use it in read-only mode for working with the host cache
for use with build-packages, and use it to create a new
transient cache for use with handling stage packages. It
will symlink the host /etc/apt configuration so it matches
the host configuration, but allow for an empty cache to mark
packages and dependencies for fetching.

The strategy of marking stage packages ourselves was more
complicated than it appeared, and this achieves a far better
result with less risk of affecting existing snaps.

Additionally this fixes primed-stage-packages, which mistakenly
was using a `:` instead of `=` to split the name and version.

TODO: unit tests and a spread test for the manifest

-----

### DIFF
--- a/snapcraft/internal/repo/_deb.py
+++ b/snapcraft/internal/repo/_deb.py
@@ -29,19 +29,25 @@ from pathlib import Path
 from typing import Dict, List, Optional, Set, Tuple  # noqa: F401
 from typing_extensions import Final
 
-import apt
 from launchpadlib.launchpad import Launchpad
 from xdg import BaseDirectory
 
 from snapcraft import file_utils
 from snapcraft.project._project_options import ProjectOptions
-from snapcraft.internal import os_release, repo
+from snapcraft.internal import os_release
 from snapcraft.internal.indicators import is_dumb_terminal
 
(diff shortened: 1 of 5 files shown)


snapcraft/internal/repo/_deb.py
```diff
@@ -381,247 +349,59 @@ def _install_packages(cls, package_names: List[str]) -> None:
...
-                skipped_blacklisted=skipped_blacklisted,
-                skipped_essential=skipped_essential,
-                unfiltered_packages=package_names,
-                target_arch=arch,
-            )
+        with tempfile.TemporaryDirectory() as cache_td:
+            with AptCache(cache_dir_path=Path(cache_td), read_only=False) as apt_cache:
+                apt_cache.mark_packages(set(package_names))
+                apt_cache.unmark_packages(
+                    set(package_names), set(_DEFAULT_FILTERED_STAGE_PACKAGES)
```
sergiusens, line 362: oh, this was something we needed for core20 too

PR #3092, opened 2020-04-26
repo: filter stage-packages using base's manifest (core20)

cjp256:
Requires #3086


sergiusens:
The project with the test is that MATCH is hiding this error:
Error: [Errno 13] Permission denied: '/snapcraft/tests/spread/plugins/v2/snaps/python-hello-staged-python/parts/python-part/install/bin/python3'

PR #3097, opened 2020-04-28
colcon v2 plugin + ros2 extension

cjp256:


snapcraft/plugins/v2/colcon.py
```diff
@@ -0,0 +1,207 @@
...
+        ]
+
+
+@click.group()
+def plugin_cli():
+    pass
+
+
+@plugin_cli.command()
+@click.option("--build-base", envvar="SNAPCRAFT_BUILD_BASE", required=True)
```
sergiusens, line 171: Considering this plugin is strictly for core20, one could _hardcode_ the base here.

cjp256: funny i didn't consider that, lol.  it could also be exported by the extension I suppose, but eventually other bases will come along anyways.

extensions/ros2/launch
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+function source_with_prefix() {
+    export COLCON_CURRENT_PREFIX="$1"
+    if [ -f "$COLCON_CURRENT_PREFIX/local_setup.bash" ]; then
+      # shellcheck disable=SC1090
+      source "$COLCON_CURRENT_PREFIX/local_setup.bash"
+    else
+      echo "warning: no setup found for $COLCON_CURRENT_PREFIX"
```
kyrofa, line 9: This would almost certainly be an error, unless this is needed at pull time before it's in place?

snapcraft/internal/project_loader/_extensions/ros2_dashing.py
```diff
@@ -0,0 +1,73 @@
...
+                    "key-server": "keyserver.ubuntu.com",
+                    "suites": ["$SNAPCRAFT_APT_RELEASE"],
+                }
+            ]
+        }
+
+        self.app_snippet = {
+            "command-chain": ["snap/command-chain/ros2-launch"],
+            "environment": {
+                "ROS_DISTRO": "dashing",
```
kyrofa, line 58: This should be set by sourcing that setup.sh above, no?

extensions/ros2/launch
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+function source_with_prefix() {
+    export COLCON_CURRENT_PREFIX="$1"
+    if [ -f "$COLCON_CURRENT_PREFIX/local_setup.bash" ]; then
```
kyrofa, line 5: Is there a reason you're using `local_setup.bash` as opposed to `setup.sh` like the colcon plugin?

snapcraft/internal/project_loader/_extensions/ros2_dashing.py
```diff
@@ -0,0 +1,73 @@
...
+                "PYTHONPATH": "$SNAP/opt/ros/dashing/lib/python3.6/site-packages:$SNAP/usr/lib/python3/dist-packages:$PYTHONPATH",
+            },
+        }
+
+        self.part_snippet = {"build-environment": [{"ROS_DISTRO": "dashing"}]}
+
+        self.parts = {
+            "ros2-dashing-colcon": {
```
kyrofa, line 66: This doesn't actually involve colcon, maybe `ros2-dashing-launcher` is a better name. Is there an established pattern for extension parts names?

extensions/ros2/launch
cjp256: This is strictly for run-time. I didn't error on it because the v1 script ignores it if it doesn't exist, and I thought there may be case where it didn't exist.  So I just warned.  I'll instead error :+1:

snapcraft/internal/project_loader/_extensions/ros2_dashing.py
cjp256: It is used to identify the source script to source.

extensions/ros2/launch
cjp256: This script should be the equivalent of setup.sh as I understand it?

snapcraft/internal/project_loader/_extensions/ros2_dashing.py
cjp256: Ah that's a good idea, will do.

extensions/ros2/launch
kyrofa: Yeah make it error. The other one had to deal with it not being there because `env()` was evaluated for pulling, before it was actually present. This is much nicer now.

cjp256: done

snapcraft/plugins/v2/colcon.py
kyrofa: Is it? Dashing is for core18, will that cause issues?

snapcraft/plugins/v2/colcon.py
```diff
@@ -0,0 +1,230 @@
...
+                "stage-runtime-dependencies",
+            ]
+        )
+
+    def get_build_commands(self) -> List[str]:
+        return [
+            ". /opt/ros/$ROS_DISTRO/setup.sh",
+            "if [ ! -f /etc/ros/rosdep/sources.list.d/20-default.list ]; then sudo rosdep init; fi",
+            "rosdep update --rosdistro $ROS_DISTRO",
+            "rosdep install --from-paths . -y -r",
```
kyrofa, line 182: I question the inclusion of `-r`, to continue installing despite errors. What is the reasoning here?

I also suggest including `-i` so we don't install debs that are actually included in the code being built here.

snapcraft/plugins/v2/colcon.py
```diff
@@ -0,0 +1,230 @@
...
+                os.path.abspath(__file__),
+                "stage-runtime-dependencies",
+            ]
+        )
+
+    def get_build_commands(self) -> List[str]:
+        return [
+            ". /opt/ros/$ROS_DISTRO/setup.sh",
+            "if [ ! -f /etc/ros/rosdep/sources.list.d/20-default.list ]; then sudo rosdep init; fi",
+            "rosdep update --rosdistro $ROS_DISTRO",
```
kyrofa, line 181:
```suggestion
            "rosdep update --include-eol-distros --rosdistro $ROS_DISTRO",
```

snapcraft/plugins/v2/colcon.py
```diff
@@ -0,0 +1,230 @@
...
+@click.option("--part-install", envvar="SNAPCRAFT_PART_INSTALL", required=True)
+@click.option("--ros-distro", envvar="ROS_DISTRO", required=True)
+def stage_runtime_dependencies(build_base: str, part_install: str, ros_distro: str):
+    # TODO: replace with logic reading manifests for exec_depend and perhaps
+    # honors the ignore/select package filters.
+    # TODO: Do we have to stage python packages, or are they all in the repo?
+    apt_packages: Set[str] = set()
+
+    proc = subprocess.run(
+        ["rosdep", "keys", "--from-paths", "."],
```
kyrofa, line 204: You seem to be [missing a few flags](https://github.com/snapcore/snapcraft/blob/master/snapcraft/plugins/v1/_ros/rosdep.py#L155-L162).

snapcraft/plugins/v2/colcon.py
```diff
@@ -0,0 +1,230 @@
...
+        check=True,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
+    )
+
+    for dep in filter(None, proc.stdout.decode().strip().split("\n")):
+        proc = subprocess.run(
+            ["rosdep", "resolve", dep, "--rosdistro", ros_distro],
```
kyrofa, line 212: Again assuming this is running in an environment with a sourced setup.sh, you don't need the `--rosdistro` arg.

snapcraft/plugins/v2/colcon.py
```diff
@@ -0,0 +1,230 @@
...
+@click.group()
+def plugin_cli():
+    pass
+
+
+@plugin_cli.command()
+@click.option("--build-base", envvar="SNAPCRAFT_BUILD_BASE", required=True)
+@click.option("--part-install", envvar="SNAPCRAFT_PART_INSTALL", required=True)
+@click.option("--ros-distro", envvar="ROS_DISTRO", required=True)
+def stage_runtime_dependencies(build_base: str, part_install: str, ros_distro: str):
```
kyrofa, line 197: This is where we should be using [catkin_pkg](https://github.com/ros-infrastructure/catkin_pkg) instead of rosdep.

snapcraft/plugins/v2/colcon.py
cjp256: this is actually foxy, I just did my prototyping with dashing because foxy it was not yet ready...

snapcraft/plugins/v2/colcon.py
cjp256: Ah, I think I thought it was recursive, removing...

snapcraft/plugins/v2/colcon.py
cjp256: I think when I updated it grabbed all distros unless explicitly defined here?

cjp256: Confirmed, will fetch for all distros unless you specify --rosdistro, even with $ROS_DISTRO

snapcraft/plugins/v2/colcon.py
cjp256: Hrm, I may have had issues with -a, I'll bring them back.

snapcraft/internal/project_loader/_extensions/ros2_dashing.py
kyrofa: Ah, fair enough.

snapcraft/plugins/v2/colcon.py
kyrofa: Ha! I didn't actually know that, nice find. I do still suggest including `--include-eol-distros`, though.

extensions/ros2/launch
cjp256: Hrm, we aren't we the ones chaining the overlay and underlay?  In this case, this launch script is supposed to effectively be that `setup.sh`...  I had reviewed the one generated by colcon:

```
# generated from colcon_bash/shell/template/prefix_chain.bash.em

# This script extends the environment with the environment of other prefix
# paths which were sourced when this file was generated as well as all packages
# contained in this prefix path.

# function to source another script with conditional trace output
# first argument: the path of the script
_colcon_prefix_chain_bash_source_script() {
  if [ -f "$1" ]; then
    if [ -n "$COLCON_TRACE" ]; then
      echo ". \"$1\""
    fi
    . "$1"
  else
    echo "not found: \"$1\"" 1>&2
  fi
}

# source chained prefixes
# setting COLCON_CURRENT_PREFIX avoids determining the prefix in the sourced script
COLCON_CURRENT_PREFIX="/opt/ros/foxy"
_colcon_prefix_chain_bash_source_script "$COLCON_CURRENT_PREFIX/local_setup.bash"

# source this prefix
# setting COLCON_CURRENT_PREFIX avoids determining the prefix in the sourced script
COLCON_CURRENT_PREFIX="$(builtin cd "`dirname "${BASH_SOURCE[0]}"`" > /dev/null && pwd)"
_colcon_prefix_chain_bash_source_script "$COLCON_CURRENT_PREFIX/local_setup.bash"

unset COLCON_CURRENT_PREFIX
unset _colcon_prefix_chain_bash_source_script
```

We could use that directly, but then would need patch it to handle `$SNAP`, which I thought was more painful than just handling it in the launcher?  If I'm missing something, we can certainly use the setup.sh, but I didn't see a good way to prescribe the $SNAP without some ugly sed foo.

snapcraft/internal/project_loader/_extensions/ros2_dashing.py
cjp256: Done.

snapcraft/plugins/v2/colcon.py
cjp256: That's right, can't use `-a` with `--from-paths`.  But they seem to be otherwise equal in this case?  Added `-i` back.

snapcraft/plugins/v2/colcon.py
cjp256: Done

snapcraft/plugins/v2/colcon.py
cjp256: I gave it an attempt, check it out!

snapcraft/plugins/v2/colcon.py
cjp256: Restore `-i`.  `-a` didn't always reliably work for me, whereas `--from-paths` did.  I'm not sure why the difference in behavior, but otherwise expect them to be equal?

It may have something to do with no definition of ROS_PACKAGE_PATH?

extensions/ros2/launch
kyrofa: Okay, let's just keep an eye on this then. Especially if we add the ability to find an underlay in the staging area like the catkin plugin does.

snapcraft/plugins/v2/colcon.py
kyrofa: Fair enough, I think you're right about that.

snapcraft/plugins/v2/colcon.py
kyrofa: Yeah that should work, nice job.

tests/spread/plugins/v2/snaps/colcon-ros2-foxy-rlcpp-hello/src/hello.cpp
```diff
@@ -0,0 +1,28 @@
...
+#include <memory>
+#include "rclcpp/rclcpp.hpp"
+
+int main(int argc, char * argv[])
+{
+  rclcpp::init(argc, argv);
+  rclcpp::executors::SingleThreadedExecutor exec;
+  rclcpp::NodeOptions options;
+
+  printf("hello world");
```
kyrofa, line 24: This barely uses ROS. The talker/listener example was far more realistic, in my opinion.

tests/spread/plugins/v2/snaps/colcon-ros2-foxy-rlcpp-hello/snap/snapcraft.yaml
```diff
@@ -0,0 +1,22 @@
...
+  colcon-ros2-foxy-rlcpp-hello:
+    command: opt/ros/foxy/bin/ros2 run colcon_ros2_rlcpp_hello colcon_ros2_rlcpp_hello
+    plugs: [network, network-bind]
+    extensions: [ros2-foxy]
+
+parts:
+  colcon-part:
+    plugin: colcon
+    source: .
+    build-packages: []
```
kyrofa, line 21: You probably want g++ and make here.

snapcraft/plugins/v2/colcon.py
```diff
@@ -0,0 +1,233 @@
...
+@click.group()
+def plugin_cli():
+    pass
+
+
+@plugin_cli.command()
+@click.option("--part-install", envvar="SNAPCRAFT_PART_INSTALL", required=True)
+@click.option("--ros-distro", envvar="ROS_DISTRO", required=True)
+def stage_runtime_dependencies(part_install: str, ros_distro: str):
+    # XXX: Do we have to stage python packages, or are they all in the repo?
```
kyrofa, line 200: Not sure what you're asking here, is this a question that I can help with?

snapcraft/plugins/v2/colcon.py
kyrofa: Yeah man, that looks nice and seems to work very well!

snapcraft/plugins/v2/colcon.py
```diff
@@ -0,0 +1,233 @@
...
+
+            parsed = _parse_rosdep_resolve_dependencies(
+                dep, proc.stdout.decode().strip()
+            )
+            apt_packages |= parsed.pop("apt", set())
+
+            if parsed:
+                click.echo(f"unhandled dependencies: {parsed!r}")
+
+    if apt_packages:
```
kyrofa, line 226: Do you plan on support pip dependencies at all?

tests/spread/plugins/v2/snaps/colcon-ros2-foxy-rlcpp-hello/src/hello.cpp
cjp256: I sort of figured if the rlcpp init succeeded, it'd be adequate.  I don't think we're really stressing ROS here so much as the plugin/extension worked fine, and I thought this was sufficient to cause verify rlcpp linkage.

The only requirement we really have is it has to say "hello world" and exit without fussing about, so it's consistent with the remaining v2 plugin tests which cover other parts of the lifecycle (clean, rebuild, etc.). In theory I was supposed to be able to write a node loop that did exactly that then exited, but it appeared buggy or my interpretation of the docs was wrong.

That said, I agree with your point, if we want to do more stressing of ROS, I think that's great and we should certainly add more tests outside of this task (or figure out how to make it work).

snapcraft/plugins/v2/colcon.py
cjp256: I was wondering if all of the ROS packages that are fetched via rosdep are (also) shipped as debs, and if pip support was really required..?

tests/spread/plugins/v2/snaps/colcon-ros2-foxy-rlcpp-hello/snap/snapcraft.yaml
cjp256: Good point, will do. Interesting how that was never a problem for me, must be all that destructive mode :P

snapcraft/plugins/v2/colcon.py
kyrofa: Sadly no, some deps are ONLY pip, see [LP: #1683036](https://bugs.launchpad.net/snapcraft/+bug/1683036).

tests/spread/plugins/v2/snaps/colcon-ros2-foxy-rlcpp-hello/src/hello.cpp
kyrofa: Are you saying that the existing talker/listener demo that exits after it receives something wasn't working properly? Yeah happy to help with that. We also want to add more spread tests for all the supported ROS distros, so it seems fine to leave this for now and tackle "better ROS tests" as a general thing.

snapcraft/plugins/v2/colcon.py
cjp256: Roger that.

snapcraft/plugins/v2/colcon.py
cjp256: I guess we'll have to :)

I don't think it'll be a problem, it just needs some stuff wired up to behave like python plugin that I didn't want to do unless necessary, but it appears to be so.

snapcraft/plugins/v2/colcon.py
kyrofa: Note that I'm not opposed to that being a follow-up, personally. This is a great step without it.

snapcraft/plugins/v2/colcon.py
kyrofa: Sorry, didn't intend to start two threads about this. I'll resolve this one and we can continue [here](https://github.com/snapcore/snapcraft/pull/3097#discussion_r442541537).

snapcraft/plugins/v2/colcon.py
cjp256: Agreed to revisit in future.

PR #3099, opened 2020-04-29
repo: always refresh cache when build packages are required

cjp256:
Make sure cache is up to date by refreshing whenever a package
being requested is not already installed.

### DIFF
--- a/snapcraft/internal/repo/_deb.py
+++ b/snapcraft/internal/repo/_deb.py
@@ -302,10 +302,15 @@ class Ubuntu(BaseRepo):
 
     @classmethod
     def _check_if_all_packages_installed(cls, package_names: List[str]) -> bool:
-        # If any requested package is not installed, return False. Will check
-        # versions if using <pkg_name>=<pkg_version> syntax parsed by
-        # get_pkg_name_parts().  Used as an optimization to skip installation
-        # and cache refresh if dependencies are already satisfied.
+        """Check if all given packages are installed.
+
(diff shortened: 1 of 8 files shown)


snapcraft/internal/repo/_deb.py
```diff
@@ -300,6 +300,34 @@ def refresh_build_packages(cls) -> None:
                 "failed to run apt update"
             ) from call_error
 
+    @classmethod
+    def _check_if_all_packages_installed(cls, package_names: List[str]) -> bool:
+        # If any requested package is not installed, return False. Will check
```
sergiusens, line 305: With minor edits, this could be a docstring for the method.

cjp256: Certainly could be, and I'll change it!  Since it's private I didn't think it necessary.

FWIW, pep8 says `Write docstrings for all public modules, functions, classes, and methods. Docstrings are not necessary for non-public methods, but you should have a comment that describes what the method does. This comment should appear after the def line.`

Would you prefer to docstring all methods, at least lightly?

sergiusens: documenting docstrings instead of code comments is nice when using the method elsewhere and the IDE supports pop-over help

sergiusens: I don't mind it staying private, as Repo is exported to plugin users and used in v1

cjp256: Understood, thanks!

cjp256: Done BTW

PR #3101, opened 2020-05-01
package repositories: avoid setting up on host when not the target

sergiusens:
The keys were loaded and setup during the project_loader's init.
This is an issue as this is loaded before launching the build
environment which causes host interactions before launching a build
environment.

Defer the key setup and build_packages/build_snaps collection until
needed and call from the lifecycle.

Tests were updated to be segregated to each concern. The Manifest
generation keys which started to fail due to changes required were
completely removed from the lifecycle tests as they are not of concern
to the scope of lifecycle (and are covered by integration tests which
these behaved like).

-----


cjp256:
Looks pretty good, some unit test failures

snapcraft/internal/project_loader/_config.py
```diff
@@ -280,6 +245,53 @@ def _ensure_no_duplicate_app_aliases(self):
...
+        build_snaps = set()
+
+        # Add the base.
+        if self.project._snap_meta.base is not None:
+            build_snaps.add(self.project._snap_meta.base)
+
+        for part in self.all_parts:
+            build_snaps |= part._grammar_processor.get_build_snaps()
+            if not isinstance(part.plugin, plugins.v1.PluginV1):
+                self.build_snaps |= part.plugin.get_build_snaps()
```
sergiusens, line 291:
```suggestion
                build_snaps |= part.plugin.get_build_snaps()
```

PR #3102, opened 2020-05-01
build providers: prevent snap refreshing in build environment

cjp256:
hold refresh for 1 day for installed snaps


sergiusens:
I just had the thought that doing this style of injection would lead to an incomplete manifest.yaml, so that would need to be taken into account. Maybe "snap switch" (or equivalent api call) is better suited for this.

cjp256:
Yes indeed that will.  I updated this to simply hold refresh for one day.

cjp256:
The use of `rc` in the versioning will break this string comparison.

I don't know, maybe we should stick to the python versioning scheme.

PR #3103, opened 2020-05-01
python v2 plugin: set path to the interpreter

sergiusens:
Add the path to the interpreter instead of activating. From
Snapcraft's point of view this is essentially the only thing done by
activation.

LP: #1876162

-----

### DIFF
--- a/snapcraft/plugins/v2/python.py
+++ b/snapcraft/plugins/v2/python.py
@@ -106,14 +106,17 @@ class PythonPlugin(PluginV2):
 
     def get_build_environment(self) -> Dict[str, str]:
         return {
+            # Add PATH to the python interpreter we always intend to use with
+            # this plugin. It can be user overridden, but that is an explicit
+            # choice made by a user.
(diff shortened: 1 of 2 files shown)


snapcraft/plugins/v2/python.py
```diff
@@ -106,14 +106,17 @@ def get_build_packages(self) -> Set[str]:
 
     def get_build_environment(self) -> Dict[str, str]:
         return {
+            # Add PATH to the python interpreter we always intend to use with
+            # this plugin. It can be user overridden, but that is an explicit
+            # choice made by a user.
+            "PATH": "${SNAPCRAFT_PART_INSTALL}/bin:${PATH}",
```
cjp256, line 112: I'd suggest changing SNAPCRAFT_PYTHON_INTERPRETER rather than modifying the path special for this plugin. Changing PATH will affect other things...

sergiusens: The side effect of setting PATH is that we no longer need the venv, but the real reason is for override-build to work with what would pip install in $SNAPCRAFT_PART_INSTALL.

PR #3104, opened 2020-05-04
packaging: use git-based versioning for python package

cjp256:
The current versioning is gathered from debian/changelog which
is no longer maintained.  Use git-describe, with an additional
check to label the package -dirty if the tree is not porcelain.

-----


cjp256:
Done.  Perhaps we should probably add a dirty check for both, but we can do that some other time.

PR #3109, opened 2020-05-07
pcache: introduce persistent cache decorator

cjp256:
Cache results of wrapped function, returning the last known-good result,
if available.  Configurable expiration times may be set, with a default
of one-day.

The recorded items have an identity key derived from:
- function name
- function arguments
- function keyword arguments

All arguments to the wrapped function must be representable via repr().
The arguments must be represented in an uniquely identifiable and
repeatable manner to ensure accuracy and prevent collisions.

Note that this current implementation is not optimized for performance.
The database is opened and closed on-demand which carries a cost.  It could
be updated to keep the database open across the lifetime of the process,
but that may complicate usage with multiple processes (e.g. snapcraftctl).
For our current use cases, this is the safe and appropriate default.

-----

### DIFF
--- a/snapcraft/internal/pcache.py
+++ b/snapcraft/internal/pcache.py
@@ -1,0 +1,132 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2020 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


cjp256:
Some open questions:

1. Do we want to use a decorator? Initially it seemed nice to match the lru_cache style, but practically speaking, there may be some knobs we want to include to forcible bypass the cache under scenarios not handled by expiry - this would currently require unlinking the db.

2. Along with the rest of the stage cache, we probably need a way for the user to wipe it in case something goes really wrong.  Maybe `snapcraft clean --purge-stage-cache`?  Or always purge on clean, but provide a `--ignore-stage-cache` option to allow it to persist?

sergiusens:
Can you please write up a spec for this?

PR #3110, opened 2020-05-07
repo: use persistent cache for fetching stage packages

cjp256:
It several seconds for the apt cache to to be updated and several
more to re-verify the hashes of archives that were already fetched.
It may even take minutes depending on performance and how many
stage-packages are being requested.

Set results to expire in 24 hours as the apt packages should be
refreshed periodically, particularly staged packages.

Depends on #3109


cjp256:
Open questions:

1. Along with the rest of the stage cache, we probably need a way for the user to wipe it in case something goes really wrong.  Maybe `snapcraft clean --purge-stage-cache`?  Or always purge on clean, but provide a `--preserve-stage-cache` option to allow it to persist?

2. Should the location of this get tied to the project directory?

PR #3111, opened 2020-05-08
elf: fix parsing of notes after patchelf mangling

cjp256:
We cannot rely on iterating on the notes via the note segment.
Patchelf may move note sections into other segments, and worse
yet, possibly leave the note segment standing with invalid data.
Here we instead iterate over each section and (more) reliably
gather the note information.

There is some improvement that should be done on the patchelf
side, as `readelf` complains on these binaries, and other issues
have surfaced in the past (e.g. mesa in classic snaps).  However,
this will hopefully address snapcraft build-time issues with not
only our patched files for classic snaps, but externally patched
binaries that are being packaged.

Add a couple of developer logs to indicate when we are parsing
an ELF, so future investigations are easier when something explodes.

-----

### DIFF
--- a/snapcraft/internal/elf.py
+++ b/snapcraft/internal/elf.py
@@ -338,16 +338,14 @@ class ElfFile:
                 elif isinstance(segment, elftools.elf.segments.InterpSegment):
                     self.interp = segment.get_interp_name()
 
-            # We cannot rely on iterating on the notes via the note segment.
-            # Patchelf may move note sections into other segments, and worse
-            # yet, possibly leave the note segment standing with invalid data.
-            # Here we instead iterate over each section and (more) reliably
-            # gather the note information.
-            for section in elf.iter_sections():
-                if isinstance(section, elftools.elf.sections.NoteSection):
-                    for note in section.iter_notes():
-                        if note.n_name == "GNU" and note.n_type == "NT_GNU_BUILD_ID":
-                            self.build_id = _ensure_str(note.n_desc)
+            build_id_section = elf.get_section_by_name(".note.gnu.build-id")
+            if (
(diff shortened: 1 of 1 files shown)


cjp256:
Fixes SNAPCRAFT-1K4

snapcraft/internal/elf.py
```diff
@@ -335,8 +337,15 @@ def _extract_attributes(self) -> None:  # noqa: C901
...
-                elif isinstance(segment, elftools.elf.segments.NoteSegment):
-                    for note in segment.iter_notes():
+
+            # We cannot rely on iterating on the notes via the note segment.
+            # Patchelf may move note sections into other segments, and worse
+            # yet, possibly leave the note segment standing with invalid data.
+            # Here we instead iterate over each section and (more) reliably
+            # gather the note information.
+            for section in elf.iter_sections():
+                if isinstance(section, elftools.elf.sections.NoteSection):
```
jhenstridge, line 347: I was curious about what the best canonical way of determining the build ID is.  I think this is the relevant part of the GDB source code:

https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=bfd/opncls.c;h=99097a9e39989e377af70e9450a970b009439e70;hb=HEAD#l1797

It seems to use the following process:

1. Find the section named `.note.gnu.build-id`
2. Read the contents of the section (this presumably makes sure its not a `SHT_NOBITS` section).
3. Look for a note with name "GNU" and type "NT_GNU_BUILD_ID".  The description field of that note is  the build ID.

So you are right to switch over to checking sections, but I think we can switch this to `section = elf.get_section_by_name('.note.gnu-build-id'), and work from there.

cjp256: Good idea, will do!  Thanks!

PR #3122, opened 2020-05-13
pluginhandler: make the build environment available to all steps

sergiusens:
LP: #1876260

-----

### DIFF
--- a/snapcraft/internal/lifecycle/_runner.py
+++ b/snapcraft/internal/lifecycle/_runner.py
@@ -27,8 +27,8 @@ from snapcraft.internal import (
     states,
     steps,
 )
-from snapcraft.internal.pluginhandler._part_build_environment import (
-    get_snapcraft_part_environment,
+from snapcraft.internal.pluginhandler._part_environment import (
+    get_snapcraft_part_directory_environment,
 )
 from snapcraft.internal.meta._snap_packaging import create_snap_packaging
 from ._status_cache import StatusCache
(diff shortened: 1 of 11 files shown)


sergiusens:
My questions,

- do we want to filter out build-environment?
- do we want to filter out the plugin environment?
- do we want this environment the above just excluded for pull?

cjp256:
I'd suggesting making `SNAPCRAFT_PART_SRC` available for all steps, like you have done.  But `SNAPCRAFT_PART_BUILD` and `SNAPCRAFT_PART_INSTALL` only for build step.  Prime should have SNAPCRAFT_PRIME, stage should have SNAPCRAFT_STAGE, etc.

I'd suggest keeping `build-environment` only on build step.  If we needed to, we could introduce other `<step>-environment`s if there is a good case...

sergiusens:
On Wed, 2020-05-13 at 20:04:38 -0300, Chris Patterson wrote:

> I'd suggesting making `SNAPCRAFT_PART_SRC` available for all steps,
> like you have done.  But `SNAPCRAFT_PART_BUILD` and
> `SNAPCRAFT_PART_INSTALL` only for build step.  Prime should have
> SNAPCRAFT_PRIME, stage should have SNAPCRAFT_STAGE, etc.

I am keeping SNAPCRAFT_STAGE everywhere to as at build time you might
need to link to things that are staged.

> I'd suggest keeping `build-environment` only on build step.  If we
> needed to, we could introduce other `<step>-environment`s if there is
> a good case...

Yeah, this lines up to what we agreed to in the past as well. Where is
my mind? (pun!)

PR #3126, opened 2020-05-15
pluginhandler: allow cleaning the build step

sergiusens:
PluginV2 based parts were not able to clean due to the improper guard
set at the start of the clean_build method.

-----


tests/spread/plugins/v2/build-and-run-hello/task.yaml
```diff
@@ -43,6 +45,8 @@ execute: |
   cd "../snaps/$SNAP"
 
   # Build what we have and verify the snap runs as expected.
+  snapcraft prime
```
cjp256, line 48: The only suggestion I would have here is to do the snap install and verify both times.  Just to make sure the first build works ;)

PR #3127, opened 2020-05-17
gnome extensions: multiarch bindtextdomain.so support

galgalesh:

-----
This commit adds i386 support to the bindtextdomain.so preload on amd64 systems. 

Before this commit, when you use the `gnome-3-28` extension with Wine on a 64-bit system and run a 32-bit windows app, you get the following error:

```txt
ERROR: ld.so: object '/snap/photoscape/x2/lib/bindtextdomain.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS64): ignored.
```

This is because the patched `bindtextdomain.so` is only compiled for the native arch of the snap, and the `LD_PRELOAD` path only points to that version.

This commit cross-compiles `bindtextdomain.so` to i386 on amd64 systems and the `$LIB` token is used in the `LD_PRELOAD` rpath so the linker automatically preloads the correct library based on the arch of the binary it is linking.

To test this, you can compile the [PhotoScape snap](https://github.com/snapcrafters/photoscape), which runs a 32-bit windows app in wine using the `gnome-3-28` extension.

More info on the `$LIB` token: [ld.so man page](http://man7.org/linux/man-pages/man8/ld.so.8.html), [Using LD_PRELOAD mixed 64bit/32bit environment in Linux](https://stackoverflow.com/a/38253231/1588555).

## Note about compatibility

Although we can not use rpath tokens in the `LD_LIBRARY_PATH` as explained in https://github.com/snapcore/snapcraft/pull/3128, we *can* use them in `LD_PRELOAD` because the `pre/append_dir` functions are not used for `LD_PRELOAD` since it contains a list of files, not directories.

*Note: This PR is a small step towards having better tooling to support 32-bit desktop apps in snapcraft.*


extensions/desktop/gnome/Makefile
```diff
@@ -18,11 +23,16 @@ clean:
 
 $(DEST_LAUNCHER):
 	@cat $(SRC_DIR)/init >> $(DEST_LAUNCHER)
-# tail -n +2 to remove the shebang
+    # tail -n +2 to remove the shebang
```
cjp256, line 26: should be a tab?

cjp256:
I wonder if there is a way we can conditionally include this?  Perhaps with a 'wine' extension?

Adding the arch to the path to support this case seems like a good idea regardless.

hellsworth:
I built this branch and with this snapcraft installed, i tried building photoscape and it failed:

```
Building gnome-3-28-extension 
make -j4
mkdir -p x86_64-linux-gnu
gcc -Wall -O2 -o x86_64-linux-gnu/bindtextdomain.so -fPIC -shared ./../src/bindtextdomain.c -ldl
mkdir -p i386-linux-gnu
gcc -m32 -Wall -O2 -o i386-linux-gnu/bindtextdomain.so -fPIC -shared ./../src/bindtextdomain.c -ldl
In file included from /usr/include/bits/errno.h:26:0,
                 from /usr/include/errno.h:28,
                 from ./../src/bindtextdomain.c:5:
/usr/include/linux/errno.h:1:10: fatal error: asm/errno.h: No such file or directory
 #include <asm/errno.h>
          ^~~~~~~~~~~~~
compilation terminated.
Makefile:25: recipe for target 'desktop-launch' failed
make: *** [desktop-launch] Error 1
Failed to run 'make -j4' for 'gnome-3-28-extension': Exited with code 2.
```

galgalesh:
> I wonder if there is a way we can conditionally include this? Perhaps with a 'wine' extension?

This is a good question. This `.so` is useful for all 32-bit apps, not only `wine` apps, but I don't think any regular 32-bit apps use the `gnome` extensions. I've updated the PR to only include the "enablement" bits so other snaps and future extensions can compile and include multiarch versions of the library if they require it.

galgalesh:
Note: you can test this functionality by building the [PhotoScape `bindtextdomain` branch](https://github.com/snapcrafters/photoscape/tree/bindtextdomain) which uses the [sommelier-core `bindtextdomain branch](https://github.com/snapcrafters/sommelier-core/tree/bindtextdomain), which [builds the 32-bit version of this library](https://github.com/snapcrafters/sommelier-core/blob/bindtextdomain/Makefile#L18) and puts it in the correct location to be used by desktop-launch.

galgalesh:
@sergiusens  the tests keep failing but I don't think it's actually related to this PR. Every time I rerun it, it stops at another point. It's really annoying that this is so flaky..

cjp256:
@galgalesh Yeah, the current errors are certainly unrelated.  The latest `pip` broke things and I'm working to address it...

galgalesh:
Ok, I must be having bad luck then, because almost every time I created a PR in the past year, I had random travis errors :P

Good to know the current issue being worked on :+1:

PR #3132, opened 2020-05-18
go v1 plugin: warn when using go.mod and go < 1.13

sergiusens:
LP: #1879315

-----


cjp256:
Does snapcraft's go mod logic work on older versions? Or should it resort to the non-mod logic?

sergiusens:
Yes, the user should set `GO111MODULE=on`, I would just prefer to warn here and let the project owner do the research (they would already know this from building out of Snapcraft)

sergiusens:
> The warning will be printed twice, once for pull and once for build.

I will look into this.

PR #3133, opened 2020-05-20
go v1 plugin: go.mod support for 1.11 and 1.12

sergiusens:
Add limited support for using Go Modules for 1.11 and 1.12 of Go.

The limitation when using 1.11 and 1.12 is that only the main module is
installed. This is due to limitations in "go build" in those versions,
to find the correct target, a new method "_get_module" is introduced.

A warning is issued when using 1.11 and 1.12 with a link on how to
enable support for Go Modules with precautions to only print the warning
once per execution.

Some general cleanup in GoPluing._build took place to remove repetition
of "domain" specific flag settings for when building and relinking.

LP: #1879315

-----

### DIFF
--- a/snapcraft/plugins/v1/go.py
+++ b/snapcraft/plugins/v1/go.py
@@ -14,7 +14,17 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-"""The go plugin can be used for go projects using `go get`.
+"""The go plugin can be used for go projects.
+
+The plugin supports Go Modules if a go.mod definition is found in the
+sources out of the box if the Go version used is 1.13 or greater.
+
+If using 1.11 or 1.12, enabling through environment flags is required.
+Read more about this at https://github.com/golang/go/wiki/Modules and
+the implementation will only install the binary corresponding to the
+main module.
+
+If a go.mod file is not found, the plugin will "go get".
 
 This plugin uses the common plugin keywords as well as those for "sources".
 For more information check the 'plugins' topic for the former and the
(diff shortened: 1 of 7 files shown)


sergiusens:
This replaces #3132

snapcraft/plugins/v1/go.py
```diff
@@ -14,7 +14,17 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-"""The go plugin can be used for go projects using `go get`.
+"""The go plugin can be used for go projects.
+
+The plugin supports Go Modules if a go.mod definition is found in the
```
cjp256, line 19: perhaps something like s/in the sources out of the box/in the root of the part's source/ ?

snapcraft/plugins/v1/go.py
```diff
@@ -161,25 +177,40 @@ def _setup_base_tools(self, go_channel: str, base: Optional[str]) -> None:
...
+        if current_version < parse_version(_GO_MOD_REQUIRED_GO_VERSION):
+            raise GoModRequiredVersionError(go_version=str(self._go_version))
 
-        if parse_version(self._go_version) < parse_version(_GO_MOD_REQUIRED_GO_VERSION):
-            raise GoModRequiredVersionError(go_version=self._go_version)
+        if not self._go_mod_warning_issued and current_version < parse_version(
+            _GO_MOD_ENV_FLAG_REQUIRED_GO_VERSION
+        ):
+            logger.warning(
+                "Ensure Go Module support is correct for this version of Go. "
```
cjp256, line 209: s/Go module support is correct/build environment configuration is correct/ ?

sergiusens:
I fear that "go.mod found, enabling go module support", could break backwards compatibility. A project might be working in dual mode (transitioning) with these versions.

sergiusens:
I would rather not have the environment variable automatically set, not in this PR at least, this would lead to feature creep. Create a bug if you wish to research if its doable, but this is a corner case scenario that very few people would be using.

cjp256:
I don't want to set GO111MODULE for them, but `GO111MODULE=off` _will_ cause the pull to fail on `go mod download` that snapcraft is running.   If it's unset, go will enable it when it sees go.mod (?).

But maybe there's a scenario I'm not accounting for, and the warning certainly doesn't hurt.

sergiusens:
@cjp256 I totally get where you are coming from. But to be perfectly honest, I was considering even just not fixing this issue and propose the use of override-build.

PR #3139, opened 2020-05-21
pluginhandler: run plugin commands in isolation

sergiusens:
Iterate over plugin commands and run them in isolation. This allows
clearer error reporting. The same exception as in PluginV1 is raised on
command execution.

Fixes SNAPCRAFT-1NQ

-----


cjp256:
Nice!  I will add that it was very handy to have the complete script to iterate on snaps.  Even if we don't use it directly, what do you think about just dumping it anyways?

sergiusens:
Dang it, now I remember why I didn't do this from the start!

PR #3143, opened 2020-05-22
cli: migrate close to use the new channel map

sergiusens:
The implementation was moved to the CLI as being done for all commands.

With this, the new implementation for displaying a channel map is used,
which avoids the bug in the channel_map_tree returned in close that is
not progressive release aware and shows channels mistakenly as open.

The closed_channels API results are skipped as well as they are not
really track aware, returning mixed results for tracks in some scenarios
and plain risks for others (when latest is used, which is not allowed
when using more than one track).

The tests have been simplified to the concerns of the close CLI results
as display variants of the channel map tree are already done under test_status.

A missing todo checklist entry was added to take care of the upload logic.

-----

### DIFF
--- a/snapcraft/cli/store.py
+++ b/snapcraft/cli/store.py
@@ -433,7 +433,40 @@ def close(snap_name, channels):
         snapcraft close my-snap beta
         snapcraft close my-snap beta edge
     """
-    snapcraft.close(snap_name, channels)
+    store = storeapi.StoreClient()
+    account_info = store.get_account_information()
+
+    try:
+        snap_id = account_info["snaps"][DEFAULT_SERIES][snap_name]["snap-id"]
+    except KeyError:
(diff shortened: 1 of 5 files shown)


snapcraft/cli/store.py
```diff
@@ -433,7 +433,40 @@ def close(snap_name, channels):
...
+        click.echo(
+            get_tabulated_channel_map(
+                snap_channel_map,
+                architectures=existing_architectures,
+                tracks=closed_tracks,
+            )
+        )
+        click.echo()
+
+    echo.info(msg)
```
cjp256, line 469: click.echo and echo.info?  Curious if there is any particular reason for the mixed usage?

sergiusens: our echo just wraps things in color or multine strings into proper paragraphs.

PR #3144, opened 2020-05-25
docker: setup a multiarch build for snapcraft docker images

adferrand:
Hello!

During my work to build the certbot snap (Python based) for multiple architectures (see https://github.com/certbot/certbot/pull/8016), I searched a way to execute snapcraft on each target architecture. Running `snapcraft` on the native architecture or through emulation is necessary since the snap Python plugin is not able to make cross-compilation, on the contrary of Go for instance that has this feature at its core in the Go compiler.

To be able to execute this workflow on a CI pipeline (that is running mostly on `amd64`), I choose to use QEMU. Doing so is done best through the `snapstore/snapcraft` docker image, since QEMU is quite well supported on this technology.

As a consequence I ended up rewriting the `snapstore/snapcraft` docker image with QEMU embedded and create it for each target architecture to let that happen. I think it could be useful to other for similar cross-compilation purposes, and more generally to be able to run the snapcraft docker image on more architectures.

This PR proposes the cross-compilation of the docker image for 4 platforms (linux/amd64, linux/386, linux/arm64 and linux/arm/v7) through QEMU using a host on linux/amd64. Appropriate hooks and scripts are added to allow this cross-compilation to happen on Docker Hub Autobuild. On top of it, a multiarch manifest takes place on `snapstore/snapcraft:RISK`, so that any user can target this tag during a `docker pull`, and get the appropriate architecture without explicitly asking for it.

I take also the occasion to merge the 4 Dockerfiles into one thanks to the `ARG` feature.

You can see the result of this on my fork here: https://hub.docker.com/r/adferrand/snapcraft/tags

At the end, 5 tags are pushed: one for each architecture, and one that is a manifest and refer to each architecture.

As a side note, a lot of the logic added in this PR is to cover the lack of features on Docker Hub Autobuild, that uses (ironically) and ancient version of Docker engine. With more recents versions you can take advantaged of of `docker buildx`, that leverages almost all the complexity to cross-compiling. If you are interested in using Travis instead of Docker Hub Autobuild, I will be glad to write the logic!

---


abitrolly:
See also https://github.com/snapcore/snapcraft/pull/2673

sergiusens:
Hi, I am sorry for the long time it took me to get to this, it changes a lot.

I have concerns about merging this as this is the source of our docker images used on docker hub which are published with Ubuntu 16.04 LTS, which means that it is used to create snaps without a base.

If this goes in and replaces the prior images, it will break CI for everyone relying on those images and I am not looking forward to that (I saw the change to focal and use of core20 in the PR).

For what it's worth, @cjp256 is going to be creating a separate repo to hold the docker images.

sergiusens:
I am going to close this, but reference it in the new repos. Again, sorry for the delays and feel free to reach out and I will do my best to provide timely feedback.

PR #3147, opened 2020-05-26
Use `pylxd` instead of `lxc exec`

abitrolly:

-----

`pylxd` ignores default LXD remote, but `lxc exec` does not. This is untested attempt to remove the todo and fix `--use-lxd` failure if default LXD remote is not local.

https://forum.snapcraft.io/t/use-lxd-doesnt-work-with-non-local-remote/17775

### DIFF
--- a/snapcraft/internal/build_providers/_lxd/_lxd.py
+++ b/snapcraft/internal/build_providers/_lxd/_lxd.py
@@ -160,7 +160,8 @@ class LXD(Provider):
     ) -> Optional[bytes]:
         self._ensure_container_running()
 
-        cmd = super()._get_env_command()
+        cmd = []
+        cmd.extend(super()._get_env_command())
         cmd.extend(command)
         logger.debug(f"Executing in {self.instance_name}: {cmd}")
         (exit_code, stdout, stderr) = self._container.execute()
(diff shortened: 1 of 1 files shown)


abitrolly:
Test fail with `AttributeError: 'FakeContainer' object has no attribute 'execute'`. Some mockery is going on, and I am not sure I can fix that.

snapcraft/internal/build_providers/_lxd/_lxd.py
```diff
@@ -159,29 +158,28 @@ def __init__(
...
+        (exit_code, stdout, stderr) = self._container.execute(cmd)
 
-        output = None
-        try:
-            if hide_output:
-                output = subprocess.check_output(cmd)
-            else:
-                subprocess.check_call(cmd)
-        except subprocess.CalledProcessError as process_error:
+        output = stdout + "\n\nSTDERR: " + stderr
```
sergiusens, line 172: This should be streamed and piped out to stdout and stderr as they occur. It may require extending pylxd to support proper websockets for this and the reason we just went with lxc exec.

sergiusens: docstring says this needs to be a callable

sergiusens:
I am closing this for now as it seems to have gone stale. We have on our list to have some functionality for LXD remotes in the future so even if you don't get a chance to work on this, something similar might come up in a future PR.

snapcraft/internal/build_providers/_lxd/_lxd.py
abitrolly: Can you advise on how to make this a callable?

PR #3165, opened 2020-06-08
Update cmake plugin to support Ninja generator

GamePad64:
v2 plugins are meant to be run with core20, so we can safely assume, that we have CMake version 3.16.3
CMake supports --install option since [3.15](https://cmake.org/cmake/help/v3.15/release/3.15.html#command-line), and -j option is available since [3.12](https://cmake.org/cmake/help/latest/release/3.12.html#command-line)
These improvements can fix builds with Ninja generator.

-----


tests/unit/plugins/v2/test_cmake.py
```diff
@@ -64,8 +64,8 @@ class Options:
             Equals(
                 [
                     'cmake "${SNAPCRAFT_PART_SRC_WORK}"',
-                    'cmake --build . -- -j"${SNAPCRAFT_PARALLEL_BUILD_COUNT}"',
-                    'cmake --build . --target install -- DESTDIR="${SNAPCRAFT_PART_INSTALL}"',
+                    'cmake --build . -j"${SNAPCRAFT_PARALLEL_BUILD_COUNT}"',
+                    'cmake --install . --prefix="${SNAPCRAFT_PART_INSTALL}"',
```
cjp256, line 68: I could be wrong, but I don't think this is the correct use of `--prefix`.  While it may certainly install there as a result, it seems likely to have other potential side effects...

tests/unit/plugins/v2/test_cmake.py
```diff
@@ -90,8 +90,8 @@ class Options:
                     "-DCMAKE_INSTALL_PREFIX=/foo "
                     '-DCMAKE_SPACED_ARGS="foo bar" '
                     '-DCMAKE_USING_ENV="$SNAPCRAFT_PART_INSTALL"/bar',
-                    'cmake --build . -- -j"${SNAPCRAFT_PARALLEL_BUILD_COUNT}"',
-                    'cmake --build . --target install -- DESTDIR="${SNAPCRAFT_PART_INSTALL}"',
+                    'cmake --build . -j"${SNAPCRAFT_PARALLEL_BUILD_COUNT}"',
+                    'cmake --install . --prefix="${SNAPCRAFT_PART_INSTALL}"',
```
cjp256, line 94: `--install` does seem to be the right way.

It seems to me that we should be using DESTDIR based on:
https://cmake.org/cmake/help/v3.18/variable/CMAKE_INSTALL_PREFIX.html#variable:CMAKE_INSTALL_PREFIX

But I'm a cmake newb, so feel free to dispute :D

tests/unit/plugins/v2/test_cmake.py
GamePad64: It seems, that you are right. `cmake --install X --prefix Y` is equivalent to `cmake -DCMAKE_INSTALL_PREFIX=Y -P X/cmake_install.cmake`, and setting CMAKE_INSTALL_PREFIX should not be used for packaging purposes.

DESTDIR is the way to go here, but when I try to use Ninja generator, I get an error:
```
cmake --build . --target install -- DESTDIR=XYZ                        
[0/2] Re-checking globbed directories...
ninja: error: unknown target 'DESTDIR=XYZ'
```
Ninja supports DESTDIR using environment variable, while plain-old make does not, and only supports it using a CLI argument.

I think, that the best way to fix Ninja builds is to add a new parameter to plugin: `cmake-generator`, that (if set) passes `-G` parameter to first `cmake` invocation and then:
- if `cmake-generator` contains "Ninja" (case-sensitive), then DESTDIR is passed using env and `cmake --install` is invoked (it is much faster, than cmake --build install).
- else pass DESTDIR as `cmake --build` argument.

This way this plugin will gain support of Ninja build system, which is much faster.

GamePad64:
I've pushed a new approach to support Ninja-specific builds.

snapcraft/plugins/v2/cmake.py
```diff
@@ -69,6 +69,6 @@ def _get_cmake_configure_command(self) -> str:
     def get_build_commands(self) -> List[str]:
         return [
             self._get_cmake_configure_command(),
-            'cmake --build . -- -j"${SNAPCRAFT_PARALLEL_BUILD_COUNT}"',
-            'cmake --build . --target install -- DESTDIR="${SNAPCRAFT_PART_INSTALL}"',
+            'cmake --build . -j"${SNAPCRAFT_PARALLEL_BUILD_COUNT}"',
+            'cmake --install . --prefix="${SNAPCRAFT_PART_INSTALL}"',
```
jhenstridge: This doesn't look equivalent to what you're replacing: `--prefix` is described as setting `CMAKE_INSTALL_PREFIX`, which is different to `DESTDIR`.

GamePad64: I've already removed it from the last revision.

GamePad64:
Please, feel free to comment this implementation. I think, that Ninja build will be beneficial for CMake users.

GamePad64:
Please, feel free to comment this implementation. I think, that Ninja build would be beneficial for CMake users.

snapcraft/plugins/v2/cmake.py
```diff
@@ -48,27 +48,43 @@ def get_schema(cls) -> Dict[str, Any]:
                     "uniqueItems": True,
                     "items": {"type": "string"},
                     "default": [],
+                },
+                "cmake-generator": {
+                    "type": "string",
+                    "default": "Unix Makefiles",
```
cjp256, line 54: Maybe this can be an enum of supported strings?

snapcraft/plugins/v2/cmake.py
```diff
@@ -48,27 +48,43 @@ def get_schema(cls) -> Dict[str, Any]:
...
                 }
             },
         }
 
+
+    def _is_ninja(self) -> bool:
+        return "Ninja" in self.options.cmake_generator
```
cjp256, line 61: Would compare directly, though I'm curious if there are (plans for) more than 1 ninja cmake generator option?

snapcraft/plugins/v2/cmake.py
```diff
@@ -48,27 +48,43 @@ def get_schema(cls) -> Dict[str, Any]:
...
-        return [
+        build_commands = [
             self._get_cmake_configure_command(),
-            'cmake --build . -- -j"${SNAPCRAFT_PARALLEL_BUILD_COUNT}"',
-            'cmake --build . --target install -- DESTDIR="${SNAPCRAFT_PART_INSTALL}"',
+            'cmake --build . -j${SNAPCRAFT_PARALLEL_BUILD_COUNT}',
         ]
+        
+        if self._is_ninja():
+            build_commands += ['DESTDIR="${SNAPCRAFT_PART_INSTALL}" cmake --install .']
```
cjp256, line 86: Can we move DESTDIR to env for both and keep it common?

cjp256: Nevermind read your comment below.

snapcraft/plugins/v2/cmake.py
GamePad64: Actually, I don't know. CMake maintains a list of generators as constants in their codebase. Here it is:

```
* Unix Makefiles               = Generates standard UNIX makefiles.
  Green Hills MULTI            = Generates Green Hills MULTI files
                                 (experimental, work-in-progress).
  Ninja                        = Generates build.ninja files.
  Watcom WMake                 = Generates Watcom WMake makefiles.
  CodeBlocks - Ninja           = Generates CodeBlocks project files.
  CodeBlocks - Unix Makefiles  = Generates CodeBlocks project files.
  CodeLite - Ninja             = Generates CodeLite project files.
  CodeLite - Unix Makefiles    = Generates CodeLite project files.
  Sublime Text 2 - Ninja       = Generates Sublime Text 2 project files.
  Sublime Text 2 - Unix Makefiles
                               = Generates Sublime Text 2 project files.
  Kate - Ninja                 = Generates Kate project files.
  Kate - Unix Makefiles        = Generates Kate project files.
  Eclipse CDT4 - Ninja         = Generates Eclipse CDT 4.0 project files.
  Eclipse CDT4 - Unix Makefiles= Generates Eclipse CDT 4.0 project files.

```

The most useful are `Unix Makefiles` -- just plain-old make and `Ninja`.
I don't think, that anyone uses `Watcom WMake` with CMake for snap projects. And if they do, they can always just use the `nil` plugin or submit a PR here.
`Green Hills MULTI` looks like an IDE for embedded systems, and these systems certainly don't support snap, so, IMO, its native support in this plugin is not necessary, too.
`CodeBlocks`, `CodeLite`, `Sublime Text 2`, `Kate` and `Eclipse CDT4` variants are project generators for use with various IDEs, and they should not be used for CI builds in valid CMake-based projects.

So, I think, we can use an enum with a full comparison and raise an exception, if an unsupported generator is encountered.

jhenstridge: Is there any particular reason a snap developer would want to pick a particular generator?  My understanding was that all backends are intended to produce equivalent builds.

I understand that there are performance benefits to using Ninja over traditional Makefiles though, so is there any reason why we wouldn't want to build all cmake parts using Ninja?

cjp256: Interesting point - I would not expect there to be a problem, but I don't know it for a fact.  I would consider:

1. Projects can set their own generators - setting a non-standard default immediately puts us in potential conflict unless we try to detect it.  If there is an edge case and the project knows best, they might be justified in specifying....

2. I expect ninja would add a dependency on `ninja-build`, though very minor in the scheme of things.

In general, if it's not the build tool's default, I would avoid it being the snapcraft default without good reason.  It tends to lead to issues down the road... just my opinion :D

cjp256: > Actually, I don't know. CMake maintains a list of generators as constants in their codebase. Here it is:
> 
> ```
> * Unix Makefiles               = Generates standard UNIX makefiles.
>   Green Hills MULTI            = Generates Green Hills MULTI files
>                                  (experimental, work-in-progress).
>   Ninja                        = Generates build.ninja files.
>   Watcom WMake                 = Generates Watcom WMake makefiles.
>   CodeBlocks - Ninja           = Generates CodeBlocks project files.
>   CodeBlocks - Unix Makefiles  = Generates CodeBlocks project files.
>   CodeLite - Ninja             = Generates CodeLite project files.
>   CodeLite - Unix Makefiles    = Generates CodeLite project files.
>   Sublime Text 2 - Ninja       = Generates Sublime Text 2 project files.
>   Sublime Text 2 - Unix Makefiles
>                                = Generates Sublime Text 2 project files.
>   Kate - Ninja                 = Generates Kate project files.
>   Kate - Unix Makefiles        = Generates Kate project files.
>   Eclipse CDT4 - Ninja         = Generates Eclipse CDT 4.0 project files.
>   Eclipse CDT4 - Unix Makefiles= Generates Eclipse CDT 4.0 project files.
> ```
> 
> The most useful are `Unix Makefiles` -- just plain-old make and `Ninja`.
> I don't think, that anyone uses `Watcom WMake` with CMake for snap projects. And if they do, they can always just use the `nil` plugin or submit a PR here.
> `Green Hills MULTI` looks like an IDE for embedded systems, and these systems certainly don't support snap, so, IMO, its native support in this plugin is not necessary, too.
> `CodeBlocks`, `CodeLite`, `Sublime Text 2`, `Kate` and `Eclipse CDT4` variants are project generators for use with various IDEs, and they should not be used for CI builds in valid CMake-based projects.
> 
> So, I think, we can use an enum with a full comparison and raise an exception, if an unsupported generator is encountered.

I'd limit it to the use cases of interesting.  makefiles and ninja.

sergiusens:
Chris Patterson <notifications@github.com> writes:

> @cjp256 commented on this pull request.
>
>
>
>> @@ -48,27 +48,43 @@ def get_schema(cls) -> Dict[str, Any]:
>                      "uniqueItems": True,
>                      "items": {"type": "string"},
>                      "default": [],
> +                },
> +                "cmake-generator": {
> +                    "type": "string",
> +                    "default": "Unix Makefiles",

This might be better as a enum.

tests/spread/plugins/v2/snaps/cmake-hello-ninja/snap/snapcraft.yaml
```diff
@@ -0,0 +1,23 @@
+name: cmake-hello
```
sergiusens, line 1: this test is not added to the v2 task.yaml and will likely fail as the name does not match the directory name holding this project

PR #3170, opened 2020-06-12
repo: decouple fetch and unpack stage-packages

sergiusens:

-----

### DIFF
--- a/snapcraft/internal/repo/_deb.py
+++ b/snapcraft/internal/repo/_deb.py
@@ -21,11 +21,11 @@ import io
 import lazr.restfulclient.errors
 import logging
 import os
+import pathlib
 import re
 import subprocess
 import sys
 import tempfile
-from pathlib import Path
 from typing import Dict, List, Optional, Set, Tuple  # noqa: F401
 from typing_extensions import Final
 
@@ -44,8 +44,10 @@ from ._base import BaseRepo, get_pkg_name_parts
 
 logger = logging.getLogger(__name__)
 
-_DEB_CACHE_DIR: Path = Path(BaseDirectory.save_cache_path("snapcraft", "download"))
-_STAGE_CACHE_DIR: Path = Path(
+_DEB_CACHE_DIR: pathlib.Path = pathlib.Path(
+    BaseDirectory.save_cache_path("snapcraft", "download")
+)
+_STAGE_CACHE_DIR: pathlib.Path = pathlib.Path(
     BaseDirectory.save_cache_path("snapcraft", "stage-packages")
 )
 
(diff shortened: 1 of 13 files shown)


snapcraft/internal/repo/_deb.py
```diff
@@ -424,20 +427,28 @@ def install_stage_packages(
...
-
-                    # Mark source of files.
-                    marked_name = f"{pkg_name}={pkg_version}"
-                    cls._mark_origin_stage_package(extract_dir, marked_name)
-
-                    # Stage files to install_dir.
-                    file_utils.link_or_copy_tree(extract_dir, install_dir)
+                file_utils.link_or_copy(
+                    str(dl_path),
+                    (stage_packages_path / os.path.basename(str(dl_path))).as_posix(),
```
cjp256, line 432: dl_path.name gets the base name

cjp256: Also mixed use of str() and as_posix().  I'd suggest always using `str` unless there is a good reason for posix?

snapcraft/plugins/v1/colcon.py
```diff
@@ -380,13 +383,17 @@ def _setup_apt_dependencies(self, apt_dependencies):
 
         logger.info("Installing apt dependencies...")
         try:
-            repo.Ubuntu.install_stage_packages(
```
cjp256, line 383: I probably would have left the `install_stage_packages` as a (refactored) helper so the ROS stuff didn't need to be touched, but it's done now :D

snapcraft/internal/repo/_deb.py
sergiusens: I just copied that over, please make the code suggestion with the change if that is what you are suggesting :-)

snapcraft/plugins/v1/colcon.py
sergiusens: Yeah, I was not aware of how many places this was in, I was thinking of using a tmp thing, but then thought it might be easier to debug in the future that what was unpacked if we can see the debs and not depend on a tmp_dir which I would need to create to solve the decoupling (or pass the list of dl_paths)

snapcraft/internal/repo/_deb.py
cjp256, line 432:
```suggestion
str(stage_packages_path / dl_path.name)
```

cjp256: My first `suggestion` lol

snapcraft/internal/repo/_deb.py
```diff
@@ -424,20 +427,28 @@ def install_stage_packages(
...
+        for pkg_path in stage_packages_path.glob("*.deb"):
+            with tempfile.TemporaryDirectory(suffix="deb-extract") as extract_dir:
+                # Extract deb package.
+                cls._extract_deb(pkg_path, extract_dir)
+                # Mark source of files.
+                marked_name = cls._extract_deb_name_version(pkg_path)
+                cls._mark_origin_stage_package(extract_dir, marked_name)
+                # Stage files to install_dir.
+                file_utils.link_or_copy_tree(extract_dir, install_path.as_posix())
+        cls.normalize(install_path.as_posix())
```
cjp256, line 450:
```suggestion
        cls.normalize(str(install_path))
```

PR #3172, opened 2020-06-16
cli: restore --target-arch with warning for LXD and Multipass

cjp256:
Electron builder currently uses --target-arch.  Until fixes are
upstreamed, restore --target-arch availability for LXD and Multipass,
replacing the error with a warning.

Pops target_arch from kwargs to prevent usage by get_project(),
which is not aware of the build provider.

-----

### DIFF
--- a/snapcraft/cli/lifecycle.py
+++ b/snapcraft/cli/lifecycle.py
@@ -73,7 +73,7 @@ def _execute(  # noqa: C901
     # Temporary fix to ignore target_arch.
     if "target_arch" in kwargs and build_provider in ["multipass", "lxd"]:
         echo.warning(
-            "Ignoring '--target-arch' flag.  This flag requires --destructive-mode, is unsupported, and will be removed in future versions."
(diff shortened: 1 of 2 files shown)


snapcraft/cli/lifecycle.py
```diff
@@ -70,6 +70,13 @@ def _execute(  # noqa: C901
 
     is_managed_host = build_provider == "managed-host"
 
+    # Temporary fix to ignore target_arch.
+    if "target_arch" in kwargs and build_provider in ["multipass", "lxd"]:
+        echo.warning(
+            "Ignoring '--target-arch' flag.  This flag requires --destructive-mode, is unsupported, and will be removed in future versions."
```
sergiusens, line 76: I think this might need some untangling. Maybe instead of "unsupported, "supported as-is with --destructive-mode" and a no-op on other modes or something like that

cjp256: How about `Ignoring '--target-arch' flag.  This flag requires --destructive-mode and is unsupported on LXD and Multipass`?

cjp256: updated

PR #3173, opened 2020-06-16
cli: use snap pack instead of mksquashfs

sergiusens:
Moving to using snap pack for everything as agreed with the snapd team.
This makes it easy to ensure snapcraft is not deviating from snapd
standards. Additionally, --check-skeleton is no longer needed as snap
pack does that out of the box.

Some test fixtures were killed along the way.
The code was moved to snapcraft.cli as most of the implementation left
in lifecycle.pack was mostly presentation related.

The downside here is that this makes it harder to support some commands
in Windows, but luckily Snapcraft for Windows is yet unreleased and the
broken commands (pack and upload-metadata) can eventually be
passthroughed through multipass.

-----


snapcraft/cli/lifecycle.py
```diff
@@ -135,9 +140,65 @@ def _execute(  # noqa: C901
...
+    with subprocess.Popen(
+        snap_command, stdout=subprocess.PIPE, stderr=subprocess.PIPE
+    ) as proc:
+        if indicators.is_dumb_terminal():
+            echo.info("Snapping...")
+            ret = proc.wait()
+        else:
+            message = f"\033[0;32mSnapping \033[0m"
+            progress_indicator = progressbar.ProgressBar(
+                widgets=[message, progressbar.AnimatedMarker()], maxval=7
```
cjp256, line 156: I assume we don't know actual progress here... do one of these progress bar options simply "bounce"?  Then drop the whole count logic below?

sergiusens:
I initially took a look at click.progressbar, it seems no one likes it and there are plenty of bugs, recommendation is to go elsewhere for progressbars. I then found an amazing project that solved most, if not all our issues, https://github.com/rsalmei/alive-progress, had a nice implementation with reversed_dots and saw that Windows is not supported :cry: 

I will take another quick look.

snapcraft/cli/lifecycle.py
sergiusens: We don't and much less now using snap pack. I have moved to a more elegant solution for counting.

PR #3174, opened 2020-06-17
link_or_copy: do not try to create hardlinks to symlinks.

hpoul:
while creating a snapcraft package within a docker image with a mounted mac volume I encountered a weird error:

```
Staging authpass 
[Errno 17] File exists: 'libbsd.so.0.8.7' -> '/code/authpass/stage/lib/x86_64-linux-gnu/libbsd.so.0'
```

It seems like on a mounted mac volume a hard link created for a symlink behaves weirdly (throws an error, but still creates the link). Although I'm not sure what the expected behaviour is. imho the expected behavior would be: if there is a symlink in the source folder, that there is a symlink in the destination folder. So my fix does exactly that. And with this fix the snapcraft build succeeded for me.

The problem can be reproduced even without python, just on the commandline in a docker running on mac:

```
root@1051363b90ac:/home/authpass/tmp# touch first ; ln -s first symlink ; ln symlink hardlink
ln: failed to create hard link 'hardlink' => 'symlink': Input/output error
root@1051363b90ac:/home/authpass/tmp# ls -l
total 0
-rw-r--r-- 2 root root 0 Jun 13 10:53 first
-rw-r--r-- 2 root root 0 Jun 13 10:53 hardlink
lrwxrwxrwx 1 root root 5 Jun 13 10:53 symlink -> first
```

creating a hardlink to the symlink throws an error, but still creates a file. In the python code this results in an exception being thrown, and falls back to copying the file, which fails with 'already exists'.


hpoul:
> Unfortunately CLA check failed, but I have updated branch just in case it was a glitch.

didn't sign it yet, wasn't sure it was worth for the fix.. and not sure what to enter at `Please add the Canonical Project Manager or contact` 🤷‍♂️️

sergiusens:
Just add my name or Martin Wimpress for the CLA paper work

hpoul:
Not sure what's the problem with the CLA, I've now tried signing it multiple times (https://ubuntu.com/legal/contributors/agreement) with both my primary launchpad email address as well as the email address i've used in git.. (which is also added to launchpad) and it's still not recognising it.

PR #3177, opened 2020-06-17
repo: consider virtual pkgs for cache invalidation

sergiusens:
Virtual packages were not considered when checking if all build-packages
are satisfied between runs. Check and resolve with a warning to avoid
use of virtual packages for deterministic results.

Logic has shifted a bit to avoid exposing more functionality from
AptCache and have a nicer separation of concerns.

LP: #1883546

-----

### DIFF
--- a/snapcraft/internal/repo/apt_cache.py
+++ b/snapcraft/internal/repo/apt_cache.py
@@ -140,24 +140,19 @@ class AptCache(ContextDecorator):
     def is_package_valid(self, package_name: str) -> bool:
         return package_name in self.cache or self.cache.is_virtual_package(package_name)
 
-    def is_package_installed(self, package_name: str, package_version: str) -> bool:
-        if self.cache.is_virtual_package(package_name):
+    def get_installed_version(
+        self, package_name: str, *, resolve_virtual_packages: bool = False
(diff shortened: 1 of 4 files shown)


snapcraft/internal/repo/apt_cache.py
```diff
@@ -140,6 +140,23 @@ def _verify_marked_install(self, package: apt.Package):
     def is_package_valid(self, package_name: str) -> bool:
         return package_name in self.cache or self.cache.is_virtual_package(package_name)
 
+    def is_package_installed(self, package_name: str, package_version: str) -> bool:
```
cjp256, line 143: it took me a minute to understand what was going on here... there still is a `is_package_installed()` in _deb.

Maybe `is_package_version_installed`?  still feels weird to have the function do the comparison.

I'd suggest instead adding a new parameter `resolve_virtual_packages: bool` to `get_installed_version(...)` and pass it along to `_deb.is_package_installed(...)`

sergiusens: I missed the one in ._deb, that removes my entire move. I'll move the logic back and add an extra parameter to recurse

PR #3179, opened 2020-06-18
Maven plugin: improve error message when target libs are not found.

edumucelli:
-----
Recently joined Canonical and tried to build a Snap out of a Java library. The error seemed a bit difficult to understand at first (have done mostly Gradle before so was not used to Maven).

I have then improved a bit the error message to explain that it could just that maven-targets declaration does not match the directory in which the Java library has been built. To help the user better understand what could be wrong the error message has been improved to add the directories in which the `arfiles` were searched and clarifying a simple scenario (the one that happened to me) that could lead to fix the situation.

What I think it would help even more is to replace to have the actual part name at this point so that we could `Could not find built jar files for part [the name of the part here] ...". But I did not knew how to get that, with some guidance I could that if that is of interest. I think it could help.

### DIFF
--- a/snapcraft/plugins/v1/maven.py
+++ b/snapcraft/plugins/v1/maven.py
@@ -80,7 +80,7 @@ class EmptyBuildTargetDirectoryError(errors.SnapcraftException):
         return textwrap.dedent(
             """\
             This could happen if your `maven-targets` points to a directory that is not the same as the one
-            in your Maven's <build><directory>target</directory></build> declaration. By default, Snap's 
+            in your Maven's <build><directory>target</directory></build> declaration. By default, Snap's
             Maven plugin won't need `maven-targets` declaration as it will work with the default Maven behavior.
             If you are not sure, try removing `maven-targets` definition and build your Snap again."""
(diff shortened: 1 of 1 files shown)


snapcraft/plugins/v1/maven.py
```diff
@@ -221,7 +224,16 @@ def build(self):
             arfiles = glob(os.path.join(src, "*.[jw]ar"))
 
             if len(arfiles) == 0:
-                raise RuntimeError("Could not find any built jar files for part")
+                raise RuntimeError(
```
sergiusens, line 227: Would you mind defining a plugin specific exception in this module and raising that instead. Inherit from SnapcraftException and implement get_brief and get_resolution, something like https://github.com/snapcore/snapcraft/blob/master/snapcraft/plugins/v1/go.py#L84

edumucelli: @sergiusens thanks for the suggestion, that looks way better, I did not knew there was something already made for this purpose of having explanation and solution. I will do it.

PR #3184, opened 2020-06-22
build providers: check revision before switching

sergiusens:
Snapcraft was switching channels on the availability of a tracking
channel, it may be that this was incorrectly implemented or that snapd
now switched to always report a tracking-channel when installing a snap
with --dangerous after previously having installed a revision from the
Snap Store.

An additional check for revision is now enforced before setting the
switch.

LP: #1884576

-----

### DIFF
--- a/tests/unit/build_providers/test_snap.py
+++ b/tests/unit/build_providers/test_snap.py
@@ -211,7 +211,12 @@ class SnapInjectionTest(unit.TestCase):
                 "channel": "stable",
                 "revision": "123",
             },
-            {"name": "snapcraft", "confinement": "classic", "revision": "x20", "tracking-channel": "latest/stable"},
+            {
+                "name": "snapcraft",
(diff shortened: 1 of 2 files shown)


tests/unit/build_providers/test_snap.py
```diff
@@ -240,6 +245,8 @@ def test_snapcraft_installed_on_host_with_dangerous(self):
             call(["snap-revision", "snap-revision=123", "snap-id=2kkitQ"]),
         ]
         self.get_assertion_mock.assert_has_calls(get_assertion_calls)
+        # Check the call count to ensure the snap switch command does not sneak in.
```
cjp256, line 248: wouldn't happen if you checked all the calls :wink:

cjp256:
./tests/unit/build_providers/test_snap.py:23:1: F401 'testtools.matchers.HasLength' imported but unused

PR #3198, opened 2020-07-03
snap: drop legacy, switch to core20 for bootstrap.

xnox:

-----

This not intended to be merged into master, but like to be maintained as a branch.

I will know in like 10 hours if this works or not.


xnox:
This works, but I don't know if any of this is sane or not, for a base:core20 snapcraft, without legacy support, for bootstrapping riscv64.

tools/snapcraft-override-build.sh
```diff
@@ -15,4 +15,8 @@ patch -s -b "$PYTHON_PACKAGE_PATH/ctypes/__init__.py" patches/ctypes_init.diff
 mkdir -p patched/ctypes
 [ -f "$PYTHON_PACKAGE_PATH/ctypes/__init__.py.orig" ] && mv "$PYTHON_PACKAGE_PATH/ctypes/__init__.py.orig" patched/ctypes
 
-sed -i "$SNAPCRAFT_PART_INSTALL/usr/lib/python3.6/site.py" -e 's/^ENABLE_USER_SITE = None$/ENABLE_USER_SITE = False/'
+sed -i "$SNAPCRAFT_PART_INSTALL/usr/lib/python3.8/site.py" -e 's/^ENABLE_USER_SITE = None$/ENABLE_USER_SITE = False/'
+
+# Upstream ws4py doesn't support python3.7+, fix it with distro patched ws4py
```
cjp256, line 20: Where does the ws4py requirement come from?

xnox: I have no idea. It get pulled as a dep of something? And the version is pulled from pypi, not from distro. Distro has older version, but patched for 3.7+ support. Pypi has newer version, but without python3.7+ support.

I think you can build this on x86_64, and see for yourself when/why ws4py is pulled in and by what?

jhenstridge: It looks to be an optional dependency of pylxd.

xnox: Also discussed in https://github.com/lxc/pylxd/issues/395

cjp256: Thanks for looking into it, since we only support LXD on Linux anyways, relying on the stage-package is fine for this.

sergiusens:
I am closing this as we are not moving to core20 anytime soon due to the required i386 support that we need to keep supporting. We can build this branch onto a channel branch for experimentation for the time being.

PR #3197, opened 2020-07-02
experimental extension support

cjp256:
- [ ] Have you followed the [guidelines for contributing](https://github.com/snapcore/snapcraft/blob/master/CONTRIBUTING.md)?
- [ ] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [ ] Have you successfully run `./runtests.sh static`?
- [ ] Have you successfully run `./runtests.sh tests/unit`?

-----

PR #3203, opened 2020-07-06
experimental ros2 extension & colcon v2 plugin

cjp256:
depends on #3197

re-PR'd from #3097 to get builds pushed to store


kyrofa:
Would love to land this so folks can start using Foxy.

cjp256:
Rebased, should be good to go.

snapcraft/internal/elf.py
```diff
@@ -36,7 +36,7 @@
 
 
 def _ldd_resolve(soname: str, soname_path: str) -> Tuple[str, str]:
-    logger.debug(f"_ldd_resolve: {soname!r} {soname_path!r}")
+    # logger.debug(f"_ldd_resolve: {soname!r} {soname_path!r}")
```
sergiusens, line 39: just noticed this

snapcraft/plugins/v2/colcon.py
```diff
@@ -0,0 +1,247 @@
...
+    @classmethod
+    def get_schema(cls) -> Dict[str, Any]:
+        return {
+            "$schema": "http://json-schema.org/draft-04/schema#",
+            "type": "object",
+            "additionalProperties": False,
+            "properties": {
+                "colcon-ament-cmake-args": {
+                    "type": "array",
+                    "minitems": 1,
```
sergiusens, line 81: IIRC, long standing bug in other places, but if the statement is not minItems then it is completely ignored

sergiusens: notice the camel casing I

cjp256: hah, well i suppose minItems should be 0 then :D did that too

PR #3207, opened 2020-07-08
packaging: use PEP-440 compliant versioning for python & snap

cjp256:
If HEAD matches a tag, then version is simply the tag (e.g. 4.1.1).

If HEAD does not match a tag, then version indicating the distance to
last tag with the ".post" indicator and the git commit revision in
the local identifier (e.g. 4.1.1.post17+git92348a7eb).

Local identifiers should not be published publically, but since we only
push tagged versions to the pip index, this should not occur.

Now we can do package version comparisons, e.g.:
```
>>> from packaging import version
>>> version.parse("4.1.1") > version.parse("4.1.2")
False
>>> version.parse("4.1.1.post17+git92348a7eb") > version.parse("4.1.1")
True
```

Finally, match the snap version to the python package version to
ensure they stay in sync. Do this by querying setup.py for the
version number so only one implementation is required.

-----

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -88,14 +88,15 @@ parts:
         # Put snapcraftctl into its own directory that can be included in the PATH
         # without including other binaries.
         bin/snapcraftctl: bin/scriptlet-bin/snapcraftctl
-    override-pull: |
-        snapcraftctl pull
-        version="$(git describe --always | sed -e 's/-/+git/;y/-/./')"
-        [ -n "$(echo $version | grep "+git")" ] && grade=devel || grade=stable
-        snapcraftctl set-version "$version"
-        snapcraftctl set-grade "$grade"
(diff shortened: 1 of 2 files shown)


snap/snapcraft.yaml
```diff
@@ -88,14 +88,15 @@ parts:
...
-    override-pull: |
-        snapcraftctl pull
-        version="$(git describe --always | sed -e 's/-/+git/;y/-/./')"
-        [ -n "$(echo $version | grep "+git")" ] && grade=devel || grade=stable
-        snapcraftctl set-version "$version"
-        snapcraftctl set-grade "$grade"
     override-build: |
         snapcraftctl build
+
+        version="$(python3 setup.py --version)"
```
sergiusens, line 94: too bad we could not use the setup.py parse-info entry for adopt-info as we are using grade too

PR #3211, opened 2020-07-13
pluginhandler: fix stage-snaps for v2 plugins

jhenstridge:
- [x] Have you followed the [guidelines for contributing](https://github.com/snapcore/snapcraft/blob/master/CONTRIBUTING.md)?
- [x] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [x] Have you successfully run `./runtests.sh static`?
- [x] Have you successfully run `./runtests.sh tests/unit`?

-----
The `PluginHandler._unpack_stage_snaps` method makes use of an attribute on the plugin object that does not seem to exist on v2 plugins, leading to errors like:

```
...
  File "/snap/snapcraft/5143/lib/python3.6/site-packages/snapcraft/internal/lifecycle/_runner.py", line 247, in _run_pull
    self._run_step(step=steps.PULL, part=part, progress="Pulling")
  File "/snap/snapcraft/5143/lib/python3.6/site-packages/snapcraft/internal/lifecycle/_runner.py", line 322, in _run_step
    self._prepare_step(step=step, part=part)
  File "/snap/snapcraft/5143/lib/python3.6/site-packages/snapcraft/internal/lifecycle/_runner.py", line 313, in _prepare_step
    preparation_function()
  File "/snap/snapcraft/5143/lib/python3.6/site-packages/snapcraft/internal/pluginhandler/__init__.py", line 469, in prepare_pull
    self._unpack_stage_snaps()
  File "/snap/snapcraft/5143/lib/python3.6/site-packages/snapcraft/internal/pluginhandler/__init__.py", line 434, in _unpack_stage_snaps
    logger.debug("Unpacking stage-snaps to {!r}".format(self.plugin.stage_snaps))
AttributeError: 'NilPlugin' object has no attribute 'stage_snaps'
```

The value being substituted into the log message doesn't really match the text of the log message, so I switched it to use `self.part_install_dir` instead.

PR #3212, opened 2020-07-13
tests: expand spread coverage for stage-snaps

cjp256:
Reproduces and requires #3211 to pass.

### DIFF
--- a/tests/spread/general/stage_snaps/task.yaml
+++ b/tests/spread/general/stage_snaps/task.yaml
@@ -1,10 +1,8 @@
 summary: stage-snaps
 
-# We currently only have core18 on a stable channel
 systems:
-  - ubuntu-18.04-64
-  - ubuntu-18.04-amd64
-  - ubuntu-18.04-i386
+  - ubuntu-20.04*
+  - ubuntu-18.04*
 
 environment:
   STAGE_SNAPS/default: "hello"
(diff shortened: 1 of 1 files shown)


tests/spread/general/stage_snaps/task.yaml
```diff
@@ -1,10 +1,8 @@
 summary: stage-snaps
 
-# We currently only have core18 on a stable channel
 systems:
-  - ubuntu-18.04-64
-  - ubuntu-18.04-amd64
-  - ubuntu-18.04-i386
+  - ubuntu-20.04*
+  - ubuntu-18.04*
```
jhenstridge, line 5: Is this sufficient?  Below it is setting up a project with `snapcraft init`, which would presumably create a core18 basedsnap on all systems.  You'd need a core20 based snap to exercise v2 plugins.

cjp256: Indeed not. A normal pattern for us is to use `set_base` to correctly set the base to match the host. Added, thanks :)

tests/spread/general/stage_snaps/task.yaml
```diff
@@ -1,20 +1,22 @@
 summary: stage-snaps
 
-# We currently only have core18 on a stable channel
 systems:
-  - ubuntu-18.04-64
-  - ubuntu-18.04-amd64
-  - ubuntu-18.04-i386
+  - ubuntu-20.04*
```
sergiusens, line 4: the architeture specific part of systems can be dropped as the used snaps are no available for all platforms under test. I am weary of this test making our arm64 suite a lot slower than it currently is and have some doubts from the benefits of running these fixtures there, but willing to give it a shot.

sergiusens:
I am closing this in favor of just landing #3211

PR #3213, opened 2020-07-14
cli: use prefix in store name registration hint

sergiusens:
Snapcraft recommended to register <snap-name>-<username> to overcome
concerns about ownership of a given name. It was agreed a while ago to
prefix the username suggestion to have a common namespace for a
publisher for ordering/listing.

Minor updates to quoting were also done.

-----

### DIFF
--- a/snapcraft/cli/store.py
+++ b/snapcraft/cli/store.py
@@ -41,7 +41,7 @@ _MESSAGE_REGISTER_PRIVATE = dedent(
     the choice of name and make sure you are confident nobody else will
     have a stronger claim to that particular name. If you are unsure
     then we suggest you prefix the name with your developer identity,
-    As ‘nessita-yoyodyne-www-site-content’."""
+    As '$username-yoyodyne-www-site-content'."""
 )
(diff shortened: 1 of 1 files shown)


cjp256:
Related LP bug: https://bugs.launchpad.net/bugs/1886223

galgalesh:
*Note: it might be useful to also remove the line about renaming snaps, since that's not possible at the moment*

Given the issues we've had with snaps like `arduino-mhall119`, I don't think snapcraft should recommend pre/suffixing the username of the packager simply because they are not the upstream.

* These snaps always need aliases, increasing the load on the review team.
* Snaps can still be called whatever they want in the GUI storefronts.
* Once upstream or another user takes over maintenance of the snap, the name of the snap becomes incorrect.
* Or alternatively, you get another orphaned snap like `arduino-mhall119`.
* The docs already ask not not include the name of the publisher in the snap.
* Multiple people in the forum recommend against pre/suffixing usernames.

sergiusens:
@galgalesh thanks for the insight, I will bring this up while syncing with the Snap Store team to see if this can actually push for renames.

PR #3218, opened 2020-07-15
lxd: enable security.syscalls.intercept.mknod if supported to allow snaps to create some device nodes

jhenstridge:

-----
When building a bootable base snap, it is common to include a few device files in `/dev` for the benefit of anything running in early boot before udev/devtmpfs is available.

However, trying to build such a snap with the LXD build provider fails because the `mknod` syscall is blocked by default for unprivileged containers.  LXD does provide a way to give unprivileged containers limited access to the syscall for device nodes it considers safe:

https://linuxcontainers.org/lxd/docs/master/syscall-interception#mknod-mknodat

This safe set of devices includes all of the ones included in the `core18` and `core20` snaps, so enabling it should allow bases like that to be built under LXD.

### DIFF
--- a/tests/spread/build-providers/lxd-mknod/task.yaml
+++ b/tests/spread/build-providers/lxd-mknod/task.yaml
@@ -1,5 +1,8 @@
 summary: Verify LXD builds can make (some) device nodes
 
+systems:
+  - -ubuntu-16.04*    # kernel does not support syscall interception
+
 environment:
   SNAP_DIR: ../snaps/devices
 
--- a/tests/spread/build-providers/lxd-mknod/task.yaml
+++ b/tests/spread/build-providers/lxd-mknod/task.yaml
@@ -1,0 +1,37 @@
+summary: Verify LXD builds can make (some) device nodes
+
+environment:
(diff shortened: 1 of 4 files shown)


jhenstridge:
So it looks like I've got something working on the 18.04 and 20.04 systems, but introduces a general failure on 16.04:

```
An error occurred with the instance when trying to start with 'LXD': Common start logic: System doesn't support syscall interception.
Ensure that 'LXD' is setup correctly and try again.
```

It seems 16.04's kernel is missing the features needed to implement this syscall interception feature.  It looks like we should be able to detect this through the pylxd client though.  I'll give that a go.

tests/spread/build-providers/lxd-mknod/task.yaml
```diff
@@ -0,0 +1,40 @@
+summary: Verify LXD builds can make (some) device nodes
+
+systems:
+  - -ubuntu-16.04*    # kernel does not support syscall interception
```
cjp256, line 4: TIL you can filter out systems :D

PR #3220, opened 2020-07-16
review tools: link or copy snap to snap common

sergiusens:
Attempt to link and then copy the snap to review into the review-tools'
SNAP_USER_COMMON area to support the cases where the snap created in an
area the review tools cannot read from or the broader case of having the
"home" interface disconnected.

-----

### DIFF
--- a/snapcraft/internal/review_tools/_runner.py
+++ b/snapcraft/internal/review_tools/_runner.py
@@ -18,16 +18,41 @@ import json
 import pathlib
 import subprocess
 
+from snapcraft import file_utils
 from . import errors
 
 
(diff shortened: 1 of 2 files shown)


cjp256:
Do you anticipate other methods to use the decorator with?  Curious why this approach was taken rather than just wrapping the run logic?

sergiusens:
Chris Patterson <notifications@github.com> writes:

> Do you anticipate other methods to use the decorator with? 
> Curious why this approach was taken rather than just wrapping 
> the run logic?

I do not, it just looked cleaner, I was initially going to go with 
a context
manager and decided on a decorator.

PR #3222, opened 2020-07-18
fix typo

snshn:


snshn:
You're welcome. I'm not sure this PR will pass the CI check since it seems to require a launchpad account which I don't to have, but please feel free to close this PR and replicate this patch in another PR of your own.

PR #3229, opened 2020-07-27
errors: introduce HostToolNotFoundError exception

cjp256:
If a tool on the host is required, this exception
would be thrown if not found.

This differentiates from the ToolMissingError which
would be scoped to tools that are packaged with the
snapcraft snap.

Signed-off-by: Chris Patterson <chris.patterson@canonical.com>

- [ ] Have you followed the [guidelines for contributing](https://github.com/snapcore/snapcraft/blob/master/CONTRIBUTING.md)?
- [ ] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [ ] Have you successfully run `./runtests.sh static`?
- [ ] Have you successfully run `./runtests.sh tests/unit`?

-----

PR #3231, opened 2020-07-27
file utils: rename get_tool_path() to get_snap_tool_path()

cjp256:
Scope get_tool_path() helper to finding tools that are expected to be
found in the snap. As before, if not running as a snap, use shutil.which().

- Rename get_tool_path() -> get_snap_tool_path().

- Remove special casing for "snap" command from get_snap_tool_path().

- Rename _command_path_in_root() to _find_command_path_in_root()
  and return None rather than empty string on error.

Requires #3229, #3230


cjp256:
@sergiusens My thought was to also add a get_host_tool_path() and migrate all of the users of shutil.which() over to it.

PR #3236, opened 2020-07-29
snap: use python3-apt stage-package

cjp256:
Currently python3-apt package is not installable with pip without
a patch.  Switch to the stage-package, which has the benefit
of matching the base (and up-to-date).

https://github.com/pypa/pip/issues/8654

-----

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,6 @@ PyYAML==5.3
 pyxdg==0.26
 requests==2.20.0
 requests_unixsocket==0.1.5
-http://archive.ubuntu.com/ubuntu/pool/main/p/python-apt/python-apt_1.6.5ubuntu0.2.tar.xz; sys_platform == 'linux'
 https://launchpad.net/python-distutils-extra/trunk/2.39/+download/python-distutils-extra-2.39.tar.gz
 requests-toolbelt==0.8.0
 responses==0.5.1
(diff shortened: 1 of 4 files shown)


spread.yaml
```diff
@@ -286,6 +287,14 @@ suites:
      # shellcheck source=/dev/null
      source "$HOME/venv/snapcraft/bin/activate"
      pip install --upgrade pip wheel setuptools
+
+     # Install python-apt using pip 20.1, last known working version.
```
sergiusens, line 291: Instead of this you can try and apt install and use --system-site-packages for pyvenv

spread.yaml
```diff
@@ -286,6 +287,14 @@ suites:
...
+     # https://github.com/pypa/pip/issues/8654
+     pip install --upgrade pip==20.1
```
sergiusens, line 293: Instead of a stage-packages, do you know if we can lock the pip version used? This might be something we should be able to do with snapcraft

requirements.txt
```diff
@@ -6,7 +6,6 @@ PyYAML==5.3
 pyxdg==0.26
 requests==2.20.0
 requests_unixsocket==0.1.5
-http://archive.ubuntu.com/ubuntu/pool/main/p/python-apt/python-apt_1.6.5ubuntu0.2.tar.xz; sys_platform == 'linux'
 https://launchpad.net/python-distutils-extra/trunk/2.39/+download/python-distutils-extra-2.39.tar.gz
```
sergiusens, line 9: IIRC, this dep is only required for python-apt

cjp256: That's what I figured, I was gonna cut it afterwards.

PR #3238, opened 2020-08-04
db: introduce generalized datastore

cjp256:
Provides a generalized datastore for snapcraft components that
need persistent storage.  This datastore provides:

- a TinyDB-based storage system with YAML file-backed storage

- schema versioning with migration support

Initial use will be the environment manager datastore, allowing
snapcraft to better track all snapcraft-created build environments.

-----

### DIFF
--- a/snapcraft/internal/db/datastore.py
+++ b/snapcraft/internal/db/datastore.py
@@ -1,0 +1,112 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2020 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import logging
+import pathlib
+from typing import Any, Dict, List, Type
+
+import tinydb
+import yaml
+
+from . import errors, migration
+
+logger = logging.getLogger()
+
+
+class _YAMLStorage(tinydb.Storage):
+    """Provide YAML-backed storage for TinyDB."""
+
+    def __init__(self, path: str):
+        self.path = pathlib.Path(path)
+
+    def read(self) -> Dict[Any, Any]:
+        """Read database from file."""
+        try:
+            text_data = self.path.read_text()
+        except FileNotFoundError:
+            return dict()
+
+        return yaml.safe_load(text_data)
+
+    def write(self, data) -> None:
+        """Write database (data) to file."""
+        logger.debug(f"saving datstore: {self.path} {data}")
+        self.path.write_text(yaml.dump(data))
+
+    def close(self):
+        """Nothing to do since we do not keep <path> open."""
+
+
+class _YAMLStorageReadOnly(_YAMLStorage):
+    def write(self, data) -> None:
+        """Ignore any writes in read-only mode."""
+
+
+class Datastore:
+    """Datastore class, providing context manager for TinyDB.
+
+    Manages migrations and storage requirements.  If migrations
+    do not indicate support for current datastore version,
+    SnapcraftDatastoreVersionUnsupported will be raised. In that
+    event, some basic fallback mode can be utilized by re-opening
+    datastore in read-only mode."""
+
+    def __init__(
+        self,
+        *,
+        path: pathlib.Path,
+        migrations: List[Type[migration.Migration]],
+        read_only: bool = False,
+    ) -> None:
+        self.path = path
+
+        if read_only:
+            storage_class = _YAMLStorageReadOnly
+        else:
+            storage_class = _YAMLStorage
+
+        self.db = tinydb.TinyDB(str(path), storage=storage_class)
+
+        logger.debug(f"opened datstore: {self.path} read_only: {read_only}")
+
+        # Nothing left to do if opening in read-only mode.
+        if read_only:
+            return
+
+        current_version: int = 0
+        supported_version: int = 0
+
+        for migration_class in migrations:
+            current_version = migration_class(self.db).apply()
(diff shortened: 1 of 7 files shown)


sergiusens:
quick one, seems tinydb was not added to requirements for the unittests

sergiusens:
seems tinydb is missing from "every" setup logic known to exist in our code base

cjp256:
> seems tinydb is missing from "every" setup logic known to exist in our code base

I'm not sure what's worse. That I forgot this obvious piece, or your reference to "every" setup logic.  I started counting... :rofl:

snapcraft/internal/db/datastore.py
```diff
@@ -0,0 +1,112 @@
...
+logger = logging.getLogger()
+
+
+class _YAMLStorage(tinydb.Storage):
+    """Provide YAML-backed storage for TinyDB."""
+
+    def __init__(self, path: str):
+        self.path = pathlib.Path(path)
+
+    def read(self) -> Dict[Any, Any]:
```
sergiusens, line 35: shouldn't we have something where we always return [str, Any]?

snapcraft/internal/db/datastore.py
```diff
@@ -0,0 +1,112 @@
...
+        """Read database from file."""
+        try:
+            text_data = self.path.read_text()
```
sergiusens, line 38: what if the text read is too large?

snapcraft/internal/db/datastore.py
```diff
@@ -0,0 +1,112 @@
...
+        except FileNotFoundError:
+            return dict()
+
+        return yaml.safe_load(text_data)
```
sergiusens, line 42: I thought safe_load could also use a file descriptor, am I misremembering?

snapcraft/internal/db/datastore.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def write(self, data) -> None:
+        """Write database (data) to file."""
+        logger.debug(f"saving datstore: {self.path} {data}")
```
sergiusens, line 46: why do we have a debug for writing but not one for reading?

tests/unit/db/test_datastore.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+@pytest.fixture(autouse=True)
+def mock_datetime(monkeypatch):
+    datetime_mock = Mock(wraps=datetime.datetime)
+    datetime_mock.now.return_value = datetime.datetime(2020, 1, 2, 3, 4, 5, 6)
+    monkeypatch.setattr(migration, "datetime", datetime_mock)
+
+
+def test_get_snapcraft_version(monkeypatch):
```
sergiusens, line 39: is there a need for monkeypatch here? I might have misunderstood autouse

snapcraft/internal/db/migration.py
```diff
@@ -0,0 +1,116 @@
...
+from typing import Any, ClassVar, Dict
+
+import tinydb
+
+import snapcraft
+
+logger = logging.getLogger()
+
+
+def get_snapcraft_version() -> str:
```
sergiusens, line 29: why not pass snapcraft_version=snapcraft.__version__ in Migration.__init__, mocking becomes less of a pain and the coupling would be a bit more obvious.

snapcraft/internal/db/datastore.py
cjp256: Hrm, this interface is really tinydb's depending on what data they send down.  It possibly could be `Any` But in all reality *should* be `Dict[str, Any]`. I'll change it to Dict[str, Any] but will validate the loaded yaml before returning.

snapcraft/internal/db/datastore.py
cjp256: Will use the fd option!

tests/unit/db/test_datastore.py
cjp256: Probably an artifact of previous implementation, will remove.

snapcraft/internal/db/migration.py
cjp256: Sure

snapcraft/internal/db/datastore.py
cjp256: fixed

snapcraft/internal/db/datastore.py
cjp256: fixed

snapcraft/internal/db/datastore.py
cjp256: fixed

snapcraft/internal/db/datastore.py
cjp256: I think I just minimized output, fixed

tests/unit/db/test_datastore.py
cjp256: fixed (removed altogether)

snapcraft/internal/db/migration.py
cjp256: did the same with Datastore, leaving the default there and no-default for Migration

snapcraft/internal/db/datastore.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def __init__(self, path: str):
+        self.path = pathlib.Path(path)
+        logger.debug(f"_YAMLStorage init {self.path}")
+
+    def read(self) -> Dict[str, Any]:
+        """Read database from file."""
+        logger.debug(f"_YAMLStorage read: {self.path}")
+        try:
+            with open(self.path, "r") as fd:
```
sergiusens, line 42: Maybe?
```suggestion
            with self.path.open() as fd:
```

snapcraft/internal/db/migration.py
```diff
@@ -0,0 +1,111 @@
...
+        control_table.insert(control_record)
+
+    def _record_migration(self) -> None:
+        """Record migration in 'migration' table."""
+        migration_table = self.db.table("migration")
+        migration_table.insert(
+            {
+                "schema_version": self.SCHEMA_VERSION,
+                "snapcraft_version": self._snapcraft_version,
+                "timestamp": str(datetime.now(tz=None)),
```
sergiusens, line 64: note about spec

snapcraft/internal/db/datastore.py
cjp256: fixed!

snapcraft/internal/db/migration.py
cjp256: fixed!

sergiusens (approved):

PR #3241, opened 2020-08-08
snap packaging: fetch remote icons configured via appstream

cjp256:
- Consolidate various icon handling into _finalize_icon()
  in _SnapPackaging.finalize_icon().

- Add support for http(s) URLs to 'icon' schema to allow for
  validation for remote URLs once merged.  To be clear, this
  has the effect of allowing http(s) URLs in snapcraft.yaml.
  Alternatively, we would either need to resolve it earlier
  before the validation, or drop the schema validation check
  on 'icon' after the appstream merge.  The approach taken
  here was to make behavior consistent.

- Add spread test for coverage.

-----


snapcraft/internal/project_loader/_config.py
```diff
@@ -19,33 +19,39 @@
...
+from ._env import build_env_for_stage, environment_to_replacements, runtime_env
+from ._extensions import apply_extensions
+from ._parts_config import PartsConfig
 
 logger = logging.getLogger(__name__)
 
 
 @jsonschema.FormatChecker.cls_checks("icon-path")
 def _validate_icon(instance):
+    parsed_url = urllib.parse.urlparse(instance)
```
sergiusens, line 45: I would keep the generic implementation, but still disallow urls as icon entries in snapcraft.yaml

cjp256: Makes sense - I removed the urlparse bits and removed the icon path checking out of this schema validation, into conduct_project_sanity_check().  That way we still get the up-front checking, but doesn't require the schema to enforce such a check (which requires state outside of the yaml config).

snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -357,6 +364,53 @@ def cleanup(self):
                 if os.path.splitext(f)[1] == ".desktop":
                     os.remove(os.path.join(self.meta_gui_dir, f))
 
+    def _finalize_icon(self) -> Optional[pathlib.Path]:
+        """Make sure icon is properly configured, fetching from a remote
```
sergiusens, line 368: This does not fit a proper docstring single-line statement.

Description can follow in the paragraph below

snapcraft/internal/meta/_snap_packaging.py
```diff
@@ -357,6 +364,53 @@ def cleanup(self):
...
+        # The 'icon' specified in the snapcraft.yaml has the same
+        # constraint as (2) and would have already been validated
+        # as existing by the schema.  So we can treat it the same
+        # at this point, regardless of the source of the icon.
+        parsed_url = urllib.parse.urlparse(icon)
+        parsed_path = pathlib.Path(parsed_url.path)
+        icon_ext = parsed_path.suffix[1:]
+        target_icon_path = pathlib.Path(self.meta_gui_dir, f"icon.{icon_ext}")
+
+        os.makedirs(self.meta_gui_dir, exist_ok=True)
```
sergiusens, line 392: you could use

```suggestion
        target_icon_path.parent.mkdir(parents=True, exist_ok=True)
```

sergiusens (changes requested):

snapcraft/internal/meta/_snap_packaging.py
cjp256: fixed

sergiusens (approved):

PR #3243, opened 2020-08-11
Add skip-keys to catkin plugin

Levi-Armstrong:

-----


snapcraft/plugins/v1/catkin.py
```diff
@@ -837,25 +848,28 @@ def snap_fileset(self):
         return fileset
 
 
-def _find_system_dependencies(catkin_packages, rosdep, catkin):
+def _find_system_dependencies(self, rosdep, catkin):
```
cjp256, line 851: This indent is wrong (to move this method into the class).

cjp256:
Welcome @Levi-Armstrong :-) Thanks for submitting a patch, this seems like a good field to add.

Pinging @kyrofa who would be interested.

snapcraft/plugins/v1/catkin.py
Levi-Armstrong: Fixed.

snapcraft/plugins/v1/catkin.py
```diff
@@ -68,6 +68,10 @@
       (string)
       The URI to ros master setting the env variable ROS_MASTER_URI. Defaults
       to http://localhost:11311.
+    - skip-keys:
+      (list of strings)
+      List of catkin packages to skip when resolving dependencies. This is useful
+      when a package has not been added to rosdep but is known to be installed.
```
kyrofa, line 74: I'm not sure "installed" is the right word in the context of building the snap. Maybe "staged"?

snapcraft/plugins/v1/catkin.py
```diff
@@ -256,6 +260,14 @@ def schema(cls):
 
         schema["required"] = ["source"]
 
+        schema["properties"]["skip-keys"] = {
+            "type": "array",
+            "minitems": 1,
```
kyrofa, line 265:
```suggestion
            "minItems": 1,
```

Capital I. We've made this mistake in a number of places as you can see :blush: .

snapcraft/plugins/v1/catkin.py
```diff
@@ -853,7 +866,10 @@ def _find_system_dependencies(catkin_packages, rosdep, catkin):
         # let's get the dependencies for the entire workspace.
         dependencies |= rosdep.get_dependencies()
 
-    for dependency in dependencies:
+    # Remove skip keys
+    filtered_dependencies = [x for x in dependencies if x not in skip_keys]
```
kyrofa, line 870: I suggest loading the skip-keys into a set so you can do this far more efficiently.

tests/unit/plugins/v1/test_catkin.py
```diff
@@ -1909,7 +1910,7 @@ def test_find_system_dependencies_system_only(self):
 
         self.assertThat(
             catkin._find_system_dependencies(
-                {"foo"}, self.rosdep_mock, self.catkin_mock
+                {"foo"}, self.rosdep_mock, self.catkin_mock, {}
```
kyrofa, line 1913: Note that you're passing a set in for skip keys, but skip keys in reality will be a list. Probably not a big deal, but something to consider and proof that we need some mypy magic in the catkin plugin.

kyrofa: Wait no... this is a dict. Why? How? :stuck_out_tongue:

snapcraft/plugins/v1/catkin.py
Levi-Armstrong: Should I fix the others?

kyrofa: Nah, not your problem, we should just have one big PR to fix the existing.

snapcraft/plugins/v1/catkin.py
kyrofa: (we wouldn't want to do that anyway, fix looks good)

snapcraft/plugins/v1/catkin.py
Levi-Armstrong: Should be fixed.

snapcraft/plugins/v1/catkin.py
Levi-Armstrong: ok

tests/unit/plugins/v1/test_catkin.py
Levi-Armstrong: should be fixed

snapcraft/plugins/v1/catkin.py
Levi-Armstrong: fixed.

Levi-Armstrong:
@kyrofa Do you know what I have missed? All the tests are failing because of the following.

`AttributeError: 'props' object has no attribute 'skip_keys'`

kyrofa:
Ah, you need to add a default `skip_keys` attribute to [this](https://github.com/snapcore/snapcraft/pull/3243/files#diff-3e83eb425890f94b528b9e0e18ca27c7R54).

snapcraft/plugins/v1/catkin.py
```diff
@@ -321,6 +334,12 @@ def __init__(self, name, options, project):
         self.catkin_packages = None
         if options.catkin_packages is not None:
             self.catkin_packages = set(options.catkin_packages)
+
+        # Get a unique set of skip-keys
+        self.skip_keys = None
```
kyrofa, line 339: If this was always a set it would simplify your later logic. Note that `None` is a special value for `cakin_packages`, so that's a part of the pattern we don't need here.

```suggestion
        self.skip_keys = set()
```

snapcraft/plugins/v1/catkin.py
```diff
@@ -321,6 +334,10 @@ def __init__(self, name, options, project):
         self.catkin_packages = None
         if options.catkin_packages is not None:
             self.catkin_packages = set(options.catkin_packages)
+
+        # Get a unique set of skip-keys
+        self.skip_keys = set(options.skip_keys)
```
kyrofa, line 339: Ah, you've removed too much, you still have the same bug now, but in a different place: if `options.skip_keys` is `None`, this will raise. This needs to be:

```python
self.skip_keys = set()
if options.skip_keys is not None:
    self.skip_keys = set(options.skip_keys)
```

tests/unit/plugins/v1/test_catkin.py
```diff
@@ -1909,7 +1912,7 @@ def test_find_system_dependencies_system_only(self):
 
         self.assertThat(
             catkin._find_system_dependencies(
-                {"foo"}, self.rosdep_mock, self.catkin_mock
+                {"foo"}, self.rosdep_mock, self.catkin_mock, set([])
```
kyrofa, line 1915: No need for the cast, you can just use `set()` directly (this applies throughout the tests):

```suggestion
                {"foo"}, self.rosdep_mock, self.catkin_mock, set()
```

Levi-Armstrong:
> Ah, you need to add a default `skip_keys` attribute to [this](https://github.com/snapcore/snapcraft/pull/3243/files#diff-3e83eb425890f94b528b9e0e18ca27c7R54).

I don't think this is taking me to the correct location. @kyrofa

kyrofa:
@Levi-Armstrong I'm afraid I don't have permission to push an update to this PR, but apply this patch and your tests will pass:

```
diff --git a/tests/unit/plugins/v1/test_catkin.py b/tests/unit/plugins/v1/test_catkin.py
index 661f2a50..c215689c 100644
--- a/tests/unit/plugins/v1/test_catkin.py
+++ b/tests/unit/plugins/v1/test_catkin.py
@@ -63,6 +63,7 @@ class CatkinPluginBaseTest(PluginsV1BaseTestCase):
             build_attributes = []
             catkin_ros_master_uri = "http://localhost:11311"
             disable_parallel = False
+            skip_keys = []
 
         self.properties = props()
         self.ros_distro = "kinetic"
diff --git a/tests/unit/plugins/v1/test_catkin_tools.py b/tests/unit/plugins/v1/test_catkin_tools.py
index 83c6817e..84b3cbc6 100644
--- a/tests/unit/plugins/v1/test_catkin_tools.py
+++ b/tests/unit/plugins/v1/test_catkin_tools.py
@@ -38,6 +38,7 @@ class CatkinToolsPluginBaseTest(PluginsV1BaseTestCase):
             underlay = None
             rosinstall_files = None
             build_attributes = []
+            skip_keys = []
 
         self.properties = props()
 
@@ -133,6 +134,7 @@ def options():
         underlay = None
         rosinstall_files = None
         build_attributes = []
+        skip_keys = []
 
     return Options()
```

sergiusens:
Hi @Levi-Armstrong mind amending your commits (--amend --reset-author), squash them to one if you like using your company's "authorship"?

I believe @kyrofa mentioned that this is still missing a test as well, in case you want to look into that (through code or conversation)

kyrofa:
> I believe @kyrofa mentioned that this is still missing a test as well, in case you want to look into that (through code or conversation)

Indeed, we're missing a test that actually provides some keys to skip and verifies that the right thing happens.

Levi-Armstrong:
I should have time next week to make the requested changes.

sergiusens:
@Levi-Armstrong will you have time to update this week? We will be cutting a new release and would be nice to have this.

Levi-Armstrong:
I will get this done this week.

sergiusens:
I am closing this PR as it has gone stale, feel free to reopen when you find the time to update the branch.

PR #3251, opened 2020-08-18
build providers: honour http proxy settings for snapd

cjp256:
Configure snapd with the proxy settings configured by the user
for http_proxy and https_proxy.  Do this on each launch so
that any changes in configuration passed are applied.

-----


cjp256:
I believe there is a bit of chicken and egg for the case(s) where snap injection isn't supported...

cjp256:
Hrmm, naturally this was going to break the spread test that passed in http proxy...

snapcraft/internal/build_providers/_base_provider.py
```diff
@@ -259,6 +258,19 @@ def launch_instance(self) -> None:
         # what is on the host
         self._setup_snapcraft()
 
+        # Configure snapd to use http/https proxy.
```
sergiusens, line 261: How about putting this in it's own method and do some "super" tricks in multipass and lxd to minimize the test spread?

tests/unit/build_providers/multipass/test_multipass.py
```diff
@@ -747,6 +746,36 @@ def test_lifecycle(
                 hide_output=False,
                 instance_name="snapcraft-project-name",
             ),
+            call(
```
sergiusens, line 749: We wouldn't need this with a shim like https://github.com/snapcore/snapcraft/pull/3251/files#diff-658f276fdea7891f007be4110b902a42R101 applied to https://github.com/snapcore/snapcraft/pull/3251/files#diff-658f276fdea7891f007be4110b902a42R106 to only the multipass specific bits of the implementation

cjp256: Well this whole test case appears to be about testing the whole thing... the LXD version uses `asserts_has_calls()`, but also enumerates the whole thing (minus these new changes because it never failed).  If the purpose of the test is to simply check that it has any calls, and that each call has the proper provider-specific prefix commands, I can do that.  Otherwise, I'm not sure?

sergiusens: I was trying to transform it into that for changes not to have this ripple effect. It is true that it is not the case for everything today

cjp256: Pushed updates to clean that up in areas touched :+1:

PR #3252, opened 2020-08-18
snapcraft: use system certificates by default for https requests

cjp256:
The `requests` library relies on `certifi` library for CA
certificates, unless the use overrides REQUESTS_CA_BUNDLE
environment variable to point to a different certificate
bundle.

Since there is no way for the user to tell certifi to use
the host certificates, unless using the OS distribution
patched package), have snapcraft always set
REQUESTS_CA_BUNDLE to the host SSL certificates, if they
exist in the usual Debian/Ubuntu location.

Log the setting to debug.

Note that this approach does allow the user to override
this default setting of REQUESTS_CA_BUNDLE with their own.

### DIFF
--- a/snapcraft/cli/_runner.py
+++ b/snapcraft/cli/_runner.py
@@ -16,26 +16,29 @@
 import functools
 import logging
 import os
+import pathlib
 import sys
 
 import click
 
 import snapcraft
 from snapcraft.internal import log
+
+from ._command_group import SnapcraftGroup
+from ._errors import exception_handler
+from ._options import add_provider_options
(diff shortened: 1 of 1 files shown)


snapcraft/cli/_options.py
```diff
@@ -346,6 +349,23 @@ def get_project(*, is_managed_host: bool = False, **kwargs):
     return project
 
 
+def apply_requests_environment() -> None:
```
sergiusens, line 352: maybe add this logic to snapcraft.cli._runner and apply it in .run

PR #3258, opened 2020-08-21
schema/snapcraft.json: allow type: os with build-base

anonymouse64:
This is to build the core snap with a modern version of snapcraft.

We specifically only allow type: os with build-base required and not allowed
with base set, because the only type: os snap that should ever be built is the
core snap, which will not set a base because it is a base.

Signed-off-by: Ian Johnson <ian.johnson@canonical.com>

-----

@cjp256 mentioned that there are probably some tests that need updating, not sure where those are, but probably we should have the following test cases:

- [ ] a snap with `build-base: core` and `type: os` is valid
- [ ] a snap with `base: ...` and `type: os` is invalid
- [ ] a snap with `type: os` and no `build-base` and no `base` still builds with legacy snapcraft (maybe this schema is not used for legacy snapcraft?) this is to ensure that the old snapcraft.yaml for the core snap can continue to be built
maybe there are other cases that should be tested, I can't think of any right now

### DIFF
--- a/schema/snapcraft.json
+++ b/schema/snapcraft.json
@@ -368,6 +368,7 @@
                 "base",
                 "gadget",
                 "kernel",
+                "os",
                 "snapd"
             ]
         },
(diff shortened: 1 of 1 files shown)


sergiusens:
@anonymouse64 will you have time to work on this?

anonymouse64:
Probably not for the next month unfortunately

sergiusens:
Feel free to reopen when you have the time!

PR #3264, opened 2020-08-25
colcon v2 plugin: don't strip environment for stage-runtime-dependencies

cjp256:
HTTP proxy settings are being lost.
    
    - Remove the `-i` argument to `env`, there are no longer variables
      that must be filtered, allow all of the build-time environment
      variables through.  This will include http(s)_proxy once the
      run scripts include it (shipped separately).
    
    - Remove parameters to `stage-runtime-dependencies` which can just
      be picked up from the environment now, after removing `-i` from
      `env`.
    
    - Remove PATH from env list as it no longer needs to be passed
      through explicitly, it is in the run script.

### DIFF
--- a/snapcraft/plugins/v2/colcon.py
+++ b/snapcraft/plugins/v2/colcon.py
@@ -157,7 +157,12 @@ class ColconPlugin(PluginV2):
     def _get_stage_runtime_dependencies_command(self):
         env = dict(LANG="C.UTF-8", LC_ALL="C.UTF-8")
 
-        for key in ["PATH", "SNAP", "SNAP_ARCH", "SNAP_NAME", "SNAP_VERSION"]:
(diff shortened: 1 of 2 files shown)


cjp256:
I've come to the conclusion that we actually should be passing the proxy vars for all of the scriplets we're generating...  we'll still need to drop the `-i` flag from env.

cjp256:
This will require some more consideration.

Currently the environment stripping causes the SNAPCRAFT_STAGE to get unset, which prevents the sitecustomize from adding snap site package directory.  I can unset it, but that's an ugly fight.  I reconsidered bringing "SNAPCRAFT_INTERPRETER=1" back, but I feel like that has the same problem as the PYTHONPATH approach.

Closing for the time being.

PR #3270, opened 2020-08-31
Add PYTHONPATH to environment (fixes #1893262)

hellsworth:

-----


hellsworth:
Tested that if you build an app using the gnome-3-34 extension will now register the correct `PYTHONPATH`:

```
heather@fenrir:/snap/video-downloader/x1$ env | grep PYTHONPATH
PYTHONPATH=:/snap/video-downloader/x1/gnome-platform/usr/lib/python3.6/site-packages:/snap/video-downloader/x1/usr/lib/python3.6/site-packages:/snap/video-downloader/x1/lib/python3.6/site-packages
```

cjp256:
I don't think it's guaranteed `SNAP_DESKTOP_RUNTIME` will be populated before PYTHONPATH.  I would suggest using `$SNAP/gnome-platform` unless we can make that ordering promise... @sergiusens?

hellsworth:
@cjp256 Ah that's a good point about not being sure if `$SNAP_DESKTOP_RUNTIME` would be set already or not so to be safe I went with your suggestion and am using `$SNAP/gnome-platform` in all the places.

kenvandine:
I tested this with the video-downloader snap and a local pygtk test app i wrote for testing, works great!

cjp256:
Any reason we need to be adding paths from the end-user snap into PYTHONPATH?  Specifically the 
`$SNAP/usr/lib/python3.6/site-packages:$SNAP/lib/python3.6/site-packages"`

The gnome-platform import makes sense naturally, but I'm not so sure about the $SNAP paths...

kenvandine:
> Any reason we need to be adding paths from the end-user snap into PYTHONPATH? Specifically the
> `$SNAP/usr/lib/python3.6/site-packages:$SNAP/lib/python3.6/site-packages"`
> 
> The gnome-platform import makes sense naturally, but I'm not so sure about the $SNAP paths...

I think we probably need that to allow a snap to provide their own staged python imports.  My only concern is they might not be a compatible version of python.  So might cause problems, but thankfully they would be apparent to the app developer right away.

hellsworth:
I don't know if `$SNAP/lib/python3.6/site-packages` is necessary actually. I'll rebuild and test..

hellsworth:
> Any reason we need to be adding paths from the end-user snap into PYTHONPATH? Specifically the
> `$SNAP/usr/lib/python3.6/site-packages:$SNAP/lib/python3.6/site-packages"`
> 
> The gnome-platform import makes sense naturally, but I'm not so sure about the $SNAP paths...

@cjp256 You're right the `$SNAP` paths weren't needed. @kenvandine tested by building both apps with a version of snapcraft that only had the gnome-platform bit in the `PYTHONPATH` to verify.

cjp256 (approved):
LGTM, thanks!

hellsworth:
I updated [pick-colour-picker](https://github.com/stuartlangridge/ColourPicker) to [use the gnome-3-34 extension](https://github.com/stuartlangridge/ColourPicker/blob/app/snap/snapcraft.yaml) since it uses python3-gi to see how this performed. ([here is the snapcraft.yaml diff](https://paste.ubuntu.com/p/VFM2vQb4RZ/))

With this gnome-3-34 change, I built it with snapcraft 4.3 that is in stable (built just fine), and installed it. When I ran it, it had a segfault. Then I built pick-colour-picker with a snapcraft with the changes in this pr, and installed it. It ran just fine, I picked a few colors and played with the app a bit. Looks fine.

sergiusens (approved):
Thanks for this extra test!

snapcraft/internal/project_loader/_extensions/gnome_3_34.py
```diff
@@ -88,6 +88,7 @@ def __init__(self, *, extension_name: str, yaml_data: Dict[str, Any]) -> None:
             "environment": {
                 "SNAP_DESKTOP_RUNTIME": "$SNAP/gnome-platform",
                 "GTK_USE_PORTALS": "1",
+                "PYTHONPATH": "$PYTHONPATH:$SNAP/gnome-platform/usr/lib/python3.6/site-packages",
```
sergiusens, line 91: This is a well used variable which might make more sense to have in the command-chain instead or will override the users input. There's also a slight issue here, there is no prior PYTHONPATH defined in the environment, so the actual python path might end up being ".:$SNAP/gnome-platform/usr/lib/python3.6/site-packages" (if the LD_LIBRARY_PATH behavior applies to PYTHONPATH)

sergiusens (changes requested):
Thank you for the additional tests. I took a more thorough look now and gave this some thought. Use of environment for this key might not be the best choice and might be better served by leveraging the extension's generated command-chain. The behavior could be that if PYTHONPATH is -z then and only then set this PYTHONPATH as to not interfere and have a way to override this behavior.

I am not use what happens when you set PYTHONPATH in your yaml with this approach.

snapcraft/internal/project_loader/_extensions/gnome_3_34.py
hellsworth: Good points @sergiusens .. let's explore this..

> The behavior could be that if PYTHONPATH is -z...

I don't quite get your meaning with this -z parameter. Can you elaborate and maybe provide an example of what you're thinking?

> This is a well used variable which might make more sense to have in the command-chain instead or will override the users input.

If the user set the `PYTHONPATH`, then wouldn't that just be the `$PYTHONPATH` going into this environment variable setting? So that the `PYTHONPATH` would end up being `<pythonpath from user setting>:$SNAP/gnome-platform/usr/lib/python3.6/site-packages`?

> so the actual python path might end up being ".:$SNAP/gnome-platform/usr/lib/python3.6/site-packages" 

What problems does this cause? I suppose we could swap the order so that `$PYTHONPATH` is at the end but wouldn't the behavior be the same?

hellsworth: we had a chat about it and decided it was best if PYTHONPATH lives in launcher-specific.

extensions/desktop/gnome/launcher-specific
```diff
@@ -11,6 +11,7 @@ if [ "$wayland_available" = true ]; then
   export QT_QPA_PLATFORM=wayland-egl
 fi
 append_dir GTK_PATH "$SNAP_DESKTOP_RUNTIME/usr/lib/$ARCH/gtk-3.0"
+append_dir PYTHONPATH "$SNAP_DESKTOP_RUNTIME/usr/lib/python3.6/site-packages"
```
hellsworth, line 14: I don't really like hardcoding `python3.6` here but it does work.  
What would be a better way of determining the python version though? On a system with the gnome3-34-1804 snap installed, I see several versions:
```
$ ll /snap/gnome-3-34-1804/current/usr/lib | grep python
drwxr-xr-x  27 root root     4327 Jun  1 07:55 python2.7
drwxr-xr-x   3 root root       36 Jan 29  2019 python3
drwxr-xr-x  30 root root     3628 Jun  1 07:54 python3.6
drwxr-xr-x   4 root root       47 Oct 28  2019 python3.7
drwxr-xr-x   4 root root       47 Oct 28  2019 python3.8
```
I'd like to replace this one append_dir line with something like this:
```
export PYTHON_NUMBERS=`python3 --version | cut -d ' ' -f2 | sed 's/..$//'`
export PYTHON_RUNTIME="python$PYTHON_NUMBERS"
append_dir PYTHONPATH "$SNAP_DESKTOP_RUNTIME/usr/lib/$PYTHON_RUNTIME/site-packages"
```
But I'm not sure if maybe PYTHON_NUMBERS would be better determined from what exists in $SNAP_DESKTOP_RUNTIME/usr/lib.

Thoughts?
```

cjp256: I imagine `launcher-specific` is quite coupled to the contents of the gnome platform snap, so I think it's fine that it is hardcoded there.  Once  a newer version of gnome platform comes out that breaks those paths (e.g. for core20) then that'll have to be forked.

cjp256: You could perhaps allow a flag (e.g. GNOME_PLATFORM_PYTHON_PATH=...) to be passed to `make` and have it insert the correct path when building it too. That may prevent forking and allow the extension to define it in the part.

sergiusens: I suspect this problem goes away if the rogue python directories there are cleaned up, this probably showed up from use of stage-packages and the use of dist-packages. I would just support the default python for the base, as soon as you diverge this becomes a problem. Which actually triggers another test in my mind, we need to test the extension using a python different that the python that is default for the base and the implications of setting this PYTHONPATH, for which, your script is probably doing the right thing but check for the directories existance before exporting :-)

cjp256 (approved):

extensions/desktop/gnome/launcher-specific
hellsworth: Awesome thanks!! Ok sounds like hardcoding is the way to go for now then :)
I tested this with pick-colour-picker and works just fine.

sergiusens (approved):

PR #3277, opened 2020-09-04
pluginhandler: support using patchelf on strict snaps

kyrofa:

-----

Patchelf is automatically used on classic snaps to patch the rpath and interpreter of ELF files in the snap to find things in the snap itself without requiring LD_LIRARY_PATH to be set, etc. This is usually desired, but in the cases that it's not, the snapcraft CLI supports a `no-patchelf` build attribute to disable this behavior.

There are situations where strictly-confined snaps could benefit from this behavior as well (e.g. content sharing, plugin models, etc.). Add a new build attribute called `enable-patchelf` to opt in to rpath and interpreter patching in a strictly-confined snap.


schema/snapcraft.json
```diff
@@ -944,6 +944,7 @@
                             "items": {
                                 "type": "string",
                                 "enum": [
+                                    "enable-patchelf",
                                     "no-patchelf",
```
sergiusens, line 948: I wonder if we should alias no-patchelf to disable-patchelf with a deprecation message when the other is used. That is not on necessarily work to do on this PR

sergiusens (commented):
A thought, do we want to disable the LD_LIBRARY_PATH in `command-chain` when this toggle is used, much like we do with "classic". I guess that is a harder question to answer considering that all-parts would need to subscribe to "enable-patchelf"

schema/snapcraft.json
kyrofa: Yeah that would seems reasonable.

cjp256:
- 'enable' vs 'no' feels odd. maybe 'force'?

- I really am not a fan of the all or nothing approach for patchelf.  Some options:
  1. export `snapcraftctl patchelf` or something to give the user the flexibility to do it themselves with an override. Not the nicest option.
  2. Introduce `patchelf` part property, acting like `prime` list (an opt in / out set of files to affect), defaulting to `[*]` for classic and empty for strict.

kyrofa:
> enable' vs 'no' feels odd. maybe 'force-patchelf' or just 'patchelf'?

That prompted @sergiusens to suggest changing `no-patchelf` to `disable-patchelf` for better parity, how do you feel about that? `force` doesn't feel right since it's not something that happens by default, and I don't like `patchelf` since it doesn't feel clear about what it means (to be fair, "patchelf" probably shouldn't be a user-facing term anyway :stuck_out_tongue:).

>  I'm not really a big fan of the all or nothing approach for patchelf as it's prone to occasionally break things

It's only all-or-nothing on a part-by-part basis. Not all parts need to use it, etc. Once you enable it, it happens for all ELF files provided by the part. Those are the only two use-cases I've seen. When I see folks doing this, it's generally because they're using content sharing in one way or another. I've never seem anyone want to get more fine-grained than this. Do you know of someone who only wanted a single ELF patched, and the rest in a part left alone, and weren't willing to just add patchelf as a build-dep and do it themselves? If not, this just feels like added complication.

sergiusens:
Kyle Fazzari <notifications@github.com> writes:

>> enable' vs 'no' feels odd. maybe 'force-patchelf' or just 
>> 'patchelf'?
>
> That prompted @sergiusens to suggest changing `no-patchelf` to 
> `disable-patchelf` for better parity, how do you feel about 
> that? `force` doesn't feel right since it's not something that 
> happens by default, and I don't like `patchelf` since it doesn't 
> feel clear about what it means (to be fair, "patchelf" probably 
> shouldn't be a user-facing term anyway :stuck_out_tongue:).
>

If I could take it back, I would have named this 
enable/disable-rpath-and-linker or as two separate entities.

>>  I'm not really a big fan of the all or nothing approach for 
>>  patchelf as it's prone to occasionally break things
>
> It's only all-or-nothing on a part-by-part basis. Not all parts 
> need to use it, etc. Once you enable it, it happens for all ELF 
> files provided by the part. Those are the only two use-cases 
> I've seen. When I see folks doing this, it's generally because 
> they're using content sharing in one way or another. I've never 
> seem anyone want to get more fine-grained than this. Do you know 
> of someone who only wanted a single ELF patched, and the rest in 
> a part left alone, and weren't willing to just add patchelf as a 
> build-dep and do it themselves? If not, this just feels like 
> added complication for no gain.

I suspect this is Chris surfacing that moment where we had to 
patch all except one library due to bugs in patchelf. This 
affected stage-packages, which are always a problem in one way or 
another, so I am not sure the special casing is necessary. Need to 
think on it though.

kyrofa:
> A thought, do we want to disable the LD_LIBRARY_PATH in command-chain when this toggle is used, much like we do with "classic". I guess that is a harder question to answer considering that all-parts would need to subscribe to "enable-patchelf"

Note that, with classic, snapcraft simply doesn't generate a command chain at all. There doesn't exist a very nice way to introspect the variables that are defined there and include some but not all. Also, nothing special happens in the case when building a classic snap but all parts are using "no-patchelf."

That said, I've taken a crack at this in the latest commit. It feels a bit fragile, but what do you think?

snapcraft/internal/pluginhandler/_build_attributes.py
```diff
@@ -19,6 +19,9 @@ class BuildAttributes:
     def __init__(self, build_attributes):
         self._attributes = build_attributes
 
+    def enable_patchelf(self):
```
cjp256, line 22: I haven't looked at the implications, but can we consolidate to`is_patchelf_enabled()` to get an interface for the single source of truth, applying the default depending on classic vs. strict? Perhaps in PluginHandler?

cjp256: Ah, I now see `_patching_required` below, but perhaps it'd make sense to move that PluginHandler and have pluginhandler just pass the correct bool..?

kyrofa: Sure, it's mostly a separation of concerns issue and I didn't want to move that knowledge out of here (it used to be below, I just moved it up). Would you also want the libc bits moved to pluginhandler? Context: https://github.com/snapcore/snapcraft/pull/3277/files#diff-b09ac10e7bee2e923d782b720b7dec51R162

cjp256: How about we drop all changes to _patchelf.py (and my suggestion for `is_patchelf_enabled()`) in favor of an updated conditional here: https://github.com/snapcore/snapcraft/blob/37eb4fd15f04a14f337cf004d02e4e54bd657c4d/snapcraft/internal/pluginhandler/__init__.py#L970

Something like:
```
        if self._build_attributes.no_patchelf() and self._project._snap_meta.confinement == "classic":
            logger.warning(
                "The primed files for part {!r} will not be verified for "
                "correctness or patched: build-attributes: [no-patchelf] "
                "is set.".format(self.name)
            )
        elif  self._project._snap_meta.confinement == "classic" or (self._build_attributes.enable_patchelf() and self._project._snap_meta.confinement != "classic"):
            part_patcher = PartPatcher(
                elf_files=elf_files,
                project=self._project,
                build_attributes=self._build_attributes,
                snap_base_path=self._snap_base_path,
                stage_packages=self._part_properties.get("stage-packages", []),
            )
            part_patcher.patch()
```

Personally I'd prefer to see that PartPatcher is only invoked when patching is required.  To reduce coupling, I'm content to have that done in the PluginHandler.  But perhaps @sergiusens has other thoughts?

kyrofa: @cjp256 right, that's fine, but I think that will regress in that it will stop patching in the libc case. I'll look into moving that logic out as well.

kyrofa: Alright @cjp256 I pushed a commit that does this. The patcher still needs a significant refactor to completely eliminate coupling, but it seems out of place here. Take a look?

tests/unit/pluginhandler/test_patcher.py
```diff
@@ -117,40 +117,133 @@ def mock_partpatcher():
...
+                confinement=None,
+                build_attributes=None,
+            ),
+        ),
+        (
+            "os",
+            dict(
+                snap_type="os",
+                snap_name="test-snap",
+                confinement=None,
```
cjp256, line 152: `confinement=strict` may avoid some issues at some point

snapcraft/internal/pluginhandler/_build_attributes.py
cjp256: Agreed that it's way out of scope here. Thanks for the update, LGTM.  Though I'm not sure why the unit tests are now failing on looking for execstack.  I see you patched it below, though I wonder why that's happening as a result of this.

kyrofa: Oh, just because I added some `app` test cases here, which is when execstack is called. Sorry I missed that, I have execstack installed so they passed locally. Should be fixed in the latest commit.

tests/unit/pluginhandler/test_patcher.py
kyrofa: That feels a bit... improper :P . But sure, fixed.

cjp256 (dismissed):
Please yank the last commit 51a65935f83f3c6d3ad05bca86d6148648920bb2 with my apologies, you are correct.

When I noted confinement in my last review, I meant to go look it up before submitting it.  I had thought confinement was a required field, but I clearly have not been creating enough gadget and kernel snaps ;) Type 'base' has confinement though.

And with that, :+1:

kyrofa:
> Please yank the last commit 51a6593 with my apologies, you are correct.

No problem, reverted.

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -965,20 +965,42 @@ def _handle_elf(self, snap_files: Sequence[str]) -> Set[str]:
...
+        #   - `enable-patchelf` is a build attribute
+        #   OR
+        #     - The base is defined
+        #     AND
+        #       - The base is not one of the static bases
+        #     AND
+        #       - The snap uses classic confinement
+        #       OR
+        #         - libc has been staged (as opposed to being in the base snap)
+        patching_required = self._build_attributes.enable_patchelf() or (
```
sergiusens, line 980: This logic feels like we have 3 groups:

- patching required (everything in the latter statement of this logical or)
- patching desired
- pathing required but ignored (no-patchelf)
- patching desired (enable-patchelf) but ignored (no-patchelf)

That last one should not be possible and yet considered in the if below.

I would change this to
```
if patching_required and no_pacthelf():
    logger.warning(...)
elif patching_required or enable_patchelf():
    patch(...)
# For precaution
elif enable_patchelf() and no_patchelf():
    raise RuntimeError(...)
```

schema/snapcraft.json
```diff
@@ -945,6 +945,7 @@
                             "items": {
                                 "type": "string",
                                 "enum": [
+                                    "enable-patchelf",
```
sergiusens, line 948: as jedi master of jsonschema would you mind adding a conflict for enable-patchelf and no-patchelf?

sergiusens (changes requested):
Nice improvement, I have some comments to go over

schema/snapcraft.json
kyrofa: As jedi master of jsonschema, I declare this impossible at least with how much I (as a jedi) have forgotten. List items are validated per-item, not as a set. The only way I see this working is if we made build attributes object properties instead of just a list of strings. So I punted and handled with a proper exception instead, seem okay?

snapcraft/internal/pluginhandler/__init__.py
kyrofa: Good call, done.

sergiusens (approved):

PR #3278, opened 2020-09-04
cli: client side check for setting default tracks

sergiusens:
Check for valid track names within the active set of tracks before
trying to set a default track.

This is in part a workaround for the generic django error provided by
the Snap Store but can live on even after that issue is fixed.

LP: #1892553

-----


snapcraft/cli/echo.py
```diff
@@ -69,6 +70,30 @@ def error(msg: str) -> None:
     click.echo("\033[0;31m{}\033[0m".format(msg))
 
 
+def exit_error(
```
cjp256, line 73: Some unit testing for exit_error would be ideal.

cjp256: I assume the intention is to migrate _print_snapcraft_exception_message() next?

cjp256: Maybe import exit_error for the snapcraft.cli package?

tests/unit/commands/test_set_default_track.py
```diff
@@ -43,17 +43,23 @@ def test_set_default_track_without_login_must_ask(self):
...
         self.fake_metadata.mock.assert_called_once_with(
-            snap_name="snap-test", metadata=dict(default_track="track"), force=True
+            snap_name="snap-test", metadata=dict(default_track="2.0"), force=True
         )
+
+    def test_invalid_track_fails(self):
+        result = self.run_command(["set-default-track", "snap-test", "3.0"])
+
+        self.assertThat(result.exit_code, Equals(2))
+        self.assertThat(result.output, Contains("'2.0', 'latest'"))
```
cjp256, line 65: This could be improved to verify all of `brief`, `resolution` and `details`.

cjp256 (dismissed):
Looks good, just needs a bit more test coverage.  This approach will reduce the number of exceptions :+1:

tests/unit/commands/test_set_default_track.py
sergiusens: This is exactly what I do not want to test, if we ever change the message output in exit_error, the ripple effect from this would eventually just be too cumbersome.

snapcraft/cli/echo.py
sergiusens: Does the extensive mocking required to test this even make it worth it?

tests/unit/commands/test_set_default_track.py
cjp256: I was thinking something like:

```
    @mock.patch("snapcraft.cli.store.echo.exit_error")
    def test_invalid_track_fails(self, mock_exit):
        self.run_command(["set-default-track", "snap-test", "3.0"])

        assert mock_exit.mock_calls == [
            mock.call(
                brief="The specified track does not exist for 'snap-test'.",
                details="Valid tracks: '2.0', 'latest'.",
                resolution="Ensure the track exists and try again.",
            )
        ]
```

cjp256: That verifies all of the strings under test, not what exit_error does with them.

sergiusens: I will think about using a spy and not this exact method as changing this cripples click's test runner

snapcraft/cli/echo.py
sergiusens: I will add errors here

cjp256 (approved):
Nice, LGTM!

PR #3280, opened 2020-09-08
Revert "lxd: update connectivity check url"

cjp256:
Reverts snapcore/snapcraft#3239

Now that the former URL is restored, a quick fix for LP #1894656


sergiusens:
How did you find out it was restored?

cjp256:
Well last week or so ago when Robie asked for 4.1, I checked and the URL doesn't 404 anymore... then assumed it was restored. Then I checked again today after noticing an email update... but perhaps I made a silly assumption :grimacing:

sergiusens (approved):

PR #3282, opened 2020-09-09
v2 plugins: add catkin-tools plugin

kyrofa:

-----

Add a new catkin-tools plugin using v2 of the plugin API. This also uses the ROS functionality shared by the existing Catkin and Colcon plugins.


snapcraft/plugins/v2/catkin_tools.py
```diff
@@ -0,0 +1,120 @@
...
+            },
+        }
+
+    def get_build_packages(self) -> Set[str]:
+        return super().get_build_packages() | {
+            "python3-catkin-tools",
+            # FIXME: Only needed because of a botched release:
+            # https://github.com/catkin/catkin_tools/issues/594#issuecomment-688149976
+            # Once fixed, remove this.
+            "python3-osrf-pycommon",
```
sergiusens, line 68: I am wondering if we should hide this bug like this or just tell people catkin-tools has a missing dependency for folks to carry in their own snapcraft.yaml. This is technical debt we are buying into here

kyrofa: Your call, of course, but once the bug is fixed fetching this package won't be problematic, just unnecessary. Doesn't feel like a heavy load to carry, but folks trying this plugin out will literally immediately be bitten by this and never look back. I don't think it's worth the UX price.

sergiusens (approved):

snapcraft/plugins/v2/catkin_tools.py
cjp256: Couple thoughts:
(1) The extension requires experimental flag anyhow, we could gate stable on the eventual fix.   We could go further if desirable and allow plugins to be marked experimental too.
(2) Could simply log a warning that this package is included and will be removed at a later date when the fix is available.

sergiusens: 1 makes me ok with this, 2 is not a bad idea, linking to the issue

kyrofa: 1. The extension being experimental is hopefully a short-term thing, there are a lot of capabilities that are impossible for folks using that flag (e.g. any of the existing CI systems that build snaps). I've seen no progress on that upstream bug since June; I would encourage not gating anything on it. 
2. That is just another kind of technical debt, but now it's way more obvious and annoying for the user :stuck_out_tongue: .

cjp256: Fair enough, :+1:

cjp256 (approved):
Nicely done!

PR #3283, opened 2020-09-11
specifications: unified provider

cjp256:

-----

### DIFF
--- a/specifications/unified-provider.org
+++ b/specifications/unified-provider.org
@@ -18,6 +18,7 @@ Multipass) and potential future providers (Remote LXD, etc.).
 * Provider Refactoring
 
 ** Current Public Interfaces
+- ensure_provider()
 - setup_provider()
 - mount_project()
 - expose_prime()
@@ -29,19 +30,24 @@ Multipass) and potential future providers (Remote LXD, etc.).
 
 ** Proposed Public Interfaces:
 
-|--------------------------------------------+---------------------------------------------------------------------------+--------------------------------------------------|
-| method                                     | description                                                               | raises                                           |
-|--------------------------------------------+---------------------------------------------------------------------------+--------------------------------------------------|
-| setup_provider(interactive: bool) -> None  | installs/configures provider, optionally allowing for user prompting      | ProviderNotInstalled, ProviderConfigurationError |
-| execute_step(step: steps.Step) -> None     | executes (up to and including) step                                       | NotImplementedError                              |
-| expose_prime(path: pathlib.Path) -> None   | mount project =prime= directory to host at specified path                 | NotImplementedError                              |
-| snap() -> List[pathlib.Path]               | snap project (building as required), writing snap(s) to host project dir. |                                                  |
-| clean_parts(part_names: List[str]) -> None | clean project parts                                                       | NotImplementedError                              |
-| clean() -> None                            | clean all project artifacts, including build-instance, if applicable.     |                                                  |
-| shell() -> None                            | launch interactive shell with build-instance, if applicable.              | NotImplementedError                              |
-|--------------------------------------------+---------------------------------------------------------------------------+--------------------------------------------------|
+|--------------------------------------------+---------------------------------------------------------------------------+-------------------------------------------------------|
+| method                                     | description                                                               | raises                                                |
+|--------------------------------------------+---------------------------------------------------------------------------+-------------------------------------------------------|
+| is_provider_ready() -> bool                | checks if provider is ready w/o installation or configuration             |                                                       |
+| setup_provider(echoer) -> None             | installs/configures provider, allowing for user prompting                 | ProviderInstallationError, ProviderConfigurationError |
+| execute_step(step: steps.Step) -> None     | executes (up to and including) step                                       | NotImplementedError                                   |
+| expose_prime(path: pathlib.Path) -> None   | mount project =prime= directory to host at specified path                 | NotImplementedError                                   |
+| snap() -> List[pathlib.Path]               | snap project (building as required), writing snap(s) to host project dir. |                                                       |
+| clean_parts(part_names: List[str]) -> None | clean project parts                                                       | NotImplementedError                                   |
+| clean() -> None                            | clean all project artifacts, including build-instance, if applicable.     |                                                       |
+| shell() -> None                            | launch interactive shell with build-instance, if applicable.              | NotImplementedError                                   |
+|--------------------------------------------+---------------------------------------------------------------------------+-------------------------------------------------------|
 
 *** Summary of changes:
+- Rename =ensure_provider()= to =is_provider_ready()=, scoping to see if
+  provider is ready without needing installation or further configuration.
+  Remove exceptions.  CLI will invoke setup_provider() if terminal is
+  interactive, allowing user to perform any install/configuration as necessary.
 - Remove =mount_project()=. Project path will be provided to __init__() and will
   be verified/mounted as required.
 - Replace =pack_project()= with =snap()=, adding the implication that build steps
--- a/specifications/unified-provider.org
+++ b/specifications/unified-provider.org
@@ -46,13 +46,12 @@ Multipass) and potential future providers (Remote LXD, etc.).
 
 #+begin_src python
 class Provider(abc.ABC):
-  def __init__(self, project_dir: pathlib.Path, project: Project, environment_manager_db: tinydb.TinyDB, ...) ->
+  def __init__(self, project: Project, **bp_kwargs) ->
   None:
     """Save instance attributes, take no actions."""
 
   def __enter__(self) -> Provider:
-    """Start-of-session actions, not to include launching VMs/instances.
-    """
+    """Start-of-session actions, not to include launching VMs/instances."""
 
   def __exit__(self) -> None:
     """End-of-session cleanup that may be configurable.
@@ -91,8 +90,6 @@ class Provider(abc.ABC):
       :raises SnapcraftEnvironmentError: for failures setting up provider.
       """
 
-      ...
-
     def execute_step(self, step: step.Step, part_name: Optional[List[str]]) -> None
       """Execute (up to, and including) specified step.
 
@@ -102,8 +99,6 @@ class Provider(abc.ABC):
       :raises SnapcraftProviderUnsupportedMethod: if provider does not support
           executing steps.
       """
-
-      ...
 
     def expose_prime(self, prime_dir: pathlib.Path) -> None:
       """Make project's prime directory available to host's prime_dir.
--- a/specifications/unified-provider.org
+++ b/specifications/unified-provider.org
@@ -1,0 +1,82 @@
+#+TITLE: Unified Provider
+#+AUTHOR: Chris Patterson <chris.patterson@canonical.com>
+#+DATE: [2020-09-11]
+
+* Scope:
+
+Snapcraft currently implements build providers for LXD and Multipass.  Host
+builds are driven with independent logic, and Launchpad is indendently
+implemented for remote-build.  With the re-introduction of remote LXD support,
+it is desirable to refactor all providers into a consistent implementation that
+has varied support for Snapcraft capabilities.
+
+This document defines an approach to refactor build provider
+=snapcraft.internal.build_providers._base_provider.Provider= such that it would
+support all current build providers (Host, Launchpad, LXD, Managed-Host,
+Multipass) and potential future providers (Remote LXD, etc.).
+
+* Provider Refactoring
+
+** Current Public Interfaces
+
+- setup_provider()
+- mount_project()
+- expose_prime()
+- execute_step()
+- pack_project()
+- clean(part_names)
+- clean_project()
+- shell()
+
+** Proposed Public Interfaces:
+
+|--------------------------------------------+------------------------------------------------------------------------+--------------------------------------------------|
+| method                                     | description                                                            | raises                                           |
+|--------------------------------------------+------------------------------------------------------------------------+--------------------------------------------------|
+| setup_provider(interactive: bool) -> None  | installs/configures provider, optionally allowing for user prompting   | ProviderNotInstalled, ProviderConfigurationError |
+| execute_step(step: steps.Step) -> None     | executes (up to and including) step                                    | NotImplementedError                              |
+| expose_prime(path: pathlib.Path) -> None   | mount project =prime= directory to host at specified path              | NotImplementedError                              |
+| snap() -> None                             | snap project (building as required), writing snap to host project dir. |                                                  |
+| clean_parts(part_names: List[str]) -> None | clean project parts                                                    | NotImplementedError                              |
+| clean()                                    | clean all project artifacts, including build-instance, if applicable.  |                                                  |
+| shell()                                    | launch interactive shell with build-instance, if applicable.           | NotImplementedError                              |
+|--------------------------------------------+------------------------------------------------------------------------+--------------------------------------------------|
+
+*** Summary of changes:
+- Remove =mount_project()=. Project path will be provided to __init__() and will
+  be verified/mounted as required.
+- Refactor =setup_provider()= and =ensure_provider()= and =ProviderNotFound= to
+  a simpler =setup_provider()= with a parameter enabling user interactive mode
+  to allow anything that needs to be installed or configured.
+- Rename =pack_project()= to =snap()=, adding the implication that build steps
+  are to be executed as required.
+- Rename =clean()= to =clean_parts()=, to improve clarity of scope.
+- Rename =clean_project()= to =clean()=, to improve clarity of scope.
+
+** Migrate current logic into new providers:
+- HostProvider
+- ManagedHostProvider, inherits much of HostProvider.
+- LaunchpadProvider
+- RemoteLXDProvider, inherits much of LXDProvider, overrides the (now internal)
+  equivalent of mount_project() behavior with a rsync from host to remote.
+
+* CLI Handling
+
+** Provider NotImplementedError
+
+Certain commands will not supported by all providers. E.g.
+- =snapcraft try= is not compatible with Launchpad, Remote LXD, Host
+- =snapcraft <step>= is not compatible with Lauchpad, Remote LXD
+- =snapcraft clean <parts>= is not compatible wiht Launchpad, Remote LXD
+
+In these cases, the applicable CLI methods will catch the NotImplementedError
+and inform the user that this command is not available with the current
+configured provider, e.g.:
+
+=Sorry, '<command>' is not available for the currently configured provider
+'<provider>'.=
+
+For =snapcraft clean <parts>=:
+
+=Sorry, specifying parts '<parts>' for 'snapcraft clean' is not supported by the
+currently configured provider '<provider>'.=
--- a/specifications/unified-provider.org
+++ b/specifications/unified-provider.org
@@ -30,17 +30,17 @@ Multipass) and potential future providers (Remote LXD, etc.).
 
 ** Proposed Public Interfaces:
 
-|--------------------------------------------+------------------------------------------------------------------------+--------------------------------------------------|
-| method                                     | description                                                            | raises                                           |
-|--------------------------------------------+------------------------------------------------------------------------+--------------------------------------------------|
-| setup_provider(interactive: bool) -> None  | installs/configures provider, optionally allowing for user prompting   | ProviderNotInstalled, ProviderConfigurationError |
-| execute_step(step: steps.Step) -> None     | executes (up to and including) step                                    | NotImplementedError                              |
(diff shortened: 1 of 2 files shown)


specifications/unified-provider.org
```diff
@@ -0,0 +1,84 @@
...
+- expose_prime()
+- execute_step()
+- pack_project()
+- clean(part_names)
+- clean_project()
+- shell()
+
+** Proposed Public Interfaces:
+
+|--------------------------------------------+---------------------------------------------------------------------------+-------------------------------------------------------|
```
sergiusens, line 33: How about making this class like (ABC) where methods are methods, descriptions and raises docstrings

specifications/unified-provider.org
```diff
@@ -0,0 +1,84 @@
...
+| snap() -> List[pathlib.Path]               | snap project (building as required), writing snap(s) to host project dir. |                                                       |
+| clean_parts(part_names: List[str]) -> None | clean project parts                                                       | NotImplementedError                                   |
+| clean() -> None                            | clean all project artifacts, including build-instance, if applicable.     |                                                       |
+| shell() -> None                            | launch interactive shell with build-instance, if applicable.              | NotImplementedError                                   |
+|--------------------------------------------+---------------------------------------------------------------------------+-------------------------------------------------------|
+
+*** Summary of changes:
+- Rename =ensure_provider()= to =is_provider_ready()=, scoping to see if
+  provider is ready without needing installation or further configuration.
+  Remove exceptions.  CLI will invoke setup_provider() if terminal is
```
sergiusens, line 49: I don't understand this statement, can you make it clearer?

specifications/unified-provider.org
```diff
@@ -0,0 +1,84 @@
...
+  interactive, allowing user to perform any install/configuration as necessary.
+- Remove =mount_project()=. Project path will be provided to __init__() and will
```
sergiusens, line 51: Can you detail what the scope of __init__ will be?

specifications/unified-provider.org
```diff
@@ -0,0 +1,84 @@
...
+  be verified/mounted as required.
+- Replace =pack_project()= with =snap()=, adding the implication that build steps
+  are to be executed as required (and packing may not be done by Snapcraft).
```
sergiusens, line 54: (and packing may not be done by Snapcraft) -> please explain

specifications/unified-provider.org
```diff
@@ -0,0 +1,84 @@
...
+- LaunchpadProvider
+- RemoteLXDProvider, inherits much of LXDProvider, overrides the (now internal)
+  equivalent of mount_project() behavior with a rsync from host to remote.
+
+* CLI Handling
+
+** Provider NotImplementedError
+
+Certain commands will not supported by all providers. E.g.
+- =snapcraft try= is not compatible with Launchpad, Remote LXD, Host
```
sergiusens, line 70: it should be a noop for Host, in the future, we should call the snap command that does the try for you

sergiusens: maybe that should be part of this work :-) (as in defined here)

specifications/unified-provider.org
```diff
@@ -0,0 +1,84 @@
...
+- =snapcraft <step>= is not compatible with Lauchpad, Remote LXD
+- =snapcraft clean <parts>= is not compatible with Launchpad, Remote LXD
```
sergiusens, line 72: these latter two would be supported with remote-lxd in the future, we will just progressively implement. Care to describe that so the full vision is presented?

There is also no comment about --shell, --shell-after and --debug

specifications/unified-provider.org
```diff
@@ -0,0 +1,84 @@
...
+
+In these cases, the applicable CLI methods will catch the NotImplementedError
+and inform the user that this command is not available with the current
+configured provider, e.g.:
+
+=Sorry, '<command>' is not available for the currently configured provider
```
sergiusens, line 78: Don't say sorry :-)

specifications/unified-provider.org
```diff
@@ -0,0 +1,84 @@
...
+'<provider>'.=
+
+For =snapcraft clean <parts>=:
+
+=Sorry, specifying parts '<parts>' for 'snapcraft clean' is not supported by the
```
sergiusens, line 83: ditto

sergiusens (changes requested):
A good first start!

specifications/unified-provider.org
cjp256: In that sense that Launchpad may pack the snap, Snapcraft just fetches it.  Will clarify.

specifications/unified-provider.org
cjp256: Good idea!

specifications/unified-provider.org
cjp256: Ah yes, good points.

I'll write it such that Remote LXD will support <step> and <parts>.  I'm actually not sure why I wrote it this way :thinking: 

Will add that Launchpad will not support --shell, --shell-after and --debug.

specifications/unified-provider.org
cjp256: Sorry!

specifications/unified-provider.org
cjp256: will fix

specifications/unified-provider.org
cjp256: But org tables are so much fun... i'll do that.

specifications/unified-provider.org
cjp256: will do

specifications/unified-provider.org
cjp256: I will put that in the class definition

specifications/unified-provider.org
cjp256: should be done

specifications/unified-provider.org
```diff
@@ -2,13 +2,27 @@
...
-implemented for remote-build.  With the re-introduction of remote LXD support,
-it is desirable to refactor all providers into a consistent implementation that
-has varied support for Snapcraft capabilities.
+** What are we trying to do
+
+Unify all build providers with a set of interfaces that can support
+all providers, each with their own limitations and capabilities.
+
+This will allow us to more easily introduce new providers, such
+as Remote LXD, and improve consistency for existing ones.
```
sergiusens, line 13:
```suggestion
This will allow us to more easily introduce new providers and improve
consistency for existing ones.
```

specifications/unified-provider.org
```diff
@@ -28,57 +42,197 @@ Multipass) and potential future providers (Remote LXD, etc.).
...
-- Remove =mount_project()=. Project path will be provided to __init__() and will
-  be verified/mounted as required.
+** Proposed draft of public interfaces
+
+#+begin_src python
+class ProviderV2:
+  def __init__(self, project_dir: pathlib.Path, project: Project, environment_manager_db: tinydb.TinyDB, ...) ->
+  None:
+    """Save instance attributes, take no actions."""
+    ...
```
sergiusens, line 52:
```suggestion
```

specifications/unified-provider.org
```diff
@@ -28,57 +42,197 @@ Multipass) and potential future providers (Remote LXD, etc.).
...
+
+  def __enter__(self) -> Provider:
+    """Start-of-session actions, not to include launching VMs/instances.
+
+    Examples include:
+    - querying datastore for relevant properties/configuration.
+    """
+    ...
+    return self
```
sergiusens, line 61: The signature explains return self
```suggestion
```

specifications/unified-provider.org
```diff
@@ -28,57 +42,197 @@ Multipass) and potential future providers (Remote LXD, etc.).
...
-- Rename =ensure_provider()= to =is_provider_ready()=, scoping to see if
-  provider is ready without needing installation or further configuration.
-  Remove exceptions.  CLI will invoke setup_provider() if terminal is
-  interactive, allowing user to perform any install/configuration as necessary.
-- Remove =mount_project()=. Project path will be provided to __init__() and will
-  be verified/mounted as required.
+** Proposed draft of public interfaces
+
+#+begin_src python
+class ProviderV2:
```
sergiusens, line 48: Is this abstract?

specifications/unified-provider.org
```diff
@@ -28,57 +42,197 @@ Multipass) and potential future providers (Remote LXD, etc.).
...
+
+      Used as a pre-check prior to initiating build. If False, indicates that
+      additional installation, or configuration must be performed by user, which
+      can be done with setup_provider().
+
+      :returns: True if ready, False if further setup is required.
+
+      """
+
+      ...
```
sergiusens, line 92:
```suggestion
```

specifications/unified-provider.org
```diff
@@ -28,57 +42,197 @@ Multipass) and potential future providers (Remote LXD, etc.).
...
+  def is_provider_ready(self) -> bool:
+    """Check if provider is accessible.
+
+      As applicable, validate that the provider is:
+      - installed
+      - configured
+      - running
+      - compatible
+
+      Used as a pre-check prior to initiating build. If False, indicates that
```
sergiusens, line 84: build is a loaded word and can mean many things, mind picking something else?

specifications/unified-provider.org
```diff
@@ -28,57 +42,197 @@ Multipass) and potential future providers (Remote LXD, etc.).
...
+      If interactive, prompt user for all necessary changes. If non-interactive,
+      attempt to make all necessary changes.
+
+      :param interactive: Toggle for user input.
+
+      :raises ProviderSetupError: if provider has dependencies that need to be
+      installed, or configuration that must be changed.
+      """
+
+      ...
```
sergiusens, line 106:
```suggestion
```

specifications/unified-provider.org
```diff
@@ -28,57 +42,197 @@ Multipass) and potential future providers (Remote LXD, etc.).
...
+
+    def setup_provider(self, interactive: bool) -> None:
+      """Install/configure as necessary.
+
+      If interactive, prompt user for all necessary changes. If non-interactive,
+      attempt to make all necessary changes.
+
+      :param interactive: Toggle for user input.
```
sergiusens, line 100: How is user input handled?

specifications/unified-provider.org
sergiusens, line 48: Just call it `Provider`, we won't have two versions of this living for long as we do for plugins and it is all internal anyways

specifications/unified-provider.org
```diff
@@ -28,57 +42,197 @@ Multipass) and potential future providers (Remote LXD, etc.).
...
+      """Execute (up to, and including) specified step.
+
+      :param step: Step to execute.
+      :param part_names: Optional list of part names to execute step for.
+
+      :raises SnapcraftProviderUnsupportedMethod: if provider does not support
+      executing steps.
+      """
+
+      ...
```
sergiusens, line 118:
```suggestion
```

specifications/unified-provider.org
```diff
@@ -28,57 +42,197 @@ Multipass) and potential future providers (Remote LXD, etc.).
...
+      for the given provider (i.e. bind mounting directory from container).
+
+      :param prime_dir: Directory to mount/sync prime directory to on host.
+
+      :raises SnapcraftProviderUnsupportedMethod: if provider does not support
+      syncing/mounting project's prime directory.
+      """
+
+    def snap(self, output_dir: pathlib.Path) -> List[pathlib.Path]:
+      """Snap project, building as required.
```
sergiusens, line 133: different word for building - > running through lifecycle steps

sergiusens: or "executing lifecycle steps" which matches the previous docstring

specifications/unified-provider.org
```diff
@@ -28,57 +42,197 @@ Multipass) and potential future providers (Remote LXD, etc.).
...
+
+      Write output snaps to host project directory.
+
+      :param output_dir: Directory to write snaps to.
+
+      :returns: Path to snap(s) created from build.
+      """
+
+      ...
```
sergiusens, line 142:
```suggestion
```

specifications/unified-provider.org
```diff
@@ -28,57 +42,197 @@ Multipass) and potential future providers (Remote LXD, etc.).
...
+     def clean_parts(self, part_names: List[str]) -> None:
+      """Clean specified parts.
+
+      :param part_names: List of parts to clean.
+
+      :raises SnapcraftProviderUnsupportedMethod: if provider does not support
+      executing steps.
+      """
+
+      ...
```
sergiusens, line 153:
```suggestion
```

specifications/unified-provider.org
```diff
@@ -28,57 +42,197 @@ Multipass) and potential future providers (Remote LXD, etc.).
...
+
+      Purges all artifacts from using the provider to build the
+      project.  This includes build-instances (containers/VMs) and
+      associated metadata and records.
+
+      This does not include any artifacts that have resulted from
+      a call to snap(), i.e. snap files or build logs.
+      """
+
+      ...
```
sergiusens, line 166:
```suggestion
```

specifications/unified-provider.org
```diff
@@ -28,57 +42,197 @@ Multipass) and potential future providers (Remote LXD, etc.).
...
+
+     def clean_parts(self, part_names: List[str]) -> None:
+      """Clean specified parts.
+
+      :param part_names: List of parts to clean.
+
+      :raises SnapcraftProviderUnsupportedMethod: if provider does not support
+      executing steps.
```
sergiusens, line 150: We usually anchor text after the colon

specifications/unified-provider.org
```diff
@@ -28,57 +42,197 @@ Multipass) and potential future providers (Remote LXD, etc.).
...
+      If available, load environment used for the given step and part_name.
+      Step and part_name must both be supplied for environment to be loaded,
+      otherwise both are ignored.
+
+      Note this method does not imply execute_step(), it is up to the caller
+      to launch the shell at the appropriate time.
+
+      :param step: step to load environment for.
+      :param part_name: part_name to load environment for.  If step is specified,
+      but not a part_name.
```
sergiusens, line 180: anchor

specifications/unified-provider.org
```diff
@@ -28,57 +42,197 @@ Multipass) and potential future providers (Remote LXD, etc.).
...
+
+      :raises SnapcraftProviderUnsupportedMethod: if provider does not support
+      providing a shell.
```
sergiusens, line 183: anchor

specifications/unified-provider.org
```diff
@@ -28,57 +42,197 @@ Multipass) and potential future providers (Remote LXD, etc.).
...
+      """
+
+      ...
```
sergiusens, line 186:
```suggestion
```

specifications/unified-provider.org
```diff
@@ -28,57 +42,197 @@ Multipass) and potential future providers (Remote LXD, etc.).
...
+- =setup_provider()= becomes an instance method, supporting non-interactive
+  configuration, if possible.
+- Removes =mount_project()= interface. Provider will mount project as required.
 - Replace =pack_project()= with =snap()=, adding the implication that build steps
-  are to be executed as required (and packing may not be done by Snapcraft).
+  are to be executed as required, removing the implication that Snapcraft itself
+  invokes the packing (Snapcraft may just fetch the output of the provider).
 - Rename =clean()= to =clean_parts()=, to improve clarity of scope.
 - Rename =clean_project()= to =clean()=, to improve clarity of scope.
+- Context manager =__enter__()= and =__exit__()= are currently used to define
```
sergiusens, line 199: This feel unbalanced, if all queries are related to the tinydb instance, why not query using tinydb directly (datastore instance) and then pass the relevant parameters into the provider?

specifications/unified-provider.org
```diff
@@ -28,57 +42,197 @@ Multipass) and potential future providers (Remote LXD, etc.).
...
-=Sorry, '<command>' is not available for the currently configured provider
+*** Lifecycle command =snapcraft <step> ...=
+
+If attempting to run a lifecycle command (other than =snap=) and
+the provider does not support it, fail with the following error:
+='<command>' is not available for the currently configured provider
 '<provider>'.=
 
-For =snapcraft clean <parts>=:
+For =--debug=, =--shell=, and =--shell-after= options:
```
sergiusens, line 224: should these be enforced by Provider sanity checks and not the cli implementation? It is not clarified here.

sergiusens (changes requested):

specifications/unified-provider.org
cjp256: I was planning on cli and having the cli handle the exception that it's not supported.  However I just realized some operations may run before it hits the exception. Perhaps it'd be best to have the providers advertise the features they support, or add to provider sanity checks.  I'll consider it further

specifications/unified-provider.org
cjp256: My original thought wass that the provider would know how to interpret the record(s) relevant to the provider.  But I can imagine another set of interfaces could manage that and pass it off as provider-specific kwargs.  That's probably preferable.

cjp256: I removed the datastore reference, I agree that shouldn't be there.

WRT 'imbalanced', are you referring to the shutdown of instance?  The other option was add a `close()` method and have the caller take care of it (probably with a `finally`).  I think I'd rather do that, but I mostly went this route because I couldn't come up with a good method name. :smirk:  Both approaches accomplish the same thing.  I'll add `close()` if you're good with it.

specifications/unified-provider.org
sergiusens: I would prefer that the "capabilties" be described by the provider and the cli to raise an exception

specifications/unified-provider.org
```diff
@@ -0,0 +1,215 @@
...
+      - configured
+      - running
+      - compatible
+
+      Used as a pre-check prior to initiating build. If False, indicates that
+      additional installation, or configuration must be performed by user, which
+      can be done with setup_provider().
+
+      :returns: True if ready, False if further setup is required.
+
```
sergiusens, line 79:
```suggestion
```

specifications/unified-provider.org
sergiusens: Why not just "stop"? Opposite to "start"

todo.org
```diff
@@ -78,3 +78,24 @@ CLOSED: [2020-06-24]
 - [ ] Error handling for setting invalid track names ([[https://bugs.launchpad.net/snapcraft/+bug/1892553][LP: 1892553]])
 - [X] Implement =snapcraft list-tracks=
 - [ ] Add documentation for track management to the [[https://snapcraft.io/docs/release-management][Release management]] docs.
+
+*** STRT [[file:specifications/unified-provider.org] Unified Provider]] [2/4]
```
sergiusens, line 82:
```suggestion
*** TODO [[file:specifications/unified-provider.org] Unified Provider]] [0/14]
```

sergiusens (changes requested):
Almost there

specifications/unified-provider.org
```diff
@@ -0,0 +1,214 @@
...
+  def is_provider_ready(self) -> bool:
+    """Check if provider is accessible.
+
+      As applicable, validate that the provider is:
+      - installed
+      - configured
+      - running
+      - compatible
+
+      Used as a pre-check prior to initiating build. If False, indicates that
```
cjp256, line 74:
```suggestion
      Used as a pre-check prior to use. If False, indicates that
```

specifications/unified-provider.org
cjp256: Need to add echoer

specifications/unified-provider.org
```diff
@@ -0,0 +1,214 @@
...
+      for the given provider (i.e. bind mounting directory from container).
+
+      :param prime_dir: Directory to mount/sync prime directory to on host.
+
+      :raises SnapcraftProviderUnsupportedMethod: if provider does not support
+          syncing/mounting project's prime directory.
+      """
+
+    def snap(self, output_dir: pathlib.Path) -> List[pathlib.Path]:
+      """Snap project, building as required.
```
cjp256, line 115:
```suggestion
      """Snap project, executing lifecycle steps as required.
```

specifications/unified-provider.org
cjp256: `start()` is not an interface..?

specifications/unified-provider.org
cjp256: :+1:

specifications/unified-provider.org
cjp256: Will come back with draft example

sergiusens:
This is going to be split out into its own module with a newer architecture, so all design work should go there. A new spec in the future should discuss how to integrate that new package into this code base.

PR #3297, opened 2020-09-25
cli: support snap --output <directory>

cjp256:
Currently one may only specify an explicit path to output snap
in the case of destructive mode.  For LXD/Multipass, one may only
specify the name of the snap within the project directory, otherwise
it gets "lost" in the container and is not retrieved.

This change allows output to point to an output directory, rather
than a snap file name.  This will help providers that may ship
multiple snaps, or users that want the snapcraft-provided name,
but rather have it shipped into a different directory.

NOTE that this change does not fix the still-broken case where
LXD/Multipass users specify a path outside of the project. This
should be addressed in upcoming work.

- Update _pack() to support the three uses of 'output'
- Introduce test coverage for _pack()

-----

### DIFF
--- a/snapcraft/cli/lifecycle.py
+++ b/snapcraft/cli/lifecycle.py
@@ -231,7 +231,7 @@ def _pack(
             )
         command.extend(["--compression", compression])
 
-    if output_file:
+    if output_file is not None:
         command.extend(["--filename", output_file])
 
     command.append(directory)
--- a/snapcraft/cli/lifecycle.py
+++ b/snapcraft/cli/lifecycle.py
@@ -202,6 +202,23 @@ def _run_pack(snap_command: List[Union[str, pathlib.Path]]) -> str:
 def _pack(
     directory: str, *, compression: Optional[str] = None, output: Optional[str]
 ) -> None:
+    output_file = None
+    output_dir = None
+
+    # Output may be:
+    # (1) a directory path to output snaps to
+    # (2) an explicit file path to output snap to
+    # (3) unspecified (None), output to current directory (project directory)
+    if output:
+        output_path = pathlib.Path(output)
+        output_parent = output_path.parent
+        if output_path.is_dir():
+            output_dir = str(output_path)
+        elif output_parent and output_parent != pathlib.Path("."):
+            output_dir = str(output_parent)
+            output_file = output_path.name
+        else:
+            output_file = output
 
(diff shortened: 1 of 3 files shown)


cjp256:
I should add that this helps in the case where the provider does not have a mounted project directory, so the provider can rely on _pack() to do the right thing and ship into into an output directory which will be fetched for the user.

snapcraft/cli/lifecycle.py
```diff
@@ -202,6 +202,23 @@ def _run_pack(snap_command: List[Union[str, pathlib.Path]]) -> str:
 def _pack(
     directory: str, *, compression: Optional[str] = None, output: Optional[str]
 ) -> None:
+    output_file = None
+    output_dir = None
+
+    # Output may be:
```
sergiusens, line 208: This should be in the doctstring for the method instead of a code comment

snapcraft/cli/lifecycle.py
```diff
@@ -213,10 +230,16 @@ def _pack(
                 f"EXPERIMENTAL: Setting the squash FS compression to {compression!r}."
             )
         command.extend(["--compression", compression])
-    if output is not None:
-        command.extend(["--filename", output])
+
+    if output_file:
```
sergiusens, line 234:
```suggestion
    if output_file is not None:
```

snapcraft/cli/lifecycle.py
```diff
@@ -213,10 +230,16 @@ def _pack(
...
+        command.extend(["--filename", output_file])
+
     command.append(directory)
 
+    if output_dir:
```
sergiusens, line 239:
```suggestion
    if output_dir is not None:
```

sergiusens (changes requested):
Looks good, just a few things to consider about moving things to a docstring instead (and creating one in the process)

snapcraft/cli/lifecycle.py
cjp256: Done!

snapcraft/cli/lifecycle.py
```diff
@@ -202,6 +202,28 @@ def _run_pack(snap_command: List[Union[str, pathlib.Path]]) -> str:
 def _pack(
     directory: str, *, compression: Optional[str] = None, output: Optional[str]
 ) -> None:
+    """Pack a snap.
+
```
sergiusens, line 206: you have whitespace here
```suggestion

```

sergiusens (approved):

PR #3302, opened 2020-09-29
build providers: configure apt according to runtime environment arch

cjp256:
Query the architecture from the running instance using the "arch"
command instead of the host's (using platform.machine()).

The instance's architecture is the relevant one when confinguring
the apt source URLs.

While here:

- Move apt configuration from _setup_environment(),
  to new _setup_environment_apt().

- Remove broken security mirror override (it overwote with
  primary configuration).  As it's broken and we don't intend
  to support an override for security configuration at this
  time, just remove it.

-----


snapcraft/internal/build_providers/_base_provider.py
```diff
@@ -357,27 +359,29 @@ def _get_code_name_from_build_base(self):
             "core20": "focal",
         }[build_base]
 
-    def _get_primary_mirror(self) -> str:
+    def _get_environment_arch(self) -> str:
+        arch = self._run(["arch"], hide_output=True)
```
sergiusens, line 363: won't this print the machine hardware? what we want is the userspace architecture though, right?

sergiusens (changes requested):

snapcraft/internal/build_providers/_base_provider.py
cjp256: indeed. will revisit.

since I can't rely on `dpkg-architecture` do you have a better suggestion?

cjp256: Though in this case, it's really just used to check for x86 or not, so should be safe.  At least as safe as platform.machine().

sergiusens: we have something for user space architecture checking in project_options. I would add a code comment to get the intent as _get_environment_arch is not returning an accurate value if this caveat is taken into account

cjp256:
Closing in favor of upcoming craft-providers switchover.

PR #3309, opened 2020-10-08
Rough draft of changes needed to separate python versions between gno…

hellsworth:
…me-3-28 and gnome-3-34 extensions.

gnome-3-28 needs a PYTHONPATH that points to $SNAP_DESKTOP_RUNTIME/usr/lib/python3/site-packages while gnome-3-34 needs $SNAP_DESKTOP_RUNTIME/usr/lib/python3.6/site-packages. For this reason, we shouldn't just assume python3.6 by having this env var set in launcher-specific and rather set some condition that checks the extension used and then sets the PYTHONPATH var.

-----

### DIFF
--- a/extensions/desktop/gnome/Makefile
+++ b/extensions/desktop/gnome/Makefile
@@ -9,7 +9,6 @@ LIB_DIR              := $(DESTDIR)/lib
 DEST_LAUNCHER        := desktop-launch
 BINDTEXTDOMAIN       := bindtextdomain.so
 DEST_CONFIGURE_HOOK  := hooks-configure-desktop
-DEST_PYTHON          := dest-python
 HW_PLATFORM          := $(shell uname --hardware-platform)
 ARCH                 := $(HW_PLATFORM)-linux-gnu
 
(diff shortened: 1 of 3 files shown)


sergiusens:
As a counter suggestion, this might be easier to maintain. These scripts are all "assembled" through a makefile target, instead of overcomplicating this I would use build-environment for the part in the extension that creates this launcher to export the useful information to set (or use a make variable). The makefile would then insert the correct snippet into the main launcher by adapting a template script that would be sed with the right parameters (instead of cat as done for the other bits) to get the correct path in place.

hellsworth:
> As a counter suggestion, this might be easier to maintain. These scripts are all "assembled" through a makefile target, instead of overcomplicating this I would use build-environment for the part in the extension that creates this launcher to export the useful information to set (or use a make variable).

Ok I can set the PYTHONPATH in the build-environment sections of gnome_3_28.py and gnome_3_34.py, but then the problem is that we're just overwriting any PYTHONPATH the user has set, right? This was the reason we went down the path of even adding PYTHONPATH to launcher-specific to begin with.

>The makefile would then insert the correct snippet into the main launcher by adapting a template script that would be sed with the right parameters (instead of cat as done for the other bits) to get the correct path in place.
Merge state

Where is this template script?

sergiusens:
https://github.com/hellsworth/snapcraft/commit/7a88304df380da1172f5b4469aa1abbcc14fcce1#diff-f9d3477e1f35aee2c2136c3645ba9518R28

Before or after (depending on requirements)
```
@tail -n +2 $(SRC_DIR)/launcher-specific >> $(DEST_LAUNCHER)
```
So if you created python-path.tmpl and templated `{{PYTHON_PATH}}`, something like
```
@sed $(SRC_DIR)/python-path -e "s/{{PYTHON_PATH}}/$(PYTHON_PATH)/"  >> $(DEST_LAUNCHER)
```

hellsworth:
> [hellsworth@7a88304#diff-f9d3477e1f35aee2c2136c3645ba9518R28](https://github.com/hellsworth/snapcraft/commit/7a88304df380da1172f5b4469aa1abbcc14fcce1#diff-f9d3477e1f35aee2c2136c3645ba9518R28)
> 
> Before or after (depending on requirements)
> 
> ```
> @tail -n +2 $(SRC_DIR)/launcher-specific >> $(DEST_LAUNCHER)
> ```

This line sets the contents of $(DEST_LAUNCHER), and now we want to modify that with sed.

> 
> So if you created python-path.tmpl and templated `{{PYTHON_PATH}}`, something like
> 
> ```
> @sed $(SRC_DIR)/python-path -e "s/{{PYTHON_PATH}}/$(PYTHON_PATH)/"  >> $(DEST_LAUNCHER)
> ```

Is there an example of another templated file somewhere? It would help me figure out where this file should go. It might also help me properly define `{{PYTHON_PATH}}` :)

sergiusens:
I am closing this as the diff shows as empty, please reopen when there is something to review again

PR #3317, opened 2020-10-14
plugin handler: properly handle snapcraftctl errors

cjp256:
Address issues including:

- snapcraftctl operations always exiting with 0, even for failure
  cases.

- snapcraft override scripts continue to run even after
  parent snapcraft process fails.  This is because `set -e` does
  not catch the error when snapcraftctl returns 0, and nothing
  is otherwise done to terminate the script.

Given the nature of these symptoms, this is likely the cause of
a number of other seemingly unrelated issues.

These issues occur because _handle_builtin_function() only
catches exceptions of type errors.ScriptletBaseError,
ignoring other cases that may result from the build/pull operations,
such as NodejsPluginMissingPackageJsonError, or
SnapcraftPluginBuildError.

Instead of trying to manage exceptions in _handle_builtin_function(),
catch _any_ unhandled exception in _run_scriplet()'s try block. This
simplifies the more complex return value and feedback handling.

Don't print an error from the feedback channel in snapcraftctl,
the parent process will take care of it.  This is necessary
because the error that is now seen by snapcraftctl would lead
to redundant output.  Just exit with -1 when an error is
read from the feedback fifo.

LP: #1898922

-----

### DIFF
--- a/tests/spread/general/scriptlets/task.yaml
+++ b/tests/spread/general/scriptlets/task.yaml
@@ -9,20 +9,16 @@ prepare: |
   set_base "$SNAP_DIR/snap/snapcraft.yaml"
 
 restore: |
-  #shellcheck source=tests/spread/tools/snapcraft-yaml.sh
-  . "$TOOLS_DIR/snapcraft-yaml.sh"
   cd "$SNAP_DIR"
-
   snapcraft clean
   rm -f ./*.snap
 
-  restore_yaml "snap/snapcraft.yaml"
-
-execute: |
   #shellcheck source=tests/spread/tools/snapcraft-yaml.sh
   . "$TOOLS_DIR/snapcraft-yaml.sh"
-  cd "$SNAP_DIR"
+  restore_yaml "snap/snapcraft.yaml"
 
+execute: |x
+  cd "$SNAP_DIR"
   snapcraft_log="$(snapcraft build 2>&1 || true)"
 
   echo "${snapcraft_log}" | NOMATCH "^should have failed set-version"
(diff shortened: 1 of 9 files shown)


sergiusens:
Can you add a spread test to the general suite here?

tests/spread/general/scriptlets/task.yaml
```diff
@@ -0,0 +1,30 @@
...
+  rm -f ./*.snap
+
+  restore_yaml "snap/snapcraft.yaml"
+
+execute: |
+  #shellcheck source=tests/spread/tools/snapcraft-yaml.sh
+  . "$TOOLS_DIR/snapcraft-yaml.sh"
+  cd "$SNAP_DIR"
+
+
```
sergiusens, line 26: random question about the amount of empty lines and the thought process put into it

cjp256: There does seem to be an extra line there for no reason. Just an accident.
```suggestion
```

tests/spread/general/scriptlets/task.yaml
```diff
@@ -9,20 +9,16 @@ prepare: |
...
 
-  restore_yaml "snap/snapcraft.yaml"
-
-execute: |
   #shellcheck source=tests/spread/tools/snapcraft-yaml.sh
   . "$TOOLS_DIR/snapcraft-yaml.sh"
-  cd "$SNAP_DIR"
+  restore_yaml "snap/snapcraft.yaml"
 
+execute: |x
```
sergiusens, line 20: there's an x here now :-)

sergiusens, line 20:
```suggestion
execute: |
```

sergiusens (approved):

PR #3319, opened 2020-10-14
[RFC] plugin handler: use bash with additional error checking for core20 scripts

cjp256:
Bash allows us to `set -o pipefail` to catch pipe failures, e.g.
`fail | true` that dash will ignore.

In addition to switching to bash:

- Set '-o pipefail' to catch pipe errors.

- Set '-u' to catch unbound variable errors.

- Set '-x' to improve visibility of scriptlets.

To address the subsequent breakage:

- Only apply additional shell flags after the environmnet.
  Otherwise user build-environment may be unbound when using
  variables not applicable to the step (e.g. SNAPCRAFT_PART_INSTALL).

- Update v2 catkin/colcon to drop -u flag as there are plenty
  of unbound variables present in their scripts.

- Only apply these changes to core20, keeping legacy behavior
  the same for core/core18.

If we don't go this route, we'll have to consider rewriting the core20 plugin scripts to avoid the use of pipes, or find some other alternative for pipefail.

With '-x' included, this PR (currently) would satisfy https://github.com/snapcore/snapcraft/pull/3318.


cjp256:
Well that's interesting.  In the case of the failing python spread test, the `build-environment` has
```
    build-environment:
      - PYTHONPATH: $SNAPCRAFT_PART_INSTALL/usr/lib/python3/dist-packages:$PYTHONPATH
```

During phases other than build, in this case pull, it's invalid.

sergiusens (approved):

sergiusens (approved):

PR #3321, opened 2020-10-19
meta: add error check for "command not found"

cjp256:
When not allowing massaged comamnds (e.g. core20), app commands
not found were triggering InvalidAppCommandNotExecutable() errors.

- Add check for presence of app command, throwing
  InvalidAppCommandNotFound.

- Remove errors.PrimedCommandNotFoundError, replacing
  usage with InvalidAppCommandNotFound for consistency.

- Add unit tests for both of these cases.

-----


sergiusens (commented):
Looks good, but why don't you just raise where appropriate instead of bubbling up a None check? https://docs.python.org/3.6/glossary.html#term-eafp

cjp256:
Because the exception requires `app_name`. The choice was to add it to the resolver (which couples in a way that makes me wonder why it's split from Command) or just return None, and leaving the error handling to the caller.   Pretty much a coin toss.

sergiusens:
Chris Patterson writes:

> Because the exception requires `app_name`. The choice was to add it to
> the resolver (which couples in a way that makes me wonder why it's
> split from Command) or just return None, and leaving the error
> handling to the caller.   Pretty much a coin toss.

You can always raise and catch where this context exists, right?

sergiusens (approved):

PR #3324, opened 2020-10-20
set ROS_PYTHON_VERSION for rosdep

artivis:
Sets `ROS_PYTHON_VERSION = 2` for the env used by `rosdep` to fix an harmless but annoying warning:
```terminal
WARNING: ROS_PYTHON_VERSION is unset. Defaulting to 2
```

Signed-off-by: artivis <deray.jeremie@gmail.com>

-----

### DIFF
--- a/snapcraft/plugins/v1/_ros/rosdep.py
+++ b/snapcraft/plugins/v1/_ros/rosdep.py
@@ -218,6 +218,8 @@ class Rosdep:
             self._rosdep_install_path, "usr", "lib", "python2.7", "dist-packages"
         )
 
+        env["ROS_PYTHON_VERSION"] = "2"
+
         # By default, rosdep uses /etc/ros/rosdep to hold its sources list. We
         # don't want that here since we don't want to touch the host machine
         # (not to mention it would require sudo), so we can redirect it via


artivis:
@kyrofa quick review for a small PR?

sergiusens (approved):
Looks good, will wait for @kyrofa or someone else in the robotics team to ack

snapcraft/plugins/v1/_ros/rosdep.py
```diff
@@ -218,6 +218,8 @@ def _run(self, arguments):
             self._rosdep_install_path, "usr", "lib", "python2.7", "dist-packages"
         )
 
+        env["ROS_PYTHON_VERSION"] = "2"
```
kyrofa, line 221: Given that this class is used by both catkin and colcon plugins, it makes me a little nervous to hard-code this here. Are we certain it won't have bad ramifications for colcon, which is python 3?

artivis: This var is only set for the env in which `rosdep` is run. Note that rosdep sets it to `2` by default. This does not leak to a colcon env and afaik colcon does not uses this env, so I believe we're good. Furthermore `PYTHONPATH` is hardcoded to a Python 2 path literally 3 lines above.

kyrofa: Yeah fair point, okay.

kyrofa (approved):

PR #3325, opened 2020-10-20
snapcraftctl: add checks for empty string for set-version & set-grade

cjp256:
It's a pretty common mistake for scriplets to end up setting "" in
set-version, leading the the following error:
`Failed to generate snap metadata: 'adopt-info' refers to part <part>,
but that part is lacking the 'parse-info' property.`

Setting an empty grade effectively just ignores the invalid value,
relying on the default grade later.

Explicitly error out with helpful messages when invalid options
are used.

-----

### DIFF
--- a/snapcraft/cli/snapcraftctl/_runner.py
+++ b/snapcraft/cli/snapcraftctl/_runner.py
@@ -67,15 +67,27 @@ def prime():
     _call_function("prime")
 
 
+def validate_version(ctx, param, value):
+    if not value:
+        raise click.BadParameter(f"Version must be a non-empty string.")
+    return value
+
+
 @run.command("set-version")
-@click.argument("version")
+@click.argument("version", callback=validate_version)
 def set_version(version):
     """Set the version of the snap"""
     _call_function("set-version", {"version": version})
 
 
+def validate_grade(ctx, param, value):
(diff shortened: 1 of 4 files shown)


snapcraft/cli/snapcraftctl/_runner.py
```diff
@@ -72,13 +71,25 @@ def prime():
 @click.argument("version")
 def set_version(version):
     """Set the version of the snap"""
+
+    if not version:
```
sergiusens, line 75: How about using click's native features for this https://click.palletsprojects.com/en/7.x/options/#callbacks-for-validation ?

snapcraft/cli/snapcraftctl/_runner.py
```diff
@@ -72,13 +71,25 @@ def prime():
...
+        raise errors.SnapcraftctlError(
```
sergiusens, line 76: I would raise a click.BadParameter error as this is strictly about the cli

snapcraft/cli/snapcraftctl/_runner.py
cjp256: Certainly can.  and makes sense here, will do.

sergiusens (approved):

PR #3327, opened 2020-10-20
Fix use case when multiple packages are compiled with catkin

facontidavide:
This solves a very specific problem with catkin:

If the file **rosinstall** pull 2 packages from 2 different repositories (A and B) and A depend on B, A will not compile unless the option `--merge` is used.

This change should not affect the "normal" use cases.

- [X ] Have you followed the [guidelines for contributing](https://github.com/snapcore/snapcraft/blob/master/CONTRIBUTING.md)?
- [X ] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?

-----


kyrofa (commented):
I don't believe this should be necessary. If memory serves, the actual issue you were trying to work around here ended up being due to missing CMake install rules, do you agree @facontidavide?

facontidavide:
yes, eventually I realized you are right!

PR #3336, opened 2020-10-24
package repositories: introduce path property for exact paths

cjp256:
Paths pecifies an exact path relative to `url`.

In sources.list(5) manpage, an "exact path" is a suite that ends
with a "/", and cannot be combined with components.

Snapcraft will make this translation for apt, but exposes these
two distinct usages differently in snapcraft.yaml configuration.

- Add path property, enforcing dependencies & exclusivity in
  validate().  Make components and suites (and path) optional
  in the schema, but validate the various cases when instantiating.

- Refactor the unit tests to migrate to the current style & do
  a little organization to make them easier to follow.

- Add validate() for PackageRepositoryPpa as done previously with
  PackageRepositoryApt.

- Add apt_repo parameter for repo dependency injection. This will
  simplify test mocking while maintaining current behavior by
  specifying a default value of repo.Ubuntu.

- Add unit tests to cover new functionality.

- Minor specification tweaks/fixes:

  - Don't require the '/' if a path is specified, Snapcraft will
    just append it if needed).

  - Fix statement about path starting with "/", it should be
    "ending".

  - Replace usage of "absolute" path with "exact".

-----

### DIFF
--- a/snapcraft/internal/meta/package_repository.py
+++ b/snapcraft/internal/meta/package_repository.py
@@ -21,7 +21,7 @@ from copy import deepcopy
 from pathlib import Path
 from typing import Any, Dict, List, Optional
 
-from snapcraft.internal.repo._deb import Ubuntu
+from snapcraft.internal import repo
 
 from . import errors
 
@@ -80,7 +80,7 @@ class PackageRepositoryAptPpa(PackageRepository):
         self.validate()
 
     def install(self, *, keys_path: Path) -> bool:
-        return Ubuntu.install_ppa(keys_path=keys_path, ppa=self.ppa)
+        return repo.Ubuntu.install_ppa(keys_path=keys_path, ppa=self.ppa)
 
     def marshal(self) -> Dict[str, Any]:
         data = dict(type="apt")
@@ -202,12 +202,12 @@ class PackageRepositoryApt(PackageRepository):
             raise RuntimeError("no suites or path")
 
         # First install associated GPG key.
-        new_key: bool = Ubuntu.install_gpg_key_id(
+        new_key: bool = repo.Ubuntu.install_gpg_key_id(
             keys_path=keys_path, key_id=self.key_id, key_server=self.key_server
         )
 
         # Now install sources file.
-        new_sources: bool = Ubuntu.install_sources(
+        new_sources: bool = repo.Ubuntu.install_sources(
             architectures=self.architectures,
             components=self.components,
             formats=self.formats,
@@ -248,23 +248,6 @@ class PackageRepositoryApt(PackageRepository):
         return data
 
     def validate(self) -> None:  # noqa: C901
-        if self.architectures is not None:
-            for arch in self.architectures:
-                if arch not in [
-                    "amd64",
-                    "arm64",
-                    "armhf",
-                    "i386",
-                    "ppc64el",
-                    "s390x",
-                ]:
-                    raise errors.PackageRepositoryValidationError(
-                        url=self.url,
-                        brief=f"Invalid architecture {arch!r}.",
-                        details="Valid architectures include: amd64, armhf, arm64, i386, ppc64el, and s390x.",
-                        resolution="Verify repository configuration and ensure that 'architectures' is correctly specified.",
-                    )
-
         if self.formats is not None:
             for repo_format in self.formats:
                 if repo_format not in ["deb", "deb-src"]:
--- a/snapcraft/internal/meta/package_repository.py
+++ b/snapcraft/internal/meta/package_repository.py
@@ -82,14 +82,18 @@ class PackageRepositoryAptPpa(PackageRepository):
             raise errors.PackageRepositoryValidationError(
                 url=self.ppa,
                 brief=f"Invalid PPA {self.ppa!r}.",
-                resolution="Update 'ppa' to a non-empty string.",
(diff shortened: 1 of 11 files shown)


snapcraft/internal/meta/package_repository.py
```diff
@@ -141,24 +152,40 @@ def __init__(
...
         self.validate()
 
     def install(self, keys_path: Path) -> bool:
+        if not self.path and not self.components and not self.suites:
+            suites = ["/"]
+        elif self.path:
+            # Suites denoting exact path must end with '/'.
+            path = self.path
+            if not path.endswith("/"):
+                path = path + "/"
```
sergiusens, line 170: FWIW, this is also valid
```suggestion
                path += "/"
```

snapcraft/internal/meta/package_repository.py
```diff
@@ -120,13 +129,15 @@ def __init__(
...
+        components: Optional[List[str]] = None,
         formats: Optional[List[str]] = None,
         key_id: str,
         key_server: Optional[str] = None,
         name: Optional[str] = None,
-        suites: List[str],
+        path: Optional[str] = None,
+        suites: Optional[List[str]] = None,
         url: str,
+        apt_repo: Type[Ubuntu] = Ubuntu,
```
sergiusens, line 140: Is there a reason for this?

snapcraft/internal/meta/package_repository.py
```diff
@@ -141,24 +152,40 @@ def __init__(
...
+        self.path = path
         self.suites = suites
         self.url = url
 
+        self._apt_repo = apt_repo
+
         self.validate()
 
     def install(self, keys_path: Path) -> bool:
+        if not self.path and not self.components and not self.suites:
```
sergiusens, line 164: Can you add a code comment for this?

snapcraft/internal/meta/package_repository.py
```diff
@@ -141,24 +152,40 @@ def __init__(
...
+        elif self.path:
+            # Suites denoting exact path must end with '/'.
+            path = self.path
+            if not path.endswith("/"):
+                path = path + "/"
+            suites = [path]
+        elif self.suites:
+            suites = self.suites
+        else:
+            raise RuntimeError("no suites or path")
```
sergiusens, line 175: This would only happen if validation fails, correct?

sergiusens (changes requested):
Pretty good, just some minor questions and code comment sprinkling

snapcraft/internal/meta/package_repository.py
cjp256: Done.

snapcraft/internal/meta/package_repository.py
cjp256: There was for dependency injection, but I have since refactored install bits out of meta, so this will be dropped when I push.

snapcraft/internal/meta/package_repository.py
cjp256: Done.

snapcraft/internal/meta/package_repository.py
cjp256: Yes, this should never happen.  I added a note (and another sanity check).

cjp256:
I'll come back to this on Tuesday

cjp256:
Should be good to go. I rewrote most of the error messages as well as improved some of the checks to make the outputs more precise.

And added validation/tests for architecture values.

snapcraft/internal/meta/package_repository.py
```diff
@@ -181,40 +236,119 @@ def marshal(self) -> Dict[str, Any]:
...
-        if self.formats is not None and any(
-            f not in ["deb", "deb-src"] for f in self.formats
-        ):
-            raise errors.PackageRepositoryValidationError(
-                url=self.url,
-                brief=f"Invalid formats {self.formats!r}.",
-                resolution="You can specify a list of formats including 'deb' and/or 'deb-src'.",
-            )
+    def validate(self) -> None:  # noqa: C901
+        if self.architectures is not None:
```
sergiusens, line 251: valid architectures are dependant on base, core20 and i386 are invalid as is core18 or less with riscv

sergiusens: This might be a fragile check, might be worth just letting apt fail and surface an error

sergiusens (changes requested):
One last thing, looking good!

snapcraft/internal/meta/package_repository.py
cjp256: Good point.   It was mostly meant as a sanity check, riscv enablement would require it to be added here.

Longer term I'd hope for a shared deb arch enum that we could check against.

I can remove it if you want.

sergiusens: All our enums are in ProjectOptions today, maybe add it there?

sergiusens: Then again, it is not uncommon for an arch to be enabled post release, so this would require doing post facto work, removing might just be the better option if we can surface the error correctly

cjp256: Removed.

sergiusens (approved):

cjp256:
Just the plainbox failures.

PR #3337, opened 2020-10-26
cli,plugins: assume that core22 base wants v2 plugins.

xnox:
When trying to use build-base core22, some portions of cli/plugins
defaulted to legacy/v1 plugins, instead of v2.

Use a list of old cores, and assume anything new to be modern/v2.

Add a test for a "future core2y" to use v2 plugins.

   - no, fails on groovy with new pyflakes?
   - no, fails on groovy, as it's trying to run pytest (python2) not pytest-3 (python3)?

### DIFF
--- a/snapcraft/plugins/_plugin_finder.py
+++ b/snapcraft/plugins/_plugin_finder.py
@@ -93,7 +93,7 @@ def get_plugin_for_base(plugin_name: str, *, build_base: str) -> PluginTypes:
     if build_base in ( "core", "core16", "core18"):
         plugin_version = "v1"
     elif build_base == "":
-        # This should never happen if 
+        # This should never happen when using build_base from Project.
         raise RuntimeError("'build_base' cannot be unset.")
         plugin_version = "v1"
 
--- a/snapcraft/plugins/_plugin_finder.py
+++ b/snapcraft/plugins/_plugin_finder.py
@@ -26,7 +26,7 @@ PluginTypes = Union[Type[v1.PluginV1], Type[v2.PluginV2]]
 # TODO: segregate into base specific plugins.
 if sys.platform == "linux" or TYPE_CHECKING:
     _PLUGINS: Dict[str, Dict[str, PluginTypes]] = {
-        "legacy": {
+        "v1": {
             "ant": v1.AntPlugin,
             "autotools": v1.AutotoolsPlugin,
             "catkin": v1.CatkinPlugin,
@@ -56,7 +56,7 @@ if sys.platform == "linux" or TYPE_CHECKING:
             "scons": v1.SconsPlugin,
             "waf": v1.WafPlugin,
         },
-        "core20": {
+        "v2": {
             "autotools": v2.AutotoolsPlugin,
             "catkin": v2.CatkinPlugin,
             "catkin-tools": v2.CatkinToolsPlugin,
@@ -74,7 +74,7 @@ if sys.platform == "linux" or TYPE_CHECKING:
     }
 else:
     # We cannot import the plugins on anything but linux.
-    _PLUGINS = {"legacy": {}}
+    _PLUGINS = {"v1": {}, "v2": {}}
 
 
 def get_plugin_for_base(plugin_name: str, *, build_base: str) -> PluginTypes:
@@ -86,12 +86,15 @@ def get_plugin_for_base(plugin_name: str, *, build_base: str) -> PluginTypes:
     """
     # TODO: segregate this more and remove the checks UnsupportedBase checks
     # from the plugins
-    if build_base in ("core", "core16", "core18"):
-        build_base = "legacy"
-    if build_base not in ("legacy", "core20"):
-        build_base = "core20"
+
+    # Default for unknown, and core20
+    plugin_version = "v2"
+    # Others use v1
+    if build_base in ("", "legacy", "core", "core16", "core18"):
+        plugin_version = "v1"
+
     try:
-        plugin_classes = _PLUGINS[build_base]
+        plugin_classes = _PLUGINS[plugin_version]
         return plugin_classes[plugin_name]
     except KeyError:
         raise errors.PluginError(f"unknown plugin: {plugin_name!r}")
--- a/snapcraft/plugins/_plugin_finder.py
+++ b/snapcraft/plugins/_plugin_finder.py
@@ -88,6 +88,8 @@ def get_plugin_for_base(plugin_name: str, *, build_base: str) -> PluginTypes:
     # from the plugins
     if build_base in ("core", "core16", "core18"):
         build_base = "legacy"
+    if build_base not in ("legacy", "core20"):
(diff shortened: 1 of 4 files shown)


snapcraft/plugins/_plugin_finder.py
```diff
@@ -88,6 +88,8 @@ def get_plugin_for_base(plugin_name: str, *, build_base: str) -> PluginTypes:
     # from the plugins
     if build_base in ("core", "core16", "core18"):
         build_base = "legacy"
+    if build_base not in ("legacy", "core20"):
```
cjp256, line 91: I think it would be appropriate to force the user to specify `build-base` in a case like this.  I don't think Snapcraft should assume core20 if left unspecified.

The schema or sanity checks should require `build-base` if `base` is not something we have build image support for.

xnox: `build-base` is always specified at this point.

The intention of this hunk of code is that, if base is not known use the highest known set of plugins, currently the core20 ones.

As otherwise, currently with `build-base: core22` the dictionary has no plugins for `core22`, and hence the build fails with "plugin nil not found". 

Or do you want me to start core22 dictinary key which is a copy of core20?

xnox: Note this is not really a `build-base`, but rather keys of `_PLUGINS` dictionary. Of which there are only two at the moment `legacy` and `core20`. Which doesn't quite make sense to me.

Maybe I should rework this a bit? to rename keys in the `_PLUGINS` dictionary to be `v1` and `v2`, and then just use modules_path at this point? rather than remap, for the second time, which build-base is using what?

xnox: See https://github.com/snapcore/snapcraft/pull/3338/files if that makes more sense..... Given that we have many-to-one mapping between the set of bases and plugin versions. It doesn't make sense to name the keys in the `_PLUGINS` dictinary after bases.

Or at least, it should become default-dict type, to return `core20` / `v2` ones for unknown bases?

cjp256: Ah yes, that's what I get for commenting before having my coffee :D I agree,  using keys "v1" and "v2" would be more ideal at this point rather than legacy/core20.

xnox: @cjp256 please re-review the renames of things. And all the tests now fixed up and pass.

snapcraft/plugins/_plugin_finder.py
```diff
@@ -86,10 +86,15 @@ def get_plugin_for_base(plugin_name: str, *, build_base: str) -> PluginTypes:
     """
     # TODO: segregate this more and remove the checks UnsupportedBase checks
     # from the plugins
-    if build_base in ("core", "core16", "core18"):
-        build_base = "legacy"
+
+    # Default for unknown, and core20
+    plugin_version = "v2"
+    # Others use v1
+    if build_base in ("", "legacy", "core", "core16", "core18"):
```
sergiusens, line 93: Seems that legacy faced a copy / paste error
```suggestion
    if build_base in ( "core", "core16", "core18"):
        plugin_version = "v1"
    elif build_base == "":
        # This should never happen if 
        raise RuntimeError("'build_base' cannot be unset.")
```

The last check is a precaution, in case someone skips this other precaution in `snapcraft.internal.meta.snap.Snap.get_build_base` which is only triggered if someone skips schema checks during development:
```python
# The schema does not allow for this when loaded from snapcraft.yaml.
        if build_base is None:
            raise RuntimeError("'build_base' cannot be None")
```

tests/unit/commands/test_list_plugins.py
```diff
@@ -117,6 +117,14 @@ def test_core20_list(self):
             snapcraft.plugins.v2.__path__
         )
 
+    def test_core2y_list(self):
+        # Note that core2y is some future core, _not_ allowed to be used from cmdline
```
sergiusens, line 121:
```suggestion
        # Note that core2y is some future base, _not_ allowed to be used from cmdline
```

tests/unit/pluginhandler/test_plugin_loader.py
```diff
@@ -61,18 +61,18 @@ class Local(snapcraft.plugins.v1.PluginV1):
 
 
 class InTreePluginsTest(unit.TestCase):
-    def test_all_known_legacy(self):
+    def test_all_known_core18_v1(self):
```
sergiusens, line 64: More of a nit
```suggestion
    def test_all_known_v1(self):
```

tests/unit/pluginhandler/test_plugin_loader.py
```diff
@@ -61,18 +61,18 @@ class Local(snapcraft.plugins.v1.PluginV1):
...
         self.useFixture(fixtures.MockPatch("jsonschema.validate"))
-        for plugin_name in _PLUGINS["legacy"]:
+        for plugin_name in _PLUGINS["v1"]:
             plugin_handler = self.load_part(
                 "test-part", plugin_name=plugin_name, base="core18"
             )
             self.expectThat(plugin_handler.plugin, IsInstance(PluginV1))
 
-    def test_all_core20(self):
+    def test_all_core20_v2(self):
```
sergiusens, line 73: More of a nit
```suggestion
    def test_all_v2(self):
```

sergiusens (changes requested):
Looks pretty good. Thanks for improving Snapcraft. I have some minor suggestions for corrections and some ignorable nits

snapcraft/plugins/_plugin_finder.py
```diff
@@ -86,10 +86,19 @@ def get_plugin_for_base(plugin_name: str, *, build_base: str) -> PluginTypes:
...
-    if build_base in ("core", "core16", "core18"):
-        build_base = "legacy"
+
+    # Default for unknown, and core20
+    plugin_version = "v2"
+    # Others use v1
+    if build_base in ( "core", "core16", "core18"):
+        plugin_version = "v1"
+    elif build_base == "":
+        # This should never happen if
```
sergiusens, line 96:
```suggestion
        # This should never happen when using build_base from Project.
```

snapcraft/plugins/_plugin_finder.py
```diff
@@ -86,10 +86,19 @@ def get_plugin_for_base(plugin_name: str, *, build_base: str) -> PluginTypes:
     """
     # TODO: segregate this more and remove the checks UnsupportedBase checks
     # from the plugins
-    if build_base in ("core", "core16", "core18"):
-        build_base = "legacy"
+
+    # Default for unknown, and core20
+    plugin_version = "v2"
+    # Others use v1
+    if build_base in ( "core", "core16", "core18"):
```
sergiusens, line 93:
```suggestion
    if build_base in ("core", "core16", "core18"):
```

sergiusens (approved):

PR #3347, opened 2020-11-03
Create an extra 'build' directory for plugins

artivis:
This PR introduces an extra 'build' directory `part-name/build_plugin` (and associated env var `SNAPCRAFT_PART_BUILD_PLUGIN`) for plugins that may need it.

This comes from the need to isolate the source files root directory (`part-name/build`) from the build artifacts for the ROS plugins.

Signed-off-by: artivis <deray.jeremie@gmail.com>

-----

### DIFF
--- a/snapcraft/internal/pluginhandler/__init__.py
+++ b/snapcraft/internal/pluginhandler/__init__.py
@@ -84,6 +84,7 @@ class PluginHandler:
         self.part_dir = os.path.join(project.parts_dir, self.name)
         self.part_source_dir = os.path.join(self.part_dir, "src")
         self.part_build_dir = os.path.join(self.part_dir, "build")
+        self.part_build_plugin_dir = os.path.join(self.part_dir, "build_plugin")
         self.part_install_dir = os.path.join(self.part_dir, "install")
         self.part_state_dir = os.path.join(self.part_dir, "state")
         self.part_snaps_dir = os.path.join(self.part_dir, "snaps")
@@ -258,6 +259,7 @@ class PluginHandler:
         dirs = [
             self.part_source_dir,
             self.part_build_dir,
+            self.part_build_plugin_dir,
             self.part_install_dir,
             self.part_state_dir,
             self._project.stage_dir,
@@ -582,6 +584,8 @@ class PluginHandler:
         ):
             if os.path.exists(self.part_build_dir):
                 shutil.rmtree(self.part_build_dir)
+            if os.path.exists(self.part_build_plugin_dir):
+                shutil.rmtree(self.part_build_plugin_dir)
 
             # No hard-links being used here in case the build process modifies
             # these files.
--- a/snapcraft/internal/pluginhandler/__init__.py
+++ b/snapcraft/internal/pluginhandler/__init__.py
@@ -801,6 +801,9 @@ class PluginHandler:
         if os.path.exists(self.part_build_dir):
             shutil.rmtree(self.part_build_dir)
 
+        if os.path.exists(self.part_build_plugin_dir):
+            shutil.rmtree(self.part_build_plugin_dir)
+
         if os.path.exists(self.part_install_dir):
             shutil.rmtree(self.part_install_dir)
 
--- a/snapcraft/internal/pluginhandler/_part_environment.py
+++ b/snapcraft/internal/pluginhandler/_part_environment.py
@@ -88,6 +88,7 @@ def get_snapcraft_part_directory_environment(
             {
                 "SNAPCRAFT_PART_BUILD": part.part_build_dir,
                 "SNAPCRAFT_PART_BUILD_WORK": part.part_build_work_dir,
+                "SNAPCRAFT_PART_BUILD_PLUGIN": part.part_build_plugin_dir,
                 "SNAPCRAFT_PART_INSTALL": part.part_install_dir,
             }
         )


snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -84,6 +84,7 @@ def __init__(
         self.part_dir = os.path.join(project.parts_dir, self.name)
         self.part_source_dir = os.path.join(self.part_dir, "src")
         self.part_build_dir = os.path.join(self.part_dir, "build")
+        self.part_build_plugin_dir = os.path.join(self.part_dir, "build_plugin")
```
kyrofa, line 87: This directory needs to be removed in the `clean_build()` method defined in this class.

snapcraft/internal/pluginhandler/_part_environment.py
```diff
@@ -88,6 +88,7 @@ def get_snapcraft_part_directory_environment(
             {
                 "SNAPCRAFT_PART_BUILD": part.part_build_dir,
                 "SNAPCRAFT_PART_BUILD_WORK": part.part_build_work_dir,
+                "SNAPCRAFT_PART_BUILD_PLUGIN": part.part_build_plugin_dir,
```
kyrofa, line 91: I'm iffy on this name, but I struggle to think of something better that still matches the established patterns. For example, I like "plugindir" better here, but that doesn't match "work". @sergiusens and @cjp256, got any better ideas?

snapcraft/internal/pluginhandler/__init__.py
artivis: Done in 24f8a494b7d9b3948c04a1684c1a67420715b242.

jhenstridge:
Is the aim here to perform an out of tree "$srcdir != $builddir" here?  And if so, what would a plugin using the `build_plugin` directory have in the `build` directory?

If the `build` directory would just be a copy of the source, might it make sense to allow a plugin to say "don't copy the source to `parts/$name/build`, and just use the source in `parts/$name/src` directly?

cjp256:
I don't really have a strong preference on this.  @jhenstridge's idea makes sense to allow the plugin to opt out of snapcraft-managed build directory - it does feel odd to hard-code an additional work directory. @cmatsuoka may have some thoughts as he's doing related work...

kyrofa:
> If the build directory would just be a copy of the source, might it make sense to allow a plugin to say "don't copy the source to parts/$name/build, and just use the source in parts/$name/src directly?

Yeah that would work for us. The problem is, that's the way the v1 plugin API worked, and v2 explicitly got rid of it, so we were trying to respect that decision while also being able to eat our cake :stuck_out_tongue: .

@cjp256 if you're open to re-introducing that feature, how would you like to see it implemented?

jhenstridge:
The v2 meson plugin is another option to consider, which creates an empty `.snapbuild` directory underneath the part's build directory to run the build:

https://github.com/snapcore/snapcraft/blob/a0ceca9d531a34c979251030ed67b5fa2abfdd9a/snapcraft/plugins/v2/meson.py#L88-L90

If Snapcraft did offer a "don't copy my sources" flag, then it would probably be cleaner to just have it build directly in the build directory from sources in the source directory.  But without that, this could be a pattern to follow in other plugins that need a clean build directory.

kyrofa:
> The v2 meson plugin is another option to consider, which creates an empty .snapbuild directory underneath the part's build directory to run the build

We considered that, but given that sources are automatically migrated into the build directory without plugin interaction, no directory can be chosen that doesn't have the possibility to clash with directories in the source. That's why we ended up going the direction in this PR. Of course, not migrating the sources into the build directory accomplishes the same thing, and I think is far preferable if folks are open to it.

jhenstridge:
Yep.  That's definitely a potential problem for the meson plugin too.  I think this is definitely something worth solving properly though, given most projects using cmake, meson, and probably qmake would expect out-of-tree builds.

Autotools projects could also benefit, but I suspect enabling it by default there would just cause problems: many projects don't test that mode, and some that do still end up creating files in the source directory.

jhenstridge:
> The problem is, that's the way the v1 plugin API worked, and v2 explicitly got rid of it, so we were trying to respect that decision while also being able to eat our cake :stuck_out_tongue: .

Was it removed because they thought it was a bad idea, or just as part of the general "simplify plugins as much as possible" changes?  Either way, I had a go at adding the feature back in with PR #3392, and doesn't look to add much complexity.  Maybe that will fit what ROS needs?

sergiusens:
Kyle Fazzari writes:

>> The v2 meson plugin is another option to consider, which creates an empty .snapbuild directory underneath the part's build directory to run the build
>
> We considered that, but given that sources are automatically migrated into the build directory without plugin interaction, no directory can be chosen that doesn't have the possibility to clash with directories in the source. That's why we ended up going the direction in this PR. Of course, not migrating the sources into the build directory accomplishes the same thing, and I think is far preferable if folks are open to it.

This might as well be a property of the part or source,
source-<something> or maybe least preferable, another build-attributes
keyword.

sergiusens:
I am +1 on @jhenstridge 's approach, if that works for you let's please continue there and close this one.

kyrofa:
> if that works for you let's please continue there and close this one.

+1, let's close this and move over there.

artivis:
Alright then, closing this and moving on to @jhenstridge 's approach.

PR #3356, opened 2020-11-05
extensions: correctly spell the GTK_USE_PORTAL variable

seb128:
Testing something in thunderbird I checked the environment and notice that GTK_USE_PORTALS was set, sounds like the intend is to define GTK_USE_PORTAL (which was recently manually added)? The commit should fix the issue


cjp256 (approved):
Interesting that this hasn't been noticed until now :thinking: LGTM, thanks!

sergiusens (approved):

PR #3362, opened 2020-11-09
project loader grammar: do not transform build-packages

cjp256:
Build packages are almost never supposed to be a foreign
architecture.  This may work if the package is effectively
a source package, but it ultimately very prone to conflicts
with the native host package.

Given that:

1. This behavior is more likely to do harm than good, it seems
   doubtful that it has ever been in use.

2. This behavior appears to be undocumented.

3. Usage will continue to work in the cases where the
   user is explicitly setting target-arch for build packages.

4. Usage will likely work in cases where the user is getting
   the implied target-arch package.

5. Cross-compiling via --target-arch is quite limited.

This patch applies this change globally.

-----


cjp256:
Alternatively, we have the option to limit this to core20+ to be "safest". But I'm not sure it's worth it.

sergiusens (approved):

PR #3364, opened 2020-11-10
project loader: ensure all key assets are utilized

cjp256:
If any keys are unutilized, raise an error to the user.

-----


snapcraft/internal/project_loader/errors.py
```diff
@@ -119,3 +121,17 @@ class SnapcraftAfterPartMissingError(ProjectLoaderError):
...
     def __init__(self, part_name, after_part_name):
         super().__init__(part_name=part_name, after_part_name=after_part_name)
+
+
+class SnapcraftProjectUnusedKeyAssetError(snapcraft.internal.errors.SnapcraftException):
+    def __init__(self, key_path: pathlib.Path):
+        self.key_path = key_path
+
+    def get_brief(self) -> str:
+        return f"Found unused key asset {self.key_path!s}."
```
sergiusens, line 131: self.key_path is of type Path, which probably prevented you from using repr, which led to this that does not add quotes to the string. You might need format here or enclose self.key_path in single quotes to keep the output consistent.

cjp256: Fixed, thnx

sergiusens (approved):

snapcraft/internal/project_loader/_config.py
```diff
@@ -255,6 +256,16 @@ def _get_required_package_repositories(self) -> List[PackageRepository]:
 
         return package_repos
 
+    def _verify_all_key_assets_utilized(
```
cmatsuoka, line 259: You're checking if all keys are installed on our key manager, but we can't tell if they're actually used to validate the packages we install, is that correct? Perhaps we could reflect that in the function name (`_verify_all_key_assets_installed` instead of `_verify_all_key_assets_utilized`)?

cmatsuoka (commented):
Thanks Chris. Just added a question about function naming.

snapcraft/internal/project_loader/_config.py
cjp256: interesting point.  Though the intent is to determine that all key assets are utilized, not installed.  Though it does that be checking that they are all installed. :thinking:

cjp256: Fixed

PR #3369, opened 2020-11-11
sources: enable 7z, bzr, hg, svn, zip for non-linux

cjp256:
Zip uses the python zipfile implementation.

The remainder will fail if the equivalent tools are not in the
user's path on the host. That error should be more obvious
than the "KeyError" uses are currently getting when using
remote-build.  Future work should update these source handlers
to use file_utils.get_host_tool_path() to perform this check
prior to executing the command.

LP: #1898877

-----

### DIFF
--- a/snapcraft/internal/sources/__init__.py
+++ b/snapcraft/internal/sources/__init__.py
@@ -112,11 +112,28 @@ if sys.platform == "linux":
         "": Local,
     }
 else:
+    from ._7z import SevenZip  # noqa
+    from ._bazaar import Bazaar  # noqa
     from ._git import Git  # noqa
     from ._local import Local  # noqa
+    from ._mercurial import Mercurial  # noqa
+    from ._subversion import Subversion  # noqa
     from ._tar import Tar  # noqa
+    from ._zip import Zip  # noqa
 
-    _source_handler = {"git": Git, "local": Local, "tar": Tar, "": Local}
+    _source_handler = {
+        "7z": SevenZip,
+        "bzr": Bazaar,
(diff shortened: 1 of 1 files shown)


sergiusens (approved):
can you just confirm this works as expected for all these source types?

cjp256:
It does not.  I pushed one required fix for sources._base.  I'll test all of the combinations on win+mac.

sergiusens:
Any update on the testing, there are also conflicts now

cjp256:
To clarify, because I apparently forgot 🤦 

When processing the parts, the KeyError is raised during _get_part_source_handler() in remote build's worktree.  By allowing these source types on OSX, we could process these additional types rather than exploding.  In most cases, the source handler will not be invoked anyways hey aren't "local" sources.  Launchpad does the actual fetching unless the user specifically requests `--package-all-sources`.

If using `--package-all-sources` and the required tool is not installed, it may fail with something like `FileNotFoundError: [Errno 2] No such file or directory: 'bzr': 'bzr'`. Hence the suggestion in the commit to eventually use `file_utils.get_host_tool_path()` in the source handlers.

However, further testing has made me realize that remote build will be broken if using any of these with local sources:
- deb
- rpm
- snap
- tar
- zip

The current checks determine that the sources are local if the source handler is an instance of `Local` or git's `is_local()` is True.  The listed handlers are instances of FileBase which may or may not be local.  If remote, behavior is OK.  If local, launchpad will eventually error out unable to find the desired source.  Need to fix this in the future too.

sergiusens (approved):

PR #3380, opened 2020-11-18
snap packaging: fix broken path check when copying local icon

cjp256:
If an icon is picked up from the desktop file, it would fail
to copy due to incorrect source parameter for file copy

The appstream-desktop spread test never picked up this case:

(1) icon is specified in appstream-desktop metainfo

(2) Snapcraft will parse metainfo from a file found relative
    to src and build.   In these cases, extract() is never
    called to search the `install` directory, which is where
    the desktop file is found.

This introduces another test case to pick this up.  It reuses
the appstream-desktop snap, but strips icon from metainfo,
forcing the parsing of the desktop file.  This also requires
that the snap now install metainfo so Snapcraft will properly
find the desktop file.

It is interesting that Snapcraft is rewriting the desktop file
Icon to point to the meta/gui/icon.svg now that the metainfo is
installed in the snap.  It's functionally the same because the
icon matches, but it is a strange change in behavior...

-----


sergiusens:
Was this always badly behaved? This might not produce the desired results for multi app snaps. We should talk more about this.

sergiusens (approved):

PR #3381, opened 2020-11-19
multipass build provider: check if instance exists before deleting

cjp256:
Snapcraft may attempt to delete a multipass instance that does
not exist. Check to see if instance exists prior to calling delete.

This may occur if Snapcraft sees project provider directory, but
the instance does not exist when it goes to clean it.

This commit adds a new exists() method to multipass commands. While 
it's not a "real" multipass command, it seems to be a fair interface 
to provide without overloading the usage of info() or another command.

LP #1904774

-----


snapcraft/internal/build_providers/_multipass/_multipass_command.py
```diff
@@ -251,6 +249,14 @@ def execute(
 
         return output
 
+    def exists(self, *, instance_name: str) -> bool:
```
sergiusens, line 252: Why not just make this part of the Multipass class?

sergiusens (changes requested):
We probably need to discuss the criteria for this :-)

sergiusens:
You could also add support for list and run:

    multipass list --format json

Issuing a warning might be a good idea if we should expect the image to exist (telling the user that they should't delete out of band)

snapcraft/internal/build_providers/_multipass/_multipass_command.py
cjp256: The current tests make assumptions on usage of `info` as part of its mock strategy.  We could refactor those tests and move it to multipass.  Strictly speaking, I agree it belongs there.

cjp256:
> You could also add support for list and run:
> 
> ```
> multipass list --format json
> ```
> 

We could...

> Issuing a warning might be a good idea if we should expect the image to exist (telling the user that they should't delete out of band)

It would seem to me to be an unnecessary warning - we can handle this case appropriately without user concerns.  I do think it would be reasonable to have an informational prompt in the future when we have a nice command/option to purge all snapcraft-created environments and suggest they use that instead.

sergiusens (approved):

PR #3382, opened 2020-11-19
project loader: purge dead code in Config

cjp256:
stage_env() does not appear to be used anywhere.
    
Purge it along with associated unit test.


sergiusens:
Mind updating?

cjp256:
Updated.

sergiusens (approved):

PR #3389, opened 2020-11-27
plugins v1: move build base test in v1 plugins to base class

cmatsuoka:
V1 plugins can be used only if the build base is core, core16 or
core18. As an additional sanity check, each plugin tests for base
consistency. To clean that up, move the base check from each
individual plugin to the V1 plugin base class constructor.
    
### DIFF
--- a/snapcraft/plugins/v1/_plugin.py
+++ b/snapcraft/plugins/v1/_plugin.py
@@ -21,6 +21,7 @@ import shlex
 from subprocess import CalledProcessError
 from typing import List
 
+from snapcraft.project import Project
 from snapcraft.internal import common, errors
 from snapcraft.internal.meta.package_repository import PackageRepository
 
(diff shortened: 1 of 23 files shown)


snapcraft/plugins/v1/_plugin.py
```diff
@@ -86,6 +87,15 @@ def __init__(self, name, options, project=None):
         self.options = options
 
         if project:
+            if isinstance(project, Project) and project._get_build_base() not in (
+                "core",
+                "core16",
```
cjp256, line 92: Maybe this is a good time to drop "core16"?

cjp256:
Other than some cleanup, is the goal of this to remove the dependency on project._get_build_base() throughout the plugins?

Somewhat unrelated thought: instead of checking build base, maybe we should be checking the `host_os_release`, i.e. what we really care about.  While targeting another base/arch is largely nonfunctional, I think it may be best to standardize on the explicit semantics where required so it's clear when/if we add limited cases.

What do you think about constraining this to a compatibility check that can be overridden and called from the v1 plugin init method? e.g.
```
def is_compatible(self, *, host_os_release: str, host_arch: str, target_os_release: str, target_arch: str) -> bool:
    return host_os_release in ["xenial", "bionic"]
```

The check could be limited to`build_base`.  We could do a matching interface for a v2 if we think there is value to it.  Just a thought.

snapcraft/plugins/v1/_plugin.py
sergiusens: We could, but it is in more places than here.

sergiusens (approved):
Thanks for the cleanup.

@cjp256 we should discuss your plugin generalization idea. Can you set something up?

PR #3391, opened 2020-12-11
plugins v2: port the qmake plugin

jhenstridge:

-----
This is a port of the v1 qmake plugin.  It makes the following changes to the plugin properties:

 * `options` renamed to `qmake-parameters` to match other V2 plugins
 * `qt-version` dropped, since Ubuntu 20.04 does not ship Qt 4, and there is no Qt 6 (yet).
 * `project-files` dropped in favour of `qmake-parameters`.  We're performing a srcdir=builddir build, so there is no reason to rewrite paths.


requirements.txt
```diff
@@ -6,7 +6,7 @@ PyYAML==5.3
 pyxdg==0.26
 requests==2.20.0
 requests_unixsocket==0.1.5
-http://archive.ubuntu.com/ubuntu/pool/main/p/python-apt/python-apt_1.6.5ubuntu0.2.tar.xz; sys_platform == 'linux'
+http://archive.ubuntu.com/ubuntu/pool/main/p/python-apt/python-apt_1.6.5ubuntu0.4.tar.xz; sys_platform == 'linux'
```
sergiusens, line 9: I know this is being a bit pedantic, but can this be its own "uprev" PR?

snapcraft/plugins/v2/qmake.py
```diff
@@ -0,0 +1,77 @@
...
+            'QMAKE_CFLAGS+="$CFLAGS"',
+            'QMAKE_CXXFLAGS+="$CXXFLAGS"',
+            'QMAKE_LFLAGS+="$LDFLAGS"',
+        ] + self.options.qmake_parameters
+        return " ".join(cmd)
+
+    def get_build_commands(self) -> List[str]:
+        return [
+            self._get_qmake_configure_command(),
+            '(unset CFLAGS CXXFLAGS LDFLAGS; make -j"${SNAPCRAFT_PARALLEL_BUILD_COUNT}")',
```
sergiusens, line 75: I guess unset is done as it is already done in _get_qmake_conflgure_commands. Mind adding a comment for that if it is the case, or if there is another reason, why the unset is done for future observers?

sergiusens (commented):
Thanks for contributing this. Is the goal to make use of #3392 in this PR?

jhenstridge:
I think this will need some reworking if #3392 is accepted, yes.  We can't fold `project-files` into `qmake-parameters`, since they would need to be rewritten to point into `${SNAPCRAFT_PART_SRC_WORK}`.

snapcraft/plugins/v2/qmake.py
cjp256: I expect that the exported ones may conflict with qmake generated ones... but certainly would appreciate a comment :D

This is certainly a fine approach.  FWIW when I need to do something like this, i'll use `env -u <var1>...`

cjp256 (dismissed):
LGTM, but a comment with the unset would be icing on the cake!

Thanks!

snapcraft/plugins/v2/qmake.py
```diff
@@ -0,0 +1,77 @@
...
+                    "default": [],
+                }
+            },
+        }
+
+    def get_build_snaps(self) -> Set[str]:
+        return set()
+
+    def get_build_packages(self) -> Set[str]:
+        return {"g++", "make", "qt5-qmake", "qtbase5-dev"}
```
sergiusens, line 58: is there a chance we can avoid installing qtbase5-dev to allow satisfying qt5 through other means (like a kde extension?)

snapcraft/plugins/v2/qmake.py
jhenstridge: In particular, the generated Makefile generated by qmake for the sample snap contains:

```
CFLAGS        = -pipe -O2 -fPIC $(DEFINES)
CXXFLAGS      = -pipe -O2 -fPIC $(DEFINES)
```

Having the environment variables set during the build will override the values from the Makefile, potentially miscompiling the project.

I would not be surprised if the same is true for a few other plugins.

snapcraft/plugins/v2/qmake.py
jhenstridge: I'd included `qtbase5-dev` because it was in the V1 plugin.  Things seem to work fine after removing it, so that's a nice simplification.

jhenstridge:
I've rebased the plugin on top of PR #3392, and switched to out-of-tree builds.  I reintroduced the `qmake-project-files` option, since those arguments need to be rewritten to be `${SNAPCRAFT_PART_SRC_WORK}` relative.

jhenstridge:
I've changed it again to just `qmake-project-file`, as passing multiple project files does not seem to be supported or really make sense.

snapcraft/plugins/v2/qmake.py
```diff
@@ -0,0 +1,99 @@
...
+
+This plugin uses the common plugin keywords as well as those for "sources".
+For more information check the 'plugins' topic for the former and the
+'sources' topic for the latter.
+
+Additionally, this plugin uses the following plugin-specific keywords:
+
+    - qmake-parameters:
+      (list of strings)
+      additional options to pass to the qmake invocation.
```
sergiusens, line 29:
```suggestion
      additional options to pass to the qmake invocation.

```

snapcraft/plugins/v2/qmake.py
```diff
@@ -0,0 +1,100 @@
...
+
+    - qmake-project-file:
+      (string)
+      the qmake project file to use. This is usually only needed if
+      qmake can not determine what project file to use on its own.
+
```
sergiusens, line 35:
```suggestion
```

snapcraft/plugins/v2/qmake.py
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+    def get_build_snaps(self) -> Set[str]:
+        return set()
+
+    def get_build_packages(self) -> Set[str]:
+        return {"g++", "make", "qt5-qmake"}
+
+    def get_build_environment(self) -> Dict[str, str]:
+        return {"QT_SELECT": "qt5"}
```
sergiusens, line 67: Taking a note to see if this is required.

sergiusens (approved):

snapcraft/plugins/v2/qmake.py
jhenstridge: IIRC, the qtchooser code is still part of the archive in 20.04, and you'd need to install `qt5-default` for things to function without the environment variable.

PR #3392, opened 2020-12-15
plugins v2: add support for out of source tree builds

jhenstridge:

-----
This is intended as an alternative to PR #3347.  It adds an `out_of_source_build` property to `PluginV2`, which is interpreted the same way it would be on `PluginV1` implementations, in that it disables the automatic copying of a part's source code to the build directory.  The default implementation of the property is `plugin.out_of_source_build == False`.

To demonstrate that this works, I've ported the meson and cmake plugins to use this mode.  For what it is worth, the cmake plugin was already configuring the build to use the copy of the source in `part_source_dir`, ignoring the copy in `part_build_dir`.

### DIFF
--- a/snapcraft/internal/pluginhandler/__init__.py
+++ b/snapcraft/internal/pluginhandler/__init__.py
@@ -257,6 +257,7 @@ class PluginHandler:
         dirs = [
             self.part_source_dir,
             self.part_build_dir,
+            self.part_build_work_dir,
             self.part_install_dir,
             self.part_state_dir,
             self._project.stage_dir,
--- a/snapcraft/internal/pluginhandler/__init__.py
+++ b/snapcraft/internal/pluginhandler/__init__.py
@@ -575,10 +575,7 @@ class PluginHandler:
     def build(self, force=False):
         self.makedirs()
 
-        if not (
-            isinstance(self.plugin, plugins.v1.PluginV1)
-            and self.plugin.out_of_source_build
-        ):
+        if not self.plugin.out_of_source_build:
             if os.path.exists(self.part_build_dir):
                 shutil.rmtree(self.part_build_dir)
 
@@ -589,10 +586,7 @@ class PluginHandler:
         self._do_build()
 
     def update_build(self):
-        if not (
-            isinstance(self.plugin, plugins.v1.PluginV1)
-            and self.plugin.out_of_source_build
-        ):
+        if not self.plugin.out_of_source_build:
             # Use the local source to update. It's important to use
             # file_utils.copy instead of link_or_copy, as the build process
             # may modify these files
--- a/snapcraft/internal/pluginhandler/__init__.py
+++ b/snapcraft/internal/pluginhandler/__init__.py
@@ -94,7 +94,10 @@ class PluginHandler:
         # part property.
         source_sub_dir = self._part_properties.get("source-subdir", "")
         self.part_source_work_dir = os.path.join(self.part_source_dir, source_sub_dir)
-        self.part_build_work_dir = os.path.join(self.part_build_dir, source_sub_dir)
+        if self.plugin.out_of_source_build:
+            self.part_build_work_dir = self.part_build_dir
+        else:
+            self.part_build_work_dir = os.path.join(self.part_build_dir, source_sub_dir)
 
         self._pull_state: Optional[states.PullState] = None
         self._build_state: Optional[states.BuildState] = None
@@ -257,7 +260,6 @@ class PluginHandler:
         dirs = [
             self.part_source_dir,
             self.part_build_dir,
-            self.part_build_work_dir,
             self.part_install_dir,
             self.part_state_dir,
             self._project.stage_dir,
--- a/snapcraft/plugins/v2/_plugin.py
+++ b/snapcraft/plugins/v2/_plugin.py
@@ -66,3 +66,13 @@ class PluginV2(abc.ABC):
         snapcraftctl can be used in the script to call out to snapcraft
         specific functionality.
         """
+
+    @property
+    def out_of_source_build(self) -> bool:
+        """Returns True if the plugin performs out-of-source-tree builds.
+
+        In practice, this controls whether the PluginHandler code will
+        copy the source code to the build directory before invoking
+        the plugin's build commands.
+        """
+        return False
--- a/snapcraft/plugins/v2/_plugin.py
+++ b/snapcraft/plugins/v2/_plugin.py
@@ -54,6 +54,14 @@ class PluginV2(abc.ABC):
         This method is called by the PluginHandler during the "build" step.
         """
 
+    out_of_source_build = False
+    """Set to True if the plugin performs out-of-source-tree builds.
+
+    In practice, this controls whether the PluginHandler code will
+    copy the source code to the build directory before invoking the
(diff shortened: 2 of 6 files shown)


sergiusens (approved):
This looks good, I think we wanted a way to force in-tree builds through user configuration, I cannot recall if we had some use cases for that already.

@cjp256 any thoughts?

snapcraft/plugins/v2/_plugin.py
```diff
@@ -66,3 +66,13 @@ def get_build_commands(self) -> List[str]:
         snapcraftctl can be used in the script to call out to snapcraft
         specific functionality.
         """
+
+    @property
+    def out_of_source_build(self) -> bool:
```
kyrofa, line 71: Why not just have this be an attribute?

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -257,6 +257,7 @@ def makedirs(self):
         dirs = [
             self.part_source_dir,
             self.part_build_dir,
+            self.part_build_work_dir,
```
kyrofa, line 260: Why create this here? Isn't the source copied over wholesale (or, as of this PR, potentially not at all)?

kyrofa:
>  For what it is worth, the cmake plugin was already configuring the build to use the copy of the source in part_source_dir, ignoring the copy in part_build_dir.

Hahaha

snapcraft/internal/pluginhandler/__init__.py
sergiusens: Still copied but wiped (code block below), and now that I see, wiped every time killing our rebuild story

snapcraft/internal/pluginhandler/__init__.py
```diff
@@ -575,10 +576,7 @@ def prepare_build(self, force=False):
     def build(self, force=False):
         self.makedirs()
 
-        if not (
-            isinstance(self.plugin, plugins.v1.PluginV1)
-            and self.plugin.out_of_source_build
-        ):
+        if not self.plugin.out_of_source_build:
```
sergiusens, line 579: for v2 we do not wipe anymore, to allow for faster iterative builds. Can that be taken into account? As @kyrofa mentions, might be better to just avoid the copy

sergiusens (changes requested):

snapcraft/internal/pluginhandler/__init__.py
kyrofa: > wiped every time killing our rebuild story

Remember we also have `update_build()`.

jhenstridge: I needed this change to allow the `cmake-hello-subdir` spread test to pass.  That snap has a part with `source-subdir: subdir` set, which meant that it tries to execute the build script in `parts/hello/build/subdir`.  Without copying the source code over, this directory wouldn't otherwise exist.

snapcraft/plugins/v2/_plugin.py
jhenstridge: I had originally marked this with `@abc.abstractmethod`, but removed it so I could have the other plugins take a default implementation.

I suppose it would make sense to either switch it to a regular attribute, or add back the `abstractmethod` decorator which would ensure every plugin made a choice.

jhenstridge:
> This looks good, I think we wanted a way to force in-tree builds through user configuration, I cannot recall if we had some use cases for that already.

Are there actually use cases for this?  As I see it, we essentially have four categories for build systems:

1. those that only support out-of-tree builds (e.g. meson)
2. those that support both in-tree and out-of-tree builds, and are normally used out-of-tree (e.g. cmake)
3. those that support both, but are normally used in-tree (e.g. autotools)
4. those that only support in-tree builds (maybe Python and Node.js?)

I would usually set up (1) and (2) to use out of tree builds, and (3) and (4) to use in-tree builds.  An override of that policy would only make sense for a small number of projects in cases (2) and (3).  How many autotools projects fail to build in-tree?  How many cmake projects fail to build out-of-tree?

snapcraft/internal/pluginhandler/__init__.py
jhenstridge: As @kyrofa mentioned in the other thread, isn't this covered by the existing `PluginHandler.build()` vs. `PluginHandler.update_build()` split?

sergiusens:
James Henstridge writes:

>> This looks good, I think we wanted a way to force in-tree builds through user configuration, I cannot recall if we had some use cases for that already.
>
> Are there actually use cases for this?  As I see it, we essentially have four categories for build systems:
>
> 1. those that only support out-of-tree builds (e.g. meson)
> 2. those that support both in-tree and out-of-tree builds, and are
> normally used out-of-tree (e.g. cmake)

And the strange change to do out of tree in v2 (it was in tree 3 months
ago) was due to some projects failing unless it was out of tree (custom
logic).

> 3. those that support both, but are normally used in-tree
> (e.g. autotools)

Normally, and mostly doesn't work out of tree.

> 4. those that only support in-tree builds (maybe Python and Node.js?)

And here is why I replied :-) Python is tricky, =setup.cfg= leaks to
dependencies as well and causes issues for some projects.

> I would usually set up (1) and (2) to use out of tree builds, and (3)
> and (4) to use in-tree builds.  An override of that policy would only
> make sense for a small number of projects in cases (2) and (3).  How
> many autotools projects fail to build in-tree?  How many cmake
> projects fail to build out-of-tree?

I guess this is a fair point, and we are moving to making our plugins
opinionanated with the hopes that once we implement =expand-plugins=
that whomever needs futher customization can copy paste de
=override-build= stanza.

snapcraft/plugins/v2/_plugin.py
cjp256: +1 for class attribute with default False like you have here

snapcraft/internal/pluginhandler/__init__.py
cjp256: v2 falls under the _do_v2_build() logic, not this.  In fact, if a v2 were to execute this, the if would evaluate to True :D

snapcraft/internal/pluginhandler/__init__.py
cjp256: That is because this snapcraft accounts for the working directory when subdir is used: 
https://github.com/snapcore/snapcraft/blob/master/snapcraft/internal/pluginhandler/__init__.py#L684

Specifically `cwd=self.part_build_work_dir`

If using out of source builds, I think `cwd=self.part_build_dir` is appropriate?

jhenstridge: I've updated the code to set part_build_work_dir to part_build_dir when a plugin uses out_of_source_build.  That will have the build occur in the base build dir, and make sure `$SNAPCRAFT_PART_BUILD_WORK` actually points at where the build is occurring, if it happens to be used by a scriptlet.

sergiusens:
Can we get the property back please? keeps it consistent

jhenstridge:
> Can we get the property back please? keeps it consistent

Done.  I guess I misunderstood @cjp256's "+1 for class attribute" feedback.

sergiusens (approved):

PR #3395, opened 2020-12-16
grammar: fix the case of disappearing statements

cjp256:
Any statement prior to an 'on.. to..' clause vanishes into
the abyss.

E.g.:

```
stage-packages:
  - on amd64 to arm64:
    - foo
  - on amd64 to i386:
    - foo2
``` 

In this case `foo` will never be added to the package list.

When starting to process an on/to statement, ensure the previous
one is added to the statement collection.

Add a test to cover this case and update the basic grammar tests
to explicitly declare the target (deb) arch.

-----

### DIFF
--- a/snapcraft/internal/project_loader/grammar/_processor.py
+++ b/snapcraft/internal/project_loader/grammar/_processor.py
@@ -134,6 +134,9 @@ class GrammarProcessor:
             if on_to_clause_match:
                 # We've come across the beginning of a compound statement
                 # with both 'on' and 'to'.
+                # The first time through this may be None, but the
+                # collection will ignore it.
+                statements.add(statement)
 
                 # First, extract each statement's part of the string
                 on, to = on_to_clause_match.groups()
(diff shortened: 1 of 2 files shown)


snapcraft/internal/project_loader/grammar/_processor.py
```diff
@@ -134,6 +134,9 @@ def _parse_dict(
             if on_to_clause_match:
                 # We've come across the beginning of a compound statement
                 # with both 'on' and 'to'.
+                # The first time through this may be None, but the
+                # collection will ignore it.
+                statements.add(statement)
```
sergiusens, line 139: why don't you just check for None?

cjp256: It's consistent with the remaining clauses (comment included).

I have some minor refactoring which takes care of this, but I wanted to push the bug fix first.

cjp256: I added a commit to do the refactor.

sergiusens (approved):
this is easier to read too

kyrofa (approved):
Thank you for your diligence, this looks good.

cmatsuoka (approved):
Thanks Chris, this looks correct to me.

PR #3397, opened 2020-12-16
cli: support getting snap status by track

kyrofa:

-----

Currently snapcraft supports getting the status by architecture. Add support for getting the status by track as well.

### DIFF
--- a/snapcraft/cli/store.py
+++ b/snapcraft/cli/store.py
@@ -497,13 +497,16 @@ def close(snap_name, channels):
 @click.option(
     "--arch", metavar="<arch>", help="The snap architecture to get the status for"
 )
+@click.option("--track", metavar="<track>", help="The snap track to get the status for")
 @click.argument("snap-name", metavar="<snap-name>")
-def status(snap_name, arch, experimental_progressive_releases):
+def status(snap_name, arch, track, experimental_progressive_releases):
     """Get the status on the store for <snap-name>.
 
     \b
     Examples:
         snapcraft status my-snap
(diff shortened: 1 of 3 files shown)


kyrofa:
The spread failure seems unrelated, but I don't have the ability to retry them.

snapcraft/cli/store.py
```diff
@@ -497,13 +497,16 @@ def close(snap_name, channels):
 @click.option(
     "--arch", metavar="<arch>", help="The snap architecture to get the status for"
 )
+@click.option("--track", metavar="<track>", help="The snap track to get the status for")
```
sergiusens, line 500: what if we make it a list of tracks? Would that be useful?

kyrofa: Definitely, I just wanted to match the architecture there. Would you like to see just `multiple=True`? Should we introduce the same for `--arch`? Should we also introduce short options, or would you like to see it spelled out each time?

sergiusens: Let's go with `multiple=True`, this would give us a no syntax change feature in the future for `--arch`.
Let's leave short options out for now, those require some bike shedding I am not prepared to do so close to the end of the year :-)

Thanks for working on this!

kyrofa: Alright will do.

kyrofa: Done.

sergiusens (approved):
Nice, thanks for adding support for arch as well!

PR #3398, opened 2020-12-17
autotools v2 plugin: support autogen.sh and bootstrap

sergiusens:

-----

### DIFF
--- a/snapcraft/plugins/v2/autotools.py
+++ b/snapcraft/plugins/v2/autotools.py
@@ -72,6 +72,8 @@ class AutotoolsPlugin(PluginV2):
 
     def get_build_commands(self) -> List[str]:
         return [
+            "[ ! -f ./configure ] && [ -f ./autogen.sh ] && env NOCONFIGURE=1 ./autogen.sh",
+            "[ ! -f ./configure ] && [ -f ./bootstrap ] && env NOCONFIGURE=1 ./bootstrap",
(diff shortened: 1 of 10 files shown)


snapcraft/plugins/v2/autotools.py
```diff
@@ -72,6 +72,8 @@ def _get_configure_command(self) -> str:
 
     def get_build_commands(self) -> List[str]:
         return [
+            "[ ! -f ./configure ] && [ -f ./autogen.sh ] && env NOCONFIGURE=1 ./autogen.sh",
```
cjp256, line 75: From my limited reading, it sounds like a project would generally use `autogen.sh` script, `bootstrap` script, or invoke `autoreconf`.  autoreconf after autogen might override the results of the autogen.sh / boostrap?

FWIW according to https://www.gnu.org/software/automake/faq/autotools-faq.html#What-does-_002e_002fbootstrap-or-_002e_002fautogen_002esh-do_003f  `bootstrap` is preferred over 'autogen.sh' (for the case both are present?)

sergiusens: we discussed this online. bootstrap or autogen.sh will generate should configure script at the end of their run so these three lines are mutually exclusive.

kenvandine:
This looks good to me.  I've tested it with gnome-3-38-2004 with success.  I was able to remove all my hacks.

PR #3399, opened 2020-12-18
cli: do not require snapd deb for assertions

sergiusens:
Assertion commands do not require snapd to be installed through the
deb, it just needs to be available for most commands.

LP: #1906506

-----

### DIFF
--- a/snapcraft/_store.py
+++ b/snapcraft/_store.py
@@ -34,8 +34,12 @@ from snapcraft import storeapi, yaml_utils
 
 # Ideally we would move stuff into more logical components
 from snapcraft.cli import echo
-from snapcraft.file_utils import calculate_sha3_384, get_snap_tool_path
-from snapcraft.internal import cache, deltas, repo
+from snapcraft.file_utils import (
+    calculate_sha3_384,
+    get_host_tool_path,
+    get_snap_tool_path,
+)
+from snapcraft.internal import cache, deltas
 from snapcraft.internal.deltas.errors import (
     DeltaGenerationError,
     DeltaGenerationTooBigError,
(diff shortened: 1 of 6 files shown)


snapcraft/_store.py
```diff
@@ -495,6 +498,14 @@ def list_keys():
             tablefmt="plain",
         )
         print(tabulated_keys)
+    elif not keys and enabled_keys:
+        registered_keys = "\n".join([f"- {key}" for key in enabled_keys])
+        print(
+            "No keys have been created on this system. "
```
cjp256, line 504: Maybe?
```suggestion
            "No keys found. "
```

Interesting that snapd doesn't seem to have an interface to import a key you exported?  Is it just expected you create a new key on a new host rather than maintain them?

cjp256 (approved):

snapcraft/_store.py
```diff
@@ -495,6 +498,14 @@ def list_keys():
...
+            " See 'snapcraft create-key --help to create a key.\n"
```
facundobatista, line 505: Missing `'` to close the suggested command.

snapcraft/_store.py
```diff
@@ -433,8 +437,9 @@ def list_registered():
 
 
 def _get_usable_keys(name=None):
+    snap_path = get_host_tool_path(command_name="snap", package_name="snapd")
```
facundobatista, line 440: What if there's no `snap`? Shouldn't this raise `storeapi.errors.MissingSnapdError`? How that case is managed?

If not, shouldn't that exception be removed from `storeapi.errors`?

facundobatista (commented):
Added a couple of details.

snapcraft/_store.py
sergiusens: get_host_tool_path raises an appropriately formatted exception which is generic enough to be valid on any system Snapcraft is ever ported to https://github.com/snapcore/snapcraft/blob/e64c08aeddbc98df01c8644ce1d2762af2c30620/snapcraft/file_utils.py#L340

I will check though if this MissingSnapdError needs to be cleaned up

PR #3410, opened 2021-01-11
project: enable experimental target-arch support for core20

cjp256:
Allow use of --target-arch with an *EXPERIMENTAL* warning.

-----

### DIFF
--- a/snapcraft/project/_sanity_checks.py
+++ b/snapcraft/project/_sanity_checks.py
@@ -62,9 +62,11 @@ def conduct_project_sanity_check(project: Project, **kwargs) -> None:
     if (
         project._get_build_base() in ["core20"]
         and kwargs.get("target_arch") is not None
+        and not os.getenv("SNAPCRAFT_ENABLE_EXPERIMENTAL_TARGET_ARCH")
     ):
         raise SnapcraftEnvironmentError(
-            "--target-arch has been deprecated and is no longer supported on core20."
+            "*EXPERIMENTAL* '--target-arch' configured, but not enabled. "
+            "Enable with '--enable-experimental-target-arch' flag."
         )
 
     # Icon should refer to project file, verify it exists.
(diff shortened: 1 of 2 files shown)


snapcraft/project/_sanity_checks.py
```diff
@@ -63,8 +63,9 @@ def conduct_project_sanity_check(project: Project, **kwargs) -> None:
         project._get_build_base() in ["core20"]
         and kwargs.get("target_arch") is not None
     ):
-        raise SnapcraftEnvironmentError(
-            "--target-arch has been deprecated and is no longer supported on core20."
+        logger.warning(
+            "*EXPERIMENTAL* '--target-arch' support for core20 enabled. "
```
sergiusens, line 67: should we not experimental flag this instead?

sergiusens (changes requested):
Question more than request, but seems more appropriate

snapcraft/project/_sanity_checks.py
cjp256: Want me to add `--enable-experimental-target-arch` / `SNAPCRAFT_ENABLE_EXPERIMENTAL_TARGET_ARCH`?

I considered it, but the current target-arch support is in such rough shape I could hardly think it mattered :rofl:

But more seriously:
(a) I think the user has some responsibility when there is a glaring warning when using _new_ functionality. 
(b) I am hoping to stabilize behavior in a short timeframe, in time for next release. I do believe stage-packages defaulting to target arch is the only major behavior change (which would be a desirable fix for other bases too, and I believe Snapcraft did actually do this previously).

sergiusens: My reasoning is that some folks have enabled CI with experimental features after some copy/pasting, so they might not even see the experimental nature.

sergiusens: We should mini spec x-compilation for a nicer back seat eagle eyed view to see if we are missing anything

sergiusens (approved):

cjp256:
Only the plainbox provider tests are failing.

PR #3416, opened 2021-01-13
repo: default to target arch for stage package cache

cjp256:


snapcraft/internal/repo/apt_cache.py
```diff
@@ -38,13 +38,20 @@
...
+        self,
+        *,
+        stage_cache: Optional[Path] = None,
+        stage_cache_arch: Optional[str] = None,
+    ) -> None:
         self.stage_cache = stage_cache
+        self.stage_cache_arch = stage_cache_arch
 
     def __enter__(self) -> "AptCache":
+
```
sergiusens, line 51: why?

sergiusens (commented):
Interesting, AFAIK, ros does not support cross compilation

snapcraft/internal/repo/apt_cache.py
cjp256, line 51:
```suggestion
```

cjp256: typo, fixed

cjp256:
> Interesting, AFAIK, ros does not support cross compilation

I don't think it does either - but for consistency's sake, I've included it.  I would expect target arch is always the host arch in that case.

cjp256:
FWIW, I'm going to add some spread test to cover this as well.

sergiusens (approved):

PR #3418, opened 2021-01-13
project: always set target arch even if not cross compiling

cjp256:

-----


sergiusens (commented):
is this so that the envvar is always set? do you foresee any side effects by doing this?

cjp256:
Target arch, if unspecified, is None.  If we allow target arch to be None, then I need to add logic to apt_cache to figure out the sane default.

I figured it was better just to set the correct target arch, and the usual case is just that it matches host arch.

sergiusens (approved):

PR #3430, opened 2021-01-26
More improvements for using python3.8 from within a snap

kenvandine:

-----


sergiusens (commented):
Can you explain this change? Also, shouldn't there be guards with SNAP_DESKTOP_RUNTIME to export one path or the other?

kenvandine:
Do you mean append_dir?  That just extends the list of values.  Python 3.8 from focal needs this dir be included in PYTHONPATH.

sergiusens:
You need something like this which will slow down startup time if you want to be sure about what is being imported and then gate on that:

    python3 -c "import sys; print('{}.{}'.format(sys.version_info[0], sys.version_info[1]))"

sergiusens (approved):

PR #3431, opened 2021-01-26
godeps spread test: use latest/stable go snap

cjp256:
`go get` fails with go 1.6 provided in Xenial.  Avoid this by
updating the version of go with the go snap.

-----


sergiusens (changes requested):
Nice, but seems this is not enough as tests fail

    The go snap from the Snap Store should not have been installed.

We might want to do this only for 16.04 too.

sergiusens (approved):

PR #3437, opened 2021-02-16
extensions: check that the platform snap is connected in desktop extensions and bail out if not (LP: #1915712)

oSoMoN:


sergiusens (changes requested):
This looks good!

There is on issue which is that I think you got the name wrong for kde snaps on core18 (the name of the snap is longer).

I would also encourage using a build parameter to make so that the content snap name gets encoded in desktop exports instead of an environment variable to not pollute user space as much, something like this https://github.com/snapcore/snapcraft/commit/05d015905a56e89661464062d3a9b304d264dbe7

oSoMoN:
@sergiusens I applied the changes you suggested, and I also modified the configure hook to exit with success when the platform snap isn't connected, because we don't want this scenario to make the snap installation fail.

sergiusens:
Looks nice, thanks.

How widely available is snapd 2.43 (due to assumes)? Maybe a question for @mvo5 or @anonymouse64

anonymouse64:
@sergiusens snapd 2.43 is rather old, the only systems which routinely show up with older snapds than that are systems like debian / raspbian that have a very old debian version of snapd (like 2.37 IIRC) that doesn't know to install the core/snapd snaps, so it ends up with old version of snapd if the first snap they try to install is core18 based.

extensions/desktop/common/fonts
```diff
@@ -25,6 +25,13 @@ else
   ARCH="$SNAP_ARCH-linux-gnu"
 fi
 
+if ! snapctl is-connected "%PLATFORM_PLUG%"; then
```
sergiusens, line 28: I talked to the snapd folks and it seems that this might just be a snapd bug proper, this is still valida workaround, but it seems that it still makes sense to create fonts.conf below so that once chromium gets connected the FONCCONFIG_FILE exist and treated appropriately at first launch.

So maybe, to be future proof, just gate this conditional to running fc-cache. I also don't see this being an issue per se anymore considering the `|| echo below`, but the check is not harmful.

oSoMoN: Right, given the `|| echo` below, this check isn't necessary anymore in the configure hook.
I'll remove it. It is still valid in desktop-exports I think.

oSoMoN:
The google:ubuntu-20.04-64 failure is a flake (apt updates failed).

sergiusens (approved):

sergiusens (approved):

PR #3439, opened 2021-02-18
Don't error out if font cache generation fails.  This can cause

kenvandine:
installation to fail on some systems.

-----

### DIFF
--- a/extensions/desktop/common/fonts
+++ b/extensions/desktop/common/fonts
@@ -40,6 +40,6 @@ EOF
 
 export FONTCONFIG_FILE="${SNAP_COMMON}/fontconfig/fonts.conf"
 
-"${SNAP_DESKTOP_RUNTIME}/usr/bin/fc-cache" --force --system-only --verbose
+"${SNAP_DESKTOP_RUNTIME}/usr/bin/fc-cache" --force --system-only --verbose || echo "configure hook: font generation failed"
 
 exec "$@"


sergiusens (changes requested):
Have a couple of questions.

- why would this fail?
- shouldn't https://github.com/snapcore/snapcraft/pull/3439/files#diff-34ec5da9fffb6d26109c684307721248d5ae60974d7697899a9c3cb1bd3f14c9R41 be modified instead?
- how does this conflict with the work on #3437 ?

kenvandine:
> Have a couple of questions.
> 
> * why would this fail?

For example, if a system has no fonts installed it could fail.  We're running into this on the Ubuntu Core based desktop now, where there aren't system fonts.  Ultimate it shouldn't fail because fonts are made available to the snap, but right now that's an issue because various implementation details aren't handled for the confined desktop yet.  I've also seen reports of this failing to install in some test environments.  

Regardless of font availability, I don't think the configure hook should actually fail.  If the cache isn't generated, it'll simply get created at app launch time.  At least this means the snap installation succeeds.

> * shouldn't https://github.com/snapcore/snapcraft/pull/3439/files#diff-34ec5da9fffb6d26109c684307721248d5ae60974d7697899a9c3cb1bd3f14c9R41 be modified instead?

This link goes to my change, did you mean something else?

> * how does this conflict with the work on #3437 ?

I don't think this affects #3437

sergiusens:
By https://github.com/snapcore/snapcraft/pull/3439/files#diff-34ec5da9fffb6d26109c684307721248d5ae60974d7697899a9c3cb1bd3f14c9R41

I meant that that is the line that could fail that we can let slip, but removing set -e globally we lose syntax checking completely, so it would be better to have that line end with something like `|| echo "configure hook: font generation failed"`, or if there is a snapd way to make this stick in some log, then that (log.warning like).

extensions/desktop/common/fonts
```diff
@@ -40,6 +40,6 @@ EOF
 
 export FONTCONFIG_FILE="${SNAP_COMMON}/fontconfig/fonts.conf"
 
-"${SNAP_DESKTOP_RUNTIME}/usr/bin/fc-cache" --force --system-only --verbose
+"${SNAP_DESKTOP_RUNTIME}/usr/bin/fc-cache" --force --system-only --verbose || echo "configure hook: font generation failed"
```
cjp256, line 43: This should go to stderr rather than stdout.

```suggestion
"${SNAP_DESKTOP_RUNTIME}/usr/bin/fc-cache" --force --system-only --verbose || echo "configure hook: font generation failed" >&2
```

cjp256 (changes requested):

sergiusens (approved):

PR #3440, opened 2021-02-19
python v2 plugin: install wheel for build environment

cjp256:
Install wheel library into `<part-dir>/python-build-extras` and
update PYTHONPATH to include libraries installed there.

This way pip will have wheel installed and available:
(1) preventing a really ugly build error
(2) not polluting the installed components

LP: #1912244

-----

### DIFF
--- a/snapcraft/plugins/v2/python.py
+++ b/snapcraft/plugins/v2/python.py
@@ -129,6 +129,9 @@ class PythonPlugin(PluginV2):
             'SNAPCRAFT_PYTHON_PATH="$(readlink -e "${SNAPCRAFT_PYTHON_PATH}")"',
             '"${SNAPCRAFT_PYTHON_PATH}" -m venv ${SNAPCRAFT_PYTHON_VENV_ARGS} "${SNAPCRAFT_PART_INSTALL}"',
             'SNAPCRAFT_PYTHON_VENV_INTERP_PATH="${SNAPCRAFT_PART_INSTALL}/bin/${SNAPCRAFT_PYTHON_INTERPRETER}"',
+            'SNAPCRAFT_PYTHON_BUILD_EXTRAS="$(readlink -f "../python-build-extras")"',
+            'export PYTHONPATH="${SNAPCRAFT_PYTHON_BUILD_EXTRAS}${PYTHONPATH:+:$PYTHONPATH}"',
+            'pip install wheel --target "${SNAPCRAFT_PYTHON_BUILD_EXTRAS}"',
         ]
 
         if self.options.constraints:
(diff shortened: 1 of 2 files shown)


snapcraft/plugins/v2/python.py
```diff
@@ -129,6 +129,9 @@ def get_build_commands(self) -> List[str]:
             'SNAPCRAFT_PYTHON_PATH="$(readlink -e "${SNAPCRAFT_PYTHON_PATH}")"',
             '"${SNAPCRAFT_PYTHON_PATH}" -m venv ${SNAPCRAFT_PYTHON_VENV_ARGS} "${SNAPCRAFT_PART_INSTALL}"',
             'SNAPCRAFT_PYTHON_VENV_INTERP_PATH="${SNAPCRAFT_PART_INSTALL}/bin/${SNAPCRAFT_PYTHON_INTERPRETER}"',
+            'SNAPCRAFT_PYTHON_BUILD_EXTRAS="$(readlink -f "../python-build-extras")"',
```
sergiusens, line 132: This should be conditional to not having wheel staged.
How would I lock down a wheel?

Might be worth using some python logic to check python-packages and conditionally extend the list

cjp256: As noted in the LP ticket, it is desirable to not install wheel (as a build time dependency, not run time).

A couple of thoughts, we can (a) honor constraints, (b) introduce `python-wheel-version` property, (c) introduce `python-build-packages` property which gets default populated with `wheel`, (d) add wheel to `python-packages` and the publisher can filter it out if desired.

sergiusens:
Chris Patterson writes:

> @cjp256 commented on this pull request.
>
>
>
>> @@ -129,6 +129,9 @@ def get_build_commands(self) -> List[str]:
>              'SNAPCRAFT_PYTHON_PATH="$(readlink -e "${SNAPCRAFT_PYTHON_PATH}")"',
>              '"${SNAPCRAFT_PYTHON_PATH}" -m venv ${SNAPCRAFT_PYTHON_VENV_ARGS} "${SNAPCRAFT_PART_INSTALL}"',
>              'SNAPCRAFT_PYTHON_VENV_INTERP_PATH="${SNAPCRAFT_PART_INSTALL}/bin/${SNAPCRAFT_PYTHON_INTERPRETER}"',
> +            'SNAPCRAFT_PYTHON_BUILD_EXTRAS="$(readlink -f "../python-build-extras")"',
>
> As noted in the LP ticket, it is desirable to not install wheel (as a build time dependency, not run time).
>
> A couple of thoughts, we can (a) honor constraints, (b) introduce `python-wheel-version` property, (c) introduce `python-build-packages` property which gets default populated with `wheel`, (d) add wheel to `python-packages` and the publisher can filter it out if desired.

My proposal is D, with a twist. If wheel is explicitly set in
python-packages then DO NOT setup the "build time dep" as it will
already be satisfied.

We can think about locking down a sepecific version when not part of the
system at a later time, I just don't want to break that use case.

cjp256:
(replaced with https://github.com/snapcore/snapcraft/pull/3453)

PR #3445, opened 2021-02-23
Fix snap

sergiusens:

-----

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -33,7 +33,6 @@ build-packages:
   - libxslt1-dev
   - libyaml-dev
   - patch
-  - rustc
   - sed
 
 parts:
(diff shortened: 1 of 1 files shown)


cjp256 (approved):

cjp256:
It could be <3.5, as I read that cryptography will require rust after that.  Maybe 3.4 will be maintained for a long time?

snap/snapcraft.yaml
```diff
@@ -83,12 +83,17 @@ parts:
   snapcraft:
     source: .
     plugin: python
+    python-packages:
+        # Lock down cryptography to 3.4 to override the rust dependency.
+        - cryptography==3.4
```
cjp256, line 88:
```suggestion
        - cryptography<3.5
```

snap/snapcraft.yaml
```diff
@@ -114,11 +119,16 @@ parts:
 
   legacy-snapcraft:
     plugin: python
+    python-packages:
+        # Lock down cryptography to 3.4 to override the rust dependency.
+        - cryptography==3.4
```
cjp256, line 124:
```suggestion
        - cryptography<3.5
```

cjp256 (approved):

xnox:
The local setup script was not adjusted to do the same, and ditto the requirements.txt When bootstrapping this on riscv, rustc is neither installed, nor is cryptography built without rust.

We could just build-dep on rust toolchain and upgrade to new cryptography however, as rust is available everywhere.... Any thoughts?

nanonyme:
@cjp256 what makes you think 3.4 will be maintained for long time? Have you gotten explicit promise about that from someone?

PR #3446, opened 2021-02-23
extensions: suppress realpath noise on headless systems

MrCarroll:
When using extensions on systems where a desktop session has never run,
desktop-launch will create noise complaining that paths cannot be found.
Simply redirect the error messages to /dev/null.

E.G, running over SSH in Multipass, using snaps in a CLI environment.

Additionally, prevent behaviour where `rmdir $HOME` would be called
by ensuring the variable is not empty.
This is unlikely to have caused problems as rmdir would not delete directories
that have content, but it's better to not rely on semantics here.

- [Partial] Have you successfully run `./runtests.sh tests/unit`?
(Tests were run but this is my first contribution and the tests are noisy, I suspect they do not relate to this commit)
-----

### DIFF
--- a/extensions/desktop/common/desktop-exports
+++ b/extensions/desktop/common/desktop-exports
@@ -211,8 +211,8 @@ fi
 # Create links for user-dirs.dirs
 if [ "$needs_xdg_links" = true ]; then
   for ((i = 0; i < ${#XDG_SPECIAL_DIRS_PATHS[@]}; i++)); do
-    b="$(realpath "${XDG_SPECIAL_DIRS_PATHS[$i]}" --relative-to="$HOME")"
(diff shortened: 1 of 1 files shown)


MrCarroll:
It appears this PR might not be essential in the Snapcraft Extensions specifically, the exact behaviour differs in the Snapcraft Desktop Helpers where it was tested, but in Gnome-3-38 which I've just been checking now, `realpath` doesn't actually spam the console at all 

I'll leave it open here incase there's still value in checking the variable is not nil, since it could still lead to attempting to `rmdir $HOME`, but otherwise there's some subtle in the desktop helpers that means this might be an unnecessary commit.

MarcusTomlinson (approved):
LGTM

sergiusens (approved):

PR #3452, opened 2021-02-28
storeapi: decouple auth and API

sergiusens:
This changes includes:

- a new Authorization component to generate requests.
- Auth and token discharge logic tied to Authorization class.
- Refresh logic centralized to the request of the Authorization class
  (as it is backend specific).
- Core login primitives simplified in StoreClient and relevant
  delegations to the Authorization class.
- A general simple inheritable requests wrapper for each .*API class
  with an abstractmethod to implement the request logic.

There is one big change in behavior, internally mostly, is that now
all requests to get_account_information are not stored anymore. This
affects a fast whoami call. This will be added back after the new
candid client is introduced.

-----
This is how it looks like now:
image


cjp256 (approved):
Looks good to me, though I'm not sure why the Requests interface exists - is it used anywhere other than an internal detail to the api client?  It seems to conflate the httplib request/session with the API client (and if it exists as an internal detail, get()/put()/post() probably shouldn't be public?)

snapcraft/storeapi/_ubuntu_sso_client.py
```diff
@@ -0,0 +1,259 @@
...
+        self.session.mount("http://", HTTPAdapter(max_retries=retries))
+        self.session.mount("https://", HTTPAdapter(max_retries=retries))
+
+    def request(
+        self, method, url, params=None, headers=None, **kwargs
+    ) -> requests.Response:
+        """Send a request to url relative to the root url.
+
+        :param str method: Method used for the request.
+        :param str url: Appended with the root url first.
```
cmatsuoka, line 49: This parameter description sounds a bit confusing, we should append (prepend?) the root url or it will be done automatically for us?

cmatsuoka (approved):
The code looks good, but I'm not familiar with the APIs involved in this PR (@cjp256 should be able to comment more on that regard).

snapcraft/storeapi/_ubuntu_sso_client.py
sergiusens: This is probably a copy paste error, I will update

sergiusens, line 49:
```suggestion
        :param str url: URL to request with method.
```

PR #3462, opened 2021-03-04
extensions: add core20 support to kde-neon

sergiusens:

-----


jriddell:
I also tend to add these plugs by default:
home
network
network-bind
unity7
pulseaudio
cups-control
removable-media

it might make sense for this extension to add them

ogra1:
> pulseaudio

note that pulseaudio has long been deprecated for ```audio-playback``` and ```audio-record``` you should probably update to use them instead ...

sergiusens:
I will address each of these

> I also tend to add these plugs by default:
> home

We should look into using portals instead, if a particular snap still needs it, they are not prevented from adding it.

> network
> network-bind

Not all snaps need network, this should also be a per snap setting.

> unity7

I don't believe this is needed, replaced with desktop-legacy (@anonymouse64 can you confirm?)

> pulseaudio

We should use audio-record and audio-playback instead, also per snap.

> cups-control

I would make this per snap too, but if we end up adding it, maybe we should for the gnome ones too (@kenvandine what are your thoughts on this?). I also believe cups is moving to some for of portal.

> removable-media

Similar to home, would make sense to use portals by default if possible here.

anonymouse64:
> I don't believe this is needed, replaced with desktop-legacy (@anonymouse64 can you confirm?)

yes unity7 is deprecated in favor of desktop-legacy, though of course if you can get away with not using desktop-legacy that's also ideal, but desktop-legacy is the "newer legacy" interface to use

> cups-control

snaps wanting to print things should probably use `cups` and then connect to the upcoming cups snap (though I will let Ken (and likely Till at the end of the ping chain) comment on when this will be available), but the advantage is that `cups` will auto-connect, while cups-control does not auto-connect by default. `cups` interface does need a rather recent snapd though, 2.47

snapcraft/plugins/v2/cmake.py
```diff
@@ -84,6 +90,7 @@ def _get_cmake_configure_command(self) -> str:
             '"${SNAPCRAFT_PART_SRC_WORK}"',
             "-G",
             f'"{self.options.cmake_generator}"',
+            "${SNAPCRAFT_CMAKE_ARGS}",
```
cjp256, line 93: Can't we just use a `cmake-parameters` in the extension?

sergiusens: The strict goal for extensions is to always extend snapcraft.yaml, if we can eventually pull it off, would be good

cjp256 (approved):

PR #3472, opened 2021-03-10
storeapi: move http client and auth to http_clients package

sergiusens:
Move the existing http clients (UbuntuOneAuthClient and Client)
together with the agent logic into its own http_clients.

Corresponding errors (which still need redoing with a cetralized
strategy around the final request) to this package too.

Additionally, migrate snapcraft.config.Config to this package in the
form of a Abstract Base Class with a specific implementation for the
UbuntuOneAuthClient class.

While migrating config, the "local credentials" implementation was
removed as it was a feature brought in from the click_toolbelt days
and we never really exposed in Snapcraft.

-----

### DIFF
--- a/snapcraft/_store.py
+++ b/snapcraft/_store.py
@@ -190,7 +190,7 @@ def _try_login(
         if not config_fd:
             print()
             echo.wrapped(storeapi.constants.TWO_FACTOR_WARNING)
-    except storeapi.errors.StoreTwoFactorAuthenticationRequired:
+    except storeapi.http_clients.errors.StoreTwoFactorAuthenticationRequired:
         one_time_password = echo.prompt("Second-factor auth")
         store.login(
             email=email,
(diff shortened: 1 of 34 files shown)


cjp256 (approved):
I didn't realize the store was the only user of Config, though makes sense.  Nice cleanup, LGTM!

snapcraft/storeapi/__init__.py
```diff
@@ -19,7 +19,7 @@
 from . import errors  # noqa: F401 isort:skip
 from . import channels  # noqa: F401 isort:skip
 from . import status  # noqa: F401 isort:skip
-
+from . import http_clients  # noqa: F401 isort: skip
```
cmatsuoka, line 22: Any special reason for isort skipping all these?

cmatsuoka (approved):
Didn't dive deep into all the details but it seems to be good.

snapcraft/storeapi/__init__.py
sergiusens: I think it was so that the last import stayed last and avoid some dependency cyclic import issues.
@cjp256 maybe with this change, it can be looked into again?

PR #3487, opened 2021-03-24
deprecations: deprecation notice for core (CRAFT-204)

cjp256:


cjp256:
@sergiusens 
use shorter warning message
add as deprecation notice

cjp256:
This breaks `snapcraft define curl` due the warning on stdout.  We could move the deprecation notices to stderr, but that would not be very consistent with the others.

cmatsuoka (approved):

sergiusens (approved):

PR #3488, opened 2021-03-24
cli, repo: add support for UA tokens

cjp256:


cmatsuoka (approved):
Thanks Chris, lgtm.

sergiusens:
Can you attach a quick spec to this? Same PR is fine

tests/unit/repo/test_ua_manager.py
```diff
@@ -0,0 +1,70 @@
...
+
+from snapcraft.internal.repo import ua_manager
+
+
+def test_ua_manager(fake_process):
+    ua_token = "test-ua-token"
+
+    fake_process.register_subprocess(
+        ["sudo", "ua", "status", "--format", "json"],
+        stdout='{"attached": false, "_doc": "Content provided in json response is currently considered Experimental and may change", "techSupportLevel": "n/a", "origin": null, "expires": "n/a", "services": [{"name": "cc-eal", "description": "Common Criteria EAL2 Provisioning Packages", "available": "no"}, {"name": "esm-infra", "description": "UA Infra: Extended Security Maintenance", "available": "yes"}, {"name": "fips", "description": "NIST-certified FIPS modules", "available": "no"}, {"name": "fips-updates", "description": "Uncertified security updates to FIPS modules", "available": "no"}, {"name": "livepatch", "description": "Canonical Livepatch service", "available": "yes"}]}',
```
facundobatista, line 27: Please move out this BIG line, so each test is not that ugly.

Define only one line outside, with a small `{whatever}` template so then in each test you can actually build the needed one with the bits that change from test to test.

This way also it's more clear to the reader what is changing from test to test.

snapcraft/internal/repo/ua_manager.py
```diff
@@ -0,0 +1,62 @@
...
+def _status() -> Dict[str, Any]:
+    stdout = subprocess.check_output(["sudo", "ua", "status", "--format", "json"])
+    return json.loads(stdout)
+
+
+@contextlib.contextmanager
+def ua_manager(ua_token: Optional[str]):
+    """Attach and detach UA token as required.
+
+    Uses try/finally to ensure a detach on error."""
```
facundobatista, line 51: The `"""` needs to be alone when closing a multiline string.

facundobatista (changes requested):
Great work! I annotated two very small details.

tests/unit/repo/test_ua_manager.py
sergiusens: this could just be a local pytest.fixture that returns a dict that you end up modifying from test to test

snapcraft/internal/repo/ua_manager.py
sergiusens: There was a tool that verified things like this, right? Might be worth a future general PR incorporating it and fixing

facundobatista: the `pep257` tool

cmars (approved):
LGTM (deferring to other reviewers on their concerns of course).

For next cycle we'll spec out designs for how we can provide entitlements more appropriately scoped to CI image builds like snaps and containers. I don't think this implementation would need to change much (if at all); in the ideal case, it would still be a token, with a different scope on the backend.

tests/unit/repo/test_ua_manager.py
cjp256: all i really need is {"attached": false}, the rest of there is just for real data. will break it up or limit it.

snapcraft/internal/repo/ua_manager.py
cjp256: Ah, something didn't feel right when I looked at this last time.  Now I know. Will fix

tests/unit/repo/test_ua_manager.py
cjp256: Fixed

snapcraft/internal/repo/ua_manager.py
cjp256: Fixed

facundobatista (approved):
Awesome, thanks!

PR #3489, opened 2021-03-25
errors: introduce details_from_called_process_error() helper

cjp256:
Many handlers of command errors are doing their own variant of
the same thing.  This provides a standardized helper to use
details to provide command failure details.

### DIFF
--- a/snapcraft/internal/errors.py
+++ b/snapcraft/internal/errors.py
@@ -31,6 +31,56 @@ if TYPE_CHECKING:
 CLEAN_RESOLUTION = "Run `snapcraft clean` and retry build."
 
 
+def details_from_command_error(
+    *,
+    cmd: List[str],
+    returncode: int,
+    stdout: Optional[Union[bytes, str]] = None,
+    stderr: Optional[Union[bytes, str]] = None,
+) -> str:
+    """Create a consistent ProviderError from command errors.
+
+    stdout and stderr, if provided, will be stringified using its object
+    representation.  This method does not decode byte strings.
+
+    :param cmd: Command executed.
+    :param returncode: Command exit code.
+    :param stdout: Optional stdout to include.
+    :param stderr: Optional stderr to include.
+
+    :returns: Details string.
+    """
+    cmd_string = " ".join([shlex.quote(c) for c in cmd])
+
+    details = [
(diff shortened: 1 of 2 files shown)


sergiusens (approved):

snapcraft/internal/errors.py
```diff
@@ -31,6 +31,56 @@
...
 
 
+def details_from_command_error(
+    *,
+    cmd: List[str],
+    returncode: int,
+    stdout: Optional[Union[bytes, str]] = None,
+    stderr: Optional[Union[bytes, str]] = None,
+) -> str:
+    """Create a consistent ProviderError from command errors.
```
cjp256, line 41:
```suggestion
    """Create a consistent details message from command errors.
```

snapcraft/internal/errors.py
```diff
@@ -31,6 +31,56 @@
...
+        details.append(f"* Command output: {stdout!r}")
+
+    if stderr:
+        details.append(f"* Command standard error output: {stderr!r}")
+
+    return "\n".join(details)
+
+
+def details_from_called_process_error(error: CalledProcessError) -> str:
+    """Create a consistent ProviderError from command errors.
```
cjp256, line 70:
```suggestion
    """Create a consistent details message from CalledProcessError.
```

snapcraft/internal/errors.py
```diff
@@ -31,6 +31,56 @@
...
+        f"* Command exit code: {returncode}",
+    ]
+
+    if stdout:
+        details.append(f"* Command output: {stdout!r}")
+
+    if stderr:
+        details.append(f"* Command standard error output: {stderr!r}")
+
+    return "\n".join(details)
```
facundobatista, line 66: This is suspiciously unbalanced: all lines end with a `\n` except the last one. 

How this will be used? It's ok for last line to not have an ENTER?

facundobatista (approved):
Great work!! Approving, but pleas check a question I made inline. 

Thanks!!

snapcraft/internal/errors.py
cjp256: Correct, none of the exception fields have trailing newlines in Snapcraft.  The top level handler will print with the newline.

PR #3497, opened 2021-04-08
repo: fix regression in fix_symlink()

cjp256:
If the file was copied from host, the symlink was not updated
to point to the new file.  It was logged as being copied, but
the copy wasn't utilized without a fixed (relative) symlink.

Add spread tests to cover relative and host copy cases.


cjp256:
Needs tests before merging.

sergiusens (approved):

cmatsuoka (approved):

PR #3501, opened 2021-04-12
store: use whoami dashboard endpoint for cli

sergiusens:
Add the MVP support for the whoami endpoint to the Store APIs under
Dashboard to serve as a replacement for using the account information
endpoint to serve as the backend to the whoami command.

Caching was removed from the command as the existing caching was too
strict (over caching).

"whoami" command test was rewritten in pure pytest.

-----

### DIFF
--- a/snapcraft/storeapi/v2/_api_schema.py
+++ b/snapcraft/storeapi/v2/_api_schema.py
@@ -144,7 +144,6 @@ RELEASES_JSONSCHEMA: Dict[str, Any] = {
 
 
 CHANNEL_MAP_JSONSCHEMA: Dict[str, Any] = {
-    "additionalProperties": False,
     "properties": {
         "channel-map": {
             "items": {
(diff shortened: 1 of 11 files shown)


snapcraft/storeapi/v2/_api_schema.py
```diff
@@ -339,3 +338,130 @@
...
+                    "items": {"description": "Valid channel name.", "type": "string"},
+                    "minItems": 1,
+                    "type": "array",
+                    "uniqueItems": True,
+                },
+                {"type": "null"},
+            ]
+        },
+        "errors": {
+            "items": {
```
cjp256, line 380: Interesting that errors is not allowed to be null.  Is there any scenario where we should be checkin this field for errors?

snapcraft/storeapi/v2/whoami.py
```diff
@@ -0,0 +1,85 @@
...
+
+class WhoAmI:
+    """Represent the data returned from the whoami Snap Store endpoint."""
+
+    @classmethod
+    def unmarshal(cls, payload: Dict[str, Any]) -> "WhoAmI":
+        jsonschema.validate(payload, WHOAMI_JSONSCHEMA)
+        import pprint
+
+        pprint.pprint(payload)
```
cjp256, line 73: Probably meant for removal? :P

cjp256 (approved):
LGTM with pprint fix

snapcraft/storeapi/v2/whoami.py
sergiusens: yes, and done

snapcraft/storeapi/v2/_api_schema.py
sergiusens: yes, we have some weird logic today for when you can login but cannot do anything until you sign the developer agreement or setup a username on the store side. This would allow for replacing that custom logic

cmatsuoka (approved):
+1

PR #3502, opened 2021-04-13
snaps: do not validate snaps before install/refresh (Fixes LP#1901733)

Saviq:
Branches are not advertised in store channels, the only way to check is to
to check with the store API, which `install` and `refresh` do already.

Fixes https://bugs.launchpad.net/snapcraft/+bug/1901733

### DIFF
--- a/snapcraft/internal/repo/snaps.py
+++ b/snapcraft/internal/repo/snaps.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright (C) 2017-2020 Canonical Ltd
+# Copyright (C) 2017-2021 Canonical Ltd
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License version 3 as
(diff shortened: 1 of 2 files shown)


snapcraft/internal/repo/snaps.py
```diff
@@ -167,6 +167,8 @@ def is_classic(self) -> bool:
 
     def is_valid(self) -> bool:
         """Check if the snap is valid."""
+        if self.installed and self.get_local_snap_info()["channel"] == self.channel:
```
sergiusens, line 170: the caveat here is that for this to work it has to be installed, which implies destructive-mode, correct?

Saviq: Or a preexisting environment?

sergiusens: yes, I was thinking about the bug state after this lands as it alleviates the problem but does not completely fix it, that's all :-)

sergiusens (approved):

cmatsuoka (approved):
Thanks Michał, looks good.

PR #3299, opened 2020-09-25
extensions: configure hook for fonts

sergiusens:
Signed-off-by: Sergio Schvezov <sergio.schvezov@canonical.com>

- [ ] Have you followed the [guidelines for contributing](https://github.com/snapcore/snapcraft/blob/master/CONTRIBUTING.md)?
- [ ] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [ ] Have you successfully run `./runtests.sh static`?
- [ ] Have you successfully run `./runtests.sh tests/unit`?

-----

PR #3504, opened 2021-04-19
extensions/desktop: use fonts from $XDG_DATA_DIRS, and remove unnecessary includes

jhenstridge:

-----
This PR cleans up the generation of the user `fonts.conf` file in the desktop extension, partially reverting some of the changes in PR #3299.  Namely:

1. Use fonts from $XDG_DATA_DIRS.  In particular, this means the small number of fonts provided by the platform snap will be visible to applications.  This avoids malfunctions when no fonts are available.
2. Don't include $FONTCONFIG_FILE, since this represents an include loop.

For (1), this has only really manifested in trying to run snaps on Ubuntu Core systems where there is no `desktop` slot available to share system fonts.

For (2), it seems to be a misunderstanding of what the `$XDG_CONFIG_HOME/fontconfig/fonts.conf` file is for.  It isn't used in place of the top level `fonts.conf` file (either identified by `$FONTCONFIG_FILE` or `/etc/fonts/fonts.conf`), but instead is included by the top level config file (via `conf.d/50-user.conf`).  Deleting the spurious `<include>` directives from a configuration file generated by `desktop-launch` does not change the list of fonts reported by `fc-list` inside the sandbox.

### DIFF
--- a/extensions/desktop/common/desktop-exports
+++ b/extensions/desktop/common/desktop-exports
@@ -301,9 +301,8 @@ function make_user_fontconfig {
       echo "  <dir>${data_dirs_array[$i]}/fonts</dir>"
     fi
   done
-  # We need to include this default cachedir first so that caching
-  # works: without it, fontconfig will try to write to the real user home
-  # cachedir and be blocked by AppArmor.
+  # We need to include a user-writable cachedir first to support
+  # caching of per-user fonts.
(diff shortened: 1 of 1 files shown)


sergiusens (approved):
TIL, thanks

cjp256 (approved):

kenvandine:
This looks great, thanks!

PR #3510, opened 2021-04-22
cli: introduce list-validation-sets

sergiusens:

-----

### DIFF
--- a/snapcraft/cli/assertions.py
+++ b/snapcraft/cli/assertions.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright (C) 2017 Canonical Ltd
+# Copyright (C) 2017-2021 Canonical Ltd
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License version 3 as
@@ -21,8 +21,10 @@ from textwrap import dedent
 from typing import Dict, List
 
 import click
+from tabulate import tabulate
 
 import snapcraft
+from snapcraft._store import StoreClientCLI
 from snapcraft import storeapi, yaml_utils
 from snapcraft.storeapi import assertions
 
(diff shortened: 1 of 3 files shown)


snapcraft/cli/assertions.py
```diff
@@ -151,6 +153,45 @@ def edit_collaborators(snap_name, key_name):
             echo.warning("The collaborators for this snap have not been altered.")
 
 
+@assertionscli.command("list-validation-sets")
+@click.option("--name", metavar="<name>", help="Name filter.")
+@click.option("--sequence", metavar="<sequence>", help="Sequence filter.")
+def list_validation_sets(name, sequence):
+    """Get the list of validation sets.
+
+    Refer to https://snapcraft.io/docs/validation-sets for further information
```
sergiusens, line 162: fyi @degville

snapcraft/cli/assertions.py
```diff
@@ -151,6 +153,45 @@ def edit_collaborators(snap_name, key_name):
             echo.warning("The collaborators for this snap have not been altered.")
 
 
+@assertionscli.command("list-validation-sets")
+@click.option("--name", metavar="<name>", help="Name filter.")
+@click.option("--sequence", metavar="<sequence>", help="Sequence filter.")
```
cmatsuoka, line 158: "Filter" suggests we could use something like regexp, glob or ranges (for numeric), but it seems that the dashboard API requires exact matches? Maybe describe as "Validation set name", "Sequence number" instead?

cmatsuoka (approved):
Looks good, left a comment on cli argument descriptions.

snapcraft/cli/assertions.py
sergiusens: Sounds good

snapcraft/cli/assertions.py
```diff
@@ -151,6 +153,57 @@ def edit_collaborators(snap_name, key_name):
...
+@click.option("--sequence", metavar="<sequence>", help="Sequences to show.")
+def list_validation_sets(name, sequence):
+    """Get the list of validation sets.
+
+    The sequence option can be a sequence number or a keyword.
+
+    \b
+    Examples:
+        snapcraft list-validation-sets
+        snapcraft list-validation-sets --sequence all
```
cjp256, line 171: Is `--sequence all` the same thing as no --sequence? Seems to be a bit odd. 

Not documented at below link either.

cjp256 (approved):

tests/unit/commands/test_list_validation_sets.py
```diff
@@ -0,0 +1,145 @@
...
+    assert result.output == dedent(
+        """\
+        Account-ID                        Name              Sequence    Revision    When
+        AccountIDXXXOfTheRequestingUserX  certification-x1  9           222         2020-10-29
+        AccountIDXXXOfTheRequestingUserX  acme-qa           2           2           2020-10-29
+        """
+    )
+    fake_dashboard_get_validation_sets.assert_called_once_with(
+        name=None, sequence=None,
+    )
```
cjp256, line 145: Maybe add one more test for sequence=2 filter, or similar?

snapcraft/cli/assertions.py
sergiusens: Should be in below link as it is a store side thing and might evolve, @degville can we add "all" and "latest" (current default) to the docs?

tests/unit/commands/test_list_validation_sets.py
sergiusens: I did that in the previous PR, this will only make its way to the mock (we pass on the string to the store as it holds the rule engine)

sergiusens:
Staging Store seems to be down for uploads, tried locally with known working versions of snapcraft and see the same effect.

snapcraft/cli/assertions.py
degville: Yes, of course. I'll do that now - thanks!

PR #3511, opened 2021-04-23
add V2 Ruby plugin

paddor:
Referring to https://forum.snapcraft.io/t/local-v2-plugins/23903

### DIFF
--- a/snapcraft/plugins/v2/ruby.py
+++ b/snapcraft/plugins/v2/ruby.py
@@ -1,0 +1,156 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2021 Patrik Wenger <paddor@gmail.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""The ruby plugin is useful for ruby based parts.
+It uses ruby-install to install Ruby and can use bundler to install
+dependencies from a `Gemfile` found in the sources.
+
+Additionally, this plugin uses the following plugin-specific keywords:
+    - ruby-version:
+      (string)
+      The version of ruby you want this snap to run. (e.g. '3.0' or '2.7.2')
+    - ruby-flavor:
+      (string)
+      Other flavors of ruby supported by ruby-install (e.g. 'jruby', ...)
+    - gems:
+      (list)
+      A list of gems to install.
+    - use-bundler
+      (boolean)
+      Use bundler to install gems from a Gemfile (defaults 'false').
+    - use-jemalloc:
+      (boolean)
+      Build ruby with libjemalloc (defaults 'false').
+    - shared:
+      (boolean)
+      Build ruby as a shared library (defaults 'false').
+    - configure-options:
+      (array of strings)
+      Additional configure options to use when configuring ruby.
+"""
+
+import os
+import re
+import logging
+
+from typing import Any, Dict, List, Set
+from snapcraft.plugins.v2 import PluginV2
+
+
+class RubyPlugin(PluginV2):
+    @classmethod
+    def get_schema(cls) -> Dict[str, Any]:
+        return {
+            "$schema": "http://json-schema.org/draft-04/schema#",
+            "type": "object",
+            "additionalProperties": False,
+            "properties": {
+                "ruby-flavor": {
+                    "type": "string",
+                    "default": "ruby",
+                },
+                "ruby-version": {
+                    "type": "string",
+                    "default": "3.0",
+                    "pattern": r"^\d+\.\d+(\.\d+)?$",
+                },
+                "use-bundler": {
+                    "type": "boolean",
+                    "default": False,
+                },
+                "use-jemalloc": {
+                    "type": "boolean",
+                    "default": False,
+                },
+                "shared": {
+                    "type": "boolean",
+                    "default": False,
+                },
+                "configure-options": {
+                    "type": "array",
+                    "items": {"type": "string"},
+                    "default": [],
+                },
+                "gems": {
+                    "type": "array",
+                    "items": {"type": "string"},
+                    "default": [],
+                },
+            },
+        }
+
+    def get_build_snaps(self) -> Set[str]:
+        return set()
+
+    def get_build_packages(self) -> Set[str]:
+        packages = {"curl", "jq"}
+
+        if self.options.use_jemalloc:
+            packages.add("libjemalloc-dev")
+
+        return packages
+
+    def get_build_environment(self) -> Dict[str, str]:
+        env = {
+            "PATH": "${SNAPCRAFT_PART_INSTALL}/bin:${PATH}",
+        }
+
+        if self.options.shared:
+            # for finding ruby.so when running `gem` or `bundle`
+            env["LD_LIBRARY_PATH"] = "${SNAPCRAFT_PART_INSTALL}/lib:${LD_LIBRARY_PATH}"
+
+        return env
+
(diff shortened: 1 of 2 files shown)


sergiusens:
Thanks for collaborating on this. Looks good, I have a couple of suggestions to make (aside from testing) that I will propose as soon as I have some time available

snapcraft/plugins/v2/ruby.py
```diff
@@ -0,0 +1,156 @@
...
+        return packages
+
+    def get_build_environment(self) -> Dict[str, str]:
+        env = {
+            "PATH": "${SNAPCRAFT_PART_INSTALL}/bin:${PATH}",
+        }
+
+        if self.options.shared:
+            # for finding ruby.so when running `gem` or `bundle`
+            env["LD_LIBRARY_PATH"] = "${SNAPCRAFT_PART_INSTALL}/lib:${LD_LIBRARY_PATH}"
```
cjp256, line 113: To expand correctly if LD_LIBRARY_PATH is empty:
```suggestion
            env["LD_LIBRARY_PATH"] = "${SNAPCRAFT_PART_INSTALL}/lib:${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}"
```

snapcraft/plugins/v2/ruby.py
```diff
@@ -0,0 +1,156 @@
...
+        ] + self.options.configure_options
+
+        if self.options.shared:
+            configure_opts.append("--enable-shared")
+        if self.options.use_jemalloc:
+            configure_opts.append("--with-jemalloc")
+
+        return configure_opts
+
+    def _get_install_commands(self) -> List[str]:
```
cjp256, line 131: Might as well combine these two since there's no meaningful extras:
```suggestion
    def get_build_commands(self) -> List[str]:
```

snapcraft/plugins/v2/ruby.py
```diff
@@ -0,0 +1,156 @@
...
+        commands = []
+        commands.append("ruby_install_url=$(curl -L --proto '=https' --tlsv1.2 'https://api.github.com/repos/postmodern/ruby-install/tags' | jq -r '.[0].tarball_url')")
+        commands.append("curl -L --proto '=https' --tlsv1.2 $ruby_install_url | tar xz")
```
cjp256, line 134: Is there anything we can do to authenticate this installer? signature perhaps?  Or maybe pin to a version via properties with an associated hash?

cjp256 (changes requested):
Looks pretty good, have a few suggestions - thanks!

snapcraft/plugins/v2/ruby.py
paddor: 👍

snapcraft/plugins/v2/ruby.py
paddor: 👍

snapcraft/plugins/v2/ruby.py
```diff
@@ -0,0 +1,156 @@
...
+                "ruby-flavor": {
+                    "type": "string",
+                    "default": "ruby",
+                },
+                "ruby-version": {
+                    "type": "string",
+                    "default": "3.0",
+                    "pattern": r"^\d+\.\d+(\.\d+)?$",
+                },
+                "use-bundler": {
```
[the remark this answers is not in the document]
paddor: 👍

snapcraft/plugins/v2/ruby.py
paddor: That would be nice. I just checked the existing plugins. Looks like only V1 plugins use `sources.Tar` with the argument `source_checksum`. Can I do the same in this V2 plugin?

paddor: @cjp256

snapcraft/plugins/v2/ruby.py
```diff
@@ -0,0 +1,160 @@
...
+        packages = {"curl", "jq"}
+
+        if self.options.ruby_use_jemalloc:
+            packages.add("libjemalloc-dev")
+
+        return packages
+
+    def get_build_environment(self) -> Dict[str, str]:
+        env = {
+            "PATH": "${{SNAPCRAFT_PART_INSTALL}}{}/bin:${PATH}".format(self.options.ruby_prefix),
```
sergiusens, line 115:
```suggestion
            "PATH": f"${{SNAPCRAFT_PART_INSTALL}}{self.options.ruby_prefix}/bin:${PATH}",
```

snapcraft/plugins/v2/ruby.py
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        if self.options.ruby_shared:
+            # for finding ruby.so when running `gem` or `bundle`
+            env["LD_LIBRARY_PATH"] = "${{SNAPCRAFT_PART_INSTALL}}{}/lib${{LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}}".format(
+                    self.options.ruby_prefix)
```
sergiusens, line 121:
```suggestion
            env["LD_LIBRARY_PATH"] = f"${{SNAPCRAFT_PART_INSTALL}}{self.options.ruby_prefix}/lib${{LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}}"
```

snapcraft/plugins/v2/ruby.py
```diff
@@ -0,0 +1,160 @@
...
+        commands.append("curl -L --proto '=https' --tlsv1.2 $ruby_install_url | tar xz")
+        commands.append("postmodern-ruby-install-*/bin/ruby-install -i ${{SNAPCRAFT_PART_INSTALL}}{} --package-manager apt -j${{SNAPCRAFT_PARALLEL_BUILD_COUNT}} {}-{} -- {}".format(
+            self.options.ruby_prefix,
+            self.options.ruby_flavor,
+            self.options.ruby_version,
+            ' '.join(self._configure_opts())))
+
+        # NOTE: Update bundler. Avoid conflicts/prompts about replacing bundler
+        #       executables by removing them first.
+        commands.append("rm -f ${{SNAPCRAFT_PART_INSTALL}}{}/bin/{{bundle,bundler}}".format(self.options.ruby_prefix))
```
sergiusens, line 151:
```suggestion
        commands.append(f"rm -f ${{SNAPCRAFT_PART_INSTALL}}{self.options.ruby_prefix}/bin/{{bundle,bundler}}")
```

sergiusens (commented):
How complicated would it be to have ruby as a content snap and deliver the entire hookup through an extension?

sergiusens:
And apologies for the late response, I had an intense event and a full disconnect afterwards.

paddor:
> How complicated would it be to have ruby as a content snap and deliver the entire hookup through an extension?

I'm sorry, but I don't understand the question. Content snap? Hookup?

What I'm already doing in a project is using this plugin in a part of its own without a `source:`-line, so Ruby doesn't have to be recompiled after every change in the project source code. Is that what you mean?

snapcraft/plugins/v2/ruby.py
paddor: Solved with [984430c](https://github.com/snapcore/snapcraft/pull/3511/commits/984430c623f792434904e5f1635008a4829c9a5f)

paddor:
Please review.

mr-cal (changes requested):
Hi @paddor,

Apologies for the long delay in reviews.  I'm happy to help you drive this and https://github.com/canonical/craft-parts/pull/341 to completion.

I think next step is an integration test that builds a "hello world" Ruby app.  You would need to add the test [here](https://github.com/snapcore/snapcraft/blob/1e2984c454f044b179fb5232b212e2feca6f838c/tests/spread/plugins/v2/build-and-run-hello/task.yaml#L5) and the source files for the app in a directory [here](https://github.com/snapcore/snapcraft/tree/1e2984c454f044b179fb5232b212e2feca6f838c/tests/spread/plugins/v2/snaps).  Testing this requires installing [spread](https://github.com/canonical/craft-parts/tree/main/tests/integration/plugins) and running it locally with Multipass. 

Alternatively, you could add the "hello world" integration test in craft-parts [here](https://github.com/canonical/craft-parts/tree/8fea6453667863903fc6d452c6b9b3d8c46b17c9/tests/integration/plugins) (which doesn't use spread) and I can help translate it to a spread test in this PR.

mr-cal:
Hi @paddor,

I'm closing this because the PR is stale.  If you want to pick it back up, please create a new PR.  Thanks!

PR #3514, opened 2021-04-27
elf: add fallback methods to detect library dependencies

cjp256:
libc6 may mismatch with the paths provided to ldd in LD_LIBRARY_PATH.

This may be avoided if invoking the binary directory and using
LD_TRACE_LOADED_OBJECTS, but this method is limited to executable
files.

This updates the library dependency search to try three variants:
(1) use ldd as today.  if ldd fails..
(2) use ldd with LD_PRELOAD to specify libc6 from host. if ldd fails...
(3) use ld trace method. if ldd trace fails...
(4) warn as before.

For #2, libc6 is expected as /lib/<arch-triplet>/libc.so.6.  This likely
will need to be more robust in the future.

-----

### DIFF
--- a/snapcraft/internal/elf.py
+++ b/snapcraft/internal/elf.py
@@ -14,6 +14,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 import contextlib
+import functools
 import glob
 import logging
 import os
@@ -31,11 +32,14 @@ from pkg_resources import parse_version
 
 from snapcraft import file_utils
 from snapcraft.internal import common, errors, repo
+from snapcraft.project._project_options import ProjectOptions
 
 logger = logging.getLogger(__name__)
 
 
-def _get_host_libc_path(arch_triplet: str) -> pathlib.Path:
+@functools.lru_cache()
+def _get_host_libc_path() -> pathlib.Path:
+    arch_triplet = ProjectOptions().arch_triplet
     return pathlib.Path("/lib") / arch_triplet / "libc.so.6"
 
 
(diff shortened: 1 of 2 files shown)


snapcraft/internal/elf.py
```diff
@@ -51,33 +58,33 @@ def _ldd_resolve(soname: str, soname_path: str) -> Tuple[str, str]:
...
-    Returns a dictionary of mappings of soname -> soname_path.
-    If library is not resolved, the soname itself is the soname_path.
-    """
+    debug_cmd = [f"{k}={v}" for k, v in extra_env.items()]
+    debug_cmd += cmd
 
-    libraries: Dict[str, str] = dict()
+    logger.debug("executing: %s", " ".join(debug_cmd))
+    output = subprocess.check_output(cmd, env=env).decode()
+    # logger.debug("output: %s", output)
```
sergiusens, line 70: extra line

cjp256: fixed

snapcraft/internal/elf.py
```diff
@@ -34,6 +36,11 @@
 logger = logging.getLogger(__name__)
 
 
+@functools.lru_cache()
+def _get_host_libc_path(arch_triplet: str) -> pathlib.Path:
+    return pathlib.Path("/lib") / arch_triplet / "libc.so.6"
```
facundobatista, line 39: The only place where this is called, it does `str()` around the result.

So in reality, you could solve this with just an `os.path.join` without going back and forward with `pathlib`. 

Furthermore, probably this `os.path.join` could just happen where this function is called.

facundobatista (approved):
+1, please check the comment.

snapcraft/internal/elf.py
cjp256: I 100% agree, but we've been standardizing on using pathlib and converting to string as necessary. For that reason I plan to leave it as-is unless there's an issue.

cmatsuoka (approved):
I don't see obvious problems, but this kind of operation is tricky and something may be hiding in a dark corner.

cjp256:
> I don't see obvious problems, but this kind of operation is tricky and something may be hiding in a dark corner.

lol that's for sure. thanks for reviewing :D

cjp256:
FWIW i was going to add a spread test, but the existing classic-rpath test should cover the same scenario.

snapcraft/internal/elf.py
cjp256: And now I'm using is_file() :D

PR #3516, opened 2021-04-28
github: update workflows that to use pull_request_target

sergiusens:
This would force using workflows from master/main on forks, which
allows using secrets. Exposure is reduced as the workflow is executed
from a already commited code.

-----

https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,7 +1,7 @@
 name: CI
 
-on: 
-  pull_request:
+on:
+    - pull_request_target
 
 jobs:
   publish:
--- a/.github/workflows/spread.yml
+++ b/.github/workflows/spread.yml
@@ -1,6 +1,7 @@
 name: Spread Tests
 
-on: [pull_request]
+on:
+    - pull_request_target
 
 jobs:
   build:


Saviq (changes requested):
This won't do what you want it to:
- by [events that trigger workflows](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target):
  > This event runs in the context of the base of the pull request, rather than in the merge commit as the pull_request event does.
- _if_ you explicitly checked the merge ref, you'd be exposed to malicious code (even if not malicious workflows) in PRs from forks.
  > This prevents executing unsafe workflow code from the head of the pull request that could alter your repository or steal any secrets you use in your workflow.

You'd need to check that the originating fork is trusted (from a trusted user?) before you check it out.

jhenstridge:
A few issues:

1. This won't actually test the code in the PR, since with pull_request_target, the default for actions/checkout@v2 is the target branch (which is considered safe).
2. If you updated the code to check out `github.event.pull_request.head.sha` (as described in the "INSECURE. Provided as an example only" example [here](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/), it wouldn't protect the secrets from an attacker.  In the "Run spread" step, it puts the secrets in the environment and calls spread.  This will use the attacker supplied `spread.yaml`, that can execute arbitrary code.
3. As a trusted workflow, the `GITHUB_TOKEN` for the workflow will have write permissions to the repository. With the arbitrary code execution noted above, this could allow a PR author to push changes directly to the main repository.

As Saviq said, this kind of change would only make sense if the workflow only runs for trusted users (and that the decision is preferably made before checking out the PR's code.

jhenstridge:
Also, rather than `github.event.pull_request.head.sha`, it should probably be `github.event.pull_request.merge_commit_sha` (i.e. a commit representing the result of merging the PR into the target).

PR #3518, opened 2021-05-04
yaml: detect and log warning on duplicate keys in snapcraft.yaml (CRAFT-553)

cjp256:
A very primitive check to warn if duplicate keys are found in
the snapcraft.yaml.  At the moment the file is loaded multiple
times which may result in duplicate warnings (ha ha) which we
need to address in future work.  Given that there shouldn't be
duplicates in YAML files, the extra verbosity should be OK.

-----

### DIFF
--- a/snapcraft/project/_project_info.py
+++ b/snapcraft/project/_project_info.py
@@ -27,7 +27,9 @@ class ProjectInfo:
 
     def __init__(self, *, snapcraft_yaml_file_path) -> None:
         self.snapcraft_yaml_file_path = snapcraft_yaml_file_path
-        self.__raw_snapcraft = yaml_utils.load_yaml_file(snapcraft_yaml_file_path)
+        self.__raw_snapcraft = yaml_utils.load_yaml_file(
+            snapcraft_yaml_file_path, warn_duplicate_keys=True
+        )
 
         try:
             self.name = self.__raw_snapcraft["name"]
--- a/snapcraft/project/_project_info.py
+++ b/snapcraft/project/_project_info.py
@@ -27,9 +27,7 @@ class ProjectInfo:
 
     def __init__(self, *, snapcraft_yaml_file_path) -> None:
         self.snapcraft_yaml_file_path = snapcraft_yaml_file_path
-        self.__raw_snapcraft = yaml_utils.load_yaml_file(
-            snapcraft_yaml_file_path, warn_duplicate_keys=True
-        )
+        self.__raw_snapcraft = yaml_utils.load_yaml_file(snapcraft_yaml_file_path)
 
         try:
             self.name = self.__raw_snapcraft["name"]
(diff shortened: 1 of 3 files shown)


cjp256:
Re-opened from #3517 to be on origin for edge publishing.

snapcraft/project/_project_info.py
```diff
@@ -27,7 +27,9 @@ class ProjectInfo:
 
     def __init__(self, *, snapcraft_yaml_file_path) -> None:
         self.snapcraft_yaml_file_path = snapcraft_yaml_file_path
-        self.__raw_snapcraft = yaml_utils.load_yaml_file(snapcraft_yaml_file_path)
+        self.__raw_snapcraft = yaml_utils.load_yaml_file(
+            snapcraft_yaml_file_path, warn_duplicate_keys=True
```
cmatsuoka, line 31: Is there a situation where we allow duplicate keys?

tests/unit/yaml_utils/test_yaml_utils.py
```diff
@@ -15,39 +15,48 @@
...
-class OctIntTest(unit.TestCase):
-    def test_octint_dump(self):
-        output = io.StringIO()
-        yaml_utils.dump(dict(number=yaml_utils.OctInt(8)), stream=output)
-        output.seek(0)
+    assert [r.message for r in caplog.records] == []
 
-        self.assertThat(output.read().strip(), Equals("number: 0010"))
+
+def test_load_yaml_file_with_duplicate_warnings(caplog, tmp_path):
```
sergiusens, line 45: can you add a nest with nest duplicates?

tests/unit/yaml_utils/test_yaml_utils.py
```diff
@@ -15,39 +15,48 @@
...
+    yaml_path = tmp_path / "test.yaml"
+    yaml_path.write_text("foo: bar\nfoo: bar2")
+
+    cfg = yaml_utils.load_yaml_file(str(yaml_path), warn_duplicate_keys=True)
```
cmatsuoka, line 49: If we make this configurable, we should also test for the situation where it's set to False.

snapcraft/project/_project_info.py
cmatsuoka: Removing this argument, warnings will be always enabled.

tests/unit/yaml_utils/test_yaml_utils.py
cmatsuoka: Adding a test with the following scenario, is this what you mean? (In this case outer `foo` must not clash with inner `foo`, and inner `baz` must be detected)
```
foo: bar
nest:
  foo: bar2
  baz: bar3
  baz: bar4
```

tests/unit/yaml_utils/test_yaml_utils.py
cmatsuoka: Removed configuration and leave it always enabled.

cmatsuoka:
Rebasing on master because this branch is very old, merging master would make the commits hard to follow.

cmatsuoka:
Original implementation only warned about the first duplicated key. Changed to issue warnings for all duplicated keys in the yaml file.

cjp256:
👻

PR #3521, opened 2021-05-13
build providers: set hostname for lxd

cjp256:
The hostname is being written to /etc/hostname, but the kernel
is not being informed of the update.  Use hostname -F to set it
after writing.

-----


cjp256:
Reported at: https://forum.snapcraft.io/t/snapcraft-under-lxd-very-slow-sudo-hostname-problem/24308/2

tests/unit/build_providers/lxd/test_lxd.py
```diff
@@ -246,6 +245,7 @@ def test_create(self):
         ]
 
         assert check_output_calls == [
+            ["hostname", "-F", "/etc/hostname"],
```
cmatsuoka, line 248: I think we should have a test that verifies if the hostname is actually being set inside the container.

sergiusens: We can use the already existing --use-lxd test and have an override-<something> perform a check

cjp256: The existing test uses make-hello, so I added a new spread test.

sergiusens (approved):

cmatsuoka (approved):

PR #3522, opened 2021-05-13
python v2 plugin: update pip properly

MrCarroll:
Prior to this commit, pip wouldn't be updated properly.
With this commit, pip wheel and setuptools will be updated automatically,
this behaviour can be turned off with the 'update-pip: false' attribute,
in which case the original pip (e.g, from the base) will be used.

- [Y] Have you successfully run `./runtests.sh tests/unit`?

-----

### DIFF
--- a/snapcraft/plugins/v2/python.py
+++ b/snapcraft/plugins/v2/python.py
@@ -38,8 +38,12 @@ Additionally, this plugin uses the following plugin-specific keywords:
 
     - python-packages
       (list)
-      A list of dependencies to get from PyPI. If needed, pip,
-      setuptools and wheel can be upgraded here.
+      A list of dependencies to get from PyPI.
+
+    - update-pip
+      (boolean)
+      If set to true, build with the latest versions
+      of pip, wheel, and setuptools.
 
 This plugin also interprets these specific build-environment entries:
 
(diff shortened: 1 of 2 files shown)


sergiusens:
Thanks for the contribution, at this point in time, this would be a breaking change. I think that what we have today is already somewhat acceptable since:

- python-packages can be used to bootstrap pre-dependencies (the docs could be clearer)
- python packages allows you to lock down to a specific dependency or just use the archive provided tooling by making this an empty list (again, doc update)
- most python packages have a install_requires and requirements.txt to install dependencies, requirements.txt can even be a remote url if required.

MrCarroll:
Would this contribution be acceptable if the default for `update-pip` was false and the defaults for `python-packages` reverted? This would retain the behaviour with existing build files but provide a way for people like myself to update pip properly in the situation only the `python-packages` attribute is used. The worst that would happen is an insignificant delay as pip might be checked for updates twice.

sergiusens:
is there a reason you cannot use a requirements file? I am hesitant on having N ways to reach the same goal. update-pip should really be 3 (wheel and setuptools) and it can also lead to N which can be problematic

MrCarroll:
No, I can use a requirements file but I'd have preferred to keep the package list explicitly in the snapcraft.yaml, mostly to try keep it as easy as possible for anyone wishing to create their own versions, though admittedly this isn't common in the overall snap ecosystem.

Thanks for the feedback, I'll close this PR and mark the thread on the forums as solved.

PR #3528, opened 2021-05-25
apt cache: improve error handling when packages do not have candidates available (CRAFT-136) (CRAFT-218)

cjp256:
(1) Add some checks to ensure package.candidate is not None before using
it.  If it is None, when expected to not be None, raise a
PackageNotFoundError with the name of the package in question.

(2) Do not iterate over all packages when unmarking.  This is the cause of
https://forum.snapcraft.io/t/snapcraft-crash-when-esm-repositories-are-present/24587/8
where a package was being inspected that did not have an available candidate,
causing Snapcraft to fail unnecessarily.

Fixes SNAPCRAFT-2NE
Fixes SNAPCRAFT-2H2
Fixes SNAPCRAFT-2BC
Fixes SNAPCRAFT-277
Fixes SNAPCRAFT-263
Fixes SNAPCRAFT-1R4
Fixes SNAPCRAFT-1Q5
Fixes SNAPCRAFT-1DH
    
Fixes a comment in unrelated bug report: LP: #1853682

-----

### DIFF
--- a/snapcraft/internal/repo/apt_cache.py
+++ b/snapcraft/internal/repo/apt_cache.py
@@ -212,17 +212,18 @@ class AptCache(ContextDecorator):
         return installed
 
     def get_packages_marked_for_installation(self) -> List[Tuple[str, str]]:
-        packages: List[Tuple[str, str]] = []
-        for package in self.cache.get_changes():
-            if not package.marked_install:
-                continue
+        changed_packages = self.cache.get_changes()
+        marked_install_packages = [p for p in changed_packages if p.marked_install]
+        package_names_missing_installation_candidate = [
+            p.name for p in marked_install_packages if p.candidate is None
+        ]
 
-            if package.candidate is None:
-                raise errors.PackageNotFoundError(package.name)
+        if package_names_missing_installation_candidate:
+            raise errors.PackagesNotFoundError(
+                package_names_missing_installation_candidate
+            )
 
-            packages.append((package.name, package.candidate.version))
-
-        return packages
+        return [(p.name, p.candidate.version) for p in marked_install_packages]
 
     def mark_packages(self, package_names: Set[str]) -> None:
         for name in package_names:
--- a/snapcraft/internal/repo/apt_cache.py
+++ b/snapcraft/internal/repo/apt_cache.py
@@ -158,6 +158,10 @@ class AptCache(ContextDecorator):
     def _verify_marked_install(self, package: apt.Package):
         if not package.installed and not package.marked_install:
             broken_deps: List[str] = list()
+
+            if package.candidate is None:
+                raise errors.PackageNotFoundError(package.name)
+
             for package_dependencies in package.candidate.dependencies:
                 for dep in package_dependencies:
                     if not dep.target_versions:
@@ -194,6 +198,9 @@ class AptCache(ContextDecorator):
             except apt.package.FetchError as e:
                 raise errors.PackageFetchError(str(e))
 
+            if package.candidate is None:
+                raise errors.PackageNotFoundError(package.name)
+
             downloaded.append((package.name, package.candidate.version, Path(dl_path)))
         return downloaded
 
@@ -205,11 +212,17 @@ class AptCache(ContextDecorator):
         return installed
 
     def get_packages_marked_for_installation(self) -> List[Tuple[str, str]]:
-        return [
-            (package.name, package.candidate.version)
-            for package in self.cache.get_changes()
-            if package.marked_install
-        ]
+        packages: List[Tuple[str, str]] = []
+        for package in self.cache.get_changes():
+            if not package.marked_install:
+                continue
+
+            if package.candidate is None:
+                raise errors.PackageNotFoundError(package.name)
+
+            packages.append((package.name, package.candidate.version))
+
+        return packages
 
     def mark_packages(self, package_names: Set[str]) -> None:
         for name in package_names:
@@ -253,6 +266,9 @@ class AptCache(ContextDecorator):
         skipped_filtered = set()
 
         for package in self.cache:
+            if package.candidate is None:
+                raise errors.PackageNotFoundError(package.name)
+
             if package.candidate.priority == "essential":
                 # Filter 'essential' packages.
                 skipped_essential.add(package.name)
--- a/snapcraft/internal/repo/apt_cache.py
+++ b/snapcraft/internal/repo/apt_cache.py
@@ -265,7 +265,7 @@ class AptCache(ContextDecorator):
         skipped_essential = set()
         skipped_filtered = set()
 
-        for package in self.cache:
+        for package in self.cache.get_changes():
             if package.candidate is None:
                 raise errors.PackageNotFoundError(package.name)
 
--- a/snapcraft/internal/repo/errors.py
+++ b/snapcraft/internal/repo/errors.py
@@ -16,6 +16,7 @@
 
 from typing import List, Optional, Sequence
 
+from snapcraft import formatting_utils
 from snapcraft.internal import errors
 from snapcraft.internal.errors import SnapcraftException
 from snapcraft.internal.os_release import OsRelease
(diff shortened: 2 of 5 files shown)


snapcraft/internal/repo/apt_cache.py
```diff
@@ -205,11 +212,17 @@ def get_installed_packages(self) -> Dict[str, str]:
...
-        ]
+        packages: List[Tuple[str, str]] = []
+        for package in self.cache.get_changes():
+            if not package.marked_install:
+                continue
+
+            if package.candidate is None:
+                raise errors.PackageNotFoundError(package.name)
+
+            packages.append((package.name, package.candidate.version))
```
sergiusens, line 223: Would this be a bit more descriptive?
```suggestion
        changed_packages = self.cache.get_changes()
        marked_install_packages = [p for p in changed_packages if p.marked_install]
        
        # TODO add PackagesNotFoundError (plural).
        packages_missing_installation_candidate = [p for p in marked_install_packages if p.candidate is None]
        if len(packages_missing_installation_candidate) > 0:
            raise errors.PackageNotFoundError(packages_missing_installation_candidate[0].name)
        
        return [(p.name, p.candidate.version) for p in marked_install_packages]
```

snapcraft/internal/repo/apt_cache.py
```diff
@@ -252,7 +265,10 @@ def unmark_packages(self, unmark_names: Set[str]) -> None:
         skipped_essential = set()
         skipped_filtered = set()
 
-        for package in self.cache:
+        for package in self.cache.get_changes():
```
sergiusens, line 268: what is the implication of this change?

sergiusens (commented):
Looks correct, but I have a functionality change question

snapcraft/internal/repo/apt_cache.py
cmatsuoka: It now iterates only over marked packages instead of all packages -- which seems correct because we're unmarking packages.

cmatsuoka:
Can we have some tests to assert that the fixes work and to prevent regressions?

snapcraft/internal/repo/apt_cache.py
cjp256: Done.

cjp256:
> Can we have some tests to assert that the fixes work and to prevent regressions?

It is interesting that none of our existing core spread tests broke on this.  Perhaps t's because they don't `dist-upgrade`?

I haven't thought of a good test case where we have a repo without viable candidates without making truly broken mirror configuration, or like this case where the mirror requires auth to actually fetch packages. I'm open to ideas.  I suppose at a minimum we can dist-upgrade the spread images?

cmatsuoka:
> I haven't thought of a good test case where we have a repo without viable candidates without making truly broken mirror configuration, or like this case where the mirror requires auth to actually fetch packages. I'm open to ideas. I suppose at a minimum we can dist-upgrade the spread images?

Maybe just mocking `cache.get_changes()` to return a couple of handcrafted packages with no candidates could do the trick.

sergiusens:
В письме от martes, 25 de mayo de 2021 15:41:14 -03 пользователь Chris 
Patterson написал:
> > Can we have some tests to assert that the fixes work and to prevent
> > regressions?
> It is interesting that none of our existing core spread tests broke on this.
>  Perhaps t's because they don't `dist-upgrade`?
> 
> I haven't thought of a good test case where we have a repo without viable
> candidates without making truly broken mirror configuration, or like this
> case where the mirror requires auth to actually fetch packages. I'm open to
> ideas.  I suppose at a minimum we can dist-upgrade the spread images?

Perhaps the ESM one with no ua attach is a good candidate.

sergiusens (approved):
As @cmatsuoka mentions, a test would be nice, a package without an installation candidate could be a package that existed in a prior series and is now only referenced by a virtual package, leading to no installation candidate.

cjp256:
Added spread test and verified it fails with current stable:
```
Fetched 21.8 MB in 0s (0 B/s)
Marking 'tree' (and its dependencies) to be fetched
package: <Package: name:'tree' architecture='amd64' id:26233>
'NoneType' object has no attribute 'priority'
  File "/snap/snapcraft/x1/bin/snapcraft", line 33, in <module>
    sys.exit(load_entry_point('snapcraft==4.7.1', 'console_scripts', 'snapcraft')())
  File "/snap/snapcraft/x1/lib/python3.6/site-packages/click/core.py", line 829, in __call__
  File "/snap/snapcraft/x1/lib/python3.6/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/snap/snapcraft/x1/lib/python3.6/site-packages/click/core.py", line 1236, in invoke
    return Command.invoke(self, ctx)
  File "/snap/snapcraft/x1/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
  File "/snap/snapcraft/x1/lib/python3.6/site-packages/click/core.py", line 610, in invoke
  File "/snap/snapcraft/x1/lib/python3.6/site-packages/click/decorators.py", line 21, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/snap/snapcraft/x1/lib/python3.6/site-packages/snapcraft/cli/_runner.py", line 132, in run
    snap_command.invoke(ctx)
  File "/snap/snapcraft/x1/lib/python3.6/site-packages/snapcraft/cli/_command.py", line 89, in invoke
    return super().invoke(ctx)
  File "/snap/snapcraft/x1/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
  File "/snap/snapcraft/x1/lib/python3.6/site-packages/click/core.py", line 610, in invoke
  File "/snap/snapcraft/x1/lib/python3.6/site-packages/snapcraft/cli/lifecycle.py", line 388, in snap
    _execute(steps.PRIME, parts=tuple(), pack_project=True, output=output, **kwargs)
  File "/snap/snapcraft/x1/lib/python3.6/site-packages/snapcraft/cli/lifecycle.py", line 103, in _execute
    lifecycle.execute(step, project_config, parts)
  File "/snap/snapcraft/x1/lib/python3.6/site-packages/snapcraft/internal/lifecycle/_runner.py", line 137, in execute
    executor.run(step, part_names)
  File "/snap/snapcraft/x1/lib/python3.6/site-packages/snapcraft/internal/lifecycle/_runner.py", line 191, in run
    self._handle_step(part_names, part, step, current_step, cli_config)
  File "/snap/snapcraft/x1/lib/python3.6/site-packages/snapcraft/internal/lifecycle/_runner.py", line 205, in _handle_step
    getattr(self, "_run_{}".format(current_step.name))(part)
  File "/snap/snapcraft/x1/lib/python3.6/site-packages/snapcraft/internal/lifecycle/_runner.py", line 247, in _run_pull
    self._run_step(step=steps.PULL, part=part, progress="Pulling")
  File "/snap/snapcraft/x1/lib/python3.6/site-packages/snapcraft/internal/lifecycle/_runner.py", line 322, in _run_step
    self._prepare_step(step=step, part=part)
  File "/snap/snapcraft/x1/lib/python3.6/site-packages/snapcraft/internal/lifecycle/_runner.py", line 313, in _prepare_step
    preparation_function()
  File "/snap/snapcraft/x1/lib/python3.6/site-packages/snapcraft/internal/pluginhandler/__init__.py", line 475, in prepare_pull
    self._fetch_stage_packages()
  File "/snap/snapcraft/x1/lib/python3.6/site-packages/snapcraft/internal/pluginhandler/__init__.py", line 460, in _fetch_stage_packages
    target_arch=self._project._get_stage_packages_target_arch(),
  File "/snap/snapcraft/x1/lib/python3.6/site-packages/snapcraft/internal/repo/_deb.py", line 462, in fetch_stage_packages
    apt_cache.unmark_packages(filtered_names)
  File "/snap/snapcraft/x1/lib/python3.6/site-packages/snapcraft/internal/repo/apt_cache.py", line 256, in unmark_packages
    if package.candidate.priority == "essential":
AttributeError: 'NoneType' object has no attribute 'priority'
snapcraft is running as a snap True, SNAP_NAME set to 'snapcraft'
You can find the traceback in file '/tmp/tmp5lr0suok/trace.txt'.
```

cmatsuoka (approved):
Thanks!

PR #3529, opened 2021-05-26
repo: normalize if stage-packages and check types (CRAFT-222)

sergiusens:
Normalization takes place across the entire working tree for a part.
As a side effect, files that are not part of stage-packages are also
normalized, for that matter, stage-snaps, as those are unpacked first.

When considering the pc snap, Snapcraft creates a snap.pc to correctly
namespace the snaps snap directory. The .pc artifact fixes does not
filter on file type, fileinput is used for the fixup and internally
creates a .bak file as a backup which fails as snap.pc is a directory
and not a file.

-----

### DIFF
--- a/snapcraft/internal/repo/_deb.py
+++ b/snapcraft/internal/repo/_deb.py
@@ -475,7 +475,12 @@ class Ubuntu(BaseRepo):
     def unpack_stage_packages(
         cls, *, stage_packages_path: pathlib.Path, install_path: pathlib.Path
     ) -> None:
-        for pkg_path in stage_packages_path.glob("*.deb"):
+        stage_packages = stage_packages_path.glob("*.deb")
+
+        if not stage_packages:
+            return
+
+        for pkg_path in stage_packages:
             with tempfile.TemporaryDirectory(suffix="deb-extract") as extract_dir:
                 # Extract deb package.
                 cls._extract_deb(pkg_path, extract_dir)
--- a/snapcraft/internal/repo/_base.py
+++ b/snapcraft/internal/repo/_base.py
@@ -253,7 +253,7 @@ class BaseRepo:
                 elif os.path.exists(path):
                     _fix_filemode(path)
 
-                if path.endswith(".pc") and not os.path.islink(path):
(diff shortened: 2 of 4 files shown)


snapcraft/internal/repo/_deb.py
```diff
@@ -475,7 +475,12 @@ def fetch_stage_packages(
     def unpack_stage_packages(
         cls, *, stage_packages_path: pathlib.Path, install_path: pathlib.Path
     ) -> None:
-        for pkg_path in stage_packages_path.glob("*.deb"):
+        stage_packages = stage_packages_path.glob("*.deb")
+
+        if not stage_packages:
```
cjp256, line 480: This hunk do anything?

`if not stage_packages` will never be true AFAIK because glob() returns a generator

cjp256 (dismissed):

snapcraft/internal/repo/_deb.py
sergiusens: Wasn't that iglob?

j5awry:
dropping in to say I reported the initial error. gave this branch a test, and it work for me (building a snap gadget that stages `pc` and then adds more files)

snapcraft/internal/repo/_deb.py
j5awry: yes. glob returns a list of paths, and iglob is returns an iterator.

sergiusens: For completion
```
>>> import glob
>>> glob.glob("*.py")
['setup.py', 'units.py']
```

cmatsuoka (approved):

snapcraft/internal/repo/_deb.py
cmatsuoka: There's a catch here, `glob.glob()` returns a list of paths but `Path.glob()` still returns a generator:
```
>>> import pathlib
>>> p = pathlib.Path("/")
>>> p.glob("*")
<generator object Path.glob at 0x7fa9a7356f20>
```

cmatsuoka: @sergiusens may want to fix that in a follow-up.

PR #3536, opened 2021-06-11
ROS plugins v2 misc fixes

artivis:
This PR most notably,

- looks for ROS packages in `$SNAPCRAFT_PART_INSTALL/opt/ros/$ROS_DISTRO` in case the workspace exists (e.g. provided by a build-snap) during the build step
- does not `rosdep resolve` packages that are available locally in `$SNAPCRAFT_PART_INSTALL/opt/ros/$ROS_DISTRO` during stage step. Such locally available package may be part of the built sources or provided by a stage-snap. Furthermore, it fixes the following current issue; if the source contains packages `foo` & `bar` - two packages that are not released in the rosdistro database - where `foo` depends on `bar`, staging fails as it tries to `rosdep resolve` the `bar` key (which does not exists since it is not released). With this fix, it finds `bar` locally and thus does not try to resolve it.
- port 'test-snapcraft-fake-ros-package-core18' to core20 to test the above 2 bullets

Note:

- I own the test-snapcraft-fake-ros-package-core20 snap that's used in the new spread test, I imagine we should transfer ownership

-----


snapcraft/plugins/v2/_ros.py
```diff
@@ -66,8 +66,17 @@ def _get_workspace_activation_commands(self) -> List[str]:
...
+            "_CATKIN_SETUP_DIR={path} . {path}/setup.sh".format(
+                path='"${SNAPCRAFT_PART_INSTALL}"/opt/ros/$ROS_DISTRO'
+            ),
+            "set -- --local --extend",
+            "else",
+            "set -- --local",
+            "fi",
             ". /opt/ros/$ROS_DISTRO/setup.sh",
-            'eval "$(state)"',
+            'eval "${state}"',
```
sergiusens, line 79: @cjp256 we seem to go back and forth on this. How about we add a code comment :-)

artivis: Here `state` is a variable, not a function,

```bash
$ state="echo ok"
$ $(state)
Command 'state' not found, did you mean: ...
$ ${state}
ok
```

Hence this fix.
A similar snippet can be found in the [Python plugin](https://github.com/snapcore/snapcraft/blob/405c576553d22be9727f681f8cda88b220b39f30/snapcraft/plugins/v2/python.py#L180).

cjp256: @sergiusens I imagine you're thinking of https://github.com/snapcore/snapcraft/commit/f0b2f95fce9b5a4360954784b67046fcd2409103

@artivis' fix is correct, thanks Jeremy!

cjp256 (approved):
LGTM, thanks!

sergiusens (approved):

masynthetic:
I am still having this issue when the launch file is in a separate package and uses an `<exec_depend>` tag in the .xml

I made a raw snap and ros workspace to try and figure out why my project was failing and I was able to make a package which has a <depend> on another package of interfaces this works within snapcraft but once I add the separate package of launch files its dependency fails [exactly like this](https://forum.snapcraft.io/t/ros2-package-fails-staging-no-rosdep-rule-for-pkg/25069).

my test repo can be found [here](https://github.com/masynthetic/rosnap)

artivis:
@masynthetic Thanks for sharing your repo, I'll have a look asap :+1:

PR #3540, opened 2021-06-30
spread tests: remove 16.04 and core (CRAFT-210)

sergiusens:
Cleaned up use of systems within existing tests where possible.

-----


sergiusens:
#3543 can be used as a check that the right tests have been removed

cmatsuoka (approved):
I see a lot of legacy-oriented integration tests removed, I wonder if it would make sense to keep some of those to run with a new system. Even if this is the case, it can be done in a later iteraction.

sergiusens:
> I see a lot of legacy-oriented integration tests removed, I wonder if it would make sense to keep some of those to run with a new system. Even if this is the case, it can be done in a later iteration.

Yeah, I have created tasks to bring them back (the ones that are not duplicates already on the new system)

PR #3541, opened 2021-07-01
Update Crystal v1 plugin

Blacksmoke16:

-----

* Clarifies the documentation that the `crystal-build-options` are passed to the `shards build` command
* Add some additional build dependencies
  * Biggest one being `git`.  Without it `shards` fails to install dependencies
* Consolidate build step into a single `shards build` command
* Update the build flag to use `--without-development`
  * This is required due to a change in behavior with the `--production` flag.  See https://github.com/crystal-lang/shards/pull/473

Manually tested by loading in the changes as a local plugin and build a Crystal shard via snap

Fixes https://github.com/Blacksmoke16/oq/issues/81


mamantoha:
@Blacksmoke16 nice. With these changes will be possible to build a package with `--ignore-crystal-version` option.

Blacksmoke16:
@mamantoha This PR mainly fixes the case where you're building a project (lib) that doesn't have a `shard.lock`.  You can already include that flag, but because `git` was missing it would fail to clone the dependent shards.

mamantoha (approved):

mamantoha:
@Blacksmoke16 yeah, I saw.

Also previously it runs 2 commands:

* `shards install --production`
* `shards build --production`

And additional options can be passed only to the second command. This didn't allow me to build snap with `--ignore-crystal-version`. 

So, thanks for fixing this :)

mamantoha:
@Blacksmoke16 btw, if you want to add additional packages you can use `build-packages` .
See https://github.com/mamantoha/zipstream/blob/master/snap/snapcraft.yaml#L34 for reference.

Blacksmoke16:
@mamantoha Oh nice, that's good to know. My thinking is that the plugin should include these base packages so that anything you do with the stdlib works.  Otherwise you'd prob run into linking issues and need to figure out what lib you need to resolve that.  `build-packages` deff seems to make more sense if your shard is using external/non-standard libs.

On the other hand I could also see an argument for the built in packages to only require the ones absolutely required.  Then require the end user to specify additional ones, e.g. if you `require "yaml"` you need to add `libyaml-dev`.  However I decided against it as it didn't seem like a big deal to me.

sergiusens:
Very nice

Blacksmoke16:
Not sure why CI is failing.  I deff signed the CLA.

Blacksmoke16:
@sergiusens I think I figured out the CLA issue.  I got an email the other day that `You have been added to contributor-agreement-canonical`.  My Launchpad account's email was different than my Github.  I've since added my other emails to my Launchpad account so hopefully that'll resolve that.

sergiusens:
That sounds like the root of @Blacksmoke16 ... I was very busy this week and I seem to have missed my comment about maybe reauthing your commits to match your email address, your solution is a lot better :-D

sergiusens (approved):

Blacksmoke16:
@sergiusens Is it expected that my builds on snapcraft.io/build still are using `4.8.3` versus the new `5.0.0` release?  This is causing the build to continue to fail due to this change only being in the new release.

> This will be transparently supported on snapcraft.io/build and Launchpad by using Snapcraft's 4.x channel track.

Does this mean that that service will always be using `4.x` flow or that it'll only use that flow _IF_ `base` is missing or set to `core`?

PR #3553, opened 2021-07-16
cli: make remote-build more user-friendly

kyrofa:

-----

Remote build currently has a few papercuts.

First of all, if there is no outstanding build, `--recover` literally means nothing: it just spins up another build as if it wasn't supplied at all, which in my experience is never what is desired. It also means there is a disconnect between how `--recover` and `--status` works. Fix that by making `--recover` function more like `--status`, and simply say that there are no builds. Harmony!

Second, one is _always_ prompted about how data being sent to the remote builders is publicly available, even when using options that don't send any meaningful data to the builders, like `--status` and (now) `--recover`. Fix that by no longer using Click to prompt, and prompting only when actually spinning up a new build.

Finally, add tests confirming this behavior.

### DIFF
--- a/snapcraft/cli/remote.py
+++ b/snapcraft/cli/remote.py
@@ -50,10 +50,6 @@ def remotecli():
 @click.option(
     "--launchpad-accept-public-upload",
     is_flag=True,
-    prompt=(
-        "All data sent to remote builders will be publicly available. "
-        "Are you sure you want to continue?"
-    ),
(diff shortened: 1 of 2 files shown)


kyrofa:
That test failure appears unrelated, can someone re-run it please?

sergiusens (approved):
Nice improvements. Thanks for your time!

PR #3554, opened 2021-07-16
cli: add support for parallel remote builds

kyrofa:

-----

Currently, when spinning up a remote build, snapcraft calculates a hash of the path for the project being built, and uses that as a unique identifier for the build. This artificially limits the number of builds that can be happening for a given project to 1, which really limits the usefulness of remote builds.

Instead of calculating a hash for the project directory, calculate a hash that represents the _state_ of the project, such that changing any file in the working tree (e.g. switching git branches) changes the hash. This allows multiple builds for different versions of a given project to be running at the same time.

Of course, this introduces the complication of needing to track such builds. Expose the concept of a "build ID" to the user and allow them to specify it for every remote-build command, so they can check the status of or recover particular builds by providing the relevant build ID.

### DIFF
--- a/snapcraft/project/_project.py
+++ b/snapcraft/project/_project.py
@@ -18,7 +18,7 @@ import hashlib
 import os
 from datetime import datetime
 from pathlib import Path
-from typing import Set
+from typing import List, Set
 
 from snapcraft.internal.deprecations import handle_deprecation_notice
 from snapcraft.internal.meta.snap import Snap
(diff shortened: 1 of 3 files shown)


snapcraft/project/_project.py
```diff
@@ -84,9 +84,25 @@ def _get_build_base(self) -> str:
...
+        # This function uses md5 hashes because they are fast, and because
+        # the hashes only need to be unique per project, so clashes are
+        # tremendously unlikely and not a big deal even if they happen.
+        hashes: List[str] = list()
+        for root, dirs, files in os.walk(self._project_dir):
+            # Sort contents, we need this to be stable so it's reproducible
+            dirs.sort()
+            files.sort()
+
+            for filename in files:
```
sergiusens, line 96: do you have impact numbers for this? What's the hit on a project like Snapcraft itself which is not so small neither so large?

kyrofa: Sure, here's a test script:

```python3
#!/usr/bin/env python3

import time

import snapcraft.project

if __name__ == "__main__":
    project = snapcraft.project.Project()
    start = time.time()
    h = project._get_project_directory_hash()
    end = time.time()

    print(f"{h}: calculated in {end - start:.2f} seconds")
```

Running it in the snapcraft dir:

```
$ ./test.py 
f41210f4761dc6bafbb01b1cb9ab7416: calculated in 0.12 seconds
```

Here's the Nextcloud snap:

```
$ ./test.py 
eb634d5edcc2a5deaada38e330197f2f: calculated in 0.01 seconds
```

sergiusens (approved):
This is an overall improvement and it seems like being able to parallelize automatically is already a benefit for larger projects where the build itself might be significantly _spendier_.

Thanks for the details

PR #3566, opened 2021-08-04
ROS 2 plugins v2 misc fixes

artivis:
This PR is a follow-up on #3536 and essentially fixes the same issues for ROS 2 plugins v2.

This PR most notably,

- changes the colcon install path from `$SNAPCRAFT_PART_INSTALL` to `$SNAPCRAFT_PART_INSTALL/opt/ros/snap` for a cleaner separation of ROS 2 workspaces (similar to what's in the v1 plugin).
- looks for ROS 2 packages in `$SNAPCRAFT_PART_INSTALL/opt/ros/snap` in case the workspace exists (e.g. provided by a stage-snap) during the build step
- does not `rosdep resolve` packages that are available locally in `$SNAPCRAFT_PART_INSTALL` during stage step. Such locally available package may be part of the built sources or provided by a stage-snap. Furthermore, it fixes the following current issue; if the source contains packages `foo` & `bar` - two packages that are not released in the rosdistro database - where `foo` depends on `bar`, staging fails as it tries to `rosdep resolve` the `bar` key (which does not exists since it is not released). With this fix, it finds `bar` locally and thus does not try to resolve it. This issue is fixed by deleting the `${SNAPCRAFT_PART_INSTALL}/opt/ros/snap/COLCON_IGNORE` marker file - if it exists - right after the actual build command. This marker file prevents catkin from crawling the directory while looking for packages during staging.
- adds `test-snapcraft-fake-ros2-package-core20` snap and `colcon-stage-snaps` snap/spread test to test the above bullet points.

Note:

- I own the `test-snapcraft-fake-ros2-package-core20` snap that's used in the new spread test, I imagine we should transfer ownership.

-----

### DIFF
--- a/tests/spread/plugins/v2/snaps/colcon-stage-snaps/snap/snapcraft.yaml
+++ b/tests/spread/plugins/v2/snaps/colcon-stage-snaps/snap/snapcraft.yaml
@@ -5,57 +5,24 @@ description: Verify that ROS 2 stage-snaps are checked for dependencies.
 confinement: strict
 base: core20
 
-package-repositories:
-  - components:
-      - main
-    formats:
-      - deb
-    key-id: C1CF6E31E6BADE8868B172B4F42ED6FBAB17C654
-    key-server: keyserver.ubuntu.com
-    suites:
-      - focal
-    type: apt
-    url: http://repo.ros2.org/ubuntu/main
-
 parts:
-  ros2-foxy-extension:
-      build-packages:
-        - python3-colcon-common-extensions
-        - ros-foxy-ros-core
-      override-build: install -D -m 0755 launch ${SNAPCRAFT_PART_INSTALL}/snap/command-chain/ros2-launch
-      plugin: nil
-      source: $SNAPCRAFT_EXTENSIONS_DIR/ros2
   colcon-part:
     plugin: colcon
     source: .
     build-packages: [g++, make]
     stage-packages: [ros-foxy-ros2launch]
     stage-snaps: [test-snapcraft-fake-ros2-package-core20/latest/edge]
-    build-environment:
-      - ROS_DISTRO: foxy
 
 apps:
   colcon-stage-snaps:
     command: opt/ros/foxy/bin/ros2 launch test_minimal_subscriber talker_subscriber.launch.py
     plugs: [network, network-bind]
-    command-chain:
-      - snap/command-chain/ros2-launch
-    environment:
-      PYTHONPATH: $SNAP/opt/ros/foxy/lib/python3.8/site-packages:$SNAP/usr/lib/python3/dist-packages:${PYTHONPATH}
-      ROS_DISTRO: foxy
+    extensions: [ros2-foxy]
   subscriber:
     command: opt/ros/foxy/bin/ros2 run test_minimal_subscriber test_minimal_subscriber_node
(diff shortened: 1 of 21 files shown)


tests/spread/plugins/v2/snaps/colcon-stage-snaps/snap/snapcraft.yaml
```diff
@@ -0,0 +1,61 @@
+name: colcon-stage-snaps
```
cjp256, line 1: Can this be tested with the extension? Looks like the snapcraft.yaml is pre-expanded.  Or is this setup not supported with the extension?

tests/spread/plugins/v2/snaps/test-snapcraft-fake-ros2-package-core20/README.md
```diff
@@ -0,0 +1,6 @@
+# What is this, and why isn't it used?
```
cjp256, line 1: Is the intention that this package is maintained from here going forward? As in, this is the source tree to publish a new test-snapcraft-fake-ros2-package-core20?

tests/spread/plugins/v2/snaps/test-snapcraft-fake-ros2-package-core20/snap/snapcraft.yaml
```diff
@@ -0,0 +1,51 @@
+name: test-snapcraft-fake-ros2-package-core20
+base: core20
+version: '0.1'
+summary: Bare-bones ROS 2 dependency for snapcraft CLI test
+description: |
+  This snap contains a single unusable ROS 2 package. This snap is not meant to
```
cjp256, line 6: Unusable in the sense that it sense that it lacks practical utility?  The app should function, correct?

tests/spread/plugins/v2/snaps/colcon-stage-snaps/snap/snapcraft.yaml
artivis: My bad, I was toying with the extension and forgot to revert that. Will fix it.

tests/spread/plugins/v2/snaps/test-snapcraft-fake-ros2-package-core20/README.md
artivis: That's correct; much like it's ROS 1 counterpart [test-snapcraft-fake-ros-package-core20](https://github.com/snapcore/snapcraft/blob/5060014453a8c549aaa79de92a61e28efd0bf14d/tests/spread/plugins/v2/snaps/test-snapcraft-fake-ros-package-core20/README.md).

tests/spread/plugins/v2/snaps/test-snapcraft-fake-ros2-package-core20/snap/snapcraft.yaml
artivis: That's right, the app function properly but in a way that's only useful to the test.

sergiusens:
@artivis yes it can, but I merged your already passing PR, so it would re-trigger once conflicts are resolved.

artivis:
Merged master and fixed conflicts; tests passed fine :+1:.

sergiusens (approved):

PR #3575, opened 2021-08-27
NodeJS V1 Plugin: check source-subdir for package.json if specified

debdutdeb:

-----

### Issue

`_install_node_and_yarn` doesn't check for the `source-subdir` value as expected or specified in the `NodejsPluginMissingPackageJsonError` error message - 

https://github.com/snapcore/snapcraft/blob/2a2342c69334a6b2a00dbbda21b09ea81111773f/snapcraft/plugins/v1/nodejs.py#L150
https://github.com/snapcore/snapcraft/blob/2a2342c69334a6b2a00dbbda21b09ea81111773f/snapcraft/plugins/v1/nodejs.py#L80

### Fix

Now it checks if `source-subdir` is specified or not, and checks for `package.json` accordingly.

Fixes https://bugs.launchpad.net/snapcraft/+bug/1694642

### DIFF
--- a/tests/unit/plugins/v1/test_nodejs.py
+++ b/tests/unit/plugins/v1/test_nodejs.py
@@ -41,7 +41,7 @@ class NodePluginBaseTest(PluginsV1BaseTestCase):
             nodejs_version = nodejs._NODEJS_VERSION
             nodejs_package_manager = "npm"
             nodejs_yarn_version = ""
-            source = "."
+            source_subdir = "."
 
         self.options = Options()
 
@@ -74,7 +74,7 @@ class NodePluginBaseTest(PluginsV1BaseTestCase):
         single_bin=False,
         skip_package_json=False,
     ):
-        for directory in (plugin.sourcedir, plugin.builddir):
+        for directory in (os.path.join(plugin.sourcedir, plugin.options.source_subdir), plugin.builddir):
             os.makedirs(directory)
             if not skip_package_json:
                 with open(os.path.join(directory, "package.json"), "w") as json_file:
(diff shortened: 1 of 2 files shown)


sergiusens:
Thanks for your contribution. If you don't mind a unit test would be appreciated.

debdutdeb:
Hi. Thanks for taking a look at it. Modified the tests according to the base plugin change.

tests/unit/plugins/v1/test_nodejs.py
```diff
@@ -41,7 +41,7 @@ class Options:
             nodejs_version = nodejs._NODEJS_VERSION
             nodejs_package_manager = "npm"
             nodejs_yarn_version = ""
-            source = "."
+            source_subdir = "."
```
cmatsuoka, line 44: We should have a test with `source_subdir` set (to an inner directory instead of `.`) and another without it, to check both codepaths. So we might define `Options` without `source_subdir` and set separate pull and build test cases for it (`@pytest.mark.parametrize` may help to run the same test method with different parameters).

tests/unit/plugins/v1/test_nodejs.py
```diff
@@ -477,13 +480,16 @@ class Options:
         nodejs_version = nodejs._NODEJS_VERSION
         nodejs_package_manager = request.param
         nodejs_yarn_version = ""
-        source = "."
+        source_subdir = "."
```
cmatsuoka, line 483: Then we could possibly leave this definition out (unless something using this fixture depends on the source location).

cmatsuoka (changes requested):
Thanks for finding, investigating, and fixing this issue. I did a pass and suggested some changes to improve testing and to handle situations where `source_subdir` is not defined.

sergiusens:
I am closing this for now. Feel free to reopen once addressed

PR #3578, opened 2021-09-08
Fix: files starting with '.' are not organized and also not caught by filesets

debdutdeb:

-----
### Background
`glob.iglob` considers files starting with a dot (or hidden files) to be special cases, thus doesn't count them. 

Similar issue: https://bugs.launchpad.net/snapcraft/+bug/1669854

### DIFF
--- a/snapcraft/internal/pluginhandler/__init__.py
+++ b/snapcraft/internal/pluginhandler/__init__.py
@@ -1290,12 +1290,12 @@ def _migrate_files(
 
 def _organize_filesets(part_name, fileset, base_dir, overwrite):
     for key in sorted(fileset, key=lambda x: ["*" in x, x]):
-        src = os.path.join(base_dir, key)
-        # Remove the leading slash if there so os.path.join
-        # actually joins
-        dst = os.path.join(base_dir, fileset[key].lstrip("/"))
+        base_dir = pathlib.Path(base_dir)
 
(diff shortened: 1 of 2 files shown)


cmatsuoka:
Thanks. The pathlib conversion is certainly welcome, however automatically including dot files that were previously filtered out could cause problems to projects using the dump plugin. For example, snaps suddenly starting to include `.git` in the final payload is not something we want. The number of projects that actually need to ship dot files seem to be quite small, why not explicitly list them in filesets?

debdutdeb:
Sounds fair.

Why don't we check if the source-type is git then exlude the .git folder? This means core snapcraft will include everything for '*', but the dump plugin, stays the same.

Although yes "the number of projects that actually need to ship dot files seem to be quite small" this is true, yet it seems like the .git folder exclusion is more of the plugin's design concern rather than the core product's.

cmatsuoka:
There are many other dot-files we won't want: things like `.github`, `.pytest_cache`, `.readthedocs.yml`, `.tox` come to mind, which may not be related to a specific plugin or source type. I propose we create another PR containing only with the pathlib-related changes (which are much less controversial) and land that first, and then think more about the consequences of changing the filter behavior.

sergiusens:
Closing this as reviewer comments were not addressed. Feel free to reopen when the time comes.

PR #3583, opened 2021-09-23
build providers: snapcraft's new base is core20 (CRAFT-544)

sergiusens:
Stop injecting core18.

-------

-----

### DIFF
--- a/tests/unit/build_providers/test_base_provider.py
+++ b/tests/unit/build_providers/test_base_provider.py
@@ -437,11 +437,10 @@ class BaseProviderProvisionSnapcraftTest(BaseProviderBaseTest):
             [
                 call(snap_name="snapd"),
                 call(snap_name="core20"),
-                call(snap_name="core18"),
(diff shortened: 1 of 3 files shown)


tests/unit/build_providers/test_base_provider.py
```diff
@@ -506,10 +505,11 @@ def test_setup_snapcraft_with_base(self):
             [
                 call(snap_name="snapd"),
                 call(snap_name="core18"),
+                call(snap_name="core20"),
```
cmatsuoka, line 508: Should we also run this test with base core20 to verify if the base snap is not injected twice?

cmatsuoka (approved):
Looks good, left a question about test coverage.

tests/unit/build_providers/test_base_provider.py
sergiusens: we already have that test :-)

These tests will mostly go away or change when we switch to build providers too

PR #3584, opened 2021-09-24
extension/desktop: Prevent keeping updating when there is no user use…

didrocks:
…r-dirs.dirs

On non real desktop (like WSL), we don’t have any service creating
HOME/.config/user-dirs.dirs.
Consequently, as the checksum file was not created on update, we were
fallbacking to always trying to update.
To prevent this, always check the original file first.

Note that CLA check will fail due to using my @ubuntu.com address and launchpad API being limited for private team.
- [Y] Have you successfully run `./runtests.sh tests/unit`?

-----


sergiusens (approved):
yu

PR #3596, opened 2021-11-16
extensions/kde-neon: use platform snap as plug name

ppd:
Using the same plug name "kde-frameworks-5-plug"
for all providers leads to problems when an
application updates its base or changes its default
provider because a new framework snap gets published
for the same base.

In that case, snapd will keep the old framework snap
connected and will happily connect the new framework
snap in addition to the old one.
This breaks applications because they won't find the correct
framework snap mounted at $SNAP/kf5.

Prevent this by using the same strategy that the gnome
extensions use: plug name = content snap name

See https://forum.snapcraft.io/t/snap-accumulates-content-interface-connections-to-multiple-platform-snaps/27523/8

- [x] Have you successfully run `./runtests.sh tests/unit/project_loader/extensions`?

-----

I built a snap (anki-ppd) with this change and everything works as expected. Also CC @jriddell.
I'm not sure about the subsystem prefix; it seemed more complete than just "extensions".

### DIFF
--- a/snapcraft/internal/project_loader/_extensions/kde_neon.py
+++ b/snapcraft/internal/project_loader/_extensions/kde_neon.py
@@ -95,7 +95,7 @@ class ExtensionImpl(Extension):
                     "target": "$SNAP/data-dir/sounds",
                     "default-provider": "gtk-common-themes",
                 },
-                "kde-frameworks-5-plug": {
+                info.provider: {
                     "content": info.content,
                     "interface": "content",
                     "default-provider": info.provider,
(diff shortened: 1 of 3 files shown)


snapcraft/internal/project_loader/_extensions/kde_neon.py
```diff
@@ -126,7 +126,7 @@ def __init__(self, *, extension_name: str, yaml_data: Dict[str, Any]) -> None:
                 "source": "$SNAPCRAFT_EXTENSIONS_DIR/desktop",
                 "source-subdir": "kde-neon",
                 "plugin": "make",
-                "make-parameters": ["PLATFORM_PLUG=kde-frameworks-5-plug"],
+                "make-parameters": ["PLATFORM_PLUG={plug}".format(plug=info.provider)],
```
sergiusens, line 129:
```suggestion
                "make-parameters": [f"PLATFORM_PLUG={info.provider}"],
```

sergiusens (commented):
Looks good, thanks.

snapcraft/internal/project_loader/_extensions/kde_neon.py
ppd: Changed, thanks

ppd:
I also removed the manual connection in the spread test. It is a) not required any more because auto-connect has been granted and b) the plug name is different now, so we'd have to change it anyway.

sergiusens (approved):

sergiusens (approved):
thank you!

PR #3607, opened 2022-01-03
Launchpad bug 1956216: Fix Python installation

mhoeher:
This change fixes some issues with the Python installation within a
Docker container containing snapcraft built from the docker/Dockerfile
in the project as described here:

https://bugs.launchpad.net/snapcraft/+bug/1956216

Background: The image built from the Dockerfile used to work fine. Some
weeks/months ago, it stopped working out of a sudden. Digging into the
failures, it seems the Python installation that comes with recent
snapcraft versions seems to be somehow broken. This change remedies this
by fixing up the Python installation after installation within the
Docker image. For this, we:

- Adjust PATH so the correct Python interpreter will be used.
- Fix a symlink.
- Add a *.pth file to make sure the Python 3 in the snapcraft snap finds
  its own site-packages (for some reason, the site-packages is not in
  sys.path by default).

-----

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -26,17 +26,29 @@ RUN curl -L $(curl -H 'X-Ubuntu-Series: 16' 'https://api.snapcraft.io/api/v1/sna
 RUN mkdir -p /snap/core18
 RUN unsquashfs -d /snap/core18/current core18.snap
 
+# Grab the core20 snap (which snapcraft uses as a base) from the stable channel
+# and unpack it in the proper place.
+RUN curl -L $(curl -H 'X-Ubuntu-Series: 16' 'https://api.snapcraft.io/api/v1/snaps/details/core20' | jq '.download_url' -r) --output core20.snap
+RUN mkdir -p /snap/core20
+RUN unsquashfs -d /snap/core20/current core20.snap
+
 # Grab the snapcraft snap from the $RISK channel and unpack it in the proper
 # place.
 RUN curl -L $(curl -H 'X-Ubuntu-Series: 16' 'https://api.snapcraft.io/api/v1/snaps/details/snapcraft?channel='$RISK | jq '.download_url' -r) --output snapcraft.snap
 RUN mkdir -p /snap/snapcraft
 RUN unsquashfs -d /snap/snapcraft/current snapcraft.snap
 
+# Fix Python3 installation: Make sure we use the interpreter from
+# the snapcraft snap:
+RUN unlink /snap/snapcraft/current/usr/bin/python3
+RUN ln -s /snap/snapcraft/current/usr/bin/python3.* /snap/snapcraft/current/usr/bin/python3
+RUN echo /snap/snapcraft/current/lib/python3.*/site-packages >> /snap/snapcraft/current/usr/lib/python3/dist-packages/site-packages.pth
+
 # Create a snapcraft runner (TODO: move version detection to the core of
 # snapcraft).
 RUN mkdir -p /snap/bin
 RUN echo "#!/bin/sh" > /snap/bin/snapcraft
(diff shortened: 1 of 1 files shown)


stuart-warren:
that's quite interesting, I've just found something similar

we were using `/snap/snapcraft/current/usr/bin/python3` it a script with snapcraft 4.6.2, but with 6.0.2 the python sys.path doesn't include `/snap/snapcraft/current/lib/python3.8/site-packages` where the snapcraft library is...

I spotted `/snap/snapcraft/current/usr/bin/python3` seems to have it though

```sh
root@29eda1e170aa:/# /snap/snapcraft/current/usr/bin/python3 -m site
sys.path = [
    '/',
    '/snap/snapcraft/current/usr/lib/python38.zip',
    '/snap/snapcraft/current/usr/lib/python3.8',
    '/snap/snapcraft/current/usr/lib/python3.8/lib-dynload',
    '/snap/snapcraft/current/usr/lib/python3/dist-packages',
]
USER_BASE: '/root/.local' (doesn't exist)
USER_SITE: '/root/.local/lib/python3.8/site-packages' (doesn't exist)
ENABLE_USER_SITE: False
```
```sh
root@29eda1e170aa:/# /snap/snapcraft/current/bin/python3 -m site
sys.path = [
    '/',
    '/snap/snapcraft/current/usr/lib/python38.zip',
    '/snap/snapcraft/current/usr/lib/python3.8',
    '/snap/snapcraft/current/usr/lib/python3.8/lib-dynload',
    '/snap/snapcraft/current/lib/python3.8/site-packages',
]
USER_BASE: '/root/.local' (doesn't exist)
USER_SITE: '/root/.local/lib/python3.8/site-packages' (doesn't exist)
ENABLE_USER_SITE: False
```

docker/Dockerfile
```diff
@@ -45,6 +57,7 @@ RUN chmod +x /snap/bin/snapcraft
 FROM ubuntu:$UBUNTU
 COPY --from=builder /snap/core /snap/core
 COPY --from=builder /snap/core18 /snap/core18
```
sergiusens, line 59: you can remove core18 (previous base for snapcraft)

sergiusens (approved):
Looks good, a minor comment (which would make default images smaller)

docker/Dockerfile
mhoeher: Hi @sergiusens,

thanks for the feedback! Just a question: Are you sure that core18 can safely be removed? My understanding is, that the image has to contain any core base that a potential downstream app might want to build against, correct? In this case, even if snapcraft itself wouldn't need the core18 base anymore, apps that build against it would start to fail. Or do I miss something here...?

sergiusens:
A segunda-feira, 7 de fevereiro de 2022 17:09:08 -03 Martin Höher escreveu:
> @mhoeher commented on this pull request.
> 
> Hi @sergiusens,
> 
> thanks for the feedback! Just a question: Are you sure that core18 can
> safely be removed? My understanding is, that the image has to contain any
> core base that a potential downstream app might want to build against,
> correct? In this case, even if snapcraft itself wouldn't need the core18
> base anymore, apps that build against it would start to fail. Or do I miss
> something here...?

So this is how it works, you need the base you are targetting on the build 
system , so if your OCI image is "FROM ubuntu:20.04" you want core20, if you 
have "FROM ubuntu:18.04" you want core18 (to support your target runtime) and 
core20 as that is snapcraft's base.

Take into account that the default build here is 16.04. What we may want is 
for the runtime base to be determined from codename (xenial, bionic, focal, 
jammy) or release number (16.04, 18.04, 20.04, 22.04)

sergiusens:
Unrelated issue, CLA not required for this

PR #3610, opened 2022-01-06
lifecycle: do not stage deps for pull on core20 (CRAFT-725)

sergiusens:
core20 uses Plugins V2, there is no plugin logic required from a dependency for
pulling like for V1 Plugins.

-----

### DIFF
--- a/snapcraft/internal/lifecycle/_runner.py
+++ b/snapcraft/internal/lifecycle/_runner.py
@@ -290,13 +290,20 @@ class _Executor:
     def _reprime(self, part, hint=""):
         self._rerun_step(step=steps.PRIME, part=part, progress="Re-priming", hint=hint)
 
-    def _prepare_step(self, *, step: steps.Step, part: pluginhandler.PluginHandler):
-        common.reset_env()
+    def _handle_part_dependencies(
+        self, *, step: steps.Step, part: pluginhandler.PluginHandler
+    ) -> None:
+        # core20 uses Plugins V2 which does not require staging parts for pull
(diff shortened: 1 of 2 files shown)


cmatsuoka (approved):
Looks good (and consistent with craft-parts).

snapcraft/internal/lifecycle/_runner.py
```diff
@@ -290,13 +290,20 @@ def _run_prime(self, part):
...
 
-    def _prepare_step(self, *, step: steps.Step, part: pluginhandler.PluginHandler):
-        common.reset_env()
+    def _handle_part_dependencies(
+        self, *, step: steps.Step, part: pluginhandler.PluginHandler
+    ) -> None:
+        # core20 uses Plugins V2 which does not require staging parts for pull
+        # like V1 Plugins do.
+        if self.project._get_build_base() == "core20" and step == steps.PULL:
+            return
```
xnox, line 299: This will fix this for core20 only, but not for core22? Would it make sense to invert the check?

```python
        if self.project._get_build_base() not in ["core18", "core16", "core"] and step == steps.PULL:
            return
```

john-preston:
Looks like this broke building of telegram-desktop snap, we're getting such a error:

/bin/bash: line 30: /root/stage/patches/libepoxy.patch: No such file or directory

That seem to go from https://github.com/telegramdesktop/tdesktop/blob/f1244e19a10ddb2f65c9fbc00d362c29d18c5ae2/snap/snapcraft.yaml#L207

PR #3612, opened 2022-01-10
Flutter extension: Force GDK to use GLES

AlanGriffiths:
Force GDK to use GLES as some drivers, like the v3d driver for RPi4, don't work with OpenGL.

- [ x] Have you followed the [guidelines for contributing](https://github.com/snapcore/snapcraft/blob/master/CONTRIBUTING.md)?
- [ x] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?

-----

### DIFF
--- a/snapcraft/internal/project_loader/_extensions/_flutter_meta.py
+++ b/snapcraft/internal/project_loader/_extensions/_flutter_meta.py
@@ -138,7 +138,7 @@ class FlutterMetaExtension(type):
             "environment": {
                 "SNAP_DESKTOP_RUNTIME": "$SNAP/gnome-platform",
                 "GDK_GL": "gles",
-                },
+            },
             "layout": {
                 "/usr/share/xml/iso-codes": {
                     "bind": "$SNAP/gnome-platform/usr/share/xml/iso-codes"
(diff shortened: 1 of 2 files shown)


sergiusens:
Question, will there be a reason someone would want to opt out of this setting? In that case I would add it to the desktop-common script to set it if unset.

AlanGriffiths:
> Question, will there be a reason someone would want to opt out of this setting? In that case I would add it to the desktop-common script to set it if unset.

tl;dr: this is an improvement and is probably safe.

In general there can be better performance from GL than GLES for some algorithms, but the differences won't affect Flutter as it doesn't use these nor expose the GL context. (Because it supports a lot of platforms it is conservative in what it uses.)

Theoretically, there could be a driver that works with GL with broken GLES support but, as GLES was designed to be easier to support, that doesn't seem to happen in practice. I don't know of any easy way to detect such drivers and, empirically, GL is broken on a popular device (https://github.com/MirServer/ubuntu-frame/issues/49).

AlanGriffiths:
> Theoretically, there could be a driver that works with GL with broken GLES support

We have an example of such a driver: `virtio-vga,virgl=on` QEMU

https://discourse.ubuntu.com/t/run-flutter-applications-on-ubuntu-core/24607/

PR #3622, opened 2022-01-13
lifecycle: core22 lifecycle conditional on build-attributes entry (CRAFT-725)

sergiusens:

-----

### DIFF
--- a/schema/snapcraft.json
+++ b/schema/snapcraft.json
@@ -1053,6 +1053,7 @@
                             "items": {
                                 "type": "string",
                                 "enum": [
+                                    "core22-step-dependencies",
                                     "enable-patchelf",
                                     "no-patchelf",
                                     "no-install",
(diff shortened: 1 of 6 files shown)


schema/snapcraft.json
```diff
@@ -1053,6 +1053,7 @@
                             "items": {
                                 "type": "string",
                                 "enum": [
+                                    "core22-lifecycle",
```
cmatsuoka, line 1056: This name suggests that the lifecycle will be exactly the same as core22, and it may not be exactly true depending on punctual changes we may decide to adopt. A name emphasizing this specific change could be safer (unless we use this switch later to run non-legacy code to process core20 projects, but new problems could appear in this scenario).

sergiusens:
Maybe in the future @xnox ... we don't have the concept of root level flags

cmatsuoka (commented):
Implementation seems fine, however the attribute name seems too broad in scope. I'd prefer to use a different, more specific attribute name such as `no-stage-pull-dependency`, `disable-stage-pull-dependency` or something similar. Names related to core22 such as `core22-step-dependencies` could be misleading because it may suggest that overlay step dependencies (that don't exist in legacy code) would also be used.

cmatsuoka (approved):

PR #3623, opened 2022-01-14
sources: make submodule recursion configurable

mr-cal:

-----
https://bugs.launchpad.net/snapcraft/+bug/1957767

Add new parameter `source_recurse_submodules`

Example snapcraft.yaml:
```
parts:
    git:
        source: https://github.com/snapcore/snapcraft
        source-type: git
        source-commit: "HEAD"
        source-recurse-submodules: False
```

### DIFF
--- a/snapcraft/internal/sources/__init__.py
+++ b/snapcraft/internal/sources/__init__.py
@@ -65,6 +65,12 @@ code for that part, and how to unpack it if necessary.
     When building, Snapcraft will set the working directory to be this
     subdirectory within the source.
 
+  - source-recurse-submodule: <boolean>
+
+    If true, snapcraft will recursively fetch submodules from the source tree.
+    If false, submodules are not fetched.
+    Default is true.
+
 Note that plugins might well define their own semantics for the 'source'
 keywords, because they handle specific build systems, and many languages
 have their own built-in packaging systems (think CPAN, PyPI, NPM). In those
@@ -147,6 +153,7 @@ __SOURCE_DEFAULTS = {
     "source-type": None,
     "source-branch": None,
     "source-subdir": None,
+    "source-recurse-submodules": None,
 }
 
 
@@ -168,6 +175,7 @@ def get(sourcedir, builddir, options):
         source_tag=getattr(options, "source_tag", None),
         source_commit=getattr(options, "source_commit", None),
         source_branch=getattr(options, "source_branch", None),
+        source_recurse_submodules=getattr(options, "source_recurse_submodules", None),
     )
 
     handler_class = get_source_handler(options.source, source_type=source_type)
--- a/tests/unit/sources/test_git.py
+++ b/tests/unit/sources/test_git.py
@@ -270,6 +270,37 @@ class TestGit(unit.sources.SourceTestCase):  # type: ignore
             ]
         )
 
+    def test_pull_with_recurse_submodules_default(self):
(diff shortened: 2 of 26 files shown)


snapcraft/internal/sources/__init__.py
```diff
@@ -65,6 +65,10 @@
     When building, Snapcraft will set the working directory to be this
     subdirectory within the source.
 
+  - source-recurse-submodule: <boolean>
+
+    Snapcraft will recursively fetch submodules from the source tree by default.
```
cmatsuoka, line 70: Perhaps we could rephrase this a little bit to tell what the option does, and then tell what the default is. Something like "Control whether submodules will be recursively fetched or not. The default is to recursively fetch submodules from the source tree."

tests/unit/sources/test_git.py
```diff
@@ -270,6 +270,28 @@ def test_pull_commit(self):
...
+            "git://my-source", "source_dir", source_recurse_submodules=True
+        )
+
+        git.pull()
+
+        self.mock_run.assert_called_once_with(
+            ["git", "clone", "--recursive", "git://my-source", "source_dir"]
+        )
+
+    def test_pull_with_recurse_submodules_false(self):
```
cmatsuoka, line 284: We should also test the default value to make sure it's enabling recursion.

cmatsuoka (changes requested):
The implementation looks good, I just added some minor notes about the help text and testing the default value.

About the commit message format: could you amend the commit to mention the subsystem being worked on in the subject line (something like `sources: make submodule recursion configurable`) and add some explanatory text? You can also add `LP: #nnnnn` with the number of the Launchpad bug to have it updated automatically when this PR is merged.  Example: see commit 83d54e5

mr-cal:
Thanks, Claudio.  I updated based on your feedback.

cmatsuoka:
> Should I also update `schema/snapcraft.json`?

Yes, the schema file must be updated to be consistent with the available options (otherwise schema validation will fail when using the new property).

mr-cal:
Updated schema.

cmatsuoka (changes requested):
Thanks, the changes look good to me. We don't usually force-push after the review starts but in this case it was necessary to amend the commit subject line. Could you do it once more to add a short description message to the commit with some more information about the change (e.g. "Git clone recursion can now be configured using the xyz option. If not declared, the previous behavior will be kept" or something along those lines. This helps us to understand what the commit was about if we need to revert or cherry-pick it later.

cmatsuoka:
Also I don't think it's a strict requirement in this project but I usually add a "Signed-off-by" line to my commits (using `git commit -s`).

sergiusens:
Looks pretty good.

I am having second thoughts about this, would you think that in the feature this could be expanded to allow for only getting specific submodules?

Mind rebasing to latest changes, and maybe amending "fix: " to something more functional like "static: "?

cmatsuoka:
I'm just wondering if the new proposal is practical to implement, does git provide per-module recursion control, or just the catch-all `--recurse-submodules`?

mr-cal:
Git does provide per-module recursion control.  

For clarity, I put the new proposal under a new pull request, https://github.com/snapcore/snapcraft/pull/3629.  I hope it will satisfies any practical use case.

cmatsuoka:
Superseded by PR #3629

PR #3624, opened 2022-01-14
npm plugin: allow running as root

om26er:

-----

Since snapcraft in lxd/multipass is run as root user, there can be cases where the installation of npm packages fails with "permission denied". This issue has been lurking around for a long time and there isn't really a workaround apart from not using the npm plugin and shipping npm in a custom part.

The reason for that is that npm plugin's `get_build_commands` method downloads npm from upstream and then in the next command calls `npm install` and `npm pack`, leaving no way for the developers to configure npm at all.

Here is one related issue https://forum.snapcraft.io/t/issue-with-nodejs-plugin-when-using-base-keyword/11109/14

### DIFF
--- a/snapcraft/plugins/v2/npm.py
+++ b/snapcraft/plugins/v2/npm.py
@@ -102,5 +102,6 @@ class NpmPlugin(PluginV2):
     def get_build_commands(self) -> List[str]:
         return [
             self._get_node_command(),
+            "npm config set unsafe-perm true",
             'npm install -g --prefix "${SNAPCRAFT_PART_INSTALL}" $(npm pack . | tail -1)',
         ]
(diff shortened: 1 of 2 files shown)


om26er:
We depend heavily on snapcraft/snapd and have a number of node based apps that face this specific issue. Currently we ship our own copy of node as a separate part because of this issue.

sergiusens (approved):

sergiusens (changes requested):
test might need update, while you are at it, mind amending the commit to be something like:

`npm plugin: allow running as root`

Please refer to https://github.com/snapcore/snapcraft/blob/main/CONTRIBUTING.md#snapcraft-contribution-guide

om26er:
> test might need update, while you are at it, mind amending the commit to be something like:

done, both. Please take a look again

om26er:
This unit test is fixed now

sergiusens (approved):

sergiusens:
Thank you!!!!

PR #3626, opened 2022-01-19
cli: integrate craft-cli and add version command (CRAFT-730)

cmatsuoka:
Handle message output and command line arguments parsing using
craft-cli. Implement the `version` command and fall back to the
legacy snapcraft implementation to handle all other functionality.

-----


snapcraft/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Publish your app for Linux users for desktop, cloud, and IoT."""
+
+import os
+
+import pkg_resources  # type: ignore
```
sergiusens, line 21: why do we need to ignore type here?

snapcraft/cli.py
```diff
@@ -16,9 +16,45 @@
...
+import craft_cli
+from craft_cli import EmitterMode, emit
+
+from snapcraft import __version__
 from snapcraft_legacy.cli import legacy
 
+from . import commands
+
+COMMAND_GROUPS = [
+    craft_cli.CommandGroup("Basic", [commands.VersionCommand]),
```
sergiusens, line 30: The snap command has this under "Other"

tests/unit/conftest.py
```diff
@@ -0,0 +1,99 @@
...
+
+import os
+import tempfile
+from pathlib import Path
+
+import pytest
+from craft_cli import messages
+
+
+class RecordingEmitter:
```
sergiusens, line 25: I assume this is copied over? Maybe add a TODO to use a common package?

sergiusens (approved):
Looks good, have a couple of non binding comments and questions

tests/unit/conftest.py
cmatsuoka: This will be removed once testing fixtures are provided by craft-cli itself, or by pytest-craft-cli.

cmatsuoka:
There's an additional issue here: we should leave `--help` to be handled by the legacy implementation until the new help is complete with all valid commands, otherwise it only lists what's implemented (currently only `version` and `--version`).

snapcraft/cli.py
cmatsuoka: Let's move it to Other then.

snapcraft/__init__.py
cmatsuoka: Good catch, this is not needed when `types-setuptools` is installed.

sergiusens (approved):

PR #3629, opened 2022-01-25
sources: make submodule fetching configurable

mr-cal:

-----
Alternative solution from https://github.com/snapcore/snapcraft/pull/3623

https://bugs.launchpad.net/snapcraft/+bug/1957767

Configure which submodules to fetch from the source tree in snapcraft.yaml with `source-submodules: <list-of-submodules>`. 

#### Defined
If `source-submodules` is defined, only the listed submodules are fetched.
```
parts:
  git-test:
    plugin: dump
    source-type: git
    source: git@github.com...
    source-submodules:
      - submodule_1
      - dir1/submodule_2
```

#### Defined and Empty
If `source-submodules` in defined and empty, no submodules are fetched.
```
parts:
  git-test:
    plugin: dump
    source-type: git
    source: git@github.com...
    source-submodules: []
```

#### Not Defined
If `source-submodules` is not defined, all submodules are fetched (default behavior).

#### Note
If a submodule is fetched, any submodules of that submodules are fetched too.  I don't think the behavior needs to be any more nuanced (i.e. no need to define which submodules to fetch recursively and which to fetch non-recursively).

### DIFF
--- a/tests/spread/general/sources/snaps/git-submodules/snapcraft.yaml
+++ b/tests/spread/general/sources/snaps/git-submodules/snapcraft.yaml
@@ -1,0 +1,15 @@
+name: git-recurse-submodules
+base: core20
+version: "0.1"
+summary: Test the use of source-submodules
+description: Make sure source-submodules works
+confinement: strict
+
+parts:
+    git:
+        plugin: dump
+        source: https://github.com/snapcore/core18
+        source-type: git
+        source-submodules:
+          - submodule_1
+          - dir/submodule_2
--- a/tests/spread/general/sources/snaps/git-submodules/snapcraft.yaml
+++ b/tests/spread/general/sources/snaps/git-submodules/snapcraft.yaml
@@ -1,4 +1,4 @@
-name: git-recurse-submodules
+name: git-submodules
 base: core20
 version: "0.1"
 summary: Test the use of source-submodules
@@ -8,7 +8,7 @@ confinement: strict
 parts:
     git:
         plugin: dump
-        source: https://github.com/snapcore/core18
+        source: https://github.com/snapcore/core20
         source-type: git
         source-submodules:
           - submodule_1
--- a/schema/snapcraft.json
+++ b/schema/snapcraft.json
@@ -983,6 +983,15 @@
                             "type": "integer",
                             "default": 0
                         },
+                        "source-submodules": {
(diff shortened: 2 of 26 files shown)


tests/spread/general/sources/snaps/git-submodules/snapcraft.yaml
```diff
@@ -0,0 +1,15 @@
...
+description: Make sure source-submodules works
+confinement: strict
+
+parts:
+    git:
+        plugin: dump
+        source: https://github.com/snapcore/core18
+        source-type: git
+        source-submodules:
+          - submodule_1
```
sergiusens, line 14: these submodules exist?

tests/spread/general/sources/snaps/git-submodules/snapcraft.yaml
```diff
@@ -0,0 +1,15 @@
+name: git-recurse-submodules
+base: core20
```
sergiusens, line 2: if this general suite runs on core20 this is fine, my brain doesn't feel comfortable with cloning core18 though

sergiusens:
Looks like state tracking tests have failed

tests/spread/general/sources/snaps/git-submodules/snapcraft.yaml
mr-cal: No.  Using `--recurse-submodules=<non-existant-directory>` with `git clone` doesn't produce any errors.  The lack of error or warning surprises me, but from what I gather, it's intentional. 

This test only proves that the parameters are parsed and the `git clone` command runs.

Perhaps a slightly better test would be to choose a different repo with a submodule.  Unfortunately I don't see any in snapcore.  The multipass repo has some submodules.  Alternatively I could clone a repo from an outside project.  Think it's worth the trouble?

tests/spread/general/sources/snaps/git-submodules/snapcraft.yaml
mr-cal: Agreed.

mr-cal:
Sorry for 2 force pushes, I made some mistakes.

cmatsuoka (approved):
LGTM. We'll need to bring this to legacy snapcraft in the snapcraft/7.0 branch and to craft-parts.

PR #3630, opened 2022-01-28
[DRAFT] cli: add lifecycle command stubs (CRAFT-764)

cmatsuoka:
Add handlers for the pull, build, stage, and prime commands. These
will be later used do run the parts lifecycle. All other commands
fall back to their legacy implementations.

Also moved CLI tests to a separate package for better organization,
and fix existing tests for the version command.

-----


snapcraft/errors.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class SnapcraftError(CraftError):
+    """Failure in a Snapcraft operation."""
+
+
+class FeatureNotImplemented(SnapcraftError):
+    """Attempt to execute an unimplemented feature."""
+
+    def __init__(self):
```
sergiusens, line 29:
```suggestion
    def __init__(self) -> None:
```

snapcraft/errors.py
```diff
@@ -0,0 +1,30 @@
...
+        super().__init__("This command or feature is not implemented in this release.")
```
sergiusens, line 30: If this is a common construct, CraftError could probably be a dataclass

snapcraft/commands/lifecycle.py
```diff
@@ -0,0 +1,103 @@
...
+        all parts.
+        """
+    )
+
+
+class PrimeCommand(_LifecycleCommand):
+    """Prepare the final payload for packing."""
+
+    name = "prime"
+    help_msg = "Build artifacts defined for a part"
```
sergiusens, line 96:
```suggestion
    help_msg = "Prime artifacts defined for a part"
```

snapcraft/commands/lifecycle.py
```diff
@@ -0,0 +1,103 @@
...
+    overview = textwrap.dedent(
+        """
+        Prepare the final payload to be packed as a snap. If part names are
+        specified only those parts will be primed. The default is to prime
+        all parts.
```
sergiusens, line 101: missing comments about also adding the metadata (snap.yaml, etc...)

setup.cfg
```diff
@@ -34,3 +34,8 @@ follow_imports = silent
 max-line-length = 88
 ignore = E203,E501
 
+[pydocstyle]
+# D107 Missing docstring in __init__ (reason: documented in class docstring)
```
sergiusens, line 38: usually parameters in init can or cannot make it to instance variables so I would not add this general rule here

setup.cfg
```diff
@@ -34,3 +34,8 @@ follow_imports = silent
...
+# D203 1 blank line required before class docstring (reason: pep257 default)
```
sergiusens, line 39: why is this needed?

sergiusens (changes requested):
Looks nice, just some comments for final touches

snapcraft/commands/lifecycle.py
cmatsuoka: The metadata is to be added post-lifecycle, so the primed artifacts will not be the final payload to be packed as a snap. Will change the docstring.

setup.cfg
cmatsuoka: If not added, it will require this unnecessary empty line. We don't add this extra blank line in the rest of our codebase.

setup.cfg
cmatsuoka: These are the warnings to be disabled, so listing it here allow us to also place the docstring in the class definition (where Sphinx expects it to be). This is the format used in craft-parts/providers.

snapcraft/commands/lifecycle.py
cmatsuoka: Oops. Will fix.

snapcraft/errors.py
cmatsuoka: Done, also changed the error to display a description of the missing feature.

setup.cfg
sergiusens: Right, my comment is wrt something like this construct:

```python
class Foo:
    """Foo class.

    :ivar foobar: a modified bar.
    """

    def __init__(self, bar):
        """Initialize foo with bar.

        :param bar: a number.
        """
        self.foobar = bar * 2
```

But now I see that Class can have `param`, I keep missing that.

sergiusens (approved):

PR #3363, opened 2020-11-10
repo: apt sources management refactor

cjp256:
- Finish removing install interfaces from PackageRepository.

- Migrate apt sources configuration from _deb.Ubuntu to
  apt_sources_manager.AptSourcesManager.

- Update project loader to use AptSourcesManager.

There should be no change in behavior.

Signed-off-by: Chris Patterson <chris.patterson@canonical.com>

- [ ] Have you followed the [guidelines for contributing](https://github.com/snapcore/snapcraft/blob/master/CONTRIBUTING.md)?
- [ ] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [ ] Have you successfully run `./runtests.sh static`?
- [ ] Have you successfully run `./runtests.sh tests/unit`?

-----

PR #3359, opened 2020-11-07
repo: key management refactor

cjp256:
- Decouple key management from _deb.Ubuntu, introducing a new
  apt_key_manager.AptKeyManager class.

- Improve unit test coverage of key management.

- Pivot to enforcing the 40-character key ID. Update the
  schema and package repository validation checks to
  enforce this.  Update the spread tests to comply.

- Update the key asset search to require matching keys to
  be found in snap/keys/12345678.asc, where 12345678 is
  the first eight characters of the key ID.  Update the
  schema and package repository validation checks to
  enforce this.  Update the spread tests to comply.

This functionality should finalize key management for the
stabilization of package-repositories, with the exception of
a related change to ensure all key assets are used.

Signed-off-by: Chris Patterson <chris.patterson@canonical.com>

- [ ] Have you followed the [guidelines for contributing](https://github.com/snapcore/snapcraft/blob/master/CONTRIBUTING.md)?
- [ ] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [ ] Have you successfully run `./runtests.sh static`?
- [ ] Have you successfully run `./runtests.sh tests/unit`?

-----

PR #3647, opened 2022-02-20
repo: add package repository handling (CRAFT-846)

cmatsuoka:
Port definitions and helpers to add apt repositories and PPAs. Original
code from PRs #2911, #3341, #3359, #3363.
    
Co-authored-by: Chris Patterson <chris.patterson@canonical.com>

-----

### DIFF
--- a/snapcraft/projects.py
+++ b/snapcraft/projects.py
@@ -25,6 +25,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
 import pydantic
 from pydantic import conlist, constr
 
+from snapcraft import repo
 from snapcraft.errors import ProjectValidationError
 from snapcraft.parts import validation as parts_validation
 
@@ -36,7 +37,7 @@ class ProjectModel(pydantic.BaseModel):
         """Pydantic model configuration."""
 
         validate_assignment = True
-        # extra = "forbid"
+        extra = "allow"  # FIXME: change to 'forbid' after model complete
         allow_mutation = False
         allow_population_by_field_name = True
         alias_generator = lambda s: s.replace("_", "-")  # noqa: E731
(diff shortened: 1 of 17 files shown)


snapcraft/repo/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Package repository helpers."""
+
+from .apt_key_manager import AptKeyManager  # noqa: F401
+from .apt_sources_manager import AptSourcesManager  # noqa: F401
+from .package_repository import PackageRepository  # noqa: F401
+from .package_repository import PackageRepositoryApt  # noqa: F401
+from .package_repository import PackageRepositoryAptPPA  # noqa: F401
```
sergiusens, line 23: instead of adding noqa consider declaring __all__ for the package

sergiusens (commented):
One comment about __init__ and noqa.
Thanks for keeping it tightly separate, we might consider making another package for this as this might be useful for other projects like rockcraft.

snapcraft/projects.py
```diff
@@ -152,6 +154,27 @@ class Architecture(ProjectModel):
     build_to: Optional[Union[str, UniqueStrList]]
 
 
+class AptDeb(ProjectModel):
```
sergiusens, line 157: if this model is moved to repo, it might make it easier to part with into a new package in the future

cmatsuoka: I'll change the boundaries between repo and the rest of snapcraft to reduce coupling and make it easier to spin off craft-repositories.

snapcraft/repo/__init__.py
cmatsuoka: Done. The decoupling change also drastically reduced the amount of exported symbols, but having `__all__` won't hurt (and it will be needed if we spin off craft-repositories).

snapcraft/projects.py
cmatsuoka: Repo changed to be more self-contained with less API surface. The API is now consistent with craft-parts (data is received as a dictionary and validated using a validator function provided by the package). This will become clear when the package repositories functionality is integrated with the rest of snapcraft.

sergiusens (approved):

PR #3676, opened 2022-04-04
ROS plugins v2: Exclude packages in stage snap from fetched packages during staging 

artivis:
This PR introduces a ROS-specific workaround that prevents from fetching debs already contained in a stage-snap during staging.

Assuming the following dependency chain `foo -> bar -> baz` and a stage snap containing package `baz`. Snapping a ROS app `foo` means that at staging the package `bar` will be fetched (as it should) together with its dependency `baz`. However, `baz` being already provided by a stage snap we would like to avoid fetching it again, potentially shadowing a desired specific `baz` version or custom changes. Afaik, there is no mechanisms at the moment to handle this case.

This PR, somewhat akin to [get_packages_in_base](https://github.com/snapcore/snapcraft/blob/0224b48ea0e1cc7378e1ad392b41466f005ae3e6/snapcraft/internal/repo/_deb.py#L261) for ROS stage snaps, proposes a workaround.

- [ ] Have you successfully run `./runtests.sh static`? -> Some unrelated `black` errors

-----

### DIFF
--- a/snapcraft/internal/repo/_deb.py
+++ b/snapcraft/internal/repo/_deb.py
@@ -450,6 +450,7 @@ class Ubuntu(BaseRepo):
         base: str,
         stage_packages_path: pathlib.Path,
         target_arch: str,
+        filters: Set[str] = set(),
     ) -> List[str]:
         logger.debug(f"Requested stage-packages: {sorted(package_names)!r}")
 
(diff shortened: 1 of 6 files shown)


sergiusens:
This PR looks out of date. We probably want to have a new discussion around this for when the time comes

PR #3679, opened 2022-04-07
Expand extensions

sergiusens:

-----
CRAFT-948

### DIFF
--- a/snapcraft/parts/lifecycle.py
+++ b/snapcraft/parts/lifecycle.py
@@ -53,7 +53,11 @@ _SNAP_PROJECT_FILES = [
 ]
 
 
-def _get_snap_project() -> _SnapProject:
+def get_snap_project() -> _SnapProject:
+    """Find the snapcraft.yaml to load.
+
+    :raises SnapcraftError: if the project yaml file cannot be found.
+    """
     for snap_project in _SNAP_PROJECT_FILES:
         if snap_project.project_file.exists():
             return snap_project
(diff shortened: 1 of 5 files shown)


snapcraft/parts/lifecycle.py
```diff
@@ -35,54 +36,58 @@
...
 
-_PROJECT_FILES = [
-    Path("snapcraft.yaml"),
-    Path("snap/snapcraft.yaml"),
-    Path("build-aux/snap/snapcraft.yaml"),
-    Path(".snapcraft.yaml"),
+@dataclass
+class _SnapProject:
+    project_file: Path
+    project_dir: Path = Path("snap")
```
cmatsuoka, line 42: This name could generate some confusion because `project_dir` is used elsewhere as the project top-level directory where you run snapcraft from. IIRC this directory (`snap` if you have `snap/snapcraft.yaml`) is called `assets_dir` in the rest of the code.

cmatsuoka (commented):
Implementation seems good (and it seems to fix an error in assets dir if snapcraft runs from a directory called "snap") -- I would only change the `project_dir` name to something else (`assets_dir` sounds like a good choice) because the name is already being used with a different meaning.

snapcraft/parts/lifecycle.py
sergiusens: fixed

cmatsuoka (approved):

PR #3685, opened 2022-04-14
utils: in-house strtobool implementation

sergiusens:
distutils is mostly going away in Python 3.12

-----

### DIFF
--- a/snapcraft/utils.py
+++ b/snapcraft/utils.py
@@ -16,7 +16,6 @@
 
 """Utilities for snapcraft."""
 
-import distutils.util
 import logging
 import os
 import pathlib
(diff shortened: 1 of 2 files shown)


snapcraft/utils.py
```diff
@@ -29,10 +28,27 @@
...
+        return False
+
+    raise ValueError(f"Invalid boolean value of {value!r}")
+
+
+def is_managed_mode() -> bool:
     """Check if snapcraft is running in a managed environment."""
     managed_flag = os.getenv("SNAPCRAFT_MANAGED_MODE", "n")
-    return distutils.util.strtobool(managed_flag) == 1
+    return strtobool(managed_flag)
```
cmatsuoka, line 51: `SNAPCRAFT_MANAGED_MODE` is set by ourselves, can't we just assume we set it consistently with a known value?

cmatsuoka (approved):
Implementation looks good but it doesn't seem necessary for `SNAPCRAFT_MANAGED_MODE`, I'm assuming it will also be needed elsewhere.

snapcraft/utils.py
sergiusens: I guess, but we do use strtobool in the older code as well

PR #3693, opened 2022-04-19
projects: adoptable fields are optional if adopt-info used

cmatsuoka:
Project fields `version`, `summary`, `description` and `grade` are
allowed to be unspecified in project if `adopt-info` is used. The final
values of such fields are still required and expected to be adopted
from metadata, otherwise an error is raised.

-----
CRAFT-1031

### DIFF
--- a/snapcraft/meta/snap_yaml.py
+++ b/snapcraft/meta/snap_yaml.py
@@ -107,7 +107,7 @@ class SnapMetadata(YamlModel):
     hooks: Optional[Dict[str, Any]]
 
 
-def write(project: Project, prime_dir: Path, *, arch: str, version: str, grade: str):
+def write(project: Project, prime_dir: Path, *, arch: str):
     """Create a snap.yaml file."""
     meta_dir = prime_dir / "meta"
     meta_dir.mkdir(parents=True, exist_ok=True)
@@ -159,9 +159,9 @@ def write(project: Project, prime_dir: Path, *, arch: str, version: str, grade: 
     snap_metadata = SnapMetadata(
         name=project.name,
         title=project.title,
-        version=version,
+        version=project.version,
         summary=project.summary,
-        description=project.description,
+        description=project.description,  # type: ignore
         license=project.license,
         type=project.type,
         architectures=[arch],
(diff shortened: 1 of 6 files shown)


snapcraft/projects.py
```diff
@@ -39,7 +36,7 @@ class Config:  # pylint: disable=too-few-public-methods
 
         validate_assignment = True
         extra = "allow"  # FIXME: change to 'forbid' after model complete
-        allow_mutation = False
+        allow_mutation = True
```
cmatsuoka, line 39: We update project fields with adopted metadata to take advantage of pydantic's field assignment validation.

snapcraft/meta/snap_yaml.py
```diff
@@ -159,9 +159,9 @@ def write(project: Project, prime_dir: Path, *, arch: str, version: str, grade:
     snap_metadata = SnapMetadata(
         name=project.name,
         title=project.title,
-        version=version,
+        version=project.version,
         summary=project.summary,
-        description=project.description,
+        description=project.description,  # type: ignore
```
sergiusens, line 164: why is the type ignored?

snapcraft/projects.py
sergiusens: I don't understand this comment, maybe a more clear code comment would be better

sergiusens:
Looks good, just some comments

snapcraft/meta/snap_yaml.py
cmatsuoka: It's because description is now optional in project. At this point we're already sure it has a value, but mypy still thinks it can be None.

snapcraft/parts/lifecycle.py
```diff
@@ -210,6 +200,47 @@ def _run_command(
...
+        if project_vars["version"]:
+            project.version = project_vars["version"]
+        if project_vars["grade"]:
+            project.grade = project_vars["grade"]  # type: ignore
+    except pydantic.ValidationError as err:
+        _raise_formatted_validation_error(err)
+        raise errors.SnapcraftError(f"error setting variable: {err}")
+
+    # Fields that must not end empty
+    for field in ("version", "grade", "summary", "description"):
```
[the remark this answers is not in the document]
cmatsuoka: Will define.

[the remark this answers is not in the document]
cmatsuoka: Defined as `MANDATORY_ADOPTABLE_FIELDS` because there are other fields that are strictly mandatory.

sergiusens (approved):

PR #3700, opened 2022-04-23
meta: adopt metadata info

cmatsuoka:
Update project fields with entries from external metadata.

-----
CRAFT-1036


snapcraft/parts/lifecycle.py
```diff
@@ -178,10 +199,19 @@ def _run_command(
...
+        metadata_list = lifecycle.extract_metadata()
+        update_project_metadata(
+            project,
+            project_vars=project_vars,
+            metadata_list=metadata_list,
+            assets_dir=assets_dir,
+            prime_dir=lifecycle.prime_dir,
+        )
+
+        # TODO: copy meta/gui assets
```
sergiusens, line 214: I assume this is copy snap/gui to meta/gui

snapcraft/parts/update_metadata.py
```diff
@@ -0,0 +1,163 @@
...
+    """Look for icons files and update project.
+
+    Existing icon in snap/gui/icon.{png,svg} has precedence over extracted data
+    """
+    for icon_file in (f"{assets_dir}/gui/icon.png", f"{assets_dir}/gui/icon.svg"):
+        if Path(icon_file).is_file():
+            break
+    else:
+        if metadata.icon and Path(prime_dir, metadata.icon).is_file():
+            project.icon = metadata.icon
```
sergiusens, line 108: Somewhere above `VALID_ICONS_EXTENSIONS: Final[List[str]] = ["png", "svg"]`
then,
```suggestion
    icon_files = (Path(f"{assets_dir}/gui/icon.{ext}") for ext in VALID_ICON_EXTENSIONS)
    existing_icon_files = [p for p in icon_files if p.is_file()]
    if metadata.icon and Path(prime_dir, metadata.icon).is_file():
        existing_icon_files.append(metadata.icon)
    
    with contextlib.suppress(IndexError):
        project.icon = existing_icon_files[0]
```

This might simplify the logic (one place) to decide how to copy the icon over.

snapcraft/projects.py
```diff
@@ -187,6 +187,24 @@ def _validate_aliases(cls, aliases):
 
         return aliases
 
+    @classmethod
+    def unmarshal(cls, data: Dict[str, Any]) -> "App":
```
sergiusens, line 191: I cannot spot where this is used.

sergiusens (approved):
Looks good, I did not look a the appstream tests in detail. Added a few comments which more questions really.

snapcraft/parts/lifecycle.py
cmatsuoka: It needs to do that and update the desktop file to replace the icon entry to the final icon location. This (somewhat complex) logic is already implemented in legacy (`setup_assets`, `find_icon_file`, `finalize_icon`, `write_snap_directory`) and can be ported with minor changes. This code already handles download/installation of remote icons.

snapcraft/parts/parts.py
```diff
@@ -166,6 +170,34 @@ def clean(self, *, part_names: Optional[List[str]] = None) -> None:
         emit.message(message, intermediate=True)
         self._lcm.clean(part_names=part_names)
 
+    def extract_metadata(self) -> List[ExtractedMetadata]:
+        """Obtain metadata information."""
+        if not self._adopt_info or self._adopt_info not in self._parse_info:
+            return []
+
+        # Currently not looking in stage and prime dirs
```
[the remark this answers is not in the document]
cmatsuoka: This comment should be removed .It says "currently" but this won't need to be changed, even legacy code only checks these same places (possibly for the same reason you mentioned).

snapcraft/parts/update_metadata.py
cmatsuoka: I like the idea of defining valid icon extensions separately. I'll refactor the code a bit to use the definition, but I'd avoid testing for the existence of all files when just testing the first one could be enough if it satisfies the exit condition. The icon copying code probably won't be affected by this implementation.

snapcraft/projects.py
cmatsuoka: It's used in tests.

PR #3710, opened 2022-04-26
tests: use craft-cli lib fixtures

facundobatista:
CRAFT-997


sergiusens:
I will update after the pylint fix lands and will amend the commit to follow the guidelines

sergiusens (approved):

PR #3760, opened 2022-05-25
extensions: add core22 gnome

sergiusens:

-----

### DIFF
--- a/extensions/desktop/command-chain/Makefile
+++ b/extensions/desktop/command-chain/Makefile
@@ -1,0 +1,10 @@
+#!/usr/bin/make -f
+
+BIN_DIR              := $(DESTDIR)/snap/command-chain
+
+*:
+	install -D -m755 "$@" "$(BIN_DIR)"/"$@"
+
+install: hooks-configure-fonts desktop-launch run
+
+.PHONY: hooks-configure-fonts desktop-launch run
--- a/snapcraft/extensions/gnome.py
+++ b/snapcraft/extensions/gnome.py
@@ -15,16 +15,31 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 """Generic GNOME extension to support core22 and onwards."""
-
-
+import dataclasses
+import functools
+import re
 from typing import Any, Dict, Optional, Tuple
 
 from overrides import overrides
 
-from .extension import Extension, get_extensions_data_dir, prepend_to_env
+from .extension import (
+    Extension,
+    get_build_snaps,
+    get_extensions_data_dir,
+    prepend_to_env,
+)
 
-_PLATFORM_SNAP = dict(core22="gnome-42-2204")
-_SDK_SNAP = dict(core22="gnome-42-2204-sdk")
+_SDK_SNAP = {"core22": "gnome-42-2204-sdk"}
+_PLATFORM_TRANSLATION = {"core22": "2204"}
+
+
+@dataclasses.dataclass
+class GNOMESnaps:
+    """A structure of GNOME related snaps."""
+
+    sdk: str
+    content: str
+    builtin: bool = True
 
 
 class GNOME(Extension):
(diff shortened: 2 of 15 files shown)


snapcraft/extensions/gnome.py
```diff
@@ -0,0 +1,212 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2022 Canonical Ltd
```
cmatsuoka, line 3: For new files we're using a different copyright line format.

snapcraft/extensions/gnome.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+from typing import Any, Dict, Optional, Tuple
+
+from overrides import overrides
+
+from .extension import Extension, get_extensions_data_dir
+
+_PLATFORM_SNAP = dict(core22="gnome-42-2204")
+_SDK_SNAP = dict(core22="gnome-42-2204-sdk")
```
cmatsuoka, line 27: Any special reason for not using literal dictionaries instead of `dict()`?

cmatsuoka (commented):
Having a single generic gnome extension looks much nicer than having multiple version-specific ones.

snapcraft/extensions/gnome.py
sergiusens: copy paste, this can change :-)

snapcraft/extensions/gnome.py
sergiusens: Good catch

snapcraft/extensions/gnome.py
```diff
@@ -0,0 +1,212 @@
...
+    At build time it ensures the right build dependencies are setup and for
+    the runtime it ensures the application is run in an environment catered
+    for GNOME applications.
+
+    It configures each application with the following plugs:
+
+    - GTK3 Themes.
+    - Common Icon Themes.
+    - Common Sound Themes.
+    - The GNOME runtime libraries and utilities corresponding to 3.38.
```
kenvandine, line 44: This should be 42, but you can probably drop the version.

snapcraft/extensions/registry.py
```diff
@@ -16,15 +16,19 @@
...
 
-ExtensionType = Type[Extension]
+if TYPE_CHECKING:
+    from .extension import Extension
 
-_EXTENSIONS: Dict[str, ExtensionType] = {}
+    ExtensionType = Type[Extension]
+
+
+_EXTENSIONS: Dict[str, "ExtensionType"] = {"gnome": GNOME}
```
kenvandine, line 31: Does this mean we won't have a way to use an extension to use a future platform snap, like gnome-44-2204 that is still based on core22?

sergiusens: this is what we agreed to, remember? If it came to it, it would be a transparent upgrade for users

kenvandine: I know, and I like the idea in principle but it doesn't give us an easy way to build things that need something newer versions before the next LTS.  Could we add an optional flag of some sort that let's us override the version of the content and build snaps?

sergiusens: latest commit has a mechanism

extensions/desktop/command-chain/Makefile
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/make -f
+
+BIN_DIR              := $(DESTDIR)/snap/command-chain
+
+*:
+	install -D -m755 "$@" "$(BIN_DIR)"/"$@"
+
+install: hooks-configure-fonts desktop-launch run
+
+.PHONY: hooks-configure-fonts desktop-launch run
```
sergio-costas, line 10: Lacks a blank line at the end of the file.

snapcraft/extensions/gnome.py
```diff
@@ -0,0 +1,264 @@
...
+
+    @functools.cached_property
+    def gnome_snaps(self) -> GNOMESnaps:
+        """Return the GNOME related snaps to use to construct the environment."""
+        base = self.yaml_data["base"]
+        sdk_snap = _SDK_SNAP[base]
+        build_snaps = get_build_snaps(parts_yaml_data=self.yaml_data.get("parts"))
+        matcher = re.compile(r"gnome-\d+-" + _PLATFORM_TRANSLATION[base] + r"-sdk.*")
+        sdk_snap_candidates = [s for s in build_snaps if matcher.match(s)]
+        if sdk_snap_candidates:
```
sergio-costas, line 109: What happens if the user puts more than one -sdk snap? Is that checked elsewhere?

sergio-costas (changes requested):

sergio-costas (changes requested):
Just some little comments.

PR #3766, opened 2022-05-27
extensions: prepend and append functions for ENV

sergio-costas:
This MR adds two functions that simplifies the management of environment variables, allowing to just pass the name of one variable and one or more paths, and they will return a piece of shell code that prepends or appends the paths to the variable, taking into account the case when the variable is empty, to avoid adding a trailing or leading colon.

-----

### DIFF
--- a/snapcraft/extensions/extension.py
+++ b/snapcraft/extensions/extension.py
@@ -20,7 +20,7 @@ import abc
 import os
 import sys
 from pathlib import Path
-from typing import Any, Dict, Optional, Tuple, final
+from typing import Any, Dict, Optional, Sequence, Tuple, final
 
 from craft_cli import emit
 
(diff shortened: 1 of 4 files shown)


snapcraft/extensions/extension.py
```diff
@@ -32,6 +32,28 @@ def get_extensions_data_dir() -> Path:
...
+    """Prepend one or more paths to an environment variable.
+
+    Returns a Shell string where one or more @paths are prepend
+    before @env_variable. The code takes into account the case
+    where the environment variable is empty, to avoid putting
+    a colon at the end.
+    """
+    return ":".join(paths) + f"${{{env_variable}:+:${env_variable}}}"
+
+
```
sergiusens, line 56: We use the documentation format below (feel free to adjust wording).
While the case now is that all separators are : it may not be the case tomorrow (we already have some that use a different separator).
While I did not make the suggestion in the code block, I believe it might be better to expect a Sequence[str] for paths instead of *paths

```suggestion
def append_to_env(env_variable: str, *paths: str, separator: str = ":") -> str:
    """Return a string for env_variable with one of more paths appended.

    :param env_variable: the variable to operate on.
    :para paths: paths to append.
    :param separator: the separator to use.
    :returns: a shell string where one or more paths are appended
                  to env_variable. The code takes into account the case
                  where the environment variable is empty, to avoid putting
                  a separator token at the start.
    """
    return f"${{{env_variable}{separator}+${env_variable}{separator}}}" + separator.join(paths)


def prepend_to_env(env_variable: str, *paths: str. separator: str = ":") -> str:
    """Return a string for env_variable with one of more paths prepended.
    
    :param env_variable: the variable to operate on.
    :param separator: the separator to use.
    :returns: a shell string where one or more paths are prepended
                  before env_variable. The code takes into account the case
                  where the environment variable is empty, to avoid putting
                  a separator token at the end.
    """
    return separator.join(paths) + f"${{{env_variable}{separator}+{separator}${env_variable}}}"


```

sergiusens (commented):
Looks good, just some nits

snapcraft/extensions/extension.py
sergio-costas: Looks good, but I've been doing some tests and checking google, and it seems that the right syntax is:

f"${{{env_variable}:+${env_variable}{separator}}}" + separator.join(paths)

This is: the first colon must be kept because it seems that it is part of the expression (although, honestly, this is the first time that I see this construction...).

I'll update it with these changes.

sergio-costas: I'll change the *paths with a Sequence[str] too.

sergio-costas: Done.

sergiusens (approved):

PR #3767, opened 2022-05-28
parts: don't copy metadata if already at destination

cmatsuoka:
When preparing metadata assets, don't attempt to copy files if they
are already where the system expects it to be.

-----

### DIFF
--- a/tests/unit/parts/test_setup_assets.py
+++ b/tests/unit/parts/test_setup_assets.py
@@ -183,11 +183,15 @@ def test_kernel_missing(yaml_data, new_dir):
     [
         (Path("foo"), Path("bar"), True),
         (Path("foo"), Path("foo"), False),
+        (Path("foo"), Path("dest/foo"), True),
     ],
 )
 def test_copy_files(source, destination, copied, new_dir, mocker):
     copy_mock = mocker.patch("shutil.copy")
     Path("foo").touch()
+    Path("dest").mkdir()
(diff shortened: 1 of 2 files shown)


tests/unit/parts/test_setup_assets.py
```diff
@@ -176,6 +178,23 @@ def test_kernel_missing(yaml_data, new_dir):
...
 
 
+@pytest.mark.parametrize(
+    "source,destination,copied",
+    [
+        (Path("foo"), Path("bar"), True),
+        (Path("foo"), Path("foo"), False),
+    ],
+)
+def test_copy_files(source, destination, copied, new_dir, mocker):
```
sergiusens, line 188: you are missing a test or scenario for "samefile" but different contents

sergiusens: also, legacy has link_or_copy (maybe already in new snapcraft) which might solve some issues around existing files

cmatsuoka: The "samefile" test should be inode-based so different content must mean different files, but added a case with same name and different content.

cmatsuoka: `link_or_copy` was not originally used because it dragged a large infrastructure to deal with only a couple of files, but now that more files are copied we can consider bringing it in a refactoring PR.

sergiusens (approved):

PR #3779, opened 2022-06-07
meta: handle version git

cmatsuoka:
Add special handling for projects with `version: git` so that the
snap version is set to the result of `git describe --dirty`.

-----

### DIFF
--- a/snapcraft/meta/snap_yaml.py
+++ b/snapcraft/meta/snap_yaml.py
@@ -20,6 +20,8 @@ from pathlib import Path
 from typing import Any, Dict, List, Optional, Set, Union, cast
 
 import yaml
+from craft_cli import emit
+from craft_parts.sources.git_source import GitSource
 from pydantic_yaml import YamlModel
 
 from snapcraft.projects import Project
@@ -175,11 +177,12 @@ def write(project: Project, prime_dir: Path, *, arch: str, arch_triplet: str):
         assumes.add("command-chain")
 
     environment = _populate_environment(project.environment, prime_dir, arch_triplet)
+    version = _process_version(project.version)
 
     snap_metadata = SnapMetadata(
         name=project.name,
         title=project.title,
-        version=project.version,
+        version=version,
         summary=project.summary,
         description=project.description,  # type: ignore
         license=project.license,
@@ -252,3 +255,19 @@ def _populate_environment(
 
     # if the environment only contained a null LD_LIBRARY_PATH and a null PATH, return None
     return None
+
+
+def _process_version(version: Optional[str]) -> str:
+    """Handle special version strings."""
+    if version is None:
+        raise ValueError("version cannot be None")
+
+    new_version = version
+    if version == "git":
+        emit.progress("Determining the version from the project repo (version: git).")
+        new_version = GitSource.generate_version()
+
+    if new_version != version:
+        emit.message(f"Version has been set to {new_version!r}", intermediate=True)
+
+    return new_version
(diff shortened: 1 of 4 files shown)


snapcraft/meta/snap_yaml.py
```diff
@@ -175,11 +177,13 @@ def write(project: Project, prime_dir: Path, *, arch: str, arch_triplet: str):
         assumes.add("command-chain")
 
     environment = _populate_environment(project.environment, prime_dir, arch_triplet)
+    # mypy false positive: at this point project.version is a string
+    version = _process_version(project.version)  # type: ignore
```
sergiusens, line 181: instead of this I would do something like this to indicate that the project at this point needs to have been processed

```suggestion
    if project.version is None:
        raise ValueError("'project.version' cannot be None")
    version = _process_version(project.version)  # type: ignore
```

sergiusens: or handle the case inside _process_version

tests/unit/meta/test_snap_yaml.py
```diff
@@ -502,3 +503,24 @@ def test_project_environment_ld_library_path_null(simple_project, new_dir):
           PATH: $SNAP/usr/sbin:$SNAP/usr/bin:$SNAP/sbin:$SNAP/bin:$PATH
         """
     )
+
+
+def test_version_git(simple_project, new_dir, mocker):
```
sergiusens, line 508: this might be better in spread where the git setup is somewhat already handled, for units just ensure the externalities are  called (GitSource)

snapcraft/meta/snap_yaml.py
cmatsuoka: Handling it inside `_process_version`.

tests/unit/meta/test_snap_yaml.py
cmatsuoka: Mocking `generate_version`, craft-parts tests should ensure it's working as expected.

cmatsuoka: Added a spread test.

cmatsuoka:
Amended version verification and unit tests, and rebased on main.

tests/spread/general/version-git/task.yaml
```diff
@@ -0,0 +1,40 @@
...
+  rm -rf ./*.snap
+
+execute: |
+  # Unset SNAPCRAFT_BUILD_ENVIRONMENT=host.
+  unset SNAPCRAFT_BUILD_ENVIRONMENT
+
+  cd test-snap
+
+  # First with lxd
+  snapcraft
```
sergiusens, line 32: since you are running on all systems, you need to pass --use-lxd here as the default on older snapcrafts is multipass

sergiusens, line 32: since you are running on all systems, you need to pass --use-lxd here as the default on older snapcrafts is multipass

tests/unit/meta/test_snap_yaml.py
```diff
@@ -14,6 +14,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import os
```
sergiusens, line 17: unused

tests/spread/general/version-git/task.yaml
cmatsuoka: Correct. Fixing to use `--use-lxd`.

tests/unit/meta/test_snap_yaml.py
cmatsuoka: Removed.

sergiusens (approved):

PR #3810, opened 2022-06-24
extensions: method to get data dir

sergiusens:

-----
CRAFT-1115

### DIFF
--- a/snapcraft/extensions/extension.py
+++ b/snapcraft/extensions/extension.py
@@ -18,6 +18,8 @@
 
 import abc
 import os
+import sys
+from pathlib import Path
 from typing import Any, Dict, Optional, Tuple, final
 
 from craft_cli import emit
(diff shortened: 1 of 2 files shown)


mr-cal (approved):

sergio-costas:
The tests do fail: the folders doesn't exist.

sergio-costas:
In fact, it seems that the extension folder isn't created inside `/usr/share...` but directly in `/share':

imagen

So I think that the tests should be rewritten to compare the returned values with a fixed string (when running the tests it seems that sys.prefix() returns "/usr"). Also, the true folder seems to be different than the one returned by "sys.prefix()"...

sergiusens:
@sergio-costas the tests passed, here and locally, where do they fail?

sergiusens:
@sergio-costas 
image
image

sergio-costas:
Mmm... odd...

PR #3819, opened 2022-06-29
pack: specify snap filename

mr-cal:

-----
Allow snapcraft to specify the output snap name from the project's name, version, and target architecture.  

This will be used when building snaps where the target architecture is different than the host architecture.

Part 5 of supporting architectures for `core22`. Full code change is proposed [here](https://github.com/snapcore/snapcraft/pull/3712).

(CRAFT-1171)

### DIFF
--- a/snapcraft/pack.py
+++ b/snapcraft/pack.py
@@ -81,7 +81,7 @@ def _get_filename(
         if not output_path.is_dir():
             return output_path.name
 
-    if name is not None and version is not None and target_arch is not None:
+    if all(i is not None for i in [name, version, target_arch]):
         return f"{name}_{version}_{target_arch}.snap"
 
     return None
--- a/snapcraft/pack.py
+++ b/snapcraft/pack.py
@@ -41,48 +41,92 @@ def _verify_snap(directory: Path) -> None:
         raise errors.SnapcraftError(msg)
 
 
+def _get_directory(output: Optional[str]) -> Path:
+    """Get directory to output the snap file to.
+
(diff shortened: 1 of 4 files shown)


facundobatista (changes requested):
Logic looks great, added some details to improve though.

Thanks!

mr-cal:
Thanks for the feedback!  This PR went from a minor feature to a minor refactor (adding an extra if statement broke `pylint` complexity requirements, which forced me to break everything into functions, which then exposed the complexity).

</br>

> If output is None I would actually return pathlib.Path.cwd

Taking this advice simplified the logic quite a bit.

snapcraft/pack.py
```diff
@@ -41,48 +41,92 @@ def _verify_snap(directory: Path) -> None:
...
+    :param target_arch: Target architecture the snap project is built to.
+
+    :return: The filename of the snap file if output or name/version/target_arch are specified.
+    """
+    if output:
+        output_path = Path(output)
+        if not output_path.is_dir():
+            return output_path.name
+
+    if name is not None and version is not None and target_arch is not None:
```
sergiusens, line 84: I somewhat prefer
```suggestion
    if all(i is not None for i in [name, version, target_arch]):
```

sergiusens (approved):

snapcraft/pack.py
mr-cal: +1, more elegant

facundobatista (approved):
Awesome, thanks!

PR #3823, opened 2022-07-01
Add 'groups' statement to parts

sergio-costas:
This patch allows to add a 'groups' statement to each part,
define one or more groups for each one, and use the groups for
dependency management.

This way, it is possible to add all the binaries and libraries
parts to a "binaries" group, and make another part deppend on
that group (just using the 'after' statement like with a part
name).

This is mainly useful in big snapcraft.yaml files with a lot
of parts, like the gnome-XX-sdk ones.

It is implemented as a preprocessor, like the one used in C,
that modifies on-the-fly the configuration file.

-----

### DIFF
--- a/snapcraft/parts/lifecycle.py
+++ b/snapcraft/parts/lifecycle.py
@@ -26,7 +26,7 @@ import craft_parts
 from craft_cli import EmitterMode, emit
 from craft_parts import ProjectInfo, StepInfo, callbacks, infos
 
-from snapcraft import errors, extensions, pack, providers, utils
+from snapcraft import errors, extensions, pack, preprocessors, providers, utils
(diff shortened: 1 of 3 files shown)


sergiusens:
If we go forward with this, then it should be a craft-parts feature and not a snapcraft exclusive; I am not sold on the "groups" name either, would need to think about it

sergio-costas:
> If we go forward with this, then it should be a craft-parts feature and not a snapcraft exclusive; I am not sold on the "groups" name either, would need to think about it

But at least you like the concept itself. I'll give it another round of thinking, then.

sergio-costas:
Ok, moved into craft-parts... https://github.com/canonical/craft-parts/pull/241

PR #3831, opened 2022-07-07
project: use correct host arch when target arch is set

cmatsuoka:
The deb arch in project configuration is set to the target architecture
when --target-arch is used. Create a new project property holding the
host deb architecture instead of obtaining it from freshly instantiated
project options to avoid problems later when converting `ProjectOptions`
to a singleton.
    
-----

### DIFF
--- a/tests/spread/cross-compile/stage-packages/snap/snapcraft.yaml
+++ b/tests/spread/cross-compile/stage-packages/snap/snapcraft.yaml
@@ -18,6 +18,7 @@ parts:
       - xxd
     - to armhf:
       - tar:armhf
+    - on amd64 to armhf:
       - xxd
     - on amd64:
       - grep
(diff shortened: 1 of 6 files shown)


tests/spread/cross-compile/stage-packages/snap/snapcraft.yaml
```diff
@@ -18,6 +18,7 @@ parts:
       - xxd
     - to armhf:
       - tar:armhf
+    - on amd64 to armhf:
```
sergiusens, line 21: what is being tested or asserted here?

cmatsuoka: We want to make sure that "on <host arch> to <target arch>" works. The bug we fixed used target arch for both.

PR #3869, opened 2022-08-11
extensions: replace execs with source

sergio-costas:
Using source it is possible to simplify the chain, reduce memory usage and make execution faster.

This MR emulates "exec $@" using "source", thus avoiding to create a new process and launching a new shell each time a shell script must be executed from another shell script.

-----
DT-504

### DIFF
--- a/extensions/desktop/command-chain/run
+++ b/extensions/desktop/command-chain/run
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 if [ -z "$1" ]; then
   echo "run <command>"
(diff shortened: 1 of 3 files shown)


extensions/desktop/command-chain/run
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
sergiusens, line 1: isn't bash heavier?

sergio-costas: Yes, but you was already launching bash for hooks-configure-fonts and desktop-launch, so with this patch we are saving the two /bin/sh launched previously.

sergio-costas: Also, bash is required for those two scripts because they use some bash-specific syntax.

sergiusens (approved):

PR #3878, opened 2022-08-17
requirements: update craft-parts to 1.12.0

cmatsuoka:
Use craft-parts 1.12.0 with changes to the cmake plugin to fix the
prefix path and install prefix. This allows libraries built using
cmake and staged from a different part to be recognized without
defining cmake plugin parameters such as `CMAKE_INCLUDE_PATH` or
`CMAKE_LIBRARY_PATH`.

LP: #1978859

-----

### DIFF
--- a/tests/spread/extensions/gnome/task.yaml
+++ b/tests/spread/extensions/gnome/task.yaml
@@ -16,7 +16,6 @@ prepare: |
   . "$TOOLS_DIR/snapcraft-yaml.sh"
   set_base "$SNAP_DIR/snap/snapcraft.yaml"
   sed -i 's/gnome-3-28/gnome/' "$SNAP_DIR/snap/snapcraft.yaml"
-  sed -i 's/bin\/gtk3-hello/usr\/local\/bin\/gtk3-hello/' "$SNAP_DIR/snap/snapcraft.yaml"
 
   # Temporarily use snaps from edge
   echo '    build-snaps: [gnome-42-2204-sdk/latest/edge, gnome-42-2204/latest/edge]' >> "$SNAP_DIR/snap/snapcraft.yaml"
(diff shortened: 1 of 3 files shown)


facundobatista (approved):
+1

tests/spread/extensions/gnome/task.yaml
```diff
@@ -16,7 +16,6 @@ prepare: |
   . "$TOOLS_DIR/snapcraft-yaml.sh"
   set_base "$SNAP_DIR/snap/snapcraft.yaml"
   sed -i 's/gnome-3-28/gnome/' "$SNAP_DIR/snap/snapcraft.yaml"
-  sed -i 's/bin\/gtk3-hello/usr\/local\/bin\/gtk3-hello/' "$SNAP_DIR/snap/snapcraft.yaml"
```
sergiusens, line 19: this is not a hack, this most likely means we are changing the expected "command" entry for all snaps

sergiusens (changes requested):
I think the change means we need to look into the casualty this might create when rebuilding existing snaps

tests/spread/extensions/gnome/task.yaml
cmatsuoka: When we set `CMAKE_INSTALL_PREFIX` to empty, binaries are installed in `bin/` instead of `usr/local/bin/`. This matches the installation path used in the original v1 plugin for core18.

cmatsuoka: The safest solution in this case is to not set the install prefix in the plugin, but orient developers using base core22 (and core20) to add `-DCMAKE_INSTALL_PREFIX=` to `cmake-parameters` to have core18 plugin compatibility. In this case, we can close this PR and update craft-parts to 1.12.1 rolling back the install prefix change, but keeping the prefix path fix.

cmatsuoka:
Closing in favor of #3879.

PR #3884, opened 2022-08-24
cli: parse http-proxy and https-proxy

mr-cal:

-----
Parse the command line arguments `--http-proxy` and `--https-proxy`.

source: https://forum.snapcraft.io/t/error-unrecognized-arguments-http-proxy/30967

(CRAFT-1217)

### DIFF
--- a/snapcraft/commands/lifecycle.py
+++ b/snapcraft/commands/lifecycle.py
@@ -74,6 +74,18 @@ class _LifecycleCommand(BaseCommand, abc.ABC):
             default=os.getenv("SNAPCRAFT_BUILD_FOR"),
             help="Set target architecture to build for",
         )
+        parser.add_argument(
+            "--http-proxy",
+            type=str,
+            default=os.getenv("http_proxy"),
+            help="Set http proxy",
+        )
+        parser.add_argument(
+            "--https-proxy",
+            type=str,
+            default=os.getenv("https_proxy"),
+            help="Set https proxy",
+        )
 
         # --enable-experimental-extensions is only available in legacy
         parser.add_argument(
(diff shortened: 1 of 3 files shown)


tests/unit/cli/test_lifecycle.py
```diff
@@ -640,11 +748,17 @@ def test_lifecycle_command_pack_build_for(mocker):
...
                 target_arch=None,
                 provider=None,
+                http_proxy=None,
+                https_proxy=None,
             )
         )
     ]
 
 
+def test_lifecycle_command_http_proxy(mocker):
```
sergiusens, line 758: what's this for?

sergiusens:
This seems to enable psasthrough, but I thought we were going to enable the proxy for core22 as well

snapcraft/commands/lifecycle.py
```diff
@@ -74,6 +74,18 @@ def fill_parser(self, parser: "argparse.ArgumentParser") -> None:
...
+            type=str,
+            default=os.getenv("http_proxy"),
+            help="Set http proxy",
+        )
+        parser.add_argument(
+            "--https-proxy",
+            type=str,
+            default=os.getenv("https_proxy"),
+            help="Set https proxy",
+        )
```
cmatsuoka, line 88: Where are these values used? Are we just allowing the user to define them? (If the actual implementation is just setting the environment variable value and allowing it to pass to the inner instance I think we should do that)

sergiusens: this is for the legacy execution to work. I believe I will let this in, but we need a task to enable it for core22

mr-cal: Is this what you mean by "enabling the proxy"?
https://github.com/snapcore/snapcraft/blob/260214788af30ea055b3170afd97b91e556341ae/snapcraft_legacy/internal/build_providers/_base_provider.py#L510

tests/unit/cli/test_lifecycle.py
mr-cal: whoops, removed

sergiusens (approved):

PR #3886, opened 2022-08-25
ua: specify and enable ua services

cmatsuoka:
UA services can be specified in the project yaml file under the optional
`ua-services` key. These services will be enabled during the execution
of the parts lifecycle.

Note: this handles ua services for core22. Support for legacy snapcraft
will be added in a separate PR.

-----
CRAFT-1289

### DIFF
--- a/snapcraft/ua_manager.py
+++ b/snapcraft/ua_manager.py
@@ -19,7 +19,7 @@
 import contextlib
 import json
 import subprocess
-from typing import Any, Dict, Iterator, Optional
+from typing import Any, Dict, Iterator, List, Optional
 
 from craft_cli import emit
 from craft_parts.packages import Repository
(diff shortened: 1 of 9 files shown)


sergiusens:
this is core22 only, correct?

sergiusens (approved):

snapcraft/ua_manager.py
```diff
@@ -63,6 +73,17 @@ def _attach(ua_token: str) -> None:
         raise UATokenAttachError from error
 
 
+def _enable_services(services: List[str]) -> None:
+    """Enable the specified UA services."""
+    try:
+        with emit.open_stream("Enable UA services") as stream:
+            subprocess.check_call(
+                ["ua", "enable", *services, "--beta"], stdout=stream, stderr=stream
```
sergiusens, line 81: why are we hardcoding --beta here?

sergiusens: If the entire concept is beta, I would argue that this should require an experimental flag

cmatsuoka: It allows us to enable any service regardless of the beta status. Otherwise I can verify each service individually to see if beta is required using a hardcoded table lookup, and then add beta if necessary -- it's easier to just add it in all cases.

sergiusens: Just to follow up from chat, we agreed to use --enable-experimental-ua-services for now

cmatsuoka: Some services require `--beta`, others don't. ROS services, specifically, require it (the primary use of this feature will be to enable ROS services). Added the `--enable-experimental-ua-services` flag.

cmatsuoka:
> this is core22 only, correct?

This is core22, support for legacy will be added in a separate PR.

sergiusens (approved):

PR #3893, opened 2022-08-30
config: add snap config model

mr-cal:

-----
Allows Snapcraft to parse data set via `snap set snapcraft <key>=<value>`

Currently, the only supported parameter is `provider`, which will be used in an upcoming PR to allow users to choose a default provider.

(CRAFT-1321)


snap/hooks/configure
```diff
@@ -0,0 +1,11 @@
+#! /bin/bash
+
+config_file="$SNAP_DATA/snap-config.yaml"
+provider="$(snapctl get provider)"
+
+cat <<EOT > "${config_file}"
+# This file is generated automatically generated by snapcraft's configure hook.
```
sergiusens, line 7: double generated

snap/hooks/configure
```diff
@@ -0,0 +1,11 @@
+#! /bin/bash
+
+config_file="$SNAP_DATA/snap-config.yaml"
+provider="$(snapctl get provider)"
+
+cat <<EOT > "${config_file}"
```
sergiusens, line 6: instead of writing to a file why don't we call get in the relevant locations?

mr-cal: Because `snap get` requires root access

snap/hooks/configure
mr-cal: Fixed

snap/hooks/configure
mr-cal: Details for posterity:
### snap get
- always requires root access
### snapctl get
- doesn't require root access when executed inside a snap
- fails when executed outside a snap (like during a unit test...)

PR #3895, opened 2022-08-30
parts: allow debug shell after the lifecycle

cmatsuoka:
Shell into the instance if `--debug` is used and a packing error happens
after the execution of the parts lifecycle.

-----
CRAFT-1286

### DIFF
--- a/snapcraft/parts/lifecycle.py
+++ b/snapcraft/parts/lifecycle.py
@@ -490,9 +490,11 @@ def _run_in_provider(
         except subprocess.CalledProcessError as err:
             capture_logs_from_instance(instance)
             raise errors.SnapcraftError(
-                f"Failed to execute {command_name} in instance. "
-                "Run the same command again with --debug to shell into "
-                "the environment if you wish to introspect this failure."
+                f"Failed to execute {command_name} in instance.",
+                details=(
+                    "Run the same command again with --debug to shell into "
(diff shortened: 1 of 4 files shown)


cmatsuoka:
@hellsworth Could you verify if this addresses the `--debug` issues you had?

hellsworth:
@cmatsuoka Well maybe? I had saved off my test case in a local branch of the app that exhibited the failure after the lifecycle. But now when I do a clean build with this snapcraft, my snap builds. I've done a clean build a couple of times, both successfully building the snap.

hellsworth:
@sergiusens pointed out that there was a cmake issue fixed with respect to how libraries are found. My project absolutely uses cmake, so perhaps this fix got in the way.

snapcraft/parts/lifecycle.py
```diff
@@ -449,7 +490,9 @@ def _run_in_provider(
         except subprocess.CalledProcessError as err:
             capture_logs_from_instance(instance)
             raise errors.SnapcraftError(
-                f"Failed to execute {command_name} in instance."
+                f"Failed to execute {command_name} in instance. "
+                "Run the same command again with --debug to shell into "
```
sergiusens, line 494: this should use the extra params from CraftError which SnapcraftError inherits from

sergiusens (approved):

PR #3899, opened 2022-08-31
providers: choose default provider from snap config for core 22

mr-cal:

-----

The default provider can be set via `snapd`'s config.

It can be set with `snap set snapcraft provider=<provider>` where valid providers are 'lxd' and 'multipass'.
To unset, run `snap unset snapcraft provider`.

This PR depends on https://github.com/snapcore/snapcraft/pull/3898, so the tests will fail until that PR is merged.

(CRAFT-1322)

### DIFF
--- a/snapcraft/providers/_get_provider.py
+++ b/snapcraft/providers/_get_provider.py
@@ -34,7 +34,7 @@ def get_provider(provider: Optional[str] = None) -> Provider:
 
     To determine the appropriate provider,
     (1) use provider specified in the function argument,
-    (2) get the provider from the environment if valid,
+    (2) get the provider from the environment,
     (3) use provider specified with snap configuration,
     (4) default to platform default (LXD on Linux, otherwise Multipass).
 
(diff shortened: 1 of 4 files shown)


snapcraft/providers/_get_provider.py
```diff
@@ -28,40 +32,57 @@
...
+        chosen_provider = "lxd"
+    else:
+        emit.debug("Using default provider 'multipass' on non-linux system.")
+        chosen_provider = "multipass"
+
+    # ignore case
+    chosen_provider = chosen_provider.lower()
+
+    if chosen_provider not in ("lxd", "multipass"):
+        raise RuntimeError(f"unsupported provider specified: {chosen_provider!r}")
```
facundobatista, line 83: Too generic. As the value is wrong, what about using `ValueError`?

facundobatista (changes requested):
Great work!! Added a couple of comments, thanks!

facundobatista (approved):
Wonderful, thanks!!

snapcraft/providers/_get_provider.py
mr-cal: Used `ValueError`.  I also made the if statements more pythonic.

PR #3903, opened 2022-09-06
providers: choose legacy default provider from snap config

mr-cal:

-----
The default provider for `snapcraft_legacy` can be set via snapd's config.

It can be set with `snap set snapcraft provider=<provider>` where valid providers are 'lxd' and 'multipass'.
To unset, run `snap unset snapcraft provider`.

The legacy implementation is much smaller than what is implemented for core22 in https://github.com/snapcore/snapcraft/pull/3899.
This is because the code for core22 validates the snap config during snapcraft's configure hook.  The legacy implementation doesn't need to re-validate the data.  It can just assume the config's data is valid.

(CRAFT-1323)

### DIFF
--- a/requirements-devel.txt
+++ b/requirements-devel.txt
@@ -120,4 +120,4 @@ python-apt @ https://launchpad.net/ubuntu/+archive/primary/+sourcefiles/python-a
 PyNaCl==1.4.0; sys.platform != "linux"
 PyNaCl @ https://files.pythonhosted.org/packages/61/ab/2ac6dea8489fa713e2b4c6c5b549cc962dd4a842b5998d9e80cf8440b7cd/PyNaCl-1.3.0.tar.gz; sys.platform == "linux"
 setuptools==49.6.0
-pyinstaller==4.3; sys.platform == "win32"
+pyinstaller==4.10; sys.platform == "win32"
--- a/snapcraft_legacy/cli/_options.py
+++ b/snapcraft_legacy/cli/_options.py
@@ -285,20 +285,49 @@ def _sanity_check_build_provider_flags(build_provider: str, **kwargs) -> None:
 
 
 def get_build_provider(**kwargs) -> str:
-    """Get build provider and determine if running as managed instance."""
+    """Get the appropriate provider and determine if running as a managed instance.
 
+    To determine the appropriate provider:
+    (1) use provider specified by `--provider=<provider>`
+    (2) use provider from the environment variable `SNAPCRAFT_BUILD_ENVIRONMENT`
+        (`click` loads this env var into `--provider=<provider>`)
+    (3) use provider 'lxd' if `--use-lxd` was passed
+    (4) use provider 'host' if `--destructive-mode` was passed
+    (5) use provider 'host' if snapcraft is running as a process container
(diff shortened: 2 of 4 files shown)


requirements-devel.txt
```diff
@@ -120,4 +120,4 @@ python-apt @ https://launchpad.net/ubuntu/+archive/primary/+sourcefiles/python-a
 PyNaCl==1.4.0; sys.platform != "linux"
 PyNaCl @ https://files.pythonhosted.org/packages/61/ab/2ac6dea8489fa713e2b4c6c5b549cc962dd4a842b5998d9e80cf8440b7cd/PyNaCl-1.3.0.tar.gz; sys.platform == "linux"
 setuptools==49.6.0
-pyinstaller==4.3; sys.platform == "win32"
+pyinstaller==4.10; sys.platform == "win32"
```
mr-cal, line 123: This old version of `pyinstaller` was causing the [appveyor test](https://ci.appveyor.com/project/sergiusens/snapcraft/builds/44697101) to fail.  The [changelog](https://pyinstaller.org/en/v5.3/CHANGES.html#id54) says improvements were made in 4.4 to metadata collection.  I think that was the root cause.

Anyways, I bumped it to the latest 4.x version.  `pyinstaller` 5.x will require some tweaks to appveyor.yml.

mr-cal:
Unit test failure looks like an unrelated issue for downloading a nodejs tarball.  I'll look at that separately.

snapcraft_legacy/cli/_options.py
```diff
@@ -285,21 +285,50 @@ def _sanity_check_build_provider_flags(build_provider: str, **kwargs) -> None:
...
+    (6) use provider specified by snap config
+    (7) else default to multipass
+
+    :return: provider name
+    :raises ValueError: if provider is not supported
+    """
+    # load provider from snap config
+    snap_config = common.get_snap_config()
+    snap_provider = snap_config.get("provider") if snap_config else None
+
```
facundobatista, line 306: Bad combinations of parameters are handled elsewhere, or we should handle them here?

I mean, passing `--provider` and `--use-lxd` together should be an error. Or `--destructive-mode` with any of those.

Note that having the envvar is fine with any of those options, command line should override environmental config. Probably `click` is messing everything here.

facundobatista (changes requested):
Added a question, something to explore. Thanks!

snapcraft_legacy/cli/_options.py
mr-cal: Those cases are checked in [`_sanity_check_build_provider_flags()`](https://github.com/snapcore/snapcraft/blob/74f1b86351d34288f16295a6257c33ca9a811e5b/snapcraft_legacy/cli/_options.py#L237), which is called at the end of this function.

The checks are not very restrictive - your example `snapcraft --provider=lxd --use-lxd` is valid.  
It does throw an error when `--destructive-mode` is used with `--provider=<provider>` or `--use-lxd`.

facundobatista: Awesome, thanks!

facundobatista (approved):
+1

PR #3906, opened 2022-09-08
providers: pass http-proxy and https-proxy to craft-providers

mr-cal:

-----

Follow up to https://github.com/snapcore/snapcraft/pull/3884.  This PR allows proxies to be used in core22 by passing `http_proxy` and `https_proxy` to `craft-providers` as environmental variables.

`craft-providers` sets the snapd proxy in [this PR](https://github.com/canonical/craft-providers/pull/149).

I tested this manually.  I used working http and https proxies from [this site](https://www.freeproxylists.net/) and checked they were set in the instance.  Then I chose an invalid proxy and received a network error inside the instance when snapcraft attempted to download and install a snap.

(CRAFT-1334)

### DIFF
--- a/snapcraft/parts/lifecycle.py
+++ b/snapcraft/parts/lifecycle.py
@@ -525,8 +525,8 @@ def _run_in_provider(
         bind_ssh=parsed_args.bind_ssh,
         build_on=project.get_build_on(),
         build_for=project.get_build_for(),
-        http_proxy=getattr(parsed_args, "http_proxy", None),
-        https_proxy=getattr(parsed_args, "https_proxy", None),
+        http_proxy=parsed_args.http_proxy,
+        https_proxy=parsed_args.https_proxy,
     ) as instance:
         try:
             with emit.pause():
--- a/snapcraft/parts/lifecycle.py
+++ b/snapcraft/parts/lifecycle.py
@@ -525,6 +525,8 @@ def _run_in_provider(
         bind_ssh=parsed_args.bind_ssh,
         build_on=project.get_build_on(),
         build_for=project.get_build_for(),
+        http_proxy=getattr(parsed_args, "http_proxy", None),
+        https_proxy=getattr(parsed_args, "https_proxy", None),
     ) as instance:
         try:
             with emit.pause():
--- a/snapcraft/providers/_provider.py
+++ b/snapcraft/providers/_provider.py
@@ -74,8 +74,17 @@ class Provider(ABC):
         """
 
     @staticmethod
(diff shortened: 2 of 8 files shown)


mr-cal:
Code coverage about 90% 🥳 🥳 🥳

snapcraft/providers/_provider.py
```diff
@@ -74,8 +74,17 @@ def ensure_provider_is_available(cls) -> None:
         """
 
     @staticmethod
-    def get_command_environment() -> Dict[str, Optional[str]]:
-        """Construct the required environment."""
+    def get_command_environment(
+        http_proxy: Optional[str] = None,
+        https_proxy: Optional[str] = None,
```
sergiusens, line 79: we should add no_proxy too

snapcraft/parts/lifecycle.py
```diff
@@ -525,6 +525,8 @@ def _run_in_provider(
         bind_ssh=parsed_args.bind_ssh,
         build_on=project.get_build_on(),
         build_for=project.get_build_for(),
+        http_proxy=getattr(parsed_args, "http_proxy", None),
+        https_proxy=getattr(parsed_args, "https_proxy", None),
```
cmatsuoka, line 529: Can `parsed_args` reach this point with these attributes undefined?

sergiusens (approved):

snapcraft/providers/_provider.py
mr-cal: From our discussion:
1. If the environmental variable `no_proxy` is defined on the host system, it gets passed to the instance [here](https://github.com/snapcore/snapcraft/blob/7cc2a529e1e694b5ba72af5e516dac0ec6c793fd/snapcraft/providers/_provider.py#L86).  I tested and confirmed this.
2. `--no-proxy` is not a command line option, which is why I didn't add extra handling for it in this PR

snapcraft/parts/lifecycle.py
mr-cal: No, they should always exist in `parsed_args`.  If the user didn't pass them as command line arguments nor their environmental variables, then `argparse` instantiates them with a value of `None`.

Even if they didn't happen to exist in `parsed_args`, `getattr` still returns the default `None`.

cmatsuoka: If the attribute is always there, getting it using `parsed_args.http_proxy` could be more natural and easier to read (like `parsed_args.bind_ssh` a few lines above).

mr-cal: Makes sense.  I switched to your suggestion.

PR #3913, opened 2022-09-10
nil.py: enhance the explanation of when the nil plugin is useful

rpjday:
The current docstring explanation is incomplete, and ignores the scenario where a part with a nil plugin can still define a source.

No functional change.

Signed-off-by: Robert P. J. Day <robert.day@canonical.com>

-----


cmatsuoka (approved):
Thanks, I'll copy this description to the nil plugin in craft-parts.

cmatsuoka (changes requested):
Actually there's an inaccurate statement in the description, even if the nil plugin is used the `build/` directory is still populated (but it's true that nothing will be added to `install/`).

rpjday:
OK, I believe it's correct now.

PR #3931, opened 2022-10-04
linters: allow ignoring files per-linter

tigarmo:
This commit restructures the lint.ignore project section to allow ignoring files on a per-linter basis. This still lets the user disable all issues for a given linter, but allows for finer-grained control.

This is the current scheme (before this PR):

```yaml
lint:
  ignore:
    linters:
      - library
      - classic
    files:
      - "elf.*"
      - lib/libfoo.so.1
```

The two entries (`linters` and `files`) are unrelated and complementary: If `linters` has e.g. `library`, then all `library` linting is ignored, regardless of `files`. If `files` has e.g. "elf.*", then any linting issue found matching that glob will be ignored, regardless of `linters`.

This is the new scheme implemented in this PR:

```yaml
lint:
  ignore:
    library:
      - lib/libfoo.so.1     # Disable "library" linter only for lib/libfoo.so.1
    classic:                # Disable "classic" linter for all files
```
Since the way that this is described in the Pydantic model is a bit "less-than-intuive" in the way that `classic` being a key with `None` as value actually means "all" files, I added a couple of methods to use as the "public" API to hopefully hide that complexity a bit.

### DIFF
--- a/snapcraft/linters/linters.py
+++ b/snapcraft/linters/linters.py
@@ -40,7 +40,7 @@ if TYPE_CHECKING:
 LinterType = Type[Linter]
 
 
-_LINTERS: Dict[str, LinterType] = {
+LINTERS: Dict[str, LinterType] = {
     "classic": ClassicLinter,
     "library": LibraryLinter,
 }
@@ -137,8 +137,8 @@ def run_linters(location: Path, *, lint: Optional[projects.Lint]) -> List[Linter
         snap_metadata = snap_yaml.read(Path())
 
         emit.progress("Running linters...")
-        for name, linter_class in _LINTERS.items():
-            if lint and name in lint.ignore.linters:
+        for name, linter_class in LINTERS.items():
+            if lint and lint.all_ignored(name):
                 continue
 
             linter = linter_class(name=name, lint=lint, snap_metadata=snap_metadata)
@@ -157,17 +157,21 @@ def _ignore_matching_filenames(
     issues: List[LinterIssue], *, lint: Optional[projects.Lint]
 ) -> None:
     """Mark any remaining filename match as ignored."""
-    if lint and lint.ignore.files:
-        for issue in issues:
-            for pattern in lint.ignore.files:
-                if (
-                    issue.filename
-                    and issue.result != LinterResult.IGNORED
-                    and fnmatch.fnmatch(issue.filename, pattern)
-                ):
-                    emit.verbose(
-                        f"Ignore {issue.name!r} linter issue ({issue.filename!r} "
-                        f"matches {pattern!r})"
-                    )
-                    issue.result = LinterResult.IGNORED
-                    break
+    if lint is None:
+        return
+
+    for issue in issues:
+
+        files = lint.ignored_files(issue.name)
+        for pattern in files:
+            if (
+                issue.filename
+                and issue.result != LinterResult.IGNORED
+                and fnmatch.fnmatch(issue.filename, pattern)
+            ):
+                emit.verbose(
+                    f"Ignore {issue.name!r} linter issue ({issue.filename!r} "
+                    f"matches {pattern!r})"
+                )
+                issue.result = LinterResult.IGNORED
+                break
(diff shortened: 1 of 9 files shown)


snapcraft/linters/linters.py
```diff
@@ -157,17 +157,21 @@ def _ignore_matching_filenames(
...
-                    and issue.result != LinterResult.IGNORED
-                    and fnmatch.fnmatch(issue.filename, pattern)
-                ):
-                    emit.verbose(
-                        f"Ignore {issue.name!r} linter issue ({issue.filename!r} "
-                        f"matches {pattern!r})"
-                    )
-                    issue.result = LinterResult.IGNORED
-                    break
+    if lint is None:
```
sergiusens, line 160: with the removal of files is it possible for lint to be None now?

snapcraft/projects.py
```diff
@@ -180,17 +180,38 @@ def _validate_list_stream(cls, listen_stream):
         return listen_stream
 
 
-class LintIgnore(ProjectModel):
-    """Lists of specific linters or globbed filenames to ignore."""
+class Lint(ProjectModel):
+    """Linter configuration.
+
+    :ivar ignore: A dict describing which files should have issues ignored for given linters.
+        The keys in the dict are the names of known linters (currently "library" and "classic"):
```
sergiusens, line 187: for maintenance, it might be better to make https://github.com/tigarmo/snapcraft/blob/38d3b738b3d946db4799092b71ffb71256cb6d0f/snapcraft/linters/linters.py#L43 public and refer to that here (instead of the "currently" list)

tests/unit/linters/test_base.py
```diff
@@ -57,3 +58,27 @@ def test_linter_issue_string_no_url(self, linter_issue):
     def test_linter_issue_string_filename_no_url(self, linter_issue):
         issue = linter_issue(filename="foo.txt", url=None)
         assert f"{issue}" == "test: foo.txt: Linter message text"
+
+
+def test_lint_ignore_api():
+    """Test projects.Lint "ignore"-related API."""
```
sergiusens, line 64: the API is rather easy to follow, but this test is hard to unpack. Maybe splitting it up a bit can add clarity

sergiusens (commented):
Looks good!

snapcraft/linters/linters.py
tigarmo: yes: It's a bit confusing but `lint` here refers to the `lint:` key in the project, so it's still optional.
The function that calls this is called by lifecycle like this:
```python
    if command_name in ("pack", "snap"):
        issues = linters.run_linters(lifecycle.prime_dir, lint=project.lint)
        status = linters.report(issues, intermediate=True)

File: tests/unit/linters/test_base.py (hunk 3 above)
Comment (tigarmo) [reply tests/unit/linters/test_base.py]: Hmm yes I see what you mean. I split this and reworded the contents a bit, please take a look

File: tests/unit/linters/test_base.py [hunk 4]
```diff
@@ -60,25 +62,37 @@ def test_linter_issue_string_filename_no_url(self, linter_issue):
...
-    # We want to ignore *no* issues from this linter.
-    ignore_none = "ignore_none"
+    The data specifies that all issues for linter "linter1" should be ignored, and that the
+    issues for "linter2" should be ignored if they refer to files matching "file1" and/or
+    "/lib/file2*".
+    """
+    yaml_data = textwrap.dedent(
+        """
+        ignore:
+          linter1:          # ignore all issues for "linter1"
```
sergiusens, line 76: I am glad you did this yaml, I missed the part where it is not a list of one key/value dict or a scalar like:
```
lint:
  ignore:
  - library:
    - lib/libfoo.so.1
  - classic  # disables all classic
 ```
Which I think might be more user friendly

tigarmo: Updated to a list. To simplify the actual implementation of the api I ended up adding a 'private' attribute that reworks the list as a dict internally.

sergiusens (approved):

PR #3932, opened 2022-10-04
providers: refactor `launched_environment()`

mr-cal:
snapcraft-specific code in `launched_environment()` is removed from the craft-providers interface.

-----
### Overview

Removes snapcraft-specific logic from `launched_environment()`.  This function is now matches the implementation in rockcraft ([lxd](https://github.com/canonical/rockcraft/blob/97306bc3d93c8a41f18432b1dadf4569c7f28854/rockcraft/providers/_lxd.py#L99) and [multipass](https://github.com/canonical/rockcraft/blob/97306bc3d93c8a41f18432b1dadf4569c7f28854/rockcraft/providers/_multipass.py#L89))

This is the most complex PR of the series, due to the translation of bases.  The base translation was added in [this PR](https://github.com/canonical/rockcraft/pull/85) to rockcraft.

#### Background
*craft applications have a concept of a "base". This base is translated a `craft-providers` base and is passed to `craft-providers` so it can choose which base image (OS and release) to use for the instance.

#### Problem

Currently, each application translates its own base inside `_lxd.py` and `_multipass.py`. This application-specific code needs to be removed from these files.

#### Solution

To support this, I created a list of what bases craft-providers can accept. Right now, `18.04`, `20.04`, and `22.04` are supported. Other OSs and releases can be added in the future without breaking backwards compatibility.

The image below shows how an application can translates its base to a `craft-providers` base. It also shows `craft-providers` translating it to the specific name needed for LXD and Multipass to download the base image.

image

### Note about the changeset
This PR shows ` 10 files changed, 768 insertions(+), 274 deletions(-)`, but it's much smaller.  It's really `8 files changed, 164 insertions(+), 122 deletions(-)`.

This is because I copied over rockcraft's files `tests/unit/providers/test_{lxd,multipass}.py`.  These files have already been reviewed and are going to be migrated to `craft-providers` later this week.  Thus, these unit test files probably don't need intense scrutiny.

### DIFF
--- a/snapcraft/parts/lifecycle.py
+++ b/snapcraft/parts/lifecycle.py
@@ -39,7 +39,11 @@ from snapcraft.projects import (
     Project,
 )
 from snapcraft.providers import capture_logs_from_instance
-from snapcraft.providers.providers import get_instance_name
+from snapcraft.providers.providers import (
+    SNAPCRAFT_BASE_TO_PROVIDER_BASE,
+    get_base_configuration,
+    get_instance_name,
+)
 from snapcraft.utils import (
     convert_architecture_deb_to_platform,
     get_host_architecture,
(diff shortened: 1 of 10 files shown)


facundobatista (approved):
Looks solid.

PR #3936, opened 2022-10-05
lint: support the existence of 'venv'

tigarmo:
I think 'venv' is a very common virtual-env name, so add it to gitignore and have codespell skip it. I'm actually not sure if this type of change is desirable, so feel free to reject it!


mr-cal (approved):
I don't see any issues with this.

I support ignoring `.venv` too, I think that's pretty common too (and my default).

PR #3942, opened 2022-10-06
commands: allow re-editing validation sets on error

sergiusens:
Validation sets on error used to not provide a way to re-edit if the set
 was rejected. Prompt the user if they want to re-edit the set after
 showing the error.

 The change here is the while loop, _generate_template and
 _submit_validation_set have been extracted from the run method.

-----
CRAFT-1388

### DIFF
--- a/snapcraft/commands/validation_sets.py
+++ b/snapcraft/commands/validation_sets.py
@@ -28,9 +28,9 @@ import yaml
 from craft_cli import BaseCommand, emit
 from overrides import overrides
 
-import snapcraft_legacy.storeapi.errors
 from snapcraft import errors, utils
 from snapcraft_legacy._store import StoreClientCLI
+from snapcraft_legacy.storeapi.errors import StoreValidationSetsError
 
 if TYPE_CHECKING:
     import argparse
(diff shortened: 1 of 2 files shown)


snapcraft/commands/validation_sets.py
```diff
@@ -87,50 +88,81 @@ def run(self, parsed_args: "argparse.Namespace"):
...
+
+        while True:
+            try:
+                _submit_validation_set(
+                    edited_validation_sets, parsed_args.key_name, store_client
+                )
+                break
+            except snapcraft_legacy.storeapi.errors.StoreValidationSetsError as validation_error:
+                emit.message(str(validation_error))
+                if not utils.prompt("Do you with to amend the validation set? "):
```
tigarmo, line 112:
```suggestion
                if not utils.prompt("Do you wish to amend the validation set? "):
```

tests/unit/commands/test_validation_sets.py
```diff
@@ -229,3 +232,177 @@ def test_edit_validation_sets_with_changes_to_existing_set(
...
+    validation_sets_payload,
+    fake_dashboard_get_validation_sets,
+    fake_dashboard_post_validation_sets_build_assertion,
+    fake_dashboard_post_validation_sets,
+    fake_snap_sign,
+    fake_edit_validation_sets,
+    monkeypatch,
+):
+    fake_edit_validation_sets.return_value = {
+        "account-id": "AccountIDXXXOfTheRequestingUserX",
```
tigarmo, line 248: This one's just a suggestion: the checks that the tests do are themselves fine, but I think the amount of "hardcoded" text makes it hard to see what's actually being tested. For example, consider putting this dict in a variable that you can use both here and in the `assert fake_dashboard_post_validation_sets_build_assertion.mock_calls` check

tigarmo (changes requested):
Nice, thanks. The test comment is just a suggestion and I'll leave it up to you, but we should fix that typo on the prompt

mr-cal (approved):
Looks good!

tests/unit/commands/test_validation_sets.py
sergiusens: sounds like a good suggestion, so yes, I will do this.

tigarmo (approved):
Leaving this approved to not delay the merge, thanks!

PR #3973, opened 2022-11-03
kernel plugin: do not copy extra-modules.conf

mbd98:
Fixes failure to copy due to non-existent source file. Source file no longer exists upstream, and intended target already exists. See https://github.com/snapcore/core-initrd/tree/main/modules

-----

### DIFF
--- a/snapcraft/parts/plugins/kernel.py
+++ b/snapcraft/parts/plugins/kernel.py
@@ -861,14 +861,6 @@ class KernelPlugin(plugins.Plugin):
                     ],
                 ),
                 "done",
-                " ".join(
-                    [
-                        "cp",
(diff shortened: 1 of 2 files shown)


sergiusens (approved):

sergiusens (changes requested):
tests have not been updated

sergiusens (approved):

kubiko:
@sergiusens @mbd98 I'm late here and I will revert this PR in #3998
Anyway, this PR might have fixed a build error if there are not defined kernel modules to be included in the initramfs, but it also causes regression, that if there are defined kernel modules to be included, now they are not.
This change also renders for loop in `_make_initrd_cmd` useless. That loop populates `${initramfs_ko_modules_conf}` with requested kernel modules. If we do not copy the populated file, those won't be included.

The correct fix is to always create an empty file.

PR #3980, opened 2022-11-10
plugins: Increase kernel plugin build environments comparability

jocado:
Increase kernel plugin build environments comparability by using port 80 for keyserver

This is a suggest that would really help me, but I'm not sure if there is a good reason not too.  

@kubiko ?

-----


julian-klode:
(I think I said most of that on internal chat yesterday, but for the record:)
Presumably this was supposed to say compatibility and not comparability, as in, you run behind a firewall/proxy that blocks the normal hkp port.

My suggestion of course is to use https, and get rid of the apt-key, and just store a key file.

This can be done multiple ways, e.g. using python-apt:

```
import apt_pkg
import os
os.makedirs("/etc/apt/keyrings", exist_ok=True)
apt_pkg.init()
acq = apt_pkg.Acquire()
file = apt_pkg.AcquireFile(acq, f"https://https://keyserver.ubuntu.com/pks/lookup?search={fingerprint}&fingerprint=on&op=get", "/etc/apt/keyrings/snappy-dev.asc")
acq.run()
```

or using apt-helper

```
import apt_pkg
import subprocess
os.makedirs("/etc/apt/keyrings", exist_ok=True)
subprocess.check_call(["/usr/lib/apt/apt-helper", "download-file",  f"https://https://keyserver.ubuntu.com/pks/lookup?search={fingerprint}&fingerprint=on&op=get", "/etc/apt/keyrings/snappy-dev.asc"])
```

and then setting `[signed-by=/etc/apt/keyrings/snappy-dev.asc]` in the sources.list entry.

These two make it use apt's proxy configuration even, which should be enabling for most people (some might have a more restrictive 
proxy for caching purposes while they have full internet access).

And /etc/apt/keyrings and signed-by follows the new best practices (the directory comes by default since 22.04).

Both of these snippets can be improved upon by creating a temporary directory owned by _apt, downloading in there (changing the target filename), and then moving the file to /etc/apt/keyrings and chown it to root. This enables the download to run as the user _apt and other sandboxing mechanisms that don't work when writing to /etc/apt/keyrings directly.

julian-klode:
Oh a fully tested example, with sandboxing, but wrong key fingerprint of course (it's my personal key):

```
import apt_pkg
import apt.progress.text
import os
import shutil
import tempfile

apt_pkg.init()

fingerprint = "0x" + "AEE1 C8AA AAF0 B768 4019  C546 021B 361B 6B03 1B00".replace(" ", "")

with tempfile.TemporaryDirectory() as tmpdir:
   shutil.chown(tmpdir, "_apt", "root")
   acq = apt_pkg.Acquire(apt.progress.text.AcquireProgress())
   file = apt_pkg.AcquireFile(acq, f"https://keyserver.ubuntu.com/pks/lookup?search={fingerprint}&fingerprint=on&op=get", destfile=f"{tmpdir}/snappy-dev.asc")
   res = acq.run()

   if file.error_text:
     raise ValueError("Could not fetch file: {}".format(file.error_text))

   if res != apt_pkg.Acquire.RESULT_CONTINUE:
     raise ValueError("Could not fetch file: Unknown error")

   os.makedirs("/etc/apt/keyrings", exist_ok=True)
   shutil.move(f"{tmpdir}/snappy-dev.asc", "/etc/apt/keyrings/snappy-dev.asc")
```

So shiny. Might need a os.chmod() if we want to ignore umask so the key has the right permissions - don't know.

jocado:
> (I think I said most of that on internal chat yesterday, but for the record:)

Hi - not sure if that message was meant for me - I don't recall having any chats relating to this yesterday :slightly_smiling_face: 

With this MR, I was just trying to add a very low impact approach to fixing the issue around commonly allowed egress destination ports.

I very much appreciate the input, but I wasn't looking to rework the implementation much, just increase compatibility as it stands.

PR #3981, opened 2022-11-10
commands, lifecycle: provide snapcraft try for core22

tigarmo:

-----
This commit adds support for ``snapcraft try`` for core22 snaps. Just
like in the legacy version, this works by creating a ``prime`` directory
on the host that is then mounted into the ``/root/prime`` directory
inside the instance. The lifecycle then runs normally and the project is
primed but not packed.

CRAFT-1121


snapcraft/parts/lifecycle.py
```diff
@@ -556,6 +559,13 @@ def _run_in_provider(
             providers.capture_logs_from_instance(instance)
 
 
+def _expose_prime(project_path: Path, instance: Executor):
+    """Expose the instance's prime directory in ``project_path`` on the host."""
+    host_prime = project_path / "prime"
+    host_prime.mkdir(exist_ok=True)
+    instance.mount(host_source=project_path / "prime", target=Path("/root/prime"))
```
cmatsuoka, line 566: We can use `utils.get_managed_environment_home_path()` instead of hard-coding `/root` (to prevent problems in case we change the work dirs location inside the instance), and obtain the prime directory name by creating a `craft_parts.ProjectDirs` object passing the managed environment home path as the work dir.

snapcraft/commands/lifecycle.py
```diff
@@ -296,3 +296,16 @@ class CleanCommand(_LifecycleStepCommand):
         remove the managed snap packing environment (VM or container).
         """
     )
+
+
+class TryCommand(_LifecycleCommand):
```
cmatsuoka, line 301: Are all `_LifecycleCommand` options applicable to the `try` command? I suppose `--destructive-mode` works because in this case the `prime` directory is created directly on the host and there's no need no expose it?

tigarmo: > Are all _LifecycleCommand options applicable to the try command? 

I think so, because `try` really is `prime + expose the prime dir` right? So any parameter that applies to `prime` could apply to `try` too. I can narrow it down if you think this is too broad though

cmatsuoka (approved):
Thanks!

PR #3983, opened 2022-11-10
repo: raise error when populating the apt cache directory fails

mr-cal:

-----

If the contents of `/etc/apt` cannot be copied to the cache directory, snapcraft will now produce a helpful error.

This can occur on a system where Ubuntu Pro is enabled, snapcraft is run in destructive mode, and snapcraft is not being run with sudo nor as root.

(CRAFT-1421)

### DIFF
--- a/snapcraft_legacy/internal/repo/apt_cache.py
+++ b/snapcraft_legacy/internal/repo/apt_cache.py
@@ -120,7 +120,15 @@ class AptCache(ContextDecorator):
 
         # Copy current cache configuration.
         cache_etc_apt_path.parent.mkdir(parents=True, exist_ok=True)
-        shutil.copytree("/etc/apt", cache_etc_apt_path)
+
+        # systems with ubuntu pro have an auth file in /etc/apt readable only by root
+        # this may raise an error when copying /etc/apt into the cache directory
+        try:
+            shutil.copytree("/etc/apt", cache_etc_apt_path)
+        except shutil.Error as error:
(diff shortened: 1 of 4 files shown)


snapcraft_legacy/internal/repo/apt_cache.py
```diff
@@ -120,7 +120,15 @@ def _populate_stage_cache_dir(self) -> None:
 
         # Copy current cache configuration.
         cache_etc_apt_path.parent.mkdir(parents=True, exist_ok=True)
-        shutil.copytree("/etc/apt", cache_etc_apt_path)
+
+        # systems with ubuntu pro have an auth file in /etc/apt readable only by root
+        # this may raise an error when copying /etc/apt into the cache directory
```
facundobatista, line 125: Code looks ok, and approving it, but this comment needs to be improved: it's not `/etc/apt` that is not readable; the directory itself is readable just fine (otherwise `copytree` will fail differently) but the files inside are only readable by root).

Furthermore, to think about: do we want to support other errors here? (e.g. `/etc/apt` being non readable, which will fail with a "standard" PermissionError).

facundobatista (approved):

cmatsuoka (approved):

snapcraft_legacy/internal/repo/apt_cache.py
mr-cal: - [X] improved the comments
- [X] caught and tested `PermissionError` when `/etc/apt` is unreadable

PR #3984, opened 2022-11-14
Fixed StoreLegacyRegisterKeyCommand's overview.

facundobatista:


snapcraft/commands/legacy.py
```diff
@@ -222,7 +222,7 @@ class StoreLegacyRegisterKeyCommand(LegacyBaseCommand):
     help_msg = "Register a key to sign assertions with the Snap Store."
     overview = textwrap.dedent(
         """
-        Register a a key with the Snap Store. Prior to registration, use register-key
+        Register a a key with the Snap Store. Prior to registration, use create-key
```
tigarmo, line 225:
```suggestion
        Register a key with the Snap Store. Prior to registration, use create-key
```

tigarmo (changes requested):

snapcraft/commands/legacy.py
facundobatista: Good catch! Improved.

tigarmo (approved):

PR #3994, opened 2022-11-25
core22 colcon plugin: re-source environment after installation

Guillaumebeuzeboc:

-----

Core22 ROS 2 `colcon` snaps couldn't build packages creating custom ROS 2 interfaces. It was sourcing the environment before downloading the dependencies. Some dependencies (`rosidl`) have specificities when sourced, hence sourcing before downloading them causes issues.

In the case of ROS 2 message, the `PYTHONPATH` only had the `site-packages` subdir and not the `dist-packages` subdir added by `rosidl`.

This issue couldn't appear with `lxd` spread tests because the `lxd` environment was not cleaned properly by `colcon` spread tests (Debian packages were still installed, hence the dependencies were mostly already installed).

Additionally, the advanced `colcon` spread tests had no entry in the `spread.yaml` and were never called.

Hence, this PR does:
- Add missing spread entry for `colcon` specific tests
- Add `package-utils.sh` to restore the state of Debian package installed before a spread tests
- Add a `colcon` spread test with a message package to highlight that there is a bug
- Fix the bug by sourcing the ROS 2 environment after installing the ROS 2 dependencies


Guillaumebeuzeboc:
@artivis you probably want to review that

spread.yaml
```diff
@@ -490,6 +490,10 @@ suites:
    summary: tests of snapcraft's craft-part's based plugins
    systems:
    - ubuntu-22.04*
+ tests/spread/plugins/craft-parts/colcon/:
```
sergiusens, line 493: instead of this, can you layout the tests to follow the pattern on https://github.com/snapcore/snapcraft/tree/main/tests/spread/plugins/craft-parts/build-and-run-hello ?

tests/spread/plugins/craft-parts/colcon/colcon-packages-ignore/task.yaml
```diff
@@ -18,6 +18,9 @@ prepare: |
   . "$TOOLS_DIR/snapcraft-yaml.sh"
   set_base "../snaps/$SNAP/snap/snapcraft.yaml"
 
+  . "$TOOLS_DIR/package-utils.sh"
```
sergiusens, line 22: mind adding this to the foreach entries in the main spread.yaml?

sergiusens:
@Guillaumebeuzeboc this looks pretty good, but I wanted to clarify, the LXD environment Snapcraft creates should be clean, I believe the issue here is that they are running in destructive mode

tests/spread/plugins/craft-parts/colcon/colcon-packages-ignore/task.yaml
Guillaumebeuzeboc: I can give it a shot, but I am scared that some tests might just break. I will let you know.

Guillaumebeuzeboc:
> @Guillaumebeuzeboc this looks pretty good, but I wanted to clarify, the LXD environment Snapcraft creates should be clean, I believe the issue here is that they are running in destructive mode

@sergiusens I tried to use LXD with the ROS spread test (with core22). My issue was that when I run my spread tests locally on my machine, it was failing due to LXD not having nested LXD enable. I couldn't find a way to enable it.

If you know a way to do so, indeed, using LXD on every ROS test is a better solution.

spread.yaml
Guillaumebeuzeboc: Done in https://github.com/snapcore/snapcraft/pull/3994/commits/6f689c8643241d135d15cc7652272588ee459be7

tests/spread/plugins/craft-parts/colcon/colcon-packages-ignore/task.yaml
Guillaumebeuzeboc: @sergiusens, do you recommand me to break the test and to fix them later, or should I use the "apt packages breakpoint" only for the ROS tests like it is right now?

artivis (approved):
lgtm

sergiusens (approved):
live reviewed

PR #3998, opened 2022-12-06
Kernel plugin: multiple fixes

kubiko:
Multiple 
- fixing a regression introduced by #3973. Please do not remove features if not sure about their functionality
- fix crash if there are no defined kernel modules to be included in initramfs
- improve documentation of multiple options by clarifying default values
- fix warning and occasional error reported when linking initramfs overlays with wild cards
- updated unit tests for the above changes


cmatsuoka (approved):
Thanks for the fixes!

PR #4003, opened 2022-12-14
Tests: Make test_patcher not require a real patchelf in /usr/bin

lengau:
Discovered this accidentally thanks to a `~/.bashrc` that put `/bin` before `/usr/bin` in `$PATH`.

I'm not entirely sure this was unintentional, so unless someone else knows we should probably wait for @sergiusens.

### DIFF
--- a/tests/unit/elf/test_patcher.py
+++ b/tests/unit/elf/test_patcher.py
@@ -190,7 +190,6 @@ def test_patcher_patch_rpath_already_set(mocker, patcher, elf_file):
     )
 
     patcher.patch(elf_file=elf_file)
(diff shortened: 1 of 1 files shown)


sergiusens (approved):

tests/unit/elf/test_patcher.py
```diff
@@ -187,10 +190,11 @@ def test_patcher_patch_rpath_already_set(mocker, patcher, elf_file):
     )
 
     patcher.patch(elf_file=elf_file)
+    # raise Exception(os.environ['PATH'])
```
sergiusens, line 193:
```suggestion
```

PR #4004, opened 2022-12-14
Suggest an arguably simpler alternative

nicolasbock:
The dev container can also be accessed via `su - ubuntu`. This suggestion does not have a lot of practical relevance, the command is slightly shorter though :)

Signed-off-by: Nicolas Bock <nicolas.bock@canonical.com>

-----

### DIFF
--- a/tools/environment-setup.sh
+++ b/tools/environment-setup.sh
@@ -42,4 +42,4 @@ lxc exec snapcraft-dev -- sudo -iu ubuntu bash -c \
     "echo 'source /home/ubuntu/.venv/snapcraft/bin/activate' >> .bashrc"
 
 echo "Container ready, enter it by running: "
-echo "lxc exec snapcraft-dev -- sudo -iu ubuntu bash"
+echo "lxc exec snapcraft-dev -- su --login ubuntu"


tools/environment-setup.sh
```diff
@@ -42,4 +42,4 @@ lxc exec snapcraft-dev -- sudo -iu ubuntu bash -c \
     "echo 'source /home/ubuntu/.venv/snapcraft/bin/activate' >> .bashrc"
 
 echo "Container ready, enter it by running: "
-echo "lxc exec snapcraft-dev -- sudo -iu ubuntu bash"
+echo "lxc exec snapcraft-dev -- su - ubuntu"
```
sergiusens, line 45: Thanks for the contribution. The man page for `su` states:

> It is recommended to always use the --login option (instead of its shortcut -) to avoid side effects caused by mixing environments.

nicolasbock: Oh sure, I prefer long options as well. I should have suggested `--login` initially.

sergiusens:
after some analysis I have decided not to move forward with this to avoid any surprises

sergiusens:
thanks!

PR #3519, opened 2021-05-07
requirements: correct setup.py install_requires and add tooling to freeze reqs

cjp256:
- Add tools/freeze-requirements.sh script to automatically update
  requirements.txt and requirements-devel.txt.

- Add make target to freeze-requirements.

- Update setup.py install_requires list and extend with [dev] extras.

- Set pinned pip version in environment-setup-local.sh.

Signed-off-by: Chris Patterson <chris.patterson@canonical.com>

- [ ] Have you followed the [guidelines for contributing](https://github.com/snapcore/snapcraft/blob/master/CONTRIBUTING.md)?
- [ ] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [ ] Have you successfully run `./runtests.sh static`?
- [ ] Have you successfully run `./runtests.sh tests/unit`?

-----

PR #4005, opened 2022-12-16
Testing: Lint and test using Tox

lengau:
Configures tox for linting and unit testing and provides a GitHub workflow for running Tox.

Does not disable old linting and testing yet for the sake of comparison.

(CRAFT-1513)


.github/workflows/tox.yaml
```diff
@@ -0,0 +1,89 @@
+name: Tox
+on:
+  push:
+    branches:
+      - "main"
+      - "snapcraft/7.0"
+      - "release/*"
+      - "hotfix/*"
```
mr-cal, line 8: What about `feature/*` for long-lived feature branches?  (I'm thinking of the `feature/kernel` branch)

.github/workflows/tox.yaml
```diff
@@ -0,0 +1,89 @@
...
+        with:
+          fetch-depth: 0
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - name: Install dependencies
+        run: |
+          echo "::group::Begin snap install"
+          echo "Installing snaps in the background while running apt and pip..."
```
mr-cal, line 26: Clever!

requirements-devel.txt
```diff
@@ -1,22 +1,24 @@
 astroid==2.13.2
 attrs==22.2.0
 black==22.12.0
+cachetools==5.2.0
 catkin-pkg==0.5.2
-certifi==2022.12.7
+certifi==2022.9.24
```
mr-cal, line 6: Are these "downgrades" are just from the need to rebase the PR?

setup.py
```diff
@@ -82,6 +83,7 @@ def recursive_data_files(directory, install_directory):
     "pytest-mock",
     "pytest-subprocess",
     "ruff",
+    "tox>=4.0",
```
mr-cal, line 86: Is this supposed to align with the version in the tox.yaml workflow?
`python -m pip install 'tox>=4.0.11<5.0' tox-gh`

setup.py
```diff
@@ -126,8 +128,13 @@ def recursive_data_files(directory, install_directory):
 ]
 
 try:
-    os_release = open("/etc/os-release").read()
-    ubuntu = "ID=ubuntu" in os_release
+    ubuntu = bool(
```
mr-cal, line 131: :thinking: I think I know why you were looking at the `os_release` functions in 3.11

setup.cfg
```diff
@@ -45,3 +40,141 @@ ignore = E203,E501
...
+    unit = py{38,310}-{withreq,noreq}-unit
+    # Legacy unit tests only
+    legacy = py{38,310}-{withreq,noreq}-legacy
+set_env =
+    SNAPCRAFT_IGNORE_YAML_BINDINGS: "1"
+
+[testenv]
+deps =
+    withreq,pylint,mypy,pyright: -r{tox_root}/requirements-devel.txt
+    noreq: PyNaCl>=1.5.0
```
mr-cal, line 72: If I understand correctly, `tox` always needs PyNaCl?  Or snapcraft needs it?

setup.cfg
```diff
@@ -45,3 +40,141 @@ ignore = E203,E501
...
 ignore_decorators = overrides
+
+[tox:tox]
+min_version = 4.0
+env_list =
+    # Parametrized environments.
+    # First parameter allows us to choose Python 3.8 or 3.10.
+    # Second parameter chooses how to define the environment:
+    #  withreq: using requirements-devel.txt
+    #  noreq: Without either requirements file (but including dev requirements)
```
tigarmo, line 51: what's the reasoning behind this 'noreq' env? Did this concept exist before?

setup.cfg
```diff
@@ -45,3 +40,141 @@ ignore = E203,E501
...
+# Note: this does not include `snapcraft_legacy` as it contains several files that need reformatting.
+commands = black setup.py snapcraft tests
+
+[testenv:format-ruff]
+description = Autoformat with ruff
+labels = fix
+deps = ruff
+skip_install = true
+commands =
+    ruff --fix setup.py snapcraft tests tools
```
tigarmo, line 102: is `--fix` safe here?

.github/workflows/tox.yaml
lengau: I can definitely add whatever we want here - I just copied this from `tests.yaml`.

`feature/*` is probably something we'd want in general on all our projects, so it might be better to have that discussion next week when I create the template repo?

setup.cfg
```diff
@@ -45,3 +40,141 @@ ignore = E203,E501
...
+    ruff --fix --config snapcraft_legacy/ruff.toml snapcraft_legacy tests/legacy
+
+[testenv:pylint]
+description = Lint with pylint
+labels = lint
+# This runs all commands even if the first fails.
```
tigarmo, line 108: :+1:

requirements-devel.txt
lengau: Other than `snap-helpers` and `craft-providers` I think they're all related to specifications from tox - certainly `certifi` is. Once we get renovate everywhere this should be more reliable.

setup.py
lengau: Yes/no? I wanted to be more explicit in the yaml file that we don't want to do a major release upgrade, but here I was trying to fit with our current theme of the least restrictive possible requirements - at the time I wrote this, 4.0 were all still considered prerelease so `pip install tox` would install 3.x. It looks like as of 4.1 that's no longer the case, but also I know for a fact that our config doesn't work with 3.x.

lengau: Long term I plan on having Renovate update the versions in the yaml file as well though :-)

setup.py
lengau: Yeah and as it turns out it's added in [3.10](https://docs.python.org/3/library/platform.html#platform.freedesktop_os_release), not 3.11 as I thought.

With 3.10 we can say:

```python
import platform
ubuntu = "ubuntu" in  platform.freedesktop_os_release().get("ID_LIKE", "").split() + [platform.freedesktop_os_release()["ID"]]
```

Likewise this dict contains everything we need for `os_release.py`, although we could also replace that with a dependency on [`distro`](https://github.com/python-distro/distro), which itself has no dependencies outside the standard library.

setup.cfg
lengau: Snapcraft was having breakages without PyNaCl. I'm not sure why as we don't explicitly require it. I would like at some point to tackle removing this and updating to the latest version of cryptography, but it was out of scope here. #3519 is the relevant PR where this was added, and I'm guessing that upgrading to the latest cryptography will let us remove this.

lengau: I've created a task for this at CRAFT-1569

setup.cfg
lengau: It was helpful for me while implementing tox to figure out why some things were breaking, so it's new coming from that. By the time I'd figured out the issues, having it had become sufficiently useful that I kept it around.

setup.cfg
lengau: The `format` environments here actually run the formatting and are for local use rather than CI use. You can either run `tox -e format-ruff` to do the ruff formatting or (preferred) run `tox -m fix` to run all the auto-fixing tools (including black, as done above).

The CI runs the `lint` label, which includes `black` with `--check` and `ruff` without `--fix`.

My hope is that since `black` and `ruff` are so fast, we can eventually incorporate these into a pre-commit hook (assuming everyone's on board with that).

.github/workflows/tox.yaml
mr-cal: Sounds good!

mr-cal (approved):
Great work!

tigarmo (approved):
nice!

PR #4011, opened 2023-01-09
requirements: update dependencies

tigarmo:
- [ ] Have you followed the [guidelines for contributing](https://github.com/snapcore/snapcraft/blob/master/CONTRIBUTING.md)?
- [ ] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [ ] Have you successfully run `make lint`?
- [ ] Have you successfully run `pytest tests/unit`?

-----

PR #4012, opened 2023-01-09
linters: migrate to ruff

tigarmo:
Most of the configuration in `pyproject.toml` comes from Alex's work in craft-parts. I had to add a separate `ruff.toml` in `snapcraft_legacy` to have legacy-specific configuration (basically just flake8). Output looks like this:

image

The PR is probably easier to review per-commit; the first commit has deps update from the #4011 PR because this branch started off that one.

(CRAFT-1505)


snapcraft_legacy/ruff.toml
```diff
@@ -0,0 +1,30 @@
...
+    ".vscode",
+    "_build",
+    "buck-out",
+    "__pycache__",
+    "build",
+    "dist",
+    "parts",
+    "stage",
+    "prime",
+]
```
lengau, line 19: Any reason to do this rather than use `extend-exclude` and rely on the [default configuration](https://github.com/charliermarsh/ruff#configuration)?

```suggestion
extend-exclude = [
    ".vscode",
    "__pycache__",
    "parts",
    "stage",
    "prime",
]
```

Not much of a difference, just curious.

pyproject.toml
```diff
@@ -51,3 +51,96 @@ load-plugins = "pylint_fixme_info,pylint_pytest"
...
+    "B009",  # Allow calling `getattr` in tests since it can be used to make the test clearer.
+    "S103", # Allow `os.chmod` setting a permissive mask `0o555` on file or directory
+    "S105", # Allow Possible hardcoded password.
+    "S106", # Allow Possible hardcoded password.
+    "S108", # Allow Probable insecure usage of temporary file or directory.
+]
+"__init__.py" = ["I001"]  # Imports in __init__ filesare allowed to be out of order
+
+[tool.ruff.flake8-annotations]
+suppress-none-returning = true  # We don't need to explicitly point out that a function doesn't return anything
```
lengau, line 146:
```suggestion
suppress-none-returning = true  # We don't need to explicitly point out that a function doesn't return anything

```
\n

lengau (approved):
LGTM

mr-cal (approved):
Looks great!

snapcraft_legacy/ruff.toml
tigarmo: no reason! This file was generated with `flake8-to-ruff`. Your way makes more sense, so updated
thanks

lengau: Oh right - `flake8-to-ruff` would make it explicit so you don't have subtle changes. That's smart.

PR #4014, opened 2023-01-11
snap: add remove hook to delete base instances and base images

mr-cal:

-----

When snapcraft is uninstalled, the base instances and images for LXD are removed.

(CRAFT-1488)


tests/spread/core22/remove-hook/task.yaml
```diff
@@ -0,0 +1,61 @@
...
+  set_base "snap/snapcraft.yaml"
+
+restore: |
+  #shellcheck source=tests/spread/tools/snapcraft-yaml.sh
+  . "$TOOLS_DIR/snapcraft-yaml.sh"
+  restore_yaml "snap/snapcraft.yaml"
+
+  # reinstall snapcraft
+
+  # If $SNAPCRAFT_CHANNEL is defined, install snapcraft from that channel.
```
mr-cal, line 15: Copied verbatim from `spread.yaml`'s project `prepare`.   I think I should move this to a function and call it to avoid duplication.

snap/hooks/remove
```diff
@@ -0,0 +1,50 @@
+#! /usr/bin/env bash
```
sergiusens, line 1: did you run shellcheck against this?

We probably want to add it to our shellcheck target

tests/spread/tools/prepare.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash -e
```
sergiusens, line 1: same shellcheck comment

snap/hooks/remove
mr-cal: Yes and it's already checked by `make test-shellcheck`

tests/spread/tools/prepare.sh
mr-cal: And same

snap/hooks/remove
sergiusens, line 1: ok, the only problem I can see with this is that we are assuming bash, and bash may or may not be installed on the host, we should normalize on sh and remove bashisms

mr-cal: Converted from bash to sh :heavy_check_mark:

sergiusens (approved):

PR #4017, opened 2023-01-18
commands: clean all parts without removing the instance

sergio-costas:
This MR adds the option of using

    snapcraft clean-parts

to clean all the parts in a single step. This allows to fully rebuild the snap, but without having to recreate the virtual machine or the container, like when using "snapcraft clean". This saves a lot of time.

-----

### DIFF
--- a/debian/snapcraft-completion
+++ b/debian/snapcraft-completion
@@ -6,7 +6,7 @@ _snapcraft()
     COMPREPLY=()
     cur="${COMP_WORDS[COMP_CWORD]}"
     prev="${COMP_WORDS[COMP_CWORD-1]}"
-    opts="help init list-plugins plugins login logout export-login list-keys keys create-key register-key register registered list-registered push release clean cleanbuild clean-parts pull build sign-build stage prime snap update define search gated validate history status close enable-ci expand-extensions extension extensions list-extensions"
+    opts="help init list-plugins plugins login logout export-login list-keys keys create-key register-key register registered list-registered push release clean cleanbuild pull build sign-build stage prime snap update define search gated validate history status close enable-ci expand-extensions extension extensions list-extensions"
 
     case "$prev" in
     help)
--- a/debian/snapcraft-completion
+++ b/debian/snapcraft-completion
@@ -6,7 +6,7 @@ _snapcraft()
     COMPREPLY=()
     cur="${COMP_WORDS[COMP_CWORD]}"
     prev="${COMP_WORDS[COMP_CWORD-1]}"
-    opts="help init list-plugins plugins login logout export-login list-keys keys create-key register-key register registered list-registered push release clean cleanbuild pull build sign-build stage prime snap update define search gated validate history status close enable-ci expand-extensions extension extensions list-extensions"
+    opts="help init list-plugins plugins login logout export-login list-keys keys create-key register-key register registered list-registered push release clean cleanbuild clean-parts pull build sign-build stage prime snap update define search gated validate history status close enable-ci expand-extensions extension extensions list-extensions"
 
     case "$prev" in
     help)
--- a/snapcraft/cli.py
+++ b/snapcraft/cli.py
@@ -40,6 +40,7 @@ COMMAND_GROUPS = [
         "Lifecycle",
         [
             commands.CleanCommand,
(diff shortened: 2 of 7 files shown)


cmatsuoka:
I think this functionality is useful but I would prefer to invoke it as an option to clean, something like `snapcraft clean --keep-environment` or similar.

tigarmo:
I like this idea! I feel like `snapcraft clean _all` is a bit too magical and dissonant with the rest of the commands, how about something like `snapcraft clean --all-parts`?

sergio-costas:
Using an extra parameter seems a good idea... I'll check if I'm able to do it.

cmatsuoka:
> I also like the idea! How about something like `snapcraft clean --all-parts`? That feels very explicit in intent

It's clear in the sense that all parts are cleaned up, but that's also the effect if you remove the entire environment. Something that says the environment will be kept will not have such ambiguity.

sergio-costas:
> > I also like the idea! How about something like `snapcraft clean --all-parts`? That feels very explicit in intent
> 
> It's clear in the sense that all parts are cleaned up, but that's also the effect if you remove the entire environment. Something that says the environment will be kept will not have such ambiguity.

I did the change to support a parameter. Now the only thing that remains is to decide which parameter to use :-D

sergio-costas:
Ok, changed to "--keep-environment", and also allows to use "-k".

cmatsuoka:
Many command-line switches in snapcraft have only the long form, I would leave this one without the short version as well.

sergio-costas:
> Many command-line switches in snapcraft have only the long form, I would leave this one without the short version as well.

I know, but in this specific case I think that it's convenient to have both: the long one for scripts, where you want to clearly see what you are doing, and the short one for the command line, when you just want to clean it but not write "El Quijote" :-)

The other option is to use instead a new command instead, like "clean-parts" or something like that...

sergio-costas:
Ok, changed to "clean-parts". I think that it is more clear and consistent now, while keeping the usability.

cmatsuoka:
I think we should keep adding this functionality to the existing "clean" command instead of adding a different one that does something very similar. I see that the temptation to use something expressing "clean all parts" makes sense if you look at the implementation where you essentially add all parts to the list of parts to be cleaned, but from the user's point of view cleaning all parts is the expected behavior of running `snapcraft clean` without specifying parts, and having an option or additional command expressing that sounds confusing and redundant in my opinion (it's like running `snapcraft do-something --really-do-something`).

It's important, however, that whatever solution we go for, we should also add an equivalent implementation to core18/core20 so that commands will be valid and consistent regardless of the base being used -- we already have a similar problem with verbosity flags and this is currently being addressed.

sergio-costas:
I put the change in a different patch to allow to revert it easily. I'll do what you decide.

cmatsuoka:
Since Snapcraft already uses "Launching instance" when it creates the build environment, maybe `--keep-instance` could be a more consistent choice.

sergio-costas:
Done.

sergio-costas:
@cmatsuoka BTW: how do I have to do to add this in legacy mode?

sergio-costas:
Added legacy support.

sergio-costas:
Oh my...!!!! All checks passed!!! :-)

cmatsuoka:
The next Snapcraft release should have significant improvements in the time needed to spin up the build instance. We decided to put this feature on hold until we evaluate the performance gain and see if this or other alternatives will still be needed, but in the meantime you can mimic its behavior by shelling into the build instance and running `snapcraft clean`.

sergio-costas:
I'm currently using this python script... It does the same: https://github.com/sergio-costas/cleanparts

cmatsuoka:
Some notes from the discussion on this feature:

- We should use only the long option, short options are reserved only for very well-known cases such as `-v` for verbose or `-h` for help.
- The rationale for not adding this now is to avoid implementing a feature that could make it *too easy* for users to create snaps without properly cleaning the environment.
- We need to add tests.
- We also need to add a warning text about the potential consequences of building using a dirty instance.
- If needed, `snapcraft pull --shell` and `snapcraft clean` inside the instance will have the same effect.

PR #4024, opened 2023-01-27
cli: pass `--verbose` and `-v` to snapcraft_legacy

mr-cal:

-----

This PR allows `--verbose` and `-v` to be used when building core18|20 snaps.

Support for other verbosity options (`-q`, `--quiet`, `--verbosity=<verbosity level>`) can be added in the same manner.

(CRAFT-1292)

### DIFF
--- a/snapcraft_legacy/cli/_options.py
+++ b/snapcraft_legacy/cli/_options.py
@@ -72,50 +72,50 @@ _HIDDEN_PROVIDERS = ["managed-host"]
 _ALL_PROVIDERS = _SUPPORTED_PROVIDERS + _HIDDEN_PROVIDERS
 _PROVIDER_OPTIONS: List[Dict[str, Any]] = [
     dict(
-        param_decls="--target-arch",
+        param_decls=["--target-arch"],
         metavar="<arch>",
         help="Target architecture to cross compile to",
         supported_providers=["host", "lxd", "multipass"],
     ),
     dict(
-        param_decls="--debug",
+        param_decls=["--debug"],
         is_flag=True,
         help="Shells into the environment if the build fails.",
         supported_providers=["host", "lxd", "managed-host", "multipass"],
     ),
     dict(
-        param_decls="--shell",
+        param_decls=["--shell"],
         is_flag=True,
         help="Shells into the environment in lieu of the step to run.",
         supported_providers=["host", "lxd", "managed-host", "multipass"],
     ),
     dict(
-        param_decls="--offline",
+        param_decls=["--offline"],
         is_flag=True,
         help="Operate in offline mode.",
         envvar="SNAPCRAFT_OFFLINE",
         supported_providers=["host", "lxd", "managed-host", "multipass"],
     ),
     dict(
-        param_decls="--shell-after",
+        param_decls=["--shell-after"],
         is_flag=True,
         help="Shells into the environment after the step has run.",
         supported_providers=["host", "lxd", "managed-host", "multipass"],
     ),
     dict(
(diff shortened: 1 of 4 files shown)


tests/legacy/unit/commands/test_verbosity.py
```diff
@@ -0,0 +1,64 @@
...
+        with mock.patch("snapcraft_legacy.internal.log.configure") as mock_configure:
+            result = self.run_command(args=["-v", "pull"])
+
+        self.assertThat(result.exit_code, Equals(0))
+        mock_configure.assert_called_once_with(log_level=logging.INFO)
+
+    def test_verbosity_verbose_long_and_debug_error(self):
+        """Test --verbose and --enabled-developer-debug cannot be used together."""
+        result = self.run_command(
+            args=["--verbose", "--enable-developer-debug", "pull"]
```
sergiusens, line 55: what about `args=["pull", "--verbose"]`?

mr-cal: It works great when using snapcraft at the command line but fails in the unit test:
```
Try 'run pull -h' for help.

Error: No such option: --verbose
```

I think I need to do 2 things:
1. Write the unit tests with pytest rather than using the existing `CommandBaseTest` fixtures.
2. Write spread tests

snapcraft_legacy/cli/_options.py
```diff
@@ -354,27 +363,28 @@ def get_build_provider_flags(build_provider: str, **kwargs) -> Dict[str, str]:
...
+            key_formatted = _param_decls_to_kwarg(key)
+            if is_flag and not kwargs.get(key_formatted):
+                continue
 
-        # Add build provider flag using envvar as key.
-        if envvar is not None and key_formatted in kwargs:
-            build_provider_flags[envvar] = kwargs[key_formatted]
+            # Add build provider flag using envvar as key.
+            if envvar is not None and key_formatted in kwargs:
+                build_provider_flags[envvar] = kwargs[key_formatted]
```
mr-cal, line 387: This new code has a side effect of adding both `v` and `verbose` to the `build_provider_flags`.  I think this is harmless, but the proper solution would be to discard any single character flags (keep `verbose` and discard `v`).  I can add this on Monday.

snapcraft_legacy/cli/_options.py
```diff
@@ -354,27 +363,30 @@ def get_build_provider_flags(build_provider: str, **kwargs) -> Dict[str, str]:
...
+        keys: List[str] = option["param_decls"]  # type: ignore
         is_flag: bool = option.get("is_flag", False)  # type: ignore
         envvar: Optional[str] = option.get("envvar")  # type: ignore
         supported_providers: List[str] = option["supported_providers"]  # type: ignore
 
-        # Skip --provider option.
-        if key == "--provider":
-            continue
+        for key in keys:
+            # TODO: skip single character flags (e.g. keep `verbose` but discard `v`)
```
mr-cal, line 372: This new code has a side effect of adding both `v` and `verbose` to the `build_provider_flags`. I think this is harmless, but the proper solution would be to discard any single character flags (keep verbose and discard v). 

I can add this on Monday to this PR or follow up in a subsequent PR.

sergiusens (approved):

snapcraft_legacy/cli/_options.py
mr-cal: `-v` and `--verbose` are not actually being added to `build_provider_flags`.  Added a docstring to clarify this in https://github.com/snapcore/snapcraft/pull/4025.

PR #4028, opened 2023-02-06
linters: add linter check for unused libraries

mr-cal:

-----

This PR adds a new linter warning when a library is packaged into a snap but not used.

The changes are split into 5 smaller commits.

These changes are inspired by some of the excellent work in https://github.com/snapcore/snapcraft/pull/3413

(CRAFT-1588)


snapcraft/linters/library_linter.py
```diff
@@ -73,6 +75,24 @@ def run(self) -> List[LinterIssue]:
                 issues=issues,
             )
 
+            # if the elf file is a library, get its resolved path
+            if elf_file.soname or self._is_library_path(elf_file.path):
+                all_libraries.add(str(elf_file.path.resolve()))
+
+            # get paths to the local libraries used by the elf file
+            used_libraries.update(
+                elf_file.load_dependencies(
```
cmatsuoka, line 84: Can we obtain the resolved path from the existing list of dependencies instead of computing it all over again?

cmatsuoka (commented):
Thanks, I see that obtaining the list of unused libraries is not trivial! Added a few questions about possible ways to obtain results faster and in a more accurate way.

snapcraft/linters/library_linter.py
```diff
@@ -116,3 +136,71 @@ def _check_dependencies_satisfied(
...
+            )
+            issues.append(issue)
+
+        return issues
+
+    def _is_library_path(self, path: Path) -> bool:
+        """Check if a file is in a library directory.
+
+        A library directory is a directory whose pathname ends in `lib/` or
+        `lib/<architecture-triplet>/` (i.e. lib/x86_64-linux-gnu/`)
```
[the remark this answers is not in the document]
mr-cal: Great (and challenging) question!

`ld --verbose` gives a nice set of library paths:
```
>>> re.findall(r'SEARCH_DIR\("=(.+?)"\);', subprocess.check_output(["ld", "--verbose"]).decode())

/lib
/lib/x86_64-linux-gnu
/lib64
/usr/lib
/usr/lib/x86_64-linux-gnu
/usr/lib/x86_64-linux-gnu64
/usr/lib64
/usr/local/lib
/usr/local/lib/x86_64-linux-gnu
/usr/local/lib64
/usr/x86_64-linux-gnu/lib
/usr/x86_64-linux-gnu/lib64
```

`ldconfig --verbose` returns a different list of directories:
```
/lib
/lib32
/lib/i386-linux-gnu
/lib/x86_64-linux-gnu
/usr/lib/x86_64-linux-gnu/libfakeroot
/usr/local/lib
```

Unlike `ldd`, `ld --verbose` and `ldconfig --verbose` ignore paths in `LD_LIBRARY_PATH`, like `/root/prime/usr/lib`.  Sergio mentioned parsing `LD_LIBRARY_PATH` is out of scope for the first implementation and can be a future improvement.

Possible options:
- [ ] Only include directories from the existing [`get_common_ld_library_paths()`](https://github.com/snapcore/snapcraft/blob/709e16c781246cc6aa6a5d926ea3ee22a95c3c7c/snapcraft/utils.py#L331) function
- [ ] Add a TODO for a future improvement to parse library paths from both `LD_LIBRARY_PATH` and `ld --verbose`
- [ ] Extend the regex in this new function to include directories like `lib64/`, `lib32/`, and  `x86_64-linux-gnu64`

My vote is for options B and C.

snapcraft/linters/library_linter.py
```diff
@@ -116,3 +136,71 @@ def _check_dependencies_satisfied(
...
+    ) -> List[LinterIssue]:
+        """Get a list of unused library issues.
+
+        :param all_libraries: a set of paths to all libraries
+        :param used_libraries: a set of libraries used by elf files in the snap
+
+        :returns: list of LinterIssues for unused libraries
+        """
+        issues: List[LinterIssue] = []
+        unused_libraries = all_libraries.difference(used_libraries)
```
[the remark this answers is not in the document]
mr-cal: Nice catch!

snapcraft/linters/library_linter.py
mr-cal: Yes.  It took awhile for me to refactor and eliminate the second computation.  The code is simpler now too.

snapcraft/linters/library_linter.py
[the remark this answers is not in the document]
cmatsuoka: Yes, we leave that for a future improvement. For now it's ok if we don't list some unused libraries if we hit a corner case, but we shouldn't list a library that's actually used as unused.

snapcraft/linters/library_linter.py
```diff
@@ -98,21 +116,92 @@ def _check_dependencies_satisfied(
...
+
+        if not path.is_file():
+            return False
+
+        # TODO: get library directories from LD_LIBRARY_PATH and `ld --verbose`
+        # instead of matching against patterns
+        if path.match("lib/*") or path.match("lib32/*") or path.match("lib64/*"):
+            return True
+
+        for architecture in elf_utils.ARCH_CONFIG.values():
```
cmatsuoka, line 203: Ideally we should keep `ARCH_CONFIG` internal to elf_utils, maybe we could add a function that returns a list of valid triplets to use here?

mr-cal: Made a function to return valid triplets and wrote unit tests.

cmatsuoka (approved):
Thanks!

PR #4044, opened 2023-02-26
plugins: add snapcraft python plugin

cmatsuoka:
Subclass the generic Python plugin provided by Craft-parts and add Snapcraft-specific logic to set venv symlinks according to the runtime environment.

-----
CRAFT-1630


snapcraft/parts/lifecycle.py
```diff
@@ -271,6 +271,8 @@ def _run_command(
         work_dir=work_dir,
         assets_dir=assets_dir,
         base=project.get_effective_base(),
+        project_base=project.base or "",
```
cmatsuoka, line 274: This is needed because craft-parts expects `base` to be the effective base (i.e. build-base if base is bare).

snapcraft/parts/parts.py
```diff
@@ -102,6 +104,9 @@ def __init__(
                 project_name=project_name,
                 project_vars_part_name=adopt_info,
                 project_vars=project_vars,
+                # custom arguments
+                project_base=project_base,
```
cmatsuoka, line 108: This is needed because Craft-parts expects `base` to be the effective base (i.e. build-base if base is bare).

cmatsuoka:
Rebased on main.

tests/spread/plugins/craft-parts/build-and-run-hello/python-hello/snap/snapcraft.yaml
```diff
@@ -0,0 +1,16 @@
...
+version: "1.0"
+summary: simple python application
+description: build a python application using core22
+base: core22
+confinement: strict
+
+apps:
+  python-hello:
+    command: bin/hello
+    extensions: [gnome]
```
lengau, line 11: Do you intend to include the gnome extension here, since it's included below in the gnome-extension tests?

tigarmo (approved):
looks good!

tests/spread/plugins/craft-parts/python-symlinks/task.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+restore: |
+  cd "${SNAP}"
+  snapcraft clean
+  rm -f ./*.snap
+
+execute: |
+  cd "${SNAP}"
+
+  if [ -f expected-symlink ]; then
```
sergiusens, line 36: I don't understand how this would work with not provisioned corner cases

sergiusens (changes requested):
This looks really good, the tests I feel need a little work. Since we are using a spread variant, it would be really nice to see the summary and description of the snapcraft.yaml describe the expected results and what is tests (considering some have and some don't have the expected symlink). It looks to me that we are testing that classic is broken when python is not provisioned. Clarity on the objective of the test would be nice to see.

tests/spread/plugins/craft-parts/python-symlinks/task.yaml
cmatsuoka: Could you provide an example of such a corner case?

snapcraft/parts/parts.py
cmatsuoka: I tried other names and found them to be slightly confusing. The value originates from `project.base` so keeping the project prefix name makes the variable instantly recognizable.

sergiusens (approved):

PR #4046, opened 2023-03-01
tools: move to the starbase tox configuration

lengau:

-----


tox.ini
```diff
@@ -0,0 +1,144 @@
+# Differences from the starbase tox.ini:
```
mr-cal, line 1: +1 for noting departures from starbase

tox.ini
```diff
@@ -0,0 +1,144 @@
...
+# * Uses requirements-devel.txt
+# * No library code for pyright type checking
+# * Documentation envs removed
+# * Pytest uses wheel package
+# * Legacy tests instead af integration tests
```
mr-cal, line 6: `af` -> `of`

tox.ini
```diff
@@ -0,0 +1,144 @@
...
+# * Additional test environment "noreq" runs all the tests without requirements.txt
+# * lint-shellcheck includes spread-shellcheck.py
+# * Additional ruff configuration for snapcraft_legacy
+# * Include isort
```
mr-cal, line 10: Curious - isort is not in starbase because ruff implements it?

tox.ini
```diff
@@ -0,0 +1,144 @@
...
+    codespell[tomli]==2.2.2
+    # renovate: datasource=pypi
+    yamllint==1.29.0
+    # renovate: datasource=pypi
+    isort==5.11.4
+env_dir = {work_dir}/linting
+runner = ignore_env_name_mismatch
+
+[shellcheck]
+find = git ls-files
```
mr-cal, line 94: full circle!

tox.ini
```diff
@@ -0,0 +1,144 @@
...
+filter = file --mime-type -Nnf- | grep shellscript | cut -f1 -d:
+
+[testenv:lint-{black,ruff,shellcheck,codespell,yaml}]
```
mr-cal, line 97: should `isort` be on this line?

Probably out-of-scope, but isort took a long time:
```
lint-isort: commands[2]> isort --check .
Skipped 71959 files
  lint-isort: OK (78.08=setup[0.01]+cmd[0.12,0.02,77.93] seconds)
  congratulations :) (78.39 seconds)
```

tox.ini
mr-cal, line 1: Do I need to install `tox-ignore-env-name-mismatch` manually?  Should it be in `requirements-devel.txt`?

All my tox runs log:
 `ROOT: will run in automatically provisioned tox, host /home/developer/dev/snapcraft/venv/bin/python is missing [requires (has)]: tox-ignore-env-name-mismatch==0.2.0.post2`

mr-cal, line 1: Do you know if the code coverage drop is related?

I know that git actions (rebase, push, merge) can confuse the code coverage.

.github/workflows/tox.yaml
```diff
@@ -49,17 +49,21 @@ jobs:
       fail-fast: false  # Run all the tests to their conclusions.
       matrix:
         platform: [ubuntu-20.04, ubuntu-22.04]
+        python_version: ["3.8", "3.10"]
```
mr-cal, line 52: Remind me to update which Github workflows are required when this is ready to merge.

tox.ini
mr-cal, line 97: Is there a tox command that I can use to replace `make lint`?

mr-cal (commented):
Looks great!  I have a few questions, a few minor typos, and some requests that can either be ignored or deferred.

tox.ini
mr-cal: Ah, and `tox -e isort` and `tox -e lint-isort` are both accepted by tox :confused:

.yamllint.yaml
```diff
@@ -0,0 +1,29 @@
+# Variances from starbase:
+# * Thruthiness checks set to warning
```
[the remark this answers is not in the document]
lengau: Believe it or not that's what yamllint calls it :-) It's for [alllllll the possible strings that become booleans](https://yaml.org/type/bool.html).

tox.ini
lengau: It should reuse the same `[project_dir]/.tox/.tox` environment. If we want to put in dev requiremnts I'm happy to do so, but tbh you don't even need to be in a venv to run the tox stuff. With this change you can even just use tox from apt regardless of the version and it'll provision all the environments for you. 

(Which also means we could even remove tox from the dev requirements...)

tox.ini
lengau:
```suggestion
# * Legacy tests instead of integration tests
```

tox.ini
lengau: Yes. Ruff's isort rules are... *slightly* different from our isort configuration here (stricter, so running ruff with isort enabled wants to make some changes in `snapcraft_legacy` that isort doesn't). Ideally I'd like to phase out isort and use ruff's isort rules, but that's its own change.

tox.ini
lengau: All of these have the `lint` label, so you can run `tox run -m lint` to run all the linters. (`tox lint -m lint` will show you which environments get run on that label.) It doesn't look like tox can list labels or factors though, so there's no way to discover this other than looking at the labels in the tox config.

The current labels in the tox config are:

* lint
* type (which is a subset of lint)
* format (runs black, ruff and codespell with auto-fixing)
* unit-tests (runs pytest for tests/unit in Python 3.8, 3.10 and 3.11)
* integration-tests (tests/legacy like above)
* tests (union of unit-tests and integration-tests)

lengau: ...I'm gonna document that in starbase

.yamllint.yaml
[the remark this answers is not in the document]
mr-cal: I mean, should `Thruthiness` be `Truthiness`? Unless thruth is also a word! :laughing:

tox.ini
mr-cal: Interesting.  So I safely can ignore that warning?

tox.ini
mr-cal: Ah, `tox run -m lint`! Thanks!

tox.ini
lengau: It's because I forgot to check the legacy tests. Those need coverage here, but they're being included as integration tests where we don't usually run coverage. Fixed!

mr-cal (approved):
Looks good!

tox.ini
lengau: Yeah, it's more of an info thing than a warning, but it's intended from the way I did the config. There doesn't appear to be a way to tell tox not to display it.

.github/workflows/tox.yaml
mr-cal: Removing:
- `tests (ubuntu-20.04)`
- `tests (ubuntu-22.04)`

Adding:
- `tests (ubuntu-20.04, 3.8)`
- `tests (ubuntu-20.04, 3.10)`
- `tests (ubuntu-22.04, 3.8)`
- `tests (ubuntu-22.04, 3.10)`

mr-cal:
Removed:
- tests (ubuntu-20.04)
- tests (ubuntu-22.04)

Added:
- tests (ubuntu-20.04, 3.8)
- tests (ubuntu-20.04, 3.10)
- tests (ubuntu-22.04, 3.8)
- tests (ubuntu-22.04, 3.10)

tox.ini
mr-cal: Did you add `isort` here?

Makefile
```diff
@@ -3,39 +3,39 @@ SOURCES_LEGACY=snapcraft_legacy tests/legacy
...
 
 .PHONY: test-codespell
 test-codespell:
-	tox run -e codespell
+	tox run -e test-codespell
 
 .PHONY: test-isort
 test-isort:
-	tox run -e isort
+	tox run -e test-isort
```
mr-cal, line 22: shouldn't this call `tox run -e lint-isort`? (same for the other linters)

mr-cal (commented):
2 follow up questions

Makefile
lengau: Hey so it turns out `git push` is a useful command that I forgot to run :upside_down_face:

PR #4062, opened 2023-03-09
fix(clean): non host cleaning shouldn't load the project

Guillaumebeuzeboc:
The call to the `lifecycle.clean` method calls the
`project_loader.load_config` which calls the `pluginhandler.load_plugin`

This `load_plugin` method is using `sys.path` to load plugins. This is using the host `sys.path` (while cleaning a container/VM snap build).
If the host has a Python module with the same name as the snapcraft plugin, it will try to load the Python module from the host. The module from the host being something completely unrelated to snapcraft plugins, the load crashes.

A simple way to reproduce the error is to:
- source a [ROS Noetic](http://wiki.ros.org/noetic) environment (Expanding `PYTHONPATH` with a ROS dedicated path on the host)
- call `snapcraft --use-lxd` to build a ROS 1 Noetic snap
- run `snapcraft clean --use-lxd` to clean to snap build env
It will generate the following error:
```
Loaded local plugin for catkin
Failed to load plugin: unknown plugin: 'catkin'
```
It happens because it tries to load a local plugin found in the `PYTHONPATH`. The "local plugin" is simply a Python module from the host that was added to the `PYTHONPATH` when sourcing our environment.
The `PYTHONPATH` from the host shouldn't be used when cleaning a lxd environment

-----

### DIFF
--- a/snapcraft_legacy/cli/lifecycle.py
+++ b/snapcraft_legacy/cli/lifecycle.py
@@ -468,9 +468,9 @@ def clean(ctx, parts, unprime, **kwargs):
                 instance.clean_parts(part_names=parts)
         else:
             build_provider_class(project=project, echoer=echo).clean_project()
-            # Clear the prime directory on the host, unless on Windows.
+            # Clear the prime directory on the host in case snapcraft try was used, unless on Windows.
             if sys.platform != "win32":
-                lifecycle.clean(project, parts, steps.PRIME)
+                lifecycle.clean_try_dir(project)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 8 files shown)


sergiusens:
what happens with this code if you run `snapcraft try` and then `snapcraft clean`?

sergiusens:
@cmatsuoka we need to disable random load paths for modules in snapcraft

sergiusens:
Can you create an issue @Guillaumebeuzeboc ... this doesn't look like the right way to go, but indeed can be a problem

Guillaumebeuzeboc:
Issue [added to launchpad](https://bugs.launchpad.net/snapcraft/+bug/2028047)

PR #4064, opened 2023-03-09
Make sure that software-properties-common package is installed …

dbruno74:
…on buld environment to avoid the add-apt-repository call to fail.

- [ ] Have you followed the [guidelines for contributing](https://github.com/snapcore/snapcraft/blob/master/CONTRIBUTING.md)? Yes
- [ ] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)? Yes
- [ ] Have you successfully run `make lint`? No
- [ ] Have you successfully run `pytest tests/unit`? No

-----


snapcraft_legacy/plugins/v2/kernel.py
```diff
@@ -1425,6 +1425,12 @@ def _add_snappy_ppa(self) -> None:
 
             # add ppa itself
             logger.warning("adding ppa:snappy-dev/image to handle initrd builds")
+            subprocess.run(
```
mr-cal, line 1428: This one will need `sudo`

sergiusens: we assume root when running, so should be fine. That said, we should not be adding PPAs through the plugin

cmatsuoka: Can´t this be added to the plugin's build-packages list?

sergiusens:
the branch is already adding the PPA and forcing an experimental option for the plugin. If something is incorrect please coordinate with @kubiko

PR #4075, opened 2023-03-14
tests: mock dependency loading for patch elf test

mr-cal:

-----

One of the lifecycle unit tests was not properly mocked.  The test was loading dependencies for `/bin/true`, which is a process that is dependent on the host's copy of `ldd` and whether the `core22` snap is installed on the host.

### DIFF
--- a/tests/unit/parts/test_lifecycle.py
+++ b/tests/unit/parts/test_lifecycle.py
@@ -27,6 +27,7 @@ from craft_parts import Action, Step, callbacks
 from craft_providers.bases.buildd import BuilddBaseAlias
 
 from snapcraft import errors
+from snapcraft.elf import ElfFile
 from snapcraft.parts import lifecycle as parts_lifecycle
 from snapcraft.parts.update_metadata import update_project_metadata
 from snapcraft.projects import MANDATORY_ADOPTABLE_FIELDS, Architecture, Project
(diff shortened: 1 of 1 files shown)


tests/unit/parts/test_lifecycle.py
```diff
@@ -1571,9 +1575,6 @@ def test_patch_elf(snapcraft_yaml, mocker, new_dir):
             patchelf_args=[
                 "--set-interpreter",
                 "/snap/core22/current/lib64/ld-linux-x86-64.so.2",
-                "--force-rpath",
```
mr-cal, line 1574: These lines can be removed because [this if statement](https://github.com/snapcore/snapcraft/blob/13719604c44d8761df3154cd964f29d290e8a563/snapcraft/elf/_patcher.py#L67) will always be false.

cmatsuoka: Shouldn't we unit-test that codepath as well?

mr-cal: That codepath in `snapcraft/elf/_patcher.py.Patcher.patch()` is already unit tested by `tests/unit/elf/test_patcher.py`.

cmatsuoka (approved):
Looks good, thanks!

PR #4077, opened 2023-03-17
linters: allow "categories" for subfiltering

tigarmo:
This commit adds the notion of "categories" for Linter subclasses. For linters that perform multiple different "types" of linting, this lets the author ignore specific paths for specific types of linting issues, with a finer granularity than ignoring the whole linter wholesale.

Currently the only Linter type that supports this is the LibraryLinter, which provides the "unused-library" and "missing-library" categories to disable checking for unused or missing libraries.

CRAFT-1642

-----

First commit has the general linter changes, second commit has the update of the ros extension.


snapcraft/linters/base.py
```diff
@@ -103,14 +104,44 @@ def run(self) -> List[LinterIssue]:
...
+        ignored_files = self._lint.ignored_files(self._name)
+
+        if category:
+            # No "extend()" because we don't want to affect the original list.
+            ignored_files = ignored_files + self._lint.ignored_files(category)
+
+        if isinstance(filepath, ElfFile):
+            path = filepath.path
+        else:
+            path = filepath
```
cmatsuoka, line 127: I wonder if we shouldn't make `filepath` just a `Path`, and change the caller to use `elf_file.path` when necessary.

tigarmo: I thought about doing this, but I didn't want to start changing a bunch of files (noisy on the review). Do you think we should just pull the trigger and do this now?

tigarmo:
>  Can we extend the ros related spread test that uses the extension to verify this is working as expected too?

Do you mean by e.g. checking the `snapcraft` output and verifying that no linter issues are emitted? Sure

>  Additionally, can we have a spread test that declares linter entries?

Yes. I'll do both in a little bit.

tigarmo:
@sergiusens thank you for the suggestion, it pointed out an unexpected interaction. First, I added a new spread test that uses the `ros2-humble` extension and verifies that no linter warnings are generated. This worked as expected.

Then I tried to add a new spread with a "user-provided" lint ignore like:

```yaml
lint:
  ignore:
    - unused-library:
      - usr/lib/*/libpng16.so*
```

... the resulting behavior is that this new `unused-library` entry "overrides" the extension-provided one, so the snap generates all the ros-related unused library warnings. I think it would be quite straightforward to [update this behavior](https://github.com/snapcore/snapcraft/blob/main/snapcraft/projects.py#L202) to have the user-generated lint ignores "add" to the extension-provided ones, but I also think we should do that in a separate task/PR. 

What do you think?

sergiusens:
@tigarmo let's ask that question during the demo

PR #4079, opened 2023-03-21
update craft-archives and support apt pinning

tigarmo:
This commit updates craft-archives to 0.0.3, which restructures the
key-handling implementation and support repo pinning.

CRAFT-1646


setup.py
```diff
@@ -96,7 +96,7 @@ def recursive_data_files(directory, install_directory):
 install_requires = [
     "attrs",
     "click",
-    "craft-archives==0.0.1",
+    "craft-archives==0.0.3",
```
tigarmo, line 99: does anyone know why we pin the version here too?

tests/spread/core22/package-repositories/task.yaml
```diff
@@ -5,15 +5,14 @@ environment:
...
   SNAP/test_apt_ppa: test-apt-ppa
+  SNAP/test_pin: test-pin
   SNAPCRAFT_BUILD_ENVIRONMENT: ""
 
 restore: |
   cd "$SNAP"
   rm -f ./*.snap
   snapcraft clean
   snapcraft clean --destructive-mode
-  ls -l /etc/apt/sources.list.d/
```
tigarmo, line 15: removed this because tests/spread/tools/restore.sh handles this per-test already.

lengau (approved):
Thanks, this is excellent!

sergiusens (approved):
very nice!

PR #4080, opened 2023-03-23
lint: initialize retcode to satisfy pyright

tigarmo:
Latest version of pyright complains that `retcode` might not be initialized. As far as I can tell, every code path sets a value (except for `run_legacy()` which does its own exit handling), so this initial setting to 1 is spurious but should be harmless.

-----


mr-cal (approved):
Looks reasonable - I like this better than ignoring the warning.

PR #4081, opened 2023-03-24
Kernel plugin improvements 1

kubiko:
code reorganisation:
- moving more code to the common implementation
- introducing the use of `textwrap.dedent` for code readability


tests/legacy/unit/plugins/v2/test_kernel.py
```diff
@@ -0,0 +1,1878 @@
...
+        		return 0
+        	else
+        		return 1
+        	fi
+        }
+        """
+    )
+]
+
+_donwload_initrd_fnc = [
```
sergiusens, line 1170:
```suggestion
_download_initrd_fnc = [
```

cmatsuoka: Fixed in both kernel plugin tests.

cmatsuoka (approved):
Thanks @kubiko. Both plugins are now very short and much better organized. I see there's still work to be done in script formatting but this change is certainly a step forward.

PR #4084, opened 2023-03-30
providers: inject current snapcraft snap into instances

mr-cal:

-----

Allows for core22 snaps to be built using a parallel-installed instance of snapcraft.

A subsequent PR will support core18|20 snaps.

(CRAFT-1660)

### DIFF
--- a/snapcraft/providers.py
+++ b/snapcraft/providers.py
@@ -176,7 +176,19 @@ def get_base_configuration(
     # install snapcraft from the store's stable channel
     snap_channel = get_managed_environment_snap_channel()
     if sys.platform != "linux" and not snap_channel:
+        emit.progress(
+            "Using snapcraft from snap store channel 'latest/stable' in instance "
+            "because snap injection is only supported on Linux hosts.",
+            permanent=True,
+        )
+        snap_name = "snapcraft"
         snap_channel = "stable"
+    else:
+        # Use SNAP_INSTANCE_NAME for snapcraft's snap name, as it may not be 'snapcraft'
+        # if the '--name' parameter was used to install snapcraft.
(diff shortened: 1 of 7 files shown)


snapcraft/providers.py
```diff
@@ -176,7 +176,19 @@ def get_base_configuration(
     # install snapcraft from the store's stable channel
     snap_channel = get_managed_environment_snap_channel()
     if sys.platform != "linux" and not snap_channel:
+        emit.progress(
+            "Using snapcraft from snap store channel 'latest/stable' in instance "
+            " because snap injection is only supported on Linux hosts.",
```
tigarmo, line 181: minor but the full sentence might be double-spaced here

tests/spread/core22/parallel-install/task.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+  # unset parallel instance support
+  snap set system experimental.parallel-instances=null
+
+  install_lxd
+  install_snapcraft
+
+execute: |
+  #shellcheck source=tests/spread/tools/prepare.sh
+  . "$TOOLS_DIR/prepare.sh"
```
tigarmo, line 38: isn't this being called in `prepare` already?

tigarmo (changes requested):

snapcraft/providers.py
mr-cal: Good catch, thanks!

tests/spread/core22/parallel-install/task.yaml
mr-cal: Yes, but it has to get re-sourced in every section where I need to call a function from that file.

tigarmo (approved):

PR #4112, opened 2023-04-25
plugins: kernel plugin: fix core20 base builds

kubiko:
Builds for core20 are often run without `sudo` when adding ppa, we determine if sudo should be used to invoke `apt-get`
The new install step for `software-properties-common` package is missing this check, breaking existing builds without `sudo`


mr-cal:
Hey @kubiko!
https://github.com/snapcore/snapcraft/pull/4111 was just merged to main.  Do you mind rebasing or resolving the conflict?  I can approve and merge afterwards.

PR #4116, opened 2023-04-27
plugins: kernel: remove use of deprecated config.common.ports config

kubiko:
'config.common.ports' config file is deprecated and not used anymore by the Ubuntu kernel, it might disappear in the future altogether, remove use of it when using Ubuntu `kconfigflavour` option, so builds do not break in the future.


xnox (approved):
ports config was last used in like 2004-2006 timeframe for non-ubuntu architectures. thus like it should not have been ever used for any architectures targetting Ubuntu Core.

mr-cal (approved):

PR #4150, opened 2023-05-17
meta: use build-for in snap.yaml architecture

mr-cal:

-----
### Details

Previously, using `build-for: all` in a snapcraft project would cause `meta/snap.yaml` to contain the `build-on` architecture, due to [this line](https://github.com/snapcore/snapcraft/blob/2f2d1eac63de39dbd4f9bbfab8ffa5df0ac9ca62/snapcraft/parts/parts.py#L88) from my first implementation of adding support for the `architectures` keyword.

Now, `build-for: all` will cause `meta/snap.yaml` to use `architectures: all`. 

### Source
https://bugs.launchpad.net/snapcraft/+bug/1999580
https://forum.snapcraft.io/t/incorrect-architecture-with-snapcraft-7-x-and-build-for-build-on/34130
(CRAFT-1535)

### DIFF
--- a/snapcraft/parts/lifecycle.py
+++ b/snapcraft/parts/lifecycle.py
@@ -410,12 +410,7 @@ def _generate_metadata(
     )
 
     emit.progress("Generating snap metadata...")
-    snap_yaml.write(
-        project,
-        lifecycle.prime_dir,
-        arch=project.get_build_for(),
-        arch_triplet=lifecycle.target_arch_triplet,
-    )
+    snap_yaml.write(project, lifecycle.prime_dir, arch=project.get_build_for())
     emit.progress("Generated snap metadata", permanent=True)
 
     if parsed_args.enable_manifest:
--- a/snapcraft/parts/lifecycle.py
+++ b/snapcraft/parts/lifecycle.py
@@ -413,7 +413,7 @@ def _generate_metadata(
     snap_yaml.write(
         project,
         lifecycle.prime_dir,
-        arch=lifecycle.target_arch,
+        arch=project.get_build_for(),
         arch_triplet=lifecycle.target_arch_triplet,
     )
     emit.progress("Generated snap metadata", permanent=True)
@@ -449,7 +449,7 @@ def _generate_manifest(
     manifest.write(
         project,
         lifecycle.prime_dir,
-        arch=lifecycle.target_arch,
+        arch=project.get_build_for(),
         parts=parts,
         start_time=start_time,
         image_information=image_information,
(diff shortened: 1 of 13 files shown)


snapcraft/parts/lifecycle.py
```diff
@@ -413,7 +413,7 @@ def _generate_metadata(
     snap_yaml.write(
         project,
         lifecycle.prime_dir,
-        arch=lifecycle.target_arch,
+        arch=project.get_build_for(),
         arch_triplet=lifecycle.target_arch_triplet,
```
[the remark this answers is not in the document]
lengau: I'd like to understand this a bit more.

If we're building for multiple architectures (not arch-independent), I presume this gets generated on a per-arch basis? (If so that's fine.)

If we're building an architecture-independent snap, does an `arch` value get written to the `snap.yaml` file here? If so, this *could* be a problem if something ends up using that value for any reason.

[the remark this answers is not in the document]
tigarmo: ping @sergiusens

tigarmo (approved):

tests/unit/parts/test_lifecycle.py
```diff
@@ -1686,3 +1686,58 @@ def test_patch_elf(snapcraft_yaml, mocker, new_dir):
...
+        "prime",
+        project=project,
+        parse_info={},
+        assets_dir=Path(),
+        start_time=datetime.now(),
+        parallel_build_count=8,
+        parsed_args=parsed_args,
+    )
+
+    assert mock_write_metadata.mock_calls == [
```
lengau, line 1725: Probably out of scope here, but I'd love to see some integration tests for creating these files.

(Out of scope because it looks like we don't currently have an integration test infrastructure at all.)

lengau (approved):

snapcraft/parts/lifecycle.py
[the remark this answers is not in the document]
mr-cal: As discussed offline, I've fixed this behavior.

tests/unit/parts/test_lifecycle.py
mr-cal: Agreed

tests/unit/meta/test_snap_yaml.py
```diff
@@ -925,3 +900,37 @@ def test_app_passthrough_snap_yaml(simple_project, new_dir):
           PATH: $SNAP/usr/sbin:$SNAP/usr/bin:$SNAP/sbin:$SNAP/bin:$PATH
         """
     )
+
+
+def test_architectures(simple_project, new_dir):
```
lengau, line 905: Please test this with another arch triplet

mr-cal: Done!

sergiusens (approved):

PR #4209, opened 2023-06-12
extensions: add no_extension entry for parts

artivis:
Add a new parts entry, `no_extension`, that prevent extensions from being applied to a part that sets this new entry to true. 

Extensions' `part_snippet` are applied to all parts systematically. However in a multi-part snap, this behavior can cause issues, e.g. if the extension defines entries that are specific to a given plugin resulting in an `Additional properties are not allowed` error. This PR introduces a new entry that allows for explicitely shielding a part from having *any* extension applied to it, e.g. see [here](https://github.com/snapcore/snapcraft/compare/main...artivis:feature/no_extension?expand=1#diff-cf7cd4f4450e2eb36d0a44171d9ea8dae69312340c8b8901f2591a41b9f11e07R522).

-----

### DIFF
--- a/snapcraft/extensions/_utils.py
+++ b/snapcraft/extensions/_utils.py
@@ -21,6 +21,8 @@ import contextlib
 import copy
 from typing import TYPE_CHECKING, Any, Dict, List, Set
 
+from snapcraft import errors
+
 from .registry import get_extension_class
 
 if TYPE_CHECKING:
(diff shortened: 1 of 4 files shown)


sergiusens:
we can rather make extensions per part plugin aware for these things

artivis:
> we can rather make extensions per part plugin aware for these things

I don't think the two features are mutually exclusive. Actually, you may want an extension to apply to a plugin type some times and not others (e.g. destkop to dump).

sergiusens:
closing in favor of #4271

PR #4217, opened 2023-06-16
parts: show the current part lifecycle in the window title

sergio-costas:
When building a snap with several parts using -v, it's usually not possible to know which part is being built due to the long text output.

This patch shows in the window title the part lifecycle that is being executed.

Although maybe a better way would be to add a method "set_title" in craft_cli... but that would require two PRs.

-----

### DIFF
--- a/snapcraft/parts/parts.py
+++ b/snapcraft/parts/parts.py
@@ -80,7 +80,7 @@ class PartsLifecycle:
         self._parse_info = parse_info
         self._all_part_names = [*all_parts]
 
-        emit.progress("Initializing parts lifecycle")
+        emit.progress("Initializing parts lifecycle", update_titlebar=True)
 
         # set the cache dir for parts package management
         cache_dir = BaseDirectory.save_cache_path("snapcraft")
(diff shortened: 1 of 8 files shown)


snapcraft/parts/parts.py
```diff
@@ -173,6 +173,7 @@ def run(
                 for action in actions:
                     message = _action_message(action)
                     emit.progress(f"Executing parts lifecycle: {message}")
+                    print(f"\033]2;Executing parts lifecycle: {message}\007")
```
sergiusens, line 176: nice addition, but your original assessment seems correct, craft-cli is about the presentation. You can work with @tigarmo on figuring out what the best param for emit.progress can be

sergio-costas:
I'm testing it. Just did this replacement in `requirements.txt`:

    -craft-cli==2.0.0
    +git+https://github.com/sergio-costas/craft-cli@UDENG-832-allow-to-set-the-window-title#craft-cli

sergio-costas:
Testing it with `gnome-42-2204-sdk`... seems to work fine.

sergio-costas:
The Tox errors are because it can't find the new parameter. When the patch for craft_cli is merged, they should disappear.

sergio-costas:
Completed the gnome-42-2204-sdk build. It works fine.

tigarmo:
> git+https://github.com/sergio-costas/craft-cli@UDENG-832-allow-to-set-the-window-title#craft-cli

can you do this in the PR too please? So we know the tests and etc are OK with the changes

sergio-costas:
> can you do this in the PR too please? So we know the tests and etc are OK with the changes

Done.

sergiusens:
I am closing this as the craft-cli work is not yet complete, please reopen once there

sergio-costas:
What do you mean that is not complete? It is... The problem in KDE is a kterm problem...

PR #4270, opened 2023-07-11
remote-build: Use personal access tokens

cjwatson:
The initial mechanism used by Launchpad to grant temporary access tokens to `snapcraft remote-build` was a bit of a hack.  They were essentially pure bearer tokens (implemented as macaroons) that weren't recorded anywhere on the Launchpad side, which meant that there was no mechanism for revoking them.  We mitigated this with a hardcoded expiry time, but it wasn't great.

We now have a new "personal access tokens" mechanism, which is much better.  These tokens show up in a "Manage access tokens" page on the repository and can be revoked by the owner; they have more specific control of the access scopes they grant, and the caller can specify an expiry time.  The old mechanism is now deprecated, and we intend to remove it after a suitable transition period.

This commit changes `snapcraft remote-build` to use the new mechanism. The unit tests can essentially only confirm that it's passing the expected arguments to the Launchpad API, so I've also run an end-to-end manual test against Launchpad production to make sure that those arguments have the intended effect.

We can use a very tight expiry time here, since it only has to last for long enough for the `git push` process to _start_.  It's OK to increase it a little if needed, but it's generally good for token lifetimes to be short where possible.

-----

### DIFF
--- a/snapcraft_legacy/internal/remote_build/_launchpad.py
+++ b/snapcraft_legacy/internal/remote_build/_launchpad.py
@@ -19,7 +19,7 @@ import logging
 import os
 import shutil
 import time
-from datetime import datetime
+from datetime import datetime, timedelta, timezone
 from typing import Any, Dict, List, Optional, Sequence, Type
 from urllib.parse import unquote, urlsplit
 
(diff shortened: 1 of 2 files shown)


sergiusens (approved):
thanks Colin!

PR #4290, opened 2023-07-23
docker: add core22 base

abitrolly:

-----

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -36,7 +36,7 @@ FROM ubuntu:$UBUNTU
 COPY --from=builder /snap/core /snap/core
 COPY --from=builder /snap/core18 /snap/core18
 COPY --from=builder /snap/core20 /snap/core20
-COPY --from=builder /snap/core20 /snap/core22
+COPY --from=builder /snap/core22 /snap/core22
 COPY --from=builder /snap/snapcraft /snap/snapcraft
 COPY --from=builder /snap/bin/snapcraft /snap/bin/snapcraft
 
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,5 +1,5 @@
-ARG RISK=edge
-ARG UBUNTU=xenial
+ARG RISK=stable
+ARG UBUNTU=22.04
 
 FROM ubuntu:$UBUNTU as builder
 ARG RISK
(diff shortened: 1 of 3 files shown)


docker/prepare.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+set -xeo pipefail
+
+
+echo "--- unpacking bases for snapcraft/$RISK ---"
+
+# Grab the core snaps (which snapcraft uses as a base) from the stable channel
+# and unpack them in the proper place.
+BASES="core core18 core20 core22"
```
sergiusens, line 10: it doesn't make much sense to build a snap with a core18 base if the base of the OCI image is 22.04

abitrolly: There are many snaps that need `core18`. Why it is wrong to build `core18` with 22.04?

docker/Dockerfile
```diff
@@ -58,11 +36,16 @@ FROM ubuntu:$UBUNTU
 COPY --from=builder /snap/core /snap/core
 COPY --from=builder /snap/core18 /snap/core18
 COPY --from=builder /snap/core20 /snap/core20
+COPY --from=builder /snap/core20 /snap/core22
```
VladRassokhin, line 39: Should be `core22` in both args

docker/Dockerfile
```diff
@@ -58,11 +36,16 @@ FROM ubuntu:$UBUNTU
...
 COPY --from=builder /snap/snapcraft /snap/snapcraft
 COPY --from=builder /snap/bin/snapcraft /snap/bin/snapcraft
 
 # Generate locale and install dependencies.
-RUN apt-get update && apt-get dist-upgrade --yes && apt-get install --yes snapd sudo locales && locale-gen en_US.UTF-8
+RUN apt-get update && apt-get -y dist-upgrade && apt-get -y install \
```
VladRassokhin, line 44: Minor: not sure whether this change is really necessary as it just changes the order of arguments and adds lines

docker/Dockerfile
```diff
@@ -1,42 +1,20 @@
-ARG RISK=edge
-ARG UBUNTU=xenial
+ARG RISK=stable
+ARG UBUNTU=22.04
```
VladRassokhin, line 2: In README.md codename of release is used, probably makes sense to stick to codenames and replace with `jammy` ?

abitrolly: I rarely remember which release name is which version. I remember that 22.04 is the latest LTS, and I guess that's just what matters here.

docker/Dockerfile
abitrolly: Missed that. Thanks. Fixed.

docker/Dockerfile
abitrolly: It is Docker best practices https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#sort-multi-line-arguments

VladRassokhin: Yeah, following best practices is a good thing, but it's not mentioned in the commit message and looks unrelated (could be in another commit and PR)

VladRassokhin:
Thank you for this PR, I've tested it and improved a bit for my needs: https://github.com/VladRassokhin/snapcraft/tree/dockerfile

Spotted problem:

`core22` uses `lxd` by default which won't work inside docker container, so you need to add `ENV SNAPCRAFT_BUILD_ENVIRONMENT=host` (see [commit](https://github.com/VladRassokhin/snapcraft/commit/fe5b53e5077d03ea06d03b7c145946be417af4d8))

sergiusens:
we do not want to change this for legacy reasons, we have however started https://github.com/canonical/snapcraft-rocks as our new initiative

VladRassokhin:
@sergiusens how is that related to rocks exactly? Why it replaces this image which is used for communication with snapcraft servers on a non-ubuntu machnes?

PR #4302, opened 2023-08-02
plugins: kernel: split kernel plugin into kernel and initrd plugins

kubiko:
kernel plugin builds two components, kernel and initrd
While initrd is dependent on the kernel, it is a separate component.
Separating those 2 steps into 2 own plugins will allow greater reuse of the
kernel plugin as well as each plugin being owned by respective teams.

Initrd steps will always depend on the kernel step, but with
the plugin separation, prebuilt kernel binaries can be used with the initrd plugin
In the future inird might even become the official way to build kernel snap
from kernel deb packages.
For simplicity now ignoring signed kernels...

-----


sergiusens (commented):
Ondra, this is fantastic, thanks for this.

I'll circle back to the kernel, foundations and core folks to get this reviewed as well

mr-cal:
@kubiko - I see you pushed some updates last week.  Let me know when this is ready for review

dilyn-corner:
What is the difference between `initrd-modules` and `initrd-configured-modules`? It seems like they basically collapse into each other; why don't I want a module in the initrd NOT to autoload?

`bison`, `flex`, `libelf-dev`, and `libssl-dev` are essentially required to build any kernel (or one that matters), so they should probably be added to the list of packages like `gcc`, `make`, `kmod`, etc. are.

dilyn-corner:
re: 3 in my last comment, this issue is relevant: https://github.com/canonical/craft-archives/issues/104

I would argue that this PR shouldn't be merged until that problem has been figured out.

mr-cal:
> Let me know when this is ready for review

It's been awhile since I left that comment above, but it is still true - just let me know when this is ready.

KDegeling:
After reading the documentation on building a kernel snap: https://documentation.ubuntu.com/core/how-to-guides/image-creation/build-a-kernel-snap/ and looking into the initrd plugin, I thought that cross-building was supported.

However, when I try to build a kernel snap using the `aarch64-linux-gnu` cross compiler, I run into issues when trying to build the initrd step. The initrd plugin mentions that for cross-building the build-environment needs to be set-up accordingly, however the issue that I run into is the following:
```
:: + chroot /root/parts/initrd/src/uc-initramfs-build-root /bin/bash -c 'apt-get update'
:: chroot: failed to run command ‘/bin/bash’: Exec format error
:: + _clean_chroot /root/parts/initrd/src/uc-initramfs-build-root
```
Which doesn't seem to be related to any of the cross-compilation tools. Am I missing some required steps? Or is cross-building not yet supported?

When building on an arm64 machine I am able to produce a valid kernel snap.

dilyn-corner:
> ```
> :: + chroot /root/parts/initrd/src/uc-initramfs-build-root /bin/bash -c 'apt-get update'
> :: chroot: failed to run command ‘/bin/bash’: Exec format error
> :: + _clean_chroot /root/parts/initrd/src/uc-initramfs-build-root
> ```


Are you building with `--use-lxd`? If snapcraft is creating a LXD container, you need to have things like the qemu-user-static package installed on your host in order for LXD to setup binfmt_misc in the environment.

dilyn-corner:
Some extra digging was done into LXD behavior here to @KDegeling's point and some work is still outstanding to make this work optimally: https://github.com/canonical/lxd/pull/16658

This should make its way into a release soon enough and get into 5.21.

For now, adding the following entry to `/var/snap/lxd/common/lxc/local.conf` should solve the issue: 

```
lxc.mount.entry = /proc/sys/fs/binfmt_misc proc/sys/fs/binfmt_misc none bind,rw,create=dir 0 0
```

kubiko:
Closing PR as superseded with https://github.com/canonical/snapcraft/pull/5814

PR #4313, opened 2023-08-11
environment: add build-on and build-for envvars for core20

mr-cal:

-----

New environment variables:
- `SNAPCRAFT_ARCH_BUILD_ON`
- `SNAPCRAFT_ARCH_TRIPLET_BUILD_ON`
- `SNAPCRAFT_ARCH_BUILD_FOR`
- `SNAPCRAFT_ARCH_TRIPLET_BUILD_FOR`

The build-for envvars are only available when the build-for architecture can be determined. They are not available for multi-architecture builds or for unknown build-for architectures.

TODO: 
- [x] add spread tests
- [x] s/run-on/build-for
- [x] change target from `hotfix/7.5` to `main`

[LP: #1997752](https://bugs.launchpad.net/bugs/1997752)
(CRAFT-1498)


sergiusens:
`s/RUN_ON/BUILD_FOR/`
add CRAFT variants

snapcraft_legacy/internal/pluginhandler/_part_environment.py
```diff
@@ -84,6 +86,12 @@ def get_snapcraft_global_environment(
         "SNAPCRAFT_TARGET_ARCH": project.target_arch,
         "SNAPCRAFT_CONTENT_DIRS": content_dirs_envvar,
     }
+    # run-on architecture may not be available
```
sergiusens, line 89: if the run on architecture is not available, then we source the build_on as the convention is that they should match, and perhaps this is the culprit to all problems

sergiusens: calling it build_for makes the intent a bit more obvious too

mr-cal: I changed `run-on` to `build-for` and reworded this comment

tests/legacy/unit/pluginhandler/test_environment.py
```diff
@@ -0,0 +1,86 @@
...
+        base="core20",
+        parts={"test-part": {"plugin": "nil"}},
+        architectures=[{"build-on": ["amd64", "arm64"], "run-on": "amd64"}],
+    )
+    snapcraft_yaml.write_snapcraft_yaml()
+    project = Project(snapcraft_yaml_file_path=snapcraft_yaml.snapcraft_yaml_file_path)
+
+    environment = get_snapcraft_global_environment(project)
+
+    assert environment["SNAPCRAFT_ARCH_BUILD_ON"] == project.arch_build_on
```
sergiusens, line 37: This might be more obvious (ditto to other places,
```suggestion
    assert environment["SNAPCRAFT_ARCH_BUILD_ON"] == "amd64"
```
And so on, if you are thinking about the case of failure when running these tests on arm64, well, they will also fail on s390x. The platform.machine call needs to be mocked for that :-) Not a rush here

sergiusens (approved):

sergiusens:
This looks good, some minor improvements would be nice

PR #4329, opened 2023-08-18
cli: prioritize `--use-lxd` over `SNAPCRAFT_BUILD_ENVIRONMENT`

mr-cal:
- [x] Have you followed the [guidelines for contributing](https://github.com/snapcore/snapcraft/blob/master/CONTRIBUTING.md)?
- [x] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [x] Have you successfully run `make lint`?
- [x] Have you successfully run `pytest tests/unit`?

-----

Prioritize the command line argument `--use-lxd` over the environment variable `SNAPCRAFT_BUILD_ENVIRONMENT=host`.

Resolves #4307 
(CRAFT-1956)

PR #4330, opened 2023-08-21
manifest: use `--enable-image-information` inside instances

mr-cal:

-----

This fixes a bug that has been present since manifests were [added to core22](https://github.com/snapcore/snapcraft/pull/3824/files#diff-81bf1b8cfa88289a0432d3e12f4b4c1c104d4463a49dbc0e6c431d4e444a35cdR372-R378).  

The bug is that `--enable-image-information` was not passed into LXD/Multipass instances.  This flag only got used when running in destructive mode.  During the implementation, there was probably a typo/mix-up between the envvar `SNAPCRAFT_BUILD_INFO` and `--enable-image-information`.

This bug was uncovered by #4329, which prevented the spread test from running inside a LXD container.

### DIFF
--- a/snapcraft/parts/lifecycle.py
+++ b/snapcraft/parts/lifecycle.py
@@ -506,9 +506,9 @@ def _run_in_provider(
 
     if getattr(parsed_args, "enable_manifest", False):
         cmd.append("--enable-manifest")
-    build_information = getattr(parsed_args, "manifest_build_information", None)
-    if build_information:
-        cmd.extend(["--manifest-build-information", build_information])
+    image_information = getattr(parsed_args, "manifest_image_information", None)
+    if image_information:
+        cmd.extend(["--manifest-image-information", image_information])
 
     cmd.append("--build-for")
     cmd.append(project.get_build_for())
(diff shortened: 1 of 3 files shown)


snapcraft/parts/lifecycle.py
```diff
@@ -506,9 +506,9 @@ def _run_in_provider(
 
     if getattr(parsed_args, "enable_manifest", False):
         cmd.append("--enable-manifest")
-    build_information = getattr(parsed_args, "manifest_build_information", None)
-    if build_information:
-        cmd.extend(["--manifest-build-information", build_information])
```
tigarmo, line 511: so is `--manifest-build-information` a completely invalid/inexisting flag?

tigarmo (approved):

snapcraft/parts/lifecycle.py
cmatsuoka: Good catch, this was likely a mix-up with `SNAPCRAFT_BUILD_INFO`. Can we have a spread test using the option in non-destructive-mode?

cmatsuoka (approved):

snapcraft/parts/lifecycle.py
mr-cal: You already wrote a spread test it with LXD and destructively!

PR #4332, opened 2023-08-23
environment: error when build-for arch variable cannot be evaluated

### LINKED ISSUE #4314: Raise error when using `BUILD_FOR` envvars on multi-arch and unknown builds
### What needs to get done

https://github.com/snapcore/snapcraft/pull/4313 added the following environment variables for core20 snaps:

- SNAPCRAFT_ARCH_BUILD_ON
- SNAPCRAFT_ARCH_TRIPLET_BUILD_ON
- SNAPCRAFT_ARCH_BUILD_FOR
- SNAPCRAFT_ARCH_TRIPLET_BUILD_FOR

Multi-arch builds and builds with unknown architectures do not evaluate `SNAPCRAFT_ARCH_BUILD_FOR` or `SNAPCRAFT_ARCH_TRIPLET_BUILD_FOR`.

When these envvars are used in a snapcraft.yaml, an error should be raised, explaining to the user why they cannot use these envvars.

### Why it needs to get done

This will improve the user experience for core20 snaps, where users are allowed to build for multiple architectures and unknown architectures.

mr-cal:

-----

For core20 snaps, if the project variable `SNAPCRAFT_ARCH_BUILD_FOR` or `SNAPCRAFT_ARCH_TRIPLET_BUILD_FOR` cannot be determined and is used in a snapcraft.yaml file, then an error is raised.

cc @cmatsuoka - this may be a good reference for https://github.com/canonical/craft-parts/issues/514

Resolves #4314 
(CRAFT-1963)


snapcraft_legacy/internal/project_loader/__init__.py
```diff
@@ -53,3 +62,38 @@ def replace_attr(
...
+    project_variables = [
+        "SNAPCRAFT_ARCH_BUILD_FOR",
+        "SNAPCRAFT_ARCH_TRIPLET_BUILD_FOR",
+    ]
+
+    # expand to shell syntax for variables (`$item` and `${item}`)
+    expanded_project_variables = (
+        [f"${item}" for item in project_variables ] +
+        [f"${{{item}}}" for item in project_variables]
+    )
```
lengau, line 89:
```suggestion
    expanded_project_variables = [
        *(f"${item}" for item in project_variables),
        *(f"${{{item}}}" for item in project_variables),
    ]
```

This is more obvious what it's doing IMO, but no strong feelings

snapcraft_legacy/internal/project_loader/__init__.py
```diff
@@ -53,3 +62,38 @@ def replace_attr(
...
+
+    if (
+        replacement in attr
+        and replacement in expanded_project_variables
+        and value is None
```
lengau, line 94: Could we put `if value is not None: return` at the top of the function and skip the variable expansion logic each time?

.github/workflows/publish.yaml
```diff
@@ -26,7 +26,8 @@ jobs:
           fetch-depth: 0
 
       - if: steps.decisions.outputs.PUBLISH == 'true'
-        uses: snapcore/action-build@v1
+        # XXX: waiting on a release with https://github.com/snapcore/action-build/pull/59
+        uses: snapcore/action-build@master
```
sergiusens, line 30: this should be released now

tests/spread/cross-compile/environment/task.yaml
```diff
@@ -1,25 +1,41 @@
...
 restore: |
   cd "./snaps/$SNAP"
-  snapcraft clean
+  snapcraft clean --destructive-mode
   rm -f ./*.snap
 
 execute: |
   cd "./snaps/$SNAP"
 
+  # test error scenarios
```
lengau, line 23: This seems like it would do better as a separate test (`environment-errors`?)

tests/spread/cross-compile/environment/snaps/build-for-multi-arch/snap/snapcraft.yaml
```diff
@@ -12,12 +12,15 @@ architectures:
...
-  SNAPCRAFT_ARCH_TRIPLET_BUILD_FOR: ${SNAPCRAFT_ARCH_TRIPLET_BUILD_FOR}
   SNAPCRAFT_ARCH_BUILD_ON: ${SNAPCRAFT_ARCH_BUILD_ON}
   SNAPCRAFT_ARCH_TRIPLET_BUILD_ON: ${SNAPCRAFT_ARCH_TRIPLET_BUILD_ON}
   SNAPCRAFT_TARGET_ARCH: ${SNAPCRAFT_TARGET_ARCH}
 
 parts:
   my-part:
     plugin: nil
+    override-pull:
+      echo ${SNAPCRAFT_ARCH_TRIPLET}
```
lengau, line 23: Maybe put a test command here to ensure it's empty as we expect?

tests/spread/cross-compile/environment/snaps/target-arch/snap/snapcraft.yaml
```diff
@@ -21,3 +21,10 @@ environment:
 parts:
   my-part:
     plugin: nil
+    override-pull:
+      echo ${SNAPCRAFT_ARCH_TRIPLET}
```
lengau, line 25: Might be worth checking that the values here are what we expect too

lengau (commented):
What do these `None` values look like in the resulting environment variables? Are they unset? Empty string?

.github/workflows/publish.yaml
lengau: Thanks! :rocket: 

The `v1` tag is still pointing to `v1.1.2` though

snapcraft_legacy/internal/project_loader/__init__.py
mr-cal: Implemented, thanks!

snapcraft_legacy/internal/project_loader/__init__.py
mr-cal: Good idea.  Done.

tests/spread/cross-compile/environment/task.yaml
mr-cal: Done.  I separated the passing and failing scenarios into 2 separate suites

tests/spread/cross-compile/environment/snaps/target-arch/snap/snapcraft.yaml
mr-cal: Done.  I updated the spread test to confirm these variables are as-expected.

tests/spread/cross-compile/environment/snaps/build-for-multi-arch/snap/snapcraft.yaml
mr-cal: `SNAPCRAFT_ARCH_TRIPLET` is a confusing variable that will always exist and not be empty.  Sometimes it's `build-for` and sometimes `build-on`, determined by convoluted business logic.

I didn't change that variable but I did add it to these tests to help catch regressions in the future.

mr-cal:
> What do these None values look like in the resulting environment variables? Are they unset? Empty string?

None values get cast to the string literal `"None"`.  With this PR, the `BUILD_FOR` should no longer evaluate to `"None"` and instead raise an error.

I'm not sure if any of the other snapcraft project variables will have a `None` value.  I'm pretty sure they won't, but I decided not to be conservative and not validate other project variables.

snapcraft_legacy/internal/project_loader/__init__.py
cmatsuoka: Why a list and not a tuple in the refactored version?

snapcraft_legacy/internal/project_loader/__init__.py
```diff
@@ -53,3 +62,38 @@ def replace_attr(
         return result
 
     return attr
+
+
+def _validate_replacement(attr: str, replacement: str, value: Optional[str]) -> None:
```
cmatsuoka, line 67: The "replacement" variable name sounds a bit strange to me as it's the thing that will be replaced and not the new value. Maybe it could be better expressed as "variable", "name" or something similar?

cmatsuoka (approved):
It looks good, just consider changing the replacement variable name because it suggests it's the new value that will be used to replace something.

snapcraft_legacy/internal/project_loader/__init__.py
mr-cal: Good point, a tuple is more appropriate.

snapcraft_legacy/internal/project_loader/__init__.py
mr-cal: Thanks, I improved the naming!

lengau (approved):
Thanks for answering! this is great

PR #4345, opened 2023-09-01
chore: remove unused list_registered legacy code

### LINKED ISSUE #4296: Make 'price' field optional
### What needs to get done

From what I can see, the snapcraft legacy client, in `snapcraft_legacy/_store.py`, requires that the 'price' field be present in the response from the dashboard's `/dev/api/account` endpoint:
```
def list_registered():
    account_info = StoreClientCLI().get_account_information()
    snaps = [
        (
            name,
            info["since"],
            "private" if info["private"] else "public",
            info["price"] or "-",  # would lead to a KeyError if 'price' is removed
            "-",
        )
        for name, info in account_info["snaps"].get(DEFAULT_SERIES, {}).items()
        # Presenting only approved snap registrations, which means name
        # disputes will be displayed/sorted some other way.
        if info["status"] == "Approved"
    ]
```

Changing the line from `info["price"] or "-"` to `info.get("price") or "-"` should take care of this. Adding a test for a missing 'price' field response from the dashboard would also be great.

### Why it needs to get done

The aforementioned API currently always returns 'price' as `null` since the feature is deprecated. This change will enable a removal path for the 'price' field altogether from the Dashboard API endpoint `/dev/api/account` in the future.

sergiusens:
This code has since been ported to the newer instances as a Craft CLI command "names", which uses a newer implementation that does not have any reference to price.

With this we remove the last references to Price, even if unused.

-----
Fixes #4296


mr-cal (approved):
:+1: linter failure looks minor

cmatsuoka (approved):

PR #4385, opened 2023-09-28
feat(remote-build): get build id

### LINKED ISSUE #4322: remote-build: migrate build-id generation
### What needs to get done

Migrate code to generate the build-id:
- get the project name from the snapcraft.yaml
- generate the hash for the build-id in [`snapcraft_legacy/project/_project.py`](https://github.com/snapcore/snapcraft/blob/8758571f08fcdb896a25dce98d8bd6ee65a7f5f2/snapcraft_legacy/project/_project.py#L86).

This needs to be migrated to `snapcraft/commands/remote.py`.

### Why it needs to get done

The build-id format is `snapcraft-<snap-name>-<project-hash>`.  The project hash is based on the entire contents of the project directory.  This allows for parallel builds of the same project (see details [here](https://github.com/snapcore/snapcraft/pull/3554)).

mr-cal:

-----

4 commits.  It's easiest to review one commit at a time.

Generates a build-id or uses the build-id provided at the command line.

Fixes #4322 
(CRAFT-1977)

### DIFF
--- a/snapcraft/commands/remote.py
+++ b/snapcraft/commands/remote.py
@@ -20,8 +20,10 @@ import argparse
 import os
 import textwrap
 from enum import Enum
+from functools import partial
+from hashlib import md5
 from pathlib import Path
-from typing import Optional
+from typing import List, Optional
 
 from craft_cli import BaseCommand, emit
 from craft_cli.helptexts import HIDDEN
@@ -133,6 +135,7 @@ class RemoteBuildCommand(BaseCommand):
 
         # pylint: disable=attribute-defined-outside-init
         self._snapcraft_yaml = yaml_utils.get_snap_project().project_file
+        self._parsed_args = parsed_args
         base = self._get_effective_base()
         self._run_new_or_fallback_remote_build(base)
 
@@ -186,12 +189,15 @@ class RemoteBuildCommand(BaseCommand):
 
     def _run_new_remote_build(self) -> None:
         """Run new remote-build code."""
+        # the build-id will be passed to the new remote-build code as part of #4323
+        # pylint: disable-next=unused-variable
+        self._get_build_id()
+
         # TODO: use new remote-build code (#4323)
         emit.debug(
             "Running fallback remote-build because new remote-build is not available."
         )
         run_legacy()
-
 
     def _get_build_strategy(self) -> Optional[_Strategies]:
         """Get the build strategy from the envvar `SNAPCRAFT_REMOTE_BUILD_STRATEGY`.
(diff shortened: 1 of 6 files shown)


snapcraft/commands/remote.py
```diff
@@ -131,6 +131,8 @@ def run(self, parsed_args: argparse.Namespace) -> None:
         ):
             raise AcceptPublicUploadError()
 
+        # pylint: disable=attribute-defined-outside-init
```
tigarmo, line 136: `disable-next` maybe?

snapcraft/commands/remote.py
```diff
@@ -243,3 +249,63 @@ def _get_effective_base(self) -> str:
...
+        will be different.
+
+        :returns: A string containing the md5 hash.
+        """
+        # sort the list of files for reproducibility
+        files = sorted([file for file in Path().glob("**/*") if file.is_file()])
+        hashes: List[str] = []
+
+        for file_path in files:
+            md5_hash = md5()  # noqa: S324 (insecure-hash-function)
```
tigarmo, line 303: I realize that this is copied-over code, but do you happen to know why the hash is built as a hash-of-hashes like this? Instead of a single hash that is updated with the contents of all files, I mean

tests/unit/commands/test_remote.py
```diff
@@ -435,3 +439,121 @@ def test_run_in_repo_newer_than_core22(emitter, mock_run_legacy, monkeypatch, ne
...
+# Build id tests #
+##################
+
+# The build-id is not currently used, so these unit tests test the log output.
+# When #4323 is complete, these tests can be rewritten to verify the build-id passed
+# to the new remote-build code.
+
+
+@pytest.mark.parametrize(
+    "create_snapcraft_yaml", CURRENT_BASES - {"core22"}, indirect=True
```
tigarmo, line 454: remind me please: why doesn't this use core22?

snapcraft/commands/remote.py
```diff
@@ -243,3 +249,63 @@ def _get_effective_base(self) -> str:
             )
 
         return base
+
+    def _get_build_id(self) -> str:
```
tigarmo, line 253: A question about the broader scope of the feature (remote build): eventually we want to move the related code somewhere other than snapcraft (possibly a new library, or craft-application) so that other tools can use it too, right? Will this build-id-related code also be moved? If so, do you think it would be worth it to refactor this a bit to have isolated functions (e.g. `get_build_id(app_name, project_name, project_path)`) that are easier to move later?

tigarmo (commented):
I have a general question

snapcraft/commands/remote.py
mr-cal: Like this?

```python
md5_hash = md5()  # noqa: S324 (insecure-hash-function)
for file_path in files:
  with open(file_path, "rb") as file:
    # read files in chunks in case they are large
    for block in iter(partial(file.read, 4096), b""):
      md5_hash.update(block)

return md5_hash.hexdigest()  # noqa: S324 (insecure-hash-function)
```

I don't see any reason why it wasn't done like this in the first place.

snapcraft/commands/remote.py
mr-cal: Which do you prefer?
```python
# pylint: disable-next=attribute-defined-outside-init
self._snapcraft_yaml = yaml_utils.get_snap_project().project_file
# pylint: disable-next=attribute-defined-outside-init
self._parsed_args = parsed_args
```

```python
# pylint: disable=attribute-defined-outside-init
self._snapcraft_yaml = yaml_utils.get_snap_project().project_file
self._parsed_args = parsed_args
# pylint: enable=attribute-defined-outside-init
```

snapcraft/commands/remote.py
mr-cal: I like that idea.  I can move this to `snapcraft/remote/utils.py` so it will be ready to migrate.

tests/unit/commands/test_remote.py
mr-cal: No reason, I need to test this on all bases.

snapcraft/commands/remote.py
cmatsuoka: I believe the reason to use concatenated hashes is to make it more collision-resistant (if you just hash all files together you could change file boundaries without affecting the final digest).

snapcraft/commands/remote.py
cmatsuoka: Method 2 is probably the most readable? Not a strong opinion though.

snapcraft/commands/remote.py
```diff
@@ -211,33 +186,96 @@ def _run_new_or_fallback_remote_build(self, base: str) -> None:
...
+                project_name=self._get_project_name(),
+                project_path=Path(),
+            )
+            emit.debug(f"Using computed build ID {build_id!r}.")
 
-        if not parsed_args.launchpad_accept_public_upload and not confirm_with_user(
-            _CONFIRMATION_PROMPT
-        ):
-            raise AcceptPublicUploadError()
+        # TODO: use new remote-build code (#4323)
```
cmatsuoka, line 226: If `build_id` is used to reuse an existing git repo, does it make sense to include files under `.git` when calculating the id? We could save some time and cpu skipping it.

cmatsuoka:
Added some questions and a comment about concatenated hash, which seems to be more robust than just hashing files together.

snapcraft/commands/remote.py
tigarmo: yeap either-or is fine with me

snapcraft/remote/utils.py
```diff
@@ -0,0 +1,70 @@
...
+
+def get_build_id(app_name: str, project_name: str, project_path: Path) -> str:
+    """Get the build id for a project.
+
+    The build id is formatted as `snapcraft-<project-name>-<hash>`.
+    The hash is a hash of all files in the project directory.
+
+    :returns: The build id.
+
+    :raises SnapcraftError: If the snapcraft.yaml does not contain a 'name' keyword.
```
tigarmo, line 32: I think this is leftover

snapcraft/commands/remote.py
mr-cal: I left it as-is and added a note to the docstring about the limitations.

snapcraft/remote/utils.py
mr-cal: Removed.

snapcraft/commands/remote.py
mr-cal: Yes, we could ignore `.git` to improve performance. I am going to leave it as-is for now.

tigarmo (approved):
nice thanks!

cmatsuoka (approved):
Nice! Thank you!

Issue #4324, opened 2023-08-14
remote-build: new command logic


mr-cal:
### What needs to get done

- [ ] Implement the new work area, where the WorkTree is not reused and removed after the build completes.
    - Each launchpad build has a unique build-id and WorkTree.  In the legacy code, WorkTrees are stored by build-id in `~/.local/share/snapcraft/remote-build/<build-id>/`.
    - In the new code, WorkTrees will be stored in the work area  `~/.cache/snapcraft/remote-build/<build-id>/`.  When the build completes successfully, the WorkTree will be removed.  WorkTrees are not reused.
- [ ] use new `_generate_build_id()` from #4322 
- [ ] linting and typing updates (e.g. use f-strings)
- [ ] move snapcraft business logic to `snapcraft/commands/remote.py`

### Why it needs to get done

This allows a user to inspect the contents of the WorkTree if the build fails and minimizes disk usage.

PR #4390, opened 2023-10-03
chore(remote-build): migrate LaunchpadClient and WorkTree

### LINKED ISSUE #4323: remote-build: rework WorkTree and LaunchpadClient classes
### What needs to get done

Migrate the legacy remote-build code to snapcraft/remote_build/

This involves:

- simplifications to WorkTree class
    - no tarballing sources
    - no modifications to snapcraft.yaml
    - use new `pygit2` wrapper

### Why it needs to get done

remote-build needs to be migrated because it does not leverage the new craft libraries, has issues with building core22 snaps, and has issues related to how the local project is bundled.

mr-cal:

-----

Migrate `WorkTree`, `LaunchpadClient`, and their tests from `snapcraft_legacy`.  Reviewing by commit is the "easiest" (this is still a large code migration to review).

Scope:
- Add extra errors and utilities
- Tweak the `GitRepo` class
- Migrate `WorkTree`, `LaunchpadClient`, and their unit tests
- Modernize to current code conventions and style
- Replace snapcraft-specific code with generic code
- Update tests from `unit.TestCase` to `pytest`

Note that the WorkTree class is pretty empty now.  It will get populated more as part of #4324.

Fixes #4323 
(CRAFT-1978)


mr-cal:
Store test failure is unrelated

snapcraft/remote/utils.py
```diff
@@ -76,3 +78,27 @@ def _compute_hash(directory: Path) -> str:
...
+    """Shutil onerror function to make read-only files writable.
+
+    Try setting file to writeable if error occurs during rmtree. Known to be required
+    on Windows where file is not writeable, but it is owned by the user (who can
+    set file permissions).
+
+    :param filepath: filepath to make writable
+    """
+    filepath.chmod(stat.S_IWRITE)
+    func(filepath)
```
cmatsuoka, line 104: What happens if this fails again?

snapcraft/remote/worktree.py
```diff
@@ -0,0 +1,38 @@
...
+from snapcraft.remote.utils import rmtree
+
+
+class WorkTree:
+    """Class to manages tree for remote builds."""
+
+    def __init__(self, worktree_dir: Path) -> None:
+        """Create remote-build WorkTree.
+
+        :param str worktree_dir: Directory to use for working tree.
```
cmatsuoka, line 30: Outdated type listed in param line.

cmatsuoka (approved):
Did a pass and it looks ok, added a couple of questions and a note about a type in a docstring param line.

snapcraft/remote/launchpad.py
```diff
@@ -0,0 +1,478 @@
...
+            # Shouldn't end up here either.
+            self.cleanup()
+            raise errors.RemoteBuildError(
+                "Unknown builder error - no build entries found."
+            )
+
+        build_number = _get_url_basename(build_request.self_link)
+        logger.debug("Build request accepted: %s", build_number)
+
+    def login(self) -> Launchpad:
```
lengau, line 236: This seems like it would be a better fit as either a protected method or having a side effect of setting `self._lp`?

snapcraft/remote/launchpad.py
```diff
@@ -0,0 +1,478 @@
...
+        """Login to launchpad."""
+        try:
+            return Launchpad.login_with(
+                f"{self._app_name} remote-build",
+                "production",
```
lengau, line 241: This might be worth making an optional parameter that defaults to `"production"`. For certain integration/e2e tests we might want to talk to staging if the test could make a mess on launchpad.

lengau (changes requested):
Strong direction, but grumpy old man is yelling at cloud today

snapcraft/remote/launchpad.py
mr-cal: Agreed, I'm going to leave it as-is and we can add this in the a future PR.

snapcraft/remote/launchpad.py
mr-cal: Agreed.  I made it protected.

snapcraft/remote/git.py
```diff
@@ -145,27 +147,40 @@ def _init_repo(self) -> None:
...
+        :param token: token in the url to hide in logs and errors
 
         :raises GitError: if the ref cannot be resolved or pushed
         """
         resolved_ref = self._resolve_ref(ref)
         refspec = f"{resolved_ref}:refs/heads/{remote_branch}"
 
+        # hide secret tokens embedded in a url
+        if token:
+            stripped_url = re.sub(token, "<token>", remote_url)
```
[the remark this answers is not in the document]
mr-cal: Good catch!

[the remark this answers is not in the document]
mr-cal: Good catch!

snapcraft/remote/errors.py
```diff
@@ -50,3 +50,22 @@ def __init__(self, message: str) -> None:
         details = message
 
         super().__init__(brief=brief, details=details)
+
+
+class RemoteBuildTimeoutError(RemoteBuildError):
+    """Remote-build timed out."""
+
+    def __init__(self) -> None:
+        brief = "Remote build timed out."
```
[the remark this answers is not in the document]
mr-cal: I agree this error isn't very helpful.  However, the user shouldn't be too surprised to see it, since the only way you can get this error is by passing the command line argument `--launchpad-timeout=<seconds>`.

On a related note, I need to ask @sergiusens on why `--launchpad-timeout` was dropped in snapcraft 7.  I can't tell if it was intentional or not.

snapcraft/remote/utils.py
```diff
@@ -76,3 +78,27 @@ def _compute_hash(directory: Path) -> str:
...
+
+    :param directory: Directory to remove.
+    """
+    shutil.rmtree(
+        str(directory.resolve()),
+        onerror=_remove_readonly,  # type: ignore
+    )
+
+
+def _remove_readonly(func, filepath):
```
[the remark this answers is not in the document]
mr-cal: I can't figure this one out (I see why it was not tested in snapcraft_legacy).

I feel like I'm close with:
```python
def test_rmtree_readonly(mocker):
    """Remove a directory tree that contains a read-only file."""
    Path("root-dir").mkdir()
    Path("root-dir/test").touch()

    mocker.patch("os.unlink", side_effect=[OSError])

    rmtree(Path("root-dir"))

    assert not Path("root-dir").exists()
```

But it's not working and I don't like how dependent the test is on shutil's implementation.

snapcraft/remote/utils.py
mr-cal: It depends on what function is passed, but the I believe it will raise an OS or shutil error.

snapcraft/remote/errors.py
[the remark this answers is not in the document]
lengau: Yeah, this is more a vague UX-related thought, not a change request because as-is there isn't any possible change.

snapcraft/remote/utils.py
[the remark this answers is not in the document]
lengau: Yeah I don't have any good ideas about it, but the update is an improvement. Since this was brought from legacy I think it's not a stopper until we actually get this in the real world, and frankly if you'd excluded this function altogether I probably wouldn't have even thought about it.

lengau (approved):
Fancy snek approves this PR.

PR #4395, opened 2023-10-06
feat(remote-build): new command logic

### LINKED ISSUE #4324: remote-build: new command logic

mr-cal:

-----

- Migrate the command logic used to orchestrate a remote build.
  - Some of this code was moved to a new class, `RemoteBuild`
- Store a copy of the project directory in the local cache
- Clean up the cache after a build completes
- Typing and linting updates

Builds on #4390, so "only" the last 8 commits are for this PR (update: #4390 was landed and this PR is rebased onto main).  The new remote-build should be functional, so I encourage you to give it try.

Fixes #4324 
(CRAFT-1979)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,8 +53,6 @@ ignore_missing_imports = true
 follow_imports = "silent"
 exclude = [
     "build",
-    # launchpadlib is not typed
-    "snapcraft/remote/launchpad.py",
     "snapcraft_legacy",
     "tests/spread",
     "tests/legacy",
@@ -67,13 +65,7 @@ plugins = [
 
 [tool.pyright]
 include = ["snapcraft", "tests"]
-exclude = [
-    "build",
-    # launchpadlib is not typed
-    "snapcraft/remote/launchpad.py",
-    "tests/legacy",
-    "tests/spread",
-]
+exclude = ["build", "tests/legacy", "tests/spread"]
 pythonVersion = "3.10"
 
 [tool.pytest.ini_options]
--- a/snapcraft/commands/remote.py
+++ b/snapcraft/commands/remote.py
@@ -314,7 +314,7 @@ class RemoteBuildCommand(BaseCommand):
 
         return base
 
-    def _get_project_architectures(self) -> List[str]:
+    def _get_project_build_on_architectures(self) -> List[str]:
         """Get a list of build-on architectures from the project's snapcraft.yaml.
 
         :returns: A list of architectures.
@@ -347,11 +347,11 @@ class RemoteBuildCommand(BaseCommand):
         :raises SnapcraftError: If `--build-on` was provided and architectures are
         defined in the project's snapcraft.yaml.
         """
-        project_architectures = self._get_project_architectures()
+        project_architectures = self._get_project_build_on_architectures()
         if project_architectures and self._parsed_args.build_for:
             raise SnapcraftError(
                 "Cannot use `--build-on` because architectures are already defined in "
-                "the snapcraft.yaml."
+                "snapcraft.yaml."
             )
 
         if project_architectures:
--- a/snapcraft/commands/remote.py
+++ b/snapcraft/commands/remote.py
@@ -30,7 +30,7 @@ from overrides import overrides
 from snapcraft.errors import MaintenanceBase, SnapcraftError
 from snapcraft.legacy_cli import run_legacy
 from snapcraft.parts import yaml_utils
-from snapcraft.remote import AcceptPublicUploadError, get_build_id, is_repo
+from snapcraft.remote import AcceptPublicUploadError, RemoteBuilder, is_repo
 from snapcraft.utils import confirm_with_user, get_host_architecture, humanize_list
 
 _CONFIRMATION_PROMPT = (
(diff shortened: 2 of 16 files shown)


pyproject.toml
```diff
@@ -65,7 +67,13 @@ plugins = [
 
 [tool.pyright]
 include = ["snapcraft", "tests"]
-exclude = ["tests/legacy", "tests/spread", "build"]
+exclude = [
+    "build",
+    # launchpadlib is not typed
+    "snapcraft/remote/launchpad.py",
```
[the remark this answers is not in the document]
lengau: Personal opinion: I'd rather have a lot of type checker noise in `launchpad.py` than exclude the file entirely

snapcraft/remote/utils.py
```diff
@@ -76,3 +98,57 @@ def _compute_hash(directory: Path) -> str:
 
     all_hashes = "".join(hashes).encode()
     return md5(all_hashes).hexdigest()  # noqa: S324 (insecure-hash-function)
+
+
+def humanize_list(
```
lengau, line 103: This replicates `snapcraft.utils.humanize_list`.

Bigger picture: This is probably a good utility function to put into `craft-application`

lengau (changes requested):

snapcraft/remote/utils.py
lengau: Based on that, your call.

If remote build is going to become its own library that doesn't depend on `craft-application` (and given how we have this function in so many places), we may even want to make a little `craft-utils` library for all the little miscellany we have. (`humanize_list`, possibly architecture related stuff, maybe even our base error class. But I would want to keep it without non-stdlib dependencies if possible.)

sergiusens: craft cli is all about presentation though, utils risks being a dumpster of things we do not know or do not want to think of where to put

pyproject.toml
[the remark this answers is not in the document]
mr-cal: Well, I feel silly for excluding this file in the first place.  A few typing fixes and `casts` (thanks @lengau for the hint) solved dozens of linter errors.

To answer your question Claudio, `#type: ignore` in the import line doesn't work.

snapcraft/commands/remote.py
```diff
@@ -279,3 +308,53 @@ def _get_effective_base(self) -> str:
...
+        :returns: A list of architectures.
+
+        :raises SnapcraftError: If `--build-on` was provided and architectures are
+        defined in the project's snapcraft.yaml.
+        """
+        project_architectures = self._get_project_architectures()
+        if project_architectures and self._parsed_args.build_for:
+            raise SnapcraftError(
+                "Cannot use `--build-on` because architectures are already defined in "
+                "the snapcraft.yaml."
```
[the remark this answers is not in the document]
mr-cal: I think that the phrases `in snapcraft.yaml` or `in the snapcraft.yaml file` are both fine (I mixed the two here).  I'll use your suggestion.

snapcraft/commands/remote.py
```diff
@@ -279,3 +308,53 @@ def _get_effective_base(self) -> str:
...
+    def _determine_architectures(self) -> List[str]:
+        """Determine architectures to build for.
+
+        The build architectures can be set via the `--build-on` parameter or determined
+        from the build-on architectures listed in the project's snapcraft.yaml.
+
+        :returns: A list of architectures.
+
+        :raises SnapcraftError: If `--build-on` was provided and architectures are
+        defined in the project's snapcraft.yaml.
```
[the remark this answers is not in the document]
mr-cal: Like this?
```
architectures:
  - build-on: [amd64]
    build-for: [amd64]
  - build-on: [amd64]
    build-for: [arm64]
```

Unfortunately, we can't define architectures in a `snapcraft.yaml` file and request for Launchpad to only build some variants.

snapcraft/remote/utils.py
mr-cal: I decided to leave it as-is.

lengau: Yeah, that is a concern. `craft-cli` is also a good spot, but I wouldn't want to make `craft-parts` depend on it (and it currently uses its own `humanize_list` in 9 places). Although if we're reducing having it in all the apps + craft-parts to just craft-parts and craft-cli that's already a huge improvement.

I've made this to start out with: https://github.com/canonical/craft-cli/issues/183

snapcraft/remote/launchpad.py
```diff
@@ -102,7 +102,7 @@ def __init__(
         self._project_name = project_name
 
         self._lp: Launchpad = self._login()
-        self.user = self._lp.me.name
+        self.user = self._lp.me.name  # type: ignore
```
lengau, line 105:
```suggestion
        self.user = cast(str, self._lp.me.name)
```

should work too

snapcraft/remote/launchpad.py
```diff
@@ -160,28 +160,30 @@ def _fetch_artifacts(self, snap: Entry) -> None:
             self._download_build_artifacts(build)
             self._download_log(build)
 
-    def _get_builds_collection_entry(self, snap: Entry) -> Optional[Entry]:
+    def _get_builds_collection_entry(self, snap: Optional[Entry]) -> Optional[Entry]:
         logger.debug("Fetching builds collection information from Launchpad...")
-        url = snap.builds_collection_link
-        return self._lp_load_url(url)
+        url = snap.builds_collection_link  # type: ignore
+        return self._lp_load_url(cast(str, url))
```
lengau, line 166:
```suggestion
        url = cast(str, snap.builds_collection_link)
        return self._lp_load_url(url)
```

lengau (approved):
I've made two suggestions about typing in launchpad.py, but I haven't actually tried them to know whether both mypy *and* pyright will behave with them, and frankly there's no need for them.

snapcraft/remote/launchpad.py
mr-cal: I tried that, but the problem is with `_lp` being incorrectly typed:
```
error: Cannot access member "name" for type "NamedOperation"
   Member "name" is unknown (reportGeneralTypeIssues)
error: "name" is not a known member of "None"  (reportOptionalMemberAccess)
```

I can rewrite it as:
```
me = self._lp.me
self.user = me.name
```
and the same two warnings persist for the first line.

So I left it as-is.

snapcraft/remote/launchpad.py
mr-cal: Oo thanks.  You got me on the right track, check it out now.

snapcraft/remote/launchpad.py
lengau: Well that's disappointing... Thanks for checking!

lengau (approved):

snapcraft/remote/launchpad.py
lengau: Fantastic!

sergiusens (approved):

cmatsuoka (approved):
Awesome, thanks!

PR #4400, opened 2023-10-12
Return 1 if any remote-build failed

Hook25:
# Description
When using `snapcraft remote-build` in CI, it would be useful to get a failure when any of the remote build fails. This is not the behaviour that we currently have (namely, even if some build fails, the return code is `0` and the workflow is marked as ok. See: [core16-failure](https://github.com/canonical/checkbox/actions/runs/6199930599/job/16833507894)). Currently we are resorting to counting artifacts, but this process is fragile and error prone (For instance, we have discovered that the snap counting mechanism was introduced [here](https://github.com/canonical/checkbox/actions/runs/6199930599/workflow#L55) but not [here](https://github.com/canonical/checkbox/actions/runs/6491088080/workflow), leading to some failures not being detected)

This PR introduces a new flag to `remote-build`: `--exit-fail-on-any-failure`. When the flag is given, the remote build command will return 1 if any build fails, if the flag is not given it will just produce a debug message signaling the failure.
# Bug

Fixes:  https://github.com/snapcore/snapcraft/issues/4142

# Testing
I can't run the following due to this exception:
```
(venv)  snapcraft (return_1_remote_builds_failed) >  pytest tests/unit
ImportError while loading conftest '/home/h25/prj/canonical/snapcraft/tests/conftest.py'.
tests/conftest.py:21: in <module>
    import xdg.BaseDirectory
E   ModuleNotFoundError: No module named 'xdg.BaseDirectory'
```
-----

### DIFF
--- a/snapcraft/commands/remote.py
+++ b/snapcraft/commands/remote.py
@@ -128,8 +128,8 @@ class RemoteBuildCommand(BaseCommand):
         :param parsed_args: Snapcraft's argument namespace.
 
         :raises AcceptPublicUploadError: If the user does not agree to upload data.
-        :raises SnapcraftError: If the project cannot be loaded and parsed.
-        :raises SnapcraftError: If any build fails to complete.
+        :raises SnapcraftError: If the project cannot be loaded and parsed or if any
+        build fails to complete.
         """
         if os.getenv("SUDO_USER") and os.geteuid() == 0:
             emit.message(
--- a/snapcraft/commands/remote.py
+++ b/snapcraft/commands/remote.py
@@ -113,10 +113,11 @@ class RemoteBuildCommand(BaseCommand):
             default=0,
             metavar="<seconds>",
             help="Time in seconds to wait for launchpad to build.",
+        )
         parser.add_argument(
-            "--exit-fail-any-failure",
+            "--allow-failure",
             action="store_true",
-            help="return 1 if any of the builds fail"
+            help="return 0 even when builds fail"
         )
 
     @overrides
@@ -160,13 +161,14 @@ class RemoteBuildCommand(BaseCommand):
 
         success = self._run_new_or_fallback_remote_build(base)
         if not success:
-            if parsed_args.exit_fail_any_failure:
-                raise SnapcraftError("Some remote build failed to complete")
-            else:
+            if parsed_args.allow_failure:
                 emit.debug(
                     "Some build failed to complete but "
-                    "--exit-fail-any-failure was not provided"
+                    "--allow-failure was provided"
                 )
+            else:
+                raise SnapcraftError("Some remote build failed to complete")
+
 
 
     def _run_new_or_fallback_remote_build(self, base: str) -> bool:
--- a/snapcraft/commands/remote.py
+++ b/snapcraft/commands/remote.py
@@ -117,7 +117,8 @@ class RemoteBuildCommand(BaseCommand):
         parser.add_argument(
             "--allow-failure",
             action="store_true",
-            help="return 0 even when builds fail"
+            default=os.getenv("SNAPCRAFT_REMOTE_BUILD_ALLOW_FAILURE"),
+            help="return 0 even when builds fail",
         )
 
     @overrides
@@ -163,13 +164,10 @@ class RemoteBuildCommand(BaseCommand):
         if not success:
             if parsed_args.allow_failure:
                 emit.debug(
-                    "Some build failed to complete but "
-                    "--allow-failure was provided"
+                    "Some build failed to complete but --allow-failure was provided"
                 )
             else:
                 raise SnapcraftError("Some remote build failed to complete")
-
-
 
     def _run_new_or_fallback_remote_build(self, base: str) -> bool:
         """Run the new or fallback remote-build.
@@ -205,7 +203,7 @@ class RemoteBuildCommand(BaseCommand):
                 f"{_STRATEGY_ENVVAR!r} is {_Strategies.FORCE_FALLBACK.value!r}"
             )
             run_legacy()
-            return True # TODO: implement return of run_legacy
+            return True  # TODO: implement return of run_legacy
 
         if is_repo(Path().absolute()):
             emit.debug(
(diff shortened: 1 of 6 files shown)


snapcraft/commands/remote.py
```diff
@@ -101,6 +101,11 @@ def fill_parser(self, parser: argparse.ArgumentParser) -> None:
             action="store_true",
             help="acknowledge that uploaded code will be publicly available.",
         )
+        parser.add_argument(
```
sergiusens, line 104: we should probably fail if any of the build fails, I would start with that as a default

mr-cal: Perhaps a flag named `--allow-failure` to opt out?

@sergiusens - should we also offer an envvar like `SNAPCRAFT_REMOTE_BUILD_ALLOW_FAILURE`?

mr-cal: You should be able to do it just like this:
https://github.com/snapcore/snapcraft/blob/2e064759459924dbe163be07b081d902b61bf5e4/snapcraft/commands/lifecycle.py#L74

Hook25: Done!

snapcraft/commands/remote.py
```diff
@@ -110,6 +116,7 @@ def run(self, parsed_args: argparse.Namespace) -> None:
 
         :raises AcceptPublicUploadError: If the user does not agree to upload data.
         :raises SnapcraftError: If the project cannot be loaded and parsed.
+        :raises SnapcraftError: If any build fails to complete.
```
mr-cal, line 119:
```suggestion
        :raises SnapcraftError: If the project cannot be loaded and parsed or if any
        build fails to complete.
```

mr-cal (commented):
Looking good, thanks!  I think this still needs:
1. Unit test updates (some are now failing)
2. A new unit test that confirms a SnapcraftError is raised when one of the builds fail
3. A new unit test that confirms a Snapcraft ERror is not raised when the builds succeed
4. The same changes made in `snapcraft_legacy`'s remote build to return true/false

Hook25:
I'm afraid I will need some guidance to patch the legacy runner. If it is a straightforward thing I can do it. I don't really know what the legacy launcher is used for/what modifying it leads to, so please give me some input for that as well. 

If it works as I think it does, for our specific need (and the issue I have claimed to fix in this PR) this patch should be enough and it should cover most CI/CD needs as I suppose that most users use the latest version of snapcraft to trigger the remote build.

mr-cal:
> I'm afraid I will need some guidance to patch the legacy runner. If it is a straightforward thing I can do it. I don't really know what the legacy launcher is used for/what modifying it leads to, so please give me some input for that as well.
> 
> If it works as I think it does, for our specific need (and the issue I have claimed to fix in this PR) this patch should be enough and it should cover most CI/CD needs as I suppose that most users use the latest version of snapcraft to trigger the remote build.

Sure, if you don't mind I'd appreciate it.  It should be straightforward as the code is extremely similar and slightly simpler.

There are two places where `monitor_build` is called ([here](https://github.com/snapcore/snapcraft/blob/4b1cff546cde4ddd933e92ae928ba140a78cef2b/snapcraft_legacy/cli/remote.py#L170) and [here](https://github.com/snapcore/snapcraft/blob/main/snapcraft_legacy/cli/remote.py#L183)), which should look identical ([here](https://github.com/snapcore/snapcraft/blob/4b1cff546cde4ddd933e92ae928ba140a78cef2b/snapcraft_legacy/internal/remote_build/_launchpad.py#L336)).

I think you can re-use your same design of `monitor_build()` returning a bool then raising an error.

The main reason I want to have both the new and fallback remote-builders to have a consistent behavior is because the fallback remote-builder is still in use.  In fact, it's the only builder available on the `latest/stable` channel.  The new remote-build is currently only available on `latest/edge` and is planned to be released to stable with snapcraft 8.

mr-cal (changes requested):
Thanks, @Hook25!  

I think you will need to touch this file:
```diff
--- a/snapcraft/legacy_cli.py
+++ b/snapcraft/legacy_cli.py
@@ -29,7 +29,7 @@ _LIB_NAMES = ("craft_parts", "craft_providers", "craft_store", "snapcraft.remote
 _ORIGINAL_LIB_NAME_LOG_LEVEL: Dict[str, int] = {}


-def run_legacy(err: Optional[Exception] = None):
+def run_legacy(err: Optional[Exception] = None) -> bool:
     """Run legacy implementation."""
     # Reset the libraries to their original log level
     for lib_name in _LIB_NAMES:
@@ -43,4 +43,4 @@ def run_legacy(err: Optional[Exception] = None):
         emit.trace(f"run legacy implementation: {err!s}")
     emit.ended_ok()

-    legacy.legacy_run()
+    return legacy.legacy_run()
```

and if you don't mind adding two more unit tests for the legacy runner:
```
@pytest.mark.parametrize(
    "create_snapcraft_yaml", LEGACY_BASES | {"core22"}, indirect=True
)
@pytest.mark.usefixtures("create_snapcraft_yaml", "mock_argv", "mock_confirm")
def test_remote_build_legacy_ok(mock_run_legacy):
    """Successful legacy remote builds should have a return code of 0."""
    mock_run_legacy.return_value = True

    return_value = cli.run()

    assert return_value == 0


@pytest.mark.parametrize(
    "create_snapcraft_yaml", LEGACY_BASES | {"core22"}, indirect=True
)
@pytest.mark.usefixtures("create_snapcraft_yaml", "mock_argv", "mock_confirm")
def test_remote_build_legacy_failure(mock_run_legacy):
    """Unsuccessful legacy remote builds should have a return code of 1."""
    mock_run_legacy.return_value = False

    return_value = cli.run()

    assert return_value == 1
```

Finally, we will need some unit tests in `tests/legacy/unit/remote/test_launchpad.py` to ensure `monitor_build()` is working as expected, but those are slower to write with the existing `unit.TestCase` design.  I can help you with those if you'd like.

sergiusens:
thanks for this, an alternative implementation made its way in so this is now superseded

PR #4418, opened 2023-10-23
fix(remote-build) exponential backoff to LaunchpadClient `_lp_load_url`

mz2:
Exponential backoff to _lp_load_url made Launchpad requests, on the same lines as done in https://github.com/snapcore/snapcraft/commit/89ce40718#diff-0923e2475a672e5461493a8db8b751c8fd7937f88705ac631258dd3452c0c2efR97-R116 

I didn't add exponential backoff to the bit that inspired this which might be something to do too?
I did change `ConnectionResetError` to `ConnectionError` intentionally.

### DIFF
--- a/snapcraft_legacy/internal/remote_build/_launchpad.py
+++ b/snapcraft_legacy/internal/remote_build/_launchpad.py
@@ -193,12 +193,32 @@ class LaunchpadClient:
         )
 
     def _lp_load_url(self, url: str) -> Entry:
-        """Load Launchpad url with a retry in case the connection is lost."""
-        try:
-            return self._lp.load(url)
-        except ConnectionResetError:
-            self._lp = self.login()
-            return self._lp.load(url)
+        """Load Launchpad url with an exponential backoff retry in case the connection is lost."""
+        max_retries = 5
+        base_delay = 1
+        retry_count = max_retries
+
+        while retry_count > 0:
+            try:
+                return self._lp.load(url)
+            except ConnectionError as conn_error:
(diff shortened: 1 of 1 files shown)


snapcraft_legacy/internal/remote_build/_launchpad.py
```diff
@@ -193,12 +193,32 @@ def _issue_build_request(self, snap: Entry) -> Entry:
...
+                    "{!r} (retries left {})".format(url, retry_count)
+                )
+
+                if retry_count == 1:
+                    raise conn_error
+
+                self._lp = self.login()
+
+                delay = base_delay * (2 ** (max_retries - retry_count))
+                time.sleep(delay)
```
mr-cal, line 216: What do you think of adding a logger.debug message like `Failed to load URL.  Waiting <x> seconds before retrying.`?

mr-cal (changes requested):
Thanks for the PR, I think this is a great feature!

Would you be able to adjust the unit tests this breaks? 
- `tests/legacy/unit/remote_build/test_launchpad.py::LaunchpadTestCase::test_load_connection_refused
- `tests/legacy/unit/remote_build`/test_launchpad.py::LaunchpadTestCase::test_load_connection_reset_twice`

The second test may be more complex to change as you'll have to mock the sleep. And it would be great if you could confirm that the math is correct and it sleeps for 1, 2, 4, 8, then 16 seconds or a total of 31 seconds.

sergiusens:
closing due to lack of activity

PR #4462, opened 2023-12-01
build(deps): bump craft-providers to 1.20.1

mr-cal:

-----

This resolves a bug introduced in `1.16.0` where improperly set-up base instances would cause snapcraft to fail with the error `Instance setup failed. Check LXD logs for more details`.

#### 1.20.1

- Update base compatibility tag from ``base-v3`` to ``base-v4``
- If an existing base instance is not setup, then it is auto-cleaned.
  If the process that created the not setup base instance is inactive, then
  ``craft-providers`` will immediately auto-clean the instance.

#### 1.20.0
- Snaps injected from the host will have their base snap injected into
  the instance.

### DIFF
--- a/requirements-devel.txt
+++ b/requirements-devel.txt
@@ -15,7 +15,7 @@ craft-archives==1.1.3
 craft-cli==1.2.0
 craft-grammar==1.1.1
 craft-parts==1.19.7
-craft-providers==1.19.2
+craft-providers==1.20.1
 craft-store==2.5.0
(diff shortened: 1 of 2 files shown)


lengau (approved):

cmatsuoka (approved):

mr-cal:
Merging with core18 spread test failures because they are unrelated (https://github.com/flutter/flutter/issues/133881)

PR #4477, opened 2023-12-04
snapcraft: Allow equal sign in commands

3v1n0:
In various cases applications require to use --option=value syntax but this is  currently blocked by snapcraft syntax checker.

So add the equal sign as a valid value in the regex since it does seems to be harmful.

See also: https://forum.snapcraft.io/t/please-allow-equals-symbol-in-snapcraft-yaml/36709

---

-----

### DIFF
--- a/schema/snapcraft.json
+++ b/schema/snapcraft.json
@@ -857,7 +857,7 @@
                             "type": "array",
                             "items": {
                                 "type": "string",
-                                "pattern": "^[A-Za-z0-9/._#:$-]*$",
+                                "pattern": "^[A-Za-z0-9/._#:$-=]*$",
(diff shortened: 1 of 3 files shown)


mr-cal (commented):
Hi @3v1n0, I appreciate the PR but unfortunately, this can't be fixed within snapcraft. 

During packaging, Snapcraft calls `snap pack`, which does not support equal signs in the `command` field (see the [launchpad bug](https://bugs.launchpad.net/snapd/+bug/1827392) for details).

The regex preventing equal signs in Snapcraft exists for two reasons:
1. To raise the error as quickly as possible
2. To present the user with a more readable error message than the one that `snap pack` raises

sergiusens:
thanks, but we cannot enable this before the feature in snapd is completed (mentioned in the forum post)

Issue #4142, opened 2023-05-15
snapcraft remote-build exits with return code 0 even when some snaps failed to build for a given arch


pieqq:
### Bug Description

We use `snapcraft remote-build` to build snaps for different architectures as part of our continuous delivery process.

Sometimes, something goes wrong and the snap for one of the architectures fails to build, yet snapcraft does not return any error, but simply a `Build failed for arch ***` message, and exits with return code `0`. We then have no idea that some of the snaps have not been properly created (and pushed to the store) unless we manually check the logs or the snapstore.

It would be better if snapcraft returned `1`, so that we could get a notification to investigate the issue and maybe re-trigger a build.

### To Reproduce

Use `snapcraft remote-build` to build a snap on different architectures (amd64, arm64, armhf, etc.)

We build a bunch of snaps, you can find the related `snapcraft.yaml` in [our git repository](https://github.com/canonical/checkbox/tree/main/checkbox-core-snap).

### Environment

- Using snapcraft from 7.x/stable channel.
- Launch command: `snapcraft remote-build --build-on amd64,arm64,armhf,i386 --launchpad-accept-public-upload`
- Ubuntu 22.04 (using Github Actions, see [workflow](https://github.com/canonical/checkbox/actions/runs/4977536729/workflow))

### snapcraft.yaml

```shell
name: checkbox20
summary: Checkbox runtime and public providers
description: "Checkbox runtime and public providers"
grade: stable
confinement: strict

version: '2.6'

base: core20

# Don't forget to add a new slot if a new provider part is added in the parts
# section below.
slots:
  provider-resource:
    interface: content
    read:
      - $SNAP/providers/checkbox-provider-resource
  provider-checkbox:
    interface: content
    read:
      - $SNAP/providers/checkbox-provider-base
  provider-docker:
    interface: content
    read:
      - $SNAP/providers/checkbox-provider-docker
  provider-tpm2:
    interface: content
    read:
      - $SNAP/providers/checkbox-provider-tpm2
  provider-iiotg:
    interface: content
    read:
      - $SNAP/providers/checkbox-provider-iiotg
  provider-engineering-tests:
    interface: content
    read:
      - $SNAP/providers/checkbox-provider-engineering-tests
  provider-sru:
    interface: content
    read:
      - $SNAP/providers/checkbox-provider-sru
  provider-gpgpu:
    interface: content
    read:
      - $SNAP/providers/checkbox-provider-gpgpu
  provider-certification-client:
    interface: content
[...]

PR #4480, opened 2023-12-05
fix(remote): return error when remote build failed

syu-w:
Fixes: #4142

### DIFF
--- a/snapcraft/commands/remote.py
+++ b/snapcraft/commands/remote.py
@@ -257,17 +257,19 @@ class RemoteBuildCommand(BaseCommand):
                 "Do you wish to recover this build?", default=True
             ):
                 emit.progress("Building")
-                remote_builder.monitor_build()
-                emit.progress("Cleaning")
-                remote_builder.clean_build()
-                emit.progress("Build completed", permanent=True)
+                try:
+                    remote_builder.monitor_build()
+                finally:
+                    emit.progress("Cleaning")
+                    remote_builder.clean_build()
+                    emit.progress("Build task(s) completed", permanent=True)
                 return
 
             # Otherwise clean running build before we start a new one.
             emit.progress("Cleaning existing build")
             remote_builder.clean_build()
         else:
-            emit.progress("No existing build found", permanent=True)
+            emit.progress("No existing build task(s) found", permanent=True)
 
         emit.progress(
             "If interrupted, resume with: 'snapcraft remote-build --recover "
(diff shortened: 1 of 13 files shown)


snapcraft/commands/remote.py
```diff
@@ -277,10 +279,12 @@ def _run_new_remote_build(self) -> None:
...
-        remote_builder.monitor_build()
-        emit.progress("Cleaning")
-        remote_builder.clean_build()
-        emit.progress("Build completed", permanent=True)
+        try:
+            remote_builder.monitor_build()
+        finally:
+            emit.progress("Cleaning")
+            remote_builder.clean_build()
+            emit.progress("Build completed", permanent=True)
```
lengau, line 287: This will say the build completed even when it raises an exception. I'm guessing the correct behaviour is closer to:

```suggestion
        try:
            remote_builder.monitor_build()
        except Exception:
            emit.progress("Build failed", permanent=True)
            raise
        finally:
            emit.progress("Cleaning")
            remote_builder.clean_build()
        emit.progress("Build completed", permanent=True)
```

lengau (changes requested):

snapcraft/commands/remote.py
syu-w: Shows "build completed" even when it raises an exception is what I want. The `monitor_build()` included all sub building task's results that some of them may fail. Like `amd64` success but `arm64` failed. But the build task is completed.

syu-w: Change it to `build task(s) completed` maybe more clearer

mr-cal (changes requested):
Nice work!  Just a minor comment.

snapcraft/cli.py
```diff
@@ -300,6 +300,9 @@ def run():  # noqa: C901 (complex-structure)
     except errors.LinterError as err:
         emit.error(craft_cli.errors.CraftError(f"linter error: {err}"))
         retcode = err.exit_code
+    except RemoteBuildFailedError as err:
```
mr-cal, line 303: I don't think we should catch `RemoteBuild` and a subclassed error of `RemoteBuild`.  I say we keep it simple and only catch `RemoteBuild` errors.

syu-w: OK, the only downside is the error message become: `remote-build error: Remote build failed.`. It feels a little duplicative.

mr-cal (approved):
Tested and it looks good :+1:

lengau (approved):
Thanks!

snapcraft/commands/remote.py
```diff
@@ -257,17 +257,19 @@ def _run_new_remote_build(self) -> None:
                 "Do you wish to recover this build?", default=True
             ):
                 emit.progress("Building")
-                remote_builder.monitor_build()
-                emit.progress("Cleaning")
-                remote_builder.clean_build()
-                emit.progress("Build completed", permanent=True)
+                try:
+                    remote_builder.monitor_build()
+                finally:
```
sergiusens, line 262: is this safe for the resume build feature?

syu-w: It has two scenarios:
1. Snapcraft got killed / reboot / network lost.
2. User Ctrl+C.

It should safe for the first one. But for the current code, if user Ctrl+C in the middle, it will run `cleanup()` and not resume-able. If we also want to resume it after Ctrl+C, it needs to remove the `finally`.

sergiusens (approved):

PR #4490, opened 2023-12-10
refactor: replace `pkg_resources` with `packaging`

### LINKED ISSUE #4407: Replace `pkg_resources.parse_version`
### What needs to get done

The `is_linker_compatible()` and `get_required_glibc()` calls need to be updated to not use this package.

### Why it needs to get done

setuptools>=66 deprecated pkg_resources

branchvincent:

---

Fixes #4407

This removes the `setuptools<66` pin (which also unblocks python 3.12 support):

- `pkg_resources.parse_version` is replaced with `packaging.version.parse` (which now only accepts PEP 440-compliant version strings. I only made the changes needed to pass the unit tests, but maybe we should be more defensive)
- `pkg_resources.require` is replaced with `importlib.metadata.version` (added in [python 3.8](https://docs.python.org/3/library/importlib.metadata.html))

I've also updated `setup.py` with env markers (see the caveat in the commit message, lmk if that's not ok) and a minimum python (not sure if this should be `3.9` or `3.10`)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -74,6 +74,7 @@ dev_requires = [
     "pycodestyle",
     "pydocstyle",
     "pyftpdlib",
+    "pyinstaller; sys_platform == 'win32'",
     "pylint",
     "pylint-fixme-info",
     "pylint-pytest",
@@ -93,11 +94,9 @@ dev_requires = [
     "yamllint",
 ]
 
-if sys.platform == "win32":
-    dev_requires.append("pyinstaller")
-
 install_requires = [
     "attrs",
+    "catkin-pkg; sys_platform == 'linux'",
     "click",
     "craft-archives",
     "craft-cli",
@@ -119,7 +118,10 @@ install_requires = [
     # Pygit2 and libgit2 need to match versions.
     # Further info: https://www.pygit2.org/install.html#version-numbers
     "pygit2~=1.13.0",
+    "pylxd; sys_platform == 'linux'",
     "pymacaroons",
+    "python-apt @ https://launchpad.net/ubuntu/+archive/primary/+sourcefiles/python-apt/2.4.0ubuntu1/python-apt_2.4.0ubuntu1.tar.xz ; sys_platform == 'linux'",
+    "python-debian; sys_platform == 'linux'",
     "pyxdg",
     "pyyaml",
     "raven",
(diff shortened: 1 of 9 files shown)


sergiusens (approved):
thank you very much!

setup.py
```diff
@@ -185,6 +162,7 @@ def recursive_data_files(directory, install_directory):
         + recursive_data_files("keyrings", "share/snapcraft")
         + recursive_data_files("extensions", "share/snapcraft")
     ),
+    python_requires=">=3.10",
```
mr-cal, line 165: Reply to your PR description:

Yes, you made the right choice since we build snapcraft on 3.10.

setup.py
```diff
@@ -114,20 +112,22 @@ def recursive_data_files(directory, install_directory):
...
     "overrides",
+    "packaging",
     "progressbar",
     "pyelftools",
     # Pygit2 and libgit2 need to match versions.
     # Further info: https://www.pygit2.org/install.html#version-numbers
     "pygit2~=1.13.0",
+    "pylxd; sys_platform == 'linux'",
     "pymacaroons",
+    "python-apt @ https://launchpad.net/ubuntu/+archive/primary/+sourcefiles/python-apt/2.4.0ubuntu1/python-apt_2.4.0ubuntu1.tar.xz ; sys_platform == 'linux'",
```
mr-cal, line 123: @lengau - Question for your packaging expertise (and I recall you touching this dep in another repo this year).

This is only going to break developers using non-ubuntu distros trying to install snapcraft from source?

I think the solution will be to require using the snapcraft dev container (or the new rock).

mr-cal (approved):
Thank you!

setup.py
lengau: Yeah this one's a weird one. I can see the following effects:

1. Prevents publishing on pypi (I don't think that's relevant since we haven't published snapcraft there since the 4.x branch.)
2. Non-apt-based systems won't be able to build snapcraft without modifying this file.
3. Ubuntu systems with an incompatible version of apt won't be able to build snapcraft without modifying this file to provide their own source tarball (already true)

I think the best strategy for our `setup.py` and `pyproject.toml` files is probably what [craft-application](https://github.com/canonical/craft-application/blob/main/pyproject.toml#L63-L65) and [craft-parts](https://github.com/canonical/craft-parts/blob/aa0416efcdc68af34fdaf60263c8a48c67e39458/setup.py#L109) have.

The reason for requiring `>=2.0` is a workaround for the fact that there are old versions of python-apt published on pypi. This forces the user to provide another source for the package. In Rockcraft and Charmcraft we do that through a [separate requirements file](https://github.com/canonical/rockcraft/blob/main/requirements-jammy.txt) for common systems, so that might be the best way to do it here too.

Unrelated, but there is also a newer source package for jammy: https://launchpad.net/ubuntu/+archive/primary/+sourcefiles/python-apt/2.4.0ubuntu2/python-apt_2.4.0ubuntu2.tar.xz

setup.py
lengau: We probably should have done this when we changed the linters over 😆

lengau (approved):
Thanks for doing this! It's a significant improvement.

setup.py
lengau: After a bit further thought, I think this is fine. It doesn't provide any regressions in behaviour, but it did bring up something I should've fixed. I made an issue for it here: https://github.com/snapcore/snapcraft/issues/4496

branchvincent:
thanks for the reviews everyone!

PR #4491, opened 2023-12-11
feat(plugins): add Matter plugin support

MonicaisHer:

-----

This PR adds a new Matter plugin which is useful for building [Matter SDK](https://github.com/project-chip/connectedhomeip) based parts.

This PR could be merged once  the following tasks been completed:
- [x] add unit test
- [x] add spread test to build a matter snap


snapcraft/parts/plugins/matter_plugin.py
```diff
@@ -0,0 +1,138 @@
...
+
+MATTER_REPO = "https://github.com/project-chip/connectedhomeip.git"
+"""The repository where the matter SDK resides."""
+
+
+class MatterPluginProperties(plugins.PluginProperties, plugins.PluginModel):
+    """The part properties used by the matter plugin."""
+
+    # matter_branch: str
+    # zap_version: str
```
MonicaisHer, line 33: The Snapcraft could be built using the new Matter plugin and successfully create a snap with hardcoded arguments (`matter_branch` and `zap_version`). However, currently it's not yet possible to pass these two arguments dynamically. Below is an example of how this plugin has been used in the snapcraft.yaml file and the build errors it throws:
```
parts:
  chip-tool:
    plugin: matter
    matter-branch: "v1.2.0.1"
    zap-version: "v2023.11.13"
    override-build: |
...
```
```
$ snap info snapcraft
installed:          6.0.post966+git4eed4be8               (x1) 78MB classic
$ snapcraft -v --debug
...
Bad snapcraft.yaml content:
- extra field 'matter-branch' not permitted in 'parts.chip-tool' configuration 
```

snapcraft/parts/plugins/matter_plugin.py
```diff
@@ -0,0 +1,138 @@
...
+    @overrides
+    def get_build_commands(self) -> List[str]:
+        commands = []
+
+        if self.snap_arch == "arm64":
+            commands.extend(
+                [
+                    f"wget --no-verbose https://github.com/project-chip/zap/releases/download/v2023.11.13/zap-linux-{self.snap_arch}.zip",
+                    f"unzip -o zap-linux-{self.snap_arch}.zip",
+                    "echo 'export ZAP_INSTALL_PATH=$PWD'",
```
sergiusens, line 123: why not put this in `get_build_environment`?

snapcraft/parts/plugins/matter_plugin.py
```diff
@@ -0,0 +1,138 @@
...
+                f"if [ ! -d matter ]; then git clone --depth 1 -b v1.2.0.1 {MATTER_REPO} matter; fi",
+                "cd matter || echo 'skip clone'",
+                f"scripts/checkout_submodules.py --shallow --platform linux",
+                f"set +u && source scripts/activate.sh && set -u",
+                "cp -vr ./* $CRAFT_PART_INSTALL/",
+                "echo 'Cloned Matter repository and activated submodules'",
+            ]
+        )
+
+        return commands
```
sergiusens, line 138: this matter plugin doesn't really build anything, does it? it might be more suitable as an extension

farshidtz: Sourcing `scripts/activate.sh` (line 132) builds the SDK and sets up the environment in preparation for the application build. In the recent versions, this is actually just a [symbolic link](https://github.com/project-chip/connectedhomeip/blob/v1.2.0.1/scripts/activate.sh) to [scripts/setup/bootstrap.sh](https://github.com/project-chip/connectedhomeip/blob/v1.2.0.1/scripts/setup/bootstrap.sh), kept there for backward compatibility. I suggest replacing it with the actual target, which has a clearer name.

snapcraft/parts/plugins/matter_plugin.py
MonicaisHer: This environment variable(`ZAP_INSTALL_PATH`) depends on the success of these two build commands(`wget...`, `unzip...`). If these two commands fail, setting the environment variable would be meaningless. Moving it to a separate `get_build_environment` might cause confusion.

snapcraft/parts/plugins/matter_plugin.py
MonicaisHer: This Matter plugin builds the Matter SDK. Some comments have been added for better clarity.

snapcraft/parts/plugins/matter_plugin.py
MonicaisHer: One argument in the snapcraft.yaml has been accepted by snapcraft after rebased the working branch onto tags 8.0.0 instead of main branch.

mr-cal: Also, rebasing on `8.0.0` shouldn't make a difference vs `main` (there are only a few commits between the two that should be unrelated).  We will have to rebase this PR onto `main` to merge anyways.

On `8.0.0` and `main`, I got the same error at the same place:
```
2023-12-13 12:50:53.026 :: pw_cli.plugins.Error: Failed to import module "pw_watch.watch" (ImportError: cannot import name 'OneStyleAndTextTuple' from 'prompt_toolkit.formatted_text' (/root/snapcraft-examples/matter/parts/chip-tool/build/matter/.environment/pigweed-venv/lib/python3.11/site-packages/prompt_toolkit/formatted_text/__init__.py))
2023-12-13 12:50:53.080 :: ./scripts/setup/bootstrap.sh: line 140: pop_var_context: head of shell_variables not a function context
2023-12-13 12:50:53.080 :: ./scripts/setup/bootstrap.sh: line 153: pop_var_context: head of shell_variables not a function context
2023-12-13 12:50:53.258 'override-build' in part 'chip-tool' failed with code 1.
Review the scriptlet and make sure it's correct.
```

MonicaisHer: Thank you for pointing out the keyword prefix. The updates have been made on both snapcraft and [example snap](https://github.com/MonicaisHer/chip-tool-snap/blob/test-matter-plugin/snap/snapcraft.yaml#L23-L25) sides. However, although snapcraft appears to accept these keywords, it still fails to map `x-y` to `x_y`, resulting in a following error when running snapcraft:
```
# clean install the snapcraft built from this PR
$ sudo snap remove --purge snapcraft
$ sudo snap install snapcraft/snapcraft_8* --dangerous --classic

# build an example snap: 
$ cd chip-tool-snap
$ snapcraft clean
$ snapcraft -v --debug
...
Building chip-tool                                                                                                                                                             
:: + craftctl default                                                                                                                                                          
'MatterPlugin' object has no attribute 'matter_tag'
```

And you are right; the rebase shouldn't matter in this context. It could be caused by cache. Regarding the error you're encountering in `./scripts/setup/bootstrap.sh`, I haven't had a chance to encounter the same error due to the previous error: `MatterPlugin' object has no attribute 'matter_tag'`.

MonicaisHer: Thank you for pointing out the keyword prefix. It is working now! And you are right; the rebase shouldn't matter in this context. It could be caused by cache.

MonicaisHer: The `OneStyleAndTextTuple` error you encountered is a [bug](https://github.com/project-chip/connectedhomeip/issues/30981) from prompt-toolkit. This issue has been [fixed](https://github.com/project-chip/connectedhomeip/pull/30987) in the latest Matter SDK. By using the fix commit and the Matter plugin, example snap can be built successfully now.

snapcraft/parts/plugins/matter_plugin.py
```diff
@@ -0,0 +1,171 @@
...
+
+MATTER_REPO = "https://github.com/project-chip/connectedhomeip.git"
+"""The repository where the matter SDK resides."""
+
+
+class MatterPluginProperties(plugins.PluginProperties, plugins.PluginModel):
+    """The part properties used by the matter plugin."""
+
+    matter_tag: str
+    matter_zap_tag: str
```
farshidtz, line 33: These could be misleading. The first property is the Matter SDK git references (tags, branches), which differ from Matter standard versions. I suggest adding "sdk" to avoid confusion with Matter standard. Moreover, replace `tag` with `version` to make it less cryptic, since the end user doesn't see that this is used for checking out a git repo.

```suggestion
    matter_sdk_version: str
    matter_zap_version: str
```

snapcraft/parts/plugins/matter_plugin.py
```diff
@@ -0,0 +1,171 @@
...
+        if self.snap_arch == "arm64":
+            commands.extend(
+                [
+                    f"wget --no-verbose https://github.com/project-chip/zap/releases/download/{options.matter_zap_tag}/zap-linux-{self.snap_arch}.zip",
+                    f"unzip -o zap-linux-{self.snap_arch}.zip",
+                    "echo 'export ZAP_INSTALL_PATH=$PWD'",
+                ]
+            )
+
+        """Clone Matter repository if not present"""
```
farshidtz, line 126: Using string literals is preferred for comment blocks. These are used in several other places for single line comments.
```suggestion
        # Clone Matter repository if not present
```

snapcraft/parts/plugins/matter_plugin.py
```diff
@@ -0,0 +1,171 @@
...
+        directory within the confined snap space on /mnt.
+        """
+        """Replace storage paths"""
+        commands.extend(
+            ["sed -i 's/\/tmp/\/mnt/g' src/platform/Linux/CHIPLinuxStorage.h"]
+        )
+        """Replace key-value store path"""
+        commands.extend(
+            ["sed -i 's/\/tmp/\/mnt/g' src/platform/Linux/CHIPPlatformConfig.h"]
+        )
```
farshidtz, line 162: Shouldn't these be done before calling `bootstrap.sh` to be effective?

farshidtz (changes requested):
It is a very good start, which can be followed by adding tests.

My biggest concern on the consumer side [based on your example](https://github.com/MonicaisHer/chip-tool-snap/blob/9a23cb333361dc5f321d7e5997163f7ffe4c0c44/snap/snapcraft.yaml#L22C1-L31) is that every time the application code changes, the Matter SDK data will be removed, and the bootstrapping process (the most resource consuming step) will be repeated from scratch. This is because the plugin and application data are maintained under the same part's build directory, which gets cleared when app sources change.

snapcraft/parts/plugins/matter_plugin.py
MonicaisHer: Updated. Please note that there are still several places where `"""` is used for single-line comments, just for formatting alignment with other plugins. Let me know if all single-line comments need to be updated to use `#`.

snapcraft/parts/plugins/matter_plugin.py
```diff
@@ -0,0 +1,173 @@
...
+import os
+
+from typing import Any, Dict, List, Set, cast
+
+from craft_parts import infos, plugins
+from overrides import overrides
+
+MATTER_REPO = "https://github.com/project-chip/connectedhomeip.git"
+"""The repository where the matter SDK resides."""
+
```
syu-w, line 27:
```suggestion
# The repository where the matter SDK resides.
MATTER_REPO = "https://github.com/project-chip/connectedhomeip.git"
```

snapcraft/parts/plugins/matter_plugin.py
```diff
@@ -0,0 +1,173 @@
...
+        See the top-level layout definition which bind mounts a persisted
+        directory within the confined snap space on /mnt.
+        """
+        """Replace storage paths"""
+        commands.extend(
+            [
+                r"sed -i 's/\/tmp/\/mnt/g' src/platform/Linux/CHIPLinuxStorage.h",
+                r"sed -i 's/\/tmp/\/mnt/g' src/platform/Linux/CHIPPlatformConfig.h",
+            ]
+        )
```
syu-w, line 156: I think you may want to put this into `/var/lib/matter` if you want to it persisted. But if it is default to `/tmp`, that should able to lose without issue.

snapcraft/parts/plugins/matter_plugin.py
```diff
@@ -0,0 +1,173 @@
...
+        commands.extend(
+            ["set +u && source setup/bootstrap.sh --platform build && set -u"]
+        )
+
+        commands.extend(
+            [
+                "cp -vr ./* $CRAFT_PART_INSTALL/",
+                "echo 'Cloned Matter repository and built Matter SDK'",
+            ]
+        )
```
syu-w, line 171: Do you really need copy everything to the part? I am not sure `bootstrap.sh` doing but there is not a finally result to use?

syu-w:
Also you can run `tox -m lint` locally to check any linters error so you fix them quicker.

snapcraft/parts/plugins/matter_plugin.py
farshidtz: Good catch. I think nothing should be staged here. Only the application's binary distribution, built by the consumer snap needs to be staged.

snapcraft/parts/plugins/matter_plugin.py
mr-cal: +1 unless you want developers to see this URL and comment in the auto-generated documentation (which currently don't exist for snapcraft)

It will render [this](https://github.com/canonical/craft-parts/blob/8ecf07c0c22b607039642af956ba0b9d400b860e/craft_parts/plugins/validator.py#L32) to [this](https://canonical-craft-parts.readthedocs-hosted.com/en/latest/reference/gen/craft_parts.plugins.validator.html#craft_parts.plugins.validator.COMMAND_NOT_FOUND)

snapcraft/parts/plugins/matter_plugin.py
```diff
@@ -0,0 +1,173 @@
...
+        part_info: infos.PartInfo,
+    ) -> None:
+        super().__init__(properties=properties, part_info=part_info)
+
+        self.matter_dir = part_info.part_build_dir
+        self.snap_arch = os.getenv("SNAP_ARCH")
+
+    @overrides
+    def get_build_packages(self) -> Set[str]:
+        return {
```
mr-cal, line 82: nit: can you alphabetize these for readability?

mr-cal (commented):
In addition to unit tests, we'll also need a spread test to build a `hello world` matter snap.

It would probably live in [`tests/spread/plugins/craft-parts`](https://github.com/snapcore/snapcraft/tree/73afa9dc6f4acd58b1be9fe886ec7e964fe1d3d3/tests/spread/plugins/craft-parts), but possibly in [`tests/spread/plugins/craft-parts/build-and-run-hello`](https://github.com/snapcore/snapcraft/tree/73afa9dc6f4acd58b1be9fe886ec7e964fe1d3d3/tests/spread/plugins/craft-parts/build-and-run-hello) if you can build an app that outputs 'Hello world" when executed.

snapcraft/parts/plugins/matter_plugin.py
MonicaisHer: Sure! it has been addressed in the latest commit.

snapcraft/parts/plugins/matter_plugin.py
MonicaisHer: Thanks. You are right, the copy is redundant. It has been removed in the latest commit.

snapcraft/parts/plugins/matter_plugin.py
MonicaisHer: Thanks for the explanation of snapcraft rendering. The suggestion has been applied in the latest commit

snapcraft/parts/plugins/matter_plugin.py
MonicaisHer: This has been placed into `/mnt`, and `/mnt` has been bound to `$SNAP_COMMON/mnt`, as shown in the example [here](https://github.com/MonicaisHer/chip-tool-snap/blob/test-matter-plugin/snap/snapcraft.yaml#L17-L19). Indeed, the location of persisted data could be improved to `/var/lib/matter` rather than `/mnt` on both the plugin side and consumer snap side.

Although it defaults to `/tmp`, the consumer snap needs the data to remain consistent after reboot.

MonicaisHer:
> It is a very good start, which can be followed by adding tests.
> 
> My biggest concern on the consumer side [based on your example](https://github.com/MonicaisHer/chip-tool-snap/blob/9a23cb333361dc5f321d7e5997163f7ffe4c0c44/snap/snapcraft.yaml#L22C1-L31) is that every time the application code changes, the Matter SDK data will be removed, and the bootstrapping process (the most resource consuming step) will be repeated from scratch. This is because the plugin and application data are maintained under the same part's build directory, which gets cleared when app sources change.

Thank you for the review. [The usage of consumer snap](https://github.com/snapcore/snapcraft/blob/61adb5e73fd39ac310cdc7f7a788bb39c6701328/tests/spread/plugins/craft-parts/matter-sdk/snapcraft.yaml#L27-L50) has been improved and added to tests. The Matter SDK plugin has now become a standalone `part` and the application building `part` depending on it. This means that each time the application building `part` is modified, the building will be much faster, as the Matter SDK plugin will not need to run again.

MonicaisHer:
> In addition to unit tests, we'll also need a spread test to build a `hello world` matter snap.
> 
> It would probably live in [`tests/spread/plugins/craft-parts`](https://github.com/snapcore/snapcraft/tree/73afa9dc6f4acd58b1be9fe886ec7e964fe1d3d3/tests/spread/plugins/craft-parts), but possibly in [`tests/spread/plugins/craft-parts/build-and-run-hello`](https://github.com/snapcore/snapcraft/tree/73afa9dc6f4acd58b1be9fe886ec7e964fe1d3d3/tests/spread/plugins/craft-parts/build-and-run-hello) if you can build an app that outputs 'Hello world" when executed.

The unit test and spread test have been added. There is [a snap using the Matter SDK plugin](https://github.com/snapcore/snapcraft/tree/61adb5e73fd39ac310cdc7f7a788bb39c6701328/tests/spread/plugins/craft-parts/matter-sdk), and in the spread test, it checks whether the application was successfully initialized and the storage path was successfully replaced.

snapcraft/parts/plugins/matter_sdk_plugin.py
```diff
@@ -0,0 +1,192 @@
...
+        commands.extend(["echo 'Built Matter SDK'"])
+
+        # Compare and output pigweed related environment variables to matter_sdk_env env file
+        commands.extend(
+            [
+                "set -a",
+                'echo "PATH=$PATH" >> matter_sdk_env',
+                'env | grep "^PW_" >> matter_sdk_env',
+                'echo "VIRTUAL_ENV=$VIRTUAL_ENV" >> matter_sdk_env',
+                'echo "CIPD_CACHE_DIR=$CIPD_CACHE_DIR" >> matter_sdk_env',
```
MonicaisHer, line 185:
```suggestion
               
```

farshidtz (changes requested):

mr-cal:
> In the spread test, right now it checks the Matter SDK related functionality whether the application was successfully initialized and whether the storage path was successfully replaced. I want to ensure that I understand the requirements correctly. Is it required for the app built in the spread test also to output "Hello world" when executed, or is it simply a suggested outcome?

Outputting "hello world" was just a suggestion, since the tests can be integrated into the `build-and-run-hello` test suite.
It makes sense for traditional plugins (like Python or autotools), where we can write a simple program to output 'hello world'.

Given that this plugin is providing an SDK, I think your approach to the spread test is good.  No need to do the 'hello world' tests.

MonicaisHer:
Convert this pull request to a draft, as there are review feedback and feedback from manual testing need to be addressed first.

snapcraft/parts/plugins/matter_sdk_plugin.py
```diff
@@ -178,14 +167,8 @@ def get_build_commands(self) -> List[str]:
...
             [
-                "set -a",
-                'echo "PATH=$PATH" >> matter_sdk_env',
-                'env | grep "^PW_" >> matter_sdk_env',
-                'echo "VIRTUAL_ENV=$VIRTUAL_ENV" >> matter_sdk_env',
-                'echo "CIPD_CACHE_DIR=$CIPD_CACHE_DIR" >> matter_sdk_env',
-                "set +a",
-                "echo 'pigweed related environment variables differences exported",
-                "to matter_sdk_env file'",
+                'echo "export PATH=$PATH" >> matter_sdk_env',
```
farshidtz, line 170: The `PATH` variable at this stage contains the following: `/root/parts/lighting-hello/build/matter/third_party/pigweed/repo/out/host/host_tools:/root/parts/lighting-hello/build/matter/.environment/pigweed-venv/bin:/root/parts/lighting-hello/build/matter/.environment/cipd/packages/arm/bin:/root/parts/lighting-hello/build/matter/.environment/cipd/packages/arm:/root/parts/lighting-hello/build/matter/.environment/cipd/packages/zap:/root/parts/lighting-hello/build/matter/.environment/cipd/packages/pigweed/bin:/root/parts/lighting-hello/build/matter/.environment/cipd/packages/pigweed:/root/parts/lighting-hello/build/matter/.environment/cipd:/snap/snapcraft/x1/libexec/snapcraft:/snap/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`.

So, the following is prepended to `PATH` during bootstrapping: `/root/parts/lighting-hello/build/matter/third_party/pigweed/repo/out/host/host_tools:/root/parts/lighting-hello/build/matter/.environment/pigweed-venv/bin:/root/parts/lighting-hello/build/matter/.environment/cipd/packages/arm/bin:/root/parts/lighting-hello/build/matter/.environment/cipd/packages/arm:/root/parts/lighting-hello/build/matter/.environment/cipd/packages/zap:/root/parts/lighting-hello/build/matter/.environment/cipd/packages/pigweed/bin:/root/parts/lighting-hello/build/matter/.environment/cipd/packages/pigweed:/root/parts/lighting-hello/build/matter/.environment/cipd`.

Writing the whole PATH and letting the consumer export it could have side effects if the original PATH ends up containing other part-specific paths. Instead, you could prepend the Matter SDK additions to the PATH, e.g.: `'echo "export PATH=$MATTER_SDK_PATH:\$PATH" >> matter_sdk_env.sh'`

The content of the `matter_sdk_env.sh` should be as follows:
```bash
export PATH=/root/parts/lighting-hello/build/matter/third_party/pigweed/repo/out/host/host_tools:/root/parts/lighting-hello/build/matter/.environment/pigweed-venv/bin:/root/parts/lighting-hello/build/matter/.environment/cipd/packages/arm/bin:/root/parts/lighting-hello/build/matter/.environment/cipd/packages/arm:/root/parts/lighting-hello/build/matter/.environment/cipd/packages/zap:/root/parts/lighting-hello/build/matter/.environment/cipd/packages/pigweed/bin:/root/parts/lighting-hello/build/matter/.environment/cipd/packages/pigweed:/root/parts/lighting-hello/build/matter/.environment/cipd:$PATH
```

MonicaisHer: Thanks for the suggestion, `MATTER_SDK_PATH` has been added in commit 9f5c229f351d9d279e904d504b5f215ece1e8036.

snapcraft/parts/plugins/matter_sdk_plugin.py
```diff
@@ -158,11 +161,26 @@ def get_build_commands(self) -> List[str]:
...
+            [
+                'IFS=: read -r -a OLD_PATH_ARRAY <<< "$OLD_PATH"',
+                'IFS=: read -r -a NEW_PATH_ARRAY <<< "$PATH"',
+                "declare -a DIFFERENCE=()",
+                'for element in "${NEW_PATH_ARRAY[@]}"; do',
+                '    if [[ ! " ${OLD_PATH_ARRAY[@]} " =~ " $element " ]]; then',
+                '        DIFFERENCE+=("$element")',
+                "    fi",
+                "done",
+                'MATTER_SDK_PATH=$(IFS=:; echo "${DIFFERENCE[*]}")',
```
farshidtz, line 176: I find this very convoluted. How about using [shell parameter expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html):
```bash
MATTER_SDK_PATHS="${PATH%$OLD_PATH}"
```
and then `'echo "export PATH=$MATTER_SDK_PATHS\\$PATH" >> matter-sdk-env.sh',` ?

Note that `MATTER_SDK_PATHS` will include the separator.

snapcraft/parts/plugins/matter_sdk_plugin.py
```diff
@@ -165,23 +165,15 @@ def get_build_commands(self) -> List[str]:
...
+                'MATTER_SDK_PATHS="${PATH%$OLD_PATH}"',
             ]
         )
 
         # Prepend the Matter SDK related PATH to the beginning of the PATH environment variable
         # and save it to the matter-sdk-env.sh file
         commands.extend(
             [
-                'echo "export PATH=$MATTER_SDK_PATH:\\$PATH" >> matter-sdk-env.sh',
+                'echo "export PATH=$MATTER_SDK_PATHS:\\$PATH" >> matter-sdk-env.sh',
```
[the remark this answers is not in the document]
MonicaisHer: You're right. Thank you, I have removed it.

tests/spread/plugins/craft-parts/matter-sdk/snapcraft.yaml
```diff
@@ -0,0 +1,48 @@
...
+  matter-sdk:
+    plugin: matter-sdk
+    matter-sdk-version: "1536ca20c5917578ca40ce509400e97b52751788" # use this commit with ptpython version fix; needs to be updated once matter sdk have a stable release
+
+  lighting:
+    plugin: nil
+    after: [matter-sdk]
+    override-build: |
+      # Source the Matter SDK environment variables
+      source ../../matter-sdk/build/matter-sdk-env.sh
```
sergiusens, line 37: this is fragile and probably hard to discover, a reason to perhaps use an extension instead

tests/spread/plugins/craft-parts/matter-sdk/snapcraft.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+      # Build the lighting app for snapcraft spread testing purposes
+      cd ../../matter-sdk/build/examples/lighting-app/linux
+      gn gen out/build
+      ninja -C out/build
+      
+      ldd out/build/chip-lighting-app
+      
+      mkdir -p $CRAFT_PART_INSTALL/bin
+      cp out/build/chip-lighting-app $CRAFT_PART_INSTALL/bin/lighting-app
```
sergiusens, line 47: should the matter plugin deal with this? Are all project different? I we cannot have nice reproduceable logic for a plugin we might need to sit down and discuss this a bit more.

tests/spread/plugins/craft-parts/matter-sdk/task.yaml
```diff
@@ -0,0 +1,51 @@
...
+  #shellcheck source=tests/spread/tools/snapcraft-yaml.sh
+  . "$TOOLS_DIR/snapcraft-yaml.sh"
+  restore_yaml "snap/snapcraft.yaml"
+
+execute: |
+  cd "$SNAP"
+
+  # Build and install the snap
+  snapcraft
+  snap install "${SNAP}"*.snap --dangerous
```
mr-cal, line 27: shellcheck should be OK with this:

```suggestion
  snap install "${SNAP}*.snap" --dangerous
```

tests/unit/parts/plugins/test_matter_sdk_plugin.py
```diff
@@ -0,0 +1,146 @@
...
+
+def test_get_pull_commands(part_info):
+    properties = MatterSdkPlugin.properties_class.unmarshal(
+        {"matter-sdk-version": "master"}
+    )
+    plugin = MatterSdkPlugin(properties=properties, part_info=part_info)
+
+    sdk_version = properties.matter_sdk_version  # type: ignore
+
+    expected_commands = []
```
mr-cal, line 45: Segmenting the list in the plugin is fine since you have comments interspersed, but in the unit tests, I recommend just creating it outright:

```suggestion
    expected_commands = [
            "if [ ! -d matter ]; then",
            "    git init",
            f"   git remote add origin {MATTER_SDK_REPO}",
            f"   git fetch --depth 1 origin {sdk_version}",
            "    git checkout FETCH_HEAD",
            "fi",
            "scripts/checkout_submodules.py --shallow --platform linux",
    ]
```

tests/unit/parts/plugins/test_matter_sdk_plugin.py
```diff
@@ -0,0 +1,146 @@
...
+    assert plugin.get_build_environment() == {}
+
+
+def test_get_build_commands(part_info):
+    properties = MatterSdkPlugin.properties_class.unmarshal(
+        {"matter-sdk-version": "master"}
+    )
+    plugin = MatterSdkPlugin(properties=properties, part_info=part_info)
+
+    expected_commands = []
```
mr-cal, line 117: Same as above, you'd save ~20 lines of code and gain readability if you create the list all at once.

mr-cal (commented):
I think everything you have implemented looks really good.  The plugin, unit, and spread tests are all well done.  And nice work satisfying all the linters!

I echo Sergio's concern - a lot of code is required in an override-build script to build the example snap.  I don't know much about Matter, but I wonder if more of that can be hidden from the user via the plugin, an extension, or both.  I agree that we should have a meeting!

tests/spread/plugins/craft-parts/matter-sdk/snapcraft.yaml
MonicaisHer: For this particular command, we could use an extension instead. As for the Matter SDK plugin itself, the reason we chose a plugin over an extension is to provide the option for the user to specify the Matter SDK version.

tests/spread/plugins/craft-parts/matter-sdk/snapcraft.yaml
MonicaisHer: The reason here we enter the SDK directory `/matter-sdk/build` is to build an example application from it, as the SDK itself contains some example applications. 

For production, users will not build these example applications; they only need to run one command `source ../../matter-sdk/build/matter-sdk-env.sh` to set up the build environment, and then add their own logic.

snapcraft/parts/plugins/matter_sdk_plugin.py
```diff
@@ -0,0 +1,180 @@
...
+                "    git init",
+                f"   git remote add origin {MATTER_SDK_REPO}",
+                f"   git fetch --depth 1 origin {options.matter_sdk_version}",
+                "    git checkout FETCH_HEAD",
+                "fi",
+            ]
+        )
+
+        # Checkout submodules for Linux platform
+        commands.extend(["scripts/checkout_submodules.py --shallow --platform linux"])
```
syu-w, line 95: You may want to do this in a subshell if it is under `matter` that was changed above
```suggestion
        commands.extend(["(cd matter && ./scripts/checkout_submodules.py --shallow --platform linux)"])
```

snapcraft/parts/plugins/matter_sdk_plugin.py
```diff
@@ -0,0 +1,180 @@
...
+
+        # Clone Matter SDK repository
+        commands.extend(
+            [
+                "if [ ! -d matter ]; then",
+                "    git init",
+                f"   git remote add origin {MATTER_SDK_REPO}",
+                f"   git fetch --depth 1 origin {options.matter_sdk_version}",
+                "    git checkout FETCH_HEAD",
+                "fi",
```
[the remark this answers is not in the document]
MonicaisHer: Thanks for pointing that out! The matter directory check was leftover from previous changes. There are no scenarios in which the consumer snap can rerun this plugin with a cached matter directory. This is because when consumers override the part that uses this plugin, the plugin directory will be completely removed and rerun, without having a chance to use the existing matter directory. I will remove this `if [ ! -d matter ]` check as it is not necessary.

The reason for not creating a matter directory is that it will make the path less readable. For example, with the current approach, the consumer snap will source the environment variables like [this](https://github.com/canonical/snapcraft/blob/12e1e6ac3e92ddaed064e8d0f07b1894cde6215a/tests/spread/plugins/craft-parts/matter-sdk/snapcraft.yaml#L37):
```
source ../../matter-sdk/build/matter-sdk-env.sh
```

With the matter directory, the path would look like this:
```
source ../../matter-sdk/build/matter/matter-sdk-env.sh
```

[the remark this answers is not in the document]
MonicaisHer: As the SDK upstream currently has a specific [commit](https://github.com/canonical/snapcraft/blob/12e1e6ac3e92ddaed064e8d0f07b1894cde6215a/tests/spread/plugins/craft-parts/matter-sdk/snapcraft.yaml#L30) with a fix we need, the repo has been [shallow cloned with this commit at depth 1](https://stackoverflow.com/a/43136160/17842525). Once the SDK upstream has tagged this fix in a branch, it could simply be done using git clone and in one line as you suggested.

snapcraft/parts/plugins/matter_sdk_plugin.py
MonicaisHer: Explained as above

mr-cal (approved):
Looks great to me, thanks!

Since this plugin is experimental and there is not an established user base, I am OK with the design.

For posterity, the scope of work and design choices are covered in IENG-830.

syu-w (approved):
Tried it locally and it seems to work fine

tests/spread/plugins/craft-parts/matter-sdk/task.yaml
```diff
@@ -0,0 +1,51 @@
...
+    exit 1
+  fi
+
+  # Check if server initialization is complete for matter-lighting
+  if ! journalctl --since "$start_time" | grep matter-lighting | grep "CHIP:SVR: Server initialization complete"; then
+    echo "Error: matter-lighting initialization failed."
+    exit 1
+  fi
+  
+
```
sergiusens, line 51:
```suggestion
```

sergiusens: there are some extra lines here

sergiusens (approved):

mr-cal:
@MonicaisHer - Do you mind resolving the last file conflict?  After that, we can land this PR if you are ready

PR #4492, opened 2023-12-11
feat(projects): add components model

### LINKED ISSUE #4464: Parse and validate component keywords in `snapcraft.yaml`
### What needs to get done

The Project model needs to parse and validate a top-level `components` keyword containing a list of `Component` pydantic models.

The schema is:
```yaml
components:
  <name>:
    type: "test"
    summary: ...
    description: ...
    version: ...
```

Rules: 
- component names are unique
- component names adhere to the naming convention for namespaced partitions (lowercase alphabetical characters and hyphens, cannot start or end with a hyphen)
- component type is `test`
- summary is a string with the same naming convention as the top-level `summary` keyword
- description is a string (same as the top-level `description`)
- version is an optional field with the same naming convention as the top-level `version` keyword

Snapcraft also needs a new function in the Project module called `get_component_names() -> List[str]`.  It should return a list of component names.  Later, this will be used to create a list of partitions to pass to the `LifecycleManager`.

### Why it needs to get done

mr-cal:

-----

Adds a `components` model.  See #4464 for the naming conventions.

You can review per-commit.

Fixes #4464
(CRAFT-2308)

### DIFF
--- a/snapcraft/projects.py
+++ b/snapcraft/projects.py
@@ -187,6 +187,26 @@ def _validate_architectures_all_keyword(architectures):
             )
 
 
+def _validate_version_name(version: str, model_name: str) -> None:
+    """Validate a version complies to the naming convention.
+
+    :param version: version string to validate
+    :param model_name: name of the model that contains the version
+
+    :raises ValueError: if the version contains invalid characters
+    """
+    if version and not re.match(
+        r"^[a-zA-Z0-9](?:[a-zA-Z0-9:.+~-]*[a-zA-Z0-9+~])?$", version
+    ):
+        raise ValueError(
+            f"Invalid version '{version}': {model_name.title()} versions consist of "
+            "upper- and lower-case alphanumeric characters, as well as periods, colons, "
+            "plus signs, tildes, and hyphens. They cannot begin with a period, colon, "
+            "plus sign, tilde, or hyphen. They cannot end with a period, colon, or "
+            "hyphen"
+        )
+
+
 class Socket(ProjectModel):
     """Snapcraft app socket definition."""
 
@@ -562,15 +582,7 @@ class Project(ProjectModel):
         if not version and "adopt_info" not in values:
             raise ValueError("Version must be declared if not adopting metadata")
 
-        if version and not re.match(
-            r"^[a-zA-Z0-9](?:[a-zA-Z0-9:.+~-]*[a-zA-Z0-9+~])?$", version
-        ):
-            raise ValueError(
-                f"Invalid version '{version}': Snap versions consist of upper- and lower-case "
-                "alphanumeric characters, as well as periods, colons, plus signs, tildes, "
-                "and hyphens. They cannot begin with a period, colon, plus sign, tilde, or "
-                "hyphen. They cannot end with a period, colon, or hyphen"
-            )
+        _validate_version_name(version, "Snap")
 
         return version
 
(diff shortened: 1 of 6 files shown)


snapcraft/projects.py
```diff
@@ -415,6 +435,23 @@ def _validate_default_provider(cls, default_provider):
...
+
+    summary: ProjectSummary
+    description: str
+    type: Literal["test"]
+    version: Optional[ProjectVersion]
+
+    @pydantic.validator("version")
+    @classmethod
+    def _validate_version(cls, version):
+        if version:
```
lengau, line 449: Do we want to allow an empty string in `version`? If not,

```suggestion
        if version is not None:
```

tests/unit/test_projects.py
```diff
@@ -1773,3 +1774,191 @@ def test_project_get_build_for_arch_triplet_all(self, project_yaml_data):
...
+                "Invalid version '1:': Component versions consist of",
+            ),  # cannot end with colon
+            (
+                "1-",
+                "Invalid version '1-': Component versions consist of",
+            ),  # cannot end with hyphen
+            (
+                "123456789012345678901234567890123",
+                "ensure this value has at most 32 characters",
+            ),  # too large
```
lengau, line 1939: I don't mind if this is convention, but the pytest-encouraged way to do this would be:

```suggestion
            pytest.param(
                "123456789012345678901234567890123",
                "ensure this value has at most 32 characters",
                id="too large",
            ),
```

lengau (approved):
Looks good!

snapcraft/projects.py
```diff
@@ -187,6 +187,26 @@ def _validate_architectures_all_keyword(architectures):
...
+    """Validate a version complies to the naming convention.
+
+    :param version: version string to validate
+    :param model_name: name of the model that contains the version
+
+    :raises ValueError: if the version contains invalid characters
+    """
+    if version and not re.match(
+        r"^[a-zA-Z0-9](?:[a-zA-Z0-9:.+~-]*[a-zA-Z0-9+~])?$", version
+    ):
```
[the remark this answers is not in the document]
cmatsuoka: I was about to suggest a pattern field but was not sure if there's a mechanism for custom exception raising, couldn't find it in docs.

tests/unit/test_projects.py
cmatsuoka: Maybe something like `"x" * 33` could be more fail-proof than a literal string.

snapcraft/projects.py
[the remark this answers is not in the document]
cmatsuoka: And at some point we may want to have pre-compiled regexes in validators.

cmatsuoka (approved):

tests/unit/test_projects.py
mr-cal: Thanks, fixed! (both here and where I copied from)

snapcraft/projects.py
mr-cal: Good catch, I added a check and an extra test.

snapcraft/projects.py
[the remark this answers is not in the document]
mr-cal: Fortunately, we will be able to take advantage of `craft-applications` [implementation](https://github.com/canonical/craft-application/blob/7fc5a733a689b07e46f250c6cbb15347f2e1d4ec/craft_application/models/constraints.py#L70) and compiled regex later this cycle!

PR #4498, opened 2023-12-14
fix(remote): fallback to legacy snapcraft if under shallow git clone

syu-w:
Git shallow clone is not pushable by design. Fallback to legacy remote build which has workaround. This only support core20 and core22. Any other bases and devel will report error.

### DIFF
--- a/snapcraft/commands/remote.py
+++ b/snapcraft/commands/remote.py
@@ -30,7 +30,12 @@ from overrides import overrides
 from snapcraft.errors import MaintenanceBase, SnapcraftError
 from snapcraft.legacy_cli import run_legacy
 from snapcraft.parts import yaml_utils
-from snapcraft.remote import AcceptPublicUploadError, RemoteBuilder, is_repo
+from snapcraft.remote import (
+    AcceptPublicUploadError,
+    RemoteBuilder,
+    is_repo,
+    is_shallow_repo,
+)
 from snapcraft.utils import confirm_with_user, get_host_architecture, humanize_list
 
 _CONFIRMATION_PROMPT = (
@@ -193,7 +198,27 @@ class RemoteBuildCommand(BaseCommand):
             run_legacy()
             return
 
-        if is_repo(Path().absolute()):
+        current_path = Path().absolute()
+
+        if is_shallow_repo(current_path):
+            emit.debug("Current git repository is shallow cloned.")
+            base = self._get_effective_base()
+            if base in ["core20", "core22"]:
+                emit.progress(
+                    "Remote build for shallow clones is deprecated "
+                    "and will be removed in core24",
+                    permanent=True,
+                )
+                emit.progress("Fallback to legacy remote-build", permanent=True)
+                run_legacy()
+                return
+
+            raise SnapcraftError(
+                "remote-build for shallow clones is not supported "
+                f"for bases newer than core22, current base is {base}"
+            )
+
+        if is_repo(current_path):
             emit.debug(
                 "Running new remote-build because project is in a git repository"
             )
@@ -228,6 +253,11 @@ class RemoteBuildCommand(BaseCommand):
 
     def _run_new_remote_build(self) -> None:
         """Run new remote-build code."""
+        if is_shallow_repo(Path().absolute()):
+            raise SnapcraftError(
+                "remote-build for shallow clones is not supported "
+                "for bases newer than core22"
+            )
         emit.progress("Setting up launchpad environment")
         remote_builder = RemoteBuilder(
             app_name="snapcraft",
--- a/snapcraft/commands/remote.py
+++ b/snapcraft/commands/remote.py
@@ -32,9 +32,9 @@ from snapcraft.legacy_cli import run_legacy
 from snapcraft.parts import yaml_utils
 from snapcraft.remote import (
     AcceptPublicUploadError,
+    GitType,
     RemoteBuilder,
-    is_repo,
-    is_shallow_repo,
+    get_git_repo_type,
 )
 from snapcraft.utils import confirm_with_user, get_host_architecture, humanize_list
 
@@ -198,34 +198,25 @@ class RemoteBuildCommand(BaseCommand):
             run_legacy()
             return
 
-        current_path = Path().absolute()
+        git_type = get_git_repo_type(Path().absolute())
 
-        if is_shallow_repo(current_path):
-            emit.debug("Current git repository is shallow cloned.")
-            base = self._get_effective_base()
-            if base in ["core20", "core22"]:
-                emit.progress(
-                    "Remote build for shallow clones is deprecated "
-                    "and will be removed in core24",
-                    permanent=True,
-                )
-                emit.progress("Fallback to legacy remote-build", permanent=True)
-                run_legacy()
-                return
-
-            raise SnapcraftError(
-                "remote-build for shallow clones is not supported "
-                f"for bases newer than core22, current base is {base}"
-            )
-
-        if is_repo(current_path):
+        if git_type == GitType.NORMAL:
             emit.debug(
                 "Running new remote-build because project is in a git repository"
             )
             self._run_new_remote_build()
             return
 
-        emit.debug("Running fallback remote-build")
+        if git_type == GitType.SHALLOW:
+            emit.debug("Current git repository is shallow cloned.")
+            emit.progress(
+                "Remote build for shallow clones is deprecated "
+                "and will be removed in core24",
+                permanent=True,
+            )
+            # fall-through to legacy remote-build
+
+        emit.progress("Running fallback remote-build", permanent=True)
         run_legacy()
 
     def _get_project_name(self) -> str:
@@ -253,11 +244,6 @@ class RemoteBuildCommand(BaseCommand):
 
     def _run_new_remote_build(self) -> None:
         """Run new remote-build code."""
-        if is_shallow_repo(Path().absolute()):
-            raise SnapcraftError(
-                "remote-build for shallow clones is not supported "
-                "for bases newer than core22"
-            )
         emit.progress("Setting up launchpad environment")
         remote_builder = RemoteBuilder(
             app_name="snapcraft",
--- a/snapcraft/commands/remote.py
+++ b/snapcraft/commands/remote.py
@@ -216,7 +216,7 @@ class RemoteBuildCommand(BaseCommand):
             )
             # fall-through to legacy remote-build
 
-        emit.progress("Running fallback remote-build", permanent=True)
+        emit.debug("Running fallback remote-build")
         run_legacy()
 
     def _get_project_name(self) -> str:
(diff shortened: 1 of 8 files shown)


tests/unit/commands/test_remote.py
```diff
@@ -523,6 +526,101 @@ def test_run_in_repo_newer_than_core22(
...
+
+    # pygit2 does not support shallow cloning, so we use git directly
+    subprocess.run(
+        [
+            "git",
+            "clone",
+            "--depth",
+            "1",
+            "file://" + str(git_normal_path.absolute()),
+            str(git_shallow_path.absolute()),
```
lengau, line 563: It probably doesn't affect anything, but I think;

```suggestion
            git_normal_path.absolute().as_uri(),
            git_shallow_path.absolute().as_posix(),
```

looks nicer

tests/unit/commands/test_remote.py
```diff
@@ -523,6 +526,101 @@ def test_run_in_repo_newer_than_core22(
     emitter.assert_debug("Running new remote-build because base is newer than core22")
 
 
+@pytest.mark.parametrize(
```
lengau, line 529: The content of these tests looks fine, but they look more like integration tests than unit tests to me.

lengau (approved):

tests/unit/commands/test_remote.py
syu-w: True, since it is most test the external library. But I feel put it in spread is too heavy.

tests/unit/commands/test_remote.py
syu-w: Sure

tests/unit/commands/test_remote.py
```diff
@@ -523,6 +526,101 @@ def test_run_in_repo_newer_than_core22(
...
+    emitter.assert_debug("Current git repository is shallow cloned.")
+    emitter.assert_progress("Fallback to legacy remote-build", permanent=True)
+
+
+@pytest.mark.parametrize("create_snapcraft_yaml", {"devel"}, indirect=True)
+@pytest.mark.usefixtures(
+    "create_snapcraft_yaml", "mock_confirm", "mock_argv", "use_new_remote_build"
+)
+def test_run_in_shallow_repo_unsupported(emitter, new_dir, mock_remote_builder):
+    """core22 and older bases run new remote-build if in a git repo."""
```
mr-cal, line 581: Is this the right docstring?

tests/unit/commands/test_remote.py
```diff
@@ -523,6 +526,101 @@ def test_run_in_repo_newer_than_core22(
     emitter.assert_debug("Running new remote-build because base is newer than core22")
 
 
+@pytest.mark.parametrize(
+    "create_snapcraft_yaml", LEGACY_BASES | {"core22"}, indirect=True
+)
+@pytest.mark.usefixtures("create_snapcraft_yaml", "mock_confirm", "mock_argv")
+def test_run_in_shallow_repo(emitter, mock_run_legacy, new_dir):
+    """core22 and older bases run new remote-build if in a git repo."""
```
mr-cal, line 534: Is this docstring correct?

tests/unit/commands/test_remote.py
```diff
@@ -523,6 +526,101 @@ def test_run_in_repo_newer_than_core22(
...
+
+    os.chdir(git_shallow_path)
+    cli.run()
+
+    mock_run_legacy.assert_called_once()
+    emitter.assert_debug("Current git repository is shallow cloned.")
+    emitter.assert_progress("Fallback to legacy remote-build", permanent=True)
+
+
+@pytest.mark.parametrize("create_snapcraft_yaml", {"devel"}, indirect=True)
```
mr-cal, line 576: Why only `devel`?

tests/unit/commands/test_remote.py
```diff
@@ -523,6 +526,101 @@ def test_run_in_repo_newer_than_core22(
...
+        check=True,
+    )
+
+    os.chdir(git_shallow_path)
+
+    # no exception because run() catches it
+    ret = cli.run()
+    assert ret != 0
+
+    mock_remote_builder.assert_not_called()
```
mr-cal, line 621: What was supposed to happen?  Was an error raised?  If so, can you assert that the correct error was raised?

tests/unit/remote/test_git.py
```diff
@@ -39,6 +40,42 @@ def test_is_not_repo(new_dir):
     assert not is_repo(new_dir)
 
 
+def test_is_shallow_repo(new_dir):
```
mr-cal, line 43: Is there a test for calling this function on a repo that is not shallow?
What about on a path that is not a repo?

snapcraft/commands/remote.py
```diff
@@ -193,7 +198,27 @@ def _run_new_or_fallback_remote_build(self, base: str) -> None:
             run_legacy()
             return
 
-        if is_repo(Path().absolute()):
+        current_path = Path().absolute()
+
+        if is_shallow_repo(current_path):
+            emit.debug("Current git repository is shallow cloned.")
+            base = self._get_effective_base()
+            if base in ["core20", "core22"]:
```
mr-cal, line 206: there is already a check on line 178 for the base, so we shouldn't be checking it again with a nested if-statement

snapcraft/commands/remote.py
```diff
@@ -193,7 +198,27 @@ def _run_new_or_fallback_remote_build(self, base: str) -> None:
...
+                emit.progress(
+                    "Remote build for shallow clones is deprecated "
+                    "and will be removed in core24",
+                    permanent=True,
+                )
+                emit.progress("Fallback to legacy remote-build", permanent=True)
+                run_legacy()
+                return
+
+            raise SnapcraftError(
```
mr-cal, line 216: I'm not sure this exception is possible to reach and it's not clear if it's being tested in the tests.  Furthermore, it's a confusing duplicate of the error on line 256, which also shouldn't be part of the business logic.

snapcraft/commands/remote.py
```diff
@@ -193,7 +198,27 @@ def _run_new_or_fallback_remote_build(self, base: str) -> None:
             run_legacy()
             return
 
-        if is_repo(Path().absolute()):
+        current_path = Path().absolute()
+
+        if is_shallow_repo(current_path):
+            emit.debug("Current git repository is shallow cloned.")
+            base = self._get_effective_base()
```
mr-cal, line 205: `base` is passed as a parameter and shouldn't be reassigned

mr-cal (commented):
Good start but there are a few code smells.  The business logic in the flowchart had 4 checks and 2 outcomes (either running the new or legacy remote-build).

The code here is now out-of-order with the flowchart, introduces more checks, and adds more outcomes (raising exceptions within the business logic).

snapcraft/commands/remote.py
```diff
@@ -228,6 +253,11 @@ def _get_project_name(self) -> str:
 
     def _run_new_remote_build(self) -> None:
         """Run new remote-build code."""
+        if is_shallow_repo(Path().absolute()):
```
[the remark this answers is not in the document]
syu-w: After that check it will create `RemoteBuilder` immediately, which I don't think is a suitable place to do the check.

tests/unit/commands/test_remote.py
syu-w: You will get invalid base if you put `core24`, which it may need update in another PR

tests/unit/commands/test_remote.py
syu-w: The error was catches in `cli.run()`, that is not able to know which error actually raised. The only way to check that is capture stdout/stderr.

tests/unit/remote/test_git.py
syu-w: The function call `is_repo` first, unless you want a nearly duplicate tests to do it.

tests/unit/commands/test_remote.py
mr-cal: We can do this and add `core24` to `CURRENT_BASES` in another PR.
```suggestion
@pytest.mark.parametrize("create_snapcraft_yaml",  CURRENT_BASES - {"core22"}, indirect=True)
```

tests/unit/remote/test_git.py
mr-cal: Yes, I do want to see this explicitly tested. We should test all branches in a function when feasible.

tests/unit/commands/test_remote.py
syu-w: That's more future proof

tests/unit/remote/test_git.py
syu-w: OK, added

tests/unit/commands/test_remote.py
mr-cal: Yep, I think that's what the other tests do.

snapcraft/commands/remote.py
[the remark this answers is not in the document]
syu-w: Do you want to put it into exist functions that will use git (that results in no changes to the `command.py`), or a separate function and called here? (like `check_git()`?)

tests/unit/commands/test_remote.py
syu-w: Added

snapcraft/remote/git.py
```diff
@@ -50,6 +61,41 @@ def is_repo(path: Path) -> bool:
...
+
+    :raises SnapcraftError: if incompatible git repo is found
+    """
+    git_type = get_git_repo_type(path.absolute())
+
+    if git_type == GitType.INVALID:
+        raise SnapcraftError(f"Could not find a git repository in {str(path)!r}")
+
+    if git_type == GitType.SHALLOW:
+        raise SnapcraftError(
```
mr-cal, line 93: This should be a `RemoteBuild` error and not mention anything specific to snapcraft or its bases (since this module will get migrated out of snapcraft into a library and used by other applications)

snapcraft/commands/remote.py
```diff
@@ -193,14 +198,25 @@ def _run_new_or_fallback_remote_build(self, base: str) -> None:
...
+        if git_type == GitType.SHALLOW:
+            emit.debug("Current git repository is shallow cloned.")
+            emit.progress(
+                "Remote build for shallow clones is deprecated "
+                "and will be removed in core24",
+                permanent=True,
+            )
+            # fall-through to legacy remote-build
+
+        emit.progress("Running fallback remote-build", permanent=True)
```
mr-cal, line 219: To align with the 4 other `emit.debug("Running [sic] remote-build because...` calls above, which we decided to keep out of brief mode.

```suggestion
        emit.debug("Running fallback remote-build")
```

I think the deprecation notice is fine.

mr-cal (changes requested):
Looks good!  Just a minor comment about an emit and an error class.

snapcraft/commands/remote.py
syu-w: OK, but I really think we should notice user they are fallback to legacy mode since the output is significant difference

lengau (approved):

mr-cal (approved):
:+1:

PR #4502, opened 2024-01-03
Colcon plugin: use release build by default

MirkoFerrati:
But I got a strange error even if everything seems ok or 10/10
But I failed one test with the same issue as reported here https://github.com/snapcore/snapcraft/pull/3810
-----

Normally Colcon does not set any build flags and leaves the choice to the user, meaning the default value is empty and uses the default cmake configuration for debug symbols and optimization.
Inside snaps the executables are not supposed to be debugged and should be as optimized as possible so we want the default to be "Release" explicitly, unless the user overrides it intentionally in their snapcraft.yaml
This PR has been tested by creating snaps of the simple ros2-talker-listener snap with different debug levels, either default or user-customized.
The results for the talker executable show a clear reduction in size and in the number of symbols.


snapcraft/parts/plugins/colcon_plugin.py
```diff
@@ -197,6 +197,12 @@ def _get_build_commands(self) -> List[str]:
         if options.colcon_packages:
             build_command.extend(["--packages-select", *options.colcon_packages])
 
+        # compile in release only if user did not set cmake_args or cmake build type in cmake_args
+        if not options.colcon_cmake_args or not any(
+            "-DCMAKE_BUILD_TYPE=" in s for s in options.colcon_cmake_args
+        ):
```
Guillaumebeuzeboc, line 203: Isn't `any` going to return false anyway when options.colcon_cmake_args is empty? Do you really need the first part of the `if`

snapcraft/parts/plugins/colcon_plugin.py
```diff
@@ -197,6 +197,12 @@ def _get_build_commands(self) -> List[str]:
...
+            build_command.extend(["--cmake-args", "-DCMAKE_BUILD_TYPE=Release"])
```
Guillaumebeuzeboc, line 204: Does it work to use two time `--cmake-args` in one `colcon` call?

tests/legacy/unit/plugins/v2/test_colcon.py
```diff
@@ -182,6 +182,7 @@ class Options:
         "colcon build "
         '--base-paths "${SNAPCRAFT_PART_SRC_WORK}" --build-base "${SNAPCRAFT_PART_BUILD}" '
         '--merge-install --install-base "${SNAPCRAFT_PART_INSTALL}"/opt/ros/snap '
+        '--cmake-args -DCMAKE_BUILD_TYPE=Release '
```
Guillaumebeuzeboc, line 185: I might be missing something but here the legacy test is changed while the [legacy plugin](https://github.com/snapcore/snapcraft/blob/main/snapcraft_legacy/plugins/v2/colcon.py) is not.

snapcraft/parts/plugins/colcon_plugin.py
artivis: The first part of the if may make sure that the variable isn't `None`, altho I'm not sure whether it can take this value.

MirkoFerrati (commented):
I will push changes right now

snapcraft/parts/plugins/colcon_plugin.py
```diff
@@ -197,7 +197,16 @@ def _get_build_commands(self) -> List[str]:
...
+        # compile in release only if user did not set the build type in cmake-args
+        if not any("-DCMAKE_BUILD_TYPE=" in s for s in options.colcon_cmake_args):
+            build_command.extend(
+                [
+                    "--cmake-args",
+                    "-DCMAKE_BUILD_TYPE=Release",
+                    *options.colcon_cmake_args,
+                ]
+            )
+        elif len(options.colcon_cmake_args) > 0:
```
MirkoFerrati, line 209: works the same with else and no condition, but I wanted to be explicit here

tests/legacy/unit/plugins/v2/test_colcon.py
```diff
@@ -310,7 +311,271 @@ class Options:
...
+        "env -i LANG=C.UTF-8 LC_ALL=C.UTF-8 PATH=/bin:/test SNAP=TESTSNAP "
+        "SNAP_ARCH=TESTARCH SNAP_NAME=TESTSNAPNAME SNAP_VERSION=TESTV1 "
+        "http_proxy=http://foo https_proxy=https://bar "
+        "/test/python3 -I /test/_ros.py "
+        'stage-runtime-dependencies --part-src "${SNAPCRAFT_PART_SRC_WORK}" --part-install "${SNAPCRAFT_PART_INSTALL}" '
+        '--ros-version "${ROS_VERSION}" --ros-distro "${ROS_DISTRO}" --target-arch "${SNAPCRAFT_TARGET_ARCH}"',
+    ]
+
+
+def test_get_build_commands_with_cmake_release(monkeypatch):
```
artivis, line 462: Isn't this test essentially the same as `test_get_build_commands_with_cmake_debug`?

tests/unit/parts/plugins/test_colcon.py
```diff
@@ -352,7 +353,287 @@ def test_get_build_commands_with_all_properties(
...
+            "http_proxy=http://foo https_proxy=https://bar "
+            "/test/python3 -I /test/_ros.py "
+            'stage-runtime-dependencies --part-src "${CRAFT_PART_SRC_WORK}" '
+            '--part-install "${CRAFT_PART_INSTALL}" '
+            '--ros-version "${ROS_VERSION}" --ros-distro "${ROS_DISTRO}" '
+            '--target-arch "${CRAFT_TARGET_ARCH}" '
+            f"--stage-cache-dir {new_dir} --base core22",
+        ]
+
+    def test_get_build_commands_with_cmake_release(
```
artivis, line 514: Isn't this test essentially the same as `test_get_build_commands_with_cmake_debug`?

tests/legacy/unit/plugins/v2/test_colcon.py
MirkoFerrati: removed

tests/unit/parts/plugins/test_colcon.py
MirkoFerrati: removed

Guillaumebeuzeboc (approved):

artivis (approved):

artivis:
@mr-cal Can we get an approval for the CI?

mr-cal (approved):

cmatsuoka (approved):
Thanks!

cmatsuoka:
Should this be also backported to 7.x?

MirkoFerrati:
@cmatsuoka no need to backport

PR #4503, opened 2024-01-03
feat(meta): generate metadata for components

### LINKED ISSUE #4466: Generate metadata for components
### What needs to get done

As a post-prime step, snapcraft should generate a `meta/component.yaml` metadata file for each component.

`component.yaml` files should contain the following, collected from the definition in `snapcraft.yaml`:
```yaml
component: <snap name>+<component name>
type: test
summary: ...
description: ...
version: ... # optional
```

### Why it needs to get done

mr-cal:

-----

Generate metadata for components and write the metadata to `prime/component/<component-name>/meta/component.yaml`.

Fixes #4466 
(CRAFT-2312)

### DIFF
--- a/snapcraft/projects.py
+++ b/snapcraft/projects.py
@@ -822,12 +822,12 @@ class Project(ProjectModel):
 
         return None
 
-    def get_component_names(self) -> Optional[List[str]]:
+    def get_component_names(self) -> List[str]:
         """Get a list of component names.
 
-        :returns: A list of component names or None if no components are defined.
+        :returns: A list of component names.
         """
-        return list(self.components.keys()) if self.components else None
+        return list(self.components.keys()) if self.components else []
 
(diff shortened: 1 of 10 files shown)


snapcraft/projects.py
```diff
@@ -822,6 +822,13 @@ def get_build_for_arch_triplet(self) -> Optional[str]:
 
         return None
 
+    def get_component_names(self) -> Optional[List[str]]:
```
lengau, line 825: Any reason for this rather than just returning an empty `List[str]` if there are no components?

mr-cal: I think that's a better idea, let me implement it that way.

lengau (approved):
I'm happy with this, but I'm also very nosy so I want to know about any other comments/changes if they come down the pipeline please :-)

syu-w (approved):

PR #4506, opened 2024-01-05
feat: pack components

### LINKED ISSUE #4467: Package snap components
### What needs to get done

Each component’s prime directory and its metadata file should be packaged with `snap pack`.

The packaged components should be placed alongside the snap package.

The final message of snapcraft should include the packaged components.  For example:
```
Snapped hello_world_1.0_amd64.snap and components `hello-world+foo_1.0.comp`, `hello-world+bar_1.0.comp`, and `hello-world+baz_1.0.comp`.
```

### Why it needs to get done

mr-cal:

-----

Pack components with `snap pack`.

Snaps can have custom filenames using `--output`  (i.e. `snapcraft pack --output my-dir/my-snap.snap`)

However, components cannot.  Their names are determined by `snap pack`.  They will still be placed in the same output directory (`my-dir` in the example above).

Fixes #4467
(CRAFT-2313)

### DIFF
--- a/snapcraft/pack.py
+++ b/snapcraft/pack.py
@@ -90,30 +90,44 @@ def _get_filename(
     return None
 
 
-def _pack(command: List[Union[str, Path]]) -> str:
+def _pack(
+    directory: Path,
+    output_dir: Path,
+    output_file: Optional[str],
+    compression: Optional[str],
+) -> str:
     """Pack a directory with `snap pack` as a snap or component.
 
-    :param command: `snap pack` command to execute
+    :param directory: Directory to pack.
+    :param output_dir: Directory to output the artefact to.
+    :param output_file: Name of the artefact.
+    :param compression: Compression type to use, None for default.
 
     :returns: The filename of the packed snap or component.
 
     :raises SnapcraftError: If the directory cannot be packed.
     """
-    emit.debug(f"Pack command: {command}")
+    command: List[Union[str, Path]] = ["snap", "pack"]
 
+    if output_file:
+        command.extend(["--filename", output_file])
+    if compression:
+        command.extend(["--compression", compression])
+    command.extend([directory, output_dir])
+
+    emit.debug(f"Pack command: {command}")
     try:
         proc = subprocess.run(
             command, capture_output=True, check=True, universal_newlines=True
         )
     except subprocess.CalledProcessError as err:
-        msg = f"{err!s}"
+        msg = str(err)
         details = None
         if err.stderr:
             details = err.stderr.strip()
         raise errors.SnapcraftError(msg, details=details) from err
 
-    filename = Path(str(proc.stdout).partition(":")[2].strip()).name
-    return filename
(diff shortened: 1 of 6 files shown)


snapcraft/pack.py
```diff
@@ -90,6 +90,64 @@ def _get_filename(
...
+    :raises SnapcraftError: If the directory cannot be packed.
+    """
+    emit.debug(f"Pack command: {command}")
+
+    try:
+        proc = subprocess.run(
+            command, capture_output=True, check=True, universal_newlines=True
+        )
+    except subprocess.CalledProcessError as err:
+        msg = f"{err!s}"
```
cmatsuoka, line 109: This could be str(err) (but I see that other parts of the code are using this pattern).

snapcraft/pack.py
```diff
@@ -90,6 +90,64 @@ def _get_filename(
...
+        details = None
+        if err.stderr:
+            details = err.stderr.strip()
+        raise errors.SnapcraftError(msg, details=details) from err
+
+    filename = Path(str(proc.stdout).partition(":")[2].strip()).name
```
cmatsuoka, line 115: The method name "partition" confused me for a moment :smile:

cmatsuoka (approved):
Looks nice, can you add a spread test for component creation?

snapcraft/pack.py
```diff
@@ -90,6 +90,64 @@ def _get_filename(
     return None
 
 
+def _pack(command: List[Union[str, Path]]) -> str:
```
lengau, line 93: As it is, this feels like a pretty generic command wrapper, with the logic for creating the `pack` command still in `pack_components` or `pack_snap`. What about making `_pack` a bit smarter like:

```suggestion
def _pack(directory: Path, output_dir: Path, *, compression: Optional[str] = None, filename: Optional[str] = None) -> str:
```

snapcraft/parts/lifecycle.py
```diff
@@ -284,6 +284,47 @@ def _run_lifecycle_and_pack(  # noqa PLR0913
...
         emit.progress(f"Created snap package {snap_filename}", permanent=True)
 
+        if project.components:
+            _pack_components(lifecycle, project, parsed_args)
+
+
+def _pack_components(
+    lifecycle: PartsLifecycle,
+    project: Project,
+    parsed_args: "argparse.Namespace",
```
lengau, line 294:
```suggestion
    output: Optional[str],
```

snapcraft/pack.py
mr-cal: Great idea!

snapcraft/pack.py
mr-cal: Good point.  This was previously existing code, but I made this change.

mr-cal:
pyright failure is unrelated.  I don't think pyright is pinned in this branch.

mr-cal:
Pyright failure fixed in https://github.com/snapcore/snapcraft/pull/4512

lengau (approved):
Fantastic!

Issue #4508, opened 2024-01-05
`build-attributes: [enable-patchelf]` uses incorrect dynamic linker when setting binary ELF interpreter on core24


NucciTheBoss:
### Bug Description

I am using core24 as the base snap for my classic SLURM snap because core24 has the necessary packages that I need for enabling AMD GPU support in the workload scheduler: https://packages.ubuntu.com/search?keywords=librocm&searchon=names

I need to patch the binaries and shared libraries in the SLURM snap since it is classically confined, however, `build-attributes: [enable-patchelf]` is using an incorrect dynamic linker as the ELF interpreter for binaries. This is causing core dumps and segfaults if I try to execute the binaries on a host with an older libc implementation such as Ubuntu 22.04 or 20.04 LTS. The rpath is being set correctly though by the automatic patching mechanism. Also, for some reason, on core24, but not core22, the classic linter is warning me about a staged libc package. Several of SLURMs dependencies require libc6 >= 2.38, but it should be using the libc6 provided by the core24 base.

[snapcraft-20240105-110231.695714.log](https://github.com/snapcore/snapcraft/files/13845290/snapcraft-20240105-110231.695714.log)

### To Reproduce

1. _Download attached snapcraft.yaml file_
2. `snapcraft -v pack  #=> ensure that you have snapcraft 8.0.1 installed`
3. `lxc launch ubuntu:22.04 snap-tester --vm`
4. `lxc file push slurm_23.11.1_amd64.snap slurm-0/root/slurm_23.11.1_amd64.snap`
5. `lxc shell snap-tester`
6. `snap install core24 --edge`
7. `snap install slurm_23.11.1_amd64.snap --dangerous --classic`
8. `snap logs slurm.munged  #=> See core dump error`
9. `slurm.squeue -h  #=> See core dump error`

### Environment

OS: Ubuntu 23.10 Mantic Minotaur
LXD: 5.19 rev 20600
Snap test environment: 22.04 and 20.04

### snapcraft.yaml

```shell
# Copyright 2023 Canonical Ltd.
# See LICENSE file for licensing details

name: slurm
version: 23.11.1
summary: "Slurm: A Highly Scalable Workload Manager"
description: |
  Slurm is an open source, fault-tolerant, and highly scalable cluster management and
  job scheduling system for large and small Linux clusters.

  Contains Slurm services slurmctld, slurmd, slurmdbd, and slurmrestd. See `slurm-clients`
  snap for Slurm CLI commands.
[...]

PR #4523, opened 2024-01-15
fix(lifecycle): return correct base when using devel for build-base

syu-w:
`get_effective_base()` should not return "devel" and pass it to the `PartsLifecycle()`, which make the parts think the coreXX is "devel". Thus checking the wrong path for dynamic linking.

`LD_LIBRARY_PATH=...:/snap/core24/current/lib:/snap/core24/current/usr/lib:...`
become
`LD_LIBRARY_PATH=...:/snap/devel/current/lib:/snap/devel/current/usr/lib:...`
which not exists.

Fix: #4508

### DIFF
--- a/snapcraft/parts/lifecycle.py
+++ b/snapcraft/parts/lifecycle.py
@@ -432,7 +432,7 @@ def _run_in_provider(  # noqa PLR0915
     snapcraft_base = project.get_effective_base()
     build_base = providers.SNAPCRAFT_BASE_TO_PROVIDER_BASE[snapcraft_base]
 
-    if snapcraft_base == "devel":
+    if snapcraft_base in ("devel", "core24"):
         emit.progress(
             "Running snapcraft with a devel instance is for testing purposes only.",
             permanent=True,
--- a/snapcraft/projects.py
+++ b/snapcraft/projects.py
@@ -591,6 +591,15 @@ class Project(ProjectModel):
             raise ValueError("grade must be 'devel' when build-base is 'devel'")
         return values
 
+    @pydantic.validator("base", always=True)
+    @classmethod
+    def _validate_base(cls, base, values):
+        """Not allowed to use unsable base without devel build-base."""
+        if values.get("base") == "core24" and values.get("build_base") != "devel":
+            raise ValueError("build-base must be 'devel' when base is 'core24'")
+
+        return base
+
     @pydantic.validator("build_base", always=True)
     @classmethod
     def _validate_build_base(cls, build_base, values):
--- a/snapcraft/projects.py
+++ b/snapcraft/projects.py
@@ -594,7 +594,7 @@ class Project(ProjectModel):
     @pydantic.validator("base", always=True)
     @classmethod
     def _validate_base(cls, base, values):
-        """Not allowed to use unsable base without devel build-base."""
+        """Not allowed to use unstable base without devel build-base."""
         if values.get("base") == "core24" and values.get("build_base") != "devel":
             raise ValueError("build-base must be 'devel' when base is 'core24'")
 
--- a/snapcraft/providers.py
+++ b/snapcraft/providers.py
@@ -42,6 +42,7 @@ SNAPCRAFT_BASE_TO_PROVIDER_BASE = {
     "core18": bases.BuilddBaseAlias.BIONIC,
     "core20": bases.BuilddBaseAlias.FOCAL,
     "core22": bases.BuilddBaseAlias.JAMMY,
+    "core24": bases.BuilddBaseAlias.DEVEL,
     "devel": bases.BuilddBaseAlias.DEVEL,
 }
 
--- a/snapcraft/parts/yaml_utils.py
+++ b/snapcraft/parts/yaml_utils.py
@@ -33,7 +33,7 @@ _CORE_PART_KEYS = ["build-packages", "build-snaps"]
 _CORE_PART_NAME = "snapcraft/core"
 
 # All bases recognized by snapcraft
-BASES = {"core", "core18", "core20", "core22", "devel"}
+BASES = {"core", "core18", "core20", "core22", "core24", "devel"}
 # Bases no longer supported by the current version of snapcraft
 ESM_BASES = {"core", "core18"}
 # Bases handled by the legacy snapcraft codebase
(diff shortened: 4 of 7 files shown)


snapcraft/parts/lifecycle.py
```diff
@@ -164,7 +164,7 @@ def _run_command(  # noqa PLR0913 # pylint: disable=too-many-branches, too-many-
         project.parts,
         work_dir=work_dir,
         assets_dir=assets_dir,
-        base=project.get_effective_base(),
+        base=project.get_effective_base(allow_return_devel=project.type == "base"),
```
mr-cal, line 167: If I understand this correctly, `craft-parts` needs to know the base is `devel` only when building the devel base snap.  Why is that?

tests/unit/parts/test_lifecycle.py
```diff
@@ -1860,3 +1860,39 @@ def test_lifecycle_write_metadata(
...
+    mocker,
+    snapcraft_yaml,
+    tmp_path,
+    base,
+    build_base,
+    grade,
+    allow_return_devel,
+    expected_base,
+):
+    """Get the effective base from the project."""
```
mr-cal, line 1891: Should this be moved to `tests/unit/test_projects.py`?

mr-cal (commented):
Nice work finding the problem.  Just a question about craft-parts.

snapcraft/parts/lifecycle.py
syu-w: `base` snap does not has `base` defined, which created an issue for the checking later.  It actually can be anything but None.

tests/unit/parts/test_lifecycle.py
syu-w: Yeah, this was not the case in the initial test, so I put it here. It can indeed be moved into `test_projects.py`

snapcraft/parts/lifecycle.py
mr-cal: Sounds good, do you mind adding a comment about that in the code?

syu-w: Sure

mr-cal (approved):
:+1:

snapcraft/parts/lifecycle.py
```diff
@@ -160,11 +160,13 @@ def _run_command(  # noqa PLR0913 # pylint: disable=too-many-branches, too-many-
...
     track_stage_packages = getattr(parsed_args, "enable_manifest", False)
 
+    # notice the "base" snap does not have a base, so we need to use the
+    # build-base instead, which is allowed to be "devel"
     lifecycle = PartsLifecycle(
         project.parts,
         work_dir=work_dir,
         assets_dir=assets_dir,
-        base=project.get_effective_base(),
+        base=project.get_effective_base(allow_return_devel=project.type == "base"),
```
sergiusens, line 169: I would remove this signature change and deal with the variance from the type in get_effective_base inside the method as the object is itself project and can access its type

snapcraft/projects.py
```diff
@@ -713,6 +713,17 @@ def get_effective_base(self) -> str:
...
+        # the actual base is the same as the base defined in the project.
+        if base == "devel":
+            if allow_return_devel:
+                return base
+
+            if self.base is None:
+                raise RuntimeError("cannot determine base")
+
+            return self.base
+
```
sergiusens, line 726: why not encode this logic in get_effective_base at https://github.com/snapcore/snapcraft/blob/d1fc926eb73d9483c64b75a9da99e49fe1530a03/snapcraft/utils.py#L210 ?

sergiusens (changes requested):
I don't understand why we are loosing the abstraction, it seems like the but is in `get_effective_base`

sergiusens:
Can you also target the hotfix/8.0 branch?

snapcraft/parts/lifecycle.py
tigarmo: I agree, it's best to hide this logic inside `Project.get_effective_base()` and `utils.get_effective_base()`

syu-w: I thought about it at first, but I don't know why it was split into 2 parts in different places. Will move it to utils

tigarmo (approved):

mr-cal (commented):
So if I understand the [failing spread test](https://github.com/snapcore/snapcraft/actions/runs/7628784472/job/20780881404?pr=4523#step:4:240) correctly, for a project with the following definition,
```yaml
base: core24
build-base: devel
```
we want to pass the base of `core24` to craft-parts and the base of `devel` to craft-providers.  The problem is that `get_effective_base()` is not a one-size-fits-all solution for this scenario.


What do you think of adding `"core24": bases.BuilddBaseAlias.DEVEL` to [this mapping](https://github.com/snapcore/snapcraft/blob/d01e1cd665bfe8dd20ad2f6080ac9676e136b3e9/snapcraft/providers.py#L41)?  This would require us add more validation logic [here](https://github.com/snapcore/snapcraft/blob/d01e1cd665bfe8dd20ad2f6080ac9676e136b3e9/snapcraft/projects.py#L588) so that
```yaml
base: core24
```
still does not work.  (and would give us an opportunity to raise a better error than the current KeyError that gets raised)

mr-cal:
Ack, I'm sorry I linked to the wrong validator and misled you.

What I meant is that we need a validator that ensures this is still not allowed:
```yaml
base: core24
```

I think you can add a new validator along the lines of `if base == "core24" and not "build-base"`?

Once the `core24` snap is available on `beta` channel, we can remove this validator.

mr-cal (approved):
:+1:

snapcraft/providers.py
```diff
@@ -42,6 +42,7 @@
     "core18": bases.BuilddBaseAlias.BIONIC,
     "core20": bases.BuilddBaseAlias.FOCAL,
     "core22": bases.BuilddBaseAlias.JAMMY,
+    "core24": bases.BuilddBaseAlias.DEVEL,
     "devel": bases.BuilddBaseAlias.DEVEL,
```
sergiusens, line 46: @mr-cal we might need to clean this up, devel is only a build-base but never a base

mr-cal: https://github.com/snapcore/snapcraft/issues/4539

PR #4524, opened 2024-01-16
test(lifecycle): add test for check patchelf when override

syu-w:
`patchelf` will not be executed when
```
override-prime: |
  craftctl default
```

The reason is snapcraft patchelf using the craft-parts states to get files list and run patchelf. However when override, the list is empty due the the parts code skipped. So the patchelf will not run on any files.


tests/spread/general/classic-patchelf-override-multi-parts/snap/snapcraft.yaml
```diff
@@ -0,0 +1,36 @@
...
+  hello2:
+    source: ../hello2
+    plugin: make
+    build-attributes:
+      - enable-patchelf
+    build-packages:
+      - gcc
+      - libc-dev
+      - libcurl4-openssl-dev
+    stage-packages: [libcurl4]
```
lengau, line 36: Based on this success & failure, what happens with multiple parts that all override prime?

tests/spread/general/classic-patchelf-override/task.yaml
```diff
@@ -0,0 +1,44 @@
...
+  cd snap
+  cat snapcraft.yaml
+  snapcraft prime
+
+  arch_triplet="$(dpkg-architecture -q DEB_HOST_MULTIARCH)"
+
+  RPATH_ORIGIN_MATCH="^\\\$ORIGIN/../usr/lib/$arch_triplet:/snap/$base/current/lib/$arch_triplet"
+
+  # Verify typical binary.
+  patchelf --print-interpreter prime/bin/hello | MATCH "^/snap/$base/current/lib.*ld.*.so.*"
```
mr-cal, line 43: Shouldn't this check that patchelf did not run (by matching against `"^/lib.*/ld.*.so.*"`)?

mattculler (approved):

PR #4540, opened 2024-01-24
fix(lifecycle): return correct base when using devel for build-base (…

syu-w:


mr-cal (approved):

snapcraft/projects.py
```diff
@@ -591,6 +591,15 @@ def _validate_grade_and_build_base(cls, values):
             raise ValueError("grade must be 'devel' when build-base is 'devel'")
         return values
 
+    @pydantic.validator("base", always=True)
+    @classmethod
+    def _validate_base(cls, base, values):
+        """Not allowed to use unstable base without devel build-base."""
+        if values.get("base") == "core24" and values.get("build_base") != "devel":
```
lengau, line 598:
```suggestion
        if base == "core24" and values.get("build_base") != "devel":
```

Out of scope for this PR, but perhaps a future PR on main should move the `*_BASES` variables from parts/yaml_utils.py into a constants file and add `DEVEL_BASES`? Just thinking that would turn devel base checks from code changes to constants changes.

lengau (approved):
LGTM!

PR #4543, opened 2024-01-24
fix(remote): parse comma-separated architectures

### LINKED ISSUE #4516: regression: remote-build doesn't accept comma-seperated architectures
### Bug Description

In snapcraft 8.0.1, `snapcraft remote-build --architectures=amd64,arm64` fails to split a list of architectures.

[This logic](https://github.com/snapcore/snapcraft/blob/ffbdabf53de21606d9d12e15f16d4bca5e4423a5/snapcraft_legacy/cli/remote.py#L267) needs to be added to the remote-build cli handler.

### To Reproduce

`snapcraft remote-build --architectures=amd64,arm64`

### Environment

n/a

### snapcraft.yaml

```shell
n/a
```

### Relevant log output

```shell
2024-01-10 10:21:08.174 remote-build error: Architecture not supported by the remote builder.
The following architectures are not supported by the remote builder: ['amd64,arm64'].
Please remove them from the architecture list and try again.
```

### Additional context

_No response_

mr-cal:

-----

2 commits:

- fix(remote): make `--build-for` and `--build-on` mutually exclusive
    - This was not working as intended (and was not tested)
    - This is only for the new remote-build and makes no changes to how legacy remote build handles these arguments
- fix(remote): parse comma-separated architectures
    - This uses the same style that legacy remote-build uses for passing architectures - a single comma-separated set

Fixes #4516
(CRAFT-2386)

### DIFF
--- a/snapcraft/commands/remote.py
+++ b/snapcraft/commands/remote.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright 2022-2023 Canonical Ltd.
+# Copyright 2022-2024 Canonical Ltd.
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License version 3 as
@@ -91,14 +91,13 @@ class RemoteBuildCommand(BaseCommand):
         parser.add_argument(
             "--status", action="store_true", help="display remote build status"
         )
-        parser_target = parser.add_mutually_exclusive_group()
-        parser_target.add_argument(
+        parser.add_argument(
             "--build-on",
             metavar="arch",
             nargs="+",
             help=HIDDEN,
         )
-        parser_target.add_argument(
+        parser.add_argument(
             "--build-for",
             metavar="arch",
             nargs="+",
@@ -141,7 +140,6 @@ class RemoteBuildCommand(BaseCommand):
 
         if parsed_args.build_on:
             emit.message("Use --build-for instead of --build-on")
-            parsed_args.build_for = parsed_args.build_on
 
         if not parsed_args.launchpad_accept_public_upload and not confirm_with_user(
             _CONFIRMATION_PROMPT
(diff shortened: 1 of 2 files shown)


snapcraft/commands/remote.py
```diff
@@ -93,15 +93,15 @@ def fill_parser(self, parser: argparse.ArgumentParser) -> None:
...
         parser.add_argument(
             "--build-on",
+            type=lambda arg: [arch.strip() for arch in arg.split(",")],
             metavar="arch",
-            nargs="+",
             help=HIDDEN,
         )
         parser.add_argument(
             "--build-for",
+            type=lambda arg: [arch.strip() for arch in arg.split(",")],
```
tigarmo, line 102: Does this support something wacky like `--build-for=amd64 --build-for=arm64,armhf`?

mr-cal: Good question.  When there are duplicates, argparse/craft-cli uses the last `--build-{on|for}`, silently ignores the other `--build-{on|for}`s, and always provides a single string for this lambda.

Normally I wouldn't add a test for standard behavior, but since this is a delicate situation, I added an extra unit test.

tigarmo: to be clear, this wouldn't be a change in behavior right? The previous code would also only use the last --build-x?

mr-cal: Correct, this behavior is unchanged.

The only "loss" is that `snapcraft remote-build --build-for amd64 arm64` doesn't work.  

And I think that "loss" is OK because it is a regression in the new remote builder: snapcraft 7 only supported comma-separated architectures, not space-separated architectures.

tigarmo (approved):
thankss

lengau (approved):

mr-cal:
Merging because spread test failures are completely unrelated (appears to be an issue with the devel image to be resolved via https://github.com/canonical/craft-providers/pull/501

PR #4553, opened 2024-01-31
fix: warning unsafe override-prime

syu-w:


tests/unit/parts/test_lifecycle.py
```diff
@@ -1811,6 +1811,82 @@ def test_patch_elf(snapcraft_yaml, mocker, new_dir):
...
+        "Warning: 'enable-patchelf' feature is unsafe if 'override-prime' "
+        "is also enabled. It's not possible to track file changes in the "
+        "prime directory.",
+        permanent=True,
+    )
+
+    # TODO: This is not working as expected, the patchelf is not being called
+    # See https://github.com/canonical/craft-parts/pull/623
+    #
+    # assert run_patchelf_mock.mock_calls == [
```
mr-cal, line 1879: What do you think of something like `run_patchelf_mock.assert_not_called()`?

snapcraft/parts/lifecycle.py
```diff
@@ -186,6 +186,20 @@ def _run_command(  # noqa PLR0913 # pylint: disable=too-many-branches, too-many-
...
 
+    if any(
+        "enable-patchelf" in _part.get("build-attributes", [])
+        for _part in getattr(project, "parts", {}).values()
+    ) and any(
+        _part.get("override-prime", None)
+        for _part in getattr(project, "parts", {}).values()
+    ):
+        emit.progress(
+            "Warning: 'enable-patchelf' feature is unsafe if 'override-prime' "
```
mr-cal, line 197: Just so I am clear, is the following statement correct?

In a project with multiple parts:
- `patchelf` will not run on files primed by parts with `override-prime` defined
- `patchelf` will run on files primed by parts without `override-prime` defined

I would slightly reword this message - 'unsafe' makes me think that the feature is expected to work but experimental. Assuming the statement above is true, perhaps something like:
```
Warning: 'enable-patchelf' feature will not apply to files primed by parts that use 
the 'override-prime' keyword. It's not possible to track file changes in the prime directory.
```

mr-cal (commented):
Overall looks good, left a question about the wording of the message.

tests/unit/parts/test_lifecycle.py
syu-w: It should call it, but due to the bug it does not. I am not sure if we want to check it is not called.

snapcraft/parts/lifecycle.py
syu-w: > Just so I am clear, is the following statement correct?
> 
> In a project with multiple parts:
> 
> * `patchelf` will not run on files primed by parts with `override-prime` defined
> * `patchelf` will run on files primed by parts without `override-prime` defined

Yes, the statement is correct for now, bug we may change it to something else very soon. So I don't know if I should say it will not work or it may broke under certain circumstances.

tests/unit/parts/test_lifecycle.py
mr-cal: I posit that we should test for 3 reasons:
1. even if the behavior is undesired, the unit test should still test the behavior.  (the same principle behind `xfail(strict=True)`)
2. when craft-parts is fixed, this test will start failing and alert us that the fix was effective
3. commented out code is a code smell

snapcraft/parts/lifecycle.py
mr-cal: I wouldn't change the wording based on how soon the fix is coming.  I think it needs to be clearer that snapcraft is not 'unsafe' but 'will not work as expected'

syu-w: OK, updated

mr-cal (approved):
:+1:

snapcraft/parts/lifecycle.py
```diff
@@ -186,6 +186,22 @@ def _run_command(  # noqa PLR0913 # pylint: disable=too-many-branches, too-many-
         lifecycle.clean(part_names=part_names)
         return
 
+    # patchelf relys on known file list to work properly, but override-prime
```
cmatsuoka, line 189: typo, s/relys/relies/

snapcraft/parts/lifecycle.py
```diff
@@ -186,6 +186,22 @@ def _run_command(  # noqa PLR0913 # pylint: disable=too-many-branches, too-many-
...
+    # parts don't have the list yet, and we can't track changes in the prime.
+    if any(
+        "enable-patchelf" in _part.get("build-attributes", [])
+        for _part in getattr(project, "parts", {}).values()
+    ) and any(
+        _part.get("override-prime", None)
+        for _part in getattr(project, "parts", {}).values()
+    ):
```
cmatsuoka, line 197: The problem should only happen if `override-prime` and `enable-patchelf` are in the same part, so we can loop over parts and test both attributes in each iteration.

cmatsuoka (changes requested):
Thanks, we only need to adjust the logic for the problem detection -- ELF patching is applied per part so the problem should only happen in patchelf and override-prime are used together in the same part.

cmatsuoka (approved):
Thanks! That should keep the developers aware until we implement a proper solution in craft-parts.

syu-w:
Force pushed due to re-target to hotfix

PR #4557, opened 2024-02-01
feat!: fail if we know the project is core24

tigarmo:
The purpose of this PR is two-fold:

1) Add breaking code that checks whether the project is known to be core24, to fail early. Otherwise, if the project has a different base (or if there's no project) then we fallback to the current codepath as usual.
2) Add a set of core24 spread tests, each with an expected failure.

A lot of the changes are copied over from existing tests, so some commits can mostly be skimmed. I tried to split them up in a way that makes functional sense (to me at least):

1) First commit has the actual feature of checking the base and adding a new RuntimeError;
2) Second commit adds core24 spread tests, but as a straight-up copy of the existing tests in `tests/spread/core22/`. These tests are unchanged and still have `core22` as base, so they don't need a big review;
3) Third commit has the mostly-automated change of base of the core24 tests. I think it's worth a look over because I had to do a few manual changes;
4) The fourth commit adds all expected failures in these new tests because none of them passes cleanly. For example, all tests that call `snapcraft pack` are expected to fail with `"pack" command is not implemented for core24`. The tests now check for this message and exit.
5) Finally, the fifth commit restricts the CI spread runs to only `tests/spread/core24`. I confirmed that the other tests are passing, and restricting this feature-branch to only core24 tests speeds up the development loop considerably.


snapcraft/application.py
```diff
@@ -41,6 +43,11 @@
 class Snapcraft(Application):
     """Snapcraft application definition."""
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # Whether we know that we should use the core24-based codepath.
+        self._known_core24 = False
```
syu-w, line 49: Maybe `self._known_core = "core24"`? So we can handle multiple core fallback.

tigarmo: Are you thinking of future `core26` cases or previous `core22` and `core20`?

tigarmo:
I had to force-push this to re-seat it on the new `feature/craft-application` branch. Sorry!

snapcraft/application.py
syu-w: I was thinking the `core22` and `core20`

tigarmo: On in this case the current code already handles this - this code checks that the base isn't `core24`, then raises the `ClassicFallback` error. The non-craft-app-based code then has [its own handling](https://github.com/snapcore/snapcraft/blob/98550856e8516d956fa32101472ceb1d0dedd025/snapcraft/cli.py#L263) if the base isn't `core22`. So there's no need to handle multiple types of fallbacks here.

syu-w (approved):

tests/spread/core24/architectures/task.yaml
```diff
@@ -29,15 +31,23 @@ execute: |
 
   # if the environmental variable file exists, then call snapcraft with the environmental variables
   if [[ -e "environmental-variables.txt" ]]; then
+    # If this next line fails, delete it and replace it with the one following it
```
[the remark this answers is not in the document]
tigarmo: yeah I thought about doing this but I thought it might get too confusing with "comments that are for shellcheck annotations" and "comments that are commenting out the real test". In the end I think this current status is fine, shouldn't be like this for too long anyway. Thanks for the review.

tigarmo:
> Excellent work! Looks like we will have to adjust the required CI jobs for this branch

thanks! I updated the workflow to keep the name of the required step

PR #4559, opened 2024-02-04
bash-completion: update and improve supported completion cases

kubiko:
multiple improvements to the completion logic
- add multiple new commands,
- replace `snap` with `pack`
- add supported architectures for `--target-arch` and `--build-for`
  - do not parse `snapcraft.yaml` as the format changes between core20 and core22 bases, and we have only a list of predefined options regardless, use those.
- build | clean | pull | stage | prime: parse `snapcraft.yaml` for suggested parts
  - use `shell`/`sed` based custom parser. we do not want to bring yq as an extra dependency, and calling back to the python parser demonstrated to be very slow


kubiko:
@cmatsuoka @sergiusens I finally for a moment to revisit this.
Given that `yq` is out as outside dependency, and callback to python as tested by Claudio was very slow, I went back to shell option. Any shell-based yaml parsers I could find are overkill and usually are also a bit slow. Processing `snapcraft.yaml` line by line was also dog slow. So while admittedly not as readable, `sed` seems to be the way.
I am sure I did not catch all the corner cases, but so far this has worked on every `snapcraft.yaml` I could throw at it,

debian/snapcraft-completion
```diff
@@ -26,6 +40,26 @@ _snapcraft()
         COMREPLY=( travis )
         return 0
         ;;
+    --target-arch | --build-for)
+        arch_options="arm64 armhf amd64 riscv64"
```
lengau, line 44:
```suggestion
        arch_options="arm64 armhf amd64 i386 ppc64el riscv64 s390x"
```

debian/snapcraft-completion
```diff
@@ -1,3 +1,17 @@
+list_snapcraft_parts() {
+    # first get intend
```
lengau, line 2:
```suggestion
    # first get indentation
```

debian/snapcraft-completion
```diff
@@ -6,15 +20,15 @@ _snapcraft()
     COMPREPLY=()
     cur="${COMP_WORDS[COMP_CWORD]}"
     prev="${COMP_WORDS[COMP_CWORD-1]}"
-    opts="help init list-plugins plugins login logout export-login list-keys keys create-key register-key register registered list-registered push release clean cleanbuild pull build sign-build stage prime snap update define search gated validate history status close enable-ci expand-extensions extension extensions list-extensions"
+    opts="help init list-plugins plugins login logout export-login list-keys keys create-key register-key register registered list-registered push release clean cleanbuild pull build sign-build stage prime pack update define search gated validate history status close enable-ci expand-extensions extension extensions list-extensions --target-arch --build-for --debug --shell --shell-after ---provider -h --help --verbose"
```
lengau, line 23: Most of these new parameters should only be available if a lifecycle command has already been declared. For example, `snapcraft plugins --build-for` isn't valid

lengau (changes requested):

debian/snapcraft-completion
kubiko: @lengau I did skip them on purpose as in practice cross building to those arches is very unlikely. But that is probably not a good enough excuse to omit those options :)

debian/snapcraft-completion
kubiko: @lengau that is fair point, let me see if this can be fixed in easy way

lengau: Probably the easiest way would be to check whether any of the lifecycle commands are in `$COMP_WORDS`:

```
for i in "${COMP_WORDSS}"; do
 case "$i" in  
  pull);&
  build);&
  stage);&
  prime);&
  pack);& 
  clean) 
    opts="${opts} --build-for --target-arch --debug --shell --shell-after --provider"
    break
 esac
done```

lengau: @kubiko any update?

mr-cal:
Closing this PR because it's stale and this script will soon be replaced by @bepri's solution in https://github.com/canonical/craft-cli/pull/319.

Issue #4470, opened 2023-12-01
feature/craft-application: implement Package service


mr-cal:
### What needs to get done

Implement the Package service partially implemented [here](https://github.com/snapcore/snapcraft/blob/feature/craft-application/snapcraft/services/package.py).

Similar to [this Rockcraft commit](https://github.com/canonical/rockcraft/pull/354/commits/f490ecb3be99a96fbd7f37444f9ec3f31b5f3c5d#diff-cf13f6f606f30afa23cf9954776527355fc805a6d81f60f29ec6899c261d83e9).

### Why it needs to get done

To rebase snapcraft onto craft-application

mr-cal:
Complete via #4565

PR #4565, opened 2024-02-06
feat: implement core24 "pack" command

tigarmo:
As before, this PR is split over multiple commits:

- First one is mostly a cherry-pick of @lengau and @mr-cal 's prototype. I had to do some minor changes to satisfy linters but nothing major;
- Second commit adds the "clean" command, just because it's so easy to implement and used by basically every spread test;
- Third commit re-enables "pack" and "clean" in a bunch of spread tests;
- Final commit updates the expected error messages for those tests that are still failing due to _other_ missing features.

One unfortunate consequence of the spread parametrization that we do in some tests is that now there's no easy way to handle some tests; for instance, `tests/spread/core24/craftctl` has three "subtests" as spread parameters; one is passing and two are failing (because they need adopt-info). Rather than doing a complicated logic in bash to handle this, I think we might just accept these failures in this feature-branch while we work on the issues. If that's acceptable, it means we'll need someone with admin access to ultimately merge this when it's approved.

### DIFF
--- a/snapcraft/models/project.py
+++ b/snapcraft/models/project.py
@@ -25,11 +25,13 @@ from craft_application.models import BuildInfo, UniqueStrList
 from craft_archives import repo
 from craft_cli import emit
 from craft_grammar.models import GrammarSingleEntryDictList, GrammarStr, GrammarStrList
+from craft_providers import bases
 from pydantic import PrivateAttr, constr
 
 from snapcraft import parts, utils
 from snapcraft.elf.elf_utils import get_arch_triplet
 from snapcraft.errors import ProjectValidationError
+from snapcraft.providers import SNAPCRAFT_BASE_TO_PROVIDER_BASE
 from snapcraft.utils import (
     convert_architecture_deb_to_platform,
     get_effective_base,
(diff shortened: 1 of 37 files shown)


mr-cal:
Fixes #4470

tigarmo:
@lengau @mr-cal this is ready for review!

tests/spread/core24/linters/library/task.yaml
```diff
@@ -1,14 +1,13 @@
...
-  # If this next line fails, delete it and replace it with the one following it
-  snapcraft clean 2>&1 | MATCH '"clean" command is not implemented for core24!'; exit 0
+  snapcraft clean 2>&1 | MATCH 'snapcraft internal error: KeyError'; exit 0
   # snapcraft clean
   rm -f ./*.snap ./*.txt
 
 execute: |
   # If this next line fails, delete it and replace it with the one following it
-  snapcraft 2>&1 | MATCH '"pack" command is not implemented for core24!'; exit 0
+  snapcraft 2>&1 | MATCH 'snapcraft internal error: KeyError'; exit 0
```
lengau, line 10: This seems like an odd error to be expecting. Can you explain why it's doing that?

lengau (approved):
Looks good, though I don't understand one of the expected errors and want to make sure it's not a `craft-application` bug 🙂

tests/spread/core24/linters/library/task.yaml
tigarmo: It's a craft-application bug that only showed up here because [the project doesn't have a version](https://github.com/canonical/craft-application/blob/main/craft_application/application.py#L567). I'll create a separate craft-app bug for it, but it's going to (or should be) addressed when we implement adopt-info/dynamic versioning

snapcraft/models/project.py
```diff
@@ -737,8 +738,34 @@ def get_build_for_arch_triplet(self) -> Optional[str]:
 
     def get_build_plan(self) -> List[BuildInfo]:
         """Get the build plan for this project."""
-        # TODO
-        raise NotImplementedError("Not implemented yet!")
+        # pylint: disable=import-outside-toplevel
+        from snapcraft.providers import SNAPCRAFT_BASE_TO_PROVIDER_BASE
```
mr-cal, line 742: Why not import at the top of the module?

mr-cal (approved):
Looks good (with the caveat that the spread tests are replacing the missing unit tests in the feature branch)

snapcraft/models/project.py
tigarmo: I guess it's a bit of a (bad) habit of mine, I'll revert this

tigarmo:
> (with the caveat that the spread tests are replacing the missing unit tests in the feature branch)

Yes, that's true. The thing is that there's so much stuff missing currently that it took long enough just to get the spread tests under "control", and I think at this point they are worth more than the individual unit tests in showing us what's missing/broken.
For the upcoming smaller tasks we should definitely add regular tests though

mr-cal:
Rebase-merging with failing CI because the failures are what we expect to fail for craft-application in this feature branch.

PR #4571, opened 2024-02-07
test: fix spread multipass backend

syu-w:
Fix spread multipass to use ssh password with newer sshd.
Also allowed use multiple workers.


lengau (commented):
Nice! Can you add this in Charmcraft too?

mr-cal (approved):
Nice, works for me!

lengau (approved):

PR #4586, opened 2024-02-15
refactor: move core22 commands for versioned implementation

syu-w:
This is a pure rename with no code changes.


lengau (approved):

mr-cal (approved):

syu-w:
> @syu-w two things:
> 
> 1. Since we're changing core22/legacy code I think we should temporarily run all spread tests to ensure the non-core24 ones are still working. You can change that [here](https://github.com/canonical/snapcraft/blob/feature/craft-application/.github/workflows/spread.yml#L62).
> 2. Thanks for noticing the wrong default command. I see that this opened up some new failures, but I don't think you should investigate them on this PR because it's out-of-scope. Once this is merged I'll review these new errors, as I put in that bug in the first place.
> 
> Thanks!

OK, full spread tests enabled. I think the new errors are come from changes in the craft-application, as we are using git version as dep.

syu-w:
All errors seems related to craft-application. The LXD related errors are failed to some snapcraft command in LXD.

tigarmo (approved):
thanks!

sergiusens (approved):

PR #4589, opened 2024-02-16
feat(lifecycle): add deprecated snap command

syu-w:


tests/unit/conftest.py
```diff
@@ -426,4 +430,26 @@ def package_service(default_project, default_factory):
     )
 
 
-# pylint: enable=import-outside-toplevel
```
sergiusens, line 429: why did you remove this line?

tests/conftest.py
```diff
@@ -59,3 +72,208 @@ def memory_keyring():
...
+            "testcraft",
+            "A fake app for testing snapcraft",
+            BuildPlannerClass=MyBuildPlanner,
+            source_ignore_patterns=["*.snap", "*.charm", "*.starcraft"],
+            features=craft_application.AppFeatures(**features),
+        )
+
+
+@pytest.fixture()
+def fake_project() -> models.Project:
```
sergiusens, line 131: we now have a default_project in unit/conftest and a variation of this here

tests/conftest.py
```diff
@@ -59,3 +72,208 @@ def memory_keyring():
...
+
+
+@pytest.fixture()
+def fake_build_plan() -> list[models.BuildInfo]:
+    arch = util.get_host_architecture()
+    return [models.BuildInfo("foo", arch, arch, bases.BaseName("ubuntu", "22.04"))]
+
+
+@pytest.fixture()
+def enable_overlay() -> Iterator[craft_parts.Features]:
```
sergiusens, line 157: why would we need overlay in here?

tests/unit/conftest.py
syu-w: Oh, I did not notice it was disabled above. Will add back.

tests/conftest.py
syu-w: Removed

tests/conftest.py
syu-w: I put the two into one as the `fake_project`

snapcraft/commands/lifecycle.py
```diff
@@ -0,0 +1,48 @@
...
+"""Snapcraft lifecycle commands."""
+
+import argparse
+from typing import Any
+
+import craft_application.commands
+from craft_cli import emit
+from typing_extensions import override
+
+# pylint: disable=too-many-ancestors
```
tigarmo, line 26: weird, how come this is needed? The class only has a single parent

tests/conftest.py
```diff
@@ -59,3 +69,201 @@ def memory_keyring():
...
+
+
+class MyBuildPlanner(models.BuildPlanner):
+    """Build planner definition for tests."""
+
+    def get_build_plan(self) -> list[models.BuildInfo]:
+        arch = util.get_host_architecture()
+        return [
+            models.BuildInfo(
+                "ubuntu-22.04", arch, arch, bases.BaseName("ubuntu", "22.04")
```
tigarmo, line 88: I'm not sure why these test fixtures are being copied from craft-application, but I think these bases should specify `24.04` at least

snapcraft/commands/lifecycle.py
syu-w: It has a single parent, but inherited 8 times

tigarmo: ah I see

tigarmo (approved):

sergiusens:
I am still not seeing why we cannot use the fixtures like we had from Rockcraft and we we had to bring in the Craft Application ones

tigarmo:
> I am still not seeing why we cannot use the fixtures like we had from Rockcraft and we we had to bring in the Craft Application ones

I agree actually. We don't need to bring in everything wholesale, we can put in just the things needed for the new tests. Simpler and less code to maintain.

syu-w:
This should significantly simplified the fixtures.

(I had to force push due to merged wrong branch)

tests/conftest.py
```diff
@@ -59,3 +68,110 @@ def memory_keyring():
...
+    @pytest.mark.enable_features("build_secrets")
+    def test_with_build_secrets(...)
+    """
+    features = {}
+
+    for feature_marker in request.node.iter_markers("enable_features"):
+        for feature_name in feature_marker.args:
+            features[feature_name] = True
+
+    return features
```
sergiusens, line 95: we do not use any features in Snapcraft, so why have this?

sergiusens:
to set things straight, I would say, rebase and drop 4f492aa6504f469c7b40b490ca713bba2e08715b and work on extending the fixtures we have with the things we need

sergiusens:
@syu-w you did not seem to extend anything still, you moved the fixtures from tests/unit/conftest.py to tests/conftests.py. What is the justification for this?

syu-w:
> @syu-w you did not seem to extend anything still, you moved the fixtures from tests/unit/conftest.py to tests/conftests.py. What is the justification for this?

The snap command test only need to add the `fake_services` which using the 2 exist fixtures in the `tests/unit/conftest.py`. And it does not need to extend more fixtures for it to work.

syu-w:
Rebase with no changes.

tigarmo:
@syu-w not sure if you are aware but there are a lot of new spread test failures
```
2024-02-22 22:35:04 Error preparing google:ubuntu-22.04-64:tests/spread/general/verbosity (feb222220-944854) : 
-----
+ snapcraft init
  File "/snap/snapcraft/x1/bin/snapcraft", line 8, in <module>
    sys.exit(main())
  File "/snap/snapcraft/x1/lib/python3.10/site-packages/snapcraft/application.py", line 342, in main
    return app.run()
  File "/snap/snapcraft/x1/lib/python3.10/site-packages/craft_application/application.py", line 451, in run
    dispatcher = self._get_dispatcher()
  File "/snap/snapcraft/x1/lib/python3.10/site-packages/snapcraft/application.py", line 184, in _get_dispatcher
    dispatcher = craft_cli.Dispatcher(
  File "/snap/snapcraft/x1/lib/python3.10/site-packages/craft_cli/dispatcher.py", line 230, in __init__
    self.commands = _get_commands_info(commands_groups)
  File "/snap/snapcraft/x1/lib/python3.10/site-packages/craft_cli/dispatcher.py", line 193, in _get_commands_info
    raise RuntimeError(
RuntimeError: Multiple commands with same name: Snap and SnapCommand
-----
```

sergiusens (approved):

PR #4596, opened 2024-02-21
fix(package): use correct target-arch on snap

tigarmo:
This PR does a few things, over 3 commits:

- Bring the existing tests inline with the corresponding craft-application changes (where the Lifecycle service now takes a build plan)
- Fix an issue with the wrong build-for in the Package service
- Update the related spread tests. Some are now passing, some are still failing from lack of SNAPCRAFT_BUILD_FOR and some are removed because we won't support those features in core24.

Leaving as a Draft until the craft-application dependency is merged.


tests/spread/core24/architectures/snaps/build-for-all/snap/snapcraft.yaml
tigarmo, line 1: This is removed because we decided to not support "build-for: all" for now

tests/spread/core24/architectures/snaps/multiple-build-for-with-grammar/snap/snapcraft.yaml
tigarmo, line 1: This is removed because it wasn't being called in the first place! And it's identical to `multiple-build-for`

tests/unit/conftest.py
```diff
@@ -413,7 +413,22 @@ def default_factory(default_project):
...
 
 @pytest.fixture()
-def lifecycle_service(default_project, default_factory, tmp_path):
+def default_build_plan():
+    from craft_application.models import BuildInfo
+    from craft_providers import bases
+
+    return [
+        BuildInfo(
+            platform="amd64",
```
sergiusens, line 422: mind calling this generic-x86-64 or something like that so no one gets confused with the intent of the field?

sergiusens (approved):

mr-cal (approved):

PR #4602, opened 2024-02-22
build(lint): use ruff from snap

lengau:

-----

### DIFF
--- a/.github/workflows/tox.yaml
+++ b/.github/workflows/tox.yaml
@@ -25,7 +25,7 @@ jobs:
           echo "::group::Begin snap install"
           echo "Installing snaps in the background while running apt and pip..."
           sudo snap install --no-wait --classic pyright --revision 735 # version 1.1.344
-          sudo snap install --no-wait shellcheck
+          sudo snap install --no-wait shellcheck ruff
           echo "::endgroup::"
           echo "::group::apt-get update"
           sudo apt-get update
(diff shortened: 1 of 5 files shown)


sergiusens (approved):
\o/

Now do pyright :-D

mr-cal (approved):

lengau:
@sergiusens I would genuinely love to! My problem with pyright isn't that it's not a great IDE tool, but that release upgrades will suddenly break CI, which can derail work. Ruff has [a policy](https://docs.astral.sh/ruff/versioning/) that has so far prevented these issues, as we seldom have to make a PR to "satisfy ruff." As the app matures, I hope to set up major release channels as well so that we can upgrade to new major releases in a controlled manner.

Similarly stable tools that I'd be comfortable moving over to snap channels (assuming the snaps have regular releases and channels gating major verssions) include black, mypy, codespell and yamllint.

I've posted a [feature request](https://github.com/microsoft/pyright/issues/7315) for pyright to provide some level of stability. Even if pyright had less overall coverage in CI, as a trade-off for being able to consistently have the latest pyright I think it'd be well worth it.

sergiusens:
I use pyright as my LSP :-)
I guess I should look into going back to the python backed one and use ruff
as an add-on to that.

lengau:
Yeah I like it as an LSP. The CI breakage is the only real downside

PR #4611, opened 2024-02-23
feat(lifecycle): generate snap manifest

cmatsuoka:
If `SNAPCRAFT_BUILD_INFO` is set, retrieve package information from
the parts lifecycle and create a manifest containing the pull assets
and primed stage packages. Also add a copy of the original project
file.

If `SNAPCRAFT_IMAGE_INFO` is set, add the provided json data to the
snap manifest.

-----


snapcraft/application.py
```diff
@@ -141,6 +141,14 @@ def _configure_services(self, platform: str | None, build_for: str | None) -> No
         self.services.set_kwargs("package", platform=platform, build_for=build_for)
         super()._configure_services(platform, build_for)
 
+        self.services.set_kwargs(
+            "lifecycle",
+            cache_dir=self.cache_dir,
+            work_dir=self._work_dir,
+            build_plan=self._build_plan,
+            project_path=self._resolve_project_path(None),
```
tigarmo, line 149: I don't think this is a workable solution; trying to resolve the project path here breaks cases that don't have projects at all, like:
image

cmatsuoka: Fixed to account for non-existing projects, failing tests are again in the expected set.

snapcraft/services/lifecycle.py
```diff
@@ -36,5 +68,47 @@ def setup(self) -> None:
...
+
+    @overrides
+    def run(self, step_name: str | None, part_names: list[str] | None = None) -> None:
+        super().run(step_name, part_names)
+
+        enable_manifest = utils.strtobool(os.getenv("SNAPCRAFT_BUILD_INFO", "n"))
+        if enable_manifest:
+            self._generate_manifest()
+
+    def _generate_manifest(self) -> None:
```
tigarmo, line 81: we should definitely have tests for the un-covered lines in this PR

snapcraft/services/lifecycle.py
```diff
@@ -36,5 +68,47 @@ def setup(self) -> None:
...
+        manifest.write(
+            project,
+            self.prime_dir,
+            arch=utils.get_host_architecture(),
+            parts=parts,
+            start_time=self._start_time,
+            image_information=image_information,
+            primed_stage_packages=list(primed_stage_packages),
+        )
+        emit.progress("Generated snap manifest", permanent=True)
```
tigarmo, line 110: should this be a permanent message?

snapcraft/services/lifecycle.py
```diff
@@ -36,5 +68,47 @@ def setup(self) -> None:
...
+                part.setdefault(key, [])
+
+            stage_packages = self.get_primed_stage_packages(part_name=name)
+            if stage_packages:
+                primed_stage_packages |= set(stage_packages)
+
+        manifest.write(
+            project,
+            self.prime_dir,
+            arch=utils.get_host_architecture(),
```
tigarmo, line 104: is the host architecture right here? Is it not supposed to be the target arch/build-for?

snapcraft/services/lifecycle.py
```diff
@@ -36,5 +68,47 @@ def setup(self) -> None:
...
+            parts=parts,
+            start_time=self._start_time,
+            image_information=image_information,
+            primed_stage_packages=list(primed_stage_packages),
+        )
+        emit.progress("Generated snap manifest", permanent=True)
+
+        # Also copy the original snapcraft.yaml
+        if self._project_path:
```
tigarmo, line 113: shouldn't it be an error if `self._project_path` is None?

snapcraft/services/provider.py
```diff
@@ -0,0 +1,32 @@
...
+from craft_application import ProviderService
+from overrides import overrides
+
+
+class Provider(ProviderService):
+    """Snapcraft specialization of the Lifecycle Service."""
+
+    @overrides
+    def setup(self) -> None:
+        if build_info := os.getenv("SNAPCRAFT_BUILD_INFO"):
```
tigarmo, line 29: this needs a test too

snapcraft/services/lifecycle.py
```diff
@@ -15,17 +15,49 @@
...
+    def __init__(  # noqa: PLR0913 (too many arguments)
+        self,
+        app: AppMetadata,
+        services: ServiceFactory,
+        *,
+        project: models.Project,
+        work_dir: Path | str,
+        cache_dir: Path | str,
+        build_plan: list[BuildInfo],
+        project_path: Path | None,
```
tigarmo, line 46: Seems like `project_path` should always be a `Path`. When could it validly be `None`?

snapcraft/application.py
```diff
@@ -155,7 +168,10 @@ def _resolve_project_path(self, project_dir: pathlib.Path | None) -> pathlib.Pat
         try:
             return super()._resolve_project_path(project_dir / "snap")
         except FileNotFoundError:
-            return super()._resolve_project_path(project_dir)
+            try:
```
mr-cal, line 171: At some point (probably not this PR), we should refactor the nested try-excepts and raise a descriptive error message.  

Otherwise, they will get a FileNotFound for `build-aux/snap/snapcraft.yaml`, which may be confusing.

mr-cal (commented):
I don't have anything notable to add ontop of Tiago's review.  

Please re-request me for a review once you address his feedback, thanks!

snapcraft/services/lifecycle.py
cmatsuoka: Since this happens at the end of the lifecycle, project path should never be None at this point. But sure, we should issue an error since we're testing it.

snapcraft/services/lifecycle.py
cmatsuoka: Given the brevity of current snapcraft output this could well be an ephemeral message in brief mode.

snapcraft/services/lifecycle.py
cmatsuoka: This must be build-for. Will extract it from the build plan which must have a single entry at this point.

snapcraft/services/lifecycle.py
cmatsuoka: Never. At this point it must always be valid.

snapcraft/services/lifecycle.py
cmatsuoka: Testing for more fields. Override scripts can be added after the multi-line representation is fixed.

snapcraft/services/provider.py
cmatsuoka: Added a provider service unit test. Functionality is covered by the metadata spread tests.

snapcraft/services/lifecycle.py
```diff
@@ -36,5 +68,49 @@ def setup(self) -> None:
...
+                part.setdefault(key, [])
+
+            stage_packages = self.get_primed_stage_packages(part_name=name)
+            if stage_packages:
+                primed_stage_packages |= set(stage_packages)
+
+        host_arch = utils.get_host_architecture()
+        build_for = self._build_plan[0].build_for if self._build_plan else host_arch
+
+        manifest.write(
```
sergiusens, line 104: writing of the manifest should happen in the pack service (write_metadata)

snapcraft/services/lifecycle.py
cmatsuoka: At this moment yes. It is checked by review tools and a selection of fields are also expected to match.

snapcraft/services/lifecycle.py
cmatsuoka: Refactored to write the manifest along the rest of the metadata. Lifecycle-related data is still collected during the execution of the lifecycle service, as this information is not available elsewhere.

snapcraft/application.py
```diff
@@ -129,6 +129,11 @@ def __init__(self, *args, **kwargs):
         # Whether we know that we should use the core24-based codepath.
         self._known_core24 = False
 
+        try:
```
sergiusens, line 132: can you add a code comment on why this is needed?

cmatsuoka: Added comment explaining what the project path is used for.

snapcraft/services/lifecycle.py
```diff
@@ -36,5 +65,75 @@ def setup(self) -> None:
...
-
         super().setup()
+
+    @overrides
+    def run(self, step_name: str | None, part_names: list[str] | None = None) -> None:
+        super().run(step_name, part_names)
+
+        enable_manifest = utils.strtobool(os.getenv("SNAPCRAFT_BUILD_INFO", "n"))
+        if enable_manifest:
+            self._manifest = self._generate_manifest()
```
tigarmo, line 76: should this be done for any `step_name`? Or only `prime`?

cmatsuoka: You are right, and the same applies to mandatory project variable verification. It can't run in a lifecycle epilogue callback because it runs at the end of any lifecycle run regardless of target step, and the post-prime callback runs for each part. It seems that specifically testing for the target step is necessary?

cmatsuoka: Updated craft-application in https://github.com/canonical/craft-application/pull/257 to make manifest data generation (and mandatory adoptable field checking) to happen in the package service.

requirements-devel.txt
```diff
@@ -11,7 +11,7 @@ click==8.1.7
 codespell==2.2.6
 colorama==0.4.6
 coverage==7.4.0
-craft-application @ git+https://github.com/canonical/craft-application@main
+craft-application @ git+https://github.com/cmatsuoka/craft-application@package-service-project-update
```
cmatsuoka, line 14: Reminder: update to main before landing.

tests/unit/conftest.py
```diff
@@ -444,14 +446,33 @@ def lifecycle_service(default_project, default_factory, default_build_plan, tmp_
...
     from snapcraft.services import Package
 
+    file_path = tmp_path / "snapcraft.yaml"
+    snapcraft_yaml(filename=file_path)
+
     return Package(
         app=APP_METADATA,
         project=default_project,
         services=default_factory,
+        project_path=file_path,
```
sergiusens, line 475: the project is much more that the snapcraft.yaml, we also have hooks, gui and other assets, I would consider project_path the snapcraft project root directory

sergiusens: in the case of charms, the project_path can also include metadata.yaml, action.yaml. It seems a bit wasteful to only use it for one yaml

cmatsuoka: I suggest we refactor this when adding hooks and assets to the payload so we can limit the scope of this PR and have it landed.

snapcraft/application.py
lengau: Agreed - also that error should probably be a form of `CraftError` when it comes out of this since we don't want it to me marked as an internal error.

tests/unit/services/test_package.py
```diff
@@ -57,18 +57,24 @@ def test_pack_target_arch(default_project, default_factory, mocker):
...
     )
 
-    package_service.pack(prime_dir=Path("prime"), dest=Path())
+    package_service.pack(prime_dir=tmp_path / "prime", dest=tmp_path)
 
     assert mock_pack_snap.call_args.kwargs["target_arch"] == "s390x"
 
 
-def test_metadata(package_service, default_factory, default_build_plan, new_dir):
+def test_metadata(
```
sergiusens, line 68: can we add a new test (new please), with setenv for the manifest to be generated?

sergiusens (approved):

tests/unit/services/test_package.py
lengau: Done! While I was at it I ensured that the manifest wasn't generated if we didn't ask for it. I also added a unit test for `generate_manifest` on the lifecycle service that we can further parametrize as needed.

sergiusens (approved):

snapcraft/application.py
lengau: Kicking the can down the road: https://github.com/canonical/snapcraft/issues/4636

I've placed this on next pulse, but I may have time to do it today.

syu-w (approved):
LGTM, some Manifest fields need to be added later.

PR #4613, opened 2024-02-26
feat: adopt more fields from the appstream metadata file

soumyaDghosh:
This commit adds initial support to adopt more metadata fields from the provided appstream metadata file. Those fields includes,

1. license
2. contact
3. source-code
4. issues
5. website
6. donations

----

-----

### DIFF
--- a/snapcraft/meta/appstream.py
+++ b/snapcraft/meta/appstream.py
@@ -106,7 +106,7 @@ def extract(relpath: str, *, workdir: str) -> Optional[ExtractedMetadata]:
     issues = _get_urls_from_xml_element(dom.findall("url"), "bugtracker")
     donation = _get_urls_from_xml_element(dom.findall("url"), "donation")
     website = _get_urls_from_xml_element(dom.findall("url"), "homepage")
-    source_code = _get_urls_from_xml_element(dom.findall("url"), "vcs-browser")
+    source_code = _get_urls_from_xml_element(dom.findall("url"), "vcs-browser")[0]
 
     desktop_file_paths = []
     desktop_file_ids = _get_desktop_file_ids_from_nodes(dom.findall("launchable"))
--- a/snapcraft/meta/appstream.py
+++ b/snapcraft/meta/appstream.py
@@ -100,13 +100,13 @@ def extract(relpath: str, *, workdir: str) -> Optional[ExtractedMetadata]:
     description = _get_value_from_xml_element(dom, "description")
     title = _get_value_from_xml_element(dom, "name")
     version = _get_latest_release_from_nodes(dom.findall("releases/release"))
-    license = _get_value_from_xml_element(dom, "project_license")
+    project_license = _get_value_from_xml_element(dom, "project_license")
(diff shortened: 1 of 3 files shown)


sergiusens (approved):
thanks!

snapcraft/meta/appstream.py
```diff
@@ -100,6 +100,13 @@ def extract(relpath: str, *, workdir: str) -> Optional[ExtractedMetadata]:
     description = _get_value_from_xml_element(dom, "description")
     title = _get_value_from_xml_element(dom, "name")
     version = _get_latest_release_from_nodes(dom.findall("releases/release"))
+    project_license = _get_value_from_xml_element(dom, "project_license")
+    contact = _get_value_from_xml_element(dom, "update_contact")
+
+    issues = _get_urls_from_xml_element(dom.findall("url"), "bugtracker")
+    donation = _get_urls_from_xml_element(dom.findall("url"), "donation")
+    website = _get_urls_from_xml_element(dom.findall("url"), "homepage")
+    source_code = _get_source_code_from_xml_element(dom.findall("url"))
```
lengau, line 109: This seems like a special case of `_get_urls_from_xml_element`?

```suggestion
    source_links = _get_urls_from_xml_element(dom.findall("url"), "vcs-browser")
    source_code = source_links[0] if source_links else None
```

lengau (changes requested):

lengau (changes requested):
I don't believe this'll do everything intended. You'll need to update `update_project_metadata` to use all of these fields:

https://github.com/canonical/snapcraft/blob/f06dd8e317e96efd427a55f11873648f393c7e30/snapcraft/parts/update_metadata.py#L53

soumyaDghosh:
> I don't believe this'll do everything intended. You'll need to update `update_project_metadata` to use all of these fields:
> 
> https://github.com/canonical/snapcraft/blob/f06dd8e317e96efd427a55f11873648f393c7e30/snapcraft/parts/update_metadata.py#L53

Yes I do have that change ready in my local machine, I'll push it in my next PR. That makes some bug changes in the Project and Metadata section, that's why I hold it.

snapcraft/meta/appstream.py
soumyaDghosh: According to the docs of snapcraft `source-code` is the only one that can be a `string`. It can't be a `List[str]`. So, if I do return it from the same function, it'll return a list, that'd change the definition in the docs. Should I do that?

mr-cal (approved):
Updating a spread test like `tests/spread/general/appstream-desktop`  would be nice but not required.  Thanks!

snapcraft/meta/appstream.py
lengau: That's what the second line in my suggestion does - takes the one item from `source-links` and turns it into a string.

It's not a big deal really though, and I'm fine to leave it as-is.

lengau (approved):

lengau:
@mr-cal I believe that spread test update will need to wait until part 2 of this PR

Issue #4605, opened 2024-02-22
Support for root packages for core24


sergiusens:
### What needs to get done

In the non Craft Application based snapcraft root level build-packages and stage-packages are bundled into a part before being passed to Craft Parts. This logic is missing in the new Craft Application based Snapcraft

### Why it needs to get done

Implement the similar logic to https://github.com/canonical/snapcraft/blob/8a9a5c21229f68a5e0daa7c89d609d7d29006477/snapcraft/parts/yaml_utils.py#L186 and its similar unit tests which would allow `tests/spread/core24/root-packages/task.yaml` to pass

mr-cal:
Fixed in https://github.com/canonical/snapcraft/pull/4617

PR #4617, opened 2024-02-27
feat(craft-application): support root packages for core24

syu-w:
This allows the user to use "build-packages" and "build-snaps" at the root level of the snapcraft.yaml file.

Fixes: #4605


snapcraft/models/project.py
```diff
@@ -171,6 +172,32 @@ def _validate_architectures_all_keyword(architectures):
             )
 
 
+def root_packages_transform(yaml_data: dict[str, Any]) -> dict[str, Any]:
```
lengau, line 175:
```suggestion
def transform_root_packages(yaml_data: dict[str, Any]) -> dict[str, Any]:
```
Or if you prefer, `apply_root_packages_transform`. Anything that's "<verb>_<noun> and descriptive 🙂

snapcraft/models/project.py
```diff
@@ -171,6 +172,32 @@ def _validate_architectures_all_keyword(architectures):
...
+
+    This allows the user to use "build-packages" and "build-snaps"
+    at the root level of the snapcraft.yaml file.
+    """
+    if "build-packages" not in yaml_data and "build-snaps" not in yaml_data:
+        return yaml_data
+
+    yaml_data = copy.deepcopy(yaml_data)
+    yaml_data.setdefault("parts", {})
+    yaml_data["parts"]["snapcraft/core"] = {"plugin": "nil"}
```
lengau, line 186: How concerned are we with the possibility of someone naming a part `snapcraft/core`? I gather the answer is "not very" since this is the same as the old logic, but it might be prudent to make more liberal use of `setdefault` anyway.

lengau (approved):
Looks good, thanks!

snapcraft/models/project.py
sergiusens: we want a predictable name, we need an initial guard to ensure that we only allow / for parts that should be overridden (like extensions).

Not part of this body of work though, please create an issue

snapcraft/models/project.py
sergiusens: I would use apply, so it keeps the same semantics as `apply_extensions`

tests/unit/models/test_projects.py
```diff
@@ -1782,3 +1783,20 @@ def test_project_get_build_for_arch_triplet_all(self, project_yaml_data):
         arch_triplet = project.get_build_for_arch_triplet()
 
         assert not arch_triplet
+
+
+class TestProjectTransform:
+    """Test Transform the Project."""
+
+    def test_root_packages_transform(self, project_yaml_data):
```
mr-cal, line 1791: May be useful to have a test to ensure projects without `build-packages` and `build-snaps` are unmodified  (i.e. the `if` statement on line 181)

mr-cal (approved):
:+1:

snapcraft/application.py
```diff
@@ -170,7 +173,8 @@ def _extra_yaml_transform(
     ) -> dict[str, Any]:
         arch = build_on
         target_arch = build_for if build_for else get_host_architecture()
-        return apply_extensions(yaml_data, arch=arch, target_arch=target_arch)
+        new_yaml_data = apply_extensions(yaml_data, arch=arch, target_arch=target_arch)
+        return apply_root_packages_transform(new_yaml_data)
```
sergiusens, line 177:
```suggestion
        return apply_root_packages(new_yaml_data)
```

snapcraft/models/project.py
```diff
@@ -171,6 +172,32 @@ def _validate_architectures_all_keyword(architectures):
             )
 
 
+def apply_root_packages_transform(yaml_data: dict[str, Any]) -> dict[str, Any]:
```
sergiusens, line 175:
```suggestion
def apply_root_packages(yaml_data: dict[str, Any]) -> dict[str, Any]:
```

snapcraft/models/project.py
```diff
@@ -171,6 +172,32 @@ def _validate_architectures_all_keyword(architectures):
...
+    """Support Root Packages in Snapcraft.
+
+    This allows the user to use "build-packages" and "build-snaps"
+    at the root level of the snapcraft.yaml file.
```
sergiusens, line 179:
```suggestion
    """Create a new part with root level attributes.

    Root level attributes such as build-packages and build-snaps
    are known to Snapcraft but not Craft Parts. Create a new part
    "snapcraft/core" with these attributes and apply it to the
    current yaml_data.
```

tests/unit/models/test_projects.py
```diff
@@ -1782,3 +1783,32 @@ def test_project_get_build_for_arch_triplet_all(self, project_yaml_data):
         arch_triplet = project.get_build_for_arch_triplet()
 
         assert not arch_triplet
+
+
+class TestProjectTransform:
```
sergiusens, line 1788:
```suggestion
class TestApplyRootPackages:
```

tests/unit/models/test_projects.py
```diff
@@ -1782,3 +1783,32 @@ def test_project_get_build_for_arch_triplet_all(self, project_yaml_data):
...
+    """Test Transform the Project."""
+
+    def test_root_packages_transform(self, project_yaml_data):
+        """Test transforming the project."""
```
sergiusens, line 1792:
```suggestion
    def test_apply_root_packages(self, project_yaml_data):
        """Test creating a part with root level build-packages and build-snaps."""
```

tests/unit/models/test_projects.py
```diff
@@ -1782,3 +1783,32 @@ def test_project_get_build_for_arch_triplet_all(self, project_yaml_data):
...
+
+        data_transformed = apply_root_packages_transform(data)
+
+        project = Project.unmarshal(data_transformed)
+
+        assert project.parts["snapcraft/core"]["build-packages"] == ["pkg1", "pkg2"]
+        assert project.parts["snapcraft/core"]["build-snaps"] == ["snap3", "snap4"]
+
+    def test_root_packages_transform_no_affect(self, project_yaml_data):
+        """Test transforming the project without root packages."""
```
sergiusens, line 1805:
```suggestion
    def test_no_new_part_if_missing_root_packages(self, project_yaml_data):
        """Test that nothing is applied if there are not build-packages or build-snaps."""
```

sergiusens (commented):
wording comments

syu-w:
Updated

PR #4630, opened 2024-03-04
feat: add platforms and drop architectures for core24

mr-cal:

-----

This PR is a draft.

- Add `platforms` in core24
- Drop `architectures` in core24

Similar to `architectures`, the `platforms` keyword is optional and defaults to building-on and building-for the host architecture.

As much as possible, I tried to follow charmcraft's and rockcraft's implementation of platform support.

(CRAFT-2535)


mr-cal:
@lengau or @tigarmo: could I get some early feedback of this draft?

snapcraft/models/project.py
```diff
@@ -426,6 +445,54 @@ def _validate_default_provider(cls, default_provider):
...
+    def _validate_platform_set(cls, values: Mapping[str, Any]) -> Mapping[str, Any]:
+        """Validate the build_on build_for combination."""
+        build_for: list[str] = values["build_for"] if values.get("build_for") else []
+        build_on: list[str] = values["build_on"] if values.get("build_on") else []
+
+        # We can only build for 1 arch at the moment
+        if len(build_for) > 1:
+            raise CraftValidationError(
+                str(
+                    f"Trying to build a rock for {build_for} "
```
tigarmo, line 481: (this refers to `rock`)

snapcraft/application.py
```diff
@@ -22,87 +22,129 @@
...
 
-    architectures: List[str | Architecture] = pydantic.Field(
-        default_factory=lambda: [get_host_architecture()]
-    )
-    base: str | None = None
+    base: str | None
     build_base: str | None = None
-    name: str | None = None
+    name: str
+    platforms: dict[str, Any]
```
tigarmo, line 54: could this be a `dict[str, Platform]`?

snapcraft/models/project.py
```diff
@@ -426,6 +445,54 @@ def _validate_default_provider(cls, default_provider):
         return default_provider
 
 
+class Platform(pydantic.BaseModel):
```
tigarmo, line 448: This can probably inherit from `CraftBaseModel` instead, and then you can drop the Config

tests/unit/conftest.py
```diff
@@ -397,18 +397,41 @@ def default_project(extra_project_params):
...
+    from snapcraft.models.project import Project
+
+    parts = extra_project_params.pop("parts", {})
+
+    return Project(
+        name="default",
+        version=VersionStr("1.0"),
+        summary=SummaryStr("default project"),
+        description="default project",
+        base="core22",
```
tigarmo, line 418: Is this really meant to be `core22`? As far as I know these fixtures (this and `default_factory`) are used with craft-application-based Snapcraft right?

tigarmo (commented):
Did a first review pass, thanks!

snapcraft/application.py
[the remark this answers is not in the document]
mr-cal: I'll double-check during standup but I believe:

1. Yes, it's a clean transition:  `core22` only accepts `architectures` and `core24` only accepts `platforms`
2. For `core24` `platforms` is mandatory, so there is no default value.

lengau: Probably a `dict[str, Platform | None]` so you can have:

```
platforms:
  amd64:
```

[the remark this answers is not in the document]
lengau: Note from meeting: platforms will be optional and it being missing will have the same behaviour as architectures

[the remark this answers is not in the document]
mr-cal: And the related issue based on @cmatsuoka's meeting with Gustavo: https://github.com/canonical/craft-application/issues/254

snapcraft/const.py
```diff
@@ -0,0 +1,38 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Constants used in snapcraft."""
+
+import enum
+
+
+class SnapArch(str, enum.Enum):
```
lengau, line 22: Gee, wherever you got this idea from was brilliant! 😉

snapcraft/models/project.py
```diff
@@ -426,6 +445,54 @@ def _validate_default_provider(cls, default_provider):
...
+
+        allow_population_by_field_name = True
+        alias_generator: Callable[[str], str] = lambda s: s.replace(  # noqa: E731
+            "_", "-"
+        )
+
+    @pydantic.validator("build_for", pre=True)
+    @classmethod
+    def _vectorise_build_for(cls, val: str | list[str]) -> list[str]:
+        """Vectorise target architecture if needed."""
```
lengau, line 465: I like this. It allows us to accept just a string in `build-for`, but when we generate a json schema it'll complain if it's not a list

snapcraft/models/project.py
```diff
@@ -426,6 +445,54 @@ def _validate_default_provider(cls, default_provider):
...
+            raise CraftValidationError(
+                str(
+                    f"Trying to build a rock for {build_for} "
+                    "but multiple target architectures are not "
+                    "currently supported. Please specify only 1 value."
+                )
+            )
+
+        # If build_for is provided, then build_on must also be
+        if not build_on and build_for:
```
lengau, line 488: Does this mean that this would be a valid platform:

```
platforms:
  amd64:
    build-on: [amd64, arm64]
```

snapcraft/application.py
```diff
@@ -22,87 +22,129 @@
...
+        """Make sure all provided platforms are tangible and sane."""
+        for platform_label in platforms:
+            platform: dict[str, Any] = (
+                platforms[platform_label] if platforms[platform_label] else {}
+            )
+            error_prefix = f"Error for platform entry '{platform_label}'"
+
+            # Make sure the provided platform_set is valid
+            try:
+                platform = Platform(**platform).dict()
```
lengau, line 69: Out of curiosity why are you converting this back to a dict? It doesn't seem like you're doing anything that couldn't be done with `platform.build_on` and `platform.build_for`?

snapcraft/application.py
```diff
@@ -22,87 +22,129 @@
...
+            except CraftValidationError as err:
+                # pylint: disable=raise-missing-from
```
lengau, line 71: Instead of silencing pylint here you could raise it `from None`

mr-cal:
Sorry for the force-push, there were file conflicts.  The already-reviewed commit has not been changed.

mr-cal:
> @mr-cal is this ready for review?

Not yet :upside_down_face:

snapcraft/application.py
mr-cal: Fixed

snapcraft/models/project.py
mr-cal: Correct. In this scenario, `build-for` would be the platform name.

snapcraft/models/project.py
```diff
@@ -426,6 +445,54 @@ def _validate_default_provider(cls, default_provider):
         return default_provider
 
 
+class Platform(pydantic.BaseModel):
+    """Snapcraft project platform definition."""
+
+    build_on: list[SnapArch] | None = pydantic.Field(min_items=1, unique_items=True)
+    build_for: list[SnapArch] | None = pydantic.Field(min_items=1, unique_items=True)
```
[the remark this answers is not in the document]
mr-cal: Yes.  I originally followed rockcraft's implementation which validated this manually and gave a custom error message but now I am letting Pydantic do the work with `max_items=1`.

snapcraft/models/project.py
mr-cal: Fixed.

snapcraft/application.py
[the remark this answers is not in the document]
mr-cal: Fixed.

mr-cal: I took this from [Rockcraft](https://github.com/canonical/rockcraft/blob/ac9cfd507237d64f3b619a42ec43d753c2de1ce7/rockcraft/models/project.py#L139).  If I make this too specific, I end up having to do a handlful of `if ...` or `if isinstance()` checks when making the build plan.  I'll see if I can tidy this up any more.

mr-cal:
This is now ready for review.

A few outstanding items:
1. Should the contents of `meta/snap.yaml` change from `architectures` to `platforms`?
2. I need to add a spread test for `--platforms`
3. What should be done for scenarios like [this](https://github.com/canonical/snapcraft/actions/runs/8191429098/job/22400874940?pr=4630#step:4:1664) spread test?  It has:

```
name: core2x
type: base
build-base: core22
```

Which means the effective base is `core2x`.  Should that use `architectures` or `platforms`?  Or should we update this test to be realistic (i.e change the snapcraft.yaml to `name: core22` or `name: core24`)?

tests/unit/test_application.py
```diff
@@ -39,155 +41,177 @@ def architectures(request):
 
 
 @pytest.mark.parametrize(
```
tigarmo, line 43: nice

tests/spread/core24/grammar/snap/snapcraft.yaml
```diff
@@ -3,14 +3,17 @@ version: "1.0"
 summary: test
 description: |
   Exercise snapcraft's advanced grammar keywords, `on`, and `to`.
-  This test leverages architecture keywords, `build-on` and `build-for`.
+  This test leverages the platform keywords, `build-on` and `build-for`.
 grade: devel
 confinement: strict
-base: core22
```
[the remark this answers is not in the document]
mr-cal: Or was this test in the `core24` suite accidentally not updated to `base: core24`?

tests/unit/commands/test_expand_extensions.py
```diff
@@ -31,14 +30,34 @@ class CoreData:
     base: str
     build_base: str
     grade: str
+    platforms: str
```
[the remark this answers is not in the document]
mr-cal: It is not.  I forgot to revert this after making `platforms` optional.

mr-cal:
tests/spread/core24/platforms/snaps/platform-match` fails because snapcraft is not using the build-for from the build plan and is instead using the host architecture.

Snapcraft creates the correct build plan but it feels like the build-for isn't getting passed from craft-application to craft-parts.  I haven't figured this out yet.

tigarmo:
> tests/spread/core24/platforms/snaps/platform-match` fails because snapcraft is not using the build-for from the build plan and is instead using the host architecture.
> 
> Snapcraft creates the correct build plan but it feels like the build-for isn't getting passed from craft-application to craft-parts. I haven't figured this out yet.

I think the problem is this: https://github.com/canonical/snapcraft/blob/main/snapcraft/application.py#L47
In the absence of a command-line `build-for`, that code is defaulting to the host. I think the right solution here is to pass the filtered build plan to the package service, [like craft-application does with the lifecycle service](https://github.com/canonical/craft-application/blob/main/craft_application/application.py#L224)

mr-cal:
Thank you! I'm not sure I would have figured that out on my own.

tests/spread/core24/grammar/snap/snapcraft.yaml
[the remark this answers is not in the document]
mr-cal: Ooh, you may be right: https://github.com/canonical/snapcraft/actions/runs/8195514975/job/22414086129?pr=4630#step:4:524

mr-cal:
Looks like some spread tests are failing because there are a handful of `get_build_for()` calls in `snapcraft/parts/lifecycle.py` that assume the `architectures` keyword exists and has been filtered.

snapcraft/models/project.py
lengau: Good to know! I don't think charmcraft supports that so I should probably modify charmcraft.

lengau (approved):
Fantastic work! This was a biggie

tigarmo:
Some of the failures are like this:

```
2024-03-08T16:26:32.6531590Z 2024-03-08 16:26:32 Error executing google:ubuntu-22.04-64:tests/spread/plugins/craft-parts/python-symlinks:bare_provisioned (mar081610-674143) : 
2024-03-08T16:26:32.6533187Z -----
2024-03-08T16:26:32.6533878Z + cd bare-provisioned
2024-03-08T16:26:32.6534676Z + '[' -f expected-symlink ']'
2024-03-08T16:26:32.6535450Z + snapcraft prime
2024-03-08T16:26:32.6536663Z Bad snapcraft.yaml content:
2024-03-08T16:26:32.6538192Z - 'architectures' keyword is not supported for base 'bare'. Use 'platforms' keyword instead.
```

Looks like this validation should a) look at the `build-base` (probably use the `get_effective_base()` function if it's not) and b) only fail if the build-base is `core24` (or `devel` I guess), instead of only accepting `core22`

cmatsuoka (approved):
After a long discussion we understood the issues and the platforms validation pattern used in this PR. Although it works, there is redundant and/or sub-optimal model validation that also exists in Rockcraft. These can be addressed in follow-up PRs for both Snapcraft and Rockcraft.

snapcraft/application.py
```diff
@@ -22,87 +22,130 @@
...
     project_type: str | None = pydantic.Field(default=None, alias="type")
 
-    @pydantic.validator("architectures", always=True)
-    def _validate_architecture_data(  # pylint: disable=no-self-argument
-        cls, architectures: list[str | Architecture]
-    ) -> list[Architecture]:
-        """Validate architecture data.
+    @pydantic.validator("platforms")
+    @classmethod
+    def _validate_all_platforms(cls, platforms: dict[str, Any]) -> dict[str, Any]:
```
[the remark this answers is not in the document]
mr-cal: https://github.com/canonical/rockcraft/issues/502
https://github.com/canonical/snapcraft/issues/4650

mr-cal:
For the test failures for google:ubuntu-24.04-64:tests/spread/core24/grammar:arm64:

I think we need something like this in craft-application: https://github.com/canonical/craft-application/compare/main...work/grammar-build-for .  Maybe this should follow the same path as https://github.com/canonical/craft-application/pull/259 now that there is a `self._build_plan` to use.

But even after that patch, `snapcraft clean --destructive-mode` fails with `Multiple builds match the current platform`, which I also think is a craft-application quandary.

Otherwise, the spread tests look good.

tigarmo:
@mr-cal I think the problem with the `snapcraft clean --destructive-mode` is again the [assumption that build-for is provided](https://github.com/canonical/craft-application/blob/c70425d9d3b0afc23c336a26eb91a547c0567817/craft_application/application.py#L554) which we need to fix upstream in craft-application
my suggestion is to ignore that particular failure for this PR

tigarmo (approved):

mr-cal:
Sounds good, I'll make an issue after 

> @mr-cal I think the problem with the `snapcraft clean --destructive-mode` is again the [assumption that build-for is provided](https://github.com/canonical/craft-application/blob/c70425d9d3b0afc23c336a26eb91a547c0567817/craft_application/application.py#L554) which we need to fix upstream in craft-application my suggestion is to ignore that particular failure for this PR

New issue in craft-application: https://github.com/canonical/craft-application/issues/261

t4saha:
@mr-cal Just migrated my app to core24. Platforms is declared only as amd64 and arm64 but remote build is trying to build for all platforms. 
https://github.com/t4saha/stellarium-snap
Screenshot_20240621-103608

sergiusens:
@t4saha thanks for raising this, do you mind creating a new issue to track it and get it addressed ASAP?

sergiusens:
@t4saha I created #4867 to track it in case you want to follow

Issue #4610, opened 2024-02-23
[core24] Non-core24 runs in managed mode are not being correctly logged


tigarmo:
### Bug Description

The fallback logic in `feature/craft-application` has an issue with handling of logs in managed (lxd/etc) instances. The flow is this - let's say that the command is "snapcraft pull".

1) "Snapcraft-1" starts up on the host;
2) Realizes that the project has a non-core24 base, and falls back;
3) core22/core20 code picks up the execution, prepares the lxd instance;
4) launches Snapcraft in the instance with `SNAPCRAFT_MANAGED_MODE` set;
5) "Snapcraft-2" starts up inside the instance;
6) Like in 2, falls back to non-core24 code. However, **before this**, the Emitter is initialized - Snapcraft-2 thinks it's running on host because it's looking at `CRAFT_MANAGED_MODE` and not at `SNAPCRAFT_MANAGED_MODE`;
7) core22/core20 code in Snapcraft-2 picks up the execution, creates the lifecycle, runs "pull", finishes;
8) core22/core20 code **in Snapcraft-1** tries to read the instance logs, but fails because Snapcraft-2 put it in `/root/.local/state/snapcraft/logs` because it thought it was running on host.

### To Reproduce

Run a lifecycle command on any core22/core20 project using a Snapcraft snap from `feature/craft-application`, then look at the generated logs. The instance logs are not present.

### Environment

Jammy

### snapcraft.yaml

```shell
name: simple-snap
version: "1.0"
summary: Build a simple confined snap
description: |
  Build a simple confined snap to test the build process.

base: core22
confinement: strict

parts:
  hello:
    plugin: nil
```

### Relevant log output

```shell
# ... removed the log before this
2024-02-23 14:52:19.913 Executing in container: lxc --project snapcraft exec local:snapcraft-hello-try-on-amd64-for-amd64-17455838 -- env PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin SNAPCRAFT_MANAGED_MODE=1 chown root:root /root/.bashrc
2024-02-23 14:52:20.022 Emitter: Pausing control of the terminal
2024-02-23 14:52:28.830 Emitter: Resuming control of the terminal
[...]

PR #4638, opened 2024-03-06
fix(application): use SNAPCRAFT_MANAGED_MODE everywhere

mattculler:
`craft-application` uses `CRAFT_MANAGED_MODE` environment variable, while `snapcraft` was still using `SNAPCRAFT_MANAGED_MODE`.  This was causing inner managed instances of snapcraft to not realize they were running in managed mode, and write their logs into an ephemeral location inside the LXD instance, rendering those logs inaccessible from the host.

Note that while this fixes the logging issue, this does not fix the spread tests that @tigarmo was hoping it would.

-----
(CRAFT-2539)

### DIFF
--- a/snapcraft/providers.py
+++ b/snapcraft/providers.py
@@ -223,7 +223,7 @@ def get_command_environment(
     :return: Dictionary of environmental variables.
     """
     env = bases.buildd.default_command_environment()
-    env["SNAPCRAFT_MANAGED_MODE"] = "1"
+    env["CRAFT_MANAGED_MODE"] = "1"
 
     # Pass-through host environment that target may need.
     for env_key in [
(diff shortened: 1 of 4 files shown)


mr-cal (commented):
What did this fix?  

I still see the `snapcraft_legacy.internal.build_providers.errors.ProviderFileCopyError: An error occurred with the instance when trying to copy files with 'LXD': Not Found.` errors in the spread tests.  

I thought that those spread tests passing (or at least failing for a different reason) was the acceptance criteria.

tigarmo:
This first commit does fix issue #4610 (I checked), but issue 4610 is not the cause of those spread failures as we thought it was.

tigarmo (approved):
Whoever merge this, please improve the summary & message

mr-cal (approved):
I see, thanks for the explanation

PR #4642, opened 2024-03-06
fix(application): fallback earlier for managed core20

tigarmo:

-----

### DIFF
--- a/snapcraft/application.py
+++ b/snapcraft/application.py
@@ -382,6 +382,7 @@ def main() -> int:
     if os.getenv("SNAPCRAFT_BUILD_ENVIRONMENT") == "managed-host":
         snapcraft.ProjectOptions = snapcraft_legacy.ProjectOptions  # type: ignore
         legacy.legacy_run()
+        return 0  # never called in normal operation
 
     # set lib loggers to debug level so that all messages are sent to Emitter
     for lib_name in _LIB_NAMES:
--- a/snapcraft/application.py
+++ b/snapcraft/application.py
@@ -18,6 +18,7 @@
 
 from __future__ import annotations
 
+import logging
 import os
 import pathlib
 import signal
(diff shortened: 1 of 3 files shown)


snapcraft/application.py
```diff
@@ -373,6 +379,16 @@ def create_app() -> Snapcraft:
 
 def main() -> int:
     """Run craft-application based snapcraft with classic fallback."""
+    if os.getenv("SNAPCRAFT_BUILD_ENVIRONMENT") == "managed-host":
+        snapcraft.ProjectOptions = snapcraft_legacy.ProjectOptions  # type: ignore
+        legacy.legacy_run()
```
tigarmo, line 384: @mattculler we need to cover these lines with an unit test. Here's how I would do it:
- Create a test that mocks `legacy.legacy_run()` and `create_app()`
- Set `SNAPCRAFT_BUILD_ENVIRONMENT` to `managed-host` 
- call this `main()` function
- check that the `legacy_run()` mock *was* called, and the `create_app()` mock was *not* called

mattculler: Will do, appreciate the suggestion

mattculler: Hey @tigarmo so looking at the code, why would you expect `create_app` to not be called?  The only way I can see is if `legacy_run` is expected to call `sys.exit` or raise an exception.  If that's the case, then testing for the lack of a `create_app` call won't be useful, since `legacy_run` is mocked.

tigarmo: oh you're right
in that case, you can add a `return` after that `legacy.legacy_run()` call to emphasize the fact that the execution doesn't progress beyond this

mattculler: Ok makes sense, thanks!

mattculler: https://github.com/canonical/snapcraft/pull/4642/commits/8750f211f84f45a2b0818252206a79d63623fd32

tests/unit/test_application.py
```diff
@@ -274,3 +275,18 @@ def test_application_build_with_extensions(monkeypatch, extension_source, new_di
...
+
+
+def test_application_managed_core20_fallback(monkeypatch, new_dir, mocker):
+    monkeypatch.setenv("CRAFT_DEBUG", "1")
+    monkeypatch.setenv("SNAPCRAFT_BUILD_ENVIRONMENT", "managed-host")
+
+    (new_dir / "snap").mkdir()
+
+    mock_legacy_run = mocker.patch("snapcraft_legacy.cli.legacy.legacy_run")
+    application.create_app = Mock()
```
tigarmo, line 287: you need to use `mocker` here too, otherwise this mock is going to outlive the test as these module objects are global

something like
```python
mock_create_app = mocker.patch.object(application, "create_app")

tigarmo: ^pretty sure that's why the tests are failing actually

mattculler: Ah ok, that's unexpected.  Will fix now.

tigarmo:
@mattculler please approve this on my behalf, I can't approve it as it was originally created by me
nice works ;)

mattculler:
> @mattculler please approve this on my behalf, I can't approve it as it was originally created by me nice works ;)

Thanks!

mattculler (approved):

sergiusens (approved):

PR #4652, opened 2024-03-09
feat: adopt more fields from the appstream metadata file(part 2) 

soumyaDghosh:

-----

This is the Part 2 of PR #4613. This adds a new method named `_update_project_links` which is called from the `update_project_metadata` function. It checks if the `project` has that particular metadata and if yes, then it appends the value into it, and else it casts the value onto it from the `extracted_metadata`

### DIFF
--- a/snapcraft/meta/appstream.py
+++ b/snapcraft/meta/appstream.py
@@ -182,7 +182,7 @@ def _get_urls_from_xml_element(nodes, url_type) -> Optional[List[str]]:
         if (
             node is not None
             and node.attrib["type"] == url_type
-            and node.text not in urls
+            and node.text.strip() not in urls
         ):
             urls.append(node.text.strip())
     if urls:
--- a/snapcraft/meta/appstream.py
+++ b/snapcraft/meta/appstream.py
@@ -101,12 +101,13 @@ def extract(relpath: str, *, workdir: str) -> Optional[ExtractedMetadata]:
     title = _get_value_from_xml_element(dom, "name")
     version = _get_latest_release_from_nodes(dom.findall("releases/release"))
     project_license = _get_value_from_xml_element(dom, "project_license")
-    contact = _get_value_from_xml_element(dom, "update_contact")
+    update_contact = _get_value_from_xml_element(dom, "update_contact")
+    contact = [update_contact] if update_contact else None
 
     issues = _get_urls_from_xml_element(dom.findall("url"), "bugtracker")
     donation = _get_urls_from_xml_element(dom.findall("url"), "donation")
     website = _get_urls_from_xml_element(dom.findall("url"), "homepage")
-    source_code = _get_source_code_from_xml_element(dom.findall("url"))
+    source_code = _get_urls_from_xml_element(dom.findall("url"), "vcs-browser")
 
     desktop_file_paths = []
     desktop_file_ids = _get_desktop_file_ids_from_nodes(dom.findall("launchable"))
@@ -178,17 +179,14 @@ def _get_value_from_xml_element(tree, key) -> Optional[str]:
 def _get_urls_from_xml_element(nodes, url_type) -> Optional[List[str]]:
     urls = []  # type: List[str]
     for node in nodes:
-        if node is not None and node.attrib["type"] == url_type:
+        if (
+            node is not None
+            and node.attrib["type"] == url_type
+            and node.text not in urls
+        ):
             urls.append(node.text.strip())
     if urls:
         return urls
-    return None
-
-
-def _get_source_code_from_xml_element(nodes) -> Optional[str]:
-    for node in nodes:
-        if node is not None and node.attrib["type"] == "vcs-browser":
-            return node.text.strip()
     return None
 
 
--- a/snapcraft/models/project.py
+++ b/snapcraft/models/project.py
@@ -608,7 +608,6 @@ class Project(models.Project):
     donation: Optional[UniqueStrList]
     # snapcraft's `source_code` is more general than craft-application
     source_code: Optional[UniqueStrList]
-    license: Optional[str]  # type: ignore[assignment]
     contact: Optional[UniqueStrList]
     issues: Optional[UniqueStrList]
     website: Optional[UniqueStrList]
@@ -860,6 +859,13 @@ class Project(models.Project):
             )
 
         return provenance
+
+    @pydantic.validator("contact", "donation", "issues", "source_code", "website", pre=True)
+    @classmethod
+    def _validate_contact(cls, field_value):
+        if isinstance(field_value, str):
+            field_value = cast(UniqueStrList, [field_value])
+        return field_value
 
     @classmethod
     def unmarshal(cls, data: Dict[str, Any]) -> "Project":
--- a/snapcraft/models/project.py
+++ b/snapcraft/models/project.py
@@ -606,7 +606,10 @@ class Project(models.Project):
     donation: Optional[Union[str, UniqueStrList]]
     # snapcraft's `source_code` is more general than craft-application
     source_code: Optional[str]  # type: ignore[assignment]
-    website: Optional[str]
+    license: Optional[str]
+    contact: Optional[str]
+    issues: Optional[Union[str, UniqueStrList]]
+    website: Optional[Union[str, UniqueStrList]]
     type: Optional[Literal["app", "base", "gadget", "kernel", "snapd"]]
     icon: Optional[str]
     confinement: Literal["classic", "devmode", "strict"]
--- a/snapcraft/models/project.py
+++ b/snapcraft/models/project.py
@@ -602,14 +602,12 @@ class Project(models.Project):
     name: ProjectName  # type: ignore[assignment]
     build_base: Optional[str]
     compression: Literal["lzo", "xz"] = "xz"
-    # TODO: ensure we have a test for version being retrieved using adopt-info
-    # snapcraft's `version` is more general than craft-application
-    version: Optional[ProjectVersion]  # type: ignore[assignment]
+    version: Optional[VersionStr]  # type: ignore[assignment]
     donation: Optional[UniqueStrList]
     # snapcraft's `source_code` is more general than craft-application
-    source_code: Optional[UniqueStrList]
-    contact: Optional[UniqueStrList]
-    issues: Optional[UniqueStrList]
+    source_code: Optional[UniqueStrList]  # type: ignore[assignment]
+    contact: Optional[UniqueStrList]  # type: ignore[assignment]
+    issues: Optional[UniqueStrList]  # type: ignore[assignment]
     website: Optional[UniqueStrList]
     type: Optional[Literal["app", "base", "gadget", "kernel", "snapd"]]
     icon: Optional[str]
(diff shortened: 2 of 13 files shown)


snapcraft/meta/appstream.py
```diff
@@ -177,20 +178,17 @@ def _get_value_from_xml_element(tree, key) -> Optional[str]:
 def _get_urls_from_xml_element(nodes, url_type) -> Optional[List[str]]:
     urls = []  # type: List[str]
     for node in nodes:
-        if node is not None and node.attrib["type"] == url_type:
+        if (
+            node is not None
+            and node.attrib["type"] == url_type
+            and node.text not in urls
```
mr-cal, line 184:
```suggestion
            and node.text.strip() not in urls
```

snapcraft/models/project.py
```diff
@@ -417,10 +417,13 @@ class Project(models.Project):
     # TODO: ensure we have a test for version being retrieved using adopt-info
     # snapcraft's `version` is more general than craft-application
     version: Optional[ProjectVersion]  # type: ignore[assignment]
-    donation: Optional[Union[str, UniqueStrList]]
+    donation: Optional[UniqueStrList]  # type: ignore[assignment]
     # snapcraft's `source_code` is more general than craft-application
-    source_code: Optional[str]  # type: ignore[assignment]
-    website: Optional[str]
+    source_code: Optional[UniqueStrList]  # type: ignore[assignment]
+    license: Optional[str]  # type: ignore[assignment]
```
mr-cal, line 423: Is the required?  I think this is the same as the parent model (https://github.com/canonical/craft-application/blob/c5792385aea32ef6ab811d65c51934fe2e40df80/craft_application/models/project.py#L92) and therefore can be dropped.

snapcraft/models/project.py
```diff
@@ -417,10 +417,13 @@ class Project(models.Project):
     # TODO: ensure we have a test for version being retrieved using adopt-info
     # snapcraft's `version` is more general than craft-application
     version: Optional[ProjectVersion]  # type: ignore[assignment]
-    donation: Optional[Union[str, UniqueStrList]]
+    donation: Optional[UniqueStrList]  # type: ignore[assignment]
```
mr-cal, line 420: Could these changes break some existing `snapcraft.yaml`s?

mr-cal (commented):
Overall the changes look good.  My biggest concern is making changes to the snapcraft.yaml that could break existing builds.

In similar situations, we have made fields that accept strings and lists of strings and added a pydantic validator that converts the string to a list.  Here's a recent example: https://github.com/canonical/snapcraft/blob/f4992faff74cd05eac2b8855974aa6f2da2de312/snapcraft/models/project.py#L514

snapcraft/models/project.py
soumyaDghosh: Isn't `Optional[str]` and `str | none` the same thing?

snapcraft/models/project.py
soumyaDghosh: I am trying to look into the tests to add some more tests with currently used styles, but I am hoping it'll not break.

soumyaDghosh: I am trying to add a test, that'll take a local `snapcraft.yaml`, then process it and extract the snap.yaml from it. Is it possible?

mr-cal (commented):
Thanks for showing up today @soumyaDghosh!

Here are notes from the Starcraft clinic:

1. The links in the Project model are supposed to be there (they were unintentionally removed)
2. `# type: ignore[assignment]` can probably be removed from the `str` and `UniqueStrList` fields
3. We need to keep the same types in the project model for backward compatibility
    - This may require moving the `normalize_values` function from the Metadata model to the Project model
4. `license: Optional[str]` can probably be dropped from the Project model (if not, let me know)

mr-cal:
Pre-validator example for strings and lists of strings: https://github.com/canonical/snapcraft/blob/07b4ee15c8e58b93d32b4ed9331b1ecec7880e01/snapcraft/models/project.py#L514

soumyaDghosh:
@mr-cal I have kept the data types for `links` as `Optional[UniqueStrList]` but added a pydantic validator as you said and it now normalises the values if given as a string to list. Also added two test with website and contact, where value is given as string but output is a list.

tests/unit/models/test_projects.py
```diff
@@ -56,6 +56,8 @@ def _project_yaml_data(
             "grade": "stable",
             "confinement": "strict",
             "parts": {},
+            "contact": "hello@world",
```
mr-cal, line 59: I would drop these from this yaml because this fixture contains the bare minimum fields for a project and is used to test default values.

Instead, can you add new tests?

It looks like this is implicitly tested in `tests/unit/parts/test_update_metadata.py` but it's good practice to test it explicitly here. I'm thinking something like this:

```python
class TestProjectValidation:
    """Validate top-level project items."""

    ...

    def test_links_scalar(self, project_yaml_data):
        data = project_yaml_data(
            contact="hello@world",
            # set other links...
        )
        project = Project.unmarshal(data)
        assert project.contact == ["hello@world"]
        # assert other links
        
    def test_links_list(self, project_yaml_data):
        data = project_yaml_data(
            contact=["hello@world"],
            # set other links...
        )
        project = Project.unmarshal(data)
        assert project.contact == ["hello@world"]
        # assert other links
```

tests/unit/parts/test_update_metadata.py
```diff
@@ -71,6 +71,12 @@ def yaml_data(extra_args: Dict[str, Any]):
             "base": "core22",
             "grade": "stable",
             "confinement": "strict",
+            "license": "license",
+            "contact": {"contact1"},
```
mr-cal, line 75: Should these be lists?
```suggestion
            "contact": ["contact1"],
```

mr-cal (changes requested):
Looking great!  The pydantic validator and the types look good, so there should be no breaking changes for existing snaps.

snapcraft/models/project.py
```diff
@@ -806,6 +808,15 @@ def _validate_provenance(cls, provenance):
 
         return provenance
 
+    @pydantic.validator(
+        "contact", "donation", "issues", "source_code", "website", pre=True
+    )
+    @classmethod
+    def _validate_contact(cls, field_value):
```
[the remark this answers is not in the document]
soumyaDghosh: I was testing this thing and forgot to change the name. Sorry for that. :-(

tests/unit/parts/test_update_metadata.py
```diff
@@ -121,6 +133,11 @@ def test_update_project_metadata(project_yaml_data, appstream_file, new_dir):
     assert project.summary == "summary"  # already set in project
     assert project.description == "description"  # already set in project
     assert project.version == "0.1"  # already set in project
+    assert project.contact == ["contact1", "contact2"]
```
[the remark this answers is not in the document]
soumyaDghosh: > I may have missed it, but I don't see a test that checks that a duplicate URL in the `snapcraft.yaml` and in the metadata are de-duplicated by this line `project_list.extend(set(getattr(metadata, field)) - set(project_list))`.

It's done once here,

https://github.com/soumyaDghosh/snapcraft/blob/main/tests/unit/parts/test_update_metadata.py#L243

But, I'll try to add more.

tests/unit/parts/test_update_metadata.py
soumyaDghosh: They can be either simple strings or list, anything due to the added pydentic validator.

mr-cal (changes requested):
This is looking great!

Last thing is to fix the [spread](https://github.com/canonical/snapcraft/actions/runs/8936358304/job/24546700259?pr=4652#step:4:361) tests:
- google:ubuntu-24.04-64:tests/spread/core24/appstream-desktop:desktop
- google:ubuntu-24.04-64:tests/spread/core24/appstream-desktop:icon_desktop

It looks like you'll need to add the `license` field to the two `expected_snap.yaml` in `tests/spread/core24/appstream-desktop/`

soumyaDghosh:
@mr-cal why aren't the link showing up here? Do you have any guess? Technically, that should also have been adopted right?

mr-cal:
> @mr-cal why aren't the link showing up here? Do you have any guess? Technically, that should also have been adopted right? https://github.com/canonical/snapcraft/actions/runs/9005523750/job/24741235250?pr=4652#step:4:294

The test `google:ubuntu-24.04-64:tests/spread/core24/appstream-desktop:icon_desktop` should be reading collecting metadata from [`tests/spread/core24/appstream-desktop/icon-desktop/io.snapcraft.appstream.metainfo.xml`](https://github.com/canonical/snapcraft/blob/707385a0d971c4838b2631a6630d24b7cd08eb08/tests/spread/core24/appstream-desktop/icon-desktop/io.snapcraft.appstream.metainfo.xml#L1), right?  If so, I don't think there are links in that metadata to adopt.

soumyaDghosh:
> If so, I don't think there are links in that metadata to adopt.

https://github.com/canonical/snapcraft/blob/707385a0d971c4838b2631a6630d24b7cd08eb08/tests/spread/core24/appstream-desktop/icon-desktop/io.snapcraft.appstream.metainfo.xml#L33

There is this link. This should be listed as `website` under `links` section.

soumyaDghosh:
@mr-cal I just built musicpod from source, using the snapcraft snap that I got from the spread tests and the links are showing up.

image

So, I am hopeful that it works. License also worked. So, this is now at a working stage.

mr-cal:
Ah, yes don't worry about the chisel test, we just fixed it this morning and it hasn't landed on `main` yet. (https://github.com/canonical/snapcraft/pull/4795)

I'll take another look at the spread test tomorrow and try to figure out what's happening

soumyaDghosh:
> I'll take another look at the spread test tomorrow and try to figure out what's happening

https://github.com/canonical/snapcraft/actions/runs/9007288288/job/24746927374?pr=4652#step:4:582

The link is showing up in core22 but not in core24?

soumyaDghosh:
Hi @mr-cal I tested this locally with https://github.com/NickvisionApps/Application, and links aren't showing up there. Is there something different in `core24`?

mr-cal:
> Hi @mr-cal I tested this locally with https://github.com/NickvisionApps/Application, and links aren't showing up there. Is there something different in `core24`?

Yikes, you're right!  I don't think snapcraft is calling [`update_project_metadata()`](https://github.com/canonical/snapcraft/blob/9c90f0b3799d800434a1b4c8b531067303e449fb/snapcraft/parts/update_metadata.py#L33) for `core24`.  Since it gets called [here](https://github.com/canonical/snapcraft/blob/9c90f0b3799d800434a1b4c8b531067303e449fb/snapcraft/parts/lifecycle.py#L346) for `core22`, it should probably be called [here](https://github.com/canonical/snapcraft/blob/9c90f0b3799d800434a1b4c8b531067303e449fb/snapcraft/services/package.py#L161)) for `core24`. 

We can still land this PR, but core24 support should be done in a separate PR.  You are welcome to make the PR, otherwise I will create an issue for it so we don't lose track of it.

mr-cal:
FYI we discussed this during the starcraft clinic and found a better solution - moving `_update_project_links()` so it is called by both core22 and core24.

mr-cal:
> Sorry, was not able to attend till the end. So, what approach do you propose. BTW, the tests seem to be passing

This is a summary of what we discussed together!  

I'm just recording it here for myself in the future because I forget things easily :)

soumyaDghosh:
@mr-cal It passed!!! :tada:

mr-cal (approved):
Great work!

lengau (approved):

PR #4661, opened 2024-03-13
fix(test): use prime dir in some tests

syu-w:
Add prime dir in some tests.

PR #4662, opened 2024-03-13
fix(test): do not create 2 snapcraft.yaml and confuse location

syu-w:
There are 2 bugs:

First, the current fixture `snapcraft_yaml` may create 2 `snapcraft.yaml` in `/` and `/snap/` due to it used in different places and it is default to `/snap/snapcraft.yaml`. 

When both `snapcraft.yaml` exists, the `/snapcraft.yaml` is used without any error / warning.

The second bug is when `/snap/snapcraft.yaml` exists,  `/build-aux/snap` will be ignored even if it had hooks in it. This is probably expected behavior since these are two different versions of the structure.

Needs #4661

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -434,9 +434,7 @@ def default_build_plan():
 
 
 @pytest.fixture()
-def lifecycle_service(
-    default_project, default_factory, default_build_plan, snapcraft_yaml, tmp_path
-):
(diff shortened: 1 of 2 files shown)


syu-w:
> can you add some more explanation to this change?

Updated

tigarmo (approved):

sergiusens (approved):

PR #4663, opened 2024-03-13
fix: use "devel" bases on build plan

tigarmo:
For the BuildInfos we want to use "devel" bases, as that is the term that craft-providers expects to create instances based on the development track. We still need to use the regular "base" (e.g. "core24") in other places, so add a customization parameter to get_effective_base() to use either behavior.

-----

### DIFF
--- a/snapcraft/models/project.py
+++ b/snapcraft/models/project.py
@@ -1139,6 +1139,7 @@ class SnapcraftBuildPlanner(models.BuildPlanner):
                     build_base=self.build_base,
                     project_type=self.project_type,
                     name=self.name,
+                    translate_devel=False,  # We want actual "devel" if set.
                 )
             )
         ].value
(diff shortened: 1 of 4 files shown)


tigarmo:
As far as I can tell the spread failures are all the unrelated (expected) ones.

syu-w:
So the `SnapcraftBuildPlanner` and `Project` will have different `get_build_plan` results?

tigarmo:
> So the `SnapcraftBuildPlanner` and `Project` will have different `get_build_plan` results?

oh good call, I wasn't aware that the `Project` still had a `get_build_plan()` method. It shouldn't, so I'm going to remove it and see if anything breaks.

sergiusens (approved):

tigarmo:
@sergiusens please squash-merge this

PR #4664, opened 2024-03-13
feat(craft-application): support remote build

syu-w:
This also support --platform and --build-for.


snapcraft/commands/remote.py
```diff
@@ -0,0 +1,293 @@
...
+        build_id = get_build_id(self._app.name, project.name, project_dir)
+        if parsed_args.recover:
+            emit.progress(f"Recovering build {build_id}")
+            builds = builder.resume_builds(build_id)
+        else:
+            emit.progress(
+                "Starting new build. It may take a while to upload large projects."
+            )
+            try:
+                builds = builder.start_builds(project_dir, architectures=architectures)
```
lengau, line 210: Other than the Launchpad side (Launchpad doesn't currently support limiting the architectures of a charm build), can you think of any reason not to include this in Charmcraft?

(Overall goal: move this class to `craft-application` and provide application-specific override hooks)

tests/unit/commands/test_remote.py
```diff
@@ -122,6 +153,26 @@ def mock_run_legacy(mocker):
 #############
 
 
+@pytest.mark.parametrize("base", CURRENT_BASES - {"core22"})
```
lengau, line 156: Why are we excluding core22 from this?

tests/unit/commands/test_remote.py
lengau, line 1: How long does this file take to run on your machine? If it's quick I don't see too much reason to filter these tests further, but if it takes a while we might want to consider whether we can drop some of the tests.

lengau (commented):
Looks good, thanks! I've got a few questions, but the only likely change request I can think of is that we might want to reduce the unit tests if they take too long.

tests/unit/commands/test_remote.py
syu-w: 280 passed, 4 xfailed, 7 warnings in 8.14s

tests/unit/commands/test_remote.py
syu-w: Because core22 will still use the old remote build, at least for now.

snapcraft/commands/remote.py
syu-w: No, I do not see any reason other than the parameters are different.

snapcraft/services/remotebuild.py
```diff
@@ -0,0 +1,78 @@
...
+        project_name = self._name.split("-", maxsplit=2)[1]
+        logs: dict[str, Path | None] = {}
+        log_downloads: dict[str, Path] = {}
+        fetch_time = datetime.now().isoformat(timespec="seconds")
+        for build in self._builds:
+            url = build.build_log_url
+            if not url:
+                logs[build.arch_tag] = None
+                continue
+            filename = (
```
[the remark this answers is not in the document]
syu-w: It simple because charmcraft did this when I used it as reference.

snapcraft/services/remotebuild.py
```diff
@@ -0,0 +1,78 @@
...
+from craft_application.services import remotebuild
+from overrides import override
+
+
+class RemoteBuild(remotebuild.RemoteBuildService):
+    """Snapcraft remote build service."""
+
+    RecipeClass = launchpad.models.SnapRecipe
+
+    def fetch_logs(self, output_dir: Path) -> Mapping[str, Path | None]:
```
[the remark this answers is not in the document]
syu-w: Or we can use the upstream one?

lengau (approved):
Looks great, and thanks for answering my questions!

cmatsuoka (approved):
This closely follows the existing implementation for Charmcraft. LGTM!

PR #4666, opened 2024-03-14
feat(plugins): support core24 in the Python plugin

tigarmo:
The interpreter for confined core24 snaps is /usr/bin/python3.12. This also includes some minor rework of the logic: if the snap's confinement is strict or devmode, then we should definitely know which interpreter to use. This can make our lives easier in the future as we support more bases.

-----


tests/spread/core24/python-hello/classic/snap/snapcraft.yaml
```diff
@@ -0,0 +1,19 @@
...
+description: build a python application using core24
+base: core24
+build-base: devel
+confinement: classic
+
+apps:
+  python-hello-classic:
+    command: bin/hello
+    environment:
+      # Note that since commit 9b6e58337e0 we need to explicitly prepend $SNAP/bin
```
tigarmo, line 13: This bit here is for when Sergio finally lets me fix classic Python snaps

sergiusens: I would actually change the command to `bin/python3 bin/hello`, but let's fix the issue at the core if you want :-), PATH in environment could lead to issues

tests/spread/core24/python-hello/task.yaml
```diff
@@ -0,0 +1,20 @@
...
+  snapcraft clean
+  rm -f ./*.snap
+  snap remove python-hello-"${CONFINEMENT}"
+
+execute: |
+  cp -r src ./"${CONFINEMENT}"/
+  cd ./"${CONFINEMENT}"
+
+  snapcraft
+  snap install python-hello-"${CONFINEMENT}"_1.0_*.snap --dangerous --classic
```
sergiusens, line 18: --classic should really be on the case of $CONFINEMENT

sergiusens (commented):
a couple of comments, looks good, would be interesting to think about a more generic variant test task, `applications` perhaps? To consider systems that use extensions and plugins in the mix

tests/spread/core24/python-hello/task.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+environment:
+  CONFINEMENT/strict: strict
+  CONFINEMENT/classic: classic
+
+restore: |
+  cd ./"${CONFINEMENT}"
+  snapcraft clean
+  rm -f ./*.snap
+  snap remove python-hello-"${CONFINEMENT}"
```
sergiusens, line 11: this is not needed due to https://github.com/canonical/snapcraft/blob/feature/craft-application/spread.yaml#L225 which calls https://github.com/canonical/snapcraft/blob/057e21b7a20bf698b7d94a9fd9ed65e00919254c/tests/spread/tools/restore.sh

sergiusens:
should we address having https://github.com/canonical/snapcraft/blob/feature/craft-application/tests/spread/core24-suites/patchelf/classic-python/task.yaml at all? Perhaps reintroduce when we move the docs to rtd?

tests/spread/core24/python-hello/classic/snap/snapcraft.yaml
tigarmo: updated this to use `bin/python3`, updated the note
another motivation to fix this is that a lot of users "fix" this issue by updating the PATH, which I agree is dangerous

tests/spread/core24/python-hello/task.yaml
tigarmo: updated this

sergiusens:
@tigarmo just remove it, it is a flaky out of tree test, should be the other way around, test in our code base, documented through our sphinx infra

sergiusens (approved):
thanks
would be nice to have better names for PartInfo's project_base and base

snapcraft/parts/plugins/python_plugin.py
```diff
@@ -19,16 +19,16 @@
...
 logger = logging.getLogger(__name__)
 
 
-_RUNTIME_PYTHON_PATH = {
-    ("core22", "strict"): "/usr/bin/python3.10",
-    ("core22", "devmode"): "/usr/bin/python3.10",
-    ("core22", "classic"): None,  # classic snaps must provision python
+_CONFINED_PYTHON_PATH = {
+    "core22": "/usr/bin/python3.10",
+    "core24": "/usr/bin/python3.12",
```
cmatsuoka, line 31: Nitpick: update the year in the copyright notice line

cmatsuoka (approved):

PR #4667, opened 2024-03-14
feat(project): add get_archs_list for Architecture model

syu-w:
`Architecture.get_archs_list` provides a quick way to get a list of all required build targets

### DIFF
--- a/snapcraft/models/project.py
+++ b/snapcraft/models/project.py
@@ -29,6 +29,7 @@ from typing import (
     Mapping,
     Optional,
     Sequence,
+    Set,
     Tuple,
     Union,
     cast,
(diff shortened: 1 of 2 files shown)


snapcraft/models/project.py
```diff
@@ -425,6 +426,33 @@ class Architecture(models.CraftBaseModel, extra=pydantic.Extra.forbid):
...
+                archs_set.add(arch)
+
+        if not archs_set:
+            if isinstance(self.build_on, str):
+                archs_set.add(self.build_on)
+            elif isinstance(self.build_on, list):
+                for arch in self.build_on:
+                    archs_set.add(arch)
+
+        return list(archs_set)
```
lengau, line 454: It's probably better either to return the set here or to provide a reliable order by doing something like `sorted(archs_set)`

syu-w:
Closed as `Architecture` no longer used in core24

PR #4672, opened 2024-03-15
Enable Snapcraft plugins in craft-application codepath

tigarmo:
Two commits: one registers the plugins in the craft-application codepath, and one adds the confinement and base variables to the lifecycle (so that plugins can use it).

### DIFF
--- a/tests/unit/test_application.py
+++ b/tests/unit/test_application.py
@@ -236,3 +236,12 @@ def test_parse_info_integrated(monkeypatch, mocker, new_dir):
     assert snap_yaml["summary"] == "Sample summary"
     assert snap_yaml["description"] == "Sample description"
     assert snap_yaml["version"] == "1.2.3"
+
+
+def test_application_plugins():
+    app = application.create_app()
+    plugins = app._get_app_plugins()
+
+    # Just do some sanity checks.
+    assert "python" in plugins
+    assert "kernel" not in plugins
--- a/snapcraft/application.py
+++ b/snapcraft/application.py
@@ -29,6 +29,7 @@ import craft_application.commands as craft_app_commands
 import craft_cli
 from craft_application import Application, AppMetadata, util
 from craft_cli import emit
+from craft_parts.plugins.plugins import PluginType
 from overrides import override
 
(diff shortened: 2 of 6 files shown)


tigarmo:
This doesn't fix any spread tests but introduces a new, correct failure for tests/spread/core24/snap-creation:package_cutoff:

```
2024-03-15T18:57:19.2265172Z :: Python interpreter not found in payload.
2024-03-15T18:57:19.2266481Z :: + symlink_target=
2024-03-15T18:57:19.2267539Z :: + '[' -z '' ']'
2024-03-15T18:57:19.2268556Z :: + echo 'No suitable Python interpreter found, giving up.'
2024-03-15T18:57:19.2269497Z :: No suitable Python interpreter found, giving up.
2024-03-15T18:57:19.2270310Z :: + exit 1

**Review by sergiusens — APPROVED** [review]

File: tests/unit/test_application.py [hunk 1]
```diff
@@ -236,3 +236,12 @@ def test_parse_info_integrated(monkeypatch, mocker, new_dir):
     assert snap_yaml["summary"] == "Sample summary"
     assert snap_yaml["description"] == "Sample description"
     assert snap_yaml["version"] == "1.2.3"
+
+
+def test_application_plugins():
```
cmatsuoka, line 241: Could you add a test verifying if the `list-plugins` command is returning the expected list of plugins for different bases?

cmatsuoka (changes requested):
We should add a test to make sure `list-plugins` is working as expected.

tests/unit/test_application.py
tigarmo: I can't do that because the `list-plugins` command hasn't been ported to core24 yet (I don't know if it's on our radar)

cmatsuoka (approved):

tests/unit/test_application.py
cmatsuoka: `list-plugins` could be an interesting thing to add to craft-application to get all tools covered.

tigarmo:
@sergiusens please **rebase** merge

PR #4679, opened 2024-03-19
refactor(docs): use more docs from craft-parts

tigarmo:
Now that the structure of the shared craft-parts docs has more content,
we can simplify our own docs and just use craft-parts' directly in many
places.

The guideline is this: If the document will be used wholesale, just reference
it directly in the TOC. Otherwise, `include` it in a Snapcraft docs.


docs/requirements.txt
```diff
@@ -5,7 +5,7 @@ craft-application==1.2.1
 craft-archives==1.1.3
 craft-cli==2.5.1
 craft-grammar==1.1.1
-craft-parts==1.27.0
+craft-parts @ git+https://github.com/canonical/craft-parts@CRAFT-2589-improve-docs-bundle
```
tigarmo, line 8: reminder to drop this before merging

syu-w (approved):

mr-cal (approved):

PR #4680, opened 2024-03-19
fix: show better error when project file is missing

syu-w:


tigarmo:
I think this PR might have been from the initial implementation in craft-application? @syu-w is this code still necessary?

syu-w:
@tigarmo it still needed as we need to link the docs.

snapcraft/application.py
```diff
@@ -111,13 +113,23 @@ def _resolve_project_path(self, project_dir: pathlib.Path | None) -> pathlib.Pat
...
-            except FileNotFoundError:
-                return super()._resolve_project_path(project_dir / "build-aux" / "snap")
+
+        with suppress(FileNotFoundError):
+            return super()._resolve_project_path(project_dir)
+
+        with suppress(FileNotFoundError):
+            return super()._resolve_project_path(project_dir / "build-aux" / "snap")
+
+        raise ProjectFileMissingError(
```
tigarmo, line 125: This is tricky because now there's a disagreement between craft-app and snapcraft; in craft-app, this method raises `FileNotFoundError`, and there's code in Snapcraft that [relies on that](https://github.com/canonical/snapcraft/blob/feature/craft-application/snapcraft/application.py#L80) (I'm surprised it didn't break any tests).

snapcraft/application.py
```diff
@@ -111,13 +113,23 @@ def _resolve_project_path(self, project_dir: pathlib.Path | None) -> pathlib.Pat
...
+            f"Project file '{self.app.name+'.yaml'}' not found in '{project_dir}'",
```
tigarmo, line 126: The lack of coverage here tells me that we should have at least one test that checks that this message is correctly raised 
also we can use "snapcraft.yaml" in this case here

tigarmo (changes requested):

snapcraft/application.py
syu-w: That because `ProjectFileMissingError` also inherited `FileNotFoundError`, so it still got caught.

tigarmo: I see, but then isn't the `ProjectFileMissingError` error that craft-app raises in get_project() "swallowing" this one?

syu-w: Oh, it did swallowed this one.

tigarmo: What do you think about overriding get_project(), instead of this _resolve_project_path()? It could look like this:

```python
def get_project(self, ...):
    try:
        return super().get_project(...)
    except ProjectFileMissingError:
        raise ProjectFileMissingError(...) from None

tigarmo (approved):
thanks!

PR #4692, opened 2024-03-22
feat: core24 using Craft Application and core22 components

sergiusens:


sergiusens:
Replaced by #4693 as this branch is protected

PR #4693, opened 2024-03-22
feat: core24 using Craft Application and core22 components

sergiusens:


.shellcheckrc
```diff
@@ -0,0 +1,5 @@
+# NOTE: these disables are only necessary while the core24 spread tests are
```
sergiusens, line 1: do we need to remove this?

syu-w (approved):

mr-cal (approved):

.shellcheckrc
tigarmo: probably, yes. I added this because of all the early `exit` we put in while the features were being ported over

PR #4701, opened 2024-03-27
feat: upload snaps with components

### LINKED ISSUE #4687: Add support for uploading components
### What needs to get done

Convert [this](https://github.com/mr-cal/snapcraft/pull/2) prototype into a PR.

This includes:
- Writing all component metadata to `meta/snap.yaml` except for the version
- Parsing `--component` command line parameter for `snapcraft upload`
- Validating the correct components are going to be uploaded
- Uploading components with a snap

### Why it needs to get done

To support components in Snapcraft

mr-cal:

-----

This PR allows uploading snaps with components to the snap store.

The UX is:
```
snapcraft upload <snap-file> --component <component-name>=<component-file> --component <component-name>=<component-file> ...
```

Fixes #4687 
(CRAFT-2626)


snapcraft/commands/core22/upload.py
```diff
@@ -18,20 +18,46 @@
...
+    """
+
+    name: str | None = None
+    filename: str | None = None
+
+    def __call__(self, value):
+        """Run by argparse to validate and convert the given argument."""
+        parts = [x.strip() for x in value.split("=")]
+        parts = [p for p in parts if p]
+        if len(parts) == 2:
```
cmatsuoka, line 54: Another way to implement this would be `parts = tuple(map(str.strip, value.split("=", 1)))`.

snapcraft/commands/core22/upload.py
```diff
@@ -106,6 +155,54 @@ def run(self, parsed_args):
...
+    :param provided_components: A list of ComponentOptions provided by the user.
+    :param snap_yaml: The snap's metadata.
+    """
+    # get the names of the components provided by the user
+    provided_names = {
+        component.name for component in provided_components if component.name
+    }
+    emit.debug(f"Components provided by the user: {provided_names}")
+
+    if snap_yaml.get("components"):
```
cmatsuoka, line 176: `"components" in snap_yaml` could be marginally better here.

snapcraft/meta/snap_yaml.py
```diff
@@ -558,3 +561,22 @@ def _populate_environment(
         del environment["PATH"]
 
     return environment if environment else None
+
+
+def _process_components(components) -> Dict[str, Any] | None:
```
cmatsuoka, line 566: No type hint for `components`?

tests/unit/commands/test_upload.py
```diff
@@ -71,6 +77,34 @@ def snap_file_with_started_at():
...
+    return str(
+        (
+            pathlib.Path(unit.__file__)
+            / ".."
+            / ".."
+            / "legacy"
+            / "data"
+            / "test-snap-with-component+test-component.comp"
+        ).resolve()
+    )
```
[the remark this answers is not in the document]
mr-cal: Good idea, I was being lazy by following the status quo

snapcraft/store/client.py
```diff
@@ -519,17 +523,22 @@ def verify_upload(
...
         snap_file_size: int,
         built_at: Optional[str],
         channels: Optional[Sequence[str]],
+        components: Optional[Dict[str, str]],
     ) -> int:
         if channels:
             raise errors.SnapcraftError("Releasing during currently unsupported")
+        if components:
+            raise errors.SnapcraftError(
+                "Components support for on-prem stores is currently unsupported"
```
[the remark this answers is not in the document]
mr-cal: Fixed

snapcraft/commands/core22/upload.py
```diff
@@ -106,6 +155,54 @@ def run(self, parsed_args):
...
+            "Use `--component <name>=<filename>`."
+        )
+
+    for component in provided_components:
+        if not component.filename:
+            # should not occur after argparse validation
+            raise RuntimeError(f"Component {component.name} has no filename.")
+
+        component_filepath = pathlib.Path(component.filename)
+        if not component_filepath.exists():
```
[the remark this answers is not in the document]
mr-cal: Fixed

snapcraft/commands/core22/upload.py
```diff
@@ -18,20 +18,46 @@
...
+
+    def __call__(self, value):
+        """Run by argparse to validate and convert the given argument."""
+        parts = [x.strip() for x in value.split("=")]
+        parts = [p for p in parts if p]
+        if len(parts) == 2:
+            name, filename = parts
+            return ComponentOption(name, filename)
+
+        raise ValueError("the `--component` format must be <name>=<filename>")
```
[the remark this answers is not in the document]
mr-cal: Great suggestion, thanks!  I did lift this directly out of Charmcraft since the implementation for charm resources and snap components are similar.

snapcraft/meta/snap_yaml.py
mr-cal: Added

snapcraft/commands/core22/upload.py
mr-cal: Updated

snapcraft/commands/core22/upload.py
mr-cal: It's a good idea, but it doesn't work for some scenarios like a missing component name (`--component "=filename"`).

I did change it to be a one-liner and it should be cleaner.

setup.cfg
```diff
@@ -1,6 +1,6 @@
 [codespell]
 ignore-words-list = buildd,crate,keyserver,comandos,ro,astroid
-skip = waf,*.tar,*.xz,*.zip,*.bz2,*.7z,*.gz,*.deb,*.rpm,*.snap,*.gpg,*.pyc,*.png,*.ico,*.jar,*.so,changelog,.git,.hg,.mypy_cache,.tox,.venv,venv,_build,buck-out,__pycache__,build,dist,.vscode,parts,stage,prime,test_appstream.py,./snapcraft.spec,./.direnv,./.pytest_cache,.ruff_cache,*.asc
+skip = waf,*.tar,*.xz,*.zip,*.bz2,*.7z,*.gz,*.deb,*.rpm,*.snap,*.comp,*.gpg,*.pyc,*.png,*.ico,*.jar,*.so,changelog,.git,.hg,.mypy_cache,.tox,.venv,venv,_build,buck-out,__pycache__,build,dist,.vscode,parts,stage,prime,test_appstream.py,./snapcraft.spec,./.direnv,./.pytest_cache,.ruff_cache,*.asc
```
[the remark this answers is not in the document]
mr-cal: Added

snapcraft/meta/snap_yaml.py
```diff
@@ -558,3 +561,26 @@ def _populate_environment(
         del environment["PATH"]
 
     return environment if environment else None
+
+
+def _process_components(components: Dict[str, Any] | None) -> Dict[str, Any] | None:
```
sergiusens, line 566: why do we even add a version if we are going to remove it?
is the type not a dict of str, Component?

tests/spread/general/store/task.yaml
```diff
@@ -18,6 +18,9 @@ prepare: |
     exit 1
   fi
 
+  # components are only supported on the edge channel
+  snap refresh snapd --edge
```
sergiusens, line 22: we do not need this with my fix

snapcraft/meta/snap_yaml.py
mr-cal: Thanks, I made the type more specific.

From my interpretation of the spec, a component's version belongs in the component's metadata but not in the snap's metadata.  I asked in the spec to see if my understanding is correct.

tests/spread/general/store/task.yaml
mr-cal: 10-4, I'll drop this when your PR lands.

mr-cal:
Ready for review, all failures are to apt package issues

snapcraft/commands/upload.py
```diff
@@ -107,6 +151,54 @@ def run(self, parsed_args):
...
+    :param provided_components: A list of ComponentOptions provided by the user.
+    :param snap_yaml: The snap's metadata.
+    """
+    # get the names of the components provided by the user
+    provided_names = {
+        component.name for component in provided_components if component.name
+    }
+    emit.debug(f"Components provided by the user: {provided_names}")
+
+    if "components" in snap_yaml:
```
sergiusens, line 172: can you add a followup to use the new model and load that instead to process this? Instead of looking at snap_yaml, look at a Snap object.

sergiusens: hmm, the model is updated in this PR, so perhaps now?

sergiusens (commented):
looks pretty good, I would encourage operating on the object instead of the dictionary, but in principle I approve of this PR. Will not leave a mark for @lengau  and @cmatsuoka to be the final validators of the code

snapcraft/commands/upload.py
cmatsuoka: Agree with Sergio, this can be done against validated structured data instead of raw content.

mr-cal: Good idea, added.

snapcraft/commands/core22/upload.py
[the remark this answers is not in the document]
lengau: Yeah I want to update that in charmcraft too. I wonder whether a generic `KeyValueOption` in craft-cli would be the right way to go about this?

Anyway I've added this too: https://github.com/canonical/charmcraft/issues/1627

snapcraft/meta/snap_yaml.py
```diff
@@ -558,3 +580,21 @@ def _populate_environment(
...
+
+    :param components: Component data from a project model.
+
+    :returns: A dictionary of ComponentMetadata or None if no components are defined.
+    """
+    if not components:
+        return None
+
+    return {
+        name: ComponentMetadata.unmarshal(data.marshal())
```
lengau, line 598: How do you feel about a `from_component` classmethod on `ComponentMetadata` that takes the component itself? The initial implementation could just be `return cls.unmarshal(component.marshal()`, but then if we later need to do anything fancy we don't have to find cases like this.

snapcraft/commands/upload.py
```diff
@@ -107,6 +153,54 @@ def run(self, parsed_args):
         emit.message(message)
 
 
+def _validate_components(
+    provided_components: list[ComponentOption], snap_metadata: SnapMetadata
```
lengau, line 157: It's pedantic and probably unnecessary, but `provided_components` could be a Collection or possibly even an Iterable.

```suggestion
    provided_components: abc.Collection[ComponentOption], snap_metadata: SnapMetadata
```

lengau (approved):
Looks good! I've offered two minor suggestions, but it's not something I'd block if you dislike them.

snapcraft/meta/snap_yaml.py
mr-cal: Good idea, that is more readable and usable.

sergiusens: yes, this is what I actually wanted :-)

sergiusens (approved):

PR #4741, opened 2024-04-17
fix(env): set global environment variables for craft-app

syu-w:


mr-cal:
Closing in favor of #4757

PR #4778, opened 2024-04-30
feat(names) add --format option

DanielDewberry:
Add --format option to snapcraft names to enable output format selection. Possible choices are:
- table (default, existing behaviour)
- json

tests/unit/commands/test_names.py:
- Add test class TestNames.
- Rename test_names to test_table and move it into the scope of TestName. Add format to argparse.Namespace.
- Add test_json to TestNames.

-----

The motivation for adding JSON formatted output to the names sub-command, is that of data pipelines. It will enable tooling to read the command output and leverage JSON parsers rather than performing text processing. A small quality of life improvement :)
The JSON structure leaves room for extension if subsequently required:

```json
{
    "snaps": [
        {
            "name": "test-snap-private",
            "since": "2016-07-26T20:18:32Z",
            "visibility": "private",
            "notes": "-",
        },
    ]
}
```


DanielDewberry:
I added the email address on the commits to Launchpad several hours ago in the hope that the CLA check would validate correctly...

snapcraft/commands/names.py
```diff
@@ -138,12 +149,25 @@ def run(self, parsed_args):
...
+            elif parsed_args.format == "json":
+                json_snaps = {
+                    "snaps": [
+                        {header.lower(): snap[n] for n, header in enumerate(headers)}
+                        for snap in snaps
+                    ]
+                }
+                emit.message(json.dumps(json_snaps))
+            else:
+                raise NotImplementedError("Format not implemented", parsed_args.format)
```
mr-cal, line 170: Can you add a unit test for this too?

Something along the lines of
```python
 def test_table_error(self, emitter, fake_store_get_names, command_class):
        cmd = command_class(None)

        with pytest.raises(NotImplementedError) as raised:
            cmd.run(
                argparse.Namespace(
                    store_id="1234",
                    private=False,
                    yes=True,
                    format="invalid",
                    **{"snap-name": "test-snap"},
                )
            )

         assert raised == NotImplementedError("Format not implemented", "invalid")
```

mr-cal (changes requested):
Great work, thanks for another contribution!

snapcraft/commands/names.py
DanielDewberry: @mr-cal Done :)  I also updated the original functionality to pretty print the json using `indent=4`. `pytest` & `tox` both pass successfully.

The tests have kicked off again which is a good sign.

mr-cal (approved):

tigarmo (approved):

lengau (approved):
Thanks! This looks fairly similar to how Charmcraft does this, but with some improvements I'll probably be copying over :-)

snapcraft/commands/names.py
```diff
@@ -138,12 +149,25 @@ def run(self, parsed_args):
...
+            if parsed_args.format == "table":
+                tabulated_snaps = tabulate(
+                    snaps,
+                    headers=headers,
+                    tablefmt="plain",
+                )
+                emit.message(tabulated_snaps)
+            elif parsed_args.format == "json":
+                json_snaps = {
+                    "snaps": [
```
lengau, line 163: Sorry for undoing my approval, but I just thought a bit more about this and would like a bit of discussion:

Why did you choose to make this an object containing a list of snaps rather than just a list of snaps? (I don't think it's necessarily "wrong," I just want to understand this decision.)

lengau:
@DanielDewberry if you have opinions on a general case for this on more commands (e.g. `list-tracks`), I'd love to hear them in [this feature request](https://github.com/canonical/craft-application/issues/342).

snapcraft/commands/names.py
DanielDewberry: @lengau I took inspiration from the `snap metrics` output format. By placing the data in a list within an object, metadata can be easily added in future.

DanielDewberry:
> Thanks! This looks fairly similar to how Charmcraft does this, but with some improvements I'll probably be copying over :-)

Thank you, that's a compliment and a half :)

tests/unit/commands/test_names.py
```diff
@@ -199,22 +200,83 @@ def test_register_store_id(emitter, fake_store_register):
...
+
+        emitter.assert_message(
+            json.dumps(
+                {
+                    "snaps": [
+                        {
+                            "name": "test-snap-private",
+                            "since": "2016-07-26T20:18:32Z",
+                            "visibility": "private",
+                            "notes": "-",
```
sergiusens, line 254: if notes is `-`, from a data model perspective, should this not just be nil?

DanielDewberry: I thought it strange that the data produced by `snaps = store_client.get_names()` set `notes` to `-`. I expected it to contain a `None`/`nil`/`null` value and `tabulate` to render that to `-`.  I decided to not transform the field so as to remain consistent with both the store and the table format. 

Should it be `null`? Yes I think it probably should. If there's appetite for this, I'll update the PR.

mr-cal: Yeah, do you mind making this change for the json output?  The table can remain as-is with the "-".

I think it should be an easy change, something along the lines of:
```diff
--- a/snapcraft/commands/names.py
+++ b/snapcraft/commands/names.py
@@ -59,6 +59,10 @@ _MESSAGE_REGISTER_SUCCESS = "Registered {!r}"
 _MESSAGE_REGISTER_NO = "Snap name {!r} not registered"


+def _set_nil(value):
+    return None if value == "-" else value
+
+
 class StoreRegisterCommand(AppCommand):
     """Register a snap name with the Snap Store."""

@@ -161,7 +165,7 @@ class StoreNamesCommand(AppCommand):
             elif parsed_args.format == "json":
                 json_snaps = {
                     "snaps": [
-                        {header.lower(): snap[n] for n, header in enumerate(headers)}
+                        {header.lower(): _set_nil(snap[n]) for n, header in enumerate(headers)}
                         for snap in snaps
                     ]
```

DanielDewberry: @mr-cal 

> do you mind making this change for the json output?

I'll happily update it in the next couple of days :)

Last time I force pushed to my own branch and that propagated through to the PR. Would it be sufficient to do that again?

mr-cal: Yep, that should do the trick.  Thanks!

DanielDewberry: @mr-cal I've updated the original commit and force pushed it.

mr-cal (approved):

tests/unit/commands/test_names.py
mr-cal: Thank you!

Issue #4780, opened 2024-05-01
New remote-build fails with some architecture definitions


mr-cal:
### Bug Description

Snapcraft does not handle some definitions of architectures for remote builds and fails immediately with no error or explanation.

It's possible that other variations of `architectures:` will fail too, so we should add a set of regression tests.

### To Reproduce

snapcraft remote-build

### Environment

n/a

### snapcraft.yaml

```shell
name: test-snap
version: '0.1'
summary: test
description: test

grade: devel
confinement: devmode

base: core22

architectures:
  - build-on: arm64


parts:
  my-part:
    plugin: nil
```

### Relevant log output

```shell
All data sent to remote builders will be publicly available. Are you sure you want to continue? [y/N]: y
remote-build is experimental and is subject to change. Use with caution.
Build completed.
Log files:
Artifacts:
```

### Additional context

From @jnsgruk on the snapcrafters channel ([source](https://matrix.to/#/!cniCqhKpzVvOBMHAVt:matrix.org/$XrCuSXWoFQNDYBnro0kQCbS2ffAabd1lwbf4uipS33Y?via=matrix.org&via=ubuntu.com&via=xentonix.net))

AlanGriffiths:
I've also seen failures without `architectures`(or `platforms`) in the .yaml. With, for example,

```
snapcraft remote-build --launchpad-accept-public-upload --build-for armhf
```

(Which works for [Miriway](https://github.com/Miriway/Miriway), but not for several of [canonical's Mir](https://github.com/canonical/mir/pull/3364) based projects. Not spotted what the significant difference between these is.)

popey:
Also:

This works: `snapcraft remote-build --build-for riscv64`

This does not: `snapcraft remote-build --build-for amd64,riscv64,arm64`

```
2024-05-02 23:33:24.773 Starting snapcraft, version 8.2.3
2024-05-02 23:33:24.773 Log verbosity level set to BRIEF
2024-05-02 23:33:24.774 Configuring application...
2024-05-02 23:33:24.774 Preparing application...
2024-05-02 23:33:24.774 Build plan: platform=None, build_for=amd64,riscv64,arm64
2024-05-02 23:33:24.774 Loading project file '/home/alan/Temp/Test/snap/snapcraft.yaml'
2024-05-02 23:33:24.779 Architecture 'amd64,riscv64,arm64' is not supported.
2024-05-02 23:33:24.779 Traceback (most recent call last):
[...]

Issue #4144, opened 2023-05-15
Schema: add "build-for"


jpnurmi:
### What needs to get done

Add `architectures.build-for` field to the [JSON schema](https://github.com/snapcore/snapcraft/blob/main/schema/snapcraft.json).

### Why it needs to get done

VS Code + [YAML](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml): 

![Screenshot from 2023-05-15 22-28-49](https://github.com/snapcore/snapcraft/assets/140617/913b580c-0ac3-4eba-a50e-723d33aa122e)

jpnurmi:
Trying to do a remote build of [workshops](https://github.com/canonical/workshops/blob/main/snap/snapcraft.yaml):

```sh
$ snapcraft remote-build
snapcraft remote-build is experimental and is subject to change - use with caution.                                                                                      
All data sent to remote builders will be publicly available. Are you sure you want to continue? [y/N]: y
Issues while validating snapcraft.yaml: The 'architectures[0]' property does not match the required schema: additional properties are not allowed ('build-for' was unexpected) or orderedDict([('build-on', 'amd64'), ('build-for', 'amd64')]) is not of type 'string'
```

mr-cal:
We have 2 other reports of this bug at https://bugs.launchpad.net/snapcraft/+bug/1992557 and https://bugs.launchpad.net/snapcraft/+bug/2007789 and an internal issue, CRAFT-1640.

The problem is that the core18|20 implementation of remote-build (the legacy remote builder) does not support core22 snapcraft.yaml syntax.

The new remote builder in snapcraft 8 accepts these keywords.  It can be used via the environment variable `SNAPCRAFT_REMOTE_BUILD_STRATEGY="disable-fallback"`

mr-cal:
We are planning to fix this.

The scope of work is:
1. add `build-for` to `schema/snapcraft.yaml`
2. in the legacy codebase, ensure `build-for` and `run-on` are mutually exclusive and the correct translation between the two occurs

I'm not confident on where and how `run-on` is consumed, so we'll need to do a quick prototype and step through the legacy code.

A welcome side-effect is that `build-for` will now be a valid keyword for core20 snaps.

mr-cal:
This work on the legacy remote builder was descoped in favor of the new remote builder. 

This is not a critical issue and in a few weeks Snapcraft 9 will remove the legacy remote builder, so I'm marking this as "won't fix".

PR #4793, opened 2024-05-07
fix(remote-build): don't fail with certain archs

lengau:
This fixes the initial reported bug in #4780. However, I haven't tested some other use cases listed in that bug report.

-----

### DIFF
--- a/snapcraft/models/project.py
+++ b/snapcraft/models/project.py
@@ -1225,10 +1225,13 @@ class SnapcraftBuildPlanner(models.BuildPlanner):
             error_prefix = f"Error for platform entry '{platform_label}'"
 
             # Make sure the provided platform_set is valid
-            try:
-                platform = Platform(**platform_data)
-            except CraftValidationError as err:
-                raise CraftValidationError(f"{error_prefix}: {str(err)}") from None
+            if isinstance(platform_data, Platform):
+                platform = platform_data
+            else:
+                try:
+                    platform = Platform(**platform_data)
+                except CraftValidationError as err:
+                    raise CraftValidationError(f"{error_prefix}: {str(err)}") from None
 
             # build_on and build_for are validated
             # let's also validate the platform label
--- a/snapcraft/models/project.py
+++ b/snapcraft/models/project.py
@@ -545,9 +545,10 @@ class Platform(models.CraftBaseModel):
             if isinstance(architecture, str):
                 build_on = build_for = [architecture]
             else:
-                build_for = (  # type: ignore[assignment]
-                    architecture.build_for or architecture.build_on
-                )
+                if architecture.build_for:
+                    build_for = architecture.build_for  # type: ignore[assignment]
+                else:
+                    build_for = architecture.build_on  # type: ignore[assignment]
                 build_on = architecture.build_on  # type: ignore[assignment]
 
             if isinstance(build_for, str):
--- a/snapcraft/models/project.py
+++ b/snapcraft/models/project.py
@@ -543,21 +543,21 @@ class Platform(models.CraftBaseModel):
         platforms: dict[str, Self] = {}
         for architecture in architectures:
             if isinstance(architecture, str):
-                build_on = build_for = [architecture]
+                build_on = build_for = cast(UniqueStrList, [architecture])
             else:
-                if architecture.build_for:
-                    build_for = architecture.build_for  # type: ignore[assignment]
+                if isinstance(architecture.build_on, str):
+                    build_on = build_for = cast(UniqueStrList, [architecture.build_on])
                 else:
-                    build_for = architecture.build_on  # type: ignore[assignment]
-                build_on = architecture.build_on  # type: ignore[assignment]
+                    build_on = build_for = cast(UniqueStrList, architecture.build_on)
+                if architecture.build_for:
+                    if isinstance(architecture.build_for, str):
+                        build_for = cast(UniqueStrList, [architecture.build_for])
+                    else:
+                        build_for = cast(UniqueStrList, architecture.build_for)
 
-            if isinstance(build_for, str):
-                build_for = [build_for]
-            if isinstance(build_on, str):
-                build_on = [build_on]
             platforms[build_for[0]] = cls(
-                build_for=build_for,  # type: ignore[assignment]
-                build_on=build_on,  # type: ignore[assignment]
+                build_for=build_for,
+                build_on=build_on,
             )
 
         return platforms
--- a/snapcraft/models/project.py
+++ b/snapcraft/models/project.py
@@ -1179,6 +1179,7 @@ class SnapcraftBuildPlanner(models.BuildPlanner):
     build_base: str | None = None
     name: str
     platforms: dict[str, Any] | None = None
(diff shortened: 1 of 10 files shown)


snapcraft/utils.py
```diff
@@ -25,13 +27,16 @@
...
 from pathlib import Path
-from typing import Iterable, List, Optional
+from typing import TYPE_CHECKING, Iterable, List, Optional
 
 from craft_cli import emit
 from craft_parts.sources.git_source import GitSource
 
 from snapcraft import errors
 
+if TYPE_CHECKING:
```
tigarmo, line 37: leftover?

snapcraft/models/project.py
```diff
@@ -1261,6 +1290,9 @@ def get_build_plan(self) -> List[BuildInfo]:
         # set default value
         if self.platforms is None:
             self.platforms = {get_host_architecture(): None}
+            # For backwards compatibility with core22, convert the platforms.
+            if self.base == "core22" and self.architectures:
```
tigarmo, line 1294: > self.base

Shouldn't we use the "effective" base here? For instance, this is a project that should fall into this right?

```yaml
base: bare
build-base: core22

Comment (tigarmo) [inline snapcraft/models/project.py:1294]: > self.architectures

I'm thinking of the other cases that we support for `core22`. Correct me if I'm wrong, but for core22 snaps:

- the `architectures` keyword isn't even [needed](https://github.com/canonical/snapcraft/blob/7.5.5/snapcraft/projects.py#L451) right?
- some fields [accept](https://github.com/canonical/snapcraft/blob/7.5.5/snapcraft/projects.py#L92) `"all"`, which at some point gets converted into the host architecture (I think?), but our notion of BuildInfo can't handle that (can it)?

I think these are some good test cases to add, at least

**Review by tigarmo — CHANGES_REQUESTED** [review]
thanks!

**Comment by tigarmo** [comment]
@lengau shouldn't this PR be targeting the hotfix branch?

File: snapcraft/utils.py (hunk 1 above)
Comment (lengau) [reply snapcraft/utils.py:37]: Yep, good catch thanks!

I wonder if it'd be possible to get ruff's formatter to just remove these sorts of if statements **ONLY IF** they're checking for `TYPE_CHECKING`

File: snapcraft/models/project.py (hunk 2 above)
Comment (lengau) [reply snapcraft/models/project.py:1295]: Good point! I actually just repurposed `effective_base` from a few lines above since that just gets us a string of "22.04" and does exactly what we need otherwise.

Comment (lengau) [reply snapcraft/models/project.py:1295]: That's a good point. What *should* be the default behaviour if there are no `architectures` and no `platforms` provided? Right now it's just "build on the host," but that's probably not what we want for remote builds. Right now this works without an `architectures` keyword, but it probably does the wrong thing in the `remote-build` case since it ends up building on just the client's architecture. Maybe we make the build plan just list all supported architectures and not do cross-compiling? Filtering the plan will result in the same behaviour for local builds.

I'll have to give a bit more thought to the second point and how we want this to behave. @mr-cal wdyt?

**Comment by lengau** [comment]
@tigarmo you're right! I'll retarget to the hotfix branch. Thanks!

File: snapcraft/models/project.py [hunk 3]
```diff
@@ -533,6 +534,34 @@ def _validate_platform_set(cls, values: Mapping[str, Any]) -> Mapping[str, Any]:
...
+    @classmethod
+    def from_architectures(
+        cls,
+        architectures: list[str | Architecture],
+    ) -> dict[str, Self]:
+        """Convert a core22 architectures configuration to core24 platforms."""
+        platforms: dict[str, Self] = {}
+        for architecture in architectures:
+            if isinstance(architecture, str):
+                build_on = build_for = [architecture]
```
mr-cal, line 546: I wonder if typing `build_on` and `build_for` would let you drop the subsequent `type: ignore`s.
```suggestion
                build_on: list[str] | str = build_for: list[str] | str = [architecture]
```

mr-cal (commented):
I like this approach. I left my main feedback as a reply in another comment.

snapcraft/models/project.py
mr-cal: I'm not sure what to do about `build-for: all` for remote-builds.  There is no precedent because it's never worked ([1](https://bugs.launchpad.net/snapcraft/+bug/1992557) [2](https://bugs.launchpad.net/snapcraft/+bug/2007789) [3](https://github.com/canonical/snapcraft/issues/4144)).  This may be a good time for snapcraft to raise an error when converting `build-for: [all]` to `platforms`.

Regarding the behavior when no `platforms` or `architectures` are defined.  Building for the client architecture makes sense if remote-build is used for development and testing.  Building for each architecture makes sense if remote-build is used for CI and distribution.  I lean towards the former because `snapcraft pack` and `snapcraft remote-build` would both produce 1 snap, built on and built for the client's architecture.

tests/unit/models/test_projects.py
```diff
@@ -710,6 +710,66 @@ def test_platforms_not_allowed_core22(self, project_yaml_data):
...
+                    ),
+                    "arm64": Platform(
+                        build_for=[const.SnapArch("arm64")],
+                        build_on=[const.SnapArch("amd64"), const.SnapArch("arm64")],
+                    ),
+                },
+            ),
+        ],
+    )
+    def test_from_architectures(self, architectures, expected):
```
tigarmo, line 770: This test is fine, but with all the validators and such I think we might want to have a (parametrized) test specifically for `get_build_plan()` values for `core22` projects. What do you think?

snapcraft/models/project.py
lengau: That doesn't work because mypy is complaining that it's not a `UniqueStrList`. However, I've popped in a change that uses 5 casts, but eliminates all the type checker silencing.

snapcraft/models/project.py
lengau: OK, so I've left the default build plan as creating for only the current arch, and I've set it to error out explicitly if there's a `build-for: all` for core22. I've also made `from_architectures` explicitly raise an exception if an architecture has `build-for: all`

tests/unit/models/test_projects.py
lengau: You're right, and implementing this test caught a bug!

tests/unit/models/test_projects.py
```diff
@@ -1969,6 +1974,74 @@ def test_build_planner_get_build_plan(platforms, expected_build_infos):
     assert actual_build_infos == expected_build_infos
 
 
+@pytest.mark.parametrize(
+    ("architectures", "expected_build_infos"),
+    [
+        pytest.param(
+            ["amd64"],
```
tigarmo, line 1981: what about the case where `architectures` is `None`? I think I lost track of where we landed with regards to remote-build in this case.

lengau: Added! In this case the remote build will retain its current behaviour (which was already tested for) of building for the host platform.

tigarmo (approved):
nice! Thanks

mr-cal (approved):
LGTM

PR #4863, opened 2024-06-18
fix(cli): properly capture store errors

sergiusens:
When raised they show up as a Snapcraft internal error if not properly
handled.

This is using the same logic as in legacy.

### DIFF
--- a/snapcraft/application.py
+++ b/snapcraft/application.py
@@ -27,6 +27,7 @@ from typing import Any, Optional
 import craft_application.commands as craft_app_commands
 import craft_cli
 import craft_parts
+import craft_store
 from craft_application import Application, AppMetadata, util
 from craft_cli import emit
 from craft_parts.plugins.plugins import PluginType
@@ -34,7 +35,7 @@ from overrides import override
 
 import snapcraft
 import snapcraft_legacy
-from snapcraft import cli, commands, errors, models, services
+from snapcraft import cli, commands, errors, models, services, store
 from snapcraft.commands import unimplemented
 from snapcraft.extensions import apply_extensions
 from snapcraft.models.project import SnapcraftBuildPlanner, apply_root_packages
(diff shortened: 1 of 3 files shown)


snapcraft/application.py
```diff
@@ -173,6 +174,34 @@ def app_config(self) -> dict[str, Any]:
         config["core24"] = self._known_core24
         return config
 
+    @override
+    def run(self) -> int:
+        try:
+            return_code = super().run()
+        except craft_store.errors.NoKeyringError as err:
+            self._emit_error(
+                craft_cli.errors.CraftError(
```
tigarmo, line 183: do you need the path to the logfile to be shown in this case? I think not, right? So we should set `logpath_report` to `False`

tigarmo (changes requested):

snapcraft/application.py
sergiusens: can you tell me more about this and in which other cases we are using that? Perhaps a topic for today's engineering sync meeting

tigarmo: Sure. I only realized this parameter exists recently. What I mean is that in cases where it's clear where there's an user error, specially in cases with known resolutions, like this:

```bash
❯ rockcraft pack
Project file 'rockcraft.yaml' not found in '/home/tiago'.                                                                                                                                   
Recommended resolution: Ensure the project file exists.                                                                                                                                     
Full execution log: '/home/tiago/.local/state/rockcraft/log/rockcraft-20240619-085058.851427.log'
```

... it seems to me like it doesn't make sense to show the path to the execution log. There's nothing there that could help anyone solve that particular issue, and in a sense it implies a bit that it was a bug in the application, doesn't it?

sergiusens: I agree, but let's discuss as I believe there was this intent to always show the logs on errors to pass on to us in case of errors.
I personally prefer not to show it ever on handled errors

tigarmo (approved):

cmatsuoka (approved):

PR #4864, opened 2024-06-19
build(docs): fix version in rtd builds

tigarmo:
Readthedocs builds don't fetch tags when cloning the repo. This is a problem for us because Snapcraft's version comes from the git tag, so do a fetch of the tags as a post-checkout step.

Additionally, explicitly install snapcraft itself (via "pip install .") so that the code that gets the version from git actually runs. Note that this already works on "latest" because that builds off of "main" where this fix has already been applied (when moving to canonical-sphinx).

-----

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -14,6 +14,10 @@ build:
     python: "3.10"
   apt_packages:
     - libapt-pkg-dev
+  jobs:
+    post_checkout:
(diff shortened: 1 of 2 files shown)


tigarmo:
image

sergiusens (approved):

dariuszd21 (approved):
Thanks! Looks really good

PR #4888, opened 2024-06-30
meta: override the links if project links are set and check the links from appstream

### LINKED ISSUE #4890: URL links in snapcraft.yaml should override appstream metadata
### What needs to get done

https://github.com/canonical/snapcraft/pull/4888

### Why it needs to get done

So developers can build a snap despite bad appstream data

soumyaDghosh:

-----

This PR checks if the values that are coming from the appstream file are either http/s urls or emails. If they are, those are added to the `ExtractedMetadata`, else they are passed, no code is executed.

Closes #4890

### DIFF
--- a/snapcraft/meta/appstream.py
+++ b/snapcraft/meta/appstream.py
@@ -23,6 +23,8 @@ from io import StringIO
 from typing import List, Optional, cast
 
 import lxml.etree
+import validators
+from craft_cli import emit
 from xdg.DesktopEntry import DesktopEntry
 
 from snapcraft import errors
@@ -102,7 +104,14 @@ def extract(relpath: str, *, workdir: str) -> Optional[ExtractedMetadata]:
     version = _get_latest_release_from_nodes(dom.findall("releases/release"))
     project_license = _get_value_from_xml_element(dom, "project_license")
     update_contact = _get_value_from_xml_element(dom, "update_contact")
-    contact = [update_contact] if update_contact else None
+    contact = None
+    if update_contact:
+        if validators.url(update_contact) or validators.email(update_contact):
+            contact = [update_contact]
+        else:
+            emit.debug(
+                f"Invalid URL {update_contact} as {validators.url(update_contact)}"
+            )
 
     issues = _get_urls_from_xml_element(dom.findall("url"), "bugtracker")
     donation = _get_urls_from_xml_element(dom.findall("url"), "donation")
@@ -184,7 +193,11 @@ def _get_urls_from_xml_element(nodes, url_type) -> Optional[List[str]]:
             and node.attrib["type"] == url_type
             and node.text.strip() not in urls
         ):
-            urls.append(node.text.strip())
+            link = node.text.strip()
+            if validators.url(link) or validators.email(link):
+                urls.append(link)
+            else:
+                emit.debug(f"Invalid URL {link} as {validators.url(link)}")
     if urls:
         return urls
     return None
--- a/snapcraft/meta/appstream.py
+++ b/snapcraft/meta/appstream.py
@@ -199,7 +199,7 @@ def _get_urls_from_xml_element(nodes, url_type) -> Optional[List[str]]:
                 urls.append(link)
             else:
                 emit.progress(
-                    f"Ignoring invalid url {link!r} in {url_type!r} from appstream metadata.",
+                    f"Ignoring invalid url or email {link!r} in {url_type!r} from appstream metadata.",
                     permanent=True,
                 )
     if urls:
--- a/snapcraft/meta/appstream.py
+++ b/snapcraft/meta/appstream.py
@@ -109,8 +109,9 @@ def extract(relpath: str, *, workdir: str) -> Optional[ExtractedMetadata]:
         if validators.url(update_contact) or validators.email(update_contact):
             contact = [update_contact]
         else:
-            emit.debug(
-                f"Invalid URL {update_contact} as {validators.url(update_contact)}"
+            emit.progress(
+                f"Ignoring invalid url {update_contact!r} in update_contact from appstream metadata.",
+                permanent=True,
             )
 
     issues = _get_urls_from_xml_element(dom.findall("url"), "bugtracker")
@@ -197,7 +198,10 @@ def _get_urls_from_xml_element(nodes, url_type) -> Optional[List[str]]:
             if validators.url(link) or validators.email(link):
                 urls.append(link)
             else:
-                emit.debug(f"Invalid URL {link} as {validators.url(link)}")
+                emit.progress(
+                    f"Ignoring invalid url {link!r} in {url_type!r} from appstream metadata.",
+                    permanent=True,
+                )
     if urls:
         return urls
     return None
--- a/snapcraft/parts/update_metadata.py
+++ b/snapcraft/parts/update_metadata.py
@@ -17,7 +17,7 @@
 """External metadata helpers."""
 
 from pathlib import Path
-from typing import Dict, Final, List, cast
+from typing import Dict, Final, List, OrderedDict, cast
 
 import pydantic
 from craft_application.models import ProjectTitle, SummaryStr, UniqueStrList, VersionStr
@@ -115,18 +115,26 @@ def _update_project_links(
     :param project: The Project model to update.
     :param metadata_list: A list of parsed information from metadata files.
     """
-    for metadata in metadata_list:
-        fields = ["contact", "donation", "source_code", "issues", "website"]
-        for field in fields:
-            metadata_field = getattr(metadata, field)
-            project_field = getattr(project, field)
-            if metadata_field and project_field and project_field != metadata_field:
-                project_list = list(project_field)
-                project_list.extend(set(metadata_field) - set(project_list))
-                setattr(project, field, cast(UniqueStrList, project_list))
+    fields = ["contact", "donation", "source_code", "issues", "website"]
+    for field in fields:
+        project_field = getattr(project, field)
 
-            if not getattr(project, field):
-                setattr(project, field, cast(UniqueStrList, metadata_field))
+        # only update the project if the project has not defined the field
+        if not project_field:
+
+            # values for a field from all metadata files
+            metadata_values = list()  # type: List[str]
+
+            # iterate through all metadata and create a set of values for the field
+            for metadata in metadata_list:
+                if metadata_field := getattr(metadata, field):
+                    metadata_values = list(
+                        OrderedDict.fromkeys(metadata_values + metadata_field)
+                    )
+
+            # update project with all new values from the metadata
+            if metadata_values:
+                setattr(project, field, cast(UniqueStrList, metadata_values))
 
 
 def _update_project_variables(project: Project, project_vars: Dict[str, str]):
--- a/snapcraft/parts/update_metadata.py
+++ b/snapcraft/parts/update_metadata.py
@@ -123,7 +123,7 @@ def _update_project_links(
         if not project_field:
 
             # values for a field from all metadata files
-            metadata_values = list()  # type: List[str]
+            metadata_values: list[str] = list()
 
             # iterate through all metadata and create a set of values for the field
             for metadata in metadata_list:
--- a/tests/unit/meta/test_appstream.py
+++ b/tests/unit/meta/test_appstream.py
@@ -782,20 +782,23 @@ class TestAppstreamContent:
               <project_license>GPL-3.0-or-later</project_license>
               <content_rating type="oars-1.1"/>
               <name>Drawing</name>
-              <url type="homepage">test_at_hello_dot_com</url>
+              <url type="homepage">bad_invalid_mail</url>
               <url type="homepage">https://hello.com</url>
-              <url type="vcs-browser">file://test.deb</url>
-              <url type="vcs-browser">http://file_huge</url>
+              <url type="vcs-browser">file://invalid.file.link</url>
+              <url type="vcs-browser">http://missing-dot-com</url>
               <url type="bugtracker">https://github.com/alainm23/planify/issues</url>
-              <url type="bugtracker">https:/helo.com</url>
+              <url type="bugtracker">https:/missing-slash.com</url>
               <url type="donation">https://www.buymeacoffee.com/alainm23</url>
-              <url type="donation">paypal.me/alainm23</url>
+              <url type="donation">paypal.me/bad</url>
+              <update_contact>rrroschan@gmail.com</update_contact>
+              <update_contact>malformed_invalid@mailcom</update_contact>
               </component>
         """
         )
         Path(file_name).write_text(content)
         metadata = appstream.extract(file_name, workdir=".")
         assert metadata is not None
+        assert metadata.contact == ["rrroschan@gmail.com"]
         assert metadata.donation == ["https://www.buymeacoffee.com/alainm23"]
         assert metadata.website == ["https://hello.com"]
         assert metadata.source_code == None
--- a/tests/unit/meta/test_appstream.py
+++ b/tests/unit/meta/test_appstream.py
@@ -50,22 +50,40 @@ class TestAppstreamData:
             ("id", {}, "common_id", "test-id", "test-id"),
             ("name", {}, "title", "test-title", "test-title"),
             ("project_license", {}, "license", "test-license", "test-license"),
-            ("update_contact", {}, "contact", "test-contact", ["test-contact"]),
-            ("url", {"type": "homepage"}, "website", "test-website", ["test-website"]),
-            ("url", {"type": "bugtracker"}, "issues", "test-issues", ["test-issues"]),
+            (
+                "update_contact",
+                {},
+                "contact",
+                "test-contact@test.com",
+                ["test-contact@test.com"],
+            ),
+            (
+                "url",
+                {"type": "homepage"},
+                "website",
+                "http://test-website.test",
+                ["http://test-website.test"],
+            ),
+            (
+                "url",
+                {"type": "bugtracker"},
+                "issues",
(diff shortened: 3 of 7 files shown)


mr-cal (commented):
What do you think about updating this:

https://github.com/canonical/snapcraft/blob/e034a5eb4a6285825b2cf3c8db325687ab48a6e2/snapcraft/parts/update_metadata.py#L123

such that metadata in `snapcraft.yaml` always override the appstream metadata?

I think this was Sergio's request in https://forum.snapcraft.io/t/call-for-testing-snapcraft-8-3-0/40789/6.  The snapcraft user needs a way to override each adopted field of the appstream metadata.

snapcraft/meta/appstream.py
```diff
@@ -102,7 +104,17 @@ def extract(relpath: str, *, workdir: str) -> Optional[ExtractedMetadata]:
...
+    if update_contact:
+        if urlparse(update_contact).scheme in ["http", "https"]:
+            contact = [update_contact]
+        else:
+            try:
+                EmailStr.validate(update_contact)
+            except:
+                pass
+            else:
+                contact = [update_contact]
```
sergiusens, line 117: Something like this might be a bit more manageable (no blank exception catching either)
```suggestion
    if update_contact and urlparse(update_contact).scheme in ["http", "https"]:
        contact = [update_contact]
    else:
        try:
            EmailStr.validate(update_contact)
            contact = [update_contact]
        except pydantic.ValidationError as err:
            emit.debug(f"Invalid email {update_contact!r} set: {err!s}")
```

sergiusens (changes requested):
Looks good, just a comment on the handling

@mr-cal this blocks 8.3.0

I would like to see an invalid appstream input scenario unit tested.
I do not think this covers being able to override the entry from snapcraft.yaml, but that might be something deeper in the infra for @cmatsuoka to look into

snapcraft/meta/appstream.py
soumyaDghosh: This one is a bit problematic as what the appstream has passed might not be an email and a malformed website url, like eg "http:/testcom", I have found another better library/tool to fix this. But, this would create a new addition to the `requirements.txt`. 

https://pypi.org/project/validators/

And I think we can emit a general error that `invalid url {update_contact}`

Regarding test, yes, I also thought about it, and I think it should have its own unit test. Although, I did add one such case here already

[L528](https://github.com/canonical/snapcraft/pull/4888/files#diff-5eaf20f392bec62c4f476a461c87baf38d7f9710774bf720459bd1ea2808e408R528)
[L553](https://github.com/canonical/snapcraft/pull/4888/files#diff-5eaf20f392bec62c4f476a461c87baf38d7f9710774bf720459bd1ea2808e408R553)

But, I would definitely want to add one such test.

soumyaDghosh:
> such that metadata in `snapcraft.yaml` always overrides the appstream metadata?

Well, appstream metadata file is one that can always be changed, and if we want to override then, we'll need to make these fields a single string, and looking at the storefront, I am not sure if that'd be a correct thing to do. For third snappers like us, we can always do this right?

https://github.com/soumyaDghosh/monophony-snap/blob/main/snap/snapcraft.yaml#L70

mr-cal:
> Is the failure due to the new module added? Because, it's there in the `requirements.txt` file and `requirements-devel.txt`

You also need it in `docs/requirements.txt`!

soumyaDghosh:
> > Is the failure due to the new module added? Because, it's there in the `requirements.txt` file and `requirements-devel.txt`
> 
> You also need it in `docs/requirements.txt`!

Yeeah, figured it out

mr-cal:
> Well, appstream metadata file is one that can always be changed, and if we want to override then, we'll need to make these fields a single string, and looking at the storefront, I am not sure if that'd be a correct thing to do.

Can you clarify what you mean by "make these fields a single string"?  

The `override-pull`/`sed` option would certainly work.

@cmatsuoka - can you choose from these two approaches?
1. Metadata in `snapcraft.yaml` for contact, donation, source_code, issues, and website will override the contents of the appstream.  I prefer this because this is what already happens for title, summary, description, version, license, and grade.
2. Add a how-to documentation page for users to use an `override-pull` and `sed` call to rewrite the appstream like this: https://github.com/soumyaDghosh/monophony-snap/blob/main/snap/snapcraft.yaml#L70

soumyaDghosh:
> Can you clarify what you mean by "make these fields a single string"?

Very sorry, that was a wrong understanding by me. I wanted to say is, if we want to block/override the appstream data then we'll have to completely ignore those. I wrongly thought, that it'd make those fields act as a `String` once again, but users can set those fields with multiple values directly in the project.  Sorry for that mistake.

But may be overriding those fields is the correct approach. I, now, think, that it's the correct approach. But, anyways, this bug also got caught by that problem. :-)

mr-cal:
Related but not relevant, here is the snapd validation code for emails and URLs: https://github.com/snapcore/snapd/blob/a58ebef2534481154b4db06ec485bed4faf55f61/snap/validate.go#L1422

soumyaDghosh:
> Related but not relevant, here is the snapd validation code for emails and URLs: https://github.com/snapcore/snapd/blob/a58ebef2534481154b4db06ec485bed4faf55f61/snap/validate.go#L1422

The validation done by the `validators` is actually stricter than what `snapd` is doing. In my first iteration of this commit, I also did somewhat similar, but that still kept a loophole. Which can be found in this [new test,](https://github.com/canonical/snapcraft/pull/4888/files#diff-5eaf20f392bec62c4f476a461c87baf38d7f9710774bf720459bd1ea2808e408R774-R803). The `snapd` validation seems to be allowing URLs like `https://test` or emails like `test@test` but these are also malformed.



> updated such that the `snapcraft.yaml` entries override the appstream for these fields and a test for this.

I was also looking into this and the only problem with this, is the fact that, `parse-info` takes `list`, so, there can be multiple objects of `ExtractedMetadata`. Now, if I do a checking like this, which I tried locally,

```python
for metadata in metadata_list:
        fields = ["contact", "donation", "source_code", "issues", "website"]
        for field in fields:
            metadata_field = getattr(metadata, field)
            project_field = getattr(project, field)
            if metadata_field and not project_field:
                project_list = list(metadata_field)
                setattr(project, field, cast(UniqueStrList, project_list))
```

this would update the metadata from the first object, but during the second iteration, it'll fail, because the `project_field` isn't empty anymore. Which essentially fails this [test](https://github.com/canonical/snapcraft/blob/main/tests/unit/parts/test_update_metadata.py#L243-L304). But, this brings it up, can one single part really have multiple `appstream` metadata files to parse from? And because `adopt-info` is a string, we can parse details from only one part.

mr-cal:
> this would update the metadata from the first object, but during the second iteration, it'll fail, because the project_field isn't empty anymore. Which essentially fails this test. But, this brings it up, can one single part really have multiple appstream metadata files to parse from? And because adopt-info is a string, we can parse details from only one part.

You're right.  You would have to update that loop to only update project metadata at the end.  Something along the lines of:

```python
fields = ["contact", "donation", "source_code", "issues", "website"]
for field in fields:
    project_field = getattr(project, field)

    # only update the project if the project has not defined the field
    if not project_field:

        # values for a field from all metadata files
        metadata_values = set()

        # iterate through all metadata and create a set of values for the field
        for metadata in metadata_list:
            if metadata_field := getattr(metadata, field):
                metadata_values.update(metadata_field)

        # update project with all new values from the metadata
        if metadata_values:
            setattr(project, field, cast(UniqueStrList, metadata_values))
```

There are many ways to attack that data structure problem, feel free to do it how you wish.

I agree that there isn't a real-world scenario for more than one metadata file to iterate but we still need to support this design for now.

> I have another suggestion. As we're already using pydantic to convert the links from string to list, we can do the validation there also, and pause the build. This way, it can save a lot of time in the build itself. Because, for snapd doing the validation while packing, the whole build is probably successful, and to see it fail, just because of a malformed link is something really painful to see 😢

I think that validation in the `Project` model is a good idea but here is my concern:

It's ok to add strict validation to the newly adopted fields from the appstream, because they've never been adopted prior to Snapcraft 8.3 and this validation won't cause builds to fail.  It only prevents the field from being adopted.

What we can't do is add validation to the fields in the ``snapcraft.yaml`` file itself that is more strict than snapd.  First, because Snapcraft shouldn't be more strict than snapd.  If we want to tighten validation, a PR/feature request to snapd is the first step.  Second, because this may cause existing snaps to fail to build (even if the change is as simple as updating a contact field in their `snapcraft.yaml`)

soumyaDghosh:
@mr-cal Facing a big issue here. Well, previously using `set` it didn't change the position of a value, like, if the list is

```python
----------------------------------------------------- Captured stdout call -----------------------------------------------------
['test@test.com']
['donation1', 'donation2']
['donation2']
['source-code']
['source-code', 'vcs-browser']
['issues1', 'issues3']
['issues2', 'issues3']
['website1']
['website2']
```
It'd get the position changed when used as `set`

```python
----------------------------------------------------- Captured stdout call -----------------------------------------------------
{'test@test.com'}
{'donation2', 'donation1'}
{'donation2'}
{'source-code'}
{'vcs-browser', 'source-code'}
{'issues3', 'issues1'}
{'issues2', 'issues3'}
{'website1'}
{'website2'}
```

Thus it's failing the sequence

soumyaDghosh:
@mr-cal Done as you requested :-)

snapcraft/parts/update_metadata.py
```diff
@@ -115,18 +115,27 @@ def _update_project_links(
...
+            # iterate through all metadata and create a set of values for the field
+            for metadata in metadata_list:
+                if metadata_field := getattr(metadata, field):
+                    metadata_values = list(
+                        OrderedDict.fromkeys(metadata_values + metadata_field)
+                    )
+
+            # update project with all new values from the metadata
+            if metadata_values:
+                print(metadata_values)
```
sergiusens, line 137: leftover?

soumyaDghosh: Uhhh... shit, yeah, leftover

snapcraft/parts/update_metadata.py
```diff
@@ -115,18 +115,26 @@ def _update_project_links(
...
-                setattr(project, field, cast(UniqueStrList, metadata_field))
+    fields = ["contact", "donation", "source_code", "issues", "website"]
+    for field in fields:
+        project_field = getattr(project, field)
+
+        # only update the project if the project has not defined the field
+        if not project_field:
+
+            # values for a field from all metadata files
+            metadata_values = list()  # type: List[str]
```
mr-cal, line 126:
```suggestion
            metadata_values: list[str] = list()
```

tests/unit/meta/test_appstream.py
```diff
@@ -749,6 +771,36 @@ def test_appstream_with_multiple_lists(self):
...
+              <metadata_license>CC0-1.0</metadata_license>
+              <project_license>GPL-3.0-or-later</project_license>
+              <content_rating type="oars-1.1"/>
+              <name>Drawing</name>
+              <url type="homepage">test_at_hello_dot_com</url>
+              <url type="homepage">https://hello.com</url>
+              <url type="vcs-browser">file://test.deb</url>
+              <url type="vcs-browser">http://file_huge</url>
+              <url type="bugtracker">https://github.com/alainm23/planify/issues</url>
+              <url type="bugtracker">https:/helo.com</url>
```
mr-cal, line 790: Can you encode something into the malformed URLs to make it obvious that they are invalid?  Some of the problems like `:/` vs `://` here are subtle and takes careful reading to understand the expected outcome of the test.


I'm thinking something along the lines of
```suggestion
              <url type="bugtracker">https:/invalid.com</url>
              <url type="bugtracker">https:/bad.com</url>
              <url type="bugtracker">https:/missing-slash.com</url>
```

tests/unit/parts/test_update_metadata.py
```diff
@@ -296,14 +295,54 @@ def test_update_project_metadata_multiple(
...
-    assert project.contact == ["test@test.com"]
+    assert project.contact == ["contact1", "contact2"]
     assert project.license == "GPL-3.0"
     assert project.donation == ["donation1", "donation2"]
     assert project.source_code == ["source-code", "vcs-browser"]
     assert project.issues == ["issues1", "issues3", "issues2"]
     assert project.website == ["website1", "website2"]
 
 
+def test_update_project_metadata_appstream(new_dir):
```
mr-cal, line 306: Which are you trying to test here:
1. `snapcraft.yaml` metadata overrides the appstream?
2. invalid metadata links are ignored?

mr-cal (changes requested):
Looking good!

tests/unit/parts/test_update_metadata.py
soumyaDghosh: The overriding of metadata. Test of malformed links can't be done here. Should I change the name?

mr-cal: Yeah, I think you should change the name or add a docstring to the test.


 Also, if you use valid URLs/emails in the appstream metadata here, then we will know they are overridden and not ignored for being invalid.

soumyaDghosh: Done

mr-cal (approved):
Looks great!

I tested on a few gnome apps with appstream files and can confirm both changes are working nicely.

snapcraft/meta/appstream.py
```diff
@@ -102,7 +104,15 @@ def extract(relpath: str, *, workdir: str) -> Optional[ExtractedMetadata]:
...
     project_license = _get_value_from_xml_element(dom, "project_license")
     update_contact = _get_value_from_xml_element(dom, "update_contact")
-    contact = [update_contact] if update_contact else None
+    contact = None
+    if update_contact:
+        if validators.url(update_contact) or validators.email(update_contact):
+            contact = [update_contact]
+        else:
+            emit.progress(
+                f"Ignoring invalid url {update_contact!r} in update_contact from appstream metadata.",
```
cmatsuoka, line 113: Can we read to `contact` directly using `_get_urls_from_xml_element` instead of assigning to `update_contact`, revalidating it, and then assigning to `contact`?

snapcraft/meta/appstream.py
```diff
@@ -184,7 +194,14 @@ def _get_urls_from_xml_element(nodes, url_type) -> Optional[List[str]]:
             and node.attrib["type"] == url_type
             and node.text.strip() not in urls
         ):
-            urls.append(node.text.strip())
+            link = node.text.strip()
+            if validators.url(link) or validators.email(link):
+                urls.append(link)
+            else:
+                emit.progress(
+                    f"Ignoring invalid url {link!r} in {url_type!r} from appstream metadata.",
```
cmatsuoka, line 202: This message could be slightly misleading if the user intended to use an email address, but it's not incorrect.

cmatsuoka (commented):
Thanks. Could you clarify the need of assigning to `update_contact` before we land this?

snapcraft/meta/appstream.py
mr-cal: Agreed, not critical but this could be simplified.

snapcraft/meta/appstream.py
mr-cal, line 202:
```suggestion
                    f"Ignoring invalid url or email {link!r} in {url_type!r} from appstream metadata.",
```

snapcraft/meta/appstream.py
mr-cal: Perhaps this?

```python
contact = _get_value_from_xml_element(dom, "update_contact")
if not validators.url(update_contact) and not validators.email(update_contact):
     contact = None
     emit.progress(
                f"Ignoring invalid url {update_contact!r} in update_contact from appstream metadata.",
                permanent=True,
    )
```

soumyaDghosh: `contact` is a `list[str]` to match the `project` model. So, can we really do so?

mr-cal: Ohh, right.  And I misunderstood what @cmatsuoka was asking about.

@cmatsuoka, I don't think we can call `_get_urls_from_xml_element` for `update_contact` because that is a single string whereas the `url` fields in an appstream are lists of strings.

Is that correct @soumyaDghosh?

soumyaDghosh: Okay, I think this is possible,

```python
    contact = _get_value_from_xml_element(dom, "update_contact")
    if contact:
        if not validators.url(contact) and not validators.email(contact):
            contact = None
            emit.progress(
                f"Ignoring invalid url {update_contact!r} in update_contact from appstream metadata.",
                permanent=True,
            )
```

And then, while returning, we'll just do this

```python
return ExtractedMetadata(
        common_id=common_id,
        title=title,
        summary=summary,
        description=description,
        version=version,
        icon=icon,
        license=project_license,
        contact=[contact],
        issues=issues,
        donation=donation,
        website=website,
        source_code=source_code,
        desktop_file_paths=desktop_file_paths,
    )
 ```

soumyaDghosh: > Ohh, right.  And I misunderstood what @cmatsuoka was asking about.
> 
> @cmatsuoka, I don't think we can call `_get_urls_from_xml_element` for `update_contact` because that is a single string whereas the `url` fields in an appstream are lists of strings.
> 
> Is that correct @soumyaDghosh?

Yeah, that's right.

mr-cal: I see.  I like your original implementation the best!  We'll see what @cmatsuoka thinks

cmatsuoka: Ah I see the list issue. My concern was about having the same validation twice in the code, but given the list vs scalar issue we could leave it as is.

cmatsuoka (approved):

PR #4899, opened 2024-07-08
fix: setup Python env for tiobe ci

sergiusens:
Also shallow checkout like the other workflows

-----

### DIFF
--- a/.github/workflows/tics.yaml
+++ b/.github/workflows/tics.yaml
@@ -11,6 +11,13 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.10'
 
       - name: Install dependencies
         run: |


.github/workflows/tics.yaml
```diff
@@ -11,6 +11,13 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+          with:
+            fetch-depth: 0
+
+      - name: Setup Python
```
lengau, line 17: Is this for consistency or is it failing with the system python3.10?

mr-cal (approved):

.github/workflows/tics.yaml
sergiusens: I copied the shallow checkout to potentially fix https://github.com/canonical/snapcraft/actions/runs/9844524213/job/27178103576 which is a version parsing issue (my view is that if configured the same it should work the same)

The python thing is to keep it consistent with the pull_request triggered flow, by keeping them in sync it is much easier to anticipate errors as it is not easy to test something that runs on main pushes only

cmatsuoka (approved):

lengau (approved):

PR #4915, opened 2024-07-13
Add "-f" for rm to ignore file not found error

AristoChen:
Sometimes there may be no `{install_dir}/modules/*/source`, so I think it would be better to add `-rf` for `rm` command, just like how it is used in other lines of the same file(_kernel_build.py)

### DIFF
--- a/snapcraft_legacy/plugins/v2/_kernel_build.py
+++ b/snapcraft_legacy/plugins/v2/_kernel_build.py
@@ -1111,7 +1111,7 @@ def _arrange_install_dir_cmd(install_dir: str) -> List[str]:
             # but snapd expects modules/ and firmware/
             mv {install_dir}/lib/modules {install_dir}/
             # remove symlinks modules/*/build and modules/*/source
-            rm {install_dir}/modules/*/build {install_dir}/modules/*/source
+            rm -f {install_dir}/modules/*/build {install_dir}/modules/*/source
             # if there is firmware dir, move it to snap root
             # this could have been from stage packages or from kernel build
             [ -d {install_dir}/lib/firmware ] && mv {install_dir}/lib/firmware {install_dir}
--- a/tests/unit/parts/plugins/test_kernel.py
+++ b/tests/unit/parts/plugins/test_kernel.py
@@ -1919,7 +1919,7 @@ _finalize_install_cmd = [
         # but snapd expects modules/ and firmware/
         mv ${CRAFT_PART_INSTALL}/lib/modules ${CRAFT_PART_INSTALL}/
         # remove symlinks modules/*/build and modules/*/source
-        rm ${CRAFT_PART_INSTALL}/modules/*/build ${CRAFT_PART_INSTALL}/modules/*/source
+        rm -f ${CRAFT_PART_INSTALL}/modules/*/build ${CRAFT_PART_INSTALL}/modules/*/source
         # if there is firmware dir, move it to snap root
         # this could have been from stage packages or from kernel build
         [ -d ${CRAFT_PART_INSTALL}/lib/firmware ] && mv ${CRAFT_PART_INSTALL}/lib/firmware ${CRAFT_PART_INSTALL}
--- a/tests/legacy/unit/plugins/v2/test_kernel.py
+++ b/tests/legacy/unit/plugins/v2/test_kernel.py
@@ -1768,7 +1768,7 @@ _finalize_install_cmd = [
         # but snapd expects modules/ and firmware/
         mv ${SNAPCRAFT_PART_INSTALL}/lib/modules ${SNAPCRAFT_PART_INSTALL}/
         # remove symlinks modules/*/build and modules/*/source
-        rm ${SNAPCRAFT_PART_INSTALL}/modules/*/build ${SNAPCRAFT_PART_INSTALL}/modules/*/source
+        rm -f ${SNAPCRAFT_PART_INSTALL}/modules/*/build ${SNAPCRAFT_PART_INSTALL}/modules/*/source
         # if there is firmware dir, move it to snap root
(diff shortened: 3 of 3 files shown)


mr-cal (changes requested):
:+1: 

You'll need to update the unit tests in `tests/unit/parts/plugins/test_kernel.py` to match this change.

snapcraft_legacy/plugins/v2/_kernel_build.py
```diff
@@ -1111,7 +1111,7 @@ def _arrange_install_dir_cmd(install_dir: str) -> List[str]:
             # but snapd expects modules/ and firmware/
             mv {install_dir}/lib/modules {install_dir}/
             # remove symlinks modules/*/build and modules/*/source
-            rm {install_dir}/modules/*/build {install_dir}/modules/*/source
+            rm -rf {install_dir}/modules/*/build {install_dir}/modules/*/source
```
dariuszd21, line 1114: If that's the symlink, I'd avoid adding `-r` as it seems unnecessary in that case:

```suggestion
            rm -f {install_dir}/modules/*/build {install_dir}/modules/*/source
```

AristoChen: hi @dariuszd21 , 

thanks for the feedback! previously I was considering adding `-f` only, the reason is same as your opinion, but I noticed that `-rf` was used in the same file when using `rm`, such as

https://github.com/canonical/snapcraft/blob/a9badaa32be6bcfa24f99302a3d6dffecd96063b/snapcraft_legacy/plugins/v2/_kernel_build.py#L739

and

https://github.com/canonical/snapcraft/blob/a9badaa32be6bcfa24f99302a3d6dffecd96063b/snapcraft_legacy/plugins/v2/_kernel_build.py#L1048

The above cases are obviously removing a file instead of a directory, so using `-r` seems meaningless. But in order to make sure the consistency in that file and prevent any regression(though looks like there won't be), I changed my mind to use `-rf`.

Or do you want me to remove all `-r` for those are actually not removing a directory?

dariuszd21: `-rf` raise a flag for me :D
I don't think it's necessary to remove it everywhere else as a part of this PR, but to keep it consistent and compatible with previous implementation I'd just stick with `-f` (or use `-e` guard, to skip removing if it does not exist).

dariuszd21: In that case if it will be unexpectedly the directory it will fail, instead of removing it silently :smile:

AristoChen: Got it, I just force-pushed the changes, also updated the `tests/unit/parts/plugins/test_kernel.py`  as @mr-cal has requested, please let me know if you have any concerns, thanks!

mr-cal (approved):

tests/unit/parts/plugins/test_kernel.py
```diff
@@ -1919,7 +1919,7 @@ def _is_sub_array(array, sub_array):
         # but snapd expects modules/ and firmware/
         mv ${CRAFT_PART_INSTALL}/lib/modules ${CRAFT_PART_INSTALL}/
         # remove symlinks modules/*/build and modules/*/source
-        rm ${CRAFT_PART_INSTALL}/modules/*/build ${CRAFT_PART_INSTALL}/modules/*/source
+        rm -r ${CRAFT_PART_INSTALL}/modules/*/build ${CRAFT_PART_INSTALL}/modules/*/source
```
dariuszd21, line 1922:
```suggestion
        rm -f ${CRAFT_PART_INSTALL}/modules/*/build ${CRAFT_PART_INSTALL}/modules/*/source
```

dariuszd21 (changes requested):

tests/unit/parts/plugins/test_kernel.py
AristoChen: Hi @dariuszd21 , thanks for that, I've fixed it

dariuszd21 (approved):

tests/unit/parts/plugins/test_kernel.py
dariuszd21: Looks good to me

mr-cal:
Can you update the legacy unit test for this as well?  Thanks

```diff
diff --git a/tests/legacy/unit/plugins/v2/test_kernel.py b/tests/legacy/unit/plugins/v2/test_kernel.py
index 939316c14..0b7729ad8 100644
--- a/tests/legacy/unit/plugins/v2/test_kernel.py
+++ b/tests/legacy/unit/plugins/v2/test_kernel.py
@@ -1768,7 +1768,7 @@ _finalize_install_cmd = [
         # but snapd expects modules/ and firmware/
         mv ${SNAPCRAFT_PART_INSTALL}/lib/modules ${SNAPCRAFT_PART_INSTALL}/
         # remove symlinks modules/*/build and modules/*/source
-        rm ${SNAPCRAFT_PART_INSTALL}/modules/*/build ${SNAPCRAFT_PART_INSTALL}/modules/*/source
+        rm -f ${SNAPCRAFT_PART_INSTALL}/modules/*/build ${SNAPCRAFT_PART_INSTALL}/modules/*/source
         # if there is firmware dir, move it to snap root
         # this could have been from stage packages or from kernel build
         [ -d ${SNAPCRAFT_PART_INSTALL}/lib/firmware ] && mv ${SNAPCRAFT_PART_INSTALL}/lib/firmware ${SNAPCRAFT_PART_INSTALL}
```

AristoChen:
Hi @mr-cal 

rebased and force-pushed with the change that you have requested, thanks!

PR #4925, opened 2024-07-17
add: extension env-injector

locnnil:

-----

Add extension to inject environment variables into snaps by its command-chain.
This extension takes snap options and transform them into environment variables


snapcraft/extensions/env_injector.py
```diff
@@ -0,0 +1,135 @@
...
+
+            }
+        }
+
+    def get_toolchain(self):
+        # Dictionary mapping architecture names
+        toolchain = {
+            'amd64': 'x86_64-unknown-linux-musl',
+            'arm64': 'aarch64-unknown-linux-musl',
+            'armhf': 'arm-unknown-linux-musleabihf',
```
farshidtz, line 129: Snaps aren't supported on armv6. The `armhf` arch refers to armv7; change to `armv7-unknown-linux-musleabihf`.

https://doc.rust-lang.org/rustc/platform-support.html

snapcraft/extensions/env_injector.py
```diff
@@ -0,0 +1,135 @@
...
+    
+    @overrides
+    def get_app_snippet(self, *, app_name: str) -> Dict[str, Any]:
+        """Return the app snippet to apply."""
+        return {
+            "command-chain": [ "bin/command-chain/env-exporter"],
+            "environment": {
+                "env_alias": f"{app_name}",
+            }
+        }
```
farshidtz, line 81: This depends on:
- #4926

snapcraft/extensions/env_injector.py
```diff
@@ -0,0 +1,135 @@
...
+
+      rustup default stable
+      rustup target add {toolchain}
+
+      cargo build --target {toolchain} --release
+      mkdir -p $SNAPCRAFT_PART_INSTALL/bin/command-chain
+
+      cp target/{toolchain}/release/env-exporter $SNAPCRAFT_PART_INSTALL/bin/command-chain
+
+      # shrink the binary
```
farshidtz, line 115: -> compress the binary

(There is no suggestion button available.)

snapcraft/extensions/env_injector.py
```diff
@@ -0,0 +1,135 @@
...
+        toolchain = self.get_toolchain()
+        return {
+            f"env-injector/env-injector": {
+                "source": f"{get_extensions_data_dir()}/env-injector",
+                "plugin": "nil",
+                "build-snaps": [
+                    f"rustup",
+                ],
+                "build-packages": [
+                    f"musl-tools",
```
farshidtz, line 102: add comment: `# for static linking`

snapcraft/extensions/env_injector.py
```diff
@@ -0,0 +1,135 @@
...
+    def get_toolchain(self):
+        # Dictionary mapping architecture names
+        toolchain = {
+            'amd64': 'x86_64-unknown-linux-musl',
+            'arm64': 'aarch64-unknown-linux-musl',
+            'armhf': 'arm-unknown-linux-musleabihf',
+            'riscv64': 'riscv64gc-unknown-linux-musl',
+            'ppc64el': 'powerpc64-unknown-linux-gnu',
+            's390x': 's390x-unknown-linux-musl',
+            'i386': 'i586-unknown-linux-musl',
```
farshidtz, line 133: I'm not sure how reliable are these builds. The targets are taken from [Tier 2](https://doc.rust-lang.org/rustc/platform-support.html#tier-2-with-host-tools) which are "guaranteed to build". Why not chose from [Tier 1](https://doc.rust-lang.org/rustc/platform-support.html#tier-1-with-host-tools) that are "guaranteed to work" targets?

farshidtz (changes requested):

snapcraft/extensions/env_injector.py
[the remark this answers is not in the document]
locnnil: I had to make a choice here: If you see [the list of available Rust toolchains](https://rust-lang.github.io/rustup-components-history/), the only toolchain available for i386 is a Tier 3 toolchain `i386-apple-ios` which doesn't fit our needs here.

So the closest we have of i386 is i586. But I can also remove this architecture of the list in this case.

locnnil: In order to static linking, we need the toolchains with the `*-musl` prefix, the tier 1 has two problems: It only have `*-gnu` toolchains and only have support for `amd64` and `arm64`, also the tendency is that the Tier 2 toolchains became each time more realiable over the time, if you see [the list of toolchains](https://rust-lang.github.io/rustup-components-history/) there is already a Tier 2.5 which the documentation doesn't mention. In other words: We have to work with the tools that we have.

locnnil: Actually, there is another way to do static linking using gnu based toolchains, but it doesn't get close to the size of  when using the `musl` ones because of the size cost when static linking glibc, [please take a look at this analysis](https://gist.github.com/locnnil/c6f7dad389444ded0928bfc4c6f027aa).

snapcraft/extensions/env_injector.py
locnnil: As suggested by @mr-cal, the dependency [was cherry-picked](https://github.com/canonical/snapcraft/pull/4925/commits/b4853bb889616a6571b5f624b80c797d903eec44) from the PR #4926.

snapcraft/extensions/env_injector.py
```diff
@@ -0,0 +1,138 @@
...
+            "environment": {
+                "env_alias": f"{app_name}",
+            }
+        }
+
+    @overrides
+    def get_part_snippet(self, *, plugin_name: str) -> Dict[str, Any]:
+        return {
+             "build-environment": [
+                 { "VARS_INJECT_EXTENSION": "true" }
```
artivis, line 87: The extension is `env_injector`, wouldn't it make sens to rename this accordingly? (`ENV_INJECTOR_EXTENSION`).
Furthermore, how about prepending `SNAPCRAFT_` to it to further improve readability?

extensions/env-injector/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "env-exporter"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+http-body-util = "0.1.2"
+hyper = {git = "https://github.com/hyperium/hyper", branch = "master"}
```
lengau, line 8: How reliably stable are these branches? Especially `hyperlocal`, since it's pre-1.0.

I think I'd prefer to have these use frozen versions and rely on renovate updating these for snapcraft. (For `hyper` "any 1.x release >= some number" plus relying on renovate for major releases would also be fine IMO.)

extensions/env-injector/src/main.rs
lengau, line 1: In case anyone's wondering about build times: This increased my build time on a nonsense snap on my laptop by about 30 seconds (including downloads - 15 seconds without). 

To compare: packing snapcraft itself with everything cached (no re-build, not even re-creating the venv) takes 25 seconds, so this is pretty small.

snapcraft/extensions/env_injector.py
```diff
@@ -0,0 +1,132 @@
...
+    def get_parts_snippet(self) -> Dict[str, Any]:
+        toolchain = self.get_toolchain()
+        if toolchain is None:
+            raise ValueError(
+                f"Unsupported architecture for env-injector extension: {self.arch}"
+            )
+        return {
+            "env-injector/env-injector": {
+                "source": f"{get_extensions_data_dir()}/env-injector",
+                "plugin": "nil",
```
lengau, line 95: Why not use the `rust` plugin here?

extensions/env-injector/src/main.rs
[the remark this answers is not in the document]
locnnil: > Is it possible to do this without a noticeable performance hit with a bash or python script (since those are included in core20-core24)?

At first, we try it with bash script, but it introduces a fixed **40ms latency** with additional **10ms** to each environment variable set.

extensions/env-injector/Cargo.toml
locnnil: Thanks for the suggestion. This was during the development/hacking phase, and I ended up forgetting to use a fixed tag/commit hash. I agree with your point, and I have now changed it to a fixed tag.

extensions/env-injector/src/main.rs
locnnil: Yeah, it's necessary to make a trade-off here. We decided to optimize binary size at the cost of slightly longer compilation time.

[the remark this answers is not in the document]
lengau: Cool, thanks for the answer.

snapcraft/extensions/env_injector.py
locnnil: I've tried, but I ended up running into problems, such as a binary showing this error:

```
env-exporter: symbol lookup error: /lib/x86_64-linux-gnu/libc.so.6: undefined symbol: __tunable_is_initialized, version GLIBC_PRIVATE
```

This is what I've tried:
```rs
        return {
            "env-injector/env-injector": {
                "source": f"{get_extensions_data_dir()}/env-injector",
                "plugin": "rust",
                "rust-use-global-lto": True,
                "rust-channel": "stable",
                "build-snaps": [
                    "rustup",
                ],
                "build-packages": [
                    "upx-ucl",  # for binary compression
                ],
                "override-build": f"""

      rustup default stable
      craftctl default
      mkdir -p $SNAPCRAFT_PART_INSTALL/bin/command-chain

      cp target/{toolchain}/release/env-exporter $SNAPCRAFT_PART_INSTALL/bin/command-chain

      # compress the binary
      upx --best --lzma target/{toolchain}/release/env-exporter
      cp target/{toolchain}/release/env-exporter $SNAPCRAFT_PART_INSTALL/bin/command-chain/env-exporter-upx
                """,
            }
        }

```

snapcraft/extensions/env_injector.py
locnnil: This has a little hacking that I did to test some stuff, I've removed. But thanks for the suggestion.

snapcraft/extensions/env_injector.py
lengau: Cool, thanks for the explanation.

lengau (approved):

extensions/env-injector/src/main.rs
```diff
@@ -0,0 +1,161 @@
...
+        for (key, value) in process_env(&serde_json::Value::Object(global_env.clone())) {
+            std::env::set_var(key, value.trim_matches('"'));
+        }
+    }
+
+    if let Some(app_env) = stdout_json["apps"][app]["env"].as_object() {
+        for (key, value) in process_env(&serde_json::Value::Object(app_env.clone())) {
+            std::env::set_var(key, value.trim_matches('"'));
+        }
+    }
```
farshidtz, line 71: This also processes env keys that contain dots, treated as a hierarchy by snapd.
Setting a snap option such as `env.x.bad=BAD` results in an unwanted export for `X={"bad":"test"}`.

tests/spread/extensions/env-injector/task.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+  snap install "${SNAP}"_1.0_*.snap --dangerous
+
+  # Create envfile
+  echo 'HELLO_WORLD="Hello, World"' >> envfile.env
+
+  # Set env vars: Global, App specific, and envfile
+  snap set env-injector-hello env.hello="Hello"
+  snap set env-injector-hello apps.hello.env.world="World"
+  snap set env-injector-hello envfile="${SNAP_DIR}"/enfile.env
```
farshidtz, line 49: Other needed test cases:
- snap options keys with hyphen converted to equivalent
- `<env>` key with dot rejected (ignored with error message): e.g. `env.hello.world`
- Order of exporting variables (those from env files overridden by snap options)

tests/spread/extensions/env-injector/task.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+  # Run the hello command
```
farshidtz, line 51:
```suggestion
  # Run the hello command
  # It will fail if environment variables aren't set as expected
```

snapcraft/extensions/env_injector.py
```diff
@@ -0,0 +1,132 @@
...
+
+    - To set environment variables for a specific application **inside** the snap:
+
+    .. code-block:: shell
+        sudo snap set <snap-name> apps.<app-name>.env.<key>=<value>
+
+    - To set environment file inside the snap:
+
+    .. code-block:: shell
+        sudo snap set <snap-name> env-file=<path-to-env-file>
```
farshidtz, line 48: Missing documentation on app-specific env file.

farshidtz (changes requested):
Thanks, I've tested and everything works as expected.

I have a few concerns, on the testing, error handling, and documentation.

tests/spread/extensions/snaps/env-injector-hello/snap/snapcraft.yaml
```diff
@@ -0,0 +1,18 @@
...
+description: This is a basic snap for testing env-injector extension
+
+grade: devel
+base: core24
+confinement: strict
+
+apps:
+  env-hello:
+    command: usr/local/bin/hello
+    extensions: [ env-injector ]
```
farshidtz, line 13: Add another app that set's its own `env_alias` and then test that we aren't overriding it.

snapcraft/extensions/env_injector.py
```diff
@@ -0,0 +1,132 @@
...
+            )
+        return {
+            "env-injector/env-injector": {
+                "source": f"{get_extensions_data_dir()}/env-injector",
+                "plugin": "nil",
+                "build-snaps": [
+                    "rustup",
+                ],
+                "build-packages": [
+                    "musl-tools",  # for static linking
```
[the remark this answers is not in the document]
locnnil: Solved! Thank you!

tests/spread/extensions/env-injector/task.yaml
locnnil: Added a comment on it! Thanks!

snapcraft/extensions/env_injector.py
locnnil: Added! Thank you!

extensions/env-injector/src/main.rs
locnnil: Fixed! Thank you!

tests/spread/extensions/env-injector/task.yaml
locnnil: Addressed! Thanks.

tests/spread/extensions/snaps/env-injector-hello/hello.c
```diff
@@ -4,34 +4,66 @@
...
             fprintf(stderr, "Got: %s\n", env);
-            return 1;
+            exit(1);
         }
     }
 
+    // Check that it's not possible to access other app ENVs 
+    const char *env_specific = getenv("SPECIFIC");
+    if (env_specific != NULL) {
+        fprintf(stderr, "\n[ERROR] Env. variable SPECIFIC is accessible from the app.\n");
```
farshidtz, line 40: Can't use [println](https://doc.rust-lang.org/std/macro.println.html)?

tests/spread/extensions/snaps/env-injector-hello/snap/snapcraft.yaml
```diff
@@ -0,0 +1,28 @@
...
+    extensions: [ env-injector ]  
+
+  hello2:
+    command: usr/local/bin/hello2
+    extensions: [ env-injector ]
+
+  hello-demo:
+    command: usr/local/bin/hello2
+    environment:
+      app_alias: hello-demo
```
farshidtz, line 22:
```suggestion
      # user-defined alias
      app_alias: myapp
```

tests/spread/extensions/env-injector/task.yaml
```diff
@@ -0,0 +1,68 @@
...
+  echo 'ORDER="From envfile"' >> envfile.env
+  snap set env-injector-hello apps.hello1.env.order="from app-specific"
+  snap set env-injector-hello envfile="${SNAP_DIR}"/envfile.env
+
+  # Run the hello command
+  # Those will fail if the env vars are not set as expected
+  env-injector-hello.hello1
+  env-injector-hello.hello2
+
+  # Test if the alias it's rewrite
```
farshidtz, line 66:
```suggestion
  # Make sure that alias is NOT rewritten
```

farshidtz: How is this checking that the alias isn't rewritten? The set application alias is same as the app name, so there is no way to know if it was rewritten. The alias should be different, as that's the whole point of allowing user-defined aliases.

Then the test can look into squashfs-root/meta/snap.yaml to check the environment before starting configurations. Alternatively, it can continue configuring it and then run it with an expectation to fail (when alias is rewritten).

tests/spread/extensions/env-injector/task.yaml
```diff
@@ -0,0 +1,68 @@
...
+  # To Check if key with dot will be rejected 
+  snap set env-injector-hello env.word.dot="wrong"
+
+  # To test order of env vars
+  echo 'ORDER="From envfile"' >> envfile.env
+  snap set env-injector-hello apps.hello1.env.order="from app-specific"
+  snap set env-injector-hello envfile="${SNAP_DIR}"/envfile.env
+
+  # Run the hello command
+  # Those will fail if the env vars are not set as expected
```
farshidtz, line 62:
```suggestion
  # These will fail if the env vars are not set as expected
```

tests/spread/extensions/env-injector/task.yaml
```diff
@@ -0,0 +1,68 @@
...
+  echo 'HELLO_WORLD=Hello World' >> envfile.env
+
+  # Set env vars: Global
+  snap set env-injector-hello env.global="World"
+
+  # Set env vars: specific to each app
+  snap set env-injector-hello apps.hello1.env.hello="Hello"
+  snap set env-injector-hello apps.hello2.env.specific="City"
+
+  # To Check if key with dot will be rejected
```
farshidtz, line 53:
```suggestion
  # To check if key with dot will be rejected 
```

farshidtz (changes requested):

tests/spread/extensions/env-injector/task.yaml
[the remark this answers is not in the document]
farshidtz: This is still not being tested.

extensions/env-injector/src/main.rs
```diff
@@ -43,20 +43,51 @@ async fn snapdapi_req() -> Result<serde_json::Value, Box<dyn Error + Send + Sync
...
+            _ => {
+                eprintln!(
+                    "ERROR: Invalid type for environment variable value detected.\n\
+                        val: {}\n\
+                        Reason: It must be a string, number or boolean\n\
+                        Action: Skipped\n",
+                    k
+                );
+                continue;
+            },
```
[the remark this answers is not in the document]
locnnil: Actually, there is a tiny difference, the fist just checks if there isn't dots (which is a nested JSON) or arrays passed.

The second one does the type conversion. But I think it would be better to just always convert to a string.

tests/spread/extensions/snaps/env-injector-hello/hello.c
locnnil: But this is **C** code and not **Rust**!

extensions/env-injector/src/main.rs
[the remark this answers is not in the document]
farshidtz: If it is neither a nested object nor an array, then it is a primitive type. No?

extensions/env-injector/src/main.rs
```diff
@@ -43,20 +43,51 @@ async fn snapdapi_req() -> Result<serde_json::Value, Box<dyn Error + Send + Sync
...
+        .ok_or("Expected an object (JSON input)").unwrap();
+    let mut map = HashMap::new();
+
+    for (k, v) in obj {
+        if v.is_object() || v.is_array() {
+            eprintln!(
+                "ERROR: Invalid value detected.\n\
+                    Key: {}\n\
+                    Reason: Environment variable names must not contain dots.\n\
+                    Action: Skipped\n",
```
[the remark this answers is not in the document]
locnnil: Agree! Fixed! Thank you

snapcraft/extensions/env_injector.py
```diff
@@ -46,6 +46,9 @@ class EnvInjectorExtension(Extension):
 
     .. code-block:: shell
         sudo snap set <snap-name> env-file=<path-to-env-file>
+    - To set environment file for a specific app:
+    .. code-block:: shell
+        sudo snap set <snap-name> apps.<app-name>.envfile=<path-to-env-file>
```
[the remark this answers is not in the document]
locnnil: Fixed. Thanks.

extensions/env-injector/src/main.rs
```diff
@@ -43,20 +43,51 @@ async fn snapdapi_req() -> Result<serde_json::Value, Box<dyn Error + Send + Sync
...
+            serde_json::Value::String(s) => s.clone(),
+            serde_json::Value::Number(n) => n.to_string(),
+            serde_json::Value::Bool(b) => b.to_string(),
+            _ => {
+                eprintln!(
+                    "ERROR: Invalid type for environment variable value detected.\n\
+                        val: {}\n\
+                        Reason: It must be a string, number or boolean\n\
+                        Action: Skipped\n",
+                    k
```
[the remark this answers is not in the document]
locnnil: Fixed! Thank you!

tests/spread/extensions/snaps/env-injector-hello/snap/snapcraft.yaml
locnnil: Done! Thanks!

tests/spread/extensions/env-injector/task.yaml
locnnil: Fixed. Thanks

extensions/env-injector/src/main.rs
```diff
@@ -0,0 +1,175 @@
...
+    let reader = std::io::BufReader::new(file);
+
+    let mut env_vars = HashMap::new();
+    for line in reader.lines() {
+        let line = line?;
+        if !line.trim().is_empty() && !line.starts_with('#') {
+            if let Some((key, value)) = line.split_once('=') {
+                env_vars.insert(key.to_string(), value.to_string());
+            }
+        }
```
farshidtz, line 113: This is taking an inline comment as part of the value.
For example:
```bash
X=val # This is the comment
```
exports `X` with value: `val # This is the comment`.

Please refer [here](https://docs.docker.com/compose/environment-variables/variable-interpolation/#env-file-syntax) for the syntax.

farshidtz (changes requested):
Thanks for the fixes.
I've just discovered another issue while testing.

extensions/env-injector/src/main.rs
locnnil: Good catch! Fixed thanks!

farshidtz (approved):
Looks good. Thanks a lot.

Nevertheless, I do have concerns over the licensing of snaps using this extension. The extension adds a binary distribution to the snap which uses the following libraries:
- https://crates.io/crates/dotenv (MIT)
- https://docs.rs/http-body-util/latest/http_body_util/index.html (MIT)
- https://github.com/hyperium/hyper (MIT)
- https://github.com/hyperium/hyper-util (MIT)
- https://github.com/softprops/hyperlocal (MIT)
- https://github.com/serde-rs/serde (Apache 2.0 or MIT)
- https://docs.rs/serde_json/latest/serde_json/ (Apache 2.0 or MIT)
- https://github.com/tokio-rs/tokio (MIT)

We need to take a deeper look into this before flagging the extension as stable. I believe a snap using this extension needs to list MIT as one of its licenses and bundle a copy of the license files for all the above. A simpler solution would be to go back to the Bash implementation (at the expense of performance) with a royalty-free license such as CC0.

tests/spread/extensions/env-injector/task.yaml
```diff
@@ -0,0 +1,84 @@
+summary: Build and run a basic hello-world snap using extensions
+
+systems:
+  - ubuntu-24.04
```
mr-cal, line 4: Spread is tricky to work with.  This isn't running in CI because you need a wildcard to match against `google:ubuntu-24.04-64`:

```suggestion
  - ubuntu-24.04*
```

tests/spread/extensions/env-injector/task.yaml
```diff
@@ -0,0 +1,84 @@
...
+  # Create global envfile
+  echo 'HELLO_WORLD=Hello World' >> envfile.env
+
+  echo "[env-injector] Creating app-specific envfile"
+  # Create app-specific envfile
+  echo 'SCOPED=Scoped' >> appenv.env
+
+  echo "[env-injector] Setting global env variable"
+  # Set env vars: Global
+  snap set env-injector-hello env.global="World"
```
mr-cal, line 53: What was the decision on a configure hook?  If I understand correctly, this will fail because there is no configure hook.

mr-cal (commented):
Overall, I think the extension and spread test look good, but I'm not sure how it will work without a configure hook.

Also, this will need a `tests/unit/extensions/test_env_injector.py`.

I would like to return to one of the original criteria.  We still need a snapcraft.io forum post about this idea to get input and feedback on this approach.  We can adjust this PR so it gets published to a branch for testing.

tests/spread/extensions/env-injector/task.yaml
locnnil: Sorry, you're right, this is missing a configure hook. 

We're not including it in the extension because the user may already have one. However, if they don't, snapd will notify them that a configure hook is required.

tests/spread/extensions/env-injector/task.yaml
locnnil: Awesome! Thank you!

tests/spread/extensions/env-injector/task.yaml
farshidtz: The design envisioned that the extension would add an empty configure hook when it wasn't present. But this wasn't straight forward because the configure hook could be added from the project source or during the parts build. Detecting the lack of a configure hook from source and a part is possible, as suggestion [here](https://matrix.to/#/!GGqzbFAUQprdPgYYCM:ubuntu.com/$3upkz8HmiNl-hcw2il79j9GWEFORRQ7XPQ_1EtL4jdc?via=ubuntu.com&via=matrix.org&via=kde.org) by first checking `$CRAFT_PROJECT_DIR/snap/hooks/configure` and also adding a part that runs `after` all others to check if no other part has primed it (e.g. a compiled hook or one added from a different source).

In the end, we decided to leave this out for this version of the implementation and instead document that the extension should be used with a snap that has a configure hook. (In the meantime, we could look into why it is even mandatory to have a configure hook for a snap to accept configurations, where many snaps work around it by adding an empty script that does no config validations).

farshidtz:
@mr-cal 
> I would like to return to one of the original criteria. We still need a snapcraft.io forum post about this idea to get input and feedback on this approach. We can adjust this PR so it gets published to a branch for testing.

We have done extensive research on the existing practices. The plan was to get this added as an experimental extension to allow widespread testing (assuming this is the whole point of experimental extensions) and validate the proposed configuration schema for the snap options keys. A forum post is a great idea to request community feedback. Will publishing to a branch make this extension available for testing via stable `snapcraft`?

mr-cal:
> We have done extensive research on the existing practices. The plan was to get this added as an experimental extension to allow widespread testing (assuming this is the whole point of experimental extensions) and validate the proposed configuration schema for the snap options keys. A forum post is a great idea to request community feedback. Will publishing to a branch make this extension available for testing via stable `snapcraft`?

It would be available in a branch along the lines of `edge/PR-4925`.  From a conversation between Gustavo and @cmatsuoka last week, I believe that feedback and agreement is needed prior to landing the PR. I think we have a misalignment/miscommunication - @cmatsuoka, can you chime in here?

tests/spread/extensions/env-injector/task.yaml
```diff
@@ -0,0 +1,84 @@
...
+  echo "[env-injector] Load envfiles"
+  # Load global and app-specific envfiles
+  snap set env-injector-hello envfile="${SNAP_DIR}"/envfile.env
+  snap set env-injector-hello apps.hello1.envfile="${SNAP_DIR}"/appenv.env
+
+  echo "[env-injector] Run hello1 and hello2 apps"
+  # Run the hello command
+  # These will fail if the env vars are not set as expected
+  env-injector-hello.hello1
+  env-injector-hello.hello2
```
cmatsuoka, line 79: I think the test evaluation shouldn't be inside the test applications in this case. They could just relay back what they see, and we can verify the correctness in this script, otherwise you end up with the test split in two parts, one in shell script and other in C making it harder to follow and modify.

snapcraft/extensions/env_injector.py
```diff
@@ -0,0 +1,135 @@
...
+      rustup target add {toolchain}
+
+      cargo build --target {toolchain} --release
+      mkdir -p $SNAPCRAFT_PART_INSTALL/bin/command-chain
+
+      cp target/{toolchain}/release/env-exporter $SNAPCRAFT_PART_INSTALL/bin/command-chain
+
+      # compress the binary
+      upx --best --lzma target/{toolchain}/release/env-exporter
+      cp target/{toolchain}/release/env-exporter $SNAPCRAFT_PART_INSTALL/bin/command-chain/env-exporter-upx
```
cmatsuoka, line 118: This could be simply unecessary as snap payloads are already compressed. Additionally, we can have other ways to produce leaner binaries that should be investigated (e.g. can we go the no_std way?)

Also we need the reasoning to implement this as a binary (as opposed to e.g. a shell script) explained somewhere. Having it as a script could make the need for an additional part unnecessary, and the whole extension would end up being very trivial, maybe to the point that any changes to snapcraft.yaml could be done manually by the user.

In the binary case, it's better to have the injector source in a separate repository instead of bundling it with snapcraft, and just point the part source to it.

snapcraft/extensions/env_injector.py
```diff
@@ -0,0 +1,135 @@
...
+        return {}
+
+    @overrides
+    def get_app_snippet(self, *, app_name: str) -> Dict[str, Any]:
+        """Return the app snippet to apply."""
+        return {
+            "command-chain": ["bin/command-chain/env-exporter"],
+            "environment": {
+                "env_alias": f"{app_name}",
+            },
```
cmatsuoka, line 82: This is a very simple snippet and snap authors are quite used to apply even more complex patterns to parts or applications in order to implement certain features, so this alone may not warrant the existence of an extension.

cmatsuoka (changes requested):
Thanks! I think we need more discussion on this before landing to main, but even for a feature branch it would be interesting to refactor the test case to avoid splitting the extension test between the spread task scriptlet and verifications done inside C code, ideally both sides should reside in the task scriptlet (and maybe the C code can be entirely dropped).

Also compressing the rust binary can be pointless since the snap payload is already compressed (we should investigate other ways to reduce the binary size if a binary is to be use at all.)

cmatsuoka:
Let's not forget that we still need this to be presented and discussed in the forum before landing it to main. In any case the injector code should reside in a separate repository instead of being part of the snapcraft codebase, and we must carefully consider if the extension is needed at all (the changes to snapcraft.yaml look too simple and can be easily added manually by the snap author).

tests/spread/extensions/env-injector/task.yaml
locnnil: I agree! I'll change this to use `/usr/bin/env` inside the snap instead of the C app.

snapcraft/extensions/env_injector.py
farshidtz: The selected snippet adds the env exporter program and the alias. The alias gets added only when the user hasn't set it (this is intended and based on snapcraft's internal handling of the dictionary).

The extension does more than that: It adds the source code for the exporter program, builds it and places it in the right location. The design also envisions additional functions, such as adding a configure hook when it isn't a present, or adding configuration validation. 

If this was to be done manually, without the extension, the developers would have to:
1. Fetch the exporter source from **trusted location**, pick and appropriate **version**, and hope that it stays **compatible with the snap**. *
2. Add build logic to compile and place the executable in a suitable location *
3. For each application that needs the functionality:
   i. Add the exporter program to the command chain
  ii. Set the application alias

In comparison, when the extension is used:
1. Add the extension to each application that needs the functionality
2. Add alias if an app's name needs to be overridden

The manual way requires average snapping knowledge, which early adopters lack. I'd emphasize that the goal is to simplify the snapping of apps and all this complexity thrown at the developer defeats its purpose.

---

\* We could make a snap or a debian package and let the developer stage it, but then we need to separately maintain that package.

farshidtz:
> Let's not forget that we still need this to be presented and discussed in the forum before landing it to main.

Sure. We weren't aware of this requirement and assumed the reviewed spec was sufficient.

> we must carefully consider if the extension is needed at all (the changes to snapcraft.yaml look too simple and can be easily added manually by the snap author).

I don't think what the extension does is too simple. Please see my response: https://github.com/canonical/snapcraft/pull/4925#discussion_r1713366664

tests/spread/extensions/env-injector/task.yaml
locnnil: Adressed on https://github.com/canonical/snapcraft/pull/4925/commits/9d95567ca5fe108e973551b841422d3dbc076ef8

Thank you for the suggestion!

tests/spread/extensions/snaps/env-injector-hello/hello.c
```diff
@@ -0,0 +1,80 @@
...
+    for (int i = 0; i < num_envs; ++i) {
+        const char *env = getenv(envs[i]);
+
+        if (env == NULL) {
+            fprintf(stderr, "\n[ERROR] Env. variable %s is not set.\n", envs[i]);
+            exit(1);
+        }
+
+        if (strcmp(env, expected[i]) != 0) {
+            fprintf(stderr, "\n[ERROR] Env. variable %s isn't set to the expected value.\n", envs[i]);
```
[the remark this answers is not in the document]
locnnil: Addressed on https://github.com/canonical/snapcraft/pull/4925/commits/9d95567ca5fe108e973551b841422d3dbc076ef8

Thank you for the suggestion!

tests/spread/extensions/env-injector/task.yaml
```diff
@@ -0,0 +1,132 @@
...
+  # [TEST] - Check if the app-specific env var IS SET for the app hello2
+  check_env_var "env-injector-hello.hello2" "SPECIFIC" "City" || exit 1
+  # [TEST] - Check if the app-specific env var IS NOT SET for the app hello1
+  ! check_env_var "env-injector-hello.hello1" "SPECIFIC" "City" || exit 1
+
+  echo "[env-injector] Rejecting invalid keys"
+  # To check if key with dot will be rejected
+  snap set env-injector-hello env.word.dot="wrong"
+  # [TEST] - Check if the key with dot was rejected
+  ! check_env_var "env-injector-hello.hello1" "DOT" "wrong" || exit 1
```
farshidtz, line 117: The test should be to check if the key was "ignored", rather than rejected. The exporter program doesn't reject.

For this test, I'd rather look for "wrong" in the whole env dump to make sure it wasn't exported with an unexpected name, such as WORDDOT, WORD_DOT, WORD. Maybe add a special handling in the function for when the second argument is empty, then call it:
```suggestion
  # [TEST] - Check if the key with dot was ignored
  ! check_env_var "env-injector-hello.hello1" "" "wrong" || exit 1
```

tests/spread/extensions/env-injector/task.yaml
```diff
@@ -0,0 +1,132 @@
...
+
+  echo "[env-injector] Testing order of env vars"
+  # To test order of env vars
```
farshidtz, line 120: Redundant. The message above makes this clear.
```suggestion
```

tests/spread/extensions/env-injector/task.yaml
```diff
@@ -0,0 +1,132 @@
...
+  [ -f squashfs-root/bin/command-chain/env-exporter ]
+  
+  # Check that the exec-env script is present
+  [ -f squashfs-root/usr/bin/exec-env ]
+
+  snap install "${SNAP}"_1.0_*.snap --dangerous --classic
+
+  echo "[env-injector] Creating global envfile"
+  touch global.env
+  echo 'HELLO_WORLD="Hello World"' >> global.env
```
farshidtz, line 70:
```suggestion
  echo 'HELLO_WORLD="Hello World"' >> global.env
```
or 
```suggestion
  echo 'HELLO_WORLD="Hello World"' > global.env
```
?

Also in several other places.

extensions/env-injector/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "env-exporter"
+version = "0.1.0"
+edition = "2021"
```
farshidtz, line 4:
```suggestion
edition = "2024"
```

farshidtz (changes requested):

extensions/env-injector/Cargo.toml
locnnil: The Rust 2024 edition [is not released yet](https://doc.rust-lang.org/edition-guide/rust-2024/index.html).

tests/spread/extensions/env-injector/task.yaml
locnnil: I was having some errors about the file not being found, this was before I realize that it was a problem with the snap sandboxing.  :smile:

extensions/env-injector/Cargo.toml
locnnil: Also, editions are the way that Rust [guarantees backward compatibility](https://doc.rust-lang.org/edition-guide/editions/index.html#what-are-editions).

tests/spread/extensions/env-injector/task.yaml
locnnil: Addressed! Thank you!

tests/spread/extensions/env-injector/task.yaml
locnnil: Addressed on https://github.com/canonical/snapcraft/pull/4925/commits/979889baf426a72237451834bfe7a3773d565526. Thank you!

tests/spread/extensions/env-injector/task.yaml
```diff
@@ -0,0 +1,132 @@
...
+
+  echo "[env-injector] Setting app-specific env variable"
+  # Set env vars: specific to each app
+  snap set env-injector-hello apps.hello1.env.hello="Hello"
+  snap set env-injector-hello apps.hello2.env.specific="City"
+
+  # [TEST] - Check if the app-specific env var IS SET for the app hello1 
+  check_env_var "env-injector-hello.hello1" "HELLO" "Hello" || exit 1
+  # [TEST] - Check if the app-specific env var IS NOT SET for the app hello2
+  ! check_env_var "env-injector-hello.hello2" "HELLO" "Hello" || exit 1
```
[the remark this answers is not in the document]
locnnil: Changed to `assert_env` thanks for the suggestion!

tests/spread/extensions/env-injector/task.yaml
```diff
@@ -0,0 +1,132 @@
...
+  echo 'ORDER="From envfile"' >> local.env
+  local_envpath=$(realpath local.env)
+  snap set env-injector-hello apps.hello1.env.order="from app-specific"
+  snap set env-injector-hello apps.hello1.envfile="$local_envpath"
+  # [TEST] - Check if local overrites global 
+  check_env_var "env-injector-hello.hello1" "ORDER" "from app-specific" || exit 1
+
+  echo "[env-injector] Run hello-demo app"
+  snap set env-injector-hello apps.myapp.env.specific="City"
+  # [TEST] Make sure that alias is NOT rewritten
```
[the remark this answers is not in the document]
locnnil: Fixed! Thank you!

tests/spread/extensions/env-injector/task.yaml
```diff
@@ -0,0 +1,132 @@
...
+  unsquashfs "${SNAP}"_1.0_*.snap
+
+
+  # Check that the env-exporter program is present
+  [ -f squashfs-root/bin/command-chain/env-exporter ]
+  
+  # Check that the exec-env script is present
+  [ -f squashfs-root/usr/bin/exec-env ]
+
+  snap install "${SNAP}"_1.0_*.snap --dangerous --classic
```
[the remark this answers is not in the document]
locnnil: > Why should this use classic confinement?

The confinement in which it is tested is irrelevant for these tests, since the extension is made to work on all confinements. Right?

Anyway, I've changed this to `devmode` installation.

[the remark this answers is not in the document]
locnnil: Here: https://github.com/canonical/snapcraft/pull/4925/commits/ab60705ffe6533e1d9d0afbdccbd75e8dfecd719 I tried to use the common directory, [but it fails anyway](https://github.com/canonical/snapcraft/actions/runs/10423760020/job/28871293025?pr=4925#step:4:352). 

So I reverted it to `classic` confinement, since it also satisfies the purpose of this test.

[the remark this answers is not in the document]
farshidtz: The path is wrong. It should be /var/snap/<snap>/common/

https://snapcraft.io/docs/data-locations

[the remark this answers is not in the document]
farshidtz: Also, the global file name on line 76 is different, missing the .env extension.

Classic confinement satisfies the purpose, but this is a special type of snap used here for the wrong purpose. By using a strictly confined snap the tests can verify that an env file can be read in confinement by the exporter.

locnnil:
@mr-cal, @cmatsuoka since this is not being merged into the `main`, could you please create a new branch (something like `work/env-injector-extension`) so I can use it as the base for this PR? Thank you in advance.

mr-cal:
> @mr-cal, @cmatsuoka since this is not being merged into the `main`, could you please create a new branch (something like `work/env-injector-extension`) so I can use it as the base for this PR? Thank you in advance.

Yep, we can get it into a feature branch this week to make testing easier.

mr-cal (approved):
From a code perspective, I think this is well written and tested.  The documentation looks great too, although I didn't see anything about needing a configure hook.

Approving so we can land this into a feature branch this week to facilitate testing.

locnnil:
@mr-cal Good point! [I've added a note section](https://forum.snapcraft.io/t/the-env-injector-extension/41477) about the need for a configure hook. Thank you!

locnnil:
@mr-cal Just saw your update! Thanks for letting me know! I'll be waiting for @cmatsuoka feedback.

extensions/env-injector/src/main.rs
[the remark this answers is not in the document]
cmatsuoka: Just for reference, what hardware was this measured on?

cmatsuoka (changes requested):
The remaining restriction I have is that the injector code should reside in a separate repository, as a stand-alone project, instead of being part of the Snapcraft codebase. That way, it can be maintained separately without requiring Snapcraft maintainers to own the injector code. After establishing it as a separate project we can land the extension code on a feature branch for tests.

Before landing on main we'll probably need to revisit the bash vs rust implementations of the injector to see if the former can be optimized to be faster, or the latter can be optimized to be smaller (e.g. using no_std and adding some local helpers).

extensions/env-injector/src/main.rs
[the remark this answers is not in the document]
locnnil: @cmatsuoka This was tested on @farshidtz computer: with a `AMD Ryzen 7 5700G CPU`

I did these tests in my computer as well: `AMD Ryzen 9 7900` /  32GB of DDR5 memory (CMK32GX5M2B5200C40)

In my tests, each environment variable adds 7~8 ms.

snapcraft/extensions/env_injector.py
```diff
@@ -0,0 +1,129 @@
...
+                rustup target add {toolchain}
+
+                cargo build --target {toolchain} --release
+                mkdir -p $SNAPCRAFT_PART_INSTALL/bin/command-chain
+
+                cp target/{toolchain}/release/env-exporter $SNAPCRAFT_PART_INSTALL/bin/command-chain
+
+                # compress the binary
+                upx --best --lzma target/{toolchain}/release/env-exporter
+                cp target/{toolchain}/release/env-exporter $SNAPCRAFT_PART_INSTALL/bin/command-chain/env-exporter-upx
```
farshidtz, line 114: This is actually copying both the original and compressed binary to the snap, where only the original one is used in the command chain on line 79.

```suggestion
                # compress the binary
                upx --best --lzma target/{toolchain}/release/env-exporter
                
                cp target/{toolchain}/release/env-exporter $SNAPCRAFT_PART_INSTALL/bin/command-chain
```

farshidtz (changes requested):

snapcraft/extensions/env_injector.py
```diff
@@ -0,0 +1,128 @@
...
+    def get_parts_snippet(self) -> Dict[str, Any]:
+        toolchain = self.get_toolchain(self.arch)
+        if toolchain is None:
+            raise ValueError(
+                f"Unsupported architecture for env-injector extension: {self.arch}"
+            )
+
+        return {
+            "env-injector/env-injector": {
+                "source": "https://github.com/canonical/snappy-env.git",
```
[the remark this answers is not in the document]
locnnil: Fixed!

locnnil:
@cmatsuoka The snappy-env (the exporter program) code is now separated from the extension and resides here: https://github.com/canonical/snappy-env

farshidtz (changes requested):
I've done some smoke testing and it looks good. There is a change to multipass command flags for spread testing which I think shouldn't be here.

Not a major issue, but env-exporter has been renamed to snappy-env (for good reasons since it also loads env files), but the surrounding refactoring is incomplete. It is still called env-exported in a few places. The extension adds a part named `env-injector/env-injector` which could also change to `env-injector/snappy-env`.

spread.yaml
```diff
@@ -82,7 +82,7 @@ backends:
         FATAL "$SPREAD_SYSTEM is not supported!"
       fi
 
-      multipass launch --disk 20G --memory 2G --name "$instance_name" "$image"
+      multipass launch --cpus 8 --disk 30G --memory 8G --name "$instance_name" "$image"
```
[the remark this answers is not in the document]
locnnil: This change can easily cause a computer to run out of memory. In this repository, spread tests run in parallel for Focal, Jammy, and Noble, so it will be 3x8GB = 24GB.

Additionally, there is no guarantee that most computers will have 8 cores available—1 CPU is always guaranteed. Therefore, these changes are suitable only when applied locally on a machine that is guaranteed to have those resources.

farshidtz (approved):

mr-cal (approved):

snapcraft/extensions/env_injector.py
```diff
@@ -0,0 +1,129 @@
...
+                "build-packages": ["upx-ucl"],  # for binary compression
+                "override-build": f"""
+                rustup default stable
+                rustup target add {toolchain}
+
+                cargo build --target {toolchain} --release
+                mkdir -p $SNAPCRAFT_PART_INSTALL/bin/command-chain
+
+                # compress the binary
+                upx --best --lzma target/{toolchain}/release/env-exporter
```
cmatsuoka, line 112: Almost there, we only need to remove this because the snap payload is already compressed.

locnnil: Fixed on https://github.com/canonical/snapcraft/pull/4925/commits/2642669c58f298d4c01fc2722c3c58f0444c6bde. Thank you.

cmatsuoka (approved):
Thank you. Now we can work on the exporter independently, and investigate if it could be no-std to reduce binary size, or experiment and compare to a fast shell script implementation.

Issue #4917, opened 2024-07-15
Prototype `edit-registry-sets` and `list-registry-sets`


mr-cal:
### What needs to get done

1. Work with store team to get access to an environment for testing API calls
2. Create a generic `AssertionService`
    - Migrate and generalize logic from `snapcraft.commands.validation_sets` for editing assertions
    - Migrate and generalize logic from `snapcraft_legacy.cli.assertions` for listing assertions
3. Create a child class `RegistryAssertionService` for registry assertions
4. Create a new command `edit-registry-sets`
5. Create a new command `list-registry-sets`
6. Add new API calls to `craft-store`(?) for:
    - `/api/v2/registries/build-assertion`
    - `/api/v2/registries`
    - `/api/v2/registries/{name}`

### Why it needs to get done

1. To support listing and editing registry assertions
2. To enable a future migration of validation assertions from `snapcraft_legacy` to `snapcraft`/`craft-application`

mr-cal:
- Keeping this as a prototype because the scope of work is too large for a single PR (it will probably be broken into ~5 separate PRs)
- Estimation is based on a 1-week timebox

mr-cal:
Store service for registries is broken and is scheduled to be fixed in pulse 16.  

This is blocking us from testing against the store's API for registries.

mr-cal:
Prototype: https://github.com/canonical/snapcraft/tree/work/CRAFT-3116/registry-sets-prototype

Outstanding store issues:
* `revision` is not returned if it is 0 from `/api/v2/registries/build-assertion` and `api/v2/registries[/name]`
* `body` is not returned from `api/v2/registries[/name]`

PR #4928, opened 2024-07-18
fix: use correct types for validation set models

### LINKED ISSUE #4909: Fix revision type in `edit-validation-sets`
### What needs to get done

1. Work with the store team to create or get access to a brand store
    - If possible, this should be documented
2. Test editing validation sets with snap revisions
3. Update unit tests to include snap revisions
4. Fix the problem by using the correct types (strings or ints) for each API call

### Why it needs to get done

Validation sets with snap revisions fail when building the assertion with the error:

```
Issues encountered with validation set: '123' is not of type 'integer' at /snaps/0/revision
```

This may occur because when editing a validation set, Snapcraft makes the API calls:

1. `GET /api/v2/validation-sets` where `snaps.revision` is a string in the response
2. `POST /api/v2/validation-sets/build-assertion` where `snaps.revision` is an integer in the request but a string in the response
3. `POST /api/v2/validation-sets` where `snaps.revision` is a string in the request

Looks like it's failing on step #2 where it sends a string instead of an int

source: https://dashboard.snapcraft.io/docs/reference/v2/en/validation-sets.html

mr-cal:

-----

1. Uses the correct data types for the correct calls (#4909)
2. Switch to pydantic models
3. Make the boilerplate validation set valid (previously, using the boilerplate would raise an error)

I'll be using this code as groundwork for #4917 and I have an alternative fix [here](https://github.com/canonical/snapcraft/commit/9489504bd6d955aed56b3afbbba019f108b78efe) that could act as a short-term fix.

Fixes #4909
(CRAFT-3103)


snapcraft_legacy/storeapi/v2/validation_sets.py
```diff
@@ -14,161 +14,119 @@
...
+    if isinstance(data, dict):
+        return {k: _to_string(v) for k, v in data.items()}
+    elif isinstance(data, list):
+        return [_to_string(i) for i in data]
+    elif data:
+        return str(data)
+    return data
+
+
+class Snap(models.CraftBaseModel):
```
lengau, line 47: Nice!

snapcraft_legacy/storeapi/v2/validation_sets.py
```diff
@@ -14,161 +14,119 @@
...
+
+def cast_dict_values_to_strings(data: dict[str, Any]) -> dict[str, Any]:
+    """Cast all scalars in a dictionary to strings."""
+    return {k: _to_string(v) for k, v in data.items()}
+
+
+def _to_string(data: list | str | int | None) -> dict | list | str | None:
+    """Recurse through nested lists and cast all scalars to strings."""
+    if isinstance(data, dict):
+        return {k: _to_string(v) for k, v in data.items()}
```
lengau, line 39: What do you expect the output of `_to_string({1: "1"})` to be?

snapcraft_legacy/storeapi/v2/validation_sets.py
```diff
@@ -14,161 +14,119 @@
...
+    elif isinstance(data, list):
```
lengau, line 40:
```suggestion
    if isinstance(data, list):
```

snapcraft_legacy/storeapi/v2/validation_sets.py
```diff
@@ -14,161 +14,119 @@
...
+        return [_to_string(i) for i in data]
+    elif data:
```
lengau, line 42:
```suggestion
    if data:
```

snapcraft_legacy/storeapi/v2/validation_sets.py
```diff
@@ -14,161 +14,119 @@
...
+        return str(data)
+    return data
```
lengau, line 44: `_to_string(0)` seems like it would output `0` rather than `"0"`

mr-cal:
I forgot to mention during standup - this branch has been tested by the bug reporter and is working for them.

snapcraft_legacy/storeapi/v2/validation_sets.py
mr-cal: Good point.  I changed the code to look for common scalars (int, float, bool) and updated the docstring to reflect that.

snapcraft_legacy/storeapi/v2/validation_sets.py
mr-cal: Nice catch.

Even though dictionary keys for these pydantic models will all be strings, I updated this function to cast keys to strings and added some extra tests.

snapcraft_legacy/storeapi/v2/validation_sets.py
```diff
@@ -14,161 +14,130 @@
...
+
+    Supported scalar values are str, int, float, and bool.
+    """
+    if isinstance(data, dict):
+        return {_to_string(key): _to_string(value) for key, value in data.items()}
+
+    if isinstance(data, list):
+        return [_to_string(i) for i in data]
+
+    if isinstance(data, (int, float, bool)):
```
lengau, line 52:
```suggestion
    if isinstance(data, (numbers.Number, bool)):
```

[Abstract base classes](https://docs.python.org/3/library/numbers.html) FTW!

snapcraft_legacy/storeapi/v2/validation_sets.py
```diff
@@ -14,161 +14,130 @@
...
+
+
+def _to_string(
+    data: dict | list | str | int | float | str | bool | None
+) -> dict[str, Any] | list | str | None:
+    """Recurse through nested dicts and lists and cast scalar values to strings.
+
+    Supported scalar values are str, int, float, and bool.
+    """
+    if isinstance(data, dict):
```
lengau, line 46: It might be worth starting out with the trivial case for performance:

```suggestion
    if isinstance(data, str):
        return data

    if isinstance(data, dict):
```

lengau (approved):
Looks good!

snapcraft/commands/validation_sets.py
```diff
@@ -127,32 +132,43 @@ def run(self, parsed_args: "argparse.Namespace"):
...
     store_client: StoreClientCLI,
 ) -> None:
+    emit.debug(f"Posting assertion to build: {edited_validation_sets.json()}")
     build_assertion = store_client.post_validation_sets_build_assertion(
-        validation_sets=edited_validation_sets
+        validation_sets=edited_validation_sets.marshal()
     )
-    signed_validation_sets = _sign_assertion(
-        build_assertion.marshal(), key_name=key_name
+    build_assertion_dict = build_assertion.marshal_as_str()
```
cmatsuoka, line 143: It sounds a bit confusing that something marshaled "as str" ends up being a dict.

snapcraft/commands/validation_sets.py
```diff
@@ -174,25 +190,33 @@ def _generate_template(
...
+                filepath=Path("validation-sets"),
             )
             return edited_validation_sets
-        except yaml.YAMLError as yaml_error:
-            emit.message(f"A YAML parsing error occurred {yaml_error!s}")
+        except (yaml.YAMLError, CraftValidationError) as err:
+            emit.message(f"{err!s}")
             if not utils.confirm_with_user("Do you wish to amend the validation set?"):
-                raise errors.SnapcraftError("Operation aborted") from yaml_error
+                raise errors.SnapcraftError("Operation aborted") from err
```
cmatsuoka, line 215: We have established (during the rockcraft error message revamp) that error messages should start with lowercase characters for consistency.

snapcraft_legacy/storeapi/v2/validation_sets.py
```diff
@@ -14,161 +14,135 @@
...
+    revision: str | None
+    """The "revision" assertion header"""
+
+    sequence: int
+    """The "sequence" assertion header"""
+
+    snaps: Annotated[list[Snap], pydantic.Field(min_items=1)]
+    """List of snaps in a Validation Set assertion"""
+
+    def marshal_as_str(self) -> Dict[str, Any]:
```
cmatsuoka, line 99: Ah I see what this is doing. Maybe we should be a bit more explicit about the dict values being converted to strings in the function name?

snapcraft_legacy/storeapi/v2/validation_sets.py
```diff
@@ -14,161 +14,135 @@
...
+    if isinstance(data, str):
+        return data
+
+    if isinstance(data, dict):
+        return {_to_string(key): _to_string(value) for key, value in data.items()}
+
+    if isinstance(data, list):
+        return [_to_string(i) for i in data]
+
+    if isinstance(data, (numbers.Number, bool)):
```
cmatsuoka, line 57: If we declare everything in the model as strings we could convert non-scalar types during validation, and let Pydantic's type coercion to handle scalar types (unless we operate on these values and need the original types). Would that work?

snapcraft/commands/validation_sets.py
mr-cal: Renamed

snapcraft/commands/validation_sets.py
mr-cal: Fixed the existing errors.

cmatsuoka (approved):
Thanks!

snapcraft_legacy/storeapi/v2/validation_sets.py
mr-cal: Discussed offline - the data needs to be an int for some API calls and strings for others, so we will use the correct types in the pydantic model.

PR #4946, opened 2024-07-31
fix: fix Python venv in core24 classic snaps

### LINKED ISSUE #4942: Python plugin is broken in core24
### Bug Description

See [https://forum.snapcraft.io/t/upgrading-classic-snap-to-core24-using-snapcraft-8-3-causes-python-3-12-errors-at-runtime/41045](https://forum.snapcraft.io/t/upgrading-classic-snap-to-core24-using-snapcraft-8-3-causes-python-3-12-errors-at-runtime/41045) 
In short, it looks like the changes to virtual environments in Python 3.12 are not working with the python plugin.

### To Reproduce

.

### Environment

.

### snapcraft.yaml

```shell
.
```

### Relevant log output

```shell
.
```

### Additional context

We still need confirmation but [this fix](https://forum.snapcraft.io/t/upgrading-classic-snap-to-core24-using-snapcraft-8-3-causes-python-3-12-errors-at-runtime/41045/11?u=tigarmo) should work and is feasible for us.

tigarmo:
Changes to virtual environment handling made our default approach - creating a venv in the build system and then just moving it into the snap - not work for Python 3.12.

My understanding of the issue is that the "home" key in the "pyvenv.cfg" file points to a path in the build system, which obviously does not exist at snap-run-time. This apparently works fine *somehow* in core22 - likely there's some fallback logic that makes the interpreter correctly find the bundled Python libraries, but not in core24. So to address this we update this "home" key to point to its final destination in the snap.

Fixes #4942

-----

### DIFF
--- a/snapcraft/parts/plugins/python_plugin.py
+++ b/snapcraft/parts/plugins/python_plugin.py
@@ -71,8 +71,8 @@ class PythonPlugin(python_plugin.PythonPlugin):
         """Perform Python-specific actions right before packing."""
         base = step_info.project_base
 
-        if base != "core24":
-            # Only fix pyvenv.cfg on core24 snaps
+        if base in ("core20", "core22"):
+            # Only fix pyvenv.cfg on core24+ snaps
             return
 
         root_path: Path = step_info.prime_dir
@@ -89,9 +89,11 @@ class PythonPlugin(python_plugin.PythonPlugin):
             step_info.stage_dir,
         )
 
-        contents = pyvenv.read_text()
+        old_contents = contents = pyvenv.read_text()
         for candidate in candidates:
             old_home = f"home = {candidate}"
             contents = contents.replace(old_home, new_home)
 
-        pyvenv.write_text(contents)
+        if old_contents != contents:
+            logger.debug("Updating pyvenv.cfg to:\n%s", contents)
+            pyvenv.write_text(contents)
--- a/snapcraft/parts/plugins/python_plugin.py
+++ b/snapcraft/parts/plugins/python_plugin.py
@@ -17,9 +17,10 @@
 """The Snapcraft Python plugin."""
 
 import logging
+from pathlib import Path
 from typing import Optional
 
-from craft_parts import errors
+from craft_parts import StepInfo, errors
 from craft_parts.plugins import python_plugin
 from overrides import override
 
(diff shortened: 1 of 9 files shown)


tigarmo:
I think the spread failures are transient (store issues looks like)

tests/spread/core24-suites/environment/test-variables/task.yaml
```diff
@@ -33,7 +33,7 @@ execute: |
     cat "$file"
     for exp in \
       "^SNAPCRAFT_PROJECT_GRADE=devel$" \
-      "^SNAPCRAFT_PROJECT_NAME=None$" \
+      "^SNAPCRAFT_PROJECT_NAME=variables$" \
```
tigarmo, line 36: I think I fixed a bug here by accident

mr-cal: Great!  I think this is different than https://github.com/canonical/craft-application/issues/320, which is about `CRAFT_PROJECT_NAME`.

snapcraft/parts/plugins/python_plugin.py
```diff
@@ -64,3 +65,33 @@ def _get_system_python_interpreter(self) -> Optional[str]:
             confinement,
         )
         return interpreter
+
+    @classmethod
+    def post_prime(cls, step_info: StepInfo) -> None:
+        """Perform Python-specific actions right before packing."""
+        base = step_info.project_base
+
+        if base != "core24":
```
mr-cal, line 74: Since only core22 and newer snaps will use this code, what do you think about flipping this logic?

This is to help future me when `core26` get released.

```suggestion
        if base == "core22":
```


Also random thought - a future refactor could implement `__eq__()` and `__gt__()` for bases so we could do:
`if base >= core24:`

snapcraft/parts/plugins/python_plugin.py
```diff
@@ -64,3 +65,33 @@ def _get_system_python_interpreter(self) -> Optional[str]:
...
+            step_info.part_install_dir,
+            step_info.stage_dir,
+        )
+
+        contents = pyvenv.read_text()
+        for candidate in candidates:
+            old_home = f"home = {candidate}"
+            contents = contents.replace(old_home, new_home)
+
+        pyvenv.write_text(contents)
```
mr-cal, line 97: WDYT about adding a `debug` statement so users are informed that snapcraft is modifying `pyvenv.cfg` after it is primed?

mr-cal:
Some of these spread failures are fixed on main and not `hotfix/8.3`.  I'll get those cherry-picked.

mr-cal (commented):
Looks good, have a few questions

snapcraft/parts/plugins/python_plugin.py
lengau: Regarding the future refactor: Once snapcraft includes `craft-platforms`, you could do:

````python
if craft_platforms.snap.get_distro_base_from_core_base(base) >= craft_platforms.DistroBase("ubuntu", "24.04"):
```

lengau (approved):
Nice fix!

snapcraft/parts/plugins/python_plugin.py
tigarmo: @mr-cal you are totally right. I struggled with this a bit because the code that calls this `post_prime()` method is on the LifecycleService, which means it doesn't get called for core22 and older - so the check could be seen as superfluous
Regardless you are correct about the future-proofing so I changed it to `if base in ("core20", "core22")` to act as a bit of documentation too

snapcraft/parts/plugins/python_plugin.py
tigarmo: ITIAGI (I think it's a good idea)

mr-cal (approved):
Looks great!

mr-cal:
Spread failures unrelated - squash-merging

PR #4950, opened 2024-08-01
fix: configure icon for snaps without apps

mr-cal:
- [x] Have you followed the [guidelines for contributing](https://github.com/snapcore/snapcraft/blob/master/CONTRIBUTING.md)?
- [x] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [x] Have you successfully run `tox run -m lint`?
- [x] Have you successfully run `tox run -e test-py310`? (supported versions: `py39`, `py310`, `py311`, `py312`)

-----

Always configure and install icons for snaps even if they do not have any `apps` defined in their `snapcraft.yaml` file.

This looks like a regression introduced in core22 [here](https://github.com/canonical/snapcraft/pull/3702/files#diff-0eaeb159bf22a93ac8d7fe8db572236d6a88d712904c605f2d458495bfcc7dc1R52) based on [this legacy code](https://github.com/canonical/snapcraft/blob/780f2aadfb260048e58a1b2d6892bc0b3f5d9b0e/snapcraft_legacy/internal/meta/_snap_packaging.py#L515).  

Fixes https://github.com/canonical/snapcraft/issues/4936

Issue #4936, opened 2024-07-24
snapcraft 7 ignores icon path in snapcraft.yaml


mapero:
### Bug Description

I have minimal example of a snapcraft.yaml
```
name: icon-snap
base: core22
version: '0.1'
summary: icon snap
icon: icon/icon.png
description: minimal
grade: stable
confinement: strict
parts:
  my-part:
    plugin: nil
```

It seems like the icon is not copied when there is no "apps" section defined in the snapcraft.yaml in snapcraft 7. With snapcraft 6 or legacy the icon is copied into the snap:

With snapcraft 6 / legacy
```
prime/
└── meta
    ├── gui
    │   └── icon.png
    └── snap.yaml

2 directories, 2 files
```

With snapcraft 7
```
prime/
└── meta
    ├── gui
    └── snap.yaml

2 directories, 1 file
```

### To Reproduce

Build the following snap with snapcraft 7:

```
name: icon-snap
base: core22
version: '0.1'
summary: icon snap
icon: icon/icon.png
description: minimal
grade: stable
confinement: strict
parts:
  my-part:
    plugin: nil
```

### Environment

Ubuntu 22.04 LTS

### snapcraft.yaml

```shell
name: icon-snap
base: core22
version: '0.1'
summary: icon snap
icon: icon/icon.png
description: minimal
grade: stable
confinement: strict
parts:
  my-part:
    plugin: nil
```

### Relevant log output

```shell
NA
```

PsikoBlock:
Relevant source: 
https://github.com/canonical/snapcraft/blob/8c2c801e7beda637e185e576969fd8ffaa9f76d7/snapcraft/parts/setup_assets.py#L75-L80
Lines 75 and 76 should be moved below the call to _finalize_icon to match Snapcraft 6's behavior. This is unchanged in Snapcraft 8.

mr-cal:
Hi @mapero and @PsikoBlock,

Thanks for the bug report and root cause.  This indeed looks like a regression.

I have proposed a fix in #4950.  Do you need this fix in snapcraft 7?  Or is it ok to just fix this in snapcraft 8?

mapero:
Hi @mr-cal ,

we would need this also fixed in snapcraft 7 since we will stick with snapcraft 7 for quite some time. 

Thanks.

mr-cal:
Sounds good, we will target both snapcraft 7 and 8.

mr-cal:
[...]

PR #4954, opened 2024-08-06
tests: update spread tests in hotfix/7.5

mr-cal:

-----

6 cherry-picks and 6 new commits to repair CI and spread tests in `hotfix/7.5`.

The readthedocs CI failure is expected as Snapcraft 7 predates readthedocs integration.

Unblocks #4950 and #4936
Unblocks CRAFT-3189

### DIFF
--- a/.github/workflows/cla-check.yaml
+++ b/.github/workflows/cla-check.yaml
@@ -3,7 +3,7 @@ on: [pull_request]
 
 jobs:
   cla-check:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - name: Check if CLA signed
         uses: canonical/has-signed-canonical-cla@v1
--- a/.github/workflows/cla-check.yaml
+++ b/.github/workflows/cla-check.yaml
@@ -3,7 +3,7 @@ on: [pull_request]
 
 jobs:
   cla-check:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     steps:
       - name: Check if CLA signed
         uses: canonical/has-signed-canonical-cla@v1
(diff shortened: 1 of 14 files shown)


lengau (approved):

tests/spread/plugins/v1/ruby/snaps/ruby-bundle-install/snap/snapcraft.yaml
```diff
@@ -16,6 +16,8 @@ parts:
     plugin: ruby
     source: .
     use-bundler: true
-    ruby-version: "2.6.10" # bundler requires ruby >= 2.6.0
+    ruby-version: "3.3.4" # bundler requires ruby >= 3.0.0
     build-environment:
       - BUNDLE_GEMFILE: "$SNAPCRAFT_PART_SRC/Gemfile"
+    build-packages:
```
mr-cal, line 22: @cmatsuoka, should I add this as a build-package when `use-bundler: true` is defined?  

Another solution would be to add a new property `ruby-bundler-version` so projects can be pinned to an older version of the bundler that isn't dependent on `libyaml-dev`.

Or I could do nothing - snap maintainers would just have to add `build-package: [libyaml-dev]` to their project.

mr-cal: Per a conversation with @cmatsuoka, we can:

* add a `bundler-version` property.  If it is not defined, then install the latest version
* check which version of the bundler required `libyaml-dev`.  If `bundler-version` is not set or is greater than that version, then snapcraft will include `libyaml-dev` as a build-package

When we port the ruby plugin to craft-parts, we can redesign `ruby-use-bundler` to accept a version string instead of a boolean flag.

.github/workflows/cla-check.yaml
```diff
@@ -3,7 +3,7 @@ on: [pull_request]
 
 jobs:
   cla-check:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
lengau, line 6: starbase uses `ubuntu-latest` for this

lengau (approved):

dariuszd21 (approved):
Looks good!

PR #5005, opened 2024-08-27
fix(remotebuild): error early for multiple artifacts per build-on

### LINKED ISSUE #4995: remote build: error if multiple artefacts can be built on the same architecture
### What needs to get done

Check if the build plan will create multiple artefacts on any particular architecture.  If it will, then raise an error before sending the project to Launchpad.

### Why it needs to get done

Launchpad cannot build multiple artefacts on the same architecture.  It prepares a build environment and runs Snapcraft in destructive mode without specifying `--build-for` or `--platform`.  Snapcraft will refuse to run if multiple builds aretfacts can be built.

This is a poor UX because it occurs after the project is uploaded to Launchpad and requires the user to read the logfile downloaded from Launchpad.

The following fails in Launchpad:
```
platforms:
  amd64:
    build-on: amd64
    build-for: amd64
  platform2:
    build-on: amd64
    build-for: riscv64
```

mr-cal:

-----

Launchpad cannot build multiple artifacts on the same build-on architecture and will fail with an esoteric error. 

Snapcraft now checks for the scenario and fails with an informative error before the build begins.

Fixes #4995
(CRAFT-3279)

### DIFF
--- a/tests/unit/commands/test_remote.py
+++ b/tests/unit/commands/test_remote.py
@@ -858,6 +858,7 @@ def test_platform_core22_error(
     ],
 )
 def test_multiple_artifacts_per_build_on(
+    check,
     base,
     build_info,
     error_messages,
@@ -879,12 +880,12 @@ def test_multiple_artifacts_per_build_on(
 
     _, err = capsys.readouterr()
 
-    assert (
-        "Remote build does not support building multiple snaps on the same architecture"
-        in err
+    check.is_in(
+        "Remote build does not support building multiple snaps on the same architecture",
+        err,
     )
     for message in error_messages:
-        assert message in err
+        check.is_in(message, err)
 
 
 ########################
--- a/tests/unit/commands/test_remote.py
+++ b/tests/unit/commands/test_remote.py
@@ -769,6 +769,124 @@ def test_platform_core22_error(
     assert "Use '--build-for' instead." in err
 
 
+@pytest.mark.parametrize(
+    ("base", "build_info", "error_messages"),
+    [
+        pytest.param(
+            "core22",
+            {
+                "architectures": [
+                    {"build-on": "amd64", "build-for": "amd64"},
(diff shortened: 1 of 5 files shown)


tests/unit/commands/test_remote.py
```diff
@@ -769,6 +769,124 @@ def test_platform_core22_error(
...
+        pytest.param(
+            "core24",
+            {
+                "platforms": {
+                    "platform1": {"build-on": "amd64", "build-for": "riscv64"},
+                    "identical-platform": {"build-on": "amd64", "build-for": "riscv64"},
+                },
+            },
+            # this will show 'riscv64' twice because the platform name is different
+            ["Building on 'amd64' will create snaps for 'riscv64' and 'riscv64'."],
```
mr-cal, line 825: This could be improved but I'm leaving this as-is because:
- it tells the user everything they need, even if it may look odd
- this is a 3-point complexity ticket
- this is an uncommon scenario
- our time would be better spent creating a `__str__()` method for the `BuildInfo` class in craft-application

tests/unit/commands/test_remote.py
```diff
@@ -769,6 +769,124 @@ def test_platform_core22_error(
...
+    assert app.run() == os.EX_CONFIG
+
+    _, err = capsys.readouterr()
+
+    assert (
+        "Remote build does not support building multiple snaps on the same architecture"
+        in err
+    )
+    for message in error_messages:
+        assert message in err
```
dariuszd21, line 887: You may consider wrapping it with `check` context manager to check always all the possibilities before failing the test

dariuszd21 (approved):
Looks good to me!

tests/unit/commands/test_remote.py
lengau: Yeah, this whole test would be a good use case for `pytest_check`

lengau (approved):

PR #5111, opened 2024-10-11
feat: use `craft-platforms` for build plans

mattculler:
- [ ] Have you followed the [guidelines for contributing](https://github.com/canonical/snapcraft/blob/main/CONTRIBUTING.md)?

-----

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -103,7 +103,7 @@ install_requires = [
     "craft-cli~=2.6",
     "craft-grammar>=2.0.1,<3.0.0",
     "craft-parts>=2.1.2,<3.0.0",
-    "craft-platforms @ file://localhost/home/mrc/dev/main/craft-platforms#egg=craft_platforms",
+    "craft-platforms @ git+https://github.com/canonical/craft-platforms@f055619f473a5f86b5e111d77dd90c2340c9bbb3", # ~=0.3.1",
     "craft-providers>=2.0.4,<3.0.0",
     "craft-store>=3.0.2,<4.0.0",
     "docutils<0.20",  # Frozen until we can update sphinx dependencies.
--- a/setup.py
+++ b/setup.py
@@ -103,7 +103,7 @@ install_requires = [
     "craft-cli~=2.6",
     "craft-grammar>=2.0.1,<3.0.0",
     "craft-parts>=2.1.2,<3.0.0",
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -103,7 +103,7 @@ def recursive_data_files(directory, install_directory):
     "craft-cli~=2.6",
     "craft-grammar>=2.0.1,<3.0.0",
     "craft-parts>=2.1.2,<3.0.0",
-    "craft-platforms~=0.1",
+    "craft-platforms~=0.4.0",
```
lengau, line 106: This would require a manual update to 0.5.0, but I think (hope?) that the snapcraft build planner can be considered to have a stable API unless we need snapcraft-specific changes (in which case we'd be changing snapcraft in lockstep), so I think we can probably say:

```suggestion
    "craft-platforms~=0.4",
```

This would be equivalent to `>=0.4.0,<1.0.0` rather than `>=0.4.0,<0.5.0`.

Not a strong preference though, more a note to ensure we're doing what you intended.

snapcraft/models/project.py
```diff
@@ -1221,16 +1223,35 @@ def get_build_plan(self) -> list[BuildInfo]:
...
+                platform=buildinfo.platform,
+                build_on=str(buildinfo.build_on),
+                build_for=str(buildinfo.build_for),
+                base=bases.BaseName(
+                    name=buildinfo.build_base.distribution,
+                    version=buildinfo.build_base.series,
+                ),
+            )
+            for buildinfo in snap.get_platforms_snap_build_plan(  # pyright: ignore[reportCallIssue]
+                base=self.base,  # type: ignore[arg-type]  # pyright: ignore[reportArgumentType]
```
lengau, line 1252: I really thought that:

```suggestion
                base=self.base,  # type: ignore[arg-type]
```

would silence pyright's reportArgumentType too?

lengau (approved):

setup.py
mattculler: Yeah that should work.

snapcraft/models/project.py
mattculler: Hey it seems like you're right!

mr-cal (approved):
Thanks!

PR #5135, opened 2024-10-28
fix(test): add patchelf to dev dependencies for testing

bepri:

-----

### DIFF
--- a/requirements-devel.txt
+++ b/requirements-devel.txt
@@ -84,6 +84,7 @@ oauthlib==3.2.2
 overrides==7.7.0
 packaging==24.1
 PasteDeploy==3.1.0
+patchelf==0.17.2.1
 pathspec==0.12.1
 pbr==6.0.0
 pexpect==4.9.0
--- a/tools/environment-setup-local.sh
+++ b/tools/environment-setup-local.sh
@@ -17,7 +17,6 @@ sudo apt install --yes \
     libxslt1-dev \
     libyaml-dev \
     make \
-    patchelf \
     python3-dev \
     python3-pip \
     python3-venv \


requirements-devel.txt
```diff
@@ -84,6 +84,7 @@ oauthlib==3.2.2
 overrides==7.7.0
 packaging==24.1
 PasteDeploy==3.1.0
+patchelf==0.17.2.1
```
dariuszd21, line 87: As far as I know `requirements*.txt` files are auto-generated and we keep dependencies in `pyproject.toml` or `setup.py`. 
If I'm not mistaken this dependency should be added here, and then you should run `tools/freeze-requirements.sh` script.
https://github.com/canonical/snapcraft/blob/main/setup.py#L61

dariuszd21 (commented):
I'm quite curious what exactly this PR fixes?

bepri:
> I'm quite curious what exactly this PR fixes?

I noticed that after a fresh clone, `tox -f test` would fail on multiple cases from patchelf not being installed. I mentioned it to Callahan and it was explained that we were using the apt package of patchelf, which was part of a weird software supply chain of us packaging our own patchelf with Snapcraft. He suggested I open this PR as a first step towards making this work without our own patchelf and setup script.

"Fix" was probably not the right commit message to include here - I made it before I understood the scope of this.

mr-cal:
> He suggested I open this PR as a first step towards making this work without our own patchelf and setup script.

Thanks for making this PR, this is a good place to discuss.

I'm not sure of the right path forward.  To be totally correct, we would need to build and install this version of patchelf: https://github.com/canonical/patchelf/tree/0.9%2Bsnapcraft as that's what snapcraft uses in it's own snap.  I haven't checked what is different between it and the upstream implementation.

I prefer using patchelf as a python package over a system package to keep snapcraft's dev deps more contained.  On the other hand, if you clone this repo, install the `install` deps only and try to run in destructive mode then snapcraft would fail because patchelf would still be missing.

Finally, I don't like that we have unit tests that require patchelf.

bepri:
Obsolete thanks to #5225

PR #5136, opened 2024-10-29
chore: migrate snapcraft itself to core24

mr-cal:

-----

Migrate snapcraft itself to core24.


setup.py
```diff
@@ -178,7 +178,7 @@ def recursive_data_files(directory, install_directory):
     package_data={
         "snapcraft": ["templates/*"],
     },
-    python_requires=">=3.10",
+    python_requires=">=3.12",
```
mr-cal, line 181: This is breaking the trivy workflow: https://github.com/canonical/snapcraft/actions/runs/12199935616/job/34034945727?pr=5136#step:10:445

I'm wondering if I should revert this change.

bepri: I'm curious where trivy gets py3.10 as its default. It seems like maybe `uv` is [selecting it here](https://github.com/canonical/starflow/blob/9c179fa633d896b32a9cbb76d70373995af627ec/.github/workflows/scan-python.yaml#L130) by default. Maybe that could be bumped?

tigarmo: Looks like [ubuntu-latest](https://github.com/canonical/starflow/blob/main/.github/workflows/scan-python.yaml#L71) is still jammy, which is still Python 3.10
so I guess either we revert this or we parametrize the starflow workflow to accept a python version as input

pyproject.toml
```diff
@@ -29,16 +29,16 @@ plugins = [
 [tool.pyright]
 include = ["snapcraft", "tests"]
 exclude = ["build", "tests/legacy", "tests/spread"]
-pythonVersion = "3.10"
+pythonVersion = "3.12"
 
 [tool.pytest.ini_options]
 minversion = "7.0"
-required_plugins = ["pytest-cov>=4.0", "pytest-mock>=3.10", "pytest-subprocess>=1.4"]
+required_plugins = ["pytest-cov>=4.0", "pytest-mock>=3.12", "pytest-subprocess>=1.4"]
```
tigarmo, line 36: I was going to say this was changed by mistake but apparently pytest-mock 3.12 adds support for python 3.12, heh

snap/snapcraft.yaml
```diff
@@ -1,5 +1,5 @@
 name: snapcraft
-base: core22
+base: core24
```
tigarmo, line 2: there is this lint warning:
```
- classic: usr/lib/x86_64-linux-gnu/librtmp.so.1: ELF rpath should be set to '$ORIGIN:/snap/core24/current/lib/x86_64-linux-gnu'.
```
it might be the same bug I bumped into with rockcraft: possibly two parts are staging this so file

tigarmo (changes requested):

setup.py
tigarmo: well, maybe sit on this [for a little while](https://github.com/actions/runner-images/issues/10636)

mr-cal: Good idea, I'll wait

snap/snapcraft.yaml
mr-cal: I finally figured this out.

This is a minimal reproducer:
```yaml
base: core24

parts:
  part-a:
    plugin: nil
    build-attributes:
      - enable-patchelf
    stage-packages: 
      - git

  part-b:
    plugin: nil
    build-attributes:
      - enable-patchelf
    stage-packages:
      - libffi8
```

### 1. `part-a` builds

The `git` package has the following deps:
```
git → libcurl3t64-gnutls →            libgnutls30t64 → libp11-kit0 → libffi8
git → libcurl3t64-gnutls → librtmp1 → libgnutls30t64 → libp11-kit0 → libffi8
```

`libffi8` is not staged because it is in the core24 snap.  When patchelf runs, it correctly updates `librtmp`'s rpath to point to `/snap/core24/current/lib/x86_64-linux-gnu/`.

### 2. `part-b` builds

`libffi8` is staged because it is explicitly listed.

### 3. the classic linter runs

It sees that `libffi8` in the staged in the snap and that `librtmp`'s rpath isn't pointing to it, so it raises a linter warning.

I think the best solution is to remove `libffi8` as a stage package because it is already in the base snap.

setup.py
mr-cal: Looks like github has updated, it's working now.

snap/snapcraft.yaml
```diff
@@ -68,15 +67,14 @@ parts:
...
-      - "usr/share/git-core"
-      - "usr/lib/x86_64-linux-gnu/libcurl-gnutls*"
-      - "usr/lib/x86_64-linux-gnu/libpsl*"
-      - "usr/lib/x86_64-linux-gnu/librtmp*"
-      - "usr/lib/x86_64-linux-gnu/libnghttp2*"
-      - "usr/lib/x86_64-linux-gnu/libldap*"
-      - "usr/lib/x86_64-linux-gnu/liblber*"
-      - "usr/lib/x86_64-linux-gnu/libsasl2*"
+      - "-usr/bin"
+      - "-usr/share/doc"
```
lengau, line 71: Chisel releases have a requirement to [include the copyright file](https://github.com/canonical/chisel-releases/blob/95a72dca1e15bbb22fa1266f1a833df712cfc87e/slices/git.yaml#L57-L59) - we might need to do that here too.

mr-cal: Is there an official requirement for snaps?

There's not an easy way to do this because the `stage` and `prime` keywords don't allow excluding everything except a particular file.  We would need extglob support to say `-usr/share/doc/*/!(copyright)`.

One alternative is to run `shopt -s extglob; rm -r usr/share/doc/*/!(copyright)` in an override-build script.

tigarmo (approved):

PR #5139, opened 2024-11-06
fix(docs): display examples correctly in RTD

dariuszd21:
- [ ] Have you followed the [guidelines for contributing](https://github.com/canonical/snapcraft/blob/main/CONTRIBUTING.md)?

Some of the commands auto-generated docs were not rendered correctly, this PR addresses that

Before:
image

After:
image

### DIFF
--- a/snapcraft/commands/legacy.py
+++ b/snapcraft/commands/legacy.py
@@ -62,7 +62,8 @@ class StoreLegacyUploadMetadataCommand(LegacyAppCommand):
         If --force is used, it will force the local metadata into the Store,
         ignoring any possible conflict.
 
-        Examples:
+        Examples::
+
             snapcraft upload-metadata my-snap_0.1_amd64.snap
             snapcraft upload-metadata my-snap_0.1_amd64.snap --force
         """
--- a/snapcraft/commands/manage.py
+++ b/snapcraft/commands/manage.py
@@ -56,7 +56,8 @@ class StoreReleaseCommand(AppCommand):
             - <branch> is optional and dynamically creates a channel with a
               specific expiration date.
 
-        Examples:
+        Examples::
+
             snapcraft release my-snap 8 stable
             snapcraft release my-snap 8 stable/my-branch
             snapcraft release my-snap 9 beta,edge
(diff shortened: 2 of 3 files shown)


mr-cal (approved):
Thanks!

As a future enhancement, we should look into doing this programmatically in craft-cli when rendering for plaintext or rst.

mr-cal:
@medubelko - I added you to this because during planning today, we discussed if usage examples and output examples belong in the command's overview.

We didn't reach a consensus and want to get your opinion.

medubelko:
@mr-cal 

> doing this programmatically

Wouldn't this require another parser, or at least another step in the doc stack?

> we discussed if usage examples and output examples belong in the command's overview.

I would prefer if we generally stuck to the manpage convention:

> Name
> Command signature ("synopsis")
> Description
> Arguments

Where things get fuzzy is whether the description should contain multiple or detailed examples. I'm fine with a few, but if we start writing paragraphs where we're _advising_ on usage (rather than agnostically describing the command) it sounds like what we really need is a how-to for the command. The LXD docs provide good examples of command how-tos. It's also best if we don't foray too far into the weeds about the nuances of individual arguments within the description. That's what the _Required_ and _Options_ argument sections should be for.

That said, are there cases where we want to provide all these examples in the description because the standard or global arguments are appreciably different for certain commands? If so, then we have a deeper problem related to transclusion.

mr-cal:
@medubelko thanks, following the manpage precedent makes sense to me.  And I agree that some help messages are filling in the place of missing how-tos.  Not a problem I'd like to tackle today :see_no_evil: 

> Wouldn't this require another parser, or at least another step in the doc stack?

I think we would enhance `tools/docs/gen_cli_docs.py` rather than add another step.  Better yet, I would upstream the rst generation from that script into craft-cli. It already has a [markdown generator](https://github.com/canonical/craft-cli/blob/1b5999014c6337d55b950f84e8e9b8de3ac1ed6d/craft_cli/helptexts.py#L360).

> That said, are there cases where we want to provide all these examples in the description because the standard or global arguments are appreciably different for certain commands? 

Nothing comes to mind for snapcraft except for `snapcraft pack --output`, which is on the deprecation chopping block.

medubelko (approved):

medubelko:
Approved.

@mr-cal I left a more detailed comment about the source-parser discussion in the other PR.

PR #4985, opened 2024-08-19
build(deps): freeze requirements for docs

mr-cal:
- [x] Have you followed the [guidelines for contributing](https://github.com/snapcore/snapcraft/blob/master/CONTRIBUTING.md)?
- [x] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [x] Have you successfully run `tox run -m lint`?
- [x] Have you successfully run `tox run -e test-py310`? (supported versions: `py39`, `py310`, `py311`, `py312`)

-----

* Update `tools/freeze-requirements.sh` to freeze doc requirements
* Bump setuptools to resolve #4919, #4986
* Freeze requirements

PR #5143, opened 2024-11-13
build(deps): update deps with freeze-requirements

mr-cal:

-----

In preparation for a 8.5 release.

Follows suit of https://github.com/canonical/rockcraft/pull/731 and https://github.com/canonical/rockcraft/pull/706.

To make the changes more clear, I froze the requirements, switched to pip compile, then refroze the requirements.

This has a desirable side effect where doc dependencies are no longer in `requirement-devel.txt`.


requirements-devel.txt
```diff
@@ -1,203 +1,431 @@
...
-babel==2.16.0
-beautifulsoup4==4.12.3
+    # via snapcraft (setup.py)
 black==24.10.0
-boolean.py==4.0
-bracex==2.5
-CacheControl==0.14.0
+    # via snapcraft (setup.py)
+boolean-py==4.0
+    # via license-expression
```
mr-cal, line 18: @bepri - the culprit has been revealed

mr-cal:
The core22 spread test for `google:ubuntu-22.04-64:tests/spread/extensions/kde-neon-6` is unrelated to this PR, due to https://invent.kde.org/neon/snap-packaging/kf6-core-sdk/-/commit/2acebc089d6c353d9e79779d282063688dee90c0

requirements-docs.txt
upils, line 1: I saw there is an issue about rebasing the docs on the new starter pack but currently the docs cannot be used the same as in other projects: there is no `docs/requirements.txt` anymore so the `docs/Makefile` is not usable anymore. Is that known?

mr-cal: No - do you have any more information on that?

tools/freeze-requirements.sh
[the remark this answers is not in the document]
mr-cal: Definitely.  If I recall correctly, the writing on the wall is to use `uv` to manage requirements with a lock file.  When that happens, we should upstream it into starbase.

(cc @lengau)

requirements-docs.txt
upils: I was unable to pinpoint the commit when `docs/requirements.txt` was deleted so I do not know if that was on purpose or not :/. But rules in `docs/Makefile` work under the assumption a venv can be created from a `docs/requirements.txt` file. I see in starbase that it looks like we are moving to something managed by the main Makefile so maybe that is normal?

I noticed that because I am used to go into the `docs` folder and `make install && make serve` the doc locally to take a look at changes in docs (to make sure things render properly). I did not see any equivalent rules in the tox conf. If we do not want to rely on `docs/Makefile` maybe we should remove it to avoid confusion?

mr-cal: I moved it from `docs/requirements.txt` to `requirements-docs.txt` in https://github.com/canonical/snapcraft/pull/4985 to match rockcraft.

You should be able to use tox for the docs with `tox run -e build-docs`, `autobuild-docs`, and `lint-docs`.  Those use the top level req file here: https://github.com/canonical/snapcraft/blob/de05a733d413ba0e4d008478c200b07ff723290a/tox.ini#L157

I agree on the point of confusion. I removed the Makefile in the docs folder.

tools/freeze-requirements.sh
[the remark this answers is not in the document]
lengau: When we move fully over to `uv`, we can remove `freeze-requirements.sh` altogether, as we'll just need to run `uv sync`. Moreover, moving to `uv` will make `renovate` more useful so we'll seldom even have to run it ourselves :-)

setup.py
```diff
@@ -142,7 +142,7 @@ def recursive_data_files(directory, install_directory):
 ]
 
 docs_requires = {
-    "canonical-sphinx",
+    "canonical-sphinx[full]",
```
lengau, line 145:
```suggestion
    "canonical-sphinx>=0.2.0[full]",
```

Take it or leave it for now, but when we move to `uv` we'll need to specify minimum versions :-)

upils (approved):

Issue #5132, opened 2024-10-24
push of locally built snap fails on arm64 device


kenvandine:
### Bug Description

When attempting to push a locally built snap to the store with my qualcomm based laptop I get the following stderr:

```
FATAL ERROR: Data queue size is too large
```

### To Reproduce

snapcraft push SOMELOCALLYBUILT.snap

### Environment

snapcraft: 12827

uname: Linux xps13-qc 6.11.0-35-qcom-x1e #35-Ubuntu SMP PREEMPT_DYNAMIC Wed Oct 23 15:18:06 UTC 2024 aarch64 aarch64 aarch64 GNU/Linux

### snapcraft.yaml

```yaml
NA
```

### Relevant log output

```shell
2024-10-22 16:18:30.743 Starting snapcraft, version 8.4.3
2024-10-22 16:18:30.743 Log verbosity level set to BRIEF
2024-10-22 16:18:30.743 Preparing application...
2024-10-22 16:18:30.744 Configuring application...
2024-10-22 16:18:30.744 Setting up ConfigService
2024-10-22 16:18:30.758 Build plan: platform=None, build_for=None
2024-10-22 16:18:30.758 Running snapcraft push on host
2024-10-22 16:18:30.819 snapcraft internal error: SnapDataExtractionError('thunderbird_128.3.3esr-1_arm64.snap')
2024-10-22 16:18:30.829 Traceback (most recent call last):
2024-10-22 16:18:30.829   File "/snap/snapcraft/12827/lib/python3.10/site-packages/snapcraft_legacy/_store.py", line 58, in get_data_from_snap_file
2024-10-22 16:18:30.829     output = subprocess.check_output(
2024-10-22 16:18:30.829   File "/snap/snapcraft/12827/usr/lib/python3.10/subprocess.py", line 421, in check_output
2024-10-22 16:18:30.829     return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
2024-10-22 16:18:30.829   File "/snap/snapcraft/12827/usr/lib/python3.10/subprocess.py", line 526, in run
2024-10-22 16:18:30.829     raise CalledProcessError(retcode, process.args,
[...]

PR #5147, opened 2024-11-18
fix(unsquashfs): fix Fatal Error when using unsquashfs

sergio-costas:
When trying to push a big snap to the snap store, snapcraft fails with the error

    FATAL ERROR: Data queue size is too large

This is a bug in unsquashfs tool 4.5. But it has been fixed in 4.6.1, the current stable.

This patch builds squashfs-tools version 4.6.1 from the GIT repository, thus fixing this problem.

Fix https://github.com/canonical/snapcraft/issues/5132
Fix https://forum.snapcraft.io/t/fatal-error-data-queue-size-is-too-large/43875

- [X] Have you followed the [guidelines for contributing](https://github.com/canonical/snapcraft/blob/main/CONTRIBUTING.md)?
- [~] Have you successfully run `tox run -m lint`?

-----

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -130,7 +130,6 @@ parts:
         - python3-distutils
         - python3-pkg-resources
         - python3.10-minimal
-        - squashfs-tools
         - xdelta3
     build-attributes:
       - enable-patchelf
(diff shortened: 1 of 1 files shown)


sergio-costas:
@mr-cal I presume that yes, that would be the case, so this is just an interim solution while snapcraft isn't migrated.

sergio-costas:
@mr-cal I see that the migration to Core24 is still a draft, so it probably will require some time to be merged and published. This one, instead, fixes the problem right now. And in the Core24 migration PR, you can just remove the squashfs part...

sergio-costas:
(in fact, I'm having this problem right now with another snap... and it isn't even as big as the previous one)

mr-cal:
> @mr-cal I see that the migration to Core24 is still a draft, so it probably will require some time to be merged and published. This one, instead, fixes the problem right now. And in the Core24 migration PR, you can just remove the squashfs part after rebasing, to use again the updated .deb version...

Hi, sorry about the delay here.  I'm planning to complete the core24 migration in the next 2 weeks if our team planning goes smoothly.  With either that core24 PR or this PR, a newer version of squashfs tools will have to go into the 8.7.x release series, which won't be until 2025-Jan at the earliest.

sergio-costas:
Oh... that makes sense, yes.

I'll close this, then.

mr-cal:
Yeah, I apologize for the overlap here - there have been a bunch of moving parts in snapcraft releases recently.  I very much appreciate the PR and your testing - I wouldn't have been able to test this on my own.

kenvandine:
@mr-cal any ETA?  I'm working with someone that is hitting this problem on x86_64 with a 13M snap.

mr-cal:
@kenvandine - I wasn't aware this was on x86 as well.  The fix is targeted for Snapcraft 8.6, which is scheduled for a release to candidate on 2024-Jan-20.

kenvandine:
> @kenvandine - I wasn't aware this was on x86 as well. The fix is targeted for Snapcraft 8.6, which is scheduled for a release to candidate on 2024-Jan-20.

I hadn't seen it on x86 myself, but @Feichtmeier is experiencing it right now with 8.5.1

mr-cal:
#5136 just landed, so this should be available to test in edge later today or tomorrow, depending on launchpad's build queue

